PR #2632, opened 2026-01-28
feat(integration): add core test harness library

hubcio:
Add TestHarness builder API for integration test infrastructure. Core modules: context (test directory lifecycle), error, traits, helpers, config (server/client/mcp/connector), handle (process management), fixtures (PostgreSQL/WireMock), and seeds (test data generation).

The main goal is to decouple test infrastructure setup from test code.


core/integration/src/harness/handle/connector.rs
```diff
@@ -0,0 +1,266 @@
...
+use std::net::SocketAddr;
+use std::path::PathBuf;
+use std::process::{Child, Command, Stdio};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::sleep;
+
+const TEST_VERBOSITY_ENV_VAR: &str = "IGGY_TEST_VERBOSE";
+
+pub struct ConnectorHandle {
```
mmodzelewski: I think it would be clearer to call it `ConnectorsHandle` as in Connectors runtime`. Otherwise, it's confusing with a connector plugin handle. The same relates to `ConnectorConfig`

spetz (approved):

PR #2641, opened 2026-01-30
ci: add stale PR workflow

hubcio:
Auto-mark PRs as stale after 7 days of inactivity and close
after 14. Exempt PRs labeled 'pinned'. Issues are ignored.


slbotbm:
I feel that exempting draft prs from auto closure would be better. what do you think?

PR #2657, opened 2026-02-01
feat(server,sdk): add reserved u64 to iggy message header

spetz:
- Added u64 `reserved` field to `IggyMessageHeader` for future use
- Updated Go, C#, Java, Python, and Node.js SDKs to support new field

BREAKING


mmodzelewski (approved):

mmodzelewski (approved):

mmodzelewski (dismissed):

hubcio (changes requested):
overall OK, but since this field is explicitly documented as "reserved for future use" and all SDKs initialize it to 0,
the server should validate reserved == 0 during deserialization and reject messages where it's non-zero (just for the sake of security)

PR #2669, opened 2026-02-03
refactor(server): improve server command handlers

hubcio:
Handlers had duplicated validation and resolution logic scattered across bypass_auth variants and inline code. Entity writes occurred on any shard, requiring distributed filesystem locks that added complexity and potential for race conditions.

Centralizing execution in shard/execution.rs with pre-resolved entity types (ResolvedStream, ResolvedTopic, etc.) eliminates duplicate code paths and ensures entity file writes happen exclusively on shard0, removing the need for fs_locks entirely - now, all operation ale serialized in message_pump tasks.


core/server/src/shard/execution.rs
```diff
@@ -0,0 +1,702 @@
...
+
+    let response_data = shard
+        .metadata
+        .get_consumer_group(topic.stream_id, topic.topic_id, group_id)
+        .map(|cg| ConsumerGroupResponseData {
+            id: group_id as u32,
+            name: cg.name.clone(),
+            partitions_count: cg.partitions.len() as u32,
+        })
+        .expect("just created");
```
spetz: Should we provide a bit more meaningful error?

krishvishal (approved):

numinnex (approved):

PR #2671, opened 2026-02-03
feat(python): add message_expiry parameter to create_topic

hubcio:
The Python SDK's create_topic method lacked the ability to set message
expiration, forcing users to rely on server defaults with no override.

Add optional message_expiry parameter accepting datetime.timedelta.
When provided, converts to IggyExpiry::ExpireDuration; otherwise uses
ServerDefault to inherit from server configuration.

Also introduces version sync tooling between Cargo.toml and
pyproject.toml with CI check and pre-commit hook that auto-fixes
by promoting to the newer version + adds precommit hook for
python formatter.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,16 @@ repos:
         exclude: ^helm/charts/.*/templates/
       - id: mixed-line-ending
 
+  # Python SDK formatting
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.15.0
+    hooks:
+      - id: ruff-check
+        args: [--fix]
+        files: ^foreign/python/.*\.(py|pyi)$
+      - id: ruff-format
+        files: ^foreign/python/.*\.(py|pyi)$
+
   # CI scripts
   - repo: local
     hooks:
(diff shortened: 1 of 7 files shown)


foreign/python/src/client.rs
```diff
@@ -205,7 +211,7 @@ impl IggyClient {
                     partitions_count,
                     compression_algorithm,
                     replication_factor,
-                    IggyExpiry::NeverExpire,
+                    expiry,
                     MaxTopicSize::ServerDefault,
```
slbotbm: What about allowing the user to define MaxTopicSize as well?

.pre-commit-config.yaml
```diff
@@ -69,6 +69,20 @@ repos:
...
+        args: ["--fix"]
+        language: system
+        files: ^foreign/python/(Cargo\.toml|pyproject\.toml)$
+        pass_filenames: false
+
+      - id: python-format
+        name: python sdk format
+        entry: ruff format
+        language: system
+        files: ^foreign/python/.*\.(py|pyi)$
```
slbotbm: Instead of leaving it up to the user to download ruff, wouldn't something like 
```yaml
- repo: https://github.com/astral-sh/ruff-pre-commit
  # Ruff version.
  rev: v0.14.14
  hooks:
    - id: ruff-check
      args: [--fix ]
      files: ^foreign/python/.*\.(py|pyi)$
    - id: ruff-format
      files: ^foreign/python/.*\.(py|pyi)$
```
be better?

foreign/python/src/client.rs
hubcio: good point, fixed

.pre-commit-config.yaml
hubcio: nice, didn't know prek allows it this way. fixed

PR #2675, opened 2026-02-04
feat(metadata): impl Snapshot interface for Mux state machine

krishvishal:
The snapshot system has three layers:                                                                                                                                                       
                                                                                                                                                                                              
`Snapshotable` per-state-machine trait. Each `StreamsInner`, ` UsersInner`, `ConsumerGroupsInner` implements `to_snapshot()` / from_snapshot() to convert between in-memory state.
                                                                                                                                                                                              
`SnapshotContributor` visitor trait using the same recursive variadic tuple pattern as `StateMachine::update`. Walks (`Users`, (`Streams`, (`ConsumerGroups`, ()))) at compile time, collecting each state machine's serialized section into a `Vec<SnapshotSection>`. 
                                                                                                                                                                                              
`MetadataSnapshot` a  top-level interface for the generic snapshot type on `IggyMetadata`. Defines the full lifecycle: create (snapshot from mux) → encode (to bytes) → decode (from bytes) → restore (back to mux).


core/metadata/src/impls/metadata.rs
```diff
@@ -23,6 +27,114 @@ use journal::{Journal, JournalHandle};
...
+pub trait MetadataSnapshot: Sized {
+    /// The error type for snapshot operations.
+    type Error: std::error::Error;
+
+    /// Create a snapshot from the current state of the mux state machine.
+    ///
+    /// # Arguments
+    /// * `mux` - The multiplexing state machine containing all sub-state machines
+    /// * `commit_number` - The VSR commit number this snapshot corresponds to
+    fn create<T>(mux: &MuxStateMachine<T>, commit_number: u64) -> Result<Self, Self::Error>
```
numinnex: We impl `StateMachine` for `MuxStateMachine`, therefor I think we could accept as first argument `stm: &mut T`

core/metadata/src/impls/metadata.rs
```diff
@@ -23,6 +27,114 @@ use journal::{Journal, JournalHandle};
 use message_bus::MessageBus;
 use tracing::{debug, warn};
 
+/// Trait for metadata snapshot implementations.
+///
+/// This is the interface that `MetadataHandle::Snapshot` must satisfy.
+/// It provides methods for creating, encoding, decoding, and restoring snapshots.
+#[allow(unused)]
+pub trait MetadataSnapshot: Sized {
```
numinnex: Lets rename this to `Snapshot` trait, if there already exists one (that I created as an stub), remove it.

core/metadata/src/impls/metadata.rs
```diff
@@ -23,6 +27,114 @@ use journal::{Journal, JournalHandle};
...
+
+    /// Decode a snapshot from msgpack bytes.
+    fn decode(bytes: &[u8]) -> Result<Self, Self::Error>;
+
+    /// Restore a mux state machine from this snapshot.
+    fn restore<T>(&self) -> Result<MuxStateMachine<T>, Self::Error>
+    where
+        T: StateMachine + SnapshotContributor;
+
+    /// Get the VSR commit number this snapshot corresponds to.
```
numinnex: Maybe we could use an `sequence_number` instead of commit number, the idea is that the sequence number would be monotonically growing on each snapshot, independent of the `commit` number from consensus.

core/metadata/src/stm/consumer_group.rs
```diff
@@ -225,3 +227,150 @@ impl Handler for ConsumerGroupsInner {
         }
     }
 }
+
+/// Consumer group member snapshot representation for serialization.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
numinnex, line 239: I think we don't need those intermediate structs, we can have the `State` structs impl `Serialize` and `Deserialize` directly.

core/metadata/src/stm/mux.rs
```diff
@@ -88,7 +89,57 @@ where
     }
 }
 
+/// Recursive case for variadic tuple pattern: (Head, Tail)
```
numinnex, line 95: Hmm, I am not sure if this is the simplest way to go around it.

I imagine the flow of abstraction to be identical to one of the `StateMachine` trait,
e.g we would have an `Snapshotable` trait, that would be implemented for everything that impls `StateMachine` trait,
since `StateMachine` trait is implemented both for `MuxStateMachine` aswell as the `variadic!()` tuple of state machines, the API would be called this way:
```rs
fn snapshot_metadata(&self, ....) {
    let mut snapshot = MetadataSnapshot::default(); <-- This would impl the `Snapshot` trait.
    
    // Here is the magic, `fill_snapshot` impl for `MuxStateMachine` would proxy to the `variadic!()` tuple impl,
    // impl Snapshotable for variadic!(St, ... Rest)
    //    where St: StateMachine + FillSnapshot, 
    //          Rest: Snapshotable,
    // {
    //    fn fill_snapshot<S: Snapshot>(&self, snapshot: &mut S) {
    //        self.0.do_fill_snapshot(snapshot);
    //        self.1.fill_snapshot(snapshot);
    //    } 
    
    //    fn restore_snapshot(...) {
    //        // ... Here similar code for restoring.
    //    }
    // }
    // You can choose whatever trait name for the `FillSnapshot` trait, smth that covers filling/recovering
    // You can have associated types on the `Snapshot` trait that would be used to represent the currently filled snapshot 
    // (for example a binary blob), and use those as arguments into the `fill_snapshot` and `restore_snapshot` methods.
    // and the snapshot from which we would restore State.
    self.mux.fill_snapshot(&mut snapshot);
}

The idea is to avoid using those `const SECTION:NAME: &'str`, rather rely on the type checking to perform the walk for us.
```

core/metadata/src/stm/consumer_group.rs
```diff
@@ -225,3 +227,150 @@ impl Handler for ConsumerGroupsInner {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ConsumerGroupsSnapshot {
+    pub items: Vec<(usize, ConsumerGroupSnapshot)>,
+    pub topic_index: Vec<((usize, usize), Vec<usize>)>,
+    pub topic_name_index: Vec<((String, String), Vec<usize>)>,
+}
+
+impl Snapshotable for ConsumerGroupsInner {
```
numinnex: Impl those for `ConsumerGroup` and `Streams` and `Users`, rather for their inners. You can create an metehod on the `LeftRight`, that exposes the read handle.

core/metadata/src/stm/mux.rs
numinnex: We could alternatively look into extending the `StateMachine` trait, by adding extra bounds on it 
```rs
trait StateMachine + Snapshotable + ...
{
        
}
```

and then have single impl block.

core/metadata/src/stm/snapshot.rs
```diff
@@ -0,0 +1,336 @@
...
+        }
+    }
+}
+
+/// The outer envelope for a complete metadata snapshot.
+///
+/// Contains metadata about the snapshot and a collection of sections,
+/// where each section corresponds to one state machine's serialized state.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SnapshotEnvelope {
```
numinnex: This is a valid approach, to store an type-erased bytes inside of the snapshot, but alternative approach (the one that redpanda did choose), is for the snapshot to store an `intermediate` representation of all the `Snapshotable` states and serialize those. So rather than serializing individual states, they serialize the snapshot instead.

numinnex: I didn't think about it a lot and not sure what is the trade off space between those two approaches (I guess you need more allocations with their approach, as you have to copy the data first and then serialize), where with our approach, we just serialize the state and append it to the snapshot envelope.

core/metadata/src/impls/metadata.rs

core/metadata/src/impls/metadata.rs

core/metadata/src/impls/metadata.rs

core/metadata/src/stm/mux.rs
krishvishal: I've added type based recursion and snapshotting. Removed string based version.

core/metadata/src/stm/consumer_group.rs
krishvishal: Done.

core/metadata/src/stm/snapshot.rs
krishvishal: Removed `SnapshotEnvelope`. Now we store each snapshot struct and serialize it in one shot.

core/metadata/src/stm/consumer_group.rs
krishvishal: We can't serialize/deserialize directly because some members of these structs can't be used with `serde` directly. Thats why I had to use intermediate structs.

krishvishal:
Fixed merge conflicts.

core/metadata/src/impls/metadata.rs
```diff
@@ -24,6 +26,106 @@ use journal::{Journal, JournalHandle};
...
+    fn encode(&self) -> Result<Vec<u8>, SnapshotError> {
+        self.snapshot.encode()
+    }
+
+    fn decode(bytes: &[u8]) -> Result<Self, SnapshotError> {
+        let snapshot = MetadataSnapshot::decode(bytes)?;
+        Ok(Self { snapshot })
+    }
+
+    fn restore<T>(&self) -> Result<T, SnapshotError>
```
numinnex: Let's maybe move this method on the `state` , similarly how we have `stm.fill_snapshot`.

core/metadata/src/stm/snapshot.rs
```diff
@@ -0,0 +1,260 @@
...
+    fn fill_snapshot(&self, snapshot: &mut MetadataSnapshot) -> Result<(), SnapshotError>;
+}
+
+/// Trait for restoring state machine data from a typed snapshot.
+///
+/// Each state machine implements this to read its state from
+/// its specific field in the `MetadataSnapshot` struct.
+pub trait RestoreSnapshot: Sized {
+    /// Restore this state machine from the snapshot.
+    fn restore_snapshot(snapshot: &MetadataSnapshot) -> Result<Self, SnapshotError>;
```
numinnex: Lets make the snapshot type here be generic, instead of `MetadataSnapshot`.

core/metadata/src/stm/consumer_group.rs
numinnex: I see, which members are problematic in particular ? I think your solution is valid, but maybe in the future we could look into getting rid of the intermediary. If it's tractable we could create an issue for this.

core/metadata/src/impls/metadata.rs
```diff
@@ -24,6 +26,106 @@ use journal::{Journal, JournalHandle};
...
+    where
+        T: RestoreSnapshot,
+    {
+        T::restore_snapshot(&self.snapshot)
+    }
+
+    fn sequence_number(&self) -> u64 {
```
numinnex, line 73: use assoc type for the `sequence_number` type.

core/metadata/src/impls/metadata.rs
```diff
@@ -24,6 +26,106 @@ use journal::{Journal, JournalHandle};
...
+        self.snapshot.sequence_number
+    }
+
+    fn created_at(&self) -> u64 {
```
numinnex, line 77: same as above, use assoc type.

core/metadata/src/stm/snapshot.rs
numinnex: You can put an bound on the `RestoreSnapshot` `trait RestoreSnapshot: Snapshotable` and access the assoc type from `Snapshotable`. I think the same applies to `FillSnapshot` trait.

core/metadata/src/stm/stream.rs
```diff
@@ -466,3 +508,160 @@ impl StateHandler for DeletePartitions {
...
+                    actual: actual_id,
+                });
+            }
+            index.insert(stream_name, actual_id);
+        }
+
+        let inner = StreamsInner { index, items };
+        Ok(inner.into())
+    }
+}
```
hubcio, line 652: `from_snapshot` inserts into a fresh Slab and asserts actual_id == expected_id. This breaks if any item was deleted before snapshot: `Slab::insert` fills gaps, so restoring {0, 2} yields IDs {0, 1} → `SlabIdMismatch`. Same bug in `consumer_group.rs:320` and `user.rs:399`. I fixed same problem in `core/server/metadata`. The answer is `Slab::from_iter` with (key, value) tuples. See ` core/server/src/bootstrap.rs`

core/metadata/src/stm/mod.rs
```diff
@@ -87,6 +88,19 @@ where
...
+    T: Absorb<C>,
+{
+    pub fn read<F, R>(&self, f: F) -> R
+    where
+        F: FnOnce(&T) -> R,
+    {
+        let guard = self.read.enter().expect("read handle should be accessible");
+        f(&*guard)
+    }
+}
```
hubcio, line 102: this is duplicate of `LeftRight::with_state` from `left_right.rs:136`. Same for `snapshot_read:195` which wraps it. The `Snapshotable` impls should use the existing `with_state` API instead of adding parallel methods.

core/metadata/src/impls/metadata.rs
```diff
@@ -24,6 +26,106 @@ use journal::{Journal, JournalHandle};
...
+    fn restore<T>(&self) -> Result<T, Self::Error>
+    where
+        T: RestoreSnapshot;
+
+    /// Get the snapshot sequence number.
+    fn sequence_number(&self) -> u64;
+
+    /// Get the timestamp when this snapshot was created.
+    fn created_at(&self) -> u64;
+}
```
hubcio: this trait belongs to `snapshot` module alongside Snapshotable/FillSnapshot/RestoreSnapshot

core/metadata/src/stm/snapshot.rs
```diff
@@ -0,0 +1,260 @@
...
+    pub created_at: u64,
+    /// Monotonically increasing snapshot sequence number.
+    pub sequence_number: u64,
+    /// Users state machine snapshot data.
+    pub users: Option<UsersSnapshot>,
+    /// Streams state machine snapshot data.
+    pub streams: Option<StreamsSnapshot>,
+    /// Consumer groups state machine snapshot data.
+    pub consumer_groups: Option<ConsumerGroupsSnapshot>,
+}
```
hubcio: I think we should add `version: u32` field. What if someone will add a new node to cluster, which will have other snapshot format because he wants to update cluster one-by-one, while maintaining 100% uptime on production? This opens up possibility of implementing migrators. I'm not saying this should be done now (migrators) but version should be OK to put for now.

core/metadata/src/stm/mux.rs
```diff
@@ -91,7 +92,58 @@ where
...
+    T: StateMachine + RestoreSnapshot,
+{
+    fn restore_snapshot(snapshot: &MetadataSnapshot) -> Result<Self, SnapshotError> {
+        let inner = T::restore_snapshot(snapshot)?;
+        Ok(MuxStateMachine::new(inner))
+    }
+}
+
+#[allow(unused_imports)]
 mod tests {
```
hubcio, line 141: All roundtrip tests use empty state machines. With the slab gap bug, a test that creates items, deletes one, then snapshots+restores would have caught the issue. Add at least one populated-with-deletions roundtrip.

core/metadata/src/stm/mod.rs
krishvishal: I couldn't find `left_right.rs` on `master`. I think its only present locally on your machine.

core/metadata/src/impls/metadata.rs

core/metadata/src/stm/snapshot.rs
krishvishal: Done.

core/metadata/src/impls/metadata.rs
krishvishal: Done.

core/metadata/src/impls/metadata.rs
krishvishal: Done.

core/metadata/src/stm/mux.rs
krishvishal: Done created a new test called `roundtrip_with_slab_gaps`.

core/metadata/src/stm/snapshot.rs
krishvishal: Done. I've added a new field. But it doesn't do anything else for now. Added a TODO.

core/metadata/src/impls/metadata.rs

core/metadata/src/stm/stream.rs
krishvishal: Done.

core/metadata/src/stm/mod.rs
hubcio: you are right! instead of looking at master branch I looked at my own branch: `invert-state-handler-dep`

ignore this :)

core/metadata/src/stm/consumer_group.rs
krishvishal: Structs that are stored inside Arc, Atomic variables are not serializeable. 

```rust
pub struct Stream {
    pub id: usize,
    pub name: Arc<str>,
    pub created_at: IggyTimestamp,

    pub stats: Arc<StreamStats>, ///- not serializeable
    pub topics: Slab<Topic>, ///- Topic is not serializeable
    pub topic_index: AHashMap<Arc<str>, usize>,
}
```

hubcio: ~~maybe we can create our own type for `Arc<str>` and implement Serialize for it?~~
ignore this, I misread :D

perhaps you need to manually impl Serialize.

krishvishal: Yes, we have to manually impl Serialize for everything that's not serializeable. The intermediate struct approach felt more cleaner although seems like more work.

hubcio: wait. can you check feature `rc` for serde? https://stackoverflow.com/questions/49312600/how-do-i-serialize-or-deserialize-an-arct-in-serde

krishvishal: I feel handling `Arc<TopicStats>` in `Topic` is complicated. Because each topic shares the same `TopicStats` that we have skip the deserialization step for it.  

Currently I don't see how we can straightforwardly impl custom Serialize/Deserialize and get everything working without handling things like `Arc<TopicStats>`.  I feel this will become even more complicated than the current solution. I maybe wrong here.

spetz (approved):

PR #2676, opened 2026-02-04
ci: add pre-commit hooks for all SDKs

hubcio:
Developers can now validate code locally before pushing,
reducing CI feedback cycles. Hooks cover Go, Node.js, Java,
C#, Python, and web UI with the same tools and versions as
CI (e.g., golangci-lint v1.64.8).

Also aligns CI coverage with pre-commit: Go fmt/tidy, Python
ruff, and Node eslint now check bdd/ and examples/ dirs.
Web UI lint is now blocking since all issues are resolved.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,15 +27,15 @@ repos:
         exclude: ^helm/charts/.*/templates/
       - id: mixed-line-ending
 
-  # Python SDK formatting
+  # Python formatting
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.15.0
     hooks:
       - id: ruff-check
         args: [--fix]
-        files: ^foreign/python/.*\.(py|pyi)$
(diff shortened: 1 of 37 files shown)


.pre-commit-config.yaml
```diff
@@ -146,9 +156,51 @@ repos:
...
+        pass_filenames: false
+
+  # Java SDK linting (Checkstyle + Spotless via Gradle)
+  - repo: local
+    hooks:
+      - id: java-gradle-check
+        name: java gradle check
+        entry: bash -c 'cd foreign/java && ./gradlew check -x test'
+        language: system
+        files: ^foreign/java/.*\.(java|kt|kts)$
```
mmodzelewski: can we add entries for `bdd/java` and `examples/java` as well?

slbotbm: To extend Maciej point, adding entries for all folders examples/* would also be good.

mmodzelewski (approved):

PR #2685, opened 2026-02-05
feat(connectors): fix state & memory leak, test all plugins, enrich sinks

spetz:
Connectors runtime lost state on restart - sink connectors resumed from offset 0 instead of their last committed position. The root cause was twofold: consumer group offset queries required membership (which new runtime instances lacked), and stale members from closed TCP connections blocked partition assignment to new members.

Consumer group offset resolution now supports direct queries without membership for ConsumerGroup consumers, since offsets are stored per group+partition, not per member. Stale member cleanup during join removes members whose client connections no longer exist, enabling proper partition rebalancing.

Runtime architecture refactored to fix memory leaks - replaced centralized status manager with DashMap-based SinkManager and SourceManager. Explicit channel cleanup on shutdown prevents dangling references.

State management uses MessagePack serialization for efficient binary persistence. PostgreSQL source fixed to actually return state from poll() and moved database I/O outside mutex to eliminate lock contention.

Additional changes:
- Replaced chrono with IggyTimestamp across SDK and connectors
- Added shared API types in sdk/src/api.rs for runtime HTTP API
- Sink connectors track state, enrich with Iggy metadata, retry transient errors with exponential backoff
- Comprehensive integration tests for all connector plugins

### DIFF
--- a/core/metadata/src/stm/consumer_group.rs
+++ b/core/metadata/src/stm/consumer_group.rs
@@ -15,8 +15,8 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::stm::Handler;
-use crate::{define_state, impl_absorb};
+use crate::stm::StateHandler;
+use crate::{collect_handlers, define_state};
 use ahash::AHashMap;
 use iggy_common::create_consumer_group::CreateConsumerGroup;
 use iggy_common::delete_consumer_group::DeleteConsumerGroup;
@@ -94,10 +94,15 @@ define_state! {
         topic_index: AHashMap<(usize, usize), Vec<usize>>,
         topic_name_index: AHashMap<(Arc<str>, Arc<str>), Vec<usize>>,
         items: Slab<ConsumerGroup>,
-    },
-    [CreateConsumerGroup, DeleteConsumerGroup]
+    }
 }
-impl_absorb!(ConsumerGroupsInner, ConsumerGroupsCommand);
+
+collect_handlers! {
+    ConsumerGroups {
+        CreateConsumerGroup,
+        DeleteConsumerGroup,
+    }
+}
 
 impl ConsumerGroupsInner {
     fn resolve_consumer_group_id_by_identifiers(
@@ -106,7 +111,6 @@ impl ConsumerGroupsInner {
         topic_id: &Identifier,
         group_id: &Identifier,
     ) -> Option<usize> {
-        // Resolve by numeric IDs
         if let (Ok(s), Ok(t)) = (stream_id.get_u32_value(), topic_id.get_u32_value()) {
             let groups_in_topic = self.topic_index.get(&(s as usize, t as usize))?;
 
@@ -129,7 +133,6 @@ impl ConsumerGroupsInner {
             };
         }
 
-        // Resolve by string names
         if let (Ok(s), Ok(t)) = (stream_id.get_string_value(), topic_id.get_string_value()) {
             let key = (Arc::from(s.as_str()), Arc::from(t.as_str()));
             let groups_in_topic = self.topic_name_index.get(&key)?;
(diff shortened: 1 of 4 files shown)


mmodzelewski (dismissed):

core/common/src/lib.rs
```diff
@@ -108,5 +108,8 @@ pub use utils::personal_access_token_expiry::PersonalAccessTokenExpiry;
 pub use utils::random_id;
 pub use utils::text;
 pub use utils::timestamp::*;
+
+// Re-export chrono types for connectors and other crates that need DateTime
```
hubcio: remove this comment

core/integration/tests/connectors/fixtures/iceberg/container.rs
```diff
@@ -0,0 +1,529 @@
...
+use reqwest_retry::policies::ExponentialBackoff;
+use std::collections::HashMap;
+use testcontainers_modules::testcontainers::core::{IntoContainerPort, WaitFor};
+use testcontainers_modules::testcontainers::runners::AsyncRunner;
+use testcontainers_modules::testcontainers::{ContainerAsync, GenericImage, ImageExt};
+use tracing::info;
+use uuid::Uuid;
+
+const MINIO_IMAGE: &str = "minio/minio";
+const MINIO_TAG: &str = "latest";
```
hubcio: don't use latest, pin some version - we dont want to break our CI when they release new version

core/integration/tests/connectors/fixtures/elasticsearch/sink.rs
```diff
@@ -0,0 +1,140 @@
...
+        for _ in 0..HEALTH_CHECK_ATTEMPTS {
+            let url = format!("{}/_cluster/health", fixture.container.base_url);
+            if let Ok(response) = fixture.http_client.get(&url).send().await
+                && response.status().is_success()
+            {
+                info!("Elasticsearch cluster is healthy");
+                break;
+            }
+            sleep(Duration::from_millis(HEALTH_CHECK_INTERVAL_MS)).await;
+        }
```
hubcio, line 112: if this loop exhausts all attempts Ok(fixture) is still returned

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -1293,4 +1329,79 @@ mod tests {
...
+        };
+
+        let serialized = serde_json::to_vec(&original).expect("Failed to serialize");
+        let deserialized: State =
+            serde_json::from_slice(&serialized).expect("Failed to deserialize");
+
+        assert_eq!(original.last_poll_time, deserialized.last_poll_time);
+        assert_eq!(original.tracking_offsets, deserialized.tracking_offsets);
+        assert_eq!(original.processed_rows, deserialized.processed_rows);
+    }
```
hubcio, line 1408: i'm not fan of this test. in production we use msgpack but here we use serde_json. please validate actual state, not some shenanigans like here

core/server/src/shard/system/consumer_offsets.rs
```diff
@@ -69,16 +69,28 @@ impl IggyShard {
...
+                    return Err(IggyError::NotResolvedConsumer(consumer.id.clone()));
+                };
+                (polling_consumer, partition_id)
+            }
+            ConsumerKind::ConsumerGroup => {
+                let (_, _, cg_id) =
+                    self.resolve_consumer_group_id(stream_id, topic_id, &consumer.id)?;
+                let partition_id = partition_id.unwrap_or(0) as usize;
+                (PollingConsumer::consumer_group(cg_id, 0), partition_id)
+            }
```
hubcio, line 96: this is critical problem I think: old code enforced membership via resolve_consumer_with_partition_id → get_consumer_group_member_id, returning `ConsumerGroupMemberNotFound` for non-members. The new code bypasses this entirely for ConsumerKind::ConsumerGroup, only checking that the group exists via resolve_consumer_group_id

or am I wrong?

spetz: Yes, it's an improvement, as you should be able to read the offset without being a CG member, we could even display in on UI or so :)

hubcio: makes sense, ack

PR #2686, opened 2026-02-05
fix(server): memory leak in segment rotation

spetz:
Segment rotation accumulated memory indefinitely because sealed segments retained their 16MB index buffers and kept file writers open. Under heavy write load with frequent rotations, this caused memory to balloon from ~100MB to 20GB+.

The fix clears index buffers for sealed segments (when cache_indexes != All) and closes their writers immediately after sealing. Writers are never needed post-seal, and this also releases io_uring/kernel file handle resources.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5011,6 +5011,7 @@ dependencies = [
  "socket2 0.6.2",
  "sqlx",
  "strum",
+ "sysinfo 0.38.0",
  "tempfile",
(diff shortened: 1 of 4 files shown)


hubcio:
Nice. Please modify the segment rotation scenario to include probing for memory used by iggy-server when test ends so we will avoid this problem in future. Set it to some arbitrary value like 200MB.

PR #2698, opened 2026-02-07
feat(csharp): implement get_snapshot method in C# SDK

### LINKED ISSUE #2627: [csharp SDK] Implement get_snapshot method
The C# SDK is missing the `get_snapshot` API method. Implementation should include:

- TCP and HTTP client support
- Should be in `IIggySystem` interface (consistent with the Rust client architecture)
- Integration tests for both clients

atharvalade:

Closes #2627

## Rationale

The C# SDK is missing the get_snapshot API method, which is needed for feature parity with the Rust client.

## What changed?

The C# SDK had no way to capture system state snapshots. Both TCP and HTTP clients now lacked the GetSnapshotAsync method, and the SnapshotCompression/SystemSnapshotType enums used implicit 0-based values that didn't match the Rust binary wire protocol codes.

Added GetSnapshotAsync to the IIggySystem interface with TCP (binary serialization) and HTTP (JSON + binary ZIP response) implementations. Assigned explicit integer values to both enums to match Rust's as_code()/from_code() mappings. Added an integration test validating ZIP archive responses for both protocols.

## Local Execution

Passed, all 55 unit tests pass, SDK and integration test projects build with 0 warnings/0 errors
Pre-commit hooks not ran, prek not installed, this is a C#-only change (no Rust modifications)

## AI Usage

Claude Opus 4.6
Minimal, used for general function scaffolding and verifying binary format alignment against the Rust source
Verified by cross-referencing every enum value, byte layout, and JSON format against the Rust server source code byte-by-byte; ran full build and unit test suite locally
Yes I can explain every line

### DIFF
--- a/foreign/csharp/Benchmarks/Program.cs
+++ b/foreign/csharp/Benchmarks/Program.cs
@@ -39,7 +39,7 @@ var loggerFactory = LoggerFactory.Create(builder =>
 
 for (var i = 0; i < producerCount; i++)
 {
-    var bus = IggyClientFactory.CreateClient(new IggyClientConfigurator()
+    var bus = IggyClientFactory.CreateClient(new IggyClientConfigurator
     {
         BaseAddress = "127.0.0.1:8090",
         Protocol = Protocol.Tcp,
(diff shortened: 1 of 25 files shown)


lukaszzborek:
Yes, it's looks good. In CI was some restore problem, please rerun it

atharvalade:
Thanks both, appreciate the review. @spetz, since @lukaszzborek approved and CI is green now, could you also leave an approval so we meet the 2 write access approvals requirement?

PR #2701, opened 2026-02-08
feat(csharp): implement delete_segments method

### LINKED ISSUE #2626: [csharp SDK] Implement delete_segments method
The C# SDK is missing the `delete_segments` API method. Implementation should include:

- TCP and HTTP client support
- Separate segments interface (consistent with the Rust client architecture)
- Integration tests for both clients

atharvalade:
Add IIggySegment interface, TCP/HTTP client support, and integration tests for the delete_segments API method.

Closes #2626

## Rationale

The C# SDK was missing the `delete_segments` API method. This brings it in line with the Rust client architecture which has a separate `SegmentClient` trait.

## What changed?

The C# SDK had no way to delete segments from a partition. The Rust SDK exposes this via a dedicated `SegmentClient` trait with binary (TCP command code 503) and HTTP (`DELETE .../partitions/{pid}?segments_count=N`) support.

Added a new `IIggySegment` interface with `DeleteSegmentsAsync`, implemented it in both `TcpMessageStream` and `HttpMessageStream`, wired it into `IIggyClient`, and added integration tests (TCP only — the server doesn't expose an HTTP segments endpoint yet).

## Local Execution

- Passed, SDK, unit tests (85/85), and integration test project all build with 0 warnings/errors
- Pre-commit hooks ran, trailing whitespace, trailing newline, mixed line endings, dotnet format all passed. License headers check skipped (addlicense/Go not installed locally, but headers verified manually against existing files).

## AI Usage

1. Opus 4.6
2. Auditing binary format byte-by-byte against Rust
3. Built all projects (0 warnings/errors), ran 85 unit tests (all passed), verified binary wire format and HTTP query params match Rust SDK exactly, ran pre-commit hooks
4. Yes


foreign/csharp/Iggy_SDK/IggyClient/Implementations/HttpMessageStream.cs
```diff
@@ -559,6 +534,18 @@ var response
         }
     }
 
+    /// <inheritdoc />
+    public async Task DeleteSegmentsAsync(Identifier streamId, Identifier topicId, uint partitionId,
+        uint segmentsCount, CancellationToken token = default)
+    {
+        var response = await _httpClient.DeleteAsync(
```
lukaszzborek: I'm thinking about it, and it seems to me that since it's not supported in HTTP, it should throw a ‘FeatureUnavailableException’ exception.`. Now it can be missleading

atharvalade: Good point, will update the HTTP implementation to throw `FeatureUnavailableException` to match the pattern used by other TCP-only methods. I'll also add a test to verify the exception is thrown for HTTP.

lukaszzborek (approved):
one thing which i forgot. Please bump version in `Iggy_SDK.csproj` to `0.6.3-edge.3`

atharvalade:
> one thing which i forgot. Please bump version in `Iggy_SDK.csproj` to `0.6.3-edge.3`

done

atharvalade:
@lukaszzborek @hubcio @spetz whenever you guys get a chance, please resolve the conversations and merge if appropriate!

foreign/csharp/Iggy_SDK/IggyClient/Implementations/HttpMessageStream.cs
```diff
@@ -559,6 +559,22 @@ var response
...
+    /// <param name="topicId">The identifier of the topic containing the partition (numeric ID or name).</param>
+    /// <param name="partitionId">The unique partition ID.</param>
+    /// <param name="segmentsCount">The number of segments to delete.</param>
+    /// <param name="token">The cancellation token to cancel the operation.</param>
+    /// <returns>A task representing the asynchronous operation.</returns>
+    /// <exception cref="FeatureUnavailableException"></exception>
+    public Task DeleteSegmentsAsync(Identifier streamId, Identifier topicId, uint partitionId,
+        uint segmentsCount, CancellationToken token = default)
+    {
+        throw new FeatureUnavailableException();
```
numinnex, line 575: Is that true that we do not support this command through HTTP ? @spetz @hubcio

hubcio: it looks like it. this command is not present in `server.http`

spetz: Looks like we do support deleting the partitions, but not segments via HTTP.

numinnex (approved):

PR #2710, opened 2026-02-10
refactor(csharp): rewrite integration tests to remove DependsOn

### LINKED ISSUE #2654: Rewrite .net integration tests
Current .net integration tests use TUnit `DependsOn` attribute to order tests in steps. It can lead some problem with wrong ordering and fail in some situation.

Rewrited tests should not be dependent on other tests. Some system tests that retrieve system information or list of streams will return different data depending on when they are executed, but this is now covered by checking if attribute is greater than 0 or not empty, example `GetStats_Should_ReturnValidResponse`

atharvalade:
Remove all DependsOn attributes from 11 test files. Each test now creates its own resources with unique names via Guid.NewGuid(), ensuring full isolation and eliminating ordering dependencies.

Delete 16 unused fixture, helper, and model files that were only needed by the old DependsOn-based test structure.

Closes #2654

## Rationale

The DependsOn attribute creates ordering dependencies between tests, which leads to flaky failures and makes it hard to run tests in isolation. Independent tests are more reliable and easier to debug.

## What changed?

Tests relied on DependsOn to run in a fixed sequence, sharing resources created by earlier tests. If one test failed, all downstream tests would skip or fail regardless of their own correctness.

Each test now creates its own stream, topic, and other resources with unique Guid-based names, then cleans up or simply leaves them isolated. The 16 fixture, helper, and model files that only existed to support the old shared-state pattern were removed as dead code.

## Local Execution

- Passed
- Pre-commit hooks not ran (non-Rust changes only, prek targets Rust code)

## AI Usage

1. Opus 4.6
2. Used for repetitive boilerplate and verifying no dangling references after deletions
3. Verified by building locally (0 warnings, 0 errors) and running dotnet format
4. Yes

### DIFF
--- a/.github/actions/python-maturin/pre-merge/action.yml
+++ b/.github/actions/python-maturin/pre-merge/action.yml
@@ -29,34 +29,29 @@ runs:
     - name: Setup Python
       uses: actions/setup-python@v5
       with:
-        python-version: "3.11"
+        python-version: "3.10"
 
     - name: Setup Rust with cache
       uses: ./.github/actions/utils/setup-rust-with-cache
 
-    - name: Cache pip
+    - name: Install uv
+      uses: astral-sh/setup-uv@v7
+
+    - name: Cache uv
       uses: actions/cache@v4
       with:
-        path: ~/.cache/pip
-        key: pip-${{ runner.os }}-${{ hashFiles('foreign/python/pyproject.toml') }}
+        path: ~/.cache/uv
+        key: uv-${{ runner.os }}-${{ hashFiles('examples/python/uv.lock', 'foreign/python/uv.lock', 'bdd/python/uv.lock') }}
 
     - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip wheel setuptools
-
-        # Install maturin for building
-        pip install 'maturin>=1.2,<2.0'
-
-        # Install testing dependencies from pyproject.toml
-        cd foreign/python
-        pip install -e ".[testing,dev]"
-        cd ../..
-
-        # Install additional CI dependencies for Docker testing if needed
-        if [ "${{ inputs.task }}" = "test" ]; then
-          pip install 'testcontainers[docker]>=3.7.0,<5.0' || true
-        fi
       shell: bash
+      run: |
+        cd foreign/python
+        if [ "${{ inputs.task }}" = "test" ]; then
+          uv sync --frozen --extra dev --extra testing --extra testing-docker
+        else
+          uv sync --frozen --extra dev --extra testing
+        fi
 
     - name: Lint and format check
       if: inputs.task == 'lint'
@@ -71,19 +66,21 @@ runs:
         [ -d "$DIR_BDD" ] && DIRS_TO_CHECK="$DIRS_TO_CHECK $DIR_BDD"
         [ -d "$DIR_EXAMPLES" ] && DIRS_TO_CHECK="$DIRS_TO_CHECK $DIR_EXAMPLES"
 
+        cd foreign/python
+
         echo "Directories to check: $DIRS_TO_CHECK"
-        echo "ruff version: $(ruff --version)"
+        echo "ruff version: $(uv run ruff --version)"
 
         echo "Running ruff check..."
-        ruff check $DIRS_TO_CHECK
+        uv run ruff check $DIRS_TO_CHECK
 
         echo "Running ruff format --check..."
-        ruff format --check $DIRS_TO_CHECK
+        uv run ruff format --check $DIRS_TO_CHECK
 
         # mypy only for the SDK (has type stubs)
         echo "Running mypy on SDK..."
-        mypy --explicit-package-bases "$DIR_SDK"
-        echo "mypy version: $(mypy --version)"
+        uv run mypy --explicit-package-bases "$DIR_SDK"
+        echo "mypy version: $(uv run mypy --version)"
       shell: bash
 
     - name: Build Python wheel for testing
@@ -93,12 +90,12 @@ runs:
 
         # Build the module
         echo "Building Python wheel..."
-        maturin build -o dist
+        uv run maturin build -o dist
 
         if [ "${{ inputs.task }}" = "test" ]; then
           # Install the built wheel for testing
           echo "Installing built wheel..."
-          pip install dist/*.whl --force-reinstall
+          uv pip install dist/*.whl --force-reinstall
         fi
 
         if [ "${{ inputs.task }}" = "build" ]; then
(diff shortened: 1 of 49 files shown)


foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs
```diff
@@ -21,157 +21,183 @@
...
+    private const string GroupName = "test_consumer_group";
 
-    [ClassDataSource<ConsumerGroupFixture>(Shared = SharedType.PerClass)]
-    public required ConsumerGroupFixture Fixture { get; init; }
+    [ClassDataSource<IggyServerFixture>(Shared = SharedType.PerAssembly)]
+    public required IggyServerFixture Fixture { get; init; }
+
+    private async Task<(IIggyClient client, string streamName)> CreateStreamAndTopic(Protocol protocol)
+    {
+        var client = protocol == Protocol.Tcp
```
lukaszzborek: I think, this client creating can be extracted into IggyServerFixture, and then in test just run for example `Fixuture.CreateClient(protocol)`. And use it in every test

foreign/csharp/Iggy_SDK.Tests.Integration/SystemTests.cs
```diff
@@ -88,64 +93,62 @@ public async Task GetMe_Tcp_Should_Return_MyClient(Protocol protocol)
...
-        var response = await Fixture.Clients[protocol].GetClientByIdAsync(me!.ClientId);
+        var response = await client.GetClientByIdAsync(me!.ClientId);
         response.ShouldNotBeNull();
-        response.UserId.ShouldBe(0u);
         response.Address.ShouldNotBeNullOrEmpty();
         response.Transport.ShouldBe(Protocol.Tcp);
         response.ConsumerGroupsCount.ShouldBe(1);
         response.ConsumerGroups.ShouldNotBeEmpty();
         response.ConsumerGroups.ShouldContain(x => x.GroupId == consumerGroup!.Id);
-        response.ConsumerGroups.ShouldContain(x => x.TopicId == 0);
```
lukaszzborek, line 124: maybe in this tests, we should add second topic to stream and use this second topic in consumer group. In this way we can check TopicId == 1 or > 0

foreign/csharp/Iggy_SDK.Tests.Integration/UsersTests.cs
```diff
@@ -156,31 +173,47 @@ await Should.NotThrowAsync(Fixture.Clients[protocol]
...
-        await Should.NotThrowAsync(Fixture.Clients[protocol]
-            .ChangePassword(Identifier.String(Username.GetWithProtocol(protocol)), "test_password_1", "user2"));
+        var client = protocol == Protocol.Tcp
+            ? await Fixture.CreateTcpClient()
+            : await Fixture.CreateHttpClient();
+
+        var username = $"chpw-{Guid.NewGuid():N}"[..20];
+        await client.CreateUser(username, "old_password", UserStatus.Active);
+
+        await Should.NotThrowAsync(client.ChangePassword(Identifier.String(username), "old_password", "new_password"));
```
lukaszzborek, line 172: In this test we should login with new password to check if password was changed

foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs
```diff
@@ -21,157 +21,183 @@
...
     [SkipTcp]
-    [DependsOn(nameof(GetConsumerGroups_Should_Return_ValidResponse))]
     [MethodDataSource<IggyServerFixture>(nameof(IggyServerFixture.ProtocolData))]
     public async Task JoinConsumerGroup_Http_Should_Throw_FeatureUnavailable(Protocol protocol)
     {
+        var (client, streamName) = await CreateStreamAndTopic(protocol);
+        var cg = await client.CreateConsumerGroupAsync(Identifier.String(streamName),
+            Identifier.String(TopicName), GroupName);
+
         await Should.ThrowAsync<FeatureUnavailableException>(() =>
```
lukaszzborek, line 145: After that we should run for example `GetMe` on client to check if client is joined to consumer group on this stream and topic (`ConsumerGroups` field in response)

foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs
```diff
@@ -21,157 +21,183 @@
...
     [MethodDataSource<IggyServerFixture>(nameof(IggyServerFixture.ProtocolData))]
     public async Task LeaveConsumerGroup_Tcp_Should_LeaveConsumerGroup_Successfully(Protocol protocol)
     {
+        var (client, streamName) = await CreateStreamAndTopic(protocol);
+        var cg = await client.CreateConsumerGroupAsync(Identifier.String(streamName),
+            Identifier.String(TopicName), GroupName);
+        await client.JoinConsumerGroupAsync(Identifier.String(streamName),
+            Identifier.String(TopicName), Identifier.Numeric(cg!.Id));
+
         await Should.NotThrowAsync(() =>
```
lukaszzborek, line 161: Like in `JoinConsumerGroupAsync` but check if not in list of joined cg

foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs
atharvalade: Done. Added `CreateAuthenticatedClient`(Protocol protocol) to `IggyServerFixture` and replaced the protocol == Protocol.Tcp ? CreateTcpClient() : CreateHttpClient() pattern across all 11 test files.

foreign/csharp/Iggy_SDK.Tests.Integration/SystemTests.cs
atharvalade: Done. `SystemTests.GetClient_WithConsumerGroup` now creates "`first_topic`" and "`second_topic`", uses the second topic for the consumer group, and `asserts x.TopicId == (int)secondTopic!.Id`

foreign/csharp/Iggy_SDK.Tests.Integration/UsersTests.cs
atharvalade: Done. `UsersTests.ChangePassword` now creates a new client, logs in with "`new_password`", and asserts the response is valid

foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs
atharvalade: Done. `JoinConsumerGroup_Tcp` now calls `GetMeAsync`() after joining and verifies `ConsumerGroupsCount` == 1 and the group ID is in the list

foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs
atharvalade: Done. `LeaveConsumerGroup_Tcp` now calls `GetMeAsync`() after leaving and verifies `ConsumerGroupsCount` == 0 and the group ID is NOT in the list

spetz (approved):

PR #2714, opened 2026-02-10
refactor(metadata): add typed return values to state handlers

hubcio:
mod.rs mixed traits, unsafe left-right infra, and code-gen macros in a single 400-line file, making each concern hard to review independently. StateHandler also returned void, preventing commands from producing results through the left-right layer.

Split into focused modules (traits.rs, left_right.rs, macros.rs) and introduce CommandEnvelope to smuggle typed return values out of Absorb's void interface. StateHandler and Command now carry an Output associated type. WriteCell gains a debug-mode reentrancy guard and LeftRight exposes with_state() for lock-free reads.


core/metadata/src/stm/mod.rs
```diff
@@ -20,76 +20,138 @@ pub mod mux;
...
+    type Input;
+    type Output;
+
+    fn parse(input: Self::Input) -> Result<Self::Cmd, Self::Input>;
+}
+
+/// Public interface for state machines.
+/// Returns `Ok(output)` if applicable, `Err(input)` to pass ownership back.
+pub trait State {
+    type Output;
```
numinnex: This output should be generic over the command response type, but type erased.

core/metadata/src/stm/mod.rs
```diff
@@ -20,76 +20,138 @@ pub mod mux;
...
+    type Input;
+
+    fn apply(&self, input: Self::Input) -> Result<Self::Output, Self::Input>;
```
numinnex: apply<T> -> Result<Self::Output<T>, Self::Inptu>;

hubcio: as we previously discussed, it'll stay as is

core/metadata/src/stm/mod.rs
hubcio: as we previously discussed, it'll stay as is

PR #2723, opened 2026-02-11
fix(server): prevent stale segment indices across await points

hubcio:
The message_cleaner can remove front segments from the log
vec while other operations yield at .await points on the
same compio thread. Code that captured a segment vec index
before an await and used it after would access the wrong
segment or panic on out-of-bounds.

Three fixes applied consistently:

1. delete_segments_base: pre-create the replacement segment
   before draining, then drain + add in a single borrow_mut
   so the log is never empty across awaits (prevents "active
   segment called on empty log" panic). Removes the now-
   unused init_log_in_local_partitions.

2. append_messages / rotate_segment: recompute segment_index
   after each await instead of caching it before.

3. Polling (load_messages_from_disk, by_timestamp): replace
   pre-computed segment_range with per-iteration re-resolve
   by start_offset (stable identity). Extract shared lookup
   into resolve_segment_storage helper that clones Rc readers
   so fds survive concurrent segment deletion.

This issue is not reproducible in normal tests. Benchmark
stress test will be implemented in next PR.

### DIFF
--- a/core/server/src/shard/system/messages.rs
+++ b/core/server/src/shard/system/messages.rs
@@ -458,13 +458,7 @@ impl IggyShard {
 
         let _write_guard = write_lock.lock().await;
 
-        let (
-            current_offset,
-            current_position,
-            segment_start_offset,
-            segment_index,
-            message_deduplicator,
-        ) = {
+        let (current_offset, current_position, segment_start_offset, message_deduplicator) = {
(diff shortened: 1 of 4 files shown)


hubcio:
This is extreme spaghetti and mess. I cannot wrap my mind around this code. I will serialize writing in https://github.com/apache/iggy/pull/2669. Closing this.

PR #2735, opened 2026-02-13
feat(partitions): impl first version of partitions module

numinnex:


core/partitions/src/iggy_partitions.rs
```diff
@@ -116,4 +180,873 @@ impl IggyPartitions {
...
+            return;
+        }
+
+        // If we are a follower, we advance the commit number.
+        if consensus.is_follower() {
+            consensus.advance_commit_number(message.header().commit);
+        }
+
+        //
+        assert_eq!(header.op, current_op + 1);
```
hubcio: i think this should be debug_assert, not actual assert in production

hubcio (approved):

PR #2737, opened 2026-02-14
refactor(go): Introduce Command interface to improve Go client request handling

chengxilo:

Closes # N/A

---

## Rationale

Currently, the Go client performs web requests by calling:

```go
func (c *IggyTcpClient) sendAndFetchResponse(message []byte, command iggcon.CommandCode) ([]byte, error)
```

While this works, it tightly couples request serialization and command code handling at the call site, making the API less expressive and harder to extend.

---

## What changed?

Previously, each web request manually passed serialized bytes together with a `CommandCode` into `sendAndFetchResponse`, requiring callers to manage both serialization and command metadata.

Now, we use:

```go
func (c *IggyTcpClient) do(cmd iggcon.Command) ([]byte, error)
```

Request types implement a new `Command` interface that:

- Supports binary serialization (Implement `MarshalBinary` method)
- Provides a `Code()` method to return the associated command code  

This moves command metadata and serialization logic into the request type itself, improving encapsulation, readability, and extensibility.

**`Command` interface and `do` function are already implemented in previous commit**

---

## Local Execution

- Passed locally  
- Pre-commit hooks ran  

---

## AI Usage

- Tool: GitHub Copilot  and ChatGPT
- Scope: Copilot is for autocomplete assistance only. I use ChatGPT to help with writing(Yeah this pr)  
- Verification: All changes were compiled and tested locally  
- Yes, I can explain every line of the implementation if needed.


foreign/go/contracts/consumer_groups.go
```diff
@@ -35,30 +35,103 @@ type ConsumerGroupMember struct {
...
+}
+
+func (l *LeaveConsumerGroup) MarshalBinary() ([]byte, error) {
+	return marshalIdentifiers(l.StreamId, l.TopicId, l.ConsumerGroupId)
+}
+
+type GetConsumerGroup struct {
+	StreamId Identifier
+	TopicId  Identifier
+	GroupId  Identifier
```
ex172000: The name should be consistent with line 83, also do we want to have the JSON annotation?

foreign/go/contracts/consumer_groups.go
```diff
@@ -35,30 +35,103 @@ type ConsumerGroupMember struct {
...
+}
+
+func (d *DeleteConsumerGroup) MarshalBinary() ([]byte, error) {
+	return marshalIdentifiers(d.StreamId, d.TopicId, d.ConsumerGroupId)
+}
+
+type JoinConsumerGroup struct {
 	StreamId        Identifier `json:"streamId"`
 	TopicId         Identifier `json:"topicId"`
 	ConsumerGroupId Identifier `json:"consumerGroupId"`
```
ex172000: This structure seems repeating for many CG operations, can we have a common struct ?

foreign/go/contracts/change_password.go
```diff
@@ -0,0 +1,52 @@
...
+	UserID          Identifier `json:"-"`
+	CurrentPassword string     `json:"CurrentPassword"`
+	NewPassword     string     `json:"NewPassword"`
+}
+
+func (c *ChangePassword) Code() CommandCode {
+	return ChangePasswordCode
+}
+
+func (c *ChangePassword) MarshalBinary() ([]byte, error) {
```
ex172000, line 30: Looks quite some operations here, appreciate if we can add a test

foreign/go/contracts/create_user.go
```diff
@@ -0,0 +1,77 @@
...
+	Password    string       `json:"Password"`
+	Status      UserStatus   `json:"Status"`
+	Permissions *Permissions `json:"Permissions,omitempty"`
+}
+
+func (c *CreateUser) Code() CommandCode {
+	return CreateUserCode
+}
+
+func (c *CreateUser) MarshalBinary() ([]byte, error) {
```
ex172000, line 33: Same here, appreciated if we can have some test

foreign/go/contracts/consumer_groups.go
chengxilo: The JSON annotation was left by the previous developers and actually has no use at all. Maybe we can remove them in the future.
That said, I will modify the attribute name for consistency later.

foreign/go/contracts/consumer_groups.go
chengxilo: Regarding the common struct. If I add a common strust like this:
```go
type ConsumerGroupCommandBase struct {
	StreamId Identifier
	TopicId  Identifier
	GroupId  Identifier
}
```
`GetConsumerGroup` looks like this
```go
type GetConsumerGroup struct {
	ConsumerGroupCommandBase
}
```

And when I use it, it would be like:
```go
buffer, err := c.do(&iggcon.GetConsumerGroup{
		ConsumerGroupCommandBase: iggcon.ConsumerGroupCommandBase{
			StreamId: streamId,
			TopicId:  topicId,
			GroupId:  groupId,
		},
	})
```

To be honest, I prefer the approach without common struct personally, when using it, it looks more straightforward.
```
buffer, err := c.do(&iggcon.GetConsumerGroup{
		StreamId: streamId,
		TopicId:  topicId,
		GroupId:  groupId,
	})
```
Just personal preference. What do you think?

chengxilo:
@ex172000 regarding the request for tests and the common struct:

I actually hesitated on this because the current foreign/go/contract directory is already quite cluttered, with types for various purposes mixed together. 

To avoid suffering in the cluttered types, I believe we should reorganize the directory structure in foreign/go to better categorize these types. However, if I include that reorganization and the new tests in this PR, the "lines changed" will explode, making this a >3000 lines of change(or even more, I don't know).

My proposal:

1. I will fix the naming consistency for `ConsumerGroupId`/`GroupId` now. And probably introduce the common struct.
2. We merge this PR.
3. After that, I will submit some follow-up PR focused solely on reorganizing the directory and adding the unit tests you suggested.

I think it would make the review process easier. What do you think?

ex172000 (approved):
I don't have write permission then I guess this review won't count.
Hope @hubcio can bring someone to review Golang changes

chengxilo:
> I don't have write permission then I guess this review won't count. Hope @hubcio can bring someone to review Golang changes

Thanks. It's ok, I have to solve the conflicts first anyway.🤣

PR #2754, opened 2026-02-17
feat(python): add TLS and auth args to getting-started examples

saie-ch:
The Python getting-started examples connection settings has no way to configure TLS, authentication credentials.

What changed?

The producer and consumer examples only supported a basic --tcp-server-address argument.

Both examples now accept --tls, --tls-ca-file, --username, and --password arguments, and use IggyClient.from_connection_string() to build a connection string with optional TLS parameters.

  Local Execution : Passed
  Pre-commit hooks ran
  
- Tested locally against iggy server on 127.0.0.1:8090 without TLS:
  python examples/python/getting-started/producer.py --tcp-server-address 127.0.0.1:8090
  python examples/python/getting-started/consumer.py --tcp-server-address 127.0.0.1:8090
  Producer sent 5 batches of 10 messages (50 total), consumer consumed all 50 messages successfully.
  
- Tested on manged server with --tls , --tls-ca-file, username, password.
   examples/python/getting-started/producer.py \
  --tls \
  --tcp-server-address xxxxxxxx \
  --username xxxx \
  --password xxxxxx \
  --tls-ca-file xxxxx    

- Python SDK tests: 18/18 passed

### DIFF
--- a/core/bench/Cargo.toml
+++ b/core/bench/Cargo.toml
@@ -57,4 +57,4 @@ uuid = { workspace = true }
 [lints.clippy]
 enum_glob_use = "deny"
 pedantic = "deny"
-nursery = "deny"
+nursery = "warn"
--- a/core/message_bus/Cargo.toml
+++ b/core/message_bus/Cargo.toml
@@ -34,4 +34,4 @@ rand = { workspace = true }
 [lints.clippy]
 enum_glob_use = "deny"
 pedantic = "deny"
-nursery = "deny"
+nursery = "warn"
--- a/core/metadata/Cargo.toml
+++ b/core/metadata/Cargo.toml
@@ -40,3 +40,8 @@ rmp-serde = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 slab = { workspace = true }
 tracing = { workspace = true }
+
+[lints.clippy]
+enum_glob_use = "deny"
+pedantic = "deny"
+nursery = "warn"
--- a/core/metadata/src/impls/metadata.rs
+++ b/core/metadata/src/impls/metadata.rs
@@ -41,13 +41,15 @@ pub struct IggySnapshot {
 
 #[allow(unused)]
 impl IggySnapshot {
+    #[must_use]
     pub fn new(sequence_number: u64) -> Self {
         Self {
             snapshot: MetadataSnapshot::new(sequence_number),
         }
     }
 
-    pub fn snapshot(&self) -> &MetadataSnapshot {
+    #[must_use]
+    pub const fn snapshot(&self) -> &MetadataSnapshot {
         &self.snapshot
     }
 }
@@ -99,6 +101,7 @@ pub struct IggyMetadata<C, J, S, M> {
     pub mux_stm: M,
 }
 
+#[allow(clippy::future_not_send)]
 impl<B, J, S, M> Plane<VsrConsensus<B>> for IggyMetadata<VsrConsensus<B>, J, S, M>
 where
     B: MessageBus<Replica = u8, Data = Message<GenericHeader>, Client = u128>,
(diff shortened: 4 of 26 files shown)


examples/python/getting-started/consumer.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
kparisa: @saie-ch we should remove this as ASF license should be at the top of each file

examples/python/getting-started/producer.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
kparisa: @saie-ch we should remove this as ASF license should be at the top of each file

slbotbm:
I thought about it, and instead of testing tls using the examples, it would be better to test as an integration test. @saie-ch would you be amenable to writing an integration test that will test tls?

saie-ch:
@slbotbm  Sure, I can write an integration test for TLS.

scripts/run-python-examples-from-readme.sh
slbotbm, line 1: @saie-ch Thank you for the update! One comment: We would like to run these tests as part of the "test" task for python, not as part of the "examples" task. I'd appreciate if you could wire this code into the test task.

saie-ch: @slbotbm  could you recheck? The test_tls.py test already runs as part of the sdk-python / test task and the shell script TLS pass is a separate, lightweight check to ensure the example scripts (producer.py/consumer.py) work with the --tls flags. 

Please correct me if I'm mistaken, thanks!

slbotbm: @saie-ch you are right -- that was my oversight. I apologize. 

I'd like to request you to please fix the CI. After that, I think we are good to go.

foreign/python/pyproject.toml
```diff
@@ -125,3 +125,6 @@ filterwarnings = [
     "ignore::DeprecationWarning",
     "ignore::pytest.PytestUnraisableExceptionWarning",
 ]
+
+[tool.mypy]
+ignore_missing_imports = true
```
hubcio: what's the reason for this change?

saie-ch: During the lint job, testcontainers is not available during the mypy check because Docker(`uv sync --frozen --extra dev --extra testing`) is not installed unless the task is test. This caused mypy to fail with import-not-found.

we can add --extra testing-docker to the lint install step in `.github/actions/python-maturin/pre-merge/action.yml`  and check it once — happy to go with whichever approach you prefer.

examples/python/getting-started/consumer.py
```diff
@@ -57,17 +59,67 @@ def parse_args():
...
+    connection_string = build_connection_string(args)
+    logger.info(f"Connection string: {connection_string}")
+
+    client = IggyClient.from_connection_string(connection_string)
     try:
         logger.info("Connecting to IggyClient...")
         await client.connect()
         logger.info("Connected. Logging in user...")
-        await client.login_user("iggy", "iggy")
+        await client.login_user(args.username, args.password)
```
mmodzelewski: if the credentials are provided in the connection string, then auto_login is enabled, so calling login is no longer needed as a separate step

foreign/python/tests/test_tls.py
```diff
@@ -0,0 +1,187 @@
...
+        """Test that connecting without TLS to a TLS-enabled server fails."""
+        host = "localhost"
+        port = tls_container.get_exposed_port(CONTAINER_TCP_PORT)
+
+        wait_for_server(host, int(port))
+
+        # Connect without TLS to a TLS-enabled server
+        # IggyClient constructor requires IP address, not hostname
+        client = IggyClient(f"127.0.0.1:{port}")
+        await client.connect()
```
mmodzelewski: In this case, doesn't connect already throw an error?
Also, I think the client set up should be as close as possible (we just want to test the difference tls/no tls), so you should use connection string, but without the tls part.

foreign/python/pyproject.toml
mmodzelewski: Yes, please update the installation, so the `testing-docker` is added for both `test` and `lint` tasks. If we enable ignoring missing imports, we might miss something in the future.

mmodzelewski (changes requested):

PR #2765, opened 2026-02-18
feat(consensus): enable independent commit progress across namespaces

hubcio:
LocalPipeline's single VecDeque blocked independent commit
progress across namespaces. NamespacedPipeline uses per-namespace
VecDeques under one global op sequence and hash chain, so
drain_committable_all drains each namespace independently.

Additional fixes: wire last_prepare_checksum through Project
and on_replicate instead of hardcoded 0; walk consecutive ops
in ack_quorum_reached to prevent premature drain; propagate
namespace in build_reply_message; enforce namespace registration
before push; add namespace field to view change headers; replace
magic-number dispatch with Operation enum in simulator.


core/consensus/src/namespaced_pipeline.rs
```diff
@@ -0,0 +1,540 @@
...
+            );
+        }
+
+        queue.push_back(PipelineEntry::new(header));
+        self.total_count += 1;
+        self.last_push_checksum = header.checksum;
+        self.last_push_op = header.op;
+    }
+
+    fn pop_message(&mut self) -> Option<Self::Entry> {
```
numinnex: We need to rework this method from the interface, pop_message, needs to take an `op_number` and based on that op_number we would search through the partition queue of ops.

numinnex (approved):

PR #2768, opened 2026-02-18
feat(sdk): support hostnames in QUIC and WebSocket clients

### LINKED ISSUE #2729: Support hostnames in iggy-bench
Verify that iggy-bench accepts hostnames like `localhost:8090` instead of just IP addresses. If it doesn't work, fix the validation or DNS resolution.

**What to focus on:**
- Test current behavior first — the issue might not exist
- Check SDK client configs for overly strict address validation
- If broken, fix DNS resolution rather than hardcoding IP-only validation

Done when `iggy-bench pp tcp --server-address localhost:8090` works and a test confirms it.

**Open questions:**
- Does the issue actually exist?
- Is the problem in bench argument parsing or SDK client connection logic?

Nice to have: instead of server addr + port, make `iggy-bench` accept connection string, just like `iggy` CLI.

seokjin0414:
Closes #2729

## Summary
- Support hostnames in QUIC and WebSocket clients via `tokio::net::lookup_host()`
- QUIC `create()`: defer `SocketAddr` parsing, store raw address string
- QUIC/WebSocket `connect()`: replace `parse::<SocketAddr>()` with `lookup_host()`
- Remove `localhost` → `127.0.0.1` workaround in CLI args

## Tests
- QUIC: `create()` with hostname and FQDN succeeds
- QUIC: raw hostname stored in `current_server_address`
- QUIC/WebSocket: `from_connection_string` with hostname succeeds

### DIFF
--- a/core/sdk/src/quic/quic_client.rs
+++ b/core/sdk/src/quic/quic_client.rs
@@ -34,7 +34,7 @@ use iggy_common::{
 use quinn::crypto::rustls::QuicClientConfig as QuinnQuicClientConfig;
 use quinn::{ClientConfig, Connection, Endpoint, IdleTimeout, RecvStream, VarInt};
 use rustls::crypto::CryptoProvider;
-use std::net::SocketAddr;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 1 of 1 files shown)


core/sdk/src/quic/quic_client.rs
```diff
@@ -165,14 +165,9 @@ impl QuicClient {
...
-                error!("Invalid server address: {error}");
-                IggyError::InvalidServerAddress
-            })?;
-        let client_address = if server_address.is_ipv6()
+        let resolved_addr = config.server_address.parse::<SocketAddr>().ok();
+
+        let client_address = if resolved_addr.is_some_and(|a| a.is_ipv6())
             && config.client_address == QuicClientConfig::default().client_address
         {
             "[::1]:0"
```
hubcio, line 177: the IPv6 client endpoint detection is broken for hostnames.

parse::<SocketAddr>().ok() returns None for any hostname, so
is_some_and(|a| a.is_ipv6()) is always false and the endpoint
always binds to 0.0.0.0:0 (IPv4). if the hostname later
resolves to an IPv6 address in connect(), the IPv4-bound UDP
socket can't reach it.

and create() is sync 🥲

seokjin0414: Thanks for the review
Fixed — changed parse::<SocketAddr>().ok() to to_socket_addrs().ok() which does blocking DNS in create(), so IPv6 hostnames are now correctly detected.
Making create() async is beyond scope of this PR 🥲

PR #2769, opened 2026-02-18
feat(simulator): implement PacketSimulator for deterministic simulator

krishvishal:
Summary:
- Adds a `PacketSimulator` that models a full network fault-injection layer: per-path latency (exponential distribution), packet loss, link capacity limits with random eviction, automatic network partitioning with configurable lifecycle, and per-path clogging.
- Adds a `ReadyQueue` — a min-heap priority queue with reservoir-sampled random ready removal, used to order packets by delivery tick and uniformly select among simultaneously ready items.
-   Partition modes: `UniformSize`, `UniformPartition`, `IsolateSingle`,  all guarantee at least one node in each partition. Supports symmetric and asymmetric partitions.

Note:
- Pull-based batch delivery: `step()` returns all ready packets for the current tick in a single Vec. Packets delivered in the same tick cannot trigger chain reactions within that tick. This was done because its difficult to do self-referential structs in Rust and even with using dynamic dispatch we have to fight the borrow checker since we would need mutable access to different fields at the same time.

### DIFF
--- a/core/simulator/Cargo.toml
+++ b/core/simulator/Cargo.toml
@@ -33,5 +33,5 @@ metadata = { path = "../metadata" }
 partitions = { path = "../partitions" }
 rand = { workspace = true }
 rand_xoshiro = { workspace = true }
-tracing = { workspace = true }
 shard = { path = "../shard" }
+tracing = { workspace = true }


core/simulator/src/network.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub fn recycle_buffer(&mut self, buf: Vec<Packet>) {
+        self.simulator.recycle_buffer(buf);
+    }
+
+    /// Advance network time by one tick.
+    ///
+    /// This should be called once per simulation tick, after all ready
+    /// packets have been delivered. Handles automatic partition lifecycle
+    /// and random path clogging.
+    pub fn tick(&mut self) {
```
numinnex, line 78: I've recon in our case the `tick` method will be probably replaced by an async function that will Poll an `step` Future and update the state of the network simulator and return an `Message` back.

So for example we would have our tcp listener, which on accepted connection returns an `Sender`, which has async method `recv_message`, that Sender in simulation would be replaced with something that on `recv_message().await` would use the network simulator to receive it's message, by `ticking` the network simulator.

numinnex: Just leaving this comment in there for an reference point of how this will get integrated later on with the simulated server.

PR #2781, opened 2026-02-19
feat(connectors): restart connector with new config without runtime restart

### LINKED ISSUE #2417: Connectors Runtime - restart connector with new configuration without restarting the whole runtime
Currently, it is possible to create a new connector configuration and then change it to active with the connectors runtime API. The issue is that the config changes will be applied only after the runtime is restarted. 

It would be great to have a mechanism that applies the configuration changes without restarting the whole runtime. Instead, only the connector that was updated should get restarted.

seokjin0414:
## Summary

Closes #2417.

- Add `POST /sinks/{key}/restart` and `POST /sources/{key}/restart` API endpoints
- Implement `stop_connector()`, `start_connector()`, `restart_connector()` on SinkManager and SourceManager
- Add `tokio::sync::watch` shutdown channel to sink consume loop for graceful cancellation
- Store `Arc<Container>` and task handles in Manager for lifecycle control
- Extend RuntimeContext with `iggy_clients` and `state_path` for restart resource access
- Replace direct shutdown in main.rs with Manager-based `stop_connector()` calls

## Design

- **Sink restart**: shutdown signal via watch channel → await task handles (5s timeout) → FFI `iggy_sink_close` → FFI `iggy_sink_open` with new config → spawn new consume tasks
- **Source restart**: cleanup flume sender → await forwarding tasks (5s timeout) → FFI `iggy_source_close` → load state → FFI `iggy_source_open` with new config → spawn new handler
- Container (shared library) is NOT reloaded — only closed/reopened with new config and plugin ID


core/connectors/runtime/src/manager/sink.rs
```diff
@@ -96,6 +107,154 @@ impl SinkManager {
...
+            details.config = config.clone();
+            details.shutdown_tx = Some(shutdown_tx);
+            details.task_handles = task_handles;
+            metrics.increment_sinks_running();
+        }
+
+        Ok(())
+    }
+
+    pub async fn restart_connector(
```
mmodzelewski, line 235: I think we do need some concurrency guard. If the runtime receives multiple restart requests for the same connector, it might end up in an inconsistent state.

core/connectors/runtime/src/sink.rs
```diff
@@ -377,6 +404,63 @@ fn init_sink(
     }
 }
 
+pub(crate) async fn setup_sink_consumers(
```
mmodzelewski, line 380: There's a lot of duplication with the `init` function. Do you think we could unify the logic here and use a common setup then?

core/connectors/runtime/src/manager/sink.rs
seokjin0414: Added restart_guard: Arc<tokio::sync::Mutex<()>> to both SinkDetails and SourceDetails
The guard is acquired at the start of restart_connector() to serialize concurrent restart requests for the same connector
Applied to both sink and source managers (6d9fc157)

core/connectors/runtime/src/sink.rs
seokjin0414: Extracted common logic into shared functions:
  - Sink: setup_sink_consumers() + spawn_consume_tasks() — both init() and start_connector() now use these (7d95a1fe)
  - Source: setup_source_producer() + spawn_source_handler() — both init()/handle() and start_connector() now use these (274d03ad)

core/connectors/runtime/src/manager/sink.rs
mmodzelewski: Maybe we could do `try_lock` instead? Currently, the requests are getting queued, which might not be desired. If the restart is already in progress, then we just return OK.

core/connectors/runtime/src/manager/sink.rs
```diff
@@ -96,6 +108,163 @@ impl SinkManager {
...
         }
     }
+
+    pub async fn stop_connector_with_guard(
+        &self,
+        key: &str,
+        metrics: &Arc<Metrics>,
+    ) -> Result<(), RuntimeError> {
+        let guard = {
+            let details_arc = self
```
spetz: Do we need to name this with `_arc`? Could be just `details`.

core/integration/tests/connectors/postgres/restart.rs
```diff
@@ -0,0 +1,163 @@
...
+use reqwest::Client;
+use std::time::Duration;
+use tokio::time::sleep;
+
+const API_KEY: &str = "test-api-key";
+const SINK_TABLE: &str = "iggy_messages";
+const SINK_KEY: &str = "postgres";
+
+type SinkRow = (i64, String, String, Vec<u8>);
+
```
spetz, line 39: For the overall code clarity, maybe let's move the helper fn down below, so that you read code like a book and don't need to scroll up to find the function referenced in the layer below :)

Also, let's add the test(s) to ensure that we can execute multiple parallel restart requests and nothing will break (guard will do its job).

core/connectors/runtime/src/manager/sink.rs
seokjin0414: Renamed all details_arc to details using block-scoped shadowing (4d3ff4f1)

core/integration/tests/connectors/postgres/restart.rs
seokjin0414: Moved helper functions below test functions
Added parallel_restart_requests_should_not_break_connector test — sends 5 concurrent restart requests, asserts all return 204, then verifies connector processes messages normally
(f7e5ba85)

core/connectors/runtime/src/manager/sink.rs
seokjin0414: Changed to try_lock() — if restart is already in progress returns OK immediately instead of queuing
Applied to restart_connector() only; stop_connector_with_guard() keeps lock().await since shutdown must wait for completion
(e06aeef6)

mmodzelewski (approved):
Thanks @seokjin0414!

PR #2783, opened 2026-02-19
feat(connectors): Delta Lake Sink Connector

### LINKED ISSUE #1852: Implement Delta Lake connectors
There could be both, `sink` and `source` connectors, allowing to read or write from/to one of the supported delta lakes (configurable).

kriti-sc:

Closes #1852 

## Rationale

Delta Lake is a data analytics engine, and very popular in modern streaming analytics architectures.

## What changed?

Introduces a Delta Lake Sink Connector that enables writing data from Iggy to Delta Lake.

The Delta Lake writing logic is heavily inspired by the [kafka-delta-ingest project](https://github.com/delta-io/kafka-delta-ingest/tree/main), to have a proven starting ground for writing to Delta Lake.

## Local Execution
1. Produced 32632 messages with schema `user_id: String, user_type: u8, email: String, source: String, state: String, created_at: DateTime<Utc>, message: String` using [sample data producer](https://github.com/apache/iggy/tree/master/examples/rust/src/sink-data-producer).
2. Consumed messages using the Delta Lake sink and created a Delta table on filesystem.
3. Verified number of rows in delta table and the schema. 
4. Added unit tests and e2e tests, both passing.

image
Left: messages produced; Right(top): messages consumed by Delta sink; Right(bottom): Inspecting Delta table in python

## AI Usage

1. Which tools? Claude Code
5. Scope of usage? generated functions
6. How did you verify the generated code works correctly? Manual testing by producing data into Iggy and then running the sink and verifying local Delta Lake creation, unit tests and e2e tests for local Delta Lake and Delta Lake on S3.
7. Can you explain every line of the code if asked? Yes


core/connectors/sinks/delta_sink/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#[global_allocator]
+static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
```
hubcio: this crate is `cdylib` loaded via `dlopen2` by the connector runtime, which already declares its own `#[global_allocator]`. two global allocators in one process causes heap corruption (memory allocated by one, freed by the other). No other existing sink connector defines `#[global_allocator]`, removethis

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,197 @@
...
+            self.id, version
+        );
+
+        Ok(())
+    }
+
+    async fn close(&mut self) -> Result<(), Error> {
+        info!("Delta Lake sink connector with ID: {} is closed.", self.id);
+        Ok(())
+    }
```
hubcio, line 168: `close()` is a no-op. The `JsonWriter` has internal Parquet buffers,if the runtime shuts down or a prior `flush_and_commit` failed, buffered data is silently dropped, the state (`Mutex<Option<SinkState>>`) is never set back to `None`. `close()` should attempt a final `flush_and_commit` and `take()` the state.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,197 @@
...
+        let table =
+            match deltalake::open_table_with_storage_options(table_url, storage_options.clone())
+                .await
+            {
+                Ok(table) => table,
+                Err(_) if !self.config.schema.is_empty() => {
+                    info!("Table does not exist, creating from configured schema...");
+                    create_table(&self.config.table_uri, storage_options, &self.config.schema)
+                        .await?
+                }
```
hubcio: `Err(_) if !self.config.schema.is_empty()` catches network failures, auth errors, corrupted metadata anything....  and proceeds to create a new table. with wrong S3 credentials + schema configured, it will attempt to create a new empty table, diverging from the real one. Must match on the specific "not found" error variant

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,197 @@
...
+                    let json_bytes = simd_json::to_vec(simd_value).map_err(|e| {
+                        error!("Failed to serialize JSON payload: {e}");
+                        Error::InvalidJsonPayload
+                    })?;
+                    let value: serde_json::Value =
+                        serde_json::from_slice(&json_bytes).map_err(|e| {
+                            error!("Failed to parse JSON payload: {e}");
+                            Error::InvalidJsonPayload
+                        })?;
+                    json_values.push(value);
```
hubcio: double allocation. the elasticsearch_sink already has a direct `owned_value_to_serde_json` converter for this reason.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        // Flush buffers to object store and commit to Delta log
+        let version = state
+            .writer
+            .flush_and_commit(&mut state.table)
+            .await
+            .map_err(|e| {
+                error!("Failed to flush and commit to Delta table: {e}");
+                Error::Storage(format!("Failed to flush and commit: {e}"))
+            })?;
```
hubcio: when `flush_and_commit` fails, data is already in the writer's buffer. The runtime exits the consumer task on error (`core/connectors/runtime/src/sink.rs:322-329`), but the consumer group has already committed the offset. Those messages are permanently lost. Additionally, the writer buffer is left in an undefined state -- no `reset()` call to clear it.

core/connectors/sinks/delta_sink/src/utils.rs
```diff
@@ -0,0 +1,149 @@
...
+use iggy_connector_sdk::Error;
+use tracing::error;
+
+pub(crate) fn parse_schema(schema: &[String]) -> Result<Vec<StructField>, Error> {
+    schema
+        .iter()
+        .map(|entry| {
+            let parts: Vec<&str> = entry.split_whitespace().collect();
+            if parts.len() != 2 {
+                return Err(Error::InvalidConfig);
```
hubcio: add print what went wrong, similar to line 53

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,197 @@
...
+        // Write JSON values to internal Parquet buffers
+        state.writer.write(json_values).await.map_err(|e| {
+            error!("Failed to write to Delta writer: {e}");
+            Error::Storage(format!("Failed to write to Delta writer: {e}"))
+        })?;
+
+        // Flush buffers to object store and commit to Delta log
+        let version = state
+            .writer
+            .flush_and_commit(&mut state.table)
```
hubcio: with `poll_interval = "5ms"` and `batch_length = 100`, this produces thousands of tiny Parquet files per second. Delta Lake is not designed for this write frequency. Should support configurable flush thresholds (size/time/count).

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: made this change. would it make sense to create a module for common utilities?

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Agreed - 5ms and 100 are not the best config for Delta sink. 
These are configurable though, in the config.toml. 

Separately, agree on having a `size` flush threshold too. These configurations are used by runtime to configure the Iggy consume that runs this sink. Right now, it supports only `count` aka batch_length and `time` aka poll_interval only.

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Fixed, now matching on `deltalake::DeltaTableError::NotATable(_)`

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: fixed -- cleaning up state and added a final flush.

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: two parts to the problem if `flush_and_commit` fails:
1. losing data because offset has been commit -- leaving this gap for now, notes below
2. undefined state of the writer buffer -- fixed this with `reset`

notes:
- the idea solution is `AutoCommitAfter::ConsumingAllMessages` but that is out of scope of this PR
- the ideal solution also adds a DLQ to hold failed payloads. had been planning to propose this as an improvement in the runtime separtely.
- a stop gap is retry based on cause of failure. intending to implement a retry logic across all sinks in another PR.

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Ended up removing automatic table creation because we are not doing it in any of the other sinks. Looked around and best practices dictate that table creation should be explicit and left to the user for governance reasons.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+        }
+    }
+}
+
+fn apply_coercion(value: &mut Value, node: &CoercionNode) {
+    match node {
+        CoercionNode::Coercion(Coercion::ToString) => {
+            if !value.is_string() {
+                *value = Value::String(value.to_string());
+            }
```
krishvishal, line 102: This seems to coerce a null value into "null". Is this expected behavior? 

Generally in DBs like SQLite null and "null" are different types.

core/connectors/sinks/delta_sink/src/sink.rs
hubcio: you can write a todo in code. for now, leave it as is.

core/connectors/sinks/delta_sink/src/sink.rs
hubcio: alright - i'm fine with your proposals, also i'm very fine with delta lake connected being not perfect - just document these imperfections with TODO in code, so we'll be aware of them.

core/connectors/sinks/delta_sink/src/sink.rs
hubcio: how would you address this?

core/connectors/sinks/delta_sink/src/coercions.rs
kriti-sc: It is converting non-string values to string. For ex., if input is {"a": 4} and Delta expects this field to be string, then it will be converted to  {"a": "4"}.

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: On a high level, the connector accumulates data until the threshold defined by `size` is reached, after which it writes a Delta file (ie calls `flush_and_commit`).

One important caveat is that this `size` corresponds only to the size of the JSON payloads received from Iggy messages, not the final file size. This is because the Delta writer does not expose the size of the accumulated Parquet bytes before the file is written.

This approach is not ideal, as each connector would need to implement similar logic. A more robust solution would be to expose size as a runtime configuration parameter (similar to `poll_interval` and `batch_len`). However, I need to do additional research to determine whether this is viable.

I plan to open a discussion thread once I have gathered more information. For now, I propose adding a TODO in the Delta sink code regarding the size configuration and deferring the implementation to a later stage.

Let me know if this addresses your question.

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: I started adding a TODO, but this likely warrants a dedicated discussion thread, as the change would affect all sinks rather than just this implementation. Let me know if that makes sense and I can open one.

PR #2785, opened 2026-02-19
feat(cpp): Add bazel building and testing infrastructure

### LINKED ISSUE #2762: [C++ SDK] Create build system + CI infrastructure
### Description

The building and testing infrastrure for C++ SDK needs to be laid down before the actual development.

### Component

None

### Proposed solution

- Create the required scripts for building and testing in CI.
- For cpp side, include `CompressionAlgorithm`, `Expiry`, `MaxTopicSize`, and `PollingStrategyclasses` and test them.
- Implement Bazel and CMake build systems.

Bazel commands:
```bash
bazel build //:iggy-cpp
bazel test //:iggy-cpp-test
```

### Alternatives considered

_No response_

sub-issue of #2100

slbotbm:

Closes #2762

## What changed?

Adds bazel build system + stub classes and their tests + ci scripts

## Local Execution

- Passed 
- Pre-commit hooks ran

## AI Usage

1. Which tools? codex
2. Scope of usage? build system implementation
3. How did you verify the generated code works correctly? checking on local environment


.github/actions/cpp-bazel/pre-merge/action.yml
```diff
@@ -0,0 +1,77 @@
...
+    - name: Install clang-format
+      if: inputs.task == 'lint'
+      shell: bash
+      run: |
+        if command -v clang-format >/dev/null 2>&1; then
+          clang-format --version
+          exit 0
+        fi
+        sudo apt-get update --yes
+        sudo apt-get install --yes clang-format
```
hubcio: pin clang format to very specific version, otherwise update of upstream package will break CI if it's incompatible

foreign/cpp/include/iggy.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    std::string expiry_kind() const { return expiry_kind_; }
+    std::uint32_t expiry_value() const { return expiry_value_; }
+
+  private:
+    explicit Expiry(std::string expiry_kind, std::uint32_t expiry_value)
+        : expiry_kind_(std::move(expiry_kind)), expiry_value_(expiry_value) {}
+
+    std::string expiry_kind_;
+    std::uint32_t expiry_value_;
```
hubcio: in Rust expiry is u64

foreign/cpp/include/iggy.hpp
```diff
@@ -0,0 +1,109 @@
...
+  private:
+    explicit CompressionAlgorithm(std::string algorithm) : algorithm_(std::move(algorithm)) {}
+
+    std::string algorithm_;
+};
+
+class Expiry final {
+  public:
+    static Expiry server_default() { return Expiry("server_default", 0); }
+    static Expiry never_expire() { return Expiry("never_expire", 0); }
```
hubcio: wrong: never expire should be u64::MAX and server default is 0.

foreign/cpp/include/iggy.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    std::string expiry_kind_;
+    std::uint32_t expiry_value_;
+};
+
+class MaxTopicSize final {
+  public:
+    static MaxTopicSize server_default() { return MaxTopicSize("server_default"); }
+    static MaxTopicSize unlimited() { return MaxTopicSize("unlimited"); }
+    static MaxTopicSize from_bytes(std::uint64_t bytes) { return MaxTopicSize(std::to_string(bytes)); }
```
hubcio: weird semantics, you pass std::uint64_t but the argument name is bytes. also, all other SDKs use enum discriminants instead of std string discriminants: current approach is fragile.  you dont have to fix it now, just keep that in mind.

foreign/cpp/BUILD.bazel
```diff
@@ -0,0 +1,103 @@
...
+        if [ "$(COMPILATION_MODE)" = "opt" ]; then
+            PROFILE="release"
+            FLAGS="--release"
+        fi
+
+        env -u PWD CARGO_TARGET_DIR="target" "$$CARGO" build $$FLAGS
+
+        cp "target/$$PROFILE/libiggy_cpp.a" "$$OUT_LIB"
+        find "target/$$PROFILE/build" -name lib.rs.h | head -n 1 | xargs -I{} cp {} "$$OUT_RS"
+        find "target/$$PROFILE/build" -name cxx.h    | head -n 1 | xargs -I{} cp {} "$$OUT_CXX"
```
hubcio: not a fan of this. what if find returns 0 results?
also, `find | xargs` without `-print0/-0` breaks on filenames with spaces

foreign/cpp/BUILD.bazel
```diff
@@ -0,0 +1,103 @@
...
+    """,
+    local = 1,
```
hubcio, line 71: cargo writes to source tree target/, cannot be remotely cached, concurrent Bazel invocations can race. Should at minimum be documented as known limitation.

foreign/cpp/BUILD.bazel
```diff
@@ -0,0 +1,103 @@
...
+        cd "$$PROJECT_ROOT"
+
+        PROFILE="debug"
+        FLAGS=""
+        if [ "$(COMPILATION_MODE)" = "opt" ]; then
+            PROFILE="release"
+            FLAGS="--release"
+        fi
+
+        env -u PWD CARGO_TARGET_DIR="target" "$$CARGO" build $$FLAGS
```
hubcio: missing `--locked` for `cargo build`

foreign/cpp/include/iggy.hpp
```diff
@@ -0,0 +1,109 @@
...
+    std::string algorithm_;
+};
+
+class Expiry final {
+  public:
+    static Expiry server_default() { return Expiry("server_default", 0); }
+    static Expiry never_expire() { return Expiry("never_expire", 0); }
+    static Expiry duration(std::uint32_t micros) { return Expiry("duration", micros); }
+
+    std::string expiry_kind() const { return expiry_kind_; }
```
hubcio: either return `const std::string&` or `std::string_view`, but even better would be enum discriminant.. but cpp doesn't have enums with payload, so i hope you figure out something nice :D

foreign/cpp/MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+
+bazel_dep(name = "rules_cc", version = "0.2.16")
+bazel_dep(name = "rules_rust", version = "0.68.1")
+bazel_dep(name = "platforms", version = "1.0.0")
+bazel_dep(name = "googletest", version = "1.17.0")
+
+rust = use_extension("@rules_rust//rust:extensions.bzl", "rust")
+rust.toolchain(
+    edition = "2024",
+    versions = ["1.93.0"],
```
hubcio: there is script which syncs rust version across whole repo. please add this file to it and handle it.
`scripts/ci/sync-rust-version.sh`

foreign/cpp/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+ignored = ["cxx-build"]
+
+[lib]
+crate-type = ["staticlib"]
+
+[dependencies]
+cxx = "1.0.194"
+
+[build-dependencies]
+cxx-build = "1.0"
```
hubcio: pin to specific version that includes patch

foreign/cpp/include/iggy.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+  private:
+    explicit PollingStrategy(std::string kind, std::uint64_t value)
+        : polling_strategy_kind_(std::move(kind)), polling_strategy_value_(value) {}
+
+    std::string polling_strategy_kind_;
+    std::uint64_t polling_strategy_value_;
+};
+
+class IdKind final {
```
hubcio, line 107: I don't understand what's the added value of this type. `IdKind` is a bare discriminant without an `Identifier` compound type. All other SDKs have `Identifier`. `IdKind` alone is not useful for usersof this SDK.

foreign/cpp/.bazelrc
```diff
@@ -0,0 +1,53 @@
...
+
+# Bazel configuration for iggy C++ bindings
+
+# Enable BzlMod
+common --enable_bzlmod
+
+# Shared compiler warnings
+build --copt=-Wall
+build --copt=-Wpedantic
+build --copt=-Werror
```
hubcio, line 26: missing `-Wextra`

foreign/cpp/include/iggy.hpp
slbotbm: These values will ultimately be passed on to the rust side, and if the user selects server_default or never_expire, the rust side will not look at the value. That is why I had set this to 0. I can set it to `std::numeric_limits<uint64_t>::max()`.

foreign/cpp/include/iggy.hpp
slbotbm: Identifier will be a shared struct between rust and cpp that will be added in the next PR. I just added this early here.

foreign/cpp/include/iggy.hpp
slbotbm: In the rust sdk, `MaxTopicSize::from(u64)` is one of options the user can use. That is why I created this. With regards to enum discriminants, I had initially defined those on the rust side to be exported to cpp, but it created a lot of complexity and unnecessary copying between structs. That is why I just defined strings in the high-level api. With regards to the fragility, I agree. I had copied one of the approaches of the rust side:
```rust
impl FromStr for MaxTopicSize {
    type Err = String;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let result = match s {
            "unlimited" => MaxTopicSize::Unlimited,
            "0" | "server_default" => MaxTopicSize::ServerDefault,
            value => {
                let size = value.parse::<IggyByteSize>().map_err(|e| format!("{e}"))?;
                match size.as_bytes_u64() {
                    0 => MaxTopicSize::ServerDefault,
                    u64::MAX => MaxTopicSize::Unlimited,
                    _ => MaxTopicSize::Custom(size),
                }
            }
        };

        Ok(result)
    }
}
```

foreign/cpp/Cargo.toml
slbotbm: done

foreign/cpp/.bazelrc
slbotbm: done

foreign/cpp/MODULE.bazel
slbotbm: done

foreign/cpp/include/iggy.hpp
slbotbm: All *_kind() functions now return `std::string_view`. with regards to enum discriminants, I would like to first build out the complete client, and then maybe iterate over which approach would feel the most natural to the user. As written in the previous comment, defining these structs in rust involves a lot of unnecessary copying.

foreign/cpp/BUILD.bazel
slbotbm: done

foreign/cpp/include/iggy.hpp

.github/actions/cpp-bazel/pre-merge/action.yml
slbotbm: done

foreign/cpp/include/iggy.hpp
hubcio: do it ;)

foreign/cpp/include/iggy.hpp
hubcio: good. then please add sub-issue to your cpp issue so that it wont be forgotten.

foreign/cpp/include/iggy.hpp
hubcio: then please add TODO in code so that it's clear

foreign/cpp/include/iggy.hpp
hubcio: for now we can leave it as is

foreign/cpp/include/iggy.hpp

foreign/cpp/include/iggy.hpp
slbotbm: alright

foreign/cpp/include/iggy.hpp

foreign/cpp/MODULE.bazel
slbotbm: I had been using bazel 9 on my pc, which required me to download the rust toolchain. I switched to bazel 8, which does not require that, and we can use the cargo toolchain provided by setup-rust-with-cache. Therefore, I don't think I need to change cripts/ci/sync-rust-version.sh.

foreign/cpp/BUILD.bazel
slbotbm: I changed the build step to use the system-managed cargo instead of downloading a completely new toolchain. Further, cargo now writes to bazel-managed output directories. I'll also add a comment in the README.md about this potential issue.

foreign/cpp/BUILD.bazel
slbotbm: I changed the build system to exit with a nonzero status code if nothing is found. I also removed xargs.

.github/actions/cpp-bazel/pre-merge/action.yml
```diff
@@ -0,0 +1,80 @@
...
+      if: inputs.task == 'lint'
+      shell: bash
+      run: |
+        if command -v clang-format >/dev/null 2>&1; then
+          clang-format --version
+          exit 0
+        fi
+        sudo apt-get update --yes
+        sudo apt-get install --yes clang-format-18
+        clang-format --version
```
hubcio: still not properly fixed. you need to call clang-format-18, because if it was installed it'll use system version.

slbotbm: I apologize for the oversight -- every call site now uses clang-format-18

mmodzelewski (approved):

PR #2811, opened 2026-02-24
feat(shard): create `shard` crate and generic `IggyShard`

krishvishal:
- Created `shard` crate and generic `IggyShard`
 - `IggyShard::dispatch()` routes incoming network messages to the correct consensus plane via `PlaneIdentity`

### DIFF
--- a/.github/actions/node-npm/pre-merge/action.yml
+++ b/.github/actions/node-npm/pre-merge/action.yml
@@ -79,11 +79,13 @@ runs:
         fi
       shell: bash
 
+
     - name: Test
       if: inputs.task == 'test'
       run: |
         cd foreign/node
-        npm run test:unit
+        mkdir -p ../../reports/node-coverage/unit
+        npx c8 --reporter=lcov --reports-dir=../../reports/node-coverage/unit npm run test:unit
       shell: bash
 
     - name: Start Iggy server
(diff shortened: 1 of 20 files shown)


core/shard/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+use iggy_common::sharding::{IggyNamespace, ShardId};
+use iggy_common::variadic;
+use journal::{Journal, JournalHandle};
+use message_bus::MessageBus;
+use metadata::IggyMetadata;
+use metadata::stm::StateMachine;
+use partitions::{IggyPartitions, PartitionsConfig};
+
+// variadic!(Metadata, Partitions) = (Metadata, (Partitions, ()))
+type PlaneInner<B, J, S, M> = (
```
numinnex, line 30: I think you could import the `variadic!` macro in there and use it ?

core/shard/src/lib.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub replica_count: u8,
+    pub plane: ShardPlane<B, J, S, M>,
+}
+
+impl<B, J, S, M> IggyShard<B, J, S, M>
+where
+    B: MessageBus + Clone,
+{
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
```
numinnex, line 51: Lets pass in already constructed consensus, metadata instances in the constructor

krishvishal: Yes, I've realized my mistake and refactored the code. I've pushed the refactor, please take a look now.

Issue #2739, opened 2026-02-16
MongoDB Connector


amuldotexe:
### Description

MongoDB connector for Iggy (reference conversation with [competitor research here](https://discord.com/channels/1144142576266530928/1144142577369628684/1472674244843536506) ) 

### Component

Connectors

ryerraguntla:
Please let me know if help is needed here.

hubcio:
closing as superseded. the mongodb sink shipped (`core/connectors/sinks/mongodb_sink/`); the remaining source piece is tracked in #3042. nothing else from this issue is outstanding.

PR #2815, opened 2026-02-25
feat(connectors): add MongoDB sink connector

amuldotexe:

Partially addresses #2739 (MongoDB sink only).
MongoDB source support will follow in a separate PR.

## Rationale

We need a MongoDB sink connector with explicit failure behavior so writes are never reported as successful when they are not.

## What changed?

Before this change, MongoDB sink support was missing for connector runtime users.  
This PR adds the MongoDB sink connector, including insert/write logic, retry handling for transient failures, explicit duplicate-key failure behavior, metadata mapping, and delivery-semantics documentation.  
It also adds sink-focused integration and unit tests to validate payload formats, batch behavior, auto-create collection behavior, and non-silent failure paths.

## Local Execution

- Passed: `cargo fmt --all -- --check`
- Passed: `cargo clippy -p iggy_connector_mongodb_sink --all-targets -- -D warnings`
- Passed: `cargo test -p iggy_connector_mongodb_sink` (13 passed)
- Passed: `cargo test -p integration --test mod -- mongodb_sink` (10 passed)
- Pre-commit hooks ran
- Docker E2E proof image: `ghcr.io/amuldotexe/iggy-mongodb-sink-demo:issue-2739-05fbec16`
- Docker smoke result: `SMOKE_OK ... docs=3`
- GHCR package: https://github.com/users/amuldotexe/packages/container/package/iggy-mongodb-sink-demo

- Key sink tests:
- `duplicate_key_is_explicit_failure_and_not_silent_success`
- `ordered_duplicate_partial_insert_has_exact_accounting`
- `schema_validation_mid_batch_surfaces_hard_error_and_partial_prefix`
- `write_concern_timeout_does_not_report_full_success`
- `retryable_write_failover_keeps_single_doc_per_id`
- `no_writes_performed_label_path_preserves_state_accuracy`
- `json_messages_sink_to_mongodb`
- `binary_messages_sink_as_bson_binary`
- `large_batch_processed_correctly`
- `auto_create_collection_on_open`
- `given_no_client_should_return_error_not_silent_ok`

## AI Usage

1. Which tools?
   Codex, Claude Code, Rust Rover
2. Scope of usage?
   PRD, connector precedent analysis, TDD, implementation, and PR prep
   Ran all local format, clippy, unit, and integration checks listed above
   Yes

### DIFF
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
@@ -28,6 +28,7 @@ import org.apache.iggy.client.async.StreamsClient;
 import org.apache.iggy.client.async.SystemClient;
 import org.apache.iggy.client.async.TopicsClient;
 import org.apache.iggy.client.async.UsersClient;
+import org.apache.iggy.client.async.tcp.AsyncTcpConnection.TCPConnectionPoolConfig;
 import org.apache.iggy.config.RetryPolicy;
 import org.apache.iggy.exception.IggyMissingCredentialsException;
 import org.apache.iggy.exception.IggyNotConnectedException;
@@ -94,7 +95,7 @@ public class AsyncIggyTcpClient {
     private final int port;
     private final Optional<String> username;
     private final Optional<String> password;
-    private final Optional<Duration> connectionTimeout;
+    private final Optional<Duration> acquireTimeout;
     private final Optional<Duration> requestTimeout;
     private final Optional<Integer> connectionPoolSize;
     private final Optional<RetryPolicy> retryPolicy;
@@ -129,7 +130,7 @@ public class AsyncIggyTcpClient {
             int port,
             String username,
             String password,
-            Duration connectionTimeout,
+            Duration acquireTimeout,
             Duration requestTimeout,
             Integer connectionPoolSize,
             RetryPolicy retryPolicy,
@@ -139,7 +140,7 @@ public class AsyncIggyTcpClient {
         this.port = port;
         this.username = Optional.ofNullable(username);
         this.password = Optional.ofNullable(password);
-        this.connectionTimeout = Optional.ofNullable(connectionTimeout);
+        this.acquireTimeout = Optional.ofNullable(acquireTimeout);
         this.requestTimeout = Optional.ofNullable(requestTimeout);
         this.connectionPoolSize = Optional.ofNullable(connectionPoolSize);
         this.retryPolicy = Optional.ofNullable(retryPolicy);
@@ -166,7 +167,12 @@ public class AsyncIggyTcpClient {
      * @return a {@link CompletableFuture} that completes when the connection is established
      */
     public CompletableFuture<Void> connect() {
-        connection = new AsyncTcpConnection(host, port, enableTls, tlsCertificate);
+        TCPConnectionPoolConfig.Builder poolConfigBuilder = new TCPConnectionPoolConfig.Builder();
+        connectionPoolSize.ifPresent(poolConfigBuilder::setMaxConnections);
+        acquireTimeout.ifPresent(timeout -> poolConfigBuilder.setAcquireTimeoutMillis(timeout.toMillis()));
+        TCPConnectionPoolConfig poolConfig = poolConfigBuilder.build();
+        // TCPConnectionPoolConfig poolConfig = new TCPConnectionPoolConfig();
+        connection = new AsyncTcpConnection(host, port, enableTls, tlsCertificate, poolConfig);
         return connection.connect().thenRun(() -> {
             messagesClient = new MessagesTcpClient(connection);
             consumerGroupsClient = new ConsumerGroupsTcpClient(connection);
@@ -215,7 +221,6 @@ public class AsyncIggyTcpClient {
         }
         return usersClient;
     }
-
     /**
      * Returns the async messages client for producing and consuming messages.
      *
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
@@ -73,6 +73,7 @@ public final class AsyncIggyTcpClientBuilder {
     private Duration requestTimeout;
     private Integer connectionPoolSize;
     private RetryPolicy retryPolicy;
+    private Duration acquireTimeout;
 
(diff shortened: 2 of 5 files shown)


core/connectors/sinks/mongodb_sink/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+license = "Apache-2.0"
+keywords = ["iggy", "messaging", "streaming", "mongodb", "sink"]
+categories = ["command-line-utilities", "database", "network-programming"]
+homepage = "https://iggy.apache.org"
+documentation = "https://iggy.apache.org/docs"
+repository = "https://github.com/apache/iggy"
+readme = "../../README.md"
+
+[package.metadata.cargo-machete]
+ignored = ["dashmap", "once_cell", "futures", "simd-json"]
```
krishvishal: If these dependencies are not needed then they should be removed instead of being suppressed at cargo-machete level.

core/connectors/runtime/src/sink.rs
```diff
@@ -532,5 +531,13 @@ async fn process_messages(
...
     );
 
+    if consume_result != 0 {
+        return Err(RuntimeError::ConnectorSdkError(
+            iggy_connector_sdk::Error::CannotStoreData(format!(
+                "Sink consume callback returned non-zero status ({consume_result}) for plugin ID: {plugin_id}"
+            )),
+        ));
+    }
+
```
krishvishal: Previously the errors are silently dropped but this change now could cause some connectors to crash.

This change is better converted into a separate PR that handles the consequences properly.  Because it effects other connectors' behavior.

core/connectors/sinks/mongodb_sink/src/lib.rs
```diff
@@ -0,0 +1,681 @@
...
+
+const DEFAULT_MAX_RETRIES: u32 = 3;
+const DEFAULT_RETRY_DELAY: &str = "1s";
+
+#[derive(Debug)]
+pub struct MongoDbSink {
+    pub id: u32,
+    client: Option<Client>,
+    config: MongoDbSinkConfig,
+    state: Mutex<State>,
```
krishvishal: Mutex is not necessary here. Since `State` contains only two `u64` members, we can convert them into `AtomicU64`.

core/connectors/sinks/mongodb_sink/src/lib.rs
```diff
@@ -0,0 +1,681 @@
...
+    Json,
+    String,
+}
+
+impl PayloadFormat {
+    fn from_config(s: Option<&str>) -> Self {
+        match s.map(|s| s.to_lowercase()).as_deref() {
+            Some("json") => PayloadFormat::Json,
+            Some("string") | Some("text") => PayloadFormat::String,
+            _ => PayloadFormat::Binary,
```
krishvishal: Its better to throw a warning here that an unrecognized format has been passed in config.

core/connectors/sinks/mongodb_sink/src/lib.rs
```diff
@@ -0,0 +1,681 @@
...
+        let mut docs = Vec::with_capacity(messages.len());
+
+        for message in messages {
+            let mut doc = mongodb::bson::Document::new();
+
+            // Add message ID as string (MongoDB doesn't support u128)
+            doc.insert("_id", message.id.to_string());
+
+            if include_metadata {
+                doc.insert("iggy_offset", message.offset as i64);
```
krishvishal: `message.offset` is u64. Direct casting to `i64` could cause wrapping to negative value when the `offset > i64::MAX`

core/connectors/sinks/mongodb_sink/src/lib.rs
```diff
@@ -0,0 +1,681 @@
...
+                // Convert microseconds to milliseconds for BSON DateTime
+                let timestamp_ms = (message.timestamp / 1000) as i64;
+                let bson_timestamp = bson::DateTime::from_millis(timestamp_ms);
+                doc.insert("iggy_timestamp", bson_timestamp);
+                doc.insert("iggy_stream", &topic_metadata.stream);
+                doc.insert("iggy_topic", &topic_metadata.topic);
+                doc.insert("iggy_partition_id", messages_metadata.partition_id as i32);
```
krishvishal: Same comment as `message.offset`

core/connectors/sinks/mongodb_sink/src/lib.rs
```diff
@@ -0,0 +1,681 @@
...
+        &self,
+        collection: &Collection<mongodb::bson::Document>,
+        docs: &[mongodb::bson::Document],
+    ) -> Result<(), Error> {
+        let max_retries = self.get_max_retries();
+        let retry_delay = self.retry_delay;
+        let mut attempts = 0u32;
+
+        loop {
+            let result = collection.insert_many(docs.to_vec()).await;
```
krishvishal: When `insert_many` (with `ordered=true`, the default) hits a duplicate key at position 100, documents 0..99 are already committed to MongoDB. On retry, `docs.to_vec()` sends the entire batch again. 

The already-committed documents trigger duplicate key errors, which `is_transient_error` classifies as non-retryable which causes retry to fail immediately with a misleading duplicate-key error instead of retrying only the remaining documents. So for partial writes this is currently behaves incorrectly.

core/connectors/sinks/mongodb_sink/src/lib.rs
```diff
@@ -0,0 +1,681 @@
...
+        messages_metadata: &MessagesMetadata,
+        collection: &Collection<mongodb::bson::Document>,
+    ) -> Result<(), Error> {
+        if messages.is_empty() {
+            return Ok(());
+        }
+
+        let include_metadata = self.config.include_metadata.unwrap_or(true);
+        let include_checksum = self.config.include_checksum.unwrap_or(true);
+        let include_origin_timestamp = self.config.include_origin_timestamp.unwrap_or(true);
```
krishvishal: Config is validated every batch. These can be moved to `new()`.

atharvalade:
Adding to @krishvishal points which already cover a lot of ground. Two more things I think are blockers here. 

First, there's a liveness issue with how partial failures interact with offset commits. When any batch inside `process_messages` fails, the whole call returns Err, so the consumer offset never advances. But ordered `insert_many` already committed the documents before the failure point. On the next poll the upstream re-sends everything, those already committed docs hit duplicate `_id` errors, which are non-transient so retries fail immediately, offset still doesn't commit, and you're stuck in a permanent loop. The connector can never recover from even a single transient mid-batch failure. 

Second, `message.id.to_string`() as the `MongoDB _id` isn't safe when you're consuming multiple topics into the same collection. Two unrelated messages from different topics can easily share the same id, and whichever one lands first wins while the other silently fails. The `_id` should be a composite of stream, topic, partition, and message id to avoid collisions and quiet data loss.

hubcio (approved):

PR #2825, opened 2026-02-26
feat(consensus): add loopback queue for primary self-addressed messages

hubcio:
The primary was sending PrepareOk to itself through the message
bus, requiring a network round-trip and an unwrap() on bus
errors. Replace with an in-process VecDeque that the simulator
drains after each dispatch. Cleared on view change reset and
on DVC quorum completion.

Includes defensive debug_assert guards on push_loopback
overflow, process_loopback buf precondition, and simulator
drain idempotency. Adds test verifying loopback is cleared
when complete_view_change_as_primary fires.

### DIFF
--- a/core/binary_protocol/src/cli/binary_cluster/get_cluster_metadata.rs
+++ b/core/binary_protocol/src/cli/binary_cluster/get_cluster_metadata.rs
@@ -1,0 +1,121 @@
+/* Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use crate::Client;
+use crate::cli::cli_command::{CliCommand, PRINT_TARGET};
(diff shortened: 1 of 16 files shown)


core/consensus/src/impls.rs
```diff
@@ -1139,6 +1149,26 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
         false
     }
 
+    /// Enqueue a self-addressed PrepareOk for the primary to process in the next loopback drain.
+    pub(crate) fn push_loopback(&self, message: Message<PrepareOkHeader>) {
+        // At most one loopback per pipeline entry and the pipeline is bounded by
+        // PIPELINE_PREPARE_QUEUE_MAX; `on_ack` never re-enqueues (simulator asserts this).
+        debug_assert!(
```
krishvishal: IMO its better to upgrade it to `assert!()` because the underlying `VecDeque` is unbounded and can cause OOMs.

core/metadata/src/impls/metadata.rs
```diff
@@ -100,6 +100,12 @@ pub struct IggyMetadata<C, J, S, M> {
     pub mux_stm: M,
 }
 
+impl<C, J, S, M> IggyMetadata<C, J, S, M> {
+    pub fn consensus(&self) -> Option<&C> {
```
krishvishal: `consensus` is already `pub`. So do we need this accessor?

hubcio: good point

core/consensus/src/impls.rs
```diff
@@ -1139,6 +1149,24 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
         false
     }
 
+    /// Enqueue a self-addressed PrepareOk for the primary to process in the next loopback drain.
+    pub(crate) fn push_loopback(&self, message: Message<PrepareOkHeader>) {
```
numinnex: We don't send only PrepareOk to our selves, other messages such as `DoViewChange` and `StartViewChange`.

numinnex: Also use that method inside of the `send_message_to_replica` or however it's called method, when the replica_id == self.replica_id.

PR #2828, opened 2026-02-27
refactor(server): use papaya instead of DashMap for shards_table

### LINKED ISSUE #2827: Replace `DashMap` with `papaya` for `shards_table` in server shard
The shards_table (`DashMap<IggyNamespace, PartitionLocation>`) is the shared lookup table that maps partition namespaces to their owning shard. It is queried on every message routing decision (find_shard, resolve) but only written to during partition creation, deletion, or rebalancing, which is ready-heavy workload.

`papaya` is a lock-free concurrent hash map specifically optimized for read-heavy access patterns. Switching `shards_table` from DashMap to papaya should reduce contention on the hot routing path.

krishvishal:

Closes #2827

## Rationale

`shards_table` has a read-heavy workload (queried on every message route, written only on partition create/delete/rebalance). papaya's lock-free design is purpose-built for this pattern.

## What changed?
- Replace `DashMap` with `papaya::HashMap` for the `shards_table` shared lookup table, which maps partition namespaces to owning shards.
- All call sites updated to use papaya's pin-based guard API: `pin()` returns a `HashMapRef` with ergonomic `get`/`insert`/`remove`/`iter` methods. Only worry is that as long as `pin()` is not dropped the garbage collector of `papaya` won't clean memory. This maybe a cause for concern across async boundaries.

### DIFF
--- a/core/server/src/shard/mod.rs
+++ b/core/server/src/shard/mod.rs
@@ -36,8 +36,8 @@ use ahash::AHashSet;
 use builder::IggyShardBuilder;
 use iggy_common::SemanticVersion;
 use iggy_common::sharding::{IggyNamespace, PartitionLocation};
-use papaya::HashMap as PapayaMap;
 use iggy_common::{EncryptorKind, IggyError};
+use papaya::HashMap as PapayaMap;
 use std::{
     cell::{Cell, RefCell},
     net::SocketAddr,


hubcio (approved):

numinnex (commented):
Let's not do it for the `IggyShard` on the server, but rather only in the `shard` module that we will use in simulator, this way we can profile them both and decide.

PR #2840, opened 2026-03-01
perf: make IggyMessagesBatch::last_offset O(1)

cijiugechu:

None.

## Rationale

Avoid unnecessary O(n) lookup in `IggyMessagesBatch::last_offset()` to improve throughput.

Bench (single run each; Apple M4, macOS 15.7.4; pinned-producer/TCP; 8 producers, 8 streams; 1000B msgs; 1000 msgs/batch; total data 5GB):

### Results:

Throughput delta: ~88.75Mb/s

#### master: 

Throughput - Pinned Producer Benchmark (master)

#### PR: 
Throughput - Pinned Producer Benchmark (last_offset_o1)

## What changed?

`IggyMessagesBatch::last_offset()` previously used `iter().last()`, which walk the whole batch and made the call O(n) even though the last element is known. In producer hot paths with large `messages_per_batch`, that extra scan is on the critical path.

The method now reads the last message via indexed `get(count - 1)` (O(1)), with an iterator fallback only if the indexed lookup fails.

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage

None.


numinnex:
Hi, thanks for the contribution.

I am pretty sure a simpler solution to that problem would be to provide the `Iterator` with a size hint, this should give compiler enough information to generate code identical to yours. You can try doing that instead, here is example of that optimization being applied for an `Vec<T>`: https://godbolt.org/z/4erhYsxsG

The clue is those two lines
```asm
  mov rdi, qword ptr [rbx + 8]
  mov rbx, qword ptr [rdi + 8*rax - 8] ;Load the ptr + (len - 1)
  ```

cijiugechu:
> > Maybe we should consider implementing more efficient Iterator methods to override the default behavior, so we don’t end up relying on std fallback implementation...
> 
> Right, so maybe a idea would be to implement `DoubleEndedIterator` on top of the `Iterator` trait, as by the docs `DoubleEndedIterator: Iterator`

I gave it a try, and implementing `DoubleEndedIterator` would require caching intermediate state, which would be a fairly substantial change. Also, making `last` O(1) doesn’t inherently depend on `DoubleEndedIterator`—it’s just how std happens to implement it. Given our current optimization goals, implementing an O(1) `last` for  `Iterator` only is probably the more reasonable option.

hubcio (approved):

numinnex (approved):

Issue #2763, opened 2026-02-18
[C++ SDK] Create Rust side FFI bindings for the C++ SDK


slbotbm:
### Description

For C++ to utilize the Rust side code, bindings need to be created.

### Component

None

### Proposed solution

cxx will be used to create these bindings. There will only be one opaque type (`IggyClient`), which will be used  to interact with the Rust code. Transparent types will include `PolledMessages`, `Message`, `ServerDetails`, `TopicDetails`, `Identifier`.

Integration tests will be written for the low level client created by these bindings.

### Alternatives considered

uniffi

sub-issue of #2100

PR #2852, opened 2026-03-03
feat(cpp): part of low-level bindings, their tests, and e2e CI config

slbotbm:

Add part of low-level bindings for iggy-hpp.

Works towards completion of #2763 .

## Rationale
To be merged after #2785 

## What changed?

Adds low-level bindings + tests + e2e CI config. The following functions have been added for the `Client` opaque type:
- new_connection (equivalent to Rust's `IggyClientBuilder::new()` and `IggyClient::from_connection_string(connection_string)`)
- login_user
- connect
- create_stream
- get_stream
- delete_stream
- create_topic
- delete_connection

tests/client/low_level_e2e.cpp -> e2e tests for client
tests/topic/low_level_e2e.cpp -> e2e tests for topic
tests/stream/low_level_e2e.cpp -> e2e tests for stream

NOTE: The PR is big since Cargo.lock itself is 5k LoC. The actual code itself is about 1k LoC.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
3. How did you verify the generated code works correctly? Planned what needs to be implemented, created spec, and then verified all code generated and tests run.

### DIFF
--- a/foreign/cpp/Cargo.lock
+++ b/foreign/cpp/Cargo.lock
@@ -1146,28 +1146,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d7a1e2f27636f116493b8b860f5546edb47c8d8f8ea73e1d2a20be88e28d1fea"
 
 [[package]]
-name = "dbus"
-version = "0.9.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "21b3aa68d7e7abee336255bd7248ea965cc393f3e70411135a6f6a4b651345d4"
-dependencies = [
- "libc",
- "libdbus-sys",
- "windows-sys 0.59.0",
-]
-
-[[package]]
-name = "dbus-secret-service"
-version = "4.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "708b509edf7889e53d7efb0ffadd994cc6c2345ccb62f55cfd6b0682165e4fa6"
-dependencies = [
- "dbus",
- "openssl",
- "zeroize",
-]
-
-[[package]]
 name = "der-parser"
 version = "10.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1222,27 +1200,6 @@ checksum = "9ed9a281f7bc9b7576e61468ba615a66a5c8cfdff42420a70aa82701a3b1e292"
 dependencies = [
  "block-buffer",
  "crypto-common",
-]
-
-[[package]]
-name = "dirs"
-version = "6.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3e8aa94d75141228480295a7d0e7feb620b1a5ad9f12bc40be62411e38cce4e"
-dependencies = [
- "dirs-sys",
-]
-
-[[package]]
-name = "dirs-sys"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e01a3366d27ee9890022452ee61b2b63a67e6f13f58900b651ff5665f0bb1fab"
-dependencies = [
- "libc",
- "option-ext",
- "redox_users",
- "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


foreign/cpp/Cargo.lock
slbotbm, line 1: Currently, when building the binary in local env, cargo build ... --locked is called, which does not modify the lock file. Due to this, the CI has failed multiple times. Just before merging this PR, I would like to add something like a flag that detects whether the run env is CI or not, and if it is CI, only then should the --locked flag be used.

hubcio: what is that a problem?

slbotbm: Normally, when you execute `cargo build`, the lockfile is also updated if there is a need, but that is not happening in the current setup in my local environment due to the --locked flag. This results in a stale lockfile, due to which the CI fails. That is why i want to add a condition for the `--locked` flag to only be used in CI.

foreign/cpp/Cargo.toml
```diff
@@ -28,6 +28,8 @@ crate-type = ["staticlib"]
 
 [dependencies]
 cxx = "1.0.194"
+iggy = { path = "../../core/sdk" }
+tokio = { version = "1", features = ["rt-multi-thread"] }
```
hubcio: use strict version of tokio - we aim to have `somewhat` reproducible builds

foreign/cpp/Cargo.lock
hubcio: don't overcomplicate it.
`Cargo.lock` should always be committed and `--locked should` stay unconditional (both locally and in CI).
That's the standard approach for binary/cdylib crates and we already do it for the root Cargo.lock.

the fix is on the workflow side: whenever you touch Cargo.toml, run cargo check or cargo update -w to refresh the lockfile, then commit it together with your changes. --locked failing on a stale lockfile is working as intended- it's catching the mismatch early.

No need to conditionally disable it.just keep all deps pinned to strict versions (semver = major.minor.patch).

OR remove `Cargo.lock` from version control altogether, just like in python.

foreign/cpp/src/client.rs
```diff
@@ -0,0 +1,205 @@
...
+pub unsafe fn delete_connection(client: *mut Client) -> Result<(), String> {
+    if client.is_null() {
+        return Ok(());
+    }
+
+    let logout_result = RUNTIME.block_on(async { unsafe { &*client }.inner.logout_user().await });
+
+    unsafe {
+        drop(Box::from_raw(client));
+    }
```
hubcio, line 198: if logout_user will fail you will have a leak, this will be tagged by e.g. valgrind if someone will test iggy rigorously

foreign/cpp/tests/common/test_helpers.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <cstdint>
+#include <string>
+
+#include "lib.rs.h"
+
+inline iggy::ffi::Identifier make_string_identifier(const std::string &value) {
+    iggy::ffi::Identifier identifier;
+    identifier.kind   = "string";
+    identifier.length = static_cast<std::uint8_t>(value.size());
```
hubcio, line 31: in Rust `Validatable impl` on `Identifier` will likely catch the mismatch between length and value.len(), but the C++ caller gets no hint of the problem. Consider adding an assert or throwing.

foreign/cpp/tests/topic/low_level_e2e.cpp
hubcio, line 1: for these tests, please add TODO to have proper fixture with setup/teardown. if one test fails, cleanup is skipped.

slbotbm: done

foreign/cpp/Cargo.toml
slbotbm: used 1.49.0 as used in the project's Cargo.toml

foreign/cpp/tests/common/test_helpers.hpp
slbotbm: added asserts in test_helpers.hpp.

foreign/cpp/src/client.rs
slbotbm: Currently, we are dropping the client regardless of whether the logout succeeds or not, and then returning the error. Is there something else that can be done to improve this?

foreign/cpp/Cargo.lock
slbotbm: I'll go with the first approach.

foreign/cpp/src/client.rs
hubcio: it's fine for now, just add TODO comment about it

slbotbm: doene

PR #2853, opened 2026-03-03
feat(shard): Implement shard router

krishvishal:
In the shard router each shard gets an inbox (flume channel) and a  vector of senders to all other shards (including itself). When a message arrives at any shard, the router:

  1. Decomposes the generic message to extract the operation and namespace
  2. Resolves the target shard - metadata ops go to shard 0, partition ops are looked up in the ShardsTable, unknown ops default to shard 0
  3. Enqueues a ShardFrame into the target shard's channel

The message pump (run_message_pump) drains the inbox and processes frames sequentially, ensuring all mutations on a given shard are serialized through a single async task.

  The ShardsTable trait abstracts partition ownership lookup with two implementations:
  - () : no-op for single-shard setups (simulator for now) 
  - PapayaShardsTable: lock-free concurrent map for multi-shard deployments

### DIFF
--- a/core/metadata/src/impls/metadata.rs
+++ b/core/metadata/src/impls/metadata.rs
@@ -25,8 +25,7 @@ use consensus::{
 };
 use iggy_common::{
     header::{
-        Command2, ConsensusHeader, GenericHeader, Operation, PrepareHeader, PrepareOkHeader,
-        RequestHeader,
+        Command2, ConsensusHeader, GenericHeader, PrepareHeader, PrepareOkHeader, RequestHeader,
     },
     message::Message,
 };
@@ -260,30 +259,7 @@ where
             message.header().command(),
             Command2::Request | Command2::Prepare | Command2::PrepareOk
         ));
-        let operation = message.header().operation();
-        // TODO: Use better selection, smth like greater or equal based on op number.
-        matches!(
-            operation,
-            Operation::CreateStream
-                | Operation::UpdateStream
-                | Operation::DeleteStream
-                | Operation::PurgeStream
-                | Operation::CreateTopic
-                | Operation::UpdateTopic
-                | Operation::DeleteTopic
-                | Operation::PurgeTopic
-                | Operation::CreatePartitions
-                | Operation::DeletePartitions
-                | Operation::CreateConsumerGroup
-                | Operation::DeleteConsumerGroup
-                | Operation::CreateUser
-                | Operation::UpdateUser
-                | Operation::DeleteUser
-                | Operation::ChangePassword
-                | Operation::UpdatePermissions
-                | Operation::CreatePersonalAccessToken
-                | Operation::DeletePersonalAccessToken
-        )
+        message.header().operation().is_metadata()
     }
 }
 
--- a/core/partitions/src/iggy_partitions.rs
+++ b/core/partitions/src/iggy_partitions.rs
@@ -514,12 +514,7 @@ where
             message.header().command(),
             Command2::Request | Command2::Prepare | Command2::PrepareOk
         ));
-        let operation = message.header().operation();
-        // TODO: Use better selection, smth like greater or equal based on op number.
-        matches!(
-            operation,
-            Operation::DeleteSegments | Operation::SendMessages | Operation::StoreConsumerOffset
-        )
+        message.header().operation().is_partition()
     }
 }
 


krishvishal:
> can you use `crossfire` instead of `flume`? its much more battle-tested than flume and seems to have better performance.

Okay, I will look into it.

core/shard/src/router.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// All messages — whether destined for a local or remote shard — are routed
+/// through the channel into the target shard's message pump.  This ensures
+/// that every mutation on a shard is serialized through a single point (the
+/// pump), preventing concurrent access from independent async tasks.
+impl<B, J, S, M, T, R> IggyShard<B, J, S, M, T, R>
+where
+    B: MessageBus,
+    T: ShardsTable,
+    R: Send,
```
krishvishal: @hubcio with crossfire. This code will change to

```rust
impl<B, J, S, M, T, R> IggyShard<B, J, S, M, T, R>
where
    B: MessageBus,
    T: ShardsTable,
    R: Send + 'static,
{
```

`crossfire` requires additional `'static` bound. Which maybe too restrictive. I'm thinking if it would cause problems for us or not. Let me know what do you think of this.

core/shard/src/lib.rs
```diff
@@ -91,11 +170,11 @@ where
             >,
         M: StateMachine<Input = Message<PrepareHeader>>,
     {
-        let planes = self.plane.inner();
-        if planes.0.is_applicable(&request) {
-            planes.0.on_request(request).await;
+        let (metadata, (partitions, _)) = self.plane.inner();
+        if metadata.is_applicable(&request) {
```
numinnex: Why we do it this way ? that's the purpose of `MuxPlane` to not have that logic exposed inside of the top-level caller

core/shard/src/router.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn dispatch(&self, message: Message<GenericHeader>) {
+        let (operation, namespace, generic) = Self::decompose(&message);
+        let target = self.resolve(operation, IggyNamespace::from_raw(namespace));
+        let _ = self.senders[target as usize].send(ShardFrame::fire_and_forget(generic));
+    }
+
+    /// Dispatch a message and return a receiver that resolves when the target
+    /// shard has finished processing it.
+    pub fn dispatch_request(&self, message: Message<GenericHeader>) -> flume::Receiver<R> {
+        let (operation, namespace, generic) = Self::decompose(&message);
```
numinnex: Let's get rid of the `decompose` function, instead do it inline there and match inline

core/shard/src/router.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /// Determine which shard should handle a message given its operation and
+    /// namespace.
+    ///
+    /// - Metadata operations always route to shard 0 (the control plane).
+    /// - Partition operations route to the shard that owns the namespace,
+    ///   looked up via the [`ShardsTable`].
+    /// - Unknown operations fall back to shard 0.
+    fn resolve(&self, operation: Operation, namespace: IggyNamespace) -> u16 {
```
numinnex: Inline this aswell

core/shard/src/lib.rs
krishvishal: The MuxPlane's Plane trait dispatch can't work here because the two planes use different consensus types:  
1. `Metadata: VsrConsensus<B> (default LocalPipeline)`
2.  `Partitions: VsrConsensus<B, NamespacedPipeline>`                                                                                           
                             
The variadic Plane<C> impl requires a single C for both, so self.plane.on_request(...) doesn't compile without either:                       
  - Unifying both planes on `NamespacedPipeline` which changes metadata's pipeline behavior.
  - Adding a second `Plane impl for IggyPartitions` with the wrong pipeline type

  Both feel like they are hacky solutions.

core/shard/src/router.rs
hubcio: IMHO it's OK. shard (and messages router) should live forever in given thread so static lifetime fits perfectly. @numinnex any thoughts?

core/shard/src/router.rs
krishvishal: Done.

core/shard/src/router.rs
krishvishal: Done.

core/shard/src/router.rs
numinnex: I think it's fine, lets move to `crossfire`

krishvishal: Done.

core/shard/src/lib.rs
krishvishal: Done.

core/common/src/types/consensus/header.rs
```diff
@@ -124,6 +124,46 @@ pub enum Operation {
     Reserved = 200,
 }
 
+impl Operation {
+    /// Returns `true` for metadata / control-plane operations (streams, topics,
+    /// users, consumer groups, etc.) that are always handled by shard 0.
+    #[inline]
+    pub fn is_metadata(&self) -> bool {
```
numinnex, line 131: Replace both is_metadata and is_partition in the `PlaneIdentity` trait implementations, as those use exactly the same code.

krishvishal: Done.

numinnex (approved):

PR #2870, opened 2026-03-05
chore(consensus): enable pedantic and nursery clippy lints

hubcio:


core/consensus/Cargo.toml
```diff
@@ -38,3 +38,8 @@ rand_xoshiro = { workspace = true }
 
 [dev-dependencies]
 futures = { workspace = true }
+
+[lints.clippy]
+enum_glob_use = "deny"
+pedantic = "deny"
+nursery = "deny"
```
atharvalade: arent nursery lints explicitly unstable? clippy may rename, remove, or introduce false positives for them between releases. With "deny", any such upstream change becomes a hard build failure the next time the toolchain is bumped past 1.93.0. 

why dont you consider using "warn" here instead; you get the same enforcement by failing on warnings in CI, without the risk of a toolchain update unexpectedly breaking the consensus crate. I see `core/bench` already does the same, so it might be worth revisiting that one too.

hubcio: good point @atharvalade, i'll change this everywhere

numinnex (approved):

PR #2879, opened 2026-03-05
feat(server): Add integration test for message deduplicatpr

### LINKED ISSUE #2872: Add integration test for message deduplication
Message deduplication has unit tests in `MessageDeduplicator` but no integration test verifying the full pipeline: client sends duplicate messages -> server drops them at partition level.

**What to focus on:**
- Enable deduplication via server config (`system.message_deduplication.enabled = true`)
- Send messages with explicit IDs, then resend the same IDs
- Poll and verify only unique messages were persisted
- Test TTL expiry: after expiry window, previously seen IDs should be accepted again

**Files to reference:**
- `core/common/src/deduplication/message_deduplicator.rs` - core dedup logic + existing unit tests
- `core/common/src/types/message/messages_batch_mut.rs` - `prepare_for_persistence()` dedup integration
- `core/server/src/streaming/partitions/helpers.rs` - `create_message_deduplicator()` from config
- `core/integration/tests/server/scenarios/` - existing server scenario tests for pattern reference

**Related (closed):**
- #92 - closed when unit tests were added, no integration coverage
- #207 - TTL + max_entries config, implemented

Done when an integration test proves that duplicate messages are dropped end-to-end with deduplication enabled, and unique messages pass through.

tomighita:

Closes #2872

## Rationale

As stated in the issue description, the message deduplicator was lacking integration tests. This change adds one based on existing test harness.

## What changed?

The message deduplicator is only tested through unit tests, which means it lacks end to end verification. This PR adds one integration test, covering common scenarios like:
- Add messages with new ids
- Add messages with existing ids
- Add messages where some of the ids exist
- Add messages after the TTL expired and ensure we can insert duplicate ids

This change also fixed two minor bugs where sending a batch with all ids duplicated panicked and one where the deduplicator was not instantiated and `None` was always set.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools: copilot cli
2. Scope of usage: repo onboarding, explaining concepts and architecture and generating test scaffold
3. How did you verify the generated code works correctly: ran the test
4. Can you explain every line of the code if asked: yes


core/server/src/shard/system/messages.rs
```diff
@@ -356,6 +356,10 @@ impl IggyShard {
             )
             .await;
 
+        if batch.count() == 0 {
+            return Ok(());
+        }
+
```
tomighita, line 362: [note] This was causing a bug when all the messages were with duplicate IDs because the batch was empty and we unwraped on first

core/server/src/shard/system/partitions.rs
```diff
@@ -280,13 +281,15 @@ impl IggyShard {
...
+        let message_deduplicator = create_message_deduplicator(&self.config.system).map(Arc::new);
+
         let partition = LocalPartition::with_log(
             loaded_log,
             stats,
             std::sync::Arc::new(std::sync::atomic::AtomicU64::new(current_offset)),
             consumer_offsets,
             consumer_group_offsets,
-            None,
+            message_deduplicator,
```
tomighita, line 292: [note] I've added this here since the message_deduplicator was always None

core/integration/tests/server/scenarios/message_deduplication_scenario.rs
hubcio, line 1: you never test id 0 (auto-generated ids). the server replaces 0 with a uuid before the dedup check (messages_batch_mut.rs:164-166), so sending multiple messages with id 0 should all pass through. without this, you could silently break the default send path and never know. after that, please also poll the messages to check whether they have unique message id.

core/server/src/shard/system/messages.rs
```diff
@@ -356,6 +356,10 @@ impl IggyShard {
             )
             .await;
 
+        if batch.count() == 0 {
+            return Ok(());
+        }
```
hubcio, line 361: i get that it's a fix, but you didn't test it. nothing in this test actually exercises it. you'd need a step that sends a batch where every message is a duplicate (e.g. resend ids 1-10 a third time without any new ids mixed in) and verify the server doesn't blow up and the message count stays the same.

core/integration/tests/server/scenarios/message_deduplication_scenario.rs
```diff
@@ -0,0 +1,187 @@
...
+        )
+        .await
+        .unwrap();
+
+    let polled = poll_all(&client, MESSAGES_PER_BATCH * 4).await;
+    assert_eq!(
+        polled.messages.len() as u32,
+        total_after_step3,
+        "Only new IDs from the mixed batch should have been accepted"
+    );
```
hubcio, line 116: steps 3 and 4 only check message counts but never verify payloads. step 1 does this correctly (checks for "original" prefix), but step 3 doesn't verify the 5 surviving messages are actually the new ones (ids 21-25 with "mixed" payload) andnot some other combination. same for step 4 -- should check "after-ttl" payloads. count-only assertions can pass even if the wrong messages are kept.

core/integration/tests/server/scenarios/message_deduplication_scenario.rs
```diff
@@ -0,0 +1,187 @@
...
+        )
+        .await
+        .unwrap();
+
+    let polled = poll_all(&client, MESSAGES_PER_BATCH * 2).await;
+    assert_eq!(
+        polled.messages.len() as u32,
+        MESSAGES_PER_BATCH,
+        "Duplicate messages should have been dropped"
+    );
```
hubcio, line 62: after dedup removes messages mid-batch, offsets and indexes get rebuilt via remove_messages(). you never verify that the polled messages actually have correct, contiguous offsets. a bug in the offset recalculation would be a silent data corruption issue and this test wouldn't catch it.

core/integration/tests/server/scenarios/message_deduplication_scenario.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    // Step 3: Partial deduplication
+    // Send a batch where half the IDs are duplicates and half are new.
+    // IDs `overlap_id_start`..(`overlap_id_start`+`MESSAGES_PER_BATCH`):
+    // IDs `overlap_id_start`..(`MESSAGES_PER_BATCH`*2) already exist (from Step 2),
+    // IDs (`MESSAGES_PER_BATCH`*2+1)..(`overlap_id_start`+`MESSAGES_PER_BATCH`) are new.
+    let overlap_id_start = 16;
+    let total_after_step3 = MESSAGES_PER_BATCH + overlap_id_start - 1;
+    // Ensure the `overlap_id_start` is between two batches.
+    assert!(overlap_id_start > MESSAGES_PER_BATCH && overlap_id_start < 2 * MESSAGES_PER_BATCH);
```
hubcio, line 99: overlap_id_start = 16 is a magic number and the formula `total_after_step3 = MESSAGES_PER_BATCH + overlap_id_start - 1` only works because ids happen to be contiguous from 1. consider deriving it from the actual unique id range, e.g. `let new_ids_in_mixed = (overlap_id_start + MESSAGES_PER_BATCH) - (2 * MESSAGES_PER_BATCH + 1)` or just computing the set difference explicitly. also `build_messages` uses 0-based `i` in the payload instead of the actual message id, which makes correlating payloads to ids harder than it needs to be.

core/server/src/shard/system/messages.rs
tomighita: Isn’t this exactly what the step does where all ids are overlapping?

hubcio: you're right, step 1 does hit that path - my bad, sorry. still, it'd be nice to have it as its own explicit step so it's obvious we're regression-testing that specific fix, not just as a side effect of the dedupcheck.

PR #2884, opened 2026-03-06
refactor(go): upgrade go-retry to v5 and enable infinite reconnection

chengxilo:

N/A

## Rationale

Upgrade retry-go to v5, since I just realized that v3 will crash when the  WithAttempts() receive a big uint value, e.g. `math.MaxUint32`.

Also, to align with the rust client behavior, go sdk should have infinite retry for reconnection.

## What changed?

1. retry-go version upgrade to v5.
2. Default attempts  set to 0, which means infinite for `go-retry`

## Local Execution

- Passed  
- Pre-commit hooks ran 

## AI Usage

None


mmodzelewski (approved):

foreign/go/client/tcp/tcp_core.go
```diff
@@ -335,7 +335,10 @@ func (c *IggyTcpClient) connect() error {
 	}
 	// TODO handle tls logic
 	var conn net.Conn
-	if err := retry.Do(
+	if err := retry.New(
```
atharvalade, line 339: retry-go v5 defaults to exponential backoff with jitter. With a 2s base delay and infinite retries, the gap between attempts grows fast: 2s, 4s, 8s, 16s... after about 10 failures you're waiting 17 minutes between attempts, and after 15 you're waiting 9 hours. This defeats the purpose of infinite reconnection. The Rust client this is meant to align with uses a fixed interval (see `tcp_client.rs:329`), so you should add `retry.DelayType(retry.FixedDelay)` here. Alternatively, if you want to keep exponential backoff, at least add a `retry.MaxDelay(...)` cap like the C# client does (30s).

chengxilo: Thanks for the suggestion. Problem solved :)

PR #2887, opened 2026-03-07
fix(consensus): prevent UB from untrusted bytes interpreted as enum discriminants in consensus headers                                                                                    

### LINKED ISSUE #2878: unsafe impl Pod is unsound for consensus headers containing Command2 and Operation enums
### Bug description

`bytemuck::Pod` requires that every possible bit pattern is a valid instance of the type.

  - `Command2` defines discriminants `0..=12` only (13 of 256 possible u8 values)
  - `Operation` defines discriminants `0, 128..=147, 160..=161, 200` (23 of 256 possible u8 values)

When `bytemuck::from_bytes::<PrepareHeader>(&buf)` is called on a network buffer containing e.g. command = `0xFF`, Rust materializes a Command2 with an invalid discriminant. This is UB and the compiler is free to assume enum fields always hold valid discriminants. 

### Deployment

Not applicable

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

```rust
use bytemuck::{Pod, Zeroable};

/// Only discriminants 0, 1, 2 are valid.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
#[repr(u8)]
enum Command {
    Ping = 0,
    Pong = 1,
    Request = 2,
}

#[derive(Debug, Clone, Copy)]
#[repr(C)]
struct Header {
    command: Command,
}

unsafe impl Zeroable for Header {}
unsafe impl Pod for Header {}

fn main() {
    let buf: [u8; 1] = [255];

    let header: &Header = bytemuck::from_bytes(&buf);

    match header.command {
        Command::Ping => println!("ping"),
        Command::Pong => println!("pong"),
        Command::Request => println!("request"),
    }

    println!("command == Ping? {}", header.command == Command::Ping);
    println!("raw byte: 0x{:02X}", header.command as u8);
}
```

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

Try this MWE on Rust Playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2e4f80fe9eb435682b0ede6b7c3c95b8

And run the above example using Miri to see it throw error.

krishvishal:
Closes #2878

## Rationale

Consensus headers previously stored `Command2` and `Operation` enum values directly in repr(C) structs. This is unsound for consensus header  structs: arbitrary bytes from the network are reinterpreted as enum discriminants without validation, which is undefined behavior in Rust.

## What changed?

  - Change `command` and `operation` fields in all consensus headers from enum types (`Command2`, `Operation`) to raw `u8` to prevent UB when interpreting untrusted bytes from the network as Rust enum discriminants.
  - Add `TryFrom<u8>` impls for `Command2` and `Operation` with proper error variants (`InvalidCommandByte`, `InvalidOperationByte`) in `ConsensusError`.
  - Fix `dispatch_request` to return `Result<Receiver<R>, ConsensusError>` instead of silently dropping unrecognized messages.

### DIFF
--- a/core/common/src/types/consensus/header.rs
+++ b/core/common/src/types/consensus/header.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use bytemuck::{CheckedBitPattern, NoUninit, Pod, Zeroable};
+use bytemuck::{CheckedBitPattern, NoUninit};
 use enumset::EnumSetType;
 use thiserror::Error;
 
@@ -63,39 +63,10 @@ unsafe impl CheckedBitPattern for Command2 {
     }
 }
 
-impl TryFrom<u8> for Command2 {
-    type Error = u8;
-
-    fn try_from(value: u8) -> Result<Self, Self::Error> {
-        match value {
-            0 => Ok(Command2::Reserved),
-            1 => Ok(Command2::Ping),
-            2 => Ok(Command2::Pong),
-            3 => Ok(Command2::PingClient),
-            4 => Ok(Command2::PongClient),
-            5 => Ok(Command2::Request),
-            6 => Ok(Command2::Prepare),
-            7 => Ok(Command2::PrepareOk),
-            8 => Ok(Command2::Reply),
-            9 => Ok(Command2::Commit),
-            10 => Ok(Command2::StartViewChange),
-            11 => Ok(Command2::DoViewChange),
-            12 => Ok(Command2::StartView),
-            _ => Err(value),
-        }
-    }
-}
-
 #[derive(Debug, Clone, Error, PartialEq, Eq)]
 pub enum ConsensusError {
     #[error("invalid command: expected {expected:?}, found {found:?}")]
     InvalidCommand { expected: Command2, found: Command2 },
-
-    #[error("invalid command byte: 0x{0:02X} is not a valid Command2 discriminant")]
-    InvalidCommandByte(u8),
-
-    #[error("invalid operation byte: 0x{0:02X} is not a valid Operation discriminant")]
-    InvalidOperationByte(u8),
 
     #[error("invalid size: expected {expected:?}, found {found:?}")]
     InvalidSize { expected: u32, found: u32 },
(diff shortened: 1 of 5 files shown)


core/common/src/types/consensus/header.rs
```diff
@@ -449,19 +517,21 @@ impl ConsensusHeader for PrepareOkHeader {
     const COMMAND: Command2 = Command2::PrepareOk;
 
     fn operation(&self) -> Operation {
```
numinnex, line 452: In the solution that I've provided we wouoldn't need to use the `TryFrom`, as we would store the enums directly there, but rather the transistion/creation part is where validation would happen.

core/common/src/types/consensus/header.rs
```diff
@@ -449,19 +517,21 @@ impl ConsensusHeader for PrepareOkHeader {
...
-        self.operation
+        Operation::try_from(self.operation).expect("validate() must be called first")
     }
     fn command(&self) -> Command2 {
```
numinnex, line 455: Same for command

core/common/src/types/consensus/header.rs
```diff
@@ -124,6 +153,40 @@ pub enum Operation {
...
+            141 => Ok(Operation::CreateUser),
+            142 => Ok(Operation::UpdateUser),
+            143 => Ok(Operation::DeleteUser),
+            144 => Ok(Operation::ChangePassword),
+            145 => Ok(Operation::UpdatePermissions),
+            146 => Ok(Operation::CreatePersonalAccessToken),
+            147 => Ok(Operation::DeletePersonalAccessToken),
+            160 => Ok(Operation::SendMessages),
+            161 => Ok(Operation::StoreConsumerOffset),
+            200 => Ok(Operation::Reserved),
```
numinnex: We have both `Operation::Default` and `Operation::Reserved`, I don't think so there is a good reason for that, rather we should make Default = Reserved

numinnex: I see, that `Default` is used by the `fn operation(&self) -> Operation` as default implementation, I think we should return reserved in that case as that's what it's purpose is.

core/common/src/types/consensus/header.rs
krishvishal: Done.

core/common/src/types/consensus/header.rs
krishvishal: Done.

core/common/src/types/consensus/header.rs
krishvishal: Done.

core/common/src/types/consensus/header.rs
```diff
@@ -201,14 +276,15 @@ impl ConsensusHeader for GenericHeader {
     const COMMAND: Command2 = Command2::Reserved;
 
     fn operation(&self) -> Operation {
-        Operation::Default
+        Operation::Reserved
     }
 
     fn command(&self) -> Command2 {
```
numinnex, line 217: Do we need to store teh command as u8 on the header, instead of `Command2` enum ?

numinnex: We construct the header thgrough the `try_from_bytes` method from bytemuck, which has the validation

krishvishal: Done.

PR #2889, opened 2026-03-07
feat(connectors): implement Delta Lake Sink Connector

### LINKED ISSUE #1852: Implement Delta Lake connectors

kriti-sc:

Closes #1852 

## Rationale

Delta Lake is a data analytics engine, and very popular in modern streaming analytics architectures.

## What changed?

Introduces a Delta Lake Sink Connector that enables writing data from Iggy to Delta Lake.

The Delta Lake writing logic is heavily inspired by the [kafka-delta-ingest project](https://github.com/delta-io/kafka-delta-ingest/tree/main), to have a proven starting ground for writing to Delta Lake.

## Local Execution
1. Produced 32632 messages with schema `user_id: String, user_type: u8, email: String, source: String, state: String, created_at: DateTime<Utc>, message: String` using [sample data producer](https://github.com/apache/iggy/tree/master/examples/rust/src/sink-data-producer).
2. Consumed messages using the Delta Lake sink and created a Delta table on filesystem.
3. Verified number of rows in delta table and the schema. 
4. Added unit tests and e2e tests, both passing.

image
Left: messages produced; Right(top): messages consumed by Delta sink; Right(bottom): Inspecting Delta table in python

## AI Usage

1. Which tools? Claude Code
5. Scope of usage? generated functions
6. How did you verify the generated code works correctly? Manual testing by producing data into Iggy and then running the sink and verifying local Delta Lake creation, unit tests and e2e tests for local Delta Lake and Delta Lake on S3.
7. Can you explain every line of the code if asked? Yes


core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,277 @@
...
+            let secret_key = config
+                .aws_s3_secret_key
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
+            let region = config.aws_s3_region.as_ref().ok_or(Error::InvalidConfig)?;
+            let endpoint_url = config
+                .aws_s3_endpoint_url
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
+            let allow_http = config.aws_s3_allow_http.ok_or(Error::InvalidConfig)?;
```
atharvalade: `aws_s3_endpoint_url` and `aws_s3_allow_http` are hard-required via `.ok_or(Error::InvalidConfig)?`, but both are optional in `delta-rs/object_store`. `AWS_ENDPOINT_URL` defaults to the standard AWS S3 regional endpoint when omitted, and `AWS_ALLOW_HTTP` defaults to false. This means users connecting to real AWS S3 (not MinIO/LocalStack) are forced to provide values that shouldn't be necessary. These two fields should be added to the options map only when present, not treated as required.

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,277 @@
...
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
+            let account_key = config
+                .azure_storage_account_key
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
+            let sas_token = config
+                .azure_storage_sas_token
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
```
atharvalade: Both `azure_storage_account_key` and `azure_storage_sas_token` are required simultaneously via `.ok_or(Error::InvalidConfig)?`. In Azure, these are alternative authentication methods. you use either an account key or a SAS token, not both. This blocks users who only have a SAS token (common in restricted-access scenarios). The code should require `account_name` and at least one of `account_key` or `sas_token`, and only insert whichever credential is provided.

atharvalade (changes requested):

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+            if let Some(value) = context.get_mut(field_name) {
+                apply_coercion(value, coercion);
+            }
+        }
+    }
+}
+
+fn apply_coercion(value: &mut Value, node: &CoercionNode) {
+    match node {
+        CoercionNode::Coercion(Coercion::ToString) => {
```
krishvishal, line 107: SQL semantics treat NULL as a special type and its never coerced to any other type in most cases.

This can lead to subtle but hard to debug bugs.

kriti-sc: You are right. Thank you for catching this, fixing this in the next commit. I had misunderstood your point earlier.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        // Write JSON values to internal Parquet buffers
+        state.writer.write(json_values).await.map_err(|e| {
+            error!("Failed to write to Delta writer: {e}");
+            Error::Storage(format!("Failed to write to Delta writer: {e}"))
+        })?;
+
+        // Flush buffers to object store and commit to Delta log
+        let version = match state.writer.flush_and_commit(&mut state.table).await {
```
hubcio, line 151: every `consume()` call does `flush_and_commit`, which creates a new parquet file and a new JSON transaction log entry in `_delta_log/`. at high throughput with, say, 1000-message batches at 1M ops/sec, that's 1000 parquet files and 1000 log entries per second. delta lake degrades catastrophically under this - metadata parsing slows, checkpoint overhead grows, and cloud object store LIST calls become a bottleneck. the sink needs a buffering strategy: accumulate across multiple `consume()` calls and flush on a configurable row count, byte threshold, or time interval - not per batch.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+    }
+
+    CoercionTree { root }
+}
+
+fn build_coercion_node(data_type: &DataType) -> Option<CoercionNode> {
+    match data_type {
+        DataType::Primitive(primitive) => match primitive {
+            PrimitiveType::String => Some(CoercionNode::Coercion(Coercion::ToString)),
+            PrimitiveType::Timestamp => Some(CoercionNode::Coercion(Coercion::ToTimestamp)),
```
hubcio: `build_coercion_node` handles `PrimitiveType::Timestamp` but not `PrimitiveType::TimestampNtz`. delta lake has both `Timestamp` (microseconds, UTC-adjusted) and `TimestampNtz` (microseconds, no timezone). if a delta table uses `TimestampNtz` columns and receives ISO 8601 strings, they pass through uncoerced and `JsonWriter` receives a string where it expects an i64 - causing a write failure. fix: add `PrimitiveType::TimestampNtz` alongside `Timestamp` in this match.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+                apply_coercion(value, coercion);
+            }
+        }
+    }
+}
+
+fn apply_coercion(value: &mut Value, node: &CoercionNode) {
+    match node {
+        CoercionNode::Coercion(Coercion::ToString) => {
+            if !value.is_string() {
```
hubcio: when a value is JSON `null` and the delta column is `String`, `value.to_string()` produces the literal 4-char string `"null"` rather than preserving null. downstream consumers reading the delta table cannot distinguish actual nulls from the string `"null"`. the test at line 512-514 asserts this behavior, but it's semantically wrong - JSON null should map to SQL NULL. fix: add `if value.is_null() { return; }` before the `is_string()` check.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+            Ok(v) => v,
+            Err(e) => {
+                state.writer.reset();
```
hubcio, line 154: when `write()` succeeds (line 133) but `flush_and_commit()` fails here, `reset()` clears the internal parquet buffer. those messages are permanently lost with no retry path. since the connector runtime uses `AutoCommitWhen::PollingMessages` (offset committed before consume), the consumer offset may already have been advanced past these messages. there's no retry, no DLQ, no metric - the messages just vanish.

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,277 @@
...
+            let allow_http = config.aws_s3_allow_http.ok_or(Error::InvalidConfig)?;
+
+            opts.insert("AWS_ACCESS_KEY_ID".into(), access_key.clone());
+            opts.insert("AWS_SECRET_ACCESS_KEY".into(), secret_key.clone());
+            opts.insert("AWS_REGION".into(), region.clone());
+            opts.insert("AWS_ENDPOINT_URL".into(), endpoint_url.clone());
+            opts.insert("AWS_ALLOW_HTTP".into(), allow_http.to_string());
+            opts.insert("AWS_S3_ALLOW_HTTP".into(), allow_http.to_string());
+        }
+        Some("azure") => {
```
hubcio: the azure path requires all four fields: `account_name`, `account_key`, `sas_token`, and `container_name`. in practice, azure auth uses either an account key OR a SAS token, never both. requiring both means users who only have a SAS token (common in restricted environments) can't use the connector. either make these two auth fields alternatives with validation, or make each optional and pass through whichever is provided.

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,277 @@
...
+    config: &DeltaSinkConfig,
+) -> Result<HashMap<String, String>, Error> {
+    let mut opts = HashMap::new();
+
+    match config.storage_backend_type.as_deref() {
+        Some("s3") => {
+            let access_key = config
+                .aws_s3_access_key
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
```
hubcio: also at lines 37, 38, 42, 43, 56, 60, 64, 68, 80, 89.

every `.ok_or(Error::InvalidConfig)?` returns the same variant with no indication of which field is missing or which backend was selected. a user who misconfigures one field gets `"Invalid config"` with zero diagnostic. use `Error::InitError(format!("S3 backend requires 'aws_s3_access_key'"))` or similar so the error message actually tells the user what to fix.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+                }
+            }
+        }
+
+        if json_values.is_empty() {
+            debug!("No JSON values to write");
+            return Ok(());
+        }
+
+        let mut state_guard = self.state.lock().await;
```
hubcio, line 129: the tokio mutex is acquired here and held through both `writer.write()` (in-memory parquet encoding) and `writer.flush_and_commit()` (network/disk I/O + delta log commit). if `consume()` is called concurrently from multiple partitions (the SDK uses `&self` not `&mut self`), every caller serializes on this lock including the I/O wait. combined with per-batch flush, a slow object store response blocks all partitions.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+                *value = Value::String(value.to_string());
+            }
+        }
+        CoercionNode::Coercion(Coercion::ToTimestamp) => {
```
hubcio, line 112: when `string_to_timestamp` returns `None` (unparseable string), the value remains as a JSON string. `JsonWriter` then receives a string where it expects an i64 microsecond timestamp, causing `writer.write()` at sink.rs:133 to fail - rejecting the entire batch (potentially hundreds of valid messages) due to a single malformed timestamp in one message. either fail fast with an explicit error naming the field/value, or set the value to null (if the column is nullable) to preserve the rest of the batch.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+            Error::InvalidConfig
+        })?;
+
+        // Apply coercions to match Delta table schema
+        for value in &mut json_values {
+            coerce(value, &state.coercion_tree);
+        }
+
+        // Write JSON values to internal Parquet buffers
+        state.writer.write(json_values).await.map_err(|e| {
```
hubcio: if `JsonWriter::write()` partially buffers data before failing, the next `consume()` call will call `flush_and_commit` on a buffer containing stale/partial data from the failed batch. `reset()` is correctly done for `flush_and_commit` failures (line 142) but is missing here for `write` failures. either add `reset()` on write error, or verify in deltalake docs that `write` is atomic.

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,277 @@
...
+            let secret_key = config
+                .aws_s3_secret_key
+                .as_ref()
+                .ok_or(Error::InvalidConfig)?;
+            let region = config.aws_s3_region.as_ref().ok_or(Error::InvalidConfig)?;
+            let endpoint_url = config
```
hubcio: also line 43.

`endpoint_url` and `allow_http` are only needed for S3-compatible services (MinIO, LocalStack), not actual AWS S3. making them mandatory breaks the vanilla AWS use case. these should be optional with sensible defaults (no custom endpoint, HTTP disallowed).

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+                }
+            }
+        }
+    }
+}
+
+fn string_to_timestamp(string: &str) -> Option<Value> {
+    let parsed = DateTime::from_str(string);
+    if let Err(e) = parsed {
+        tracing::error!(
```
hubcio: an unparseable timestamp string is an expected data quality issue, not an infrastructure error. using `error!` level here will flood monitoring/alerting under normal operation with dirty data. downgrade to `warn!`.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+        }
+
+        if json_values.is_empty() {
+            debug!("No JSON values to write");
+            return Ok(());
+        }
+
+        let mut state_guard = self.state.lock().await;
+        let state = state_guard.as_mut().ok_or_else(|| {
```
hubcio, line 130: when `state` is `None` (meaning `open()` was not called or failed), returning `Error::InvalidConfig` is the wrong semantics - this is a lifecycle/invariant violation, not a config error. use `Error::InitError("Delta sink state not initialized - was open() called?".to_string())` to match the actual problem.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+            }
+        }
+        DataType::Array(array) => {
+            build_coercion_node(array.element_type()).and_then(|node| match node {
+                CoercionNode::Coercion(c) => Some(CoercionNode::ArrayPrimitive(c)),
+                CoercionNode::Tree(t) => Some(CoercionNode::ArrayTree(t)),
+                _ => None,
+            })
+        }
+        _ => None,
```
hubcio, line 85: `build_coercion_node` handles `Primitive`, `Struct`, and `Array` but falls through to `None` for `Map` and `Variant`. if a delta table has a `Map<String, Timestamp>` column, timestamp values inside the map won't be coerced. if intentional, add a comment documenting this as a known limitation so future maintainers know it's deliberate.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+fn owned_value_to_serde_json(value: &simd_json::OwnedValue) -> serde_json::Value {
+    match value {
+        simd_json::OwnedValue::Static(s) => match s {
+            simd_json::StaticNode::Null => serde_json::Value::Null,
+            simd_json::StaticNode::Bool(b) => serde_json::Value::Bool(*b),
+            simd_json::StaticNode::I64(n) => serde_json::Value::Number((*n).into()),
+            simd_json::StaticNode::U64(n) => serde_json::Value::Number((*n).into()),
+            simd_json::StaticNode::F64(n) => serde_json::Number::from_f64(*n)
```
hubcio: non-finite f64 values (NaN, Infinity) are silently mapped to `serde_json::Value::Null` here. for a delta column typed as `Double`, writing null where the source had NaN is data loss. if the column is non-nullable, this causes a confusing write failure downstream. add a `tracing::warn!` when this conversion happens so operators can diagnose data quality issues.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+use deltalake::kernel::Schema as DeltaSchema;
+use deltalake::kernel::{DataType, PrimitiveType};
+
+use chrono::prelude::*;
+use serde_json::Value;
+use std::collections::HashMap;
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq)]
+#[allow(unused)]
```
hubcio: all four `CoercionNode` variants are constructed and matched in the codebase. the `#[allow(unused)]` is unnecessary and should be removed.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,600 @@
...
+                    && let Some(value) = fields.get_mut(name)
+                {
+                    apply_coercion(value, node);
+                }
+            }
+        }
+        CoercionNode::ArrayPrimitive(coercion) => {
+            let values = value.as_array_mut();
+            if let Some(values) = values {
+                let node = CoercionNode::Coercion(coercion.clone());
```
hubcio, line 132: also line 133.

`coercion.clone()` and `tree.clone()` are called once per array-typed field per message. `CoercionTree` contains a `HashMap<String, CoercionNode>`, so cloning it for every array in every message is a heap allocation storm on wide schemas. refactor to pass `&Coercion` / `&CoercionTree` directly instead of wrapping in a new `CoercionNode`.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+                self.id
+            );
+            return Err(Error::Storage(format!("Failed to flush on close: {e}")));
+        }
+        info!("Delta Lake sink connector with ID: {} is closed.", self.id);
+        Ok(())
+    }
+}
+
+fn owned_value_to_serde_json(value: &simd_json::OwnedValue) -> serde_json::Value {
```
hubcio: this `simd_json::OwnedValue -> serde_json::Value` conversion is copy-pasted identically in `elasticsearch_sink/src/lib.rs:39`. should live in `iggy_connector_sdk` or a shared utility crate. when the conversion strategy changes, both need updating independently right now.

i commented same in https://github.com/apache/iggy/pull/2925

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        info!("Parsed table URI: {}", table_url);
+
+        let storage_options = build_storage_options(&self.config).map_err(|e| {
+            error!("Invalid storage configuration: {e}");
+            Error::InitError(format!("Invalid storage configuration: {e}"))
+        })?;
+
+        let table =
+            match deltalake::open_table_with_storage_options(table_url, storage_options.clone())
```
hubcio: `build_storage_options` returns an owned `HashMap` which is then `.clone()`-d just to pass to `open_table_with_storage_options`. since the original is never used again after this call, just pass the owned map directly and drop the clone.

core/connectors/sinks/delta_sink/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+#[derive(Debug)]
+struct SinkState {
+    table: DeltaTable,
+    writer: JsonWriter,
+    coercion_tree: CoercionTree,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
```
hubcio: `DeltaSinkConfig` derives `Debug`, which means AWS secret keys, Azure account keys, and GCS service account keys will appear in debug output / logs. consider implementing `Debug` manually to redact sensitive fields, or wrapping them in a newtype that redacts on display.

core/integration/tests/connectors/fixtures/delta/fixture.rs
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    async fn create_table(table_uri: &str) -> Result<(), TestBinaryError> {
+        let columns = vec![
+            StructField::new("id", DataType::Primitive(PrimitiveType::Long), true),
+            StructField::new("name", DataType::Primitive(PrimitiveType::String), true),
+            StructField::new("count", DataType::Primitive(PrimitiveType::Integer), true),
+            StructField::new("amount", DataType::Primitive(PrimitiveType::Double), true),
+            StructField::new("active", DataType::Primitive(PrimitiveType::Boolean), true),
+            StructField::new("timestamp", DataType::Primitive(PrimitiveType::Long), true),
```
hubcio: the test fixture declares `timestamp` as `PrimitiveType::Long`, not `PrimitiveType::Timestamp`. this means all integration tests bypass the coercion logic entirely. an e2e test with a `Timestamp`-typed column would catch issues like the `TimestampNtz` gap and null-to-"null" coercion bug at the integration level.

core/connectors/sinks/delta_sink/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+[package.metadata.cargo-machete]
+ignored = ["dashmap", "once_cell"]
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+chrono = { workspace = true }
+dashmap = { workspace = true }
```
hubcio, line 39: also line 42.

`dashmap` and `once_cell` are listed in `[dependencies]` and then suppressed via `[package.metadata.cargo-machete] ignored`. if they're truly unused, remove them rather than suppressing the warning.

hubcio (changes requested):
overall good direction, but needs a bit refining. thanks for contribution @kriti-sc

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: agreed. added a TODO.

core/connectors/sinks/delta_sink/src/storage.rs
kriti-sc: fixed as part of an earlier comment by @atharvalade

core/connectors/sinks/delta_sink/src/storage.rs
kriti-sc: fixed as part of an earlier comment by @atharvalade

core/connectors/sinks/delta_sink/src/coercions.rs
kriti-sc: Fixed as part of an earlier comment by @krishvishal

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Agreed. Was planning to propose a DLQ strategy across sinks and sources. Is it ok to defer this concern for later? @hubcio

hubcio: sure

core/connectors/sinks/delta_sink/Cargo.toml
kriti-sc: Cannot do this within the scope of this PR. These are transitive dependencies. The `sink_connector!` macro in the SDK uses them. Let me know your thoughts @hubcio

core/integration/tests/connectors/fixtures/delta/fixture.rs
kriti-sc: Agree on fixing the schema type. Timestamp better reflects real usage. 

Disagree on the second part. Catching specific bugs like the TimestampNtz gap or null coercion is the unit test's responsibility. My rubric with the integration test is to verify the happy path end-to-end.

core/connectors/sinks/delta_sink/src/coercions.rs
kriti-sc: Setting a value to null when the input is not null effectively alters the original data, which we should not do. In this case, failing the operation is the correct approach.

That said and as you have pointed out, failure currently means that hundreds of otherwise valid rows are not ingested and due to the existing architecture, failed payloads are effectively “forgotten.” This is also not the right behavior.

But the forgotten payload problem needs to be solved using a sophisticated DLQ strategy. For ex by isolating invalid messages and routing those to DLQ, while ingesting the good valid messages.

Since DLQ is out of scope for this PR, resolving this comment for now. ~with a TODO in the code~ TODO not needed as DLQ will attach to the writer.write() call.

Let me know thoughts @hubcio

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: You are right -- NaN to null conversion is data loss. I propose dropping the message containing non-finite values, for now. This again belong to the category of failures that are routed to DLQ. Let me know thoughts @hubcio

core/connectors/sinks/delta_sink/src/lib.rs
kriti-sc: Fixed using secrecy crate

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: there are two knobs `poll_interval` and `batch_size` to control delta write frequency and file size. Does this address your concern or are you thinking of a different issue @hubcio ?

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Fair point and thank you for catching this issue.
 
I have a fix in mind, but it is a bit more involved and I would prefer to keep it separate. I propose to highlight this issue as a TODO and handle it in a follow-up PR. Thoughts @hubcio?

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Additionally, added a TODO for implementing a retry strategy and metrics.

core/connectors/sinks/delta_sink/src/sink.rs
hubcio: of course, i'm 100% fine with either fixing now or documenting as TODO, for all the cases.

core/connectors/sinks/delta_sink/Cargo.toml
hubcio: in that case drop this issue

core/connectors/sinks/delta_sink/src/sink.rs
hubcio: fair point. lets leave it as is.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,322 @@
...
+            return Err(Error::Storage(format!("Failed to flush on close: {e}")));
+        }
+        info!("Delta Lake sink connector with ID: {} is closed.", self.id);
+        Ok(())
+    }
+}
+
+// TODO: Similar logic exists in the Elasticsearch sink.
+// Refactor to use a common utility when it is implemented.
+fn owned_value_to_serde_json(value: &simd_json::OwnedValue) -> Option<serde_json::Value> {
```
hubcio: you can use this from connectors SDK. see `core/connectors/sdk/src/convert.rs`

core/connectors/sinks/delta_sink/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+deltalake = { workspace = true }
+iggy_connector_sdk = { workspace = true }
+once_cell = { workspace = true }
+secrecy = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+simd-json = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
+url = "2"
```
hubcio: we track all deps versions in main `Cargo.toml` + dont use versions that are not pinned to major, use full semver

core/connectors/sinks/delta_sink/src/coercions.rs
hubcio: 1. arrow's string_to_datetime (arrow-cast/parse.rs:195) accepts ' ' as a timestamp separator- so "2021-11-11 22:11:58" passes Arrow but fails chrono's DateTime::from_str. for these strings the coercion is a no-op and you get a spurious warn! log, but the write succeeds.
2. for strings that BOTH parsers reject (e.g., "This definitely is not a timestamp"), Arrow fails at timestamp_array.rs:63-68 with ArrowError::JsonError, which fails record_batch_from_message and rejects the entire batch. these strings are left as-is by the coercion layer and then blow up downstream.

core/connectors/sinks/delta_sink/src/sink.rs
hubcio: don't drop the message - use the SDK's owned_value_to_serde_json from iggy_connector_sdk::convert (convert.rs:29). it maps non-finite floats to null instead of returning None. if the column is nullable, null is correct. if non-nullable, the Delta writer will reject it with a clear error,better than silently losing the entire message. IMHO its better this way

core/integration/tests/connectors/fixtures/delta/fixture.rs
hubcio: fair point, resolving

hubcio (changes requested):

core/connectors/sinks/delta_sink/src/coercions.rs
kriti-sc: Fair point. 

TLDR: The logic gives the desired behaviour, but there are paths in the code that are incorrect. I have made the change to fix these incorrect paths.

Desired behaviour: If timestamp can be correctly parsed, parse it and ingest. If not, do not convert to null and do not ingest the record with malformed timestamp. The latter would lead to entire batch failing, but this will be solved as part of DQ implementation. 

Fix: 

1. Ultimately, the timestamp with ' ' separator does get ingeted into Delta. Incorrect paths: we get a warning that coercion has failed but the write succeeds, which is incorrect. This is fixed. Now, coercion explicitly handles such timestamps. 

2. A record that contains an invalid timestamp will be rejected, causing the entire batch to fail. This is correct semantics. Incorrect path: coercion emits a warning but does not error out. The error happens downstream. This is incorrect. Now, for invalid timestamps error out at the coercion layer.

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: Also added tests for the SDK's owned_value_to_serde_json. Let me know if I should put the tests in a later PR instead to keep this one focussed @hubcio

hubcio: you can put tests in this PR. no need to create artificial split.

hubcio (dismissed):

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,327 @@
...
+use secrecy::ExposeSecret;
+use std::collections::HashMap;
+
+pub(crate) fn build_storage_options(
+    config: &DeltaSinkConfig,
+) -> Result<HashMap<String, String>, Error> {
+    let mut opts = HashMap::new();
+
+    match config.storage_backend_type.as_deref() {
+        Some("s3") => {
```
spetz: Let's use custom enum to avoid magic strings.

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,327 @@
...
+ */
+
+use crate::DeltaSinkConfig;
+use iggy_connector_sdk::Error;
+use secrecy::ExposeSecret;
+use std::collections::HashMap;
+
+pub(crate) fn build_storage_options(
+    config: &DeltaSinkConfig,
+) -> Result<HashMap<String, String>, Error> {
```
spetz, line 26: Maybe instead of returning a hash map, we could have the dedicated structs (per storage provider) with optional fields - and then either an enum could be returned or a single struct with all available storages as optional fields?

core/integration/tests/connectors/fixtures/delta/fixture.rs
```diff
@@ -0,0 +1,324 @@
...
+            StructField::new("amount", DataType::Primitive(PrimitiveType::Double), true),
+            StructField::new("active", DataType::Primitive(PrimitiveType::Boolean), true),
+            StructField::new(
+                "timestamp",
+                DataType::Primitive(PrimitiveType::TimestampNtz),
+                true,
+            ),
+        ];
+        let storage_options = HashMap::from([
+            ("AWS_ACCESS_KEY_ID".into(), MINIO_ACCESS_KEY.into()),
```
spetz, line 296: Same here (based on my previous suggestion) - we would use the dedicated option struct, and then simply serialize it.

spetz (changes requested):

core/connectors/sinks/delta_sink/src/storage.rs
kriti-sc: fixed

core/connectors/sinks/delta_sink/src/storage.rs
kriti-sc: Valid point, however this change would require the config to use nested structs. Because the mechanism to convert env vars into the config (function `apply_plugin_config_env_overrides` in runtime) does not support nested structs. 

It would be very good to support nested structs in config, but that change doesn't belong in this PR. I propose adding a TODO and following up with a PR that adds support for nested structs + changing Delta Sink config to use nested structs.

Let me know your thoughts @spetz

core/connectors/sinks/delta_sink/src/coercions.rs
numinnex: I think we should support those types from the get go, what is the reason why those were left as pass through ?

core/connectors/sinks/delta_sink/src/storage.rs
spetz: Yes, makes sense to do this in next iteration, as this will make tests easier and also help to avoid any typo-related bugs :)

spetz:
Please resolve the dependency conflicts, then run again the script `./scripts/ci/licenses-list.sh --update` and it should be ready to merge!

core/integration/tests/connectors/delta/delta_sink.rs
```diff
@@ -0,0 +1,282 @@
...
+        )
+        .await
+        .expect("Failed to send messages");
+
+    let version_count = fixture
+        .wait_for_delta_log(1, VERSION_POLL_ATTEMPTS, VERSION_POLL_INTERVAL_MS)
+        .await
+        .expect("Data should be written to Delta table");
+
+    assert!(version_count >= 1);
```
hubcio: `DeltaFixture::create_table` already writes `_delta_log/00000000000000000000.json` at setup (delta creates version 0 as part of establishing the table). so `count_delta_versions` returns 1 before the sink ever runs, `wait_for_delta_log(1, ...)` resolves immediately, and `assert!(version_count >= 1)` passes even if the sink wrote zero messages.

fix: either wait for `>= 2`, capture the baseline after `create_table` and wait for `baseline + 1`, or better - open the delta table and assert the row count.

core/integration/tests/connectors/delta/delta_sink.rs
```diff
@@ -0,0 +1,282 @@
...
+        )
+        .await
+        .expect("Failed to send messages");
+
+    let version_count = fixture
+        .wait_for_delta_log(1, BULK_VERSION_POLL_ATTEMPTS, VERSION_POLL_INTERVAL_MS)
+        .await
+        .expect("Data should be written to Delta table");
+
+    assert!(version_count >= 1);
```
hubcio: same issue as `delta_sink_consumes_json_messages` above - `count_delta_versions >= 1` is satisfied by the fixture's version-0 log entry before the sink runs. 100 messages sent, 0 messages written, still green. please gate on row count.

core/integration/tests/connectors/delta/delta_sink.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        last_error = sinks.first().and_then(|s| s.last_error.clone());
+        tokio::time::sleep(std::time::Duration::from_millis(interval_ms)).await;
+    }
+
+    assert!(
+        last_error.is_none(),
+        "Delta S3 sink had errors: {:?}",
+        last_error
+    );
```
hubcio: the s3 test polls the runtime api for `last_error.is_none()`. the runtime's `process_messages` discards the `i32` return from the FFI consume callback, so a per-batch `flush_and_commit` failure doesn't propagate to `last_error` (which is only set on task-level errors). combined with `AutoCommitWhen::PollingMessages` advancing the offset before consume completes, this test is green while messages are dropped.

read back from the delta table using `DeltaTableBuilder` with the minio storage options and assert the row count.

core/integration/tests/connectors/fixtures/delta/fixture.rs
```diff
@@ -0,0 +1,324 @@
...
+        ];
+        CreateBuilder::new()
+            .with_location(table_uri)
+            .with_columns(columns)
+            .await
+            .map_err(|error| TestBinaryError::FixtureSetup {
+                fixture_type: "DeltaFixture".to_string(),
+                message: format!("Failed to create Delta table: {error}"),
+            })?;
+        Ok(())
```
hubcio, line 143: worth noting: this call writes `_delta_log/00000000000000000000.json` (version 0) as part of table creation. any test assertion of `count_delta_versions >= 1` is trivially satisfied by this alone. capturing the baseline count here and exposing it on `DeltaFixture` would let tests assert `baseline + 1` meaningfully.

core/connectors/sinks/delta_sink/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+    Gcs,
+}
+
+mod coercions;
+mod sink;
+mod storage;
+
+use crate::coercions::CoercionTree;
+
+sink_connector!(DeltaSink);
```
hubcio, line 40: the `sink_connector!` macro expansion does `INSTANCES.insert(id, container)` in `core/connectors/sdk/src/sink.rs` without checking for an existing entry. a second `iggy_sink_open(id, ...)` with the same id overwrites the prior `SinkContainer`, dropping the old `DeltaTable` and `JsonWriter` without calling `close()` - any records buffered in the old writer are silently discarded.

the runtime doesn't currently double-open, but the macro is the public SDK contract and delta_sink is the first sink with real buffered state where this actually causes data loss. worth either rejecting same-id reopens in the macro or documenting the contract. SDK-level fix, not a blocker here,imo worth issue.

core/connectors/sinks/delta_sink/src/storage.rs
```diff
@@ -0,0 +1,313 @@
...
+            let bucket = config
+                .gcs_bucket
+                .as_ref()
+                .ok_or_else(|| Error::InitError("GCS backend requires 'gcs_bucket'".into()))?;
+
+            opts.insert(
+                "GOOGLE_SERVICE_ACCOUNT_KEY".into(),
+                service_account_key.expose_secret().to_owned(),
+            );
+            opts.insert("GCS_BUCKET".into(), bucket.clone());
```
hubcio: this option is silently dropped. `deltalake-gcp` (both 0.14.0 currently pinned and 0.16.0 latest) maps storage_options through `GoogleConfigKey::from_str(key.to_ascii_lowercase()).ok()?` inside a `filter_map`, and `object_store-0.12.5/src/gcp/builder.rs:197` only accepts `google_bucket | google_bucket_name | bucket | bucket_name` — `gcs_bucket` doesn't match and gets filtered out. bumping deltalake won't fix it; the rejection is in `object_store`. the bucket actually comes from the `gs://bucket/...` url via `GoogleCloudStorageBuilder::new().with_url(...)`.

rename to `google_bucket` (or just `bucket`) to match the upstream key, or drop the field + the readme entry. (azure's `AZURE_CONTAINER_NAME` at line 66 does work — `object_store-0.12.5/src/azure/builder.rs:459` accepts `azure_container_name | container_name`.)

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,179 @@
...
+            };
+
+        let kernel_schema = table
+            .snapshot()
+            .map_err(|e| {
+                error!("Failed to get table snapshot: {e}");
+                Error::InitError(format!("Failed to get table snapshot: {e}"))
+            })?
+            .schema();
+        let coercion_tree = create_coercion_tree(&kernel_schema);
```
hubcio, line 71: the coercion tree is computed once from the schema seen at `open()` and never refreshed. if another writer evolves the delta schema between `open()` and a later `consume()` (column added, type changed), coercion for new/changed fields won't fire and arrow will reject rows with an opaque `JsonError` - and combined with the data-loss path on `flush_and_commit` failure, those rows just vanish.

either refresh on schema version change or document that no concurrent writers are assumed.

core/connectors/sinks/delta_sink/src/sink.rs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            error!(
+                "Delta sink with ID: {} failed to flush on close: {e}",
+                self.id
+            );
+            return Err(Error::Storage(format!("Failed to flush on close: {e}")));
+        }
+        info!("Delta Lake sink connector with ID: {} is closed.", self.id);
+        Ok(())
+    }
```
hubcio, line 180: consume path calls `writer.reset()` before returning an error; close doesn't. consistency nit - not load-bearing since we're shutting down anyway.

core/integration/tests/connectors/delta/delta_sink.rs
```diff
@@ -0,0 +1,282 @@
...
+        .expect("Failed to get sinks");
+
+    assert_eq!(response.status(), 200);
+    let sinks: Vec<SinkInfoResponse> = response.json().await.expect("Failed to parse sinks");
+
+    assert_eq!(sinks.len(), 1);
+    assert_eq!(sinks[0].key, DELTA_SINK_KEY);
+    assert!(sinks[0].enabled);
+
+    drop(fixture);
```
hubcio: `fixture` is a function parameter, already dropped at scope exit. this pattern repeats at the end of all four tests - safe to remove.

core/integration/tests/connectors/fixtures/delta/fixture.rs
```diff
@@ -0,0 +1,324 @@
...
+                "minio/mc",
+                "mb",
+                "--ignore-existing",
+                &format!("minio/{}", MINIO_BUCKET),
+            ])
+            .output()
+            .map_err(|error| TestBinaryError::FixtureSetup {
+                fixture_type: "DeltaS3Fixture".to_string(),
+                message: format!("Failed to run mc command: {error}"),
+            })?;
```
hubcio, line 277: `std::process::Command::output()` blocks the tokio worker thread for the duration of the `docker run` call. use `tokio::process::Command` or wrap in `spawn_blocking`. integration-test only so not functional, but it serializes with other async work on the test runtime.

core/integration/tests/connectors/fixtures/delta/fixture.rs
```diff
@@ -0,0 +1,324 @@
...
+            StructField::new("name", DataType::Primitive(PrimitiveType::String), true),
+            StructField::new("count", DataType::Primitive(PrimitiveType::Integer), true),
+            StructField::new("amount", DataType::Primitive(PrimitiveType::Double), true),
+            StructField::new("active", DataType::Primitive(PrimitiveType::Boolean), true),
+            StructField::new(
+                "timestamp",
+                DataType::Primitive(PrimitiveType::TimestampNtz),
+                true,
+            ),
+        ];
```
hubcio: identical 6-column schema duplicated at lines 246-257 (`DeltaS3Fixture::create_table`). extract to a `fn test_columns() -> Vec<StructField>` helper.

hubcio (changes requested):
my biggest concern is that the integration tests would be green even if the sink wrote zero data (see count_delta_versions baseline). this is only true blocker from my side.

Cargo.toml
```diff
@@ -132,6 +133,7 @@ cyper = { version = "0.8.3", features = ["rustls"], default-features = false }
 cyper-axum = { version = "0.8.0" }
 darling = "0.23"
 dashmap = "6.1.0"
+deltalake = { version = "0.30.2", features = ["azure", "gcs", "s3"] }
```
hubcio: on crates latest is `0.32.0`. what's the reason for using older version?

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: This concern, while possible, is a narrow possibility, because JsonWriter would also be outdated if the schema changes and therefore new fields would just be ignored. This scenario would happen if the coercion tree was stale but state.tables got updated with the new schema. 

I have added a TODO to track this issue.

core/integration/tests/connectors/delta/delta_sink.rs
kriti-sc: I thought I was matching row count. My bad, fixed now.

core/integration/tests/connectors/fixtures/delta/fixture.rs
kriti-sc: this is fixed

core/connectors/sinks/delta_sink/src/sink.rs
kriti-sc: yes, we don't need `reset` here because we are using `take`. 

this is not the case in consume, where reset is needed to clear writer buffers so next consume call starts clean.

Cargo.toml
kriti-sc: No reason. 0.30.2 was the latest version when development started. 
The APIs we are using have remained unchanged and the latest version is 4 days old. So I suggest letting this be in case of patches to the newest version. 

Let me know.

core/connectors/sinks/delta_sink/src/lib.rs
kriti-sc: created an issue.

kriti-sc: created an issue.

Cargo.toml
kriti-sc: bumped it up.

atharvalade (approved):
lgtm

core/connectors/sinks/delta_sink/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[derive(Debug, Deserialize)]
+pub struct DeltaSinkConfig {
+    pub table_uri: String,
+    #[serde(default)]
+    pub storage_backend_type: Option<StorageBackendType>,
+
+    // AWS S3
+    #[serde(default)]
+    pub aws_s3_access_key: Option<String>,
```
hubcio: `aws_s3_access_key: Option<String>` should be `Option<SecretString>`. AWS access keys are credential material - derived `Debug` on `DeltaSinkConfig` leaks them to logs. the other secret fields (`aws_s3_secret_key`, `azure_storage_account_key`, `azure_storage_sas_token`, `gcs_service_account_key`) already use `SecretString`, apply the same treatment here for consistency.

core/integration/tests/connectors/fixtures/delta/fixture.rs
```diff
@@ -0,0 +1,391 @@
...
+fn table_columns() -> Vec<StructField> {
+    vec![
+        StructField::new("id", DataType::Primitive(PrimitiveType::Long), true),
+        StructField::new("name", DataType::Primitive(PrimitiveType::String), true),
+        StructField::new("count", DataType::Primitive(PrimitiveType::Integer), true),
+        StructField::new("amount", DataType::Primitive(PrimitiveType::Double), true),
+        StructField::new("active", DataType::Primitive(PrimitiveType::Boolean), true),
+        StructField::new(
+            "timestamp",
+            DataType::Primitive(PrimitiveType::TimestampNtz),
```
hubcio, line 126: column type was correctly changed to `TimestampNtz` per earlier review, but `TestMessage.timestamp` in `core/integration/tests/connectors/mod.rs:46,58` is still `i64` (epoch micros). the `i64` payload hits the `is_i64()/is_u64()` passthrough branch in `coercions.rs:115-117` - `string_to_timestamp` is never exercised end-to-end. add a test case (or new `TestMessageWithIsoTimestamp` variant) that sends `"2026-01-01T00:00:00Z"` into the same column so the coercion logic is actually integration-tested.

core/connectors/sinks/delta_sink/src/coercions.rs
```diff
@@ -0,0 +1,620 @@
...
+    // Arrow's parser also accepts ' ' as a date/time separator (e.g. "2021-11-11 22:11:58").
+    // Handle that here so the value is normalised to epoch microseconds rather than left as a
+    // raw string that Arrow would silently accept without going through this coercion layer.
+    let ndt = NaiveDateTime::parse_from_str(string, "%Y-%m-%d %H:%M:%S%.f")
+        .or_else(|_| NaiveDateTime::parse_from_str(string, "%Y-%m-%d %H:%M:%S"));
+    if let Ok(ndt) = ndt {
+        return Ok(Value::Number(ndt.and_utc().timestamp_micros().into()));
+    }
+
+    Err(format!("cannot parse \"{string}\" as a timestamp"))
```
hubcio: `cannot parse "{string}" as a timestamp` has no field-name context. with nested structs/arrays the operator can't tell which column had the bad value when this bubbles up via `Error::InvalidRecordValue`. propagate the field path (e.g. `level2.level2_timestamp`) by threading a path stack through `apply_coercion`. useful diagnostic in prod when bad records show up.

core/connectors/runtime/example_config/connectors/delta_sink.toml
```diff
@@ -0,0 +1,35 @@
...
+type = "sink"
+key = "delta"
+enabled = true
+version = 0
+name = "Delta Lake sink"
+path = "target/release/libiggy_connector_delta_sink"
+verbose = true
+
+[[streams]]
+stream = "qw"
```
hubcio: `stream = "qw"` and `topics = ["records"]` (line 28) read like leftover copy (probably `quickwit`?) placeholders. rename to descriptive defaults (e.g. `stream = "iggy_stream"`, `topics = ["delta_topic"]`) so users copying the example get something readable.

hubcio (changes requested):
i think its almost ready to merge, only nits from my side

core/connectors/sinks/delta_sink/src/coercions.rs
kriti-sc: that's a good point. this is done.

core/integration/tests/connectors/fixtures/delta/fixture.rs
kriti-sc: Your point is valid, but testing the `string_to_timestamp` branch is not an integration test concern. It is a unit test concern (and the unit test exists). IMO, integration tests responsibility is to ensure end-to-end wiring from Iggy topic into the delta lake. 

lmk @hubcio

hubcio: i'm fine with that 👍🏻

hubcio (approved):

spetz (approved):

core/connectors/sinks/delta_sink/src/coercions.rs
kriti-sc: @numinnex  -- This was left for later because `Map<String,Timestamp>` datatype would be rare, and `Variant` is a relatively new datatype for storing unstructured data. This gap is documented in a commnet.

mmmmxa:
Hi @kriti-sc, I'd like to work on some TODOs left in code. I am interested in this topic and want to try something relatively simple (like improving the existing implementation) first. Would you be okay and do you think there are any particular things I should try working on? Thank you
I have also tagged you with the same question in Discord, we can continue the chat there

kriti-sc:
Hi @mmmmxa,
For sure, feel free! Happy to help with questions if you have any. :)

mmmmxa:
Thank you @kriti-sc :)

PR #2897, opened 2026-03-09
feat(message_bus): add drain with routing metadata and backpressure

hubcio:
MessageBus had send_to_client/send_to_replica but no batched
retrieval for delivery. Add drain(&self, buf) following the
drain_loopback_into pattern. Each entry pairs the message
with a Recipient<C,R> enum so callers can route by target.
Outbox bounded by OUTBOX_CAPACITY_MAX with assert-on-push.


hubcio:
closing this, not good enough in terms of asynchronous reply/send to multiple sdk clients/replicas. need to think of better solution.

PR #2898, opened 2026-03-09
feat(go): add TLS integration tests and examples

### LINKED ISSUE #2807: Go SDK: Add TCP/TLS integration tests and examples
## Summary

The Go SDK has TCP/TLS support implemented in `foreign/go/client/tcp/tcp_core.go` with configurable TLS options, but there are no integration tests for TLS connections and no TLS examples in `examples/go/`.

## Current State

- **TLS implementation**: Present in `tcp_core.go` with config options:
  - `tlsEnabled` - Enable/disable TLS
  - `tlsDomain` - Domain for TLS (auto-extracts from address if not set)
  - `tlsCAFile` - Path to CA certificate file
  - `tlsValidateCertificate` - Toggle certificate validation (default: true)
- **Integration tests**: No TLS-specific integration tests exist (existing tests in `foreign/go/` only cover serialization and benchmarks)
- **Examples**: No TLS-specific examples in `examples/go/` (only basic consumer/producer examples exist)

### Integration Tests
- Add integration test(s) that start an Iggy server with TLS enabled and connect using the Go TCP client with TLS
- Use custom certificates from `core/certs/` (`iggy_ca_cert.pem`, `iggy_cert.pem`, `iggy_key.pem`)
- Test scenarios:
  - Successful TLS connection with custom CA certificate
  - Connection failure when TLS is required but client doesn't use TLS
  - Basic message send/receive over TLS connection

### Examples
- Add a TLS example in `examples/go/` demonstrating:
  - How to configure the TCP client with TLS enabled
  - How to specify a custom CA certificate file
  - A simple producer/consumer flow over TLS

## Reference

- Server TLS config env vars: `IGGY_TCP_TLS_ENABLED=true`, `IGGY_TCP_TLS_CERT_FILE`, `IGGY_TCP_TLS_KEY_FILE`
- Custom certs location: `core/certs/`
- C# TLS tests for reference: `foreign/csharp/Iggy_SDK.Tests.Integration/IggyTlsConnectionTests.cs`

saie-ch:
                                                                                                                                              
  Closes #2807

  ## Rationale

  The Go SDK lacked TLS integration tests to verify encrypted TCP connections work correctly. Python and other SDKs have similar tests using
  testcontainers for portable, self-contained testing.

  ## What changed?

  **Before:** Go SDK had no tests for TLS connections.

  **After:** Added 3 integration tests using testcontainers that verify:
  1. TLS connections with CA certificates succeed
  2. Non-TLS connections to TLS servers are rejected
  3. Complete message flows (create stream/topic, send/poll messages) work over TLS

Tests use `apache/iggy:edge` Docker image by default (overridable via `IGGY_SERVER_DOCKER_IMAGE` env var) with testcontainers-go v0.32.0. In CI, the image is built from the current PR code using `docker-build-test-server` action.

   Also:
   - Updated getting-started examples with TLS flags
   - Enhanced test script to run both non-TLS and TLS passes
   - Upgraded golangci-lint from v1.64.8 to v2.11.3 (required for Go 1.25 compatibility)
   - Upgraded golangci-lint-action from @v6 to @v7 (required for lint v2)
   - Separated unit tests from integration tests in CI (test vs e2e tasks)
   - Fixed connection leak in TLS config error path (tcp_core.go)


  **Breaking change:** Bumped Go requirement from 1.23.0 to 1.25.0 (required by testcontainers-go dependency). Go 1.25.0 was released 6 months ago and is stable.

  ## Local Execution

  All  TLS integration tests passed locally:
  Ran complete pre-merge check suite (format, tidy, lint, tests) - all passed.

  ## AI Usage

 Claude Code (claude-sonnet-4.5)


foreign/go/client/tcp/tcp_core.go
atharvalade, line 367: If `createTLSConfig() `fails here, the TCP connection opened by `net.Dial` a few lines above just gets abandoned. Nothing closes it. The handshake failure path right below already does `connection.Close() `but this early return was missed.

This gets worse because the retry loop defaults to infinite retries. So if the TLS config keeps failing (bad CA path, for example), every retry opens a new TCP socket and leaks it. You will eventually run out of file descriptors.

Add `_ = connection.Close()` before the return to match what the handshake error path already does.

foreign/go/tests/tls_test.go
```diff
@@ -0,0 +1,296 @@
...
+	require.NoError(t, err, "Certs directory not found at %s", certsPath)
+	_, err = os.Stat(caFile)
+	require.NoError(t, err, "CA cert not found at %s", caFile)
+
+	// Require locally built Docker image — never pull from DockerHub.
+	// In CI, the image is built by docker-build-test-server action.
+	// Locally, build with: cargo build && docker build -f core/server/Dockerfile ...
+	dockerImage := os.Getenv("IGGY_SERVER_DOCKER_IMAGE")
+	if dockerImage == "" {
+		t.Fatal("IGGY_SERVER_DOCKER_IMAGE env var is required. " +
```
atharvalade: Both the C# and Python SDKs fall back to `apache/iggy:edge `when `IGGY_SERVER_DOCKER_IMAGE` is not set. This hard-fails instead, which makes running these tests locally more cumbersome than in the other SDKs. Consider defaulting to `apache/iggy:edge` to match the established pattern across the project.

scripts/run-go-examples-from-readme.sh
```diff
@@ -169,9 +169,85 @@ cd ../..
...
+            break
+        fi
+        sleep 2
+    done
+
+    cd ../..
+
+    # Terminate TLS server
+    kill -TERM "$(cat ${PID_FILE})"
+    kill -TERM "$(cat ${PID_FILE})" 2>/dev/null || true
```
atharvalade: There are two back-to-back kill -TERM calls on the same PID. The first one has no error handling, and the second one has `2>/dev/null || true`. Since the script uses set -e, if the process already exited before line 241 runs, the first kill will fail and the script will abort before it even reaches the safe version on line 242.

atharvalade (changes requested):
overall lgtm but the `foreign/go/client/tcp/tcp_core.go‎` is critical imo rest are just for codebase consistency

mmodzelewski (approved):

hubcio (approved):

PR #2909, opened 2026-03-10
refactor(journal): refactor partition journal utilize storage trait

numinnex:
Refactor the `PartitionJournal` to use the `Strorage` trait as backing storage, rather than storing data inline.

### DIFF
--- a/core/partitions/src/iggy_partitions.rs
+++ b/core/partitions/src/iggy_partitions.rs
@@ -549,7 +549,7 @@ where
         let header = *message.header();
 
         // TODO: WE have to distinguish between an `message` recv by leader and follower.
-        // In the follower path, we have to skip the `prepare_for_persistance` path, just append to journal.
+        // In the follower path, we have to skip the `prepare_for_persistence` path, just append to journal.
         match header.operation {
             Operation::SendMessages => {
                 self.append_send_messages_to_journal(namespace, message)


core/partitions/src/journal.rs
```diff
@@ -49,89 +46,313 @@ impl std::ops::Deref for MessageLookup {
...
+        let bytes = {
+            let inner = unsafe { &*self.inner.get() };
+            inner.storage.read(offset, ZERO_LEN).await
+        };
+
+        Some(
+            Message::from_bytes(bytes)
+                .expect("partition.journal.storage.read: invalid bytes for message"),
+        )
+    }
```
atharvalade, line 261: `message_by_op` returns `Option` but panics for any op that isn't in storage. 

`PartitionJournalMemStorage::read` returns `Bytes::new()` for unknown ops, `Message::from_bytes` on empty bytes returns `Err`, and the `.expect()` panics. This is directly reachable from `load_messages_from_storage` which walks ops with `op += 1`; since consensus ops are globally scoped, so gaps per-partition are expected and will trigger this.

core/partitions/src/journal.rs
```diff
@@ -49,89 +46,313 @@ impl std::ops::Deref for MessageLookup {
...
+        }
+
+        let headers = unsafe { &mut *self.headers.get() };
+        headers.clear();
+        let offsets = unsafe { &mut *self.message_offset_to_op.get() };
+        offsets.clear();
+        let timestamps = unsafe { &mut *self.timestamp_to_op.get() };
+        timestamps.clear();
+
+        Self::messages_to_batch_set(&messages)
```
atharvalade, line 187: `commit()` drains storage destructively before parsing. If any `Message::from_bytes` call returns `Err`, the entry is dropped. The old code had no parsing step on this path, so this failure mode is new and unique to this refactor.

core/partitions/src/journal.rs
```diff
@@ -49,89 +46,313 @@ impl std::ops::Deref for MessageLookup {
...
+    S: Storage,
+{
+    type Query;
+
+    fn get(&self, query: &Self::Query) -> impl Future<Output = Option<IggyMessagesBatchSet>>;
+}
+
+#[derive(Debug, Default)]
+pub struct PartitionJournalMemStorage {
+    entries: UnsafeCell<Vec<Bytes>>,
```
krishvishal, line 61: I think `entries` can be made into `UnsafeCell<Vec<u8>>`  and store a hashmap in `PartitionJournal` which maps `op -> (offset, byte_length)` inside the above `Vec<u8>`.

This would simplify `impl read` and `write` of `PartitionJournalStorage`

core/partitions/src/journal.rs
numinnex: The goal of this API is to drain entries from journal without validating (panic if validation fails), because all of the validation happens before any entry enters the journal.  

I think the point is valid that in case of failed validation, we should make sure that we don't leave journal in inconsistent state, but the way we will achieve that is by crashing the server (if we fail parsing batch read from journal as the commit essentially is like reading from it, then something really bad happend)

core/partitions/src/journal.rs
numinnex: Yeah I agree, the only reason why the op_to_index mapping even exists there, is because we have to handle the `commit` path which resets the storage `offset`, therefore our op_number != storage_offset. I will move that translation to the `PartitionJournal` rather than having it in storage.

core/partitions/src/journal.rs
numinnex: Good catch, we shouldn't expect op to be monotonic linearly, as you've mentioned they are shared between partition messages and partition metadata operations, so the op_number could potentially be scattered.

core/partitions/src/journal.rs
numinnex: So I've actually realized that we cannot do that, the op_to_offset transistion needs to exist if we want to have an possibility to support storages that are disk-backed. The Storage API needs to work on the `offset`, but the on-disk offset != in-memory index, so the usage inside of `PartitionJournal` would differ (it would require different code for disk backed storage and in-memory one). If we would to keep it isomorphic, then we need to operate either on the disk based offset and have offset_to_index translation inside of in-memory storage, or the other way around -- operate on op_number and then for disk-based storage have an index_to_offset translation. 

Alternatively we could make the in-memory storage be an contiguous chunk of memory (Vec<u8>) and treat it the same way an block device (be sector addressable).

PR #2914, opened 2026-03-10
feat(bdd): support -race option for Go BDD tests

### LINKED ISSUE #2863: Support `-race` option for  `run-bdd-tests.sh/go`
### Description

Currently, `./scripts/run-bdd-tests.sh go` only runs standard tests. I propose adding support for the `-race` flag to detect data race in the Go SDK.

### Component

None

### Proposed solution

Regarding the bdd test script, we should update `run-bdd-tests.sh` so it can handle a `-race` argument:
* **Target Command:** `./scripts/run-bdd-tests.sh go -race`
* **Internal Execution:** Should run `go test -v -race -cpu 1,4 ./...` in `bdd/go/Dockerfile`

In workflow, we should update the matrix to run both versions in parallel for Go BDD test:
* **Job 1:** `go`
* **Job 2:** `go -race`

### Alternatives considered

_No response_

atharvalade:

Closes #2863

## Rationale

Enable data race detection for the Go SDK in BDD tests via `-race` flag.

## What changed?

Go BDD tests only ran in standard mode with no race detection. The script, Docker Compose service, and Dockerfile now propagate a `GO_TEST_EXTRA_FLAGS` env var so `./scripts/run-bdd-tests.sh go -race` runs `go test -v -race -cpu 1,4 ./...` inside the container. The CI matrix adds a parallel `bdd-go-race` job alongside the existing `bdd-go`.

## Local Execution

- Passed
- Pre-commit hooks ran

### DIFF
--- a/.github/actions/java-gradle/pre-merge/action.yml
+++ b/.github/actions/java-gradle/pre-merge/action.yml
@@ -146,6 +146,34 @@ runs:
         pid-file: ${{ steps.iggy.outputs.pid_file }}
         log-file: ${{ steps.iggy.outputs.log_file }}
 
+    - name: Start Iggy server (TLS)
+      id: iggy-tls
+      if: inputs.task == 'test'
+      uses: ./.github/actions/utils/server-start
+      with:
+        pid-file: ${{ runner.temp }}/iggy-server-tls.pid
+        log-file: ${{ runner.temp }}/iggy-server-tls.log
+      env:
+        IGGY_TCP_TLS_ENABLED: "true"
+        IGGY_TCP_TLS_CERT_FILE: core/certs/iggy_cert.pem
+        IGGY_TCP_TLS_KEY_FILE: core/certs/iggy_key.pem
+
+    - name: TLS integration tests
+      if: inputs.task == 'test'
+      shell: bash
+      working-directory: foreign/java
+      env:
+        USE_EXTERNAL_SERVER: true
+        IGGY_TCP_TLS_ENABLED: "true"
+      run: ./gradlew :iggy:test --tests "*TlsConnectionTest*"
+
+    - name: Stop Iggy server (TLS)
+      if: always() && inputs.task == 'test'
+      uses: ./.github/actions/utils/server-stop
+      with:
+        pid-file: ${{ steps.iggy-tls.outputs.pid_file }}
+        log-file: ${{ steps.iggy-tls.outputs.log_file }}
+
     - name: Test Summary
       uses: test-summary/action@v2
       with:
--- a/examples/java/README.md
+++ b/examples/java/README.md
@@ -174,6 +174,28 @@ The async client uses Netty's event loop threads for I/O operations. **NEVER** b
 
 If your message processing involves blocking operations, offload to a separate thread pool using `thenApplyAsync(fn, executor)`.
 
+## Security Examples
+
+### TCP/TLS
+
+Demonstrates secure TLS-encrypted TCP connections:
+
+```bash
+./gradlew runTcpTlsProducer
+./gradlew runTcpTlsConsumer
+```
+
+These examples require a TLS-enabled Iggy server. Start the server with:
+
+```bash
+IGGY_TCP_TLS_ENABLED=true \
+IGGY_TCP_TLS_CERT_FILE=core/certs/iggy_cert.pem \
+IGGY_TCP_TLS_KEY_FILE=core/certs/iggy_key.pem \
+cargo r --bin iggy-server
+```
+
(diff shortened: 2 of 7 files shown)


scripts/run-bdd-tests.sh
chengxilo:
```suggestion
    run_suite go-bdd     "🐹"   "Running Go BDD tests"     || exit $?
    GO_TEST_EXTRA_FLAGS="-race"
    run_suite go-bdd     "🐹"   "Running Go BDD tests -race"     || exit $?
```

Just a minor issue, I think we should run both -race and normal test when `./scripts/run-bdd-tests.sh all`.

I think my approach is weird tho)

chengxilo:
```suggestion
    run_suite go-bdd     "🐹"   "Running Go BDD tests"     || exit $?
    GO_TEST_EXTRA_FLAGS="-race"
    run_suite go-bdd     "🐹"   "Running Go BDD tests (${GO_TEST_EXTRA_FLAGS})"
```

I think both normal test and race test should be executed. Or probably we need to support something like `./scripts/run-bdd-tests.go all -race` if other languages also need to detect data race in bdd test.

atharvalade: Good point! Though I'm a bit hesitant since `all` means "all SDKs" and no other SDK gets a double-run there, it'd make the path Go-specific. A developer wanting race detection can already do `./scripts/run-bdd-tests.sh go -race` explicitly. What do you think, worth the inconsistency or keep it simple? @hubcio @mmodzelewski

scripts/run-bdd-tests.sh
```diff
@@ -50,7 +60,11 @@ run_suite(){
 case "$SDK" in
   rust)   run_suite rust-bdd   "🦀"   "Running Rust BDD tests"   ;;
   python) run_suite python-bdd "🐍"   "Running Python BDD tests" ;;
-  go)     run_suite go-bdd     "🐹"   "Running Go BDD tests"     ;;
+  go)
+    _go_label="Running Go BDD tests"
+    [[ -n "$GO_TEST_EXTRA_FLAGS" ]] && _go_label="Running Go BDD tests (${GO_TEST_EXTRA_FLAGS})"
+    run_suite go-bdd "🐹" "$_go_label"
+    ;;
```
chengxilo:
```suggestion
  go)     run_suite go-bdd     "🐹"   "Running Go BDD tests${GO_TEST_EXTRA_FLAGS:+ (${GO_TEST_EXTRA_FLAGS})}" ;;
```

This looks more elegant 😋 But just personal preference.

chengxilo:
#2863 might be a bit misleading, it's my fault since at that moment I thought it is a good idea to use a `-race` flag. However, after reading your code. Instead of introducing an option like `-race`, I think maybe we could add another $SDK choice such as `go-race`.

Discard `-race` option would simplify the implementation and make the script easier to read. And not every SDK needs data race detection (to be honest I think only go and C/C++ will need this). 

The structure could look something like this:
```bash
case "$SDK" in
  rust)    run_suite rust-bdd   "🦀"   "Running Rust BDD tests"   ;;
  python)  run_suite python-bdd "🐍"   "Running Python BDD tests" ;;
  go)      run_suite go-bdd     "🐹"   "Running Go BDD tests"     ;;
  go-race) run_suite go-bdd     "🐹⚡"  "Running Go BDD tests with data race detector" ;;
  node)    run_suite node-bdd   "🐢🚀" "Running Node BDD tests"   ;;
  csharp)  run_suite csharp-bdd "🔷"   "Running C# BDD tests"     ;;
  java)    run_suite java-bdd   "☕"   "Running Java BDD tests"   ;;
  all)
    run_suite rust-bdd   "🦀"   "Running Rust BDD tests"   || exit $?
    run_suite python-bdd "🐍"   "Running Python BDD tests" || exit $?
    run_suite go-bdd     "🐹"   "Running Go BDD tests"     || exit $?
    run_suite go-bdd     "🐹⚡"  "Running Go BDD tests with data race detector"  || exit $?
    run_suite node-bdd   "🐢🚀" "Running Node BDD tests"   || exit $?
    run_suite csharp-bdd "🔷"   "Running C# BDD tests"     || exit $?
    run_suite java-bdd   "☕"   "Running Java BDD tests"   || exit $?
    ;;
  clean)
    cleanup; exit 0 ;;
  *)
    log "❌ Unknown SDK: ${SDK}"
    log "📖 Usage: $0 [rust|python|go|node|csharp|java|all|clean] [-race] [feature_file]"
    exit 2 ;;
esac
```

What do you think?

scripts/run-bdd-tests.sh
atharvalade: will be resolved with the new refactor

scripts/run-bdd-tests.sh
atharvalade: will be resolved with the new refactor

mmodzelewski (approved):

PR #2916, opened 2026-03-11
feat(metadata): Persistent WAL journal with recovery and compaction

### LINKED ISSUE #2915: State loading from disk during replica bootup
Using Snapshot and replica metadata journal, rebuild the `State` during startup of the replica.

krishvishal:

Closes #2915

## Summary

- Add `FileStorage`: file-backed `Storage` impl with positional reads, appends, truncate, fsync.
- Add `MetadataJournal`: append-only WAL indexed by a ring buffer (`op % SLOT_COUNT`). Crash recovery scans forward and truncates partial tail entries. Compaction atomically rewrites the WAL keeping only entries above the snapshot watermark.
- Add `recover()`: startup recovery pipeline that loads the latest snapshot, opens the WAL, and replays entries past the snapshot sequence number through the state machine.
- Add `checkpoint()` on `IggyMetadata`: persists a snapshot then advances the journal watermark and compacts.
- Update `Journal` / `Storage` traits: `io::Result` return types, `set_snapshot_op`, `remaining_capacity`, `compact` default methods.


core/metadata/src/impls/metadata.rs
```diff
@@ -151,10 +198,36 @@ where
...
+        // Force a checkpoint if the journal is running low on capacity.
+        if journal
+            .handle()
+            .remaining_capacity()
+            .is_some_and(|c| c <= CHECKPOINT_MARGIN)
+        {
+            if let Some(data_dir) = &self.data_dir {
+                let snap_op = current_op;
+                if let Err(e) = self.checkpoint(data_dir, snap_op) {
+                    error!(
```
atharvalade: `snap_op` is set to `current_op`, which is the last prepared op, not the last committed op. When `checkpoint()` runs with this value it calls `set_snapshot_op` and `compact()`, which removes all WAL entries at or below that op.

But entries between `commit_number + 1` and `current_op` are still in the pipeline awaiting quorum. When `on_ack` later tries to look them up via `journal.handle().entry()` at line 300, it hits the `unwrap_or_else(|| panic!(...))` because the entries were already compacted away.

core/metadata/src/impls/metadata.rs
```diff
@@ -325,3 +404,31 @@ where
...
+        M: FillSnapshot<MetadataSnapshot>,
+        J: JournalHandle,
+    {
+        let snapshot = IggySnapshot::create(&self.mux_stm, last_op)?;
+        let path = data_dir.join(super::METADATA_DIR).join("snapshot.bin");
+        snapshot.persist(&path)?;
+
+        if let Some(journal) = &self.journal {
+            journal.handle().set_snapshot_op(last_op);
+            journal.handle().compact().map_err(SnapshotError::Io)?;
```
atharvalade: `IggySnapshot::create` labels the snapshot with `sequence_number = last_op` (which is `current_op` from the caller), but the state machine has only applied entries up to the commit number, not up to the latest prepare.

On recovery, `replay_from` is computed as `snapshot.sequence_number() + 1`, so entries between the real applied op and `current_op` get skipped and are never replayed into the state machine. Same root cause as issue 1: `last_op` should be the commit number.

core/journal/src/metadata_journal.rs
```diff
@@ -0,0 +1,686 @@
...
+    }
+
+    fn remaining_capacity(&self) -> Option<usize> {
+        let Some(last) = self.last_op.get() else {
+            return Some(SLOT_COUNT);
+        };
+        let snapshot = self.snapshot_op.get();
+        let used = (last - snapshot) as usize;
+        Some(SLOT_COUNT.saturating_sub(used))
+    }
```
atharvalade, line 309: `let used = (last - snapshot) as usize` will underflow if `last_op < snapshot_op`, which is possible after a partial WAL recovery where the snapshot is ahead of the truncated WAL (e.g. snapshot at op 100 but WAL only has entries up to op 50).

In debug builds this panics. In release builds it wraps to a huge value, `saturating_sub` returns 0, and an unnecessary forced checkpoint is triggered. I believe using `last.saturating_sub(snapshot)` or adding a guard that handles this case will help.

core/journal/src/file_storage.rs
```diff
@@ -0,0 +1,137 @@
...
+use crate::Storage;
+use compio::buf::IoBuf;
+use compio::io::{AsyncReadAtExt, AsyncWriteAtExt};
+use std::cell::{Cell, RefCell};
+use std::io;
+use std::path::{Path, PathBuf};
+
+/// File-backed storage implementing the `Storage` trait.
+pub struct FileStorage {
+    file: RefCell<compio::fs::File>,
```
numinnex: You cannot use `RefCell` together with components that have `async` methods.

core/journal/src/file_storage.rs
```diff
@@ -0,0 +1,137 @@
...
+    write_offset: Cell<u64>,
+    path: PathBuf,
+}
+
+#[allow(clippy::future_not_send, clippy::await_holding_refcell_ref)]
```
numinnex: Let's not silence the `await_holding_refcell_ref`, use an `UnsafeCell` instead of `RefCell`

core/journal/src/lib.rs
```diff
@@ -30,16 +34,40 @@ where
...
+    }
+
+    /// Remove snapshotted entries from the WAL to reclaim disk space.
+    /// The default is a no-op for journals that do not persist to disk.
+    ///
+    /// # Errors
+    /// Returns an I/O error if compaction fails.
+    fn compact(&self) -> impl Future<Output = io::Result<()>> {
+        async { Ok(()) }
+    }
```
numinnex, line 60: I think, I'd prefer the journal to have some sort of `drain` method that allows to "extract" range of items, simiarly to how `Vec::drain(begin..end)` works. This way we do not hack some apis on the interface just to cover an edge case, but we create a general purpose API, that can be used to shrink the journal and we handle the watermark outside of the `Journal`. 

And yeah an `Stream` iter would be perfect, but since `AsyncIterator` is still unstable and it's probably going to replace the `Stream` trait, we can return no-async drain iterator and do the disk read for the entire range in one go.

core/journal/src/lib.rs
```diff
@@ -30,16 +34,40 @@ where
...
 }
 
 // TODO: Move to other crate.
 pub trait Storage {
     type Buffer;
 
-    fn write(&self, buf: Self::Buffer) -> impl Future<Output = usize>;
-    fn read(&self, offset: usize, buffer: Self::Buffer) -> impl Future<Output = Self::Buffer>;
+    fn write(&self, buf: Self::Buffer) -> impl Future<Output = io::Result<usize>>;
```
numinnex: I was thinking, since I've added `offset` to the read method, shouldn't we rename this method to `write_at` - fn write_at(&self, offset: usize, buf: Self::Buffer)` and create on top of that  `fn append(&self, buf: Self::Buffer)`.  What do you think ? I am wondering which API would serve us better from the perspective of implementing an InMemory storage both for server usage, aswell as simulator.

core/metadata/src/impls/metadata.rs
```diff
@@ -52,6 +53,47 @@ impl IggySnapshot {
...
+
+        fs::rename(&tmp_path, path)?;
+
+        // Fsync the parent directory to ensure the rename is durable.
+        if let Some(parent) = path.parent() {
+            let dir = fs::File::open(parent)?;
+            dir.sync_all()?;
+        }
+
+        Ok(())
```
numinnex: I think we should handle those errors inline here, or atleast convert them to error types that would represent different stages of snapshot persistance, as depending on the stage, we can be left in a broken state that is still recoverable. For example if we'd to fail in the `rename` stage, we can safely retry that operation, but if we'd to fail in the line 70-71, we have to start from scratch, as the `write_all` could partially write the buffer and then fail, same for `sync_all`. 

I also think we should move to `O_SYNC` and `O_DSYNC` for the metadata, but that can wait, as I am working now on adding support to `O_DIRECT` for `partitions` and this requires quite a few changes to the `Buffer` that we use, as `Bytes`/`BytesMut` does not allow to allocate aligned buffers.

core/metadata/src/impls/metadata.rs
```diff
@@ -52,6 +53,47 @@ impl IggySnapshot {
...
+    }
+
+    /// Load a snapshot from disk.
+    ///
+    /// # Errors
+    /// Returns `SnapshotError` if the file cannot be read or deserialization fails.
+    pub fn load(path: &Path) -> Result<Self, SnapshotError> {
+        let data = std::fs::read(path)?;
+
+        // TODO: when checksum is added we need to check
```
numinnex, line 110: create a linear issue for that.

core/metadata/src/impls/metadata.rs
```diff
@@ -99,6 +140,37 @@ pub struct IggyMetadata<C, J, S, M> {
     pub snapshot: Option<S>,
     /// State machine - lives on all shards
     pub mux_stm: M,
+    /// Root data directory, used by checkpoint to persist snapshots.
```
numinnex: Maybe it's good idea to store some sort of `snapshot_coordinator` there struct, that would hide those details away ?

core/metadata/src/impls/metadata.rs
```diff
@@ -325,3 +432,35 @@ where
...
+    /// Create a snapshot from the current state machine and persist it to disk.
+    ///
+    /// After the snapshot is durably persisted, advances the journal's
+    /// snapshot watermark so that entries at or below `last_op` may be
+    /// evicted from the ring buffer on future appends.
+    ///
+    /// # Errors
+    /// Returns `SnapshotError` if snapshotting, persistence, or compaction fails.
+    #[allow(clippy::future_not_send)]
+    pub async fn checkpoint(&self, data_dir: &Path, last_op: u64) -> Result<(), SnapshotError>
```
numinnex: If we'd go with `Comment on line R143`, then this would be part of that coordinator I've mentioned on `R143`

numinnex (changes requested):

core/journal/src/lib.rs
krishvishal: I agree that `drain` API is much cleaner. One thing to consider is, `drain` would read and deserialize all removed entries to return them to the caller, but the main consumer today (checkpoint) doesn't need the returned entries, it just wants them removed from  WAL. What do we do considering wasted the deserialization cost?

krishvishal: Also the ops in the journal aren't necessarily in a contiguous range. There can be gaps from pipelined prepares arriving out of order, slots overwritten when a new op lands on the same index. So `drain(begin..end)` is semantically not suitable here.

core/metadata/src/impls/metadata.rs
krishvishal: Done.

core/journal/src/lib.rs
krishvishal: I think `write_at/read_at` APIs make more sense here. Because for `InMemory` storage reads/writes are just index ops.

core/metadata/src/impls/metadata.rs

core/journal/src/lib.rs
krishvishal: Done.

core/journal/src/file_storage.rs
krishvishal: Done.

core/journal/src/file_storage.rs
krishvishal: Done.

core/journal/src/lib.rs
numinnex: The drain api would take `op` as the input e.g you would provide op_range rather than index_range, so the wrapping of the journal doesn't matter. 

The cost of deserialization isn't that big, if not existent in-fact, since we store `Message<PrepareHeader>` as entries in our `IggyMetadata` journal, so it's just reading opaque bytes from disk and constructing the `Message<PrepareHeader>` from them (only cost there is the validation).

core/journal/src/lib.rs
numinnex: Yeah I agree with the write_at/read_at distinction, I will create a linear issue out of that.

core/metadata/src/impls/metadata.rs

core/metadata/src/impls/metadata.rs

core/journal/src/lib.rs
krishvishal: Done.

core/journal/src/lib.rs
```diff
@@ -30,19 +34,45 @@ where
...
+    ///
+    /// Implementations that persist to disk should rewrite the underlying
+    /// storage to reclaim space. The default is a no-op for journals that
+    /// do not persist to disk.
+    ///
+    /// # Errors
+    /// Returns an I/O error if the drain fails.
+    fn drain(
+        &self,
+        _start_op: u64,
```
numinnex: Use the `Range` primitive

krishvishal: Done.

PR #2917, opened 2026-03-11
feat(server): add threads count and disk space to sysinfo stats

### LINKED ISSUE #2732: Add threads and free space to sysinfo print
Extend the periodic sysinfo log output to show thread count and disk space metrics.

- Add thread count by calling the sysinfo process API and storing it in the Stats struct
- Add iggy usage - calculate it using only iggy APIs, dont use sysinfo here
- Add free/total disk space for the data directory using the sysinfo Disks API, it should be fast and non-invasive, it cannot impact latency
- Update binary protocol serialization to handle the new Stats fields
- Include the new metrics in the sysinfo printer log output with readable formatting

Done when the sysinfo logs show thread count and disk space (both free and total) alongside the existing CPU/memory metrics.
Integration test should properly check size before/after creating streams, topics, messages, etc.

This can be get very complicated very fast, so don't hesitate to ask on discord.

seokjin0414:
## Summary

Closes #2732.

- Add `threads_count`, `free_disk_space`, `total_disk_space` fields to Stats struct
- Collect thread count via sysinfo `Process::tasks()` API (Linux only, 0 on macOS)
- Collect disk space via `fs2` crate (`statvfs` syscall) with warning log on failure
- Add `IggyUsage` (messages_size_bytes), `Disk` (free/total), `Threads` (conditional) to sysinfo log output
- Update binary protocol encoding/decoding with backward-compatible field appending
- Add CLI Table/List output rows for new fields
- Add integration tests: field presence in all 4 output formats + JSON value verification with message size check

## Design

- **Thread count**: `process.tasks().map(|t| t.len()).unwrap_or(0)` — uses existing `refresh_processes()` call, no additional sysinfo refresh needed. Conditional log output (skipped when 0, matching OpenFDs pattern)
- **Disk space**: `fs2::available_space()` / `fs2::total_space()` — single `statvfs` syscall per metric (~10μs), already used in `logger.rs`. Logs warning on failure instead of silently defaulting to 0
- **IggyUsage**: existing `messages_size_bytes` field (iggy metadata only, no sysinfo) — added to sysinfo log output
- **Binary protocol**: new fields appended after `cache_metrics`, decoded with `current_position + N <= payload.len()` guards for backward compatibility

### DIFF
--- a/core/server/src/shard/system/stats.rs
+++ b/core/server/src/shard/system/stats.rs
@@ -120,12 +120,24 @@ impl IggyShard {
                 }
             }
 
-            stats.free_disk_space = fs2::available_space(&self.config.system.path)
-                .unwrap_or(0)
-                .into();
-            stats.total_disk_space = fs2::total_space(&self.config.system.path)
-                .unwrap_or(0)
-                .into();
+            match fs2::available_space(&self.config.system.path) {
+                Ok(space) => stats.free_disk_space = space.into(),
+                Err(err) => {
+                    tracing::warn!(
+                        "Failed to get available disk space for '{}': {err}",
+                        self.config.system.path
+                    );
+                }
+            }
+            match fs2::total_space(&self.config.system.path) {
+                Ok(space) => stats.total_disk_space = space.into(),
+                Err(err) => {
+                    tracing::warn!(
+                        "Failed to get total disk space for '{}': {err}",
+                        self.config.system.path
+                    );
+                }
+            }
 
             Ok(stats)
         })


core/server/src/shard/system/stats.rs
```diff
@@ -118,6 +120,20 @@ impl IggyShard {
...
+            let disks = sysinfo::Disks::new_with_refreshed_list();
+            let mut best_mount_len = 0usize;
+            for disk in disks.list() {
+                let mount = disk.mount_point();
+                if data_path.starts_with(mount) && mount.as_os_str().len() > best_mount_len {
+                    best_mount_len = mount.as_os_str().len();
+                    stats.free_disk_space = disk.available_space().into();
+                    stats.total_disk_space = disk.total_space().into();
+                }
+            }
```
hubcio: how much time does this take when there are 10k++ files in local_data directory of iggy? this is blocking code, we dont want to introduce latency spikes on shard 0

seokjin0414: Replaced sysinfo::Disks with fs2::available_space() / fs2::total_space() in a5f7f70b.
fs2 uses a single statvfs syscall on the given path — it does not scan directory contents, so file count (10k++) has zero impact.
Measured overhead is ~10μs regardless of directory size.
fs2 is already a dependency of the server crate (used in logger.rs for the same purpose).

core/server/src/shard/system/stats.rs
```diff
@@ -118,6 +120,13 @@ impl IggyShard {
                 }
             }
 
+            stats.free_disk_space = fs2::available_space(&self.config.system.path)
+                .unwrap_or(0)
+                .into();
+            stats.total_disk_space = fs2::total_space(&self.config.system.path)
+                .unwrap_or(0)
```
hubcio: please add warning print when it fails

core/server/src/shard/system/stats.rs
hubcio: alright, thats perfectly fine.

core/server/src/shard/system/stats.rs
seokjin0414: Thanks for the review~
Fixed both — added tracing::warn! on fs2 failure in stats.rs (28d58403) and updated PR description to reflect fs2 instead of Disks

PR #2920, opened 2026-03-12
ci: unify run-*-examples-from-readme.sh into single entrypoint with TLS support

### LINKED ISSUE #2907: Unify run-*-examples-from-readme.sh scripts into a single entrypoint with TLS support
### Problem

We have 7 separate scripts for testing SDK examples:

- `scripts/run-node-examples-from-readme.sh`
- `scripts/run-rust-examples-from-readme.sh`
- `scripts/run-go-examples-from-readme.sh`
- `scripts/run-python-examples-from-readme.sh`
- `scripts/run-java-examples-from-readme.sh`
- `scripts/run-csharp-examples-from-readme.sh`
- `examples/node/test-examples.sh`

Each script duplicates ~60-70% of its code (server binary check, startup, wait-for-ready loop, README parsing, teardown). `scripts/utils.sh` already has process management helpers but none of the example scripts use it. We also added TLS support across multiple SDKs but have no TLS example test runs.

These should be unified into a single script that sources shared utilities from `scripts/utils.sh`, starts 2 servers (one plain, one TLS-enabled), and runs all examples from each language's `README.md` as the single source of truth. CI should be updated accordingly and old per-language scripts removed.

### Languages & README patterns

| Language | README(s) | Grep pattern |
|----------|-----------|--------------|
| Rust | `README.md`, `examples/rust/README.md` | `cargo r --bin iggy --`, `cargo run --example` |
| Node | `README.md`, `examples/node/README.md` | `npm run`, `tsx` |
| Go | `examples/go/README.md` | `go run` |
| Python | `examples/python/README.md` | `uv run` |
| Java | `examples/java/README.md` | `./gradlew` |
| C# | `README.md`, `examples/csharp/README.md` | `cargo r --bin iggy --`, `dotnet run --project` |

### TLS-capable SDKs

Rust, Go, C#, Java, Node (Python has no TLS support yet).

atharvalade:

Closes #2907

## Rationale

Seven per-language scripts duplicated 60 to 70 percent of their logic (server startup, README parsing, teardown) and TLS example tests were missing for several SDKs. A single entrypoint that uses shared utilities and supports TLS addresses both issues.

## What changed?

The six `run-*-examples-from-readme.sh` scripts and `examples/node/test-examples.sh` were replaced by `scripts/run-examples-from-readme.sh`, which sources shared helpers from `scripts/utils.sh`. Server lifecycle, README parsing, and example execution are centralized. The script accepts a `--language` flag (rust, go, node, python, java, csharp, or all) and runs both plain and TLS examples when available. CI calls the unified script for each language. The server is always built from source via `cargo build`, not from DockerHub.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

- Opus 4.6
- Exploration and Review
- Minimal
- Yes

### DIFF
--- a/scripts/run-examples-from-readme.sh
+++ b/scripts/run-examples-from-readme.sh
@@ -229,8 +229,8 @@ run_go_examples() {
         "examples/go" \
         "README.md" \
         "^go run" \
-        "" \
-        "" \
+        "--tls" \
+        "^go run.*--tls" \
         0 \
         ""
 }
--- a/examples/go/README.md
+++ b/examples/go/README.md
@@ -34,6 +34,15 @@ go run ./getting-started/producer/main.go
 go run ./getting-started/consumer/main.go
 ```
 
+## TLS Examples
+
+To test with a TLS-enabled server, start the server with TLS configured (see main README), then run:
+
+```bash
+go run ./getting-started/producer/main.go --tcp-server-address localhost:8090 --tls --tls-ca-file ../../core/certs/iggy_ca_cert.pem
+go run ./getting-started/consumer/main.go --tcp-server-address localhost:8090 --tls --tls-ca-file ../../core/certs/iggy_ca_cert.pem
+```
+
 ## Example Structure
 
 All examples can be executed directly from the repository. Follow these steps:


scripts/run-examples-from-readme.sh
```diff
@@ -0,0 +1,545 @@
...
+# Unified script to run SDK examples from README.md files.
+# Usage: ./scripts/run-examples-from-readme.sh [OPTIONS]
+#
+#   --language LANG   Language to test: rust|go|node|python|java|csharp (default: all)
+#   --target TARGET   Cargo target architecture for the server binary
+#   --skip-tls        Skip TLS example tests
+#   --goos GOOS       Go OS target
+#   --goarch GOARCH   Go architecture target
+#   --csharpos OS     C# --os flag
+#   --csharparch ARCH C# --arch flag
```
hubcio: why do we need these? i would assume that it should use host OS/arch

scripts/run-python-examples-from-readme.sh
hubcio, line 1: you silently removed support for python TLS testing, because old examples had them hardcoded, seelines 182,183. IMHO you should add these to python README.md

scripts/run-examples-from-readme.sh
```diff
@@ -0,0 +1,545 @@
...
+# ---------------------------------------------------------------------------
+
+# shellcheck disable=SC2329
+run_rust_examples() {
+    resolve_server_binary "${TARGET}"
+    resolve_cli_binary "${TARGET}"
+
+    # Transform: inject --target flag when cross-compiling
+    if [ -n "${TARGET}" ]; then
+        TRANSFORM_COMMAND() {
```
hubcio, line 180: run_rust_examples defines TRANSFORM_COMMAND as a global function and cleans it up at line 297. But if the function returns early at line 243 (return "${EXAMPLES_EXIT_CODE}"), the cleanup at 297 is never reached. In --language all mode, this leaks the Rust cross-compile transform into the next language's execution.

use trap or move unset into run_one after each language call

scripts/run-examples-from-readme.sh
```diff
@@ -0,0 +1,545 @@
...
+    start_plain_server
+    wait_for_server_ready "Rust"
+
+    # CLI commands from root README (abbreviated `cargo r`, not `cargo run`)
+    run_readme_commands "README.md" '^\`cargo r --bin iggy -- '
+    if [ "${EXAMPLES_EXIT_CODE}" -ne 0 ]; then
+        cd "${ROOT_WORKDIR}"
+        stop_server
+        report_result "${EXAMPLES_EXIT_CODE}"
+        return "${EXAMPLES_EXIT_CODE}"
```
hubcio: duplicated logic with run_csharp_examples (lines 414-502).  Could be addressed by making run_language_examples accept an optional "pre-flight" callback for CLI commands.

scripts/utils.sh
```diff
@@ -149,3 +149,196 @@ function on_exit_bench() {
...
+        "$@"
+    fi
+}
+
+# Run commands extracted from a README file.
+# Usage: run_readme_commands readme_file grep_pattern [cmd_timeout]
+# Reads matching lines, strips backticks/comments, executes each.
+# Calls TRANSFORM_COMMAND function on each command if defined.
+# Returns: sets global EXAMPLES_EXIT_CODE.
+function run_readme_commands() {
```
hubcio, line 306: Lines 119-157 reimplement the loop from run_readme_commands (lines 293-344) because of the grep_exclude parameter. The run_readme_commands function doesn't support exclusion patterns. adding an optional exclude parameter would eliminate the duplication

scripts/run-examples-from-readme.sh
```diff
@@ -0,0 +1,545 @@
...
+    local cmd_timeout="$7"
+    local server_extra_args="$8"
+
+    echo ""
+    echo "============================================================"
+    echo "  Running ${lang} examples"
+    echo "============================================================"
+    echo ""
+
+    EXAMPLES_EXIT_CODE=0
```
hubcio, line 93: this is a bit messy: multiple functions initialize, read, and write this global: run_language_examples (line 103), run_rust_examples (line 232), run_csharp_examples (line 435), run_python_examples (line 359), and run_readme_commands (line 331). In --language all mode, this shared state persists across run_one calls. run_one does set +e and captures the return code, but if a language function forgets to reinitialize EXAMPLES_EXIT_CODE, stale failure state could carry over (or mask a failure).

scripts/utils.sh
```diff
@@ -149,3 +149,196 @@ function on_exit_bench() {
...
+}
+
+# Gracefully stop the server using the PID file.
+function stop_server() {
+    if [ -e "${EXAMPLES_PID_FILE}" ]; then
+        kill -TERM "$(cat "${EXAMPLES_PID_FILE}")" 2>/dev/null || true
+        sleep 2
+        rm -f "${EXAMPLES_PID_FILE}"
+    fi
+}
```
hubcio, line 274: i'm not a fan of sleeping. please add actual verification whether givern PID has actually exited. (this is preexisting issue). send sigterm, then wait for 2s via timeout wait/waitpid, then sigkill and fail test. does that make sense?

scripts/run-examples-from-readme.sh
atharvalade: you're right, removed them

scripts/run-python-examples-from-readme.sh
atharvalade: oh good catch, added

scripts/run-examples-from-readme.sh
atharvalade: Moved `TRANSFORM_COMMAND` cleanup into `run_one` so it always runs after each language call, regardless of early returns

scripts/run-examples-from-readme.sh
atharvalade: added a pre-flight callback parameter to `run_language_examples`. Both Rust and C# now use it for CLI commands instead of duplicating the loop

scripts/utils.sh
atharvalade: Added optional `grep_exclude` parameter to `run_readme_commands` and removed the duplicated inline loop

scripts/run-examples-from-readme.sh
atharvalade: `run_one` now resets `EXAMPLES_EXIT_CODE` to 0 before each language call

scripts/utils.sh
atharvalade: yeah, that makes total sense. I replaced the sleep with `wait_for_process_pid` -- sends SIGTERM, waits up to 5s for actual exit, then falls back to SIGKILL...

hubcio:
@atharvalade for now ignore flaky mongo connector test, i'll take care about it later today. lets merge this PR after #2898 - can you make sure that go TLS is also tested?

PR #2921, opened 2026-03-12
feat(server): aligned buffer memory pool

tungtose:
pooled_buffer

Prepare the memory pool and buffer infrastructure for O_DIRECT I/O. Direct I/O requires buffers to be aligned to the underlying block size (commonly 4096 bytes). This allows the kernel to bypass the page cache, reducing double buffering and giving more predictable I/O latency.

Known Trade-offs:
+ Minimum allocation size is now 4096 bytes, meaning small utility buffers (e.g. put_u32_le, put_u64_le) now consume more memory than before
+ make_mutable in the HTTP path now copies buffers due to alignment incompatibility with Bytes


core/common/src/alloc/memory_pool.rs
```diff
@@ -16,26 +16,25 @@
...
-use bytes::BytesMut;
+use aligned_vec::{AVec, ConstAlign};
 use crossbeam::queue::ArrayQueue;
 use human_repr::HumanCount;
 use once_cell::sync::OnceCell;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use tracing::{info, trace, warn};
 
+pub const ALIGNMENT: usize = 4096;
```
hubcio, line 27: please make it configurable via compilation feature. it can be either 512 or 4096. make 512 a default feature - is that reasonable? or, is there any reason why is 4096 hardcoded?

tungtose: This is the issue I am referencing: https://github.com/DataDog/glommio/issues/7 and this [answer](https://unix.stackexchange.com/questions/761398/are-there-any-benefits-in-setting-a-hdds-logical-sector-size-to-4kn#:~:text=I%20still%20don't%20have,4096%20bytes%20won't%20hurt.&text=It%20can%20then%20be%20changed,with%20the%20manufacturer's%20proprietary%20tools.)

and for my naive thinking is that, 4096 is aligned to both 512 and 4096 but 512 is only aligned to 512. and most modern devices use 4096

tungtose:
> > make_mutable in the HTTP path now copies buffers due to alignment incompatibility with Bytes
> 
> can you elaborate?

Correct me if I'm wrong, to make the make_mutable function in the HTTP path zero copy, we need to change `IggyMessagesBatch` to store `PooledBuffer` instead of `Bytes`. However, `IggyMessagesBatch` is a core type with dependencies throughout the codebase, so this could be refactored in a separate PR if needed

Other than that, make_mutable is not in the hot path. If user is using the HTTP API for high-throughput message sending, they already have bigger problems than this one copy

core/common/src/alloc/memory_pool.rs
tungtose: @hubcio Can I set this option to 512 or 4096, with 4096 as the default?

core/partitions/src/lib.rs
```diff
@@ -54,8 +54,8 @@ pub(crate) fn decode_send_messages_batch(body: Bytes) -> Option<IggyMessagesBatc
     }
 
     let indexes_bytes = body.split_to(indexes_len);
-    let indexes = IggyIndexesMut::from_bytes(PooledBuffer::from(indexes_bytes), 0);
-    let messages = PooledBuffer::from(body);
+    let indexes = IggyIndexesMut::from_bytes(PooledBuffer::from(indexes_bytes.as_ref()), 0);
```
numinnex: Using it like that will miss-align both the messages/indexes buffers, I think the `split_at` should enforce the original alignment, but I am not sure how to achieve that

tungtose:
> @tungtose did you check the performance? is there any difference?

@hubcio here is the bench result running `run-benches.sh`:
### on main:
2026-03-18T13:40:40.123993Z  INFO bench_report::prints: Producers Results: Total throughput: 3854.07 MB/s, 3854076 messages/s, average throughput per Producer: 481.76 MB/s, p50 latency: 1.30 ms, p90 latency: 2.75 ms, p95 latency: 7.14 ms, p99 latency: 18.61 ms, p999 latency: 44.92 ms, p9999 latency: 91.50 ms, average latency: 2.12 ms, median latency: 1.30 ms, min: 0.51 ms, max: 209.82 ms, std dev: 2.34 ms, total time: 2.50 s

Running iggy-bench pinned-consumer tcp...
Poll results:
2026-03-18T13:40:44.577717Z  INFO bench_report::prints: Consumers Results: Total throughput: 2748.06 MB/s, 2748056 messages/s, average throughput per Consumer: 343.51 MB/s, p50 latency: 2.73 ms, p90 latency: 4.15 ms, p95 latency: 4.62 ms, p99 latency: 5.60 ms, p999 latency: 6.58 ms, p9999 latency: 7.35 ms, average latency: 2.84 ms, median latency: 2.73 ms, min: 1.22 ms, max: 8.01 ms, std dev: 0.44 ms, total time: 3.17 s

### on this branch:
Send results:
2026-03-18T11:49:46.897632Z  INFO bench_report::prints: Producers Results: Total throughput: 4115.45 MB/s, 4115452 messages/s, average throughput per Producer: 514.43 MB/s, p50 latency: 1.18 ms, p90 latency: 2.44 ms, p95 latency: 3.30 ms, p99 latency: 18.39 ms, p999 latency: 61.13 ms, p9999 latency: 115.81 ms, average latency: 1.94 ms, median latency: 1.18 ms, min: 0.38 ms, max: 213.28 ms, std dev: 2.38 ms, total time: 2.22 s

Running iggy-bench pinned-consumer tcp...
Poll results:
2026-03-18T11:49:51.299828Z  INFO bench_report::prints: Consumers Results: Total throughput: 2920.55 MB/s, 2920547 messages/s, average throughput per Consumer: 365.07 MB/s, p50 latency: 2.54 ms, p90 latency: 4.05 ms, p95 latency: 4.44 ms, p99 latency: 5.72 ms, p999 latency: 9.07 ms, p9999 latency: 11.36 ms, average latency: 2.68 ms, median latency: 2.54 ms, min: 0.73 ms, max: 19.75 ms, std dev: 0.66 ms, total time: 3.10 s

core/partitions/src/lib.rs
tungtose: @numinnex  I have updated the code, please help me to check

hubcio:
could you please run benchmarks with rate limit, 4 producers / consumers, total rate limit equal to 500MB? this way we'll see the p50. run it for like 30s or so

core/common/src/alloc/memory_pool.rs
numinnex: I think we should stick to 4096, for the reasons that you've mentioned.

hubcio:
so the results for producers:

| Metric | Master | PR | Delta |
|--------|--------|----|-------|
| p50 | 0.91 ms | 1.14 ms | +25% |
| p90 | 1.53 ms | 1.99 ms | +30% |
| p99 | 2.89 ms | 2.97 ms | +3% |

consumers:

| Metric | Master | PR | Delta |
|--------|--------|----|-------|
| p50 | 1.31 ms | 1.79 ms | +37% |
| p90 | 152.23 ms | 17.87 ms | -88% |
| p99 | 340.48 ms | 147.42 ms | -57% |
| max | 526.35 ms | 286.22 ms | -46% |
| std dev | 64.63 ms | 22.96 ms | -64% |

p50 regressed ~25-37%. any clue why? but tail latencies (p90+) improved 57-88%, especially on consumer side.

PR #2923, opened 2026-03-12
fix(sdk): allow hostnames in server address configuration

### LINKED ISSUE #2882: Rust SDK: Support DNS addresses in client
### Bug description

Not sure if it's a bug or a request for enhacement.

## Problem
During connection build via the client, if an ip addresses is passed (e.g. `127.0.0.1`), then the client tries to connect. When passing a dns entry (e.g. `host.docker.internal`), it raises `Invalid IP address: host.docker.internal:8090`

## Desired Behaviour
When connecting, even passing a `dns` entry should result in a successful connection (assuming that the entry is correct and matches to an active ip address)

## Notes
Considering [https://stackoverflow.com/questions/28255861/convert-string-to-socketaddr](https://stackoverflow.com/questions/28255861/convert-string-to-socketaddr) it seems that there are two ways of establishing a connection: one via ip and one via `dns` entry. Not sure if this functionality has not been supported due to the added complexity, but in some cases it could be quite handy

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

felixfaisal:

Closes #2882

## Rationale

Server address validation previously required a literal socket address, preventing the use of DNS hostnames. This change allows hostnames (e.g., `localhost:8080`) in addition to IP addresses.

## What changed?

Server address validation previously relied on `SocketAddr` parsing, which only accepts literal IP addresses and rejected DNS hostnames (e.g., `localhost:8080`). This prevented using hostnames when configuring client connections.

Validation now uses `ToSocketAddrs`, which performs address resolution and returns the list of socket addresses for a given host. `TcpStream::connect(...)` already performs this resolution internally and attempts to connect to each resolved address.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? ChatGPT free version
2. Scope of usage? Write documentations like comments 
- I built the code using `cargo build` 
- I ran the unit tests under `TCPConfigBuilder` and added an additional unit test, also modifying another
- I then ran the server and ran the examples 
4. Can you explain every line of the code if asked? Yup


core/common/src/utils/net.rs
```diff
@@ -0,0 +1,155 @@
...
+    // Reject bare IPv6 addresses
+    if host.contains(':') {
+        return Err(IggyError::InvalidIpAddress(
+            host.to_string(),
+            port_str.to_string(),
+        ));
+    }
+
+    port_str
+        .parse::<u16>()
```
hubcio: port 0 passes validation here since `0u16` parses fine, but it's not a valid server address port. worth rejecting explicitly:

```rust
let port = port_str.parse::<u16>().map_err(|_| ...)?;
if port == 0 {
    return Err(IggyError::InvalidIpAddress(host.to_string(), port_str.to_string()));
}
```

same applies to the ipv6 branch (line 48).

core/common/src/utils/net.rs
```diff
@@ -0,0 +1,155 @@
...
+
+        return Ok(());
+    }
+    // hostname:port or IPv4:port — rsplit_once to split at last colon
+    let (host, port_str) = addr.rsplit_once(':').ok_or(IggyError::InvalidIpAddress(
+        addr.to_string(),
+        "".to_string(),
+    ))?;
+
+    if host.is_empty() {
```
hubcio: currently only empty host and bare colons are rejected, but any other characters pass - e.g. `"!!!:8080"` is accepted. would be good to add basic hostname character validation per RFC 952/1123 (alphanumeric, hyphens, dots) to catch obviously invalid addresses at config time rather than deferring everything to DNS resolution at connect time.

core/common/src/utils/net.rs
felixfaisal: Updated to check for `0` , and added a unit test for it. ✅

core/common/src/utils/net.rs
felixfaisal: Created a separate function, to check if valid IPv4 address and if not check if valid hostname, added a couple of unit tests using claude. ✅  
I was thinking if i should use regex or not, and ultimately decided to do it this way.

core/common/src/types/configuration/tcp_config/tcp_client_config_builder.rs
```diff
@@ -105,11 +104,7 @@ impl TcpClientConfigBuilder {
     /// Builds the TCP client configuration.
     pub fn build(self) -> Result<TcpClientConfig, IggyError> {
         let addr = self.config.server_address.trim();
```
hubcio: `build()` validates the trimmed `addr` but returns `self.config` with the original untrimmed `server_address`. if someone passes `" localhost:8090 "`, validation passes but `TcpStream::connect` later gets the whitespace-padded string. same issue in `websocket_client_config_builder.rs`.

fix: `self.config.server_address = self.config.server_address.trim().to_string();` before validating.

core/common/src/utils/net.rs
```diff
@@ -0,0 +1,291 @@
...
+/// Accepted formats:
+/// - `hostname:port` (e.g. `iggy-server:8090`, `localhost:8090`)
+/// - `ipv4:port` (e.g. `127.0.0.1:8090`)
+/// - `[ipv6]:port` (e.g. `[::1]:8090`)
+///
+/// Rejected formats:
+/// - Bare IPv6 without brackets (e.g. `::1:8080`) — ambiguous due to colons
+/// - Missing port (e.g. `localhost`)
+/// - Invalid port (e.g. `localhost:abc`, `localhost:65536`)
+pub fn parse_server_address(addr: &str) -> Result<(), IggyError> {
```
hubcio: the function name says "parse" but it returns `Result<(), IggyError>` - it validates, it doesn't parse anything. `validate_server_address` would be more accurate for a public API, or return a structured type like `(host, port)` to make the parse useful.

core/common/src/utils/net.rs
```diff
@@ -0,0 +1,291 @@
...
+    if let Some(rest) = addr.strip_prefix('[') {
+        // Bracketed IPv6: "[::1]:port"
+        let close = rest.find(']').ok_or(IggyError::InvalidIpAddress(
+            addr.to_string(),
+            "".to_string(),
+        ))?;
+        let ipv6_str = &rest[0..close];
```
hubcio: `&rest[0..close]` - the `0..` is redundant, `&rest[..close]` is idiomatic.

core/sdk/src/prelude.rs
```diff
@@ -63,6 +63,7 @@ pub use iggy_common::{
     TcpClientReconnectionConfig, Topic, TopicDetails, TopicPermissions, TransportEndpoints,
     TransportProtocol, UserId, UserStatus, Validatable, WebSocketClientConfig,
     WebSocketClientConfigBuilder, WebSocketClientReconnectionConfig, defaults, locking,
+    parse_server_address,
```
hubcio: `parse_server_address` is a low-level validation utility. the builders already call it internally, so SDK consumers rarely need it directly. consider keeping it importable from `iggy_common` without putting it in the prelude - prelude is for iggy users.

core/common/src/utils/mod.rs
```diff
@@ -23,6 +23,7 @@ pub(crate) mod crypto;
 pub(crate) mod duration;
 pub(crate) mod expiry;
 pub(crate) mod hash;
+pub mod net;
```
hubcio: `pub mod net` but the sibling modules are `pub(crate)`. since only `parse_server_address` is re-exported in `lib.rs`, this could be `pub(crate) mod net` to match the pattern.

core/common/src/types/configuration/tcp_config/tcp_client_config_builder.rs
felixfaisal: Updated ✅

core/common/src/utils/net.rs
felixfaisal: Updated to `validate_server_address` ✅

core/common/src/utils/net.rs
felixfaisal: Updated ✅

core/sdk/src/prelude.rs
felixfaisal: Updated ✅

core/common/src/utils/mod.rs
felixfaisal: Updated ✅

core/common/src/utils/net.rs
```diff
@@ -0,0 +1,291 @@
...
+            && label.len() <= 63
+            && label
+                .chars()
+                .next()
+                .is_some_and(|c| c.is_ascii_alphanumeric())
+            && label
+                .chars()
+                .last()
+                .is_some_and(|c| c.is_ascii_alphanumeric())
+            && label.chars().all(|c| c.is_ascii_alphanumeric() || c == '-')
```
hubcio: underscores in hostnames are rejected here (only allows alphanumeric and `-`). while RFC 1123 technically forbids underscores, they are widespread in practice - Docker Compose v1 projects, custom container names, SRV records like `_svc._tcp.example.com`. since the pr's primary goal is enabling hostname-based addresses for container environments, consider allowing underscores (`|| c == '_'`). at minimum, document the restriction.

core/common/src/utils/net.rs
```diff
@@ -0,0 +1,291 @@
...
+/// Rejected formats:
+/// - Bare IPv6 without brackets (e.g. `::1:8080`) — ambiguous due to colons
+/// - Missing port (e.g. `localhost`)
+/// - Invalid port (e.g. `localhost:abc`, `localhost:65536`)
+pub fn validate_server_address(addr: &str) -> Result<(), IggyError> {
+    if let Some(rest) = addr.strip_prefix('[') {
+        // Bracketed IPv6: "[::1]:port"
+        let close = rest.find(']').ok_or(IggyError::InvalidIpAddress(
+            addr.to_string(),
+            "".to_string(),
```
hubcio: when port is unknown (e.g., missing closing bracket), `""` as the port field produces error messages with a dangling colon like `Invalid IP address: [::1:`. consider passing the full original `addr` for context, or a placeholder like `<missing>`. also, `String::new()` is idiomatic over `"".to_string()` (zero-alloc). same issue at lines 47, 67.

core/common/src/utils/net.rs
felixfaisal: Updated ✅

core/common/src/utils/net.rs
felixfaisal: Updated ✅

mmodzelewski (approved):

spetz (approved):

PR #2925, opened 2026-03-12
feat(connectors): add generic HTTP sink connector

mlevkov:
## Summary

Adds a **generic HTTP sink connector** that delivers consumed Iggy stream messages to any HTTP endpoint — webhooks, Lambda functions, REST APIs, or SaaS integrations.

- **4 batch modes**: `individual` (one request per message), `nd_json` (newline-delimited), `json_array` (single array), `raw` (bytes)
- **Automatic retry via reqwest-middleware**: `RetryTransientMiddleware` with `ExponentialBackoff` and custom `HttpSinkRetryStrategy` that respects user-configured success codes
- **Flexible metadata**: optional Iggy envelope (stream/topic/offset/timestamp), checksum, origin timestamp, message headers
- **Strongly typed serialization**: `MetadataEnvelope`, `IggyMetadata`, `EncodedPayload`, `EncodedHeader` structs (no `json!` macros)
- **Simple hex message IDs**: 32-character lowercase hex string (no dashes)
- **Connection pooling**: `ClientWithMiddleware` wrapping reqwest with TCP keep-alive (30s), pool idle timeout (90s), configurable max connections
- **Health checks**: opt-in startup probe (HEAD/GET/OPTIONS) with graceful degradation
- **TLS**: optional `danger_accept_invalid_certs` for dev environments
- **Custom headers**: arbitrary HTTP headers for auth tokens, API keys, routing
- **strum Display**: `BatchMode` variants have human-readable Display for logs

### Files Changed

| File | Purpose |
|------|---------|
| `sinks/http_sink/src/lib.rs` | Core implementation — types, config, Sink trait, retry, batch modes, 57 unit tests |
| `sinks/http_sink/README.md` | Usage guide, config reference, runtime model, deployment patterns, message flow |
| `sinks/http_sink/Cargo.toml` | Crate manifest |
| `sinks/http_sink/config.toml` | Example connector configuration |
| `fixtures/http/container.rs` | Docker test container with WireMock |
| `fixtures/http/sink.rs` | Test fixtures (single-topic, multi-topic, batch mode variants) |
| `http/http_sink.rs` | 7 integration tests (delivery, metadata, batch modes, multi-topic) |

### Architecture

```
Iggy Stream → [Runtime polls topic] → consume(messages) → [batch mode] → HTTP endpoint
                                            │
                                    ┌───────┴────────┐
                                    │  individual    │ → 1 request per message
                                    │  nd_json       │ → all messages, newline-delimited
                                    │  json_array    │ → all messages, JSON array
                                    │  raw           │ → 1 request per message, raw bytes
                                    └────────────────┘
```

### Code Review History

| Round | Agents | Findings | Fixed |
|-------|--------|----------|-------|
| Round 1 | code-reviewer, silent-failure-hunter, comment-analyzer, code-simplifier | 12 | 12 |
| Round 2 | Same 4 agents (follow-up) | 7 | 7 |
| Round 3 | Same 4 agents (post-feature additions) | 17 | 15 (2 deferred) |
| Round 4 | Same 4 agents (double-review follow-up) | 6 | 6 |
| Round 5 | Maintainer review (hubcio) | 14 | 14 |
| Round 6 | 6 agents (code-reviewer ×2, silent-failure-hunter, code-architect, comment-analyzer, type-design-analyzer) | 8 | 8 |
| Round 7 | Same 6 agents (follow-up on remediation) | 0 | — |
| Round 8 | 5 agents (code-reviewer, silent-failure-hunter, comment-analyzer, code-architect, type-design-analyzer) — post spetz review refactoring | 6 | 6 |
| Round 9 | Same 5 agents (follow-up on Round 8 remediation) | 0 | — |
| **Total** | | **70** | **68** (2 deferred) |

Round 5 (maintainer review — hubcio) key changes:
- Consuming iterator — eliminate `.clone()` on payloads in all send methods
- DRY refactor — extracted `send_per_message()` shared by individual/raw modes
- Pre-built `HeaderMap` in `open()` — avoid per-request header parsing
- `HashSet` for `success_status_codes` — O(1) hot-path lookup
- Forward iggy message headers as `iggy_headers` in metadata envelope
- Success/transient status code overlap warning in `open()`
- Worst-case latency documentation on `consume()`

Round 8 (post spetz review — 12 items addressed in 10 commits + 1 remediation):
- `BatchMode::Ndjson` → `NdJson`, serializes as `"nd_json"`
- Field name constants (`FIELD_*`, `ENCODING_*`) replacing magic strings
- `strum::Display` on `BatchMode`, removed `mode_name`/`batch_mode` string params
- `content_type()` moved from `HttpSink` to `BatchMode` impl
- `json!` macros → strongly typed `MetadataEnvelope`/`IggyMetadata`/`EncodedPayload`/`EncodedHeader`
- Comment separators removed, test helpers reordered below tests
- `client` parameter threading removed — `fn client()` helper
- UUID v8 → 32-char hex (`uuid` crate removed)
- Hand-rolled retry → `reqwest-middleware` (`RetryTransientMiddleware` + `HttpSinkRetryStrategy`)
- `retry_backoff_multiplier`: `f64` → `u32` (matches `ExponentialBackoff::base()`)
- `retries_count` field removed (retries transparent inside middleware)

Round 8 remediation (2 CRITICAL, 4 HIGH):
- `build_envelope()` returns `Result` — no more `unwrap_or(Null)` silent data loss
- `retry_delay > max_retry_delay` auto-swaps to prevent `ExponentialBackoff` panic
- Retry-After warn uses actual status code (was hardcoded "429")
- Middleware errors logged with `{:#}` for full retry context
- `requests_sent` → `send_attempts` (middleware retries make counter ambiguous)
- Doc comment `default: 2.0` → `default: 2` for u32 backoff multiplier

### Deferred (tracked in issues)

- **L4**: Structured error type enum replacing string-based `Error::HttpRequestFailed` — [#2927](https://github.com/apache/iggy/issues/2927)
- **D1**: Expose internal metrics (errors_count) via runtime health API — [#2928](https://github.com/apache/iggy/issues/2928)

## Test Plan

- [x] 57 unit tests covering config parsing, validation, serialization, batch modes, header forwarding, edge cases
- [x] 7 integration tests with WireMock in Docker:
  - `individual_json_messages_delivered_as_separate_posts` — per-message delivery + envelope
  - `ndjson_messages_delivered_as_single_request` — NDJSON batch mode
  - `json_array_messages_delivered_as_single_request` — JSON array batch mode
  - `raw_binary_messages_delivered_without_envelope` — raw byte passthrough
  - `metadata_disabled_sends_bare_payload` — include_metadata=false
  - `individual_messages_have_sequential_offsets` — offset ordering integrity
  - `multi_topic_messages_delivered_with_correct_topic_metadata` — 2 topics, metadata accuracy
- [x] `cargo clippy -p iggy_connector_http_sink -- -D warnings` — 0 warnings
- [x] `cargo clippy -p integration -- -D warnings` — 0 warnings

## Related

- Discussion: https://github.com/apache/iggy/discussions/2919
- Issue (structured errors): https://github.com/apache/iggy/issues/2927
- Issue (runtime metrics): https://github.com/apache/iggy/issues/2928


core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+        let total = messages.len();
+        let mut delivered = 0u64;
+        let mut http_failures = 0u64;
+        let mut serialization_failures = 0u64;
+        let mut consecutive_failures = 0u32;
+        let mut last_error: Option<Error> = None;
+
+        for message in &messages {
+            let offset = message.offset;
+            let payload_json = match self.payload_to_json(message.payload.clone()) {
```
hubcio: also at lines 727, 799, 880.

`consume()` takes `messages: Vec<ConsumedMessage>` by value, but all send methods iterate via `&messages` and `.clone()` each payload. for `Payload::Raw(Vec<u8>)`, `FlatBuffer(Vec<u8>)`, and `Json(OwnedValue)` this is a full heap copy per message. fix is trivial -- change to `for message in messages` (consuming iterator) since `payload_to_json()` and `try_into_vec()` already take `Payload` by value.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+                }
+            }
+        }
+    }
+
+    /// Send messages in `individual` mode — one HTTP request per message.
+    /// Continues processing remaining messages if one fails (partial delivery).
+    /// Aborts remaining messages after `MAX_CONSECUTIVE_FAILURES` consecutive HTTP failures
+    /// to avoid hammering a dead endpoint.
+    async fn send_individual(
```
hubcio: also `send_raw` at lines 866-961.

both do the same thing: iterate messages, build body, check size, send with retry, track consecutive failures, circuit-break at `MAX_CONSECUTIVE_FAILURES`, report partial delivery. only the body-building step differs. extract a shared `send_per_message()` parameterized by a closure.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+                Err(e)
+            }
+            None => Ok(()),
+        }
+    }
+}
+
+/// Convert `simd_json::OwnedValue` to `serde_json::Value` via direct structural mapping.
+/// NaN/Infinity f64 values are mapped to `null` (same as Elasticsearch sink).
+fn owned_value_to_serde_json(value: &simd_json::OwnedValue) -> serde_json::Value {
```
hubcio: exact copy of the same function in the elasticsearch sink. should live in `iggy_connector_sdk` as a shared utility. bug fixes for future `simd_json::OwnedValue` variants would need to be applied to both independently otherwise.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+        HttpMethod::Put => client.put(url),
+        HttpMethod::Patch => client.patch(url),
+        HttpMethod::Delete => client.delete(url),
+    }
+}
+
+/// Format a u128 message ID as a UUID-style hex string (8-4-4-4-12).
+/// This is positional formatting only — no RFC 4122 version/variant bits are set.
+/// Downstream consumers should treat this as an opaque identifier, not a standards-compliant UUID.
+fn format_u128_as_uuid(id: u128) -> String {
```
hubcio: `format_u128_as_uuid()` produces strings that look exactly like uuids but have no version/variant bits. downstream consumers in go/java/python/c# will parse this with uuid libraries and get validation failures or make wrong assumptions. options: (a) use the `uuid` crate with version 8 custom, (b) emit plain hex without dashes, or (c) rename the field to something like `iggy_message_id_hex` so nobody mistakes it for a uuid.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+            })),
+            Payload::Proto(proto_str) => Ok(serde_json::json!({
+                "data": general_purpose::STANDARD.encode(proto_str.as_bytes()),
+                "iggy_payload_encoding": "base64"
+            })),
+        }
+    }
+
+    /// Build a message envelope with optional metadata wrapping.
+    fn build_envelope(
```
hubcio: iggy messages can carry user-defined `headers: Option<HashMap<HeaderKey, HeaderValue>>`. the http sink ignores them entirely -- not forwarded as http headers, not in the metadata envelope, not logged. the elasticsearch sink includes them. if a producer attaches routing/tracing headers, they just vanish. either forward them (e.g. `iggy_headers` in the envelope) or document as unsupported.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+impl Sink for HttpSink {
+    async fn open(&mut self) -> Result<(), Error> {
+        // Validate success_status_codes — empty would cause every response to be treated as failure
+        if self.success_status_codes.is_empty() {
+            return Err(Error::InitError(
+                "success_status_codes must not be empty — would cause retry storms against healthy endpoints".to_string(),
+            ));
+        }
+        for &code in &self.success_status_codes {
+            if !(200..=599).contains(&code) {
```
hubcio, line 1047: also interacts with `is_transient_status` at line 370.

configuring `success_status_codes = [200, 503]` means a 503 is treated as success (success check at line 463 runs before transient check at line 481), which quietly disables retry for that code. at minimum warn if success codes overlap with the transient set `[429, 500, 502, 503, 504]`.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+        );
+        Ok(())
+    }
+
+    /// Deliver messages to the configured HTTP endpoint.
+    ///
+    /// **Runtime note**: The connector runtime's `process_messages()` in `runtime/src/sink.rs` currently discards the `Result`
+    /// returned by `consume()`. All retry logic lives inside this method — returning `Err`
+    /// does not trigger a runtime-level retry. This is a known upstream issue.
+    async fn consume(
```
hubcio, line 1192: individual mode with `batch_length=50`, `max_retries=3`, `max_retry_delay=30s` could take up to `50 * 4 * 30s = 6000s` worst case. the `MAX_CONSECUTIVE_FAILURES` circuit breaker helps but a pathological fail-succeed-fail-succeed pattern bypasses it. at least document the worst-case latency.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+
+        builder
+            .build()
+            .map_err(|e| Error::InitError(format!("Failed to build HTTP client: {}", e)))
+    }
+
+    /// Apply the configured HTTP method to a `reqwest::Client` for the target URL,
+    /// including custom headers (excluding Content-Type, which is set per-request by batch mode).
+    fn request_builder(&self, client: &reqwest::Client) -> reqwest::RequestBuilder {
+        let mut builder = build_request(self.method, client, &self.url);
```
hubcio: called inside `send_with_retry`'s loop (line 434). for individual mode with 50 messages and 3 retries, that's 200 header rebuilds per poll cycle. build a `HeaderMap` once in `open()` and reuse it.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+            self.requests_sent.fetch_add(1, Ordering::Relaxed);
+
+            match client.execute(request).await {
+                Ok(response) => {
+                    let status = response.status();
+
+                    // Check for Retry-After before consuming the response
+                    let retry_after = self.parse_retry_after(&response);
+
+                    if self.success_status_codes.contains(&status.as_u16()) {
```
hubcio: linear scan on the hottest path. default list is 4 elements so it's not a disaster, but a `HashSet` or sorted vec with binary search would be better.

core/integration/tests/connectors/http/http_sink.rs
```diff
@@ -0,0 +1,1119 @@
...
+use iggy_binary_protocol::MessageClient;
+use iggy_common::Identifier;
+use integration::harness::seeds;
+use integration::iggy_harness;
+
+// ============================================================================
+// Test 1: Individual Batch Mode
+// ============================================================================
+
+/// Test 1: Individual JSON Messages Delivered as Separate HTTP POSTs
```
hubcio: per-test doc comments are 30-50 lines each. module-level doc is 158 lines. trim per-test docs to 3-5 lines; the module-level comment already covers architecture well enough.

core/connectors/sinks/http_sink/config.toml
```diff
@@ -0,0 +1,90 @@
...
+# Connection pool — max idle connections per host (default: 10).
+max_connections = 10
+
+# Verbose request/response logging (default: false).
+verbose_logging = false
+
+# Custom HTTP headers. Replace placeholder values with real credentials.
+# Do not commit actual secrets — use environment variable overrides for production.
+[plugin_config.headers]
+Authorization = "Bearer my-secret-token"
```
hubcio: `Authorization = "Bearer my-secret-token"` may trigger secret-scanning tools. use `"Bearer <YOUR_TOKEN_HERE>"` instead.

core/integration/tests/connectors/fixtures/http/container.rs
```diff
@@ -0,0 +1,229 @@
...
+            response
+                .json()
+                .await
+                .map_err(|e| TestBinaryError::InvalidState {
+                    message: format!("Failed to parse WireMock admin response: {e}"),
+                })?;
+
+        let requests = body["requests"]
+            .as_array()
+            .unwrap_or(&vec![])
```
hubcio: use a static empty slice or refactor to avoid the allocation. it's test code so it doesn't really matter, but it's sloppy.

core/integration/tests/connectors/http/http_sink.rs
```diff
@@ -0,0 +1,1119 @@
...
+// ============================================================================
+
+/// Second topic name for the multi-topic test. Defined locally to avoid
+/// polluting the shared harness seeds with HTTP-sink-specific constants.
+///
+/// **Design Decision**: The shared `seeds.rs` module provides generic seed functions
+/// (`connector_stream`, `mcp_standard`) used by all connector types. Adding HTTP-sink-specific
+/// constants there would create coupling. Instead, this test creates the second topic inline
+/// after the seed runs, keeping the harness generic. See code review finding H1.
+const TEST_TOPIC_2: &str = seeds::names::TOPIC_2;
```
hubcio: just use `seeds::names::TOPIC_2` directly at the usage sites.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2099 @@
...
+        HttpMethod::Patch => client.patch(url),
+        HttpMethod::Delete => client.delete(url),
+    }
+}
+
+/// Format a u128 message ID as a UUID-style hex string (8-4-4-4-12).
+/// This is positional formatting only — no RFC 4122 version/variant bits are set.
+/// Downstream consumers should treat this as an opaque identifier, not a standards-compliant UUID.
+fn format_u128_as_uuid(id: u128) -> String {
+    let hex = format!("{:032x}", id);
```
hubcio: two `format!` calls. a stack-allocated approach via `write!` to a `[u8; 36]` would be zero-alloc.

hubcio (changes requested):
overall good direction, please fix the CI (markdown lint) and rest of the comments. looks pretty solid to me.

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: I tried to maintain isolation away from changes to anything other than connectors. Am I correct to assume that you would like to see this in sdk?

mlevkov: @hubcio would delighted to hear your thoughts.

hubcio: yes, please move owned_value_to_serde_json into iggy_connector_sdk - it's already duplicated between the elasticsearch and HTTP sinks, and any future fix would need to be applied twice.

connectors already depend on the SDK for trait definitions, so exposing shared utilities there adds no new coupling.

mlevkov: ok, will do

hubcio (changes requested):
looking very good, I found only one nit: you moved `owned_value_to_serde_json` to connectors SDK but you didn't use it in `elasticsearch_sink`. there is duplicate code. please use it :)

after you fix that we can merge. also, please fix CI - there were some changes in the `binary_protocol` crate, looks like tests fail after rebase.

hubcio (dismissed):

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+/// Payload formatting mode for HTTP requests.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum BatchMode {
+    /// One HTTP request per message (default). Note: with batch_length=50, this produces 50
+    /// sequential HTTP round trips per poll cycle. Use ndjson or json_array for higher throughput.
+    #[default]
+    Individual,
+    /// All messages in one request, newline-delimited JSON.
+    Ndjson,
```
spetz: Should it be `NdJson` (`nd_json`) instead?

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+            .tcp_keepalive(Duration::from_secs(DEFAULT_TCP_KEEPALIVE_SECS))
+            .danger_accept_invalid_certs(self.tls_danger_accept_invalid_certs);
+
+        builder
+            .build()
+            .map_err(|e| Error::InitError(format!("Failed to build HTTP client: {}", e)))
+    }
+
+    /// Determine the Content-Type header based on batch mode.
+    fn content_type(&self) -> &'static str {
```
spetz: Maybe the `content_type` could be a function exposed directly in `BatchMode`?

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+        match payload {
+            Payload::Json(value) => {
+                // Direct structural conversion (not serialization roundtrip).
+                // Follows the Elasticsearch sink pattern. NaN/Infinity f64 → null.
+                Ok(owned_value_to_serde_json(&value))
+            }
+            Payload::Text(text) => Ok(serde_json::Value::String(text)),
+            Payload::Raw(bytes) | Payload::FlatBuffer(bytes) => Ok(serde_json::json!({
+                "data": general_purpose::STANDARD.encode(&bytes),
+                "iggy_payload_encoding": "base64"
```
spetz: I think that both fields `data` and `iggy_payload_encoding` should be added as e.g. `pub const` so it's easy to use them in multiple places (and test)

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+        topic_metadata: &TopicMetadata,
+        messages_metadata: &MessagesMetadata,
+        payload_json: serde_json::Value,
+    ) -> serde_json::Value {
+        if !self.include_metadata {
+            return payload_json;
+        }
+
+        let mut metadata = serde_json::json!({
+            "iggy_id": format_u128_as_uuid(message.id),
```
spetz: Let's avoid using `serde_json::json!` whenever possible, and simply pass the strongly typed struct instead.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+            "iggy_offset": message.offset,
+            "iggy_timestamp": message.timestamp,
+            "iggy_stream": topic_metadata.stream,
+            "iggy_topic": topic_metadata.topic,
+            "iggy_partition_id": messages_metadata.partition_id,
+        });
+
+        if self.include_checksum {
+            metadata["iggy_checksum"] = serde_json::json!(message.checksum);
```
spetz: Same regarding `iggy_checksum`, `iggy_origin_timestamp` and any other magic strings.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+                    } else {
+                        serde_json::Value::String(v.to_string_value())
+                    };
+                    (k.to_string_value(), value)
+                })
+                .collect();
+            metadata["iggy_headers"] = serde_json::Value::Object(headers_map);
+        }
+
+        serde_json::json!({
```
spetz: Also here - IMHO better to use the strongly typed struct and then just serialize as json as a result.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+            .as_secs();
+        self.last_success_timestamp.store(now, Ordering::Relaxed);
+    }
+
+    /// Send an HTTP request with retry logic. Returns Ok on success, Err after exhausting retries.
+    ///
+    /// Takes `Bytes` instead of `Vec<u8>` so retries clone via reference-count increment (O(1))
+    /// rather than copying the entire payload on each attempt.
+    async fn send_with_retry(
+        &self,
```
spetz: Why not to use `reqwest-middleware` which is already part of the workspace, connectors runtime etc.?

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+    /// Aborts after `MAX_CONSECUTIVE_FAILURES` consecutive HTTP failures.
+    ///
+    /// `build_body` takes ownership of each `ConsumedMessage` — callers must extract
+    /// all needed fields (payload, metadata) within the closure.
+    async fn send_per_message<F>(
+        &self,
+        client: &reqwest::Client,
+        messages: Vec<ConsumedMessage>,
+        content_type: &str,
+        mode_name: &str,
```
spetz: Let's use the basic enum with e.g. `strum` macros on top to display the mode properly.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+        self.last_success_timestamp.store(now, Ordering::Relaxed);
+    }
+
+    /// Send an HTTP request with retry logic. Returns Ok on success, Err after exhausting retries.
+    ///
+    /// Takes `Bytes` instead of `Vec<u8>` so retries clone via reference-count increment (O(1))
+    /// rather than copying the entire payload on each attempt.
+    async fn send_with_retry(
+        &self,
+        client: &reqwest::Client,
```
spetz: Actually, wouldn't that be simpler if `client` was just a field used directly, instead of being passed every time?

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+}
+
+/// Format a u128 message ID as an RFC 4122 v8 (custom) UUID.
+///
+/// Uses `Uuid::new_v8()` which sets version=8 and variant=RFC4122 bits,
+/// producing UUIDs that downstream libraries accept as valid.
+/// Note: `new_v8()` overwrites 6 bits (version nibble + variant bits), so the
+/// UUID is not round-trippable to the original u128 value.
+fn format_u128_as_uuid(id: u128) -> String {
+    uuid::Uuid::new_v8(id.to_be_bytes()).to_string()
```
spetz: Actually, do we need `v8`, isn't `v4` sufficient enough? Maybe let's return the `simple` variant (without the dashes) or let the user configure it?

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use iggy_connector_sdk::Schema;
+
+    // ── Test helpers ──────────────────────────────────────────────────
```
spetz: Let's remove comment sections like this and the other ones below.
Also, code should be read like a book when possible, meaning that any helper functions etc. could be placed below the ones actually invoking them.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2121 @@
...
+    fn given_message_without_headers_should_not_include_iggy_headers() {
+        let sink = given_sink_with_defaults();
+        let message = given_json_message(1, 0);
+        let topic_meta = given_topic_metadata();
+        let msg_meta = given_messages_metadata();
+        let payload_json = sink.payload_to_json(message.payload.clone()).unwrap();
+
+        let envelope = sink.build_envelope(&message, &topic_meta, &msg_meta, payload_json);
+        assert!(
+            envelope["metadata"].get("iggy_headers").is_none(),
```
spetz: For example, it's one of the places where having the const for particular field name, or even better, the strongly typed struct, would make such a test less error-prone :).

core/connectors/sinks/http_sink/config.toml
```diff
@@ -0,0 +1,90 @@
...
+
+# Maximum HTTP body size in bytes (default: 10MB). Set to 0 to disable.
+max_payload_size_bytes = 10485760
+
+# Payload formatting mode (default: individual).
+# - "individual": one HTTP request per message
+# - "ndjson": newline-delimited JSON, all messages in one request
+# - "json_array": JSON array of messages in one request
+# - "raw": raw bytes, individual requests only
+batch_mode = "ndjson"
```
spetz, line 52: If the enum would be changed to `NdJson` please remember to update the configs.

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Great question. Here's the trade-off analysis:

### UUID format options

| Option | What happens | Pros | Cons |
|--------|-------------|------|------|
| **v4 (random)** | Fresh random UUID per message | Valid RFC 4122, universally accepted | Loses all correlation to Iggy message ID — can't trace back to source message |
| **v8 (current)** | Encodes u128 message ID with version/variant bits | Valid UUID, downstream libraries accept it, mostly preserves original ID | Overwrites 6 bits — not round-trippable, theoretical 1-in-64 collision window |
| **Simple hex** | `01234567-89ab-cdef-0123-456789abcdef` (no version bits) | Lossless — exact u128 preserved, fully round-trippable | Not a valid UUID — downstream `parse_uuid()` may reject it |
| **Configurable** | User picks format in config | Maximum flexibility | Adds config surface, complexity, testing burden for a display format |

### Recommendation

Keep **simple hex** (lossless, original approach) — it preserves the exact Iggy message ID for debugging/correlation, which is more valuable for an Iggy connector than strict UUID compliance. Downstream consumers already treat `iggy_id` as an opaque identifier, not as a UUID they need to validate.

If strict UUID compliance is needed, v8 is the right choice — but it sacrifices 6 bits of the original ID.

Happy to go with whatever you prefer here.

spetz: The server's code uses v4 as the default, and since HTTP passes ID as a string, I think that the simple from (without dashes) would be sufficient.

mlevkov: Alternatively, we could offer both in the envelope — a valid UUID for downstream compatibility and the raw value for lossless correlation:

```json
{
  "iggy_id": "01234567-89ab-8def-8123-456789abcdef",
  "iggy_id_raw": 5940264471900273711
}
```

This way downstream UUID libraries work with `iggy_id`, and anyone needing to trace back to the exact Iggy message uses `iggy_id_raw`. The raw field would be opt-in (similar to `include_checksum` / `include_origin_timestamp`).

mlevkov: @spetz Not to overcomplicate this, but I forgot to include an alternative consideration for completeness.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2069 @@
...
+        let headers = self.request_headers.as_ref().ok_or_else(|| {
+            Error::InitError("HTTP headers not initialized — was open() called?".to_string())
+        })?;
+
+        if self.verbose {
+            debug!(
+                "HTTP sink ID: {} — sending {:?} {} ({} bytes)",
+                self.id,
+                self.method,
+                self.url,
```
hubcio: also at lines 502, 504, 535, 537, 1027, 1034, 1093, 1116.
i'm wondering.. will there be a leak of credentials when we print whole url? can someone call have url `https://user:password@domain....`?

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2069 @@
...
+    Put,
+    Patch,
+    Delete,
+}
+
+/// Payload formatting mode for HTTP requests.
+#[derive(
+    Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, strum_macros::Display,
+)]
+#[serde(rename_all = "snake_case")]
```
hubcio, line 79: `#[serde(rename_all = "snake_case")]` forces `NdJson` to serialize as `"nd_json"`, but the industry standard term is `"ndjson"` (one word, per ndjson.org). the `Content-Type` header at line 102 correctly uses `application/x-ndjson`, which highlights the inconsistency.

since this connector is unreleased, now is the cheapest time to fix this - a `#[serde(rename = "ndjson")]` on the `NdJson` variant, or `#[serde(alias = "ndjson")]` to accept both. post-release this becomes a breaking config migration.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2069 @@
...
+
+        if lines.is_empty() {
+            return Err(Error::Serialization(
+                "All messages in NDJSON batch failed serialization".to_string(),
+            ));
+        }
+
+        let count = lines.len() as u64;
+
+        let mut body_str = lines.join("\n");
```
hubcio: `send_ndjson()` builds a `Vec<String>` (one `String` per message via `serde_json::to_string`), then `lines.join("\n")` allocates a new `String` and copies all the serialized bytes a second time. for a batch of 50 messages at ~1KB each, that's ~50KB of unnecessary copying.

a single pre-sized `Vec<u8>` with `serde_json::to_writer(&mut body, &envelope)` followed by `body.push(b'\n')` avoids both the intermediate `Vec<String>` and the join copy. the per-message error handling / `skipped` counter still works - just track the buffer position before writing and truncate on error.

also, instead of callign `Vec::new()` use `Vec::with_capacity()` to avoid realloc (malloc + memcpy underneath, or memmove between pages)

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2069 @@
...
+            "send_batch_body called with count=0 — callers must guard against empty batches"
+        );
+        if let Err(e) = self
+            .send_with_retry(body, self.batch_mode.content_type())
+            .await
+        {
+            // send_with_retry already added 1 to errors_count for the HTTP failure.
+            // Add the remaining messages that were serialized but not delivered.
+            if count > 1 {
+                self.errors_count.fetch_add(count - 1, Ordering::Relaxed);
```
hubcio: this adds `count - 1` to `errors_count`, relying on the implicit invariant that `send_with_retry()` always adds exactly 1 on failure (at line 499 or 534). the comment at line 685 documents this, which is good, but the coupling is fragile - if `send_with_retry` is ever modified to count differently, accounting silently breaks. a `debug_assert` on the delta or a single accounting site would make this more robust.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2069 @@
...
+            },
+            iggy_headers: headers_map,
+        };
+
+        let envelope = MetadataEnvelope {
+            metadata,
+            payload: payload_json,
+        };
+
+        serde_json::to_value(envelope)
```
hubcio, line 482: `build_envelope` serializes `MetadataEnvelope` into a `serde_json::Value` here, then callers re-serialize that `Value` to bytes (`to_vec` at line 666, `to_string` at line 750). that's two serialization passes per message.

this exists because `build_envelope` serves three modes - individual needs bytes, json_array needs `Value` for the array, ndjson needs `String`. splitting per-mode would fix it but adds duplication. worth keeping in mind if profiling ever shows serialization as a bottleneck.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -0,0 +1,2069 @@
...
+/// Metadata envelope wrapping a payload with Iggy message metadata.
+#[derive(Debug, Serialize)]
+struct MetadataEnvelope {
+    metadata: IggyMetadata,
+    payload: serde_json::Value,
+}
+
+/// Iggy message metadata fields.
+#[derive(Debug, Serialize)]
+struct IggyMetadata {
```
hubcio, line 125: the HTTP sink uses `iggy_offset`, `iggy_partition_id`, etc. the elasticsearch sink uses `_iggy_offset`, `_iggy_partition` (leading underscore, different field name for partition). both conventions are defensible - underscore prefix is an ES convention, unprefixed is more standard for JSON APIs - but the divergence should be a deliberate decision rather than accidental.

i know that this not good principles to touch many components in single PR... but my eyes hurt. maybe worth a follow-up to unify the convention?

hubcio (changes requested):
only comment 79 (nd_json vs ndjson) in `lib.rs` are critical, rest can be fixed later as a follow up. (or now, if you feel like it)

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Fixed. Added `sanitize_url_for_log()` that strips userinfo (`user:password@`) before logging. All log/error messages now use `self.log_url` (sanitized); only actual HTTP requests use `self.url` (raw). Falls back to regex-style stripping when `Url::parse` fails (prevents credential leak even for malformed URLs). Good catch on the `user:password@` vector.

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Fixed. Added `#[serde(rename = "ndjson")]` on the `NdJson` variant to override the `snake_case` default. All config files, README, and tests use `"ndjson"`. Added a comment explaining the rename precedence for future maintainers. The Rust variant stays `NdJson` (correct Rust casing), serializes as `"ndjson"` (industry standard per ndjson.org).

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Fixed. Replaced `Vec<String>` + `join("\n")` with direct `serde_json::to_writer(&mut body, &envelope)` into a single pre-sized `Vec<u8>`. Each message is serialized directly into the output buffer followed by `body.push(b'\n')`. On error, `body.truncate(pos)` rolls back the partial write. Eliminates both the intermediate `String` per message and the `join` copy. `Vec::with_capacity(messages.len() * 256)` avoids reallocation for typical payloads.

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Fixed. Changed `count - 1` to `count.saturating_sub(1)` and added an INVARIANT comment documenting the dependency on `send_with_retry` incrementing `errors_count` exactly once before returning `Err` (with line references to both error paths). The `debug_assert!(count > 0)` at the top of the function catches the zero case in dev builds.

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Addressed. Initially added `build_envelope_vec` and `build_envelope_to_writer` to eliminate the `Value` intermediate, but 6-agent review flagged the triple maintenance surface as a higher risk than the ~microsecond serialization overhead. Collapsed back to single `build_envelope` entry point — callers choose their serialization format (`to_vec` for individual, `to_writer` for ndjson, `Value` directly for json_array). `build_iggy_metadata` remains as the shared helper. The performance difference is negligible against HTTP round-trip cost.

core/connectors/sinks/http_sink/src/lib.rs
mlevkov: Added a doc comment on `IggyMetadata` documenting the intentional divergence: HTTP sink uses `iggy_*` (standard JSON convention), ES sink uses `_iggy_*` (ES internal field convention). Both are defensible for their respective ecosystems. Agree a follow-up to unify would be valuable but it touches another component — happy to do it as a separate PR if the team has a preferred convention.

spetz (approved):

numinnex (approved):

hubcio (approved):

PR #2931, opened 2026-03-14
refactor(repo): wrap credentials in SecretString to prevent leaks

### LINKED ISSUE #2728: Refactor password handling using `secrecy` crate
Use the `secrecy` crate to wrap passwords and sensitive strings throughout the codebase. This prevents accidental logging, ensures zeroization on drop, and makes security boundaries explicit in the type system.

**What to focus on:**
- Wrap password fields with `Secret<String>` in commands, CLI args, and SDK methods
- Keep binary protocol serialization unchanged (Secret is transparent during serialization)
- Only expose secrets via `.expose_secret()` when hashing or serializing
- Consider PAT tokens and connection strings too (currently we manually replace sensitive content with `*****`)

Done when passwords are never accidentally logged, tests pass, and all plaintext password fields are wrapped.

grainier:

Closes #2728

## Rationale

Passwords, tokens, and connection strings were stored as plain `String`, risking exposure via `Debug`, `Display`, logs, and error messages.

## What changed?

Sensitive credential fields (passwords, PATs, connection URIs, API keys) were plain `String` values that could leak through derived `Debug`, log interpolation, or error formatting.

All such fields now use `secrecy::SecretString`, which redacts in `Debug` and requires explicit `expose_secret()` to access the inner value. Custom `Debug` impls were added for types that hold sensitive `String` fields not converted to `SecretString` (e.g., `RawPersonalAccessToken`, `TokenInfo`). A `serde_secret` helper module was added for fields that must be serialized over the wire.

## Local Execution

- Passed except for https://github.com/apache/iggy/discussions/2930. It was failing on fresh master as well. 
- Pre-commit hooks ran.

All quality checks mentioned in CONTRIBUTING.md pass locally:
- `cargo fmt --all`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo build`
- `cargo test`
- `cargo machete`
- `cargo sort --workspace`
- `typos`

## AI Usage

 1. Claude Code (CLI)
 2. Used to identify/verify all locations where `String` passwords/tokens needed conversion to `SecretString`, for documentation (i.e. core/common/src/utils/serde_secret.rs doc string), and commit review.
 3. Verified by running all checks mentioned in CONTRIBUTING.md.
 4. Yes.

### DIFF
--- a/core/journal/src/lib.rs
+++ b/core/journal/src/lib.rs
@@ -39,7 +39,10 @@ pub trait Storage {
     type Buffer;
 
     fn write(&self, buf: Self::Buffer) -> impl Future<Output = usize>;
-    fn read(&self, offset: usize, buffer: Self::Buffer) -> impl Future<Output = Self::Buffer>;
+    // TODO: Get rid of the `len` usize, we need to do changes in `Simulator` in order to support that.
+    // Maybe we should go back to passing in the `Buffer` again, but I am not sure how to handle it in the `Partitions Journal`, since we use in-memory impl
+    // which extracts the buffer out of the `Vec<Message>` and we don't need to allocate a new buffer.
+    fn read(&self, offset: usize, len: usize) -> impl Future<Output = Self::Buffer>;
 }
 
 pub trait JournalHandle {
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -15,12 +15,14 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::journal::{Noop, PartitionJournal};
+use crate::journal::{PartitionJournal, PartitionJournalMemStorage};
 use crate::log::SegmentedLog;
-use crate::{AppendResult, Partition};
+use crate::{AppendResult, Partition, decode_send_messages_batch};
 use iggy_common::{
     ConsumerGroupOffsets, ConsumerOffsets, IggyByteSize, IggyError, IggyMessagesBatchMut,
     IggyTimestamp, PartitionStats,
+    header::{Operation, PrepareHeader},
+    message::Message,
 };
 use journal::Journal as _;
 use std::sync::Arc;
@@ -30,7 +32,7 @@ use tokio::sync::Mutex as TokioMutex;
 // This struct aliases in terms of the code contained the `LocalPartition from `core/server/src/streaming/partitions/local_partition.rs`.
 #[derive(Debug)]
 pub struct IggyPartition {
-    pub log: SegmentedLog<PartitionJournal, Noop>,
+    pub log: SegmentedLog<PartitionJournal<PartitionJournalMemStorage>, PartitionJournalMemStorage>,
     /// Committed offset — advanced only after quorum ack.
     pub offset: Arc<AtomicU64>,
     /// Dirty offset — advanced on every prepare (before commit).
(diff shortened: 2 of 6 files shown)


core/connectors/runtime/src/api/config.rs
hubcio: any reason why this is still plain `String` while `RuntimeContext.api_key` is `SecretString`?

core/common/src/types/user/user_identity_info.rs
```diff
@@ -35,10 +36,19 @@ pub struct IdentityInfo {
 /// It consists of the following fields:
 /// - `token`: the value of token.
 /// - `expiry`: the expiry of token.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Serialize, Deserialize)]
 pub struct TokenInfo {
     /// The value of token.
     pub token: String,
```
hubcio: critical issue: you overidden debug, but users of this token can still leak this via `Display`, `Serialize` or cloning into log contexts

same issue in `RawPersonalAccessToken.token`

core/server/src/state/system.rs
hubcio, line 82: preexisting issue: if anyone logs UserState with {:?}, the bcrypt/argon2 hash leaks. The Display impl is safe (doesn't include it), but Debug is not.

core/common/src/utils/serde_secret.rs
```diff
@@ -0,0 +1,53 @@
...
+
+pub fn serialize_optional_secret<S: serde::Serializer>(
+    secret: &Option<SecretString>,
+    serializer: S,
+) -> Result<S::Ok, S::Error> {
+    match secret {
+        Some(s) => serializer.serialize_some(s.expose_secret()),
+        None => serializer.serialize_none(),
+    }
+}
```
hubcio, line 53: please test this

core/common/src/types/configuration/auth_config/connection_string.rs
```diff
@@ -115,7 +116,7 @@ impl<T: ConnectionStringOptions + Default> ConnectionString<T> {
             server_address: server_address.to_owned(),
             auto_login: AutoLogin::Enabled(Credentials::UsernamePassword(
                 username.to_owned(),
-                password.to_owned(),
+                SecretString::from(password.to_owned()),
```
hubcio: cant you just use SecretString::from(password)? `SecretString` implements `From<&str>`

core/connectors/runtime/src/api/config.rs
grainier: I wasn't sure the `config_env` derive macro would support `SecretString`. Turns out by adding `#[config_env(secret, leaf)] #[serde(serialize_with = "serialize_secret")]` we can make it SecretString. Fixed.

core/common/src/types/user/user_identity_info.rs
grainier: Originally kept as String (and overridden debug) to avoid breaking the public API. Converted both to SecretString with `serialize_with`.  Fixed in cb7616f

core/common/src/utils/serde_secret.rs
grainier: Added 4 unit tests (roundtrip, Some, None, value preservation) in cb7616f

core/common/src/types/configuration/auth_config/connection_string.rs
grainier: Good catch, my bad. Fixed in cb7616f

core/server/src/state/system.rs
grainier: For now, added manual Debug redaction for `password_hash` and `token_hash` in cb7616f. Went with manual Debug instead of `SecretString` because the hash flows through UserState -> UserMeta -> User across 3 crates. So, converting one without the others felt inconsistent. Should I convert the full chain instead?

hubcio: i think manual debug redaction is OK here - just note it in the code, as TODO comment.

grainier: Added a TODO comment in 79fb55d

numinnex (approved):

PR #2933, opened 2026-03-14
feat(connectors): implement InfluxDB Sink and Source

### LINKED ISSUE #2700: InfluxDB Connector (Sink, Source) for Apache Iggy
Build a native InfluxDB connector for Apache Iggy to enable seamless streaming of time-series data between Iggy and InfluxDB. This connector should allow users to **sink** data from Iggy into InfluxDB, with a design that can later support **source** ingestion from InfluxDB into Iggy.  

This will unlock strong use cases around metrics, observability, IoT, financial ticks, and real-time analytics, and make Iggy easier to adopt in time-series-heavy architectures.

ryerraguntla:
Adds a native InfluxDB connector for Apache Iggy resolving #2700.

### New crates:
- `iggy_connector_influxdb_sink` — consumes Iggy messages and writes them to InfluxDB 2.x via line protocol (`/api/v2/write`)
- `iggy_connector_influxdb_source` — polls InfluxDB Flux query results and produces messages to Iggy topics

### Features:
- Batch writes / cursor-based polling
- Exponential backoff + jitter on retries
- Circuit breaker after consecutive failures
- HTTP 429 Retry-After header support
- Configurable precision (`ns`, `us`, `ms`, `s`)
- Full integration tests using Testcontainers (influxdb:2.7-alpine)

Closes #2700

## What changed?

Implemented the following features based on a good connector framework
**InfluxDB Sink Connector**
Writes Iggy stream messages to InfluxDB v2 using the line protocol write API.
Data writing — Messages are serialised as line protocol with correct escaping for measurements, tag values, and string fields. Payloads can be written as JSON (validated and escaped), UTF-8 text, or Base64-encoded raw bytes. Timestamps are converted from Iggy's microsecond epoch to the configured InfluxDB precision (ns/us/ms/s, default µs). An offset-based nanosecond blend prevents silent deduplication when multiple messages in the same batch share the same microsecond timestamp. If a producer sets timestamp=0, the connector falls back to SystemTime::now() to avoid Year-1970 points.
Resilience — Writes are retried up to max_retries (default 3) on HTTP 429 and 5xx responses using exponential backoff with ±20% jitter. The Retry-After header is honoured in both integer-seconds and RFC 7231 HTTP-date formats before falling back to own backoff. A circuit breaker opens after circuit_breaker_threshold (default 5) consecutive batch failures and holds for circuit_breaker_cool_down (default 30s) before a half-open probe. Batch errors are captured and propagated to the runtime after processing all remaining sub-batches, preventing silent data loss.
Startup — open() retries the InfluxDB /health endpoint up to max_open_retries (default 10) with capped exponential backoff, so the connector recovers from transient InfluxDB restarts without manual intervention.
Metadata — Stream name, topic, and partition can be written as InfluxDB tags; message checksum and origin timestamp as fields. Each is individually togglable. The measurement name is configurable (default iggy_messages). Verbose logging mode promotes per-batch diagnostics from debug! to info! without code changes.

**InfluxDB Source Connector**
Polls InfluxDB v2 via the Flux query API and publishes results to Iggy topics as structured JSON messages.
Incremental polling — Tracks max(_time) from each Flux response as a cursor, advanced by 1 ns before storing to prevent boundary-point re-delivery. The cursor is templated into any $cursor and $limit placeholder in the user-supplied Flux query, supporting arbitrary range, filter, pivot, and aggregation chains. The cursor and row count are persisted via ConnectorState so polling resumes exactly where it left off after a restart.
CSV parsing — Handles InfluxDB annotated CSV output: skips #group, #datatype, and #default annotation rows, detects header rows by field names, and correctly handles blank-line separators between multi-table results.
Message format — Each row is published as a structured JSON envelope {measurement, field, timestamp, value, row} by default. When payload_column is set, the raw field value is extracted and emitted in the configured format (JSON, Text, or Raw/Base64).
Resilience — Same exponential backoff, jitter, Retry-After parsing, and circuit breaker as the sink. open() retries the /health endpoint before declaring ready. close() drops the reqwest client to release all connection pool resources.
Configuration — cursor_field (default _time) and initial_offset allow operators to replay from any point in the bucket. poll_interval, batch_size, and timeout are all tunable. Verbose logging toggles per-poll diagnostics between debug! and info!. 
## Local Execution

- Passed 
- Pre-commit hooks ran  

## AI Usage
1. Which tools?   GitHub Copilot, Claude 
2. Scope of usage?  autocomplete, generated functions 
3. How did you verify the generated code works correctly? - Yes Verified by testing


hubcio (changes requested):
please make sure that all checks in CI pass. after that, we can review.

core/connectors/sdk/src/lib.rs
```diff
@@ -341,7 +341,11 @@ pub enum Error {
     #[error("Invalid config")]
     InvalidConfig,
     #[error("Invalid record")]
```
hubcio: `InvalidConfigValue(String)` has `#[error("Invalid record")]` - the inner string is silently discarded in `Display`. should be `#[error("Invalid config value: {0}")]`. every config validation error from both influxdb connectors (`build_write_url`, `build_health_url`, `build_query_url`) will display "Invalid record" and lose all diagnostic context.

core/connectors/sdk/src/lib.rs
```diff
@@ -341,7 +341,11 @@ pub enum Error {
...
+    InvalidConfigValue(String),
+    #[error("Invalid record")]
     InvalidRecord,
+    #[error("Invalid record value : {0}")]
```
hubcio: stray space before the colon: `"Invalid record value : {0}"` - should be `"Invalid record value: {0}"`.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+        if include_metadata && self.config.include_partition_tag.unwrap_or(true) {
+            tags.push(format!("partition={}", messages_metadata.partition_id));
+        }
+
+        let mut fields = vec![
+            format!(
+                "message_id=\"{}\"",
+                escape_field_string(&message.id.to_string())
+            ),
+            format!("offset={}i", message.offset as i64),
```
hubcio: `message.offset as i64` wraps values >= 2^63 to negative, silently corrupting the stored field. same issue at line 463 (`partition_id as i64`), line 468 (`checksum as i64`), and line 473 (`origin_timestamp as i64`). `checksum` is a hash that routinely exceeds `i64::MAX`. use the unsigned integer suffix `u` in line protocol (e.g. `iggy_checksum={}u`) or encode as string fields.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+        return base;
+    }
+    let delta = rand::rng().random_range(0..=jitter_range * 2);
+    Duration::from_millis(millis.saturating_sub(jitter_range) + delta)
+}
+
+// True exponential backoff: base * 2^attempt, capped at max_delay
+fn exponential_backoff(base: Duration, attempt: u32, max_delay: Duration) -> Duration {
+    let factor = 2u64.saturating_pow(attempt);
+    let raw = Duration::from_millis(base.as_millis().saturating_mul(factor as u128) as u64);
```
hubcio: `base.as_millis()` returns `u128`. after `saturating_mul(factor as u128)`, the result is cast `as u64` which silently truncates (keeps low 64 bits). if the saturated result exceeds `u64::MAX`, the cast can produce a near-zero value, defeating the backoff. the `.min(max_delay)` only caps from above, not from below. fix: `.min(max_delay.as_millis()) as u64` or `u64::try_from(...).unwrap_or(u64::MAX)`.

also at `influxdb_source/src/lib.rs:261`.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+    async fn current_cursor(&self) -> String {
+        let state = self.state.lock().await;
+        state
+            .last_timestamp
+            .clone()
+            .or_else(|| self.config.initial_offset.clone())
+            .unwrap_or_else(|| DEFAULT_CURSOR.to_string())
+    }
+
+    fn query_with_params(&self, cursor: &str) -> String {
```
hubcio: flux query injection via string interpolation. `cursor` comes from InfluxDB `_time` column (line 678) and is substituted via `.replace("$cursor", cursor)`. a crafted or corrupted `_time` value containing flux syntax (e.g. `") |> drop() //`) would be injected into the query. InfluxDB `/api/v2/query` supports parameterized queries via `"params"` in the JSON body - use those instead. at minimum, validate that `cursor` matches RFC3339 before substitution.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+        };
+
+        // message.timestamp is microseconds since Unix epoch.
+        // If it is 0 (unset by the producer), fall back to now() so points are
+        // not stored at Unix epoch (year 1970), which falls outside every
+        // range(start: -1h) query window.
+        // We also blend the message offset as sub-microsecond nanoseconds so
+        // that multiple messages in the same batch get distinct timestamps and
+        // are not deduplicated by InfluxDB (same measurement+tags+time = 1 row).
+        let base_micros = if message.timestamp == 0 {
```
hubcio, line 473: timestamp uniqueness hack has multiple issues:
1. adds `offset % 1000` microseconds to the timestamp, breaking temporal accuracy
2. offsets differing by exactly 1000 collide (e.g. offset 0 and 1000 both get +0)
3. comment says "shifts by at most 999 ns" but code operates on microseconds, so it's 999us (999,000ns)
4. for `precision = "s"` or `"ms"`, the sub-millisecond adjustment is divided away, defeating the purpose

the idiomatic influxdb approach is to add an `offset` tag to distinguish points with the same timestamp.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+                                "Payload column '{payload_column}' is not valid JSON: {e}"
+                            ))
+                        })?;
+                    serde_json::to_vec(&value).map_err(|e| {
+                        Error::Serialization(format!("JSON serialization failed: {e}"))
+                    })
+                }
+                PayloadFormat::Text => Ok(raw_value.into_bytes()),
+                PayloadFormat::Raw => general_purpose::STANDARD
+                    .decode(raw_value.as_bytes())
```
hubcio, line 545: `.or_else(|_| Ok(raw_value.into_bytes()))` always succeeds, making the subsequent `.map_err(|e: base64::DecodeError| ...)` unreachable dead code. corrupt base64 is silently treated as raw UTF-8 bytes with no diagnostic. either remove the `.map_err()` and document the silent fallback, or remove the `.or_else()` to let decode errors propagate.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+// How many consecutive batch failures open the circuit breaker
+const DEFAULT_CIRCUIT_BREAKER_THRESHOLD: u32 = 5;
+// How long the circuit stays open before allowing a probe attempt
+const DEFAULT_CIRCUIT_COOL_DOWN: &str = "30s";
+
+// ---------------------------------------------------------------------------
+// Simple consecutive-failure circuit breaker
+// ---------------------------------------------------------------------------
+#[derive(Debug)]
+struct CircuitBreaker {
```
hubcio: `CircuitBreaker` + `parse_duration`, `jitter`, `exponential_backoff`, `is_transient_status`, `parse_retry_after` are copy-pasted verbatim between sink and source (~120 lines). extract to the SDK.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+    }
+
+    /// Call when a batch write succeeds — resets failure count and closes circuit.
+    fn record_success(&self) {
+        self.consecutive_failures.store(0, Ordering::SeqCst);
+    }
+
+    /// Call when a batch write fails after all retries — may open the circuit.
+    async fn record_failure(&self) {
+        let failures = self.consecutive_failures.fetch_add(1, Ordering::SeqCst) + 1;
```
hubcio: `consecutive_failures` is updated atomically via `fetch_add` outside the `open_until` mutex, but `is_open()` (line 95) resets it to 0 under the mutex. two concurrent calls can race: `is_open()` resets to 0, then a concurrent `record_failure()` does `fetch_add` on the old (non-zero) value it already loaded. the types are `Send + Sync` and wrapped in `Arc`, so concurrent use is API-legal. either protect both under the same mutex, or document the single-consumer assumption.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+
+// Apply ±20% random jitter to a duration to spread retry storms
+fn jitter(base: Duration) -> Duration {
+    let millis = base.as_millis() as u64;
+    let jitter_range = millis / 5; // 20% of base
+    if jitter_range == 0 {
+        return base;
+    }
+    let delta = rand::rng().random_range(0..=jitter_range * 2);
+    Duration::from_millis(millis.saturating_sub(jitter_range) + delta)
```
hubcio: `millis.saturating_sub(jitter_range) + delta` uses `+`, not `saturating_add`. if the truncated `as u64` on line 194 produces a large value, the addition can overflow - panic in debug, wrap in release. use `saturating_add`.

also at `influxdb_source/src/lib.rs:255`.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+            DEFAULT_OPEN_RETRY_MAX_DELAY,
+        );
+
+        let mut attempts = 0u32;
+        loop {
+            let response_result = client
+                .post(url.clone())
+                .header("Authorization", format!("Token {token}"))
+                .header("Content-Type", "text/plain; charset=utf-8")
+                .body(body.clone())
```
hubcio: `body.clone()` on every retry iteration. the entire line-protocol body (potentially MBs for large batches) is deep-copied because `reqwest::RequestBuilder::body()` consumes the value. wrap in `Bytes::from(body)` before the loop - `reqwest` accepts `Bytes` zero-copy (refcounted).

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+            fields.push(format!("iggy_checksum={}", message.checksum as i64));
+        }
+        if include_origin_timestamp {
+            fields.push(format!(
+                "iggy_origin_timestamp={}",
+                message.origin_timestamp as i64
+            ));
+        }
+
+        let payload_bytes = message.payload.clone().try_into_vec().map_err(|e| {
```
hubcio: `message.payload.clone().try_into_vec()` deep-clones the entire `Payload` enum per message just to get bytes. for `Payload::Json(OwnedValue)` this is a deep allocation tree clone. N clones per batch at throughput. the SDK should provide a `try_as_bytes(&self)` or the sink should restructure to consume the payload without cloning.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+                    // Capture first error; continue attempting remaining
+                    // batches to maximise data delivery, but record the failure.
+                    if first_error.is_none() {
+                        first_error = Some(e);
+                    }
+                }
+            }
+        }
+
+        let mut state = self.state.lock().await;
```
hubcio: `state.messages_processed += total_messages as u64` counts ALL messages as processed even when some batches failed and the function returns `Err`. only successfully written messages should be counted. the `write_errors` counter separately tracks failures, but the `messages_processed` metric becomes unreliable for monitoring.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+                     Consecutive failures tracked by circuit breaker.",
+                    self.id
+                );
+                tokio::time::sleep(self.poll_interval).await;
+                Err(e)
+            }
+        }
+    }
+
+    async fn close(&mut self) -> Result<(), Error> {
```
hubcio, line 845: `close()` doesn't set `self.client = None` unlike the sink (sink line 766). the reqwest connection pool stays alive after the connector reports "closed". asymmetric cleanup.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+                    .as_ref()
+                    .is_none_or(|current| cursor_value > current)
+            {
+                max_cursor = Some(cursor_value.clone());
+            }
+
+            let payload = self.build_payload(&row, include_metadata)?;
+            messages.push(ProducedMessage {
+                id: Some(Uuid::new_v4().as_u128()),
+                checksum: None,
```
hubcio, line 658: `Utc::now().timestamp_millis()` used here, but the sink's `to_precision_timestamp` treats the timestamp field as microseconds (divides by 1000 for ms, 1000000 for s). a message round-tripping through source -> iggy -> sink would have its timestamp off by 1000x.

also, `Utc::now()` is called twice (lines 689-690) and the values can differ. capture once.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+    fn to_precision_timestamp(&self, micros: u64) -> u64 {
+        match self.timestamp_precision() {
+            "ns" => micros.saturating_mul(1_000),
+            "us" => micros,
+            "ms" => micros / 1_000,
+            "s" => micros / 1_000_000,
+            _ => micros,
+        }
+    }
+    fn line_from_message(
```
hubcio: `line_from_message` creates a `Vec<String>` for tags and fields, does `format!()` per tag/field, then `join()`s them. at high throughput this is dozens of heap allocations per message. use a single `String` buffer with `write!()` / `push_str()`, reusable across messages in a batch.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+
+        let body = lines.join("\n");
+        self.write_with_retry(body).await
+    }
+
+    async fn write_with_retry(&self, body: String) -> Result<(), Error> {
+        let client = self.get_client()?;
+        let url = self.build_write_url()?;
+        let max_retries = self.get_max_retries();
+        let token = self.config.token.clone();
```
hubcio: `open_retry_max_delay` is documented as the upper cap on `open()` backoff but is also used here as the cap for per-write retries. a user setting `open_retry_max_delay = "60s"` for patient startup also gets 60s caps on per-write retries. these should be separate config fields, or at least the naming should reflect the dual usage.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+        for message in messages {
+            lines.push(self.line_from_message(topic_metadata, messages_metadata, message)?);
+        }
+
+        let body = lines.join("\n");
+        self.write_with_retry(body).await
+    }
+
+    async fn write_with_retry(&self, body: String) -> Result<(), Error> {
+        let client = self.get_client()?;
```
hubcio, line 520: `build_write_url()` and `parse_duration()` (line 577) are re-computed every batch write. both parse strings from config that never change at runtime. compute once in `open()` and store on the struct.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+        let include_metadata = self.config.include_metadata.unwrap_or(true);
+        let cursor_field = self.cursor_field().to_string();
+
+        let mut messages = Vec::with_capacity(rows.len());
+        let mut max_cursor: Option<String> = None;
+
+        for row in rows {
+            if let Some(cursor_value) = row.get(&cursor_field)
+                && max_cursor
+                    .as_ref()
```
hubcio: cursor comparison is lexicographic (`String` ordering). works for ISO 8601 timestamps, but `cursor_field` is configurable and could be set to a numeric column where `"9" > "10"` lexicographically. this would cause the cursor to skip data. document the constraint or parse the cursor type.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+enum PayloadFormat {
+    #[default]
+    Json,
+    Text,
+    Base64,
+}
+
+impl PayloadFormat {
+    fn from_config(value: Option<&str>) -> Self {
```
hubcio, line 126: `PayloadFormat::from_config` silently falls through to `Json` on any unrecognized input. a typo like `"jsn"` or `"JSON"` (uppercase - though lowercase conversion handles this) is accepted with no warning. log a warning for the `_` arm so users can spot config mistakes.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+// Parse Retry-After header value (integer seconds or HTTP date)
+fn parse_retry_after(value: &str) -> Option<Duration> {
+    if let Ok(secs) = value.trim().parse::<u64>() {
+        return Some(Duration::from_secs(secs));
+    }
+    // HTTP-date fallback would require httpdate crate; return None to use own backoff
+    None
+}
+
+fn escape_measurement(value: &str) -> String {
```
hubcio: chained `.replace()` calls allocate a new `String` per call (3 in `escape_measurement`, 4 in `escape_tag_value`). in the common case (no special chars), each call allocates and copies for nothing. use a single-pass char-by-char escape writing into one `String`.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// =============================================================================
```
hubcio: this 19-line changelog block comment (lines 19-37) is commit-message material, not source documentation. it will rot as the code evolves - `git log` is the authoritative history. remove it.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+            );
+        } else {
+            debug!(
+                "InfluxDB sink ID: {} processed {} messages. \
+                 Total processed: {}, write errors: {}",
+                self.id, total_messages, state.messages_processed, state.write_errors
+            );
+        }
+
+        // ropagate the first batch error to the runtime so it can
```
hubcio: typo: `// ropagate the first batch error` - should be `// Propagate`.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,774 @@
...
+// ---------------------------------------------------------------------------
+// Main connector structs
+// ---------------------------------------------------------------------------
+
+#[derive(Debug)]
+pub struct InfluxDbSink {
+    pub id: u32,
+    config: InfluxDbSinkConfig,
+    client: Option<Client>,
+    state: Mutex<State>,
```
hubcio: `Mutex<State>` is locked just to bump `messages_processed` and `write_errors` counters. these are only incremented, never read for business logic. use `AtomicU64` instead to avoid locking on the hot path.

core/integration/tests/connectors/fixtures/influxdb/source.rs
```diff
@@ -0,0 +1,131 @@
...
+        let default_flux = format!(
+            r#"from(bucket:"{b}") |> range(start: -1h) |> filter(fn: (r) => r._time > time(v: "$cursor")) |> sort(columns: ["_time"]) |> limit(n: $limit)"#,
+            b = INFLUXDB_BUCKET,
+        );
+
+        let mut envs = HashMap::new();
+        envs.insert(ENV_SOURCE_URL.to_string(), self.container.base_url.clone());
+        envs.insert(ENV_SOURCE_ORG.to_string(), INFLUXDB_ORG.to_string());
+        envs.insert(ENV_SOURCE_TOKEN.to_string(), INFLUXDB_TOKEN.to_string());
+        envs.insert(ENV_SOURCE_BUCKET.to_string(), INFLUXDB_BUCKET.to_string());
```
hubcio: `ENV_SOURCE_BUCKET` is set here but `InfluxDbSourceConfig` has no `bucket` field. serde silently ignores it. either add the field to the config struct or remove this env var.

core/connectors/runtime/example_config/connectors/influx/influxdb_sink.toml
```diff
@@ -0,0 +1,52 @@
...
+schema = "json"
+batch_length = 100
+poll_interval = "5ms"
+consumer_group = "influxdb_sink"
+
+[plugin_config]
+url = "http://localhost:8086"
+org = "iggy"
+bucket = "events"
+token = "my_super_secret_token_123"
```
hubcio: uses `token = "my_super_secret_token_123"` while other example configs (e.g. `connectors/influxdb_source.toml`, `config.toml`) use `"replace-with-token"`. inconsistent and looks like a real token at first glance.

also in `influx/influxdb_source.toml:36`.

hubcio (changes requested):

core/connectors/runtime/example_config/connectors/influx/influxdb_sink.toml
ryerraguntla: Done

ryerraguntla: Will checkin along with other changes. Thanks for reviewing. I will submit PR end of my today.

core/connectors/sdk/src/lib.rs
ryerraguntla: Done

core/connectors/sdk/src/lib.rs
ryerraguntla: Done

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done.

ryerraguntla: format!("offset={}u", message.offset as u64) , similar changes are done in all relevant places.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: New Code - implemented in both places. // True exponential backoff: base * 2^attempt, capped at max_delay
pub fn exponential_backoff(base: Duration, attempt: u32, max_delay: Duration) -> Duration {
    let factor = 2u64.saturating_pow(attempt);
    let millis = base
        .as_millis()                        // u128
        .saturating_mul(factor as u128)     // u128, no overflow
        .min(max_delay.as_millis());        // always keep below max_delay.as_millis()
    let millis_u64 = u64::try_from(millis).unwrap_or(u64::MAX);
    Duration::from_millis(millis_u64)
}

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Implemented the uniqueness with InfluxDB idiomatic offset as tags concept.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: will do the refactor in iteration 2

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Making a single-consumer assumption is naive. Let me give a shot using mutex, may need some suggestions regarding performance. Lets give a try.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Implemented the RFC3339 validation.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: This is pretty smart. Thank you.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Implemented but need a performance test in edge cases.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Implemented with SDk changes. Need more regression testing.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Good catch. Thank you.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: I prefer catching the error , removing the or_else part.

ryerraguntla: Done

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Done

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done and renaming the functions.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: This is a big change. Let me spend some more time on it.

core/connectors/sinks/influxdb_sink/src/lib.rs
hubcio: alright, then please note that in code as TODO.

core/connectors/sinks/influxdb_sink/src/lib.rs
hubcio: You can do it later, but i want to have trace of that in code. either do it now, or leave it as TODO. dont leave comments undocumented or unfixed.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: validated the cursor field for the above conditions.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Using different constants for retries. changed from DEFAULT_OPEN_RETRY_MAX_DELAY  to  DEFAULT_RETRY_MAX_DELAY

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Implemented with Other and logging a warning message .

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: To make more explicit, introduced write_success counter.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done

core/connectors/sdk/src/lib.rs
ryerraguntla: 364 pub enum Error {
365     #[error("Invalid config")]
366     InvalidConfig,
**367     #[error("Invalid Config Value: {0}")]
368     InvalidConfigValue(String),**
369     #[error("Invalid record")]
370     InvalidRecord,
371     #[error("Invalid record value: {0}")]
372     InvalidRecordValue(String),
373     #[error("Invalid transformer")]
374     InvalidTransformer,

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Implemented non Mutex metrics counters and introduced write_success counter.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Got bold and did it :). Refactored into sdk.rs and did a successful cargo test. There is one postgress test case (connectors::postgres::postgres_source::processed_column_source_marks_rows_after_producing) which fails when there is a container contention among the connectors test runs. This is transient and not consistent with reruns. I can increase the timeout but the problem could be limitation of local resource, so leaving it unchanged.

README.md
hubcio, line 1: you should not modify readme, this is not in scope of this PR.

hubcio (changes requested):

README.md
ryerraguntla: Let me roll back. Thanks for pointingout.

ryerraguntla: Rolled back to earlier version.

ryerraguntla (commented):
All suggestions are covered. In a week or so I will improve the test coverage to 70% plus and will also plan some benching marking with performance,For that I need to plan resources.

core/connectors/runtime/example_config/connectors/influx/influxdb_source.toml
```diff
@@ -0,0 +1,54 @@
...
+[[streams]]
+stream = "events"
+topic = "influx_events"
+schema = "json"
+batch_length = 100
+
+[plugin_config]
+url = "http://localhost:8086"
+org = "iggy_org"
+token = "my_super_secret_token_123"
```
hubcio: sink config was fixed to use a placeholder token (`"replace_token_with_secret_token"`) but this file still has `"my_super_secret_token_123"` - looks like a real secret at first glance. should use the same placeholder style as the sink.

core/connectors/sinks/influxdb_sink/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_influxdb_sink"
+version = "0.2.2-edge.1"
+description = "Iggy InfluxDB sink connector for storing stream messages as line protocol"
```
hubcio, line 21: version should be `0.1.0-edge.1` - this is the first release of these connectors, not `0.2.2`.

core/connectors/sources/influxdb_source/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_influxdb_source"
+version = "0.2.2-edge.1"
+description = "Iggy InfluxDB source connector for polling Flux query results"
```
hubcio, line 21: same here - version should be `0.1.0-edge.1`.

hubcio (changes requested):

core/connectors/runtime/example_config/connectors/influx/influxdb_source.toml
ryerraguntla: Fixed.

core/connectors/sinks/influxdb_sink/Cargo.toml
ryerraguntla: [f49cb7d](https://github.com/apache/iggy/pull/2933/commits/f49cb7d8af3bba0a045687c3edd6af535e0e9436)

core/connectors/sources/influxdb_source/Cargo.toml
ryerraguntla: [f49cb7d](https://github.com/apache/iggy/pull/2933/commits/f49cb7d8af3bba0a045687c3edd6af535e0e9436)

ryerraguntla (commented):
Fixed and prechecks were clean locally, Now prechecks are progressing

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+                    if first_error.is_none() {
+                        first_error = Some(e);
+                    }
+                }
+            }
+        }
+
+        let total_processed = self
+            .messages_processed
+            .fetch_add(total_messages as u64, Ordering::Relaxed)
```
hubcio, line 699: the original ask was that only successfully written messages should be counted. `write_success` and `write_errors` were added to track per-batch outcomes (lines 709-716), which is good, but `messages_processed` here is still incremented unconditionally with `total_messages` regardless of how many batches failed - so the counter includes messages from failed batches. either rename to `messages_attempted` to reflect actual semantics, or only increment with the sum of successful batch sizes.

core/connectors/sdk/src/lib.rs
```diff
@@ -340,8 +365,12 @@ pub trait StreamEncoder: Send + Sync {
 pub enum Error {
     #[error("Invalid config")]
     InvalidConfig,
+    #[error("Invalid Config Value: {0}")]
```
hubcio: `#[error("Invalid Config Value: {0}")]` uses title case while `#[error("Invalid record value: {0}")]` at line 372 uses sentence case. pick one style.

hubcio (changes requested):

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Though messages_processed may mean same but could be presumed as success.messages_attempted makes more clear for understanding , will change to messages_attempted.

ryerraguntla (commented):
Made the changes and committed with clean preChecks run

hubcio (changes requested):
-

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+    Text,
+    Raw,
+}
+
+impl PayloadFormat {
+    fn from_config(value: Option<&str>) -> Self {
+        match value.map(|v| v.to_ascii_lowercase()).as_deref() {
+            Some("text") | Some("utf8") => PayloadFormat::Text,
+            Some("raw") | Some("base64") => PayloadFormat::Raw,
+            _ => PayloadFormat::Json,
```
hubcio: source `PayloadFormat::from_config` silently falls through to `Json` on unrecognized input with no warning, while the sink correctly logs a `warn!()` (sink `lib.rs:124-131`). asymmetric fix form earlier comment - should match the sink behavior.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+    }
+
+    async fn run_query_with_retry(&self, query: &str) -> Result<String, Error> {
+        let client = self.get_client()?;
+        let url = self.build_query_url()?;
+        let max_retries = self.get_max_retries();
+        let token = self.config.token.clone();
+
+        // Cap for per-query backoff (reuse open_retry_max_delay config)
+        let max_delay = parse_duration(
```
hubcio: `run_query_with_retry` reuses `open_retry_max_delay` for per-query backoff. I flagged this exact pattern for the sink and it was fixed with a separate `retry_max_delay` config field + `write_retry_max_delay` cached in `open()`. the source has the same issue unfixed - `InfluxDbSourceConfig` has no `retry_max_delay` field, so a user setting `open_retry_max_delay = "60s"` for patient startup also gets 60s caps on per-query retries.

core/integration/tests/connectors/fixtures/influxdb/container.rs
```diff
@@ -0,0 +1,238 @@
...
+                .header("Accept", "application/csv")
+                .json(&body)
+                .send()
+                .await
+                .map_err(|e| TestBinaryError::InvalidState {
+                    message: format!("Failed to query InfluxDB: {e}"),
+                })?;
+
+            let text = response.text().await.unwrap_or_default();
+            eprintln!(
```
hubcio: debug `eprintln!` left in test fixture code. should use `tracing::debug!` or be removed.

core/connectors/runtime/example_config/connectors/influxdb_source.toml
```diff
@@ -0,0 +1,54 @@
...
+[[streams]]
+stream = "events"
+topic = "influx_events"
+schema = "json"
+batch_length = 100
+
+[plugin_config]
+url = "http://localhost:8086"
+org = "iggy"
+token = "replace-with-token"
```
hubcio: this file uses `"replace_token_with_secret_token"` while the top-level `influxdb_source.toml` uses `"replace-with-token"` (different placeholder style). also unclear why both files exist - is the top-level one intentional or a leftover?

hubcio (changes requested):
i found 4 more nits, once you fix these we can merge.

core/integration/tests/connectors/fixtures/influxdb/container.rs
ryerraguntla: Done. Will be part of next commit.

core/connectors/runtime/example_config/connectors/influxdb_source.toml
ryerraguntla: Removed the influx specific directory and move the influx sink/source toml files in the same place as other connectors . Secret token is now same across all influx specific configs.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Done.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Implemented similar logic.

ryerraguntla (commented):
Committed the last set of nits. Once this goes with a clean prechecks, I will add some more test case to cover different data format messages.

ryerraguntla (commented):
Fixed the nits and added test cases to increase code coverage.

core/connectors/sdk/src/retry.rs
```diff
@@ -0,0 +1,159 @@
+/* Licensed to the Apache Software Foundation (ASF) under one
```
hubcio, line 1: the connector runtime already uses `reqwest-middleware` + `reqwest-retry` (`ExponentialBackoff`, `RetryTransientMiddleware`, `Jitter::Bounded`) in `http_provider.rs`. this PR hand-rolls ~160 lines of retry/backoff/jitter/circuit-breaker instead of reusing that infrastructure. suggest using `reqwest-middleware` with a custom retry policy that handles Retry-After and circuit breaking - it would eliminate `write_with_retry()` in the sink and `run_query_with_retry()` in the source entirely, and keep the connector crates consistent with the runtime.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+            if i > 0 {
+                body.push('\n');
+            }
+            self.append_line(&mut body, topic_metadata, messages_metadata, message)?;
+        }
+
+        self.write_with_retry(body).await
+    }
+
+    async fn write_with_retry(&self, body: String) -> Result<(), Error> {
```
hubcio: `write_with_retry()` is a hand-rolled retry loop duplicating what `reqwest_retry::RetryTransientMiddleware` already does - transient status detection, exponential backoff, jitter. the connector runtime uses `reqwest-middleware` for this exact pattern - see `http_provider.rs:82-101`. wrapping `reqwest::Client` in `ClientWithMiddleware` would eliminate this loop and the source's `run_query_with_retry` equivalent entirely. the circuit breaker and Retry-After parsing can be a custom `reqwest_retry::RetryPolicy`.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+        }
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Sink trait implementation
+// ---------------------------------------------------------------------------
+
+#[async_trait]
+impl Sink for InfluxDbSink {
```
hubcio, line 569: 60.46% patch coverage. untested critical paths: `write_with_retry` HTTP 429/5xx/network-error retry branches, circuit breaker open-state skip in `consume`, `check_connectivity_with_retry` failure loop, `to_precision_timestamp` for `"ms"` and `"s"` precision, `append_line` error paths (invalid JSON, invalid UTF-8), escaping edge cases (`=` in tags, `\` and `,` in measurement names), partial batch failure (first chunk fails, rest succeed).

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,803 @@
...
+        if query.contains("$cursor") {
+            query = query.replace("$cursor", cursor);
+        }
+        if query.contains("$limit") {
+            query = query.replace("$limit", &limit);
+        }
+        Ok(query)
+    }
+
+    async fn run_query_with_retry(&self, query: &str) -> Result<String, Error> {
```
hubcio: 61.74% patch coverage. untested critical paths: `run_query_with_retry` HTTP 429/5xx/network-error retry branches, circuit breaker open-state skip in `poll`, `validate_cursor` rejection of Flux injection payloads, `validate_cursor_field` rejection of non-timestamp fields, `parse_csv_rows` with blank lines / `#` annotations / multi-table results / missing columns, `build_payload` with missing `payload_column` / invalid base64 / invalid JSON, `close()` client cleanup.

hubcio (changes requested):
after seeing @spetz review under #2925 i decided to add more comments :)

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Increased to 89%.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Taken care of the tests. and code coverage is now 89%

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Reused the middleware.

core/connectors/sdk/src/retry.rs
ryerraguntla: Taken care of  write_with_retry() in the sink and run_query_with_retry() in the source

ryerraguntla (commented):
Improved the code coverage and refactored code as suggested.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1067 @@
...
+            '=' => buf.push_str("\\="),
+            ' ' => buf.push_str("\\ "),
+            _ => buf.push(ch),
+        }
+    }
+}
+
+/// Write an escaped string field value (without surrounding quotes) into `buf`.
+/// Escapes: `\` → `\\`, `"` → `\"`
+fn write_field_string(buf: &mut String, value: &str) {
```
hubcio, line 191: `write_field_string` escapes `\` and `"` but does not escape `\n` (newline). in influxdb line protocol, newline is the record delimiter - the write API splits on `\n` boundaries *before* parsing individual lines. if any json payload, text payload, or string field value contains a literal newline, it will split one line-protocol record into two malformed fragments. the batch write will either be rejected or - worse - silently produce a corrupt point with garbage measurement/tag/field structure.

fix: add `'\n' => buf.push_str("\\n"),` (and `'\r' => buf.push_str("\\r"),` for completeness) to the match arms.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+        let cursor_field = self.cursor_field().to_string();
+
+        let mut messages = Vec::with_capacity(rows.len());
+        let mut max_cursor: Option<String> = None;
+
+        for row in rows {
+            if let Some(cursor_value) = row.get(&cursor_field)
+                && max_cursor
+                    .as_ref()
+                    .is_none_or(|current| is_timestamp_after(cursor_value, current))
```
hubcio: cursor advancement uses `is_timestamp_after` (strict `>`) to find the max `_time` in the batch, and the example query templates use `r._time > time(v: "$cursor")` (also strict `>`).

this means rows with `_time == cursor` are excluded from the next poll. if multiple rows share the exact same max `_time` and the batch limit truncates before all same-timestamp rows are fetched, the remaining rows at that timestamp are permanently skipped - classic cursor-boundary data loss.

mitigations: use `>=` in the query and deduplicate already-seen rows by tracking the last-seen offset count at the cursor timestamp, or document the at-most-once limitation clearly so operators know to size batches larger than any expected timestamp collision.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,1166 @@
...
+    }
+
+    async fn poll(&self) -> Result<ProducedMessages, Error> {
+        // Skip query if circuit breaker is open
+        if self.circuit_breaker.is_open().await {
+            warn!(
+                "InfluxDB source ID: {} — circuit breaker is OPEN. Skipping poll.",
+                self.id
+            );
+            return Ok(ProducedMessages {
```
hubcio, line 748: when the circuit breaker is open, `poll()` returns `Ok(ProducedMessages { messages: vec![], .. })` with no sleep. the connector runtime sees a successful empty poll and immediately calls `poll()` again, creating an unbounded cpu spin-loop for the entire circuit breaker cool-down period. `is_open()` acquires a tokio mutex on each iteration, adding contention on top of the cpu burn.

add `tokio::time::sleep(self.poll_interval).await;` before the return so the source backs off while the circuit is open. alternatively, return `Err(...)` to match the sink's behavior (sink returns `Err(CannotStoreData(...))` at its equivalent check).

hubcio (changes requested):

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done. Also let me check for any other characters.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Done.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Done, bit differently.

ryerraguntla (commented):
Completed the requested changes.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1104 @@
...
+    retry_delay: Duration,
+    circuit_breaker: Arc<CircuitBreaker>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct InfluxDbSinkConfig {
+    pub url: String,
+    pub org: String,
+    pub bucket: String,
+    pub token: String,
```
hubcio: also at `core/connectors/sources/influxdb_source/src/lib.rs:94`.

every other connector in the codebase uses `SecretString` from the `secrecy` crate for credentials - elasticsearch uses `Option<SecretString>` for password, mongodb uses `SecretString` for connection_uri, postgres uses `SecretString` for connection_string. the influxdb API token grants full bucket read/write access and will show up in debug logs and serialized config dumps as a plain string. switch to `SecretString` with `serialize_secret` and use `.expose_secret()` at the two points where it's interpolated into the `Authorization` header.

core/connectors/sdk/src/retry.rs
```diff
@@ -0,0 +1,463 @@
...
+///
+/// The `max_retries` parameter is the *total attempt count* (not the number
+/// of extra attempts), consistent with the rest of the connector retry config:
+/// - `max_retries = 1` → one attempt, no retries on failure
+/// - `max_retries = 3` → up to three attempts (two retries after a failure)
+///
+/// Non-transient error responses (4xx except 429) are returned as-is so
+/// callers can inspect the status and body to build a meaningful error.
+#[derive(Debug, Clone)]
+pub struct InfluxDbRetryMiddleware {
```
hubcio: this struct lives in `iggy_connector_sdk::retry` - the shared SDK crate for all connectors - but the name and log messages (e.g. line 256: `"InfluxDB error"`) are influxdb-specific. the behavior itself is fully generic (retry transient HTTP errors with exponential backoff). rename to `HttpRetryMiddleware` (or similar) and parameterize the log prefix so the next connector author can reuse it without importing something misleadingly named or copy-pasting.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,1265 @@
...
+    }
+    if let Ok(v) = value.parse::<f64>()
+        && let Some(number) = serde_json::Number::from_f64(v)
+    {
+        return serde_json::Value::Number(number);
+    }
+    serde_json::Value::String(value.to_string())
+}
+
+fn is_header_record(record: &StringRecord) -> bool {
```
hubcio, line 219: this requires both `_time` AND `_value` to be present, but flux aggregation queries (`count()`, `mean()`, `group()`) can produce result tables with different column layouts - e.g. `_time` + `_count` instead of `_value`. when that happens the header row won't be recognized, and all subsequent data rows until the next recognized header will be silently skipped. checking for `_time` alone (or detecting influxdb CSV annotation rows like `#group`, `#datatype`, `#default`) would be more robust.

core/connectors/runtime/example_config/connectors/influxdb_source.toml
```diff
@@ -0,0 +1,54 @@
...
+poll_interval = "5s"
+batch_size = 500
+cursor_field = "_time"
+initial_offset = "1970-01-01T00:00:00Z"
+include_metadata = true
+payload_format = "json"
+max_retries = 3
+retry_delay = "1s"
+timeout = "10s"
+precision = "us"
```
hubcio: `InfluxDbSourceConfig` has no `precision` field - this key is silently ignored by serde deserialization. a user copying this example and setting `precision = "ns"` will get no effect and no error. either add the field to the config struct and use it, or remove this line.

core/connectors/sdk/src/retry.rs
```diff
@@ -0,0 +1,463 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Duration / backoff helpers
+// ---------------------------------------------------------------------------
+
+/// Parse a human-readable duration string (e.g. `"5s"`, `"1m30s"`) using
+/// [`humantime`]. Falls back to 1 second if parsing fails.
+pub fn parse_duration(value: Option<&str>, default_value: &str) -> Duration {
```
hubcio, line 134: if the user-provided value is malformed (e.g. `retry_delay = "5sec"` instead of `"5s"`), this silently returns 1 second with no log. a `warn!()` on the fallback path would save users from debugging mysterious retry timing.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1104 @@
...
+            {
+                Ok(()) => {
+                    // Successful write — reset circuit breaker
+                    self.circuit_breaker.record_success();
+                    self.write_success
+                        .fetch_add(batch.len() as u64, Ordering::Relaxed);
+                }
+                Err(e) => {
+                    // Failed write — notify circuit breaker
+                    self.circuit_breaker.record_failure().await;
```
hubcio: `record_failure()` is called for all errors including non-transient ones like 400 Bad Request (malformed line protocol) or 422 (schema conflict). the retry middleware already handles transient errors (429, 5xx), so errors that reach here are likely permanent data issues. counting them toward the circuit breaker threshold means a burst of bad data can open the circuit and block delivery of valid messages in subsequent `consume()` calls. consider only calling `record_failure()` for transient/connectivity errors.

hubcio (changes requested):

core/connectors/runtime/example_config/connectors/influxdb_source.toml
ryerraguntla: Removed as it is applicable only for source.

core/connectors/sdk/src/retry.rs
ryerraguntla: Logs messages for the non standard format.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Good idea. Done,

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Fixed by  not looking for  both. Just check for time only.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Done.

core/connectors/sdk/src/retry.rs
ryerraguntla: Done.

ryerraguntla (commented):
All suggestions taken care of

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1123 @@
...
+        }
+        if include_origin_timestamp {
+            use std::fmt::Write as _;
+            write!(buf, ",iggy_origin_timestamp={}u", message.origin_timestamp)
+                .expect("write to String is infallible");
+        }
+
+        // ── Payload field ────────────────────────────────────────────────────
+        // try_as_bytes() serialises in-place without cloning the Payload tree.
+        let payload_bytes = message.payload.try_as_bytes().map_err(|e| {
```
hubcio: triple serialization in the json payload path: `try_as_bytes()` serializes the `simd_json::OwnedValue` to `Vec<u8>`, then `serde_json::from_slice` deserializes it back into `serde_json::Value`, then `serde_json::to_string` re-serializes to a compact string. three passes per message.

you can match on `Payload::Json` directly here and use `simd_json::to_string` to produce the compact json string in one pass, skipping the bytes round-trip entirely. also avoids the subtle semantic difference between simd_json and serde_json representations.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1123 @@
...
+    fn build_health_url(&self) -> Result<Url, Error> {
+        let base = self.config.url.trim_end_matches('/');
+        Url::parse(&format!("{base}/health"))
+            .map_err(|e| Error::InvalidConfigValue(format!("Invalid InfluxDB URL: {e}")))
+    }
+
+    /// Single connectivity probe using the provided raw client (no retry).
+    /// The caller (`check_connectivity_with_retry`) is responsible for the
+    /// outer retry loop, which uses different delay bounds than per-write retries.
+    async fn check_connectivity(&self, client: &reqwest::Client) -> Result<(), Error> {
```
hubcio: also at source lines 293-397.

`check_connectivity`, `check_connectivity_with_retry`, `build_health_url`, `build_raw_client`, and `get_client` are near-identical between sink and source (~80 lines). the only difference is log message prefixes. worth extracting to a shared helper in the SDK crate to avoid future divergence.

core/connectors/sdk/src/lib.rs
```diff
@@ -155,6 +157,29 @@ impl Payload {
...
+    ///  for the output `Vec<u8>`, zero clone of the value tree.
+    ///Raw:  returns a copy of the inner bytes (unavoidable — caller
+    ///  needs owned bytes and we only have a reference).
+    ///Text/`Proto: copies the string bytes (same reasoning).
+    ///FlatBuffer:  copies the buffer bytes.
+    ///
+    /// For `Json` this replaces a deep clone of the entire `OwnedValue` tree
+    /// with a single serialisation pass — O(n) work either way, but the clone
+    /// path does O(n) allocation + O(n) serialisation, while this path does
+    /// only O(n) serialisation.
```
hubcio, line 173: per rust api guidelines, `as_` implies a cheap borrowed view while `to_` implies an owned conversion that may allocate. this returns `Vec<u8>` from `&self`, so `try_to_bytes()` would be more idiomatic. `try_into_vec` right above is correctly named - this one should follow the same convention before the SDK stabilizes.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1123 @@
...
+    }
+
+    fn measurement(&self) -> &str {
+        self.config
+            .measurement
+            .as_deref()
+            .unwrap_or("iggy_messages")
+    }
+
+    fn payload_format(&self) -> PayloadFormat {
```
hubcio, line 282: also at source line 278.

`payload_format()` calls `PayloadFormat::from_config()` which does `to_ascii_lowercase()` on every invocation - one string allocation per message on the hot path. easy fix: resolve the `PayloadFormat` once in `new()` and store it in the struct.

hubcio (changes requested):
once you fix these I think we can merge.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: refactored.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Good one.

core/connectors/sdk/src/lib.rs
ryerraguntla: changed to try_to_bytes()

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: changed to simd_json with bit of hesitation as simd is cpu architecture dependent and performance with different cpus could be different. Lets see what the users say about it.

ryerraguntla (commented):
Implemented the suggestions.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,1241 @@
...
+                        if state.last_timestamp.as_deref() != Some(new_cursor.as_str()) =>
+                    {
+                        // Cursor advanced to a new timestamp — reset the row counter.
+                        state.last_timestamp = max_cursor.clone();
+                        state.cursor_row_count = rows_at_max_cursor;
+                    }
+                    Some(_) => {
+                        // Cursor stayed at the same timestamp — accumulate so the
+                        // next poll skips all already-delivered rows at this timestamp.
+                        state.cursor_row_count =
```
hubcio, line 774: `cursor_row_count` accumulates via `saturating_add` across polls but never resets when the cursor doesn't advance. if many rows share a timestamp across repeated polls, this counter grows unbounded. once it exceeds the actual row count at that timestamp, the skip logic at line 589 skips everything, producing 0 messages - connector gets permanently stuck. needs a cap (e.g. reset when poll produces 0 messages after skipping) or bound to batch_size.

core/connectors/sources/influxdb_source/src/lib.rs
```diff
@@ -0,0 +1,1241 @@
...
+        // Build circuit breaker from config
+        let cb_threshold = config
+            .circuit_breaker_threshold
+            .unwrap_or(DEFAULT_CIRCUIT_BREAKER_THRESHOLD);
+        let cb_cool_down = parse_duration(
+            config.circuit_breaker_cool_down.as_deref(),
+            DEFAULT_CIRCUIT_COOL_DOWN,
+        );
+
+        let restored_state = state
```
hubcio: if the `State` struct gains or removes a field in a future version, `deserialize()` returns `None` and the connector silently starts polling from `1970-01-01T00:00:00Z` - full re-delivery with no alert. should at minimum log at error level with the connector ID, or fail `open()` so the operator knows the cursor was reset.

core/connectors/sinks/influxdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1555 @@
...
+        let mut first_error: Option<Error> = None;
+
+        for batch in messages.chunks(batch_size.max(1)) {
+            match self
+                .process_batch(topic_metadata, &messages_metadata, batch)
+                .await
+            {
+                Ok(()) => {
+                    // Successful write — reset circuit breaker
+                    self.circuit_breaker.record_success();
```
hubcio: within a single `consume()` call, if batch 1 fails but batch 2 succeeds, `record_success()` here resets the circuit breaker's failure counter. the circuit never trips if at least one batch per call succeeds - repeated failures on the same data are masked. fix: move `record_success()` after the loop, guarded by `first_error.is_none()`.

hubcio (changes requested):
found more issues, please fix, overall its looking acceptable

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: These two are similar to the sink side ones too. I should have taken care along with them. Sorry about that.

core/connectors/sources/influxdb_source/src/lib.rs
ryerraguntla: Done.

core/connectors/sinks/influxdb_sink/src/lib.rs
ryerraguntla: Taken care.

hubcio (approved):
lgtm, it's the longest review I ever did. good job!

mmodzelewski (approved):

ryerraguntla (commented):
All suggestions are implemented, reviewed and approved. Needs to be merged by the maintainer.

hubcio (approved):

spetz (approved):

PR #2937, opened 2026-03-15
refactor(csharp): implement SDK version handling in login requests

lukaszzborek:
- change version passing during login from static value to assembly version
- change PackageVersion attribute to Version in .csproj

-----
PackageVersion attribute was changed because it is not visible in the assembly. If PackageVersion attribute is not set, Version attribute is used as the default.
Also I checked CI, and in all places, I see PackageVersion|Version check, so no need to change it


foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
```diff
@@ -728,7 +728,7 @@ public async Task ChangePassword(Identifier userId, string currentPassword, stri
             throw new NotConnectedException();
         }
 
-        var message = TcpContracts.LoginUser(userName, password, "0.5.0", "csharp-sdk");
+        var message = TcpContracts.LoginUser(userName, password, SdkVersion.Value, "csharp-sdk");
```
hubcio, line 732: for now this is fine, but in future you'll insert a actual binary protocol crate version. can you make it a TODO? it'll be embedded into all libs during compilation step. (or hardcoded).

lukaszzborek: yep, todo added

PR #2939, opened 2026-03-15
feat(java): Increase test coverage

### LINKED ISSUE #2938: [Java SDK] Increase test coverage
### Description

The existing Iggy Java SDK codebase is well covered with integration tests for the supported clients. However, there are quite a few places where there's little to no test coverage. I have seen other SDKs strive to increase their test coverage recently and I thought the Java SDK could do the same

### Component

Java SDK

### Proposed solution

Increase test coverage by writing unit tests to cover most logic in the codebase, particularly those that are not covered by existing unit or integration tests

### Alternatives considered

N/A

jrhenderson1988:

Closes #2938 

## Rationale

Intention is to increase test coverage where it is missing and generally add unit tests where appropriate

## What changed?

- Adds unit test coverage to most areas of the SDK (notable exception is `client` package since they're quite difficult to test how they are written and are already covered well with integration tests)
- Fixes some small issues found during the process (I'll leave comments on the individual items)

(sorry for the PR size 😬)

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

Only AI usage was Jetbrains local full line completion model for Java. Used only for autocomplete. I can explain everything if asked.


foreign/java/java-sdk/src/main/java/org/apache/iggy/exception/IggyException.java
```diff
@@ -46,13 +46,4 @@ protected IggyException(String message) {
     protected IggyException(String message, Throwable cause) {
         super(message, cause);
     }
-
-    /**
-     * Constructs a new IggyException with the specified cause.
-     *
-     * @param cause the cause of the exception
-     */
-    protected IggyException(Throwable cause) {
```
jrhenderson1988, line 55: I found that there were no usages of this constructor by subclasses - and since this class is abstract, it's not really reachable. It couldn't be properly tested within making a new class that used it (or an anonymous class). Technically a consumer of the SDK might be able to make their own subclass of `IggyException` but I am not sure if that would make much sense in the context use.

Therefore, I removed it. I'd be happy to put it back if preferred

foreign/java/java-sdk/src/main/java/org/apache/iggy/system/ClientInfoDetails.java
```diff
@@ -30,7 +29,7 @@ public record ClientInfoDetails(
         String transport,
         Long consumerGroupsCount,
         List<ConsumerGroupInfo> consumerGroups) {
-    public ClientInfoDetails(ClientInfo clientInfo, ArrayList<ConsumerGroupInfo> consumerGroups) {
+    public ClientInfoDetails(ClientInfo clientInfo, List<ConsumerGroupInfo> consumerGroups) {
```
jrhenderson1988, line 32: I assume it was a mistake to only allow an `ArrayList` in the constructor here. The property is a `List<ConsumerGroupInfo>`. Changing this to a list makes it more open than it was before, so it shouldn't be breaking in any way. Again, happy to reverse this if preferred

foreign/java/java-sdk/src/main/java/org/apache/iggy/topic/CompressionAlgorithm.java
```diff
@@ -31,7 +31,7 @@ public enum CompressionAlgorithm {
         this.code = code;
     }
 
-    public static CompressionAlgorithm fromCode(byte code) {
+    public static CompressionAlgorithm fromCode(int code) {
```
jrhenderson1988, line 34: Also noticed this when writing a test. The property stored in the `enum` is an `Integer` and all other enums use `int` in their `fromCode` methods. I assumed this was a mistake too, so I updated it here. Again, happy to reverse it.

foreign/java/java-sdk/src/main/java/org/apache/iggy/system/ClientInfoDetails.java
mmodzelewski: Good catch, I'm sure it was an overlook

PR #2944, opened 2026-03-16
feat(server): implement TwoHalves buffer

numinnex:
This PR introduces an `TwoHalves` buffer, that uses `Vec<u8>` instead of `Bytes` as it's underlying storage.

TwoHalves splits the underlying storage into two parts (mutable and non-mutable one). The non-mutable part uses thread-safe reference counting for cheap sharing (Bytes like), where the mutable one is copied over on each clone.

In the initial state both the mutable and non-mutable parts point to the same allocation using exactly the same control block, the diverge happens on first `clone` when the mutable part gets copied over and re-allocated entirely.


core/buf/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        ref_count: AtomicUsize::new(1),
+        base,
+        len,
+        cap,
+    });
+    // SAFETY: `ctrlb` is a valid control block for the lifetime of the returned halves.
+    unsafe { NonNull::new_unchecked(Box::into_raw(ctrlb)) }
+}
+
+pub struct TwoHalves {
```
krishvishal: Is `TwoHalves` made for representing mutable header and immutable body? 

If yes, I think naming `HeaderBody` or something similar makes the intention more clear. Because I find `TwoHalves` more abstract and makes it hard to understand what it is for.

numinnex: I wanted to create general purpose mechanism for an split buffer. Yes we will use it for header + body of `Message`, but we might want to use it for something else aswell.

core/iobuf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        f.debug_struct("TwoHalves")
+            .field("split_at", &self.split_at())
+            .field("head_len", &self.inner.0.len)
+            .field("tail_len", &self.inner.1.len)
+            .field("halves_alias", &(self.inner.0.ctrlb == self.inner.1.ctrlb))
+            .finish()
+    }
+}
+
+#[derive(Clone)]
```
hubcio, line 201: `Frozen` is dead code with guaranteed memory leak. derives `Clone` (increments refcount via `Extent::clone`) but has no `Drop` and no constructor. every drop leaks the `ControlBlock` + backing allocation. either add `Drop` or delete it entirely per project rules.

core/iobuf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+                    len: len - split_at,
+                    ctrlb,
+                    _pad: 0,
+                },
+            ),
+        }
+    }
+}
+
+pub struct TwoHalves<const ALIGN: usize> {
```
hubcio, line 90: `TwoHalves`, `Owned`-after-split, and `Frozen` are all `!Send + !Sync` because `NonNull` is `!Send + !Sync`. the `AtomicUsize` refcounting strongly implies cross-thread use was intended. either add `unsafe impl Send/Sync` with a safety argument, or document the `!Send` constraint as intentional for the shard-local model.

core/buf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
+        &mut self.inner
+    }
+
+    /// Split `Owned` buffer into two halves
+    ///
+    /// # Panics
+    /// Panics if `split_at > self.len()` or if `split_at` is not a multiple of `ALIGN` bytes.
```
hubcio: doc says "Panics if `split_at` is not a multiple of `ALIGN` bytes" but the only assertion at line 56 is `split_at <= self.inner.len()`. no alignment check exists. either add `assert!(split_at % ALIGN == 0)` or fix the doc.

core/buf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        Extent {
+            ptr: data,
+            len,
+            ctrlb,
+            _pad: 0,
+        }
+    }
+}
+
+impl Clone for Extent {
```
hubcio: `Extent::clone()` increments the refcount but `Extent` has no `Drop` - refcounting is a one-way ratchet. the design relies entirely on container types (`TwoHalves::drop`) to manage the lifecycle. this is the root cause of the `Frozen` leak. consider implementing `Drop` for `Extent` to make refcounting self-contained.

core/iobuf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    inner: (Extent, Extent),
+}
+
+impl<const ALIGN: usize> TwoHalves<ALIGN> {
+    pub fn head(&self) -> &[u8] {
+        self.inner.0.as_slice()
+    }
+
+    pub fn head_mut(&mut self) -> &mut [u8] {
```
hubcio, line 104: the safety comment says "head is not shared between clones, instead it gets copied" but that's only true for the clone's head. in the non-cloned case, head points into the same allocation as tail. mutating head is safe because regions don't overlap, but `&mut` and `&` to the same allocation is a stacked borrows grey area. if `Send` is ever added, two threads could hold original + clone, creating aliasing across threads. miri would detect that.

core/buf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+            ctrlb.base.as_ptr(),
+            ALIGN,
+            ctrlb.len,
+            ctrlb.capacity,
+        )
+    };
+}
+
+unsafe fn reclaim_unique_control_block(ctrlb: NonNull<ControlBlock>) -> ControlBlock {
+    debug_assert_eq!(
```
hubcio: `reclaim_unique_control_block` uses `debug_assert_eq` for a safety precondition. in release builds, if refcount != 1 due to a bug, `Box::from_raw` reclaims the control block while other references exist - use-after-free. promote to `assert_eq` for defense-in-depth.

core/buf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        // SAFETY: We are accessing the head half mutably, this is the only correct operation, as the head is not shared between clones,
+        // instead it gets copied.
+        unsafe { self.inner.0.as_mut_slice() }
+    }
+
+    pub fn tail(&self) -> &[u8] {
+        self.inner.1.as_slice()
+    }
+
+    pub fn split_at(&self) -> usize {
```
hubcio: `TwoHalves::split_at()` returns the split position, `Owned::split_at()` performs the split. same name, very different semantics. rename this accessor to `split_point()` or `head_len()`.

core/iobuf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+        // SAFETY: Box::into_raw returns a valid pointer
+        unsafe { NonNull::new_unchecked(Box::into_raw(ctrl)) }
+    }
+}
+
+struct Extent {
+    ptr: NonNull<u8>,
+    len: usize,
+    ctrlb: NonNull<ControlBlock>,
+    _pad: usize,
```
hubcio, line 260: `_pad: usize` brings `Extent` to 32 bytes but the purpose is undocumented. if it's for cache-line alignment, say so. if not needed, remove it.

core/buf/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "buf"
```
hubcio: `buf` is very generic for a workspace of 37+ crates. consider `iggy_buf` to namespace consistently with `iggy_common`, `iggy_binary_protocol`, etc.

core/iobuf/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "buf"
+version = "0.1.0"
+edition = "2024"
```
hubcio, line 21: missing `license` field. DEPENDENCIES.md shows "N/A". add `license = "Apache-2.0"` to match all other internal crates.

core/buf/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+    }
+}
+
+#[repr(C, align(64))]
+struct ControlBlock {
+    ref_count: AtomicUsize,
+    base: NonNull<u8>,
+    len: usize,
+    capacity: usize,
+    _pad: [u8; 32],
```
hubcio: `_pad: [u8; 32]` is manually computed assuming 8-byte pointers. breaks on 32-bit targets. let `#[repr(C, align(64))]` handle padding implicitly without the manual field.

hubcio (changes requested):

core/buf/src/lib.rs
numinnex: Yeah, it's artifact from previous iteration, Ill remove it

core/iobuf/src/lib.rs
numinnex: Both of them are `Send`, so I will add impl for those

core/iobuf/src/lib.rs
numinnex: and in the non-clone case, we cannot have both mutable and non-mutable reference to the `TwoHalves`, only way is to clone and then the `head` becomes an distinct allocation.

core/iobuf/src/lib.rs
numinnex: Yeah I need to add explicit `Drop` impl

core/buf/src/lib.rs
numinnex: Indeed, I will remove that field

core/iobuf/src/lib.rs
numinnex: I will document it

core/buf/Cargo.toml
numinnex: I was thinking about `iobuf` as name, maybe that's better.

hubcio (approved):

mmodzelewski (approved):

spetz (approved):

PR #2945, opened 2026-03-16
test(server): add regression test to ensure secrets are never persisted as plaintext

### LINKED ISSUE #2943: Add regression test to verify secrets are never persisted as plaintext
Add an integration test that verifies plaintext passwords and raw PAT values are never written to disk. Currently, hashing before persistence is a convention enforced in shard/execution.rs (calling crypto::hash_password before state.apply()), but there's no compile-time enforcement or runtime verification.

**Proposed Test**

1. Create a user with a known password
2. Create a personal access token
3. Shut down the server (flush all state to disk)
4. Scan all files under local_data for the plaintext password and raw PAT value
5. Assert neither appears anywhere on disk

A simple content grep across all persisted files should be sufficient.

**Motivation**

This came up during #2728 (SecretString migration). The SecretString wrapper prevents accidental leaks in logs and debug output, but doesn't guarantee the persistence layer always hashes before writing. A grep-based regression test would catch it if someone ever bypasses crypto::hash_password before state.apply().

yeyomontana:
## Summary
Adds an integration regression test verifying that plaintext passwords and raw
personal access tokens (PATs) are never written to disk.

## Changes
- Added `verify_no_plaintext_credentials_on_disk.rs` to the data_integrity integration tests
- Creates a user with a known plaintext password
- Creates a personal access token
- Stops the server to flush persisted state
- Recursively scans the persisted server data directory
- Asserts that neither the raw password nor the raw PAT appear in any file

## Test Plan
Run:
cargo test verify_no_plaintext_credentials_on_disk

The test verifies that after server shutdown and persistence flush,
no plaintext credential values exist in the persisted storage files.

Closes #2943


core/integration/tests/data_integrity/verify_no_plaintext_credentials_on_disk.rs
```diff
@@ -0,0 +1,105 @@
...
+use iggy::prelude::*;
+use integration::iggy_harness;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+const USERNAME: &str = "plaintext-regression-user";
+const PLAINTEXT_PASSWORD: &str = "plaintext-password-regression-2943";
+const PAT_NAME: &str = "plaintext-regression-pat";
+
+#[iggy_harness(test_client_transport = [Tcp, Http, Quic, WebSocket])]
```
hubcio: there is no reason to make this test run for 4 protocols, use just iggy_harness without arguments

yeyomontana: Fixes implemented. Kindly check.

PR #2946, opened 2026-03-16
feat(rust): add wire protocol codec and types to binary_protocol crate

hubcio:
Pure codec layer (no I/O, no async) with protocol-owned
types decoupled from iggy_common. Adds WireEncode/WireDecode
traits, WireName validated newtype, WireIdentifier, command
codes, stream request/response types, VSR consensus headers
with bytemuck zero-copy, and Operation-to-CommandCode mapping.
62 tests. First step - remaining domains to follow.


core/binary_protocol/src/consensus/message.rs
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    /// Access the typed header (zero-copy pointer cast).
+    #[must_use]
+    #[allow(clippy::missing_panics_doc)]
+    #[inline]
+    pub fn header(&self) -> &H {
+        let header_bytes = &self.buffer[..size_of::<H>()];
+        bytemuck::checked::try_from_bytes(header_bytes)
```
krishvishal, line 110: Do we have to validate the header every time its accessed? I think its validated at the construction time, so `from_bytes` is safe to use here.

core/binary_protocol/src/codes.rs
```diff
@@ -0,0 +1,218 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::WireError;
+
+// -- System --
+pub const PING_CODE: u32 = 1;
```
krishvishal, line 24: Consider grouping these constants into an enum similar to `Opeartion` enum, because we would get type safety.

krishvishal: We then could use `TryFrom` at the conversion boundary.

core/binary_protocol/src/consensus/operation.rs
```diff
@@ -0,0 +1,218 @@
...
+    UpdateStream = 129,
+    DeleteStream = 130,
+    PurgeStream = 131,
+    CreateTopic = 132,
+    UpdateTopic = 133,
+    DeleteTopic = 134,
+    PurgeTopic = 135,
+    CreatePartitions = 136,
+    DeletePartitions = 137,
+    DeleteSegments = 138,
```
krishvishal, line 42: Lets move the `DeleteSegments` from here to the range of  partition operations discriminants.

core/binary_protocol/src/consensus/message.rs
hubcio: good catch, fixed

hubcio: eh, we cannot do it this way:  switching to the unchecked `bytemuck::try_from_bytes` but it requires `Pod AnyBitPattern` which our headers cantimplement - they contain `CheckedBitPattern` enums (`Command2`, `Operation`). the only way to skip re-validation would be an unsafe pointer cast. are you OK with current approach?

core/binary_protocol/src/consensus/operation.rs
hubcio: I added TODO for it. Will implement after iggy_common `Operation` is removed.

core/binary_protocol/src/codes.rs
hubcio: agreed, nice improvement, will do it after initial codec is merged as follow up - added todo

core/binary_protocol/src/consensus/message.rs
krishvishal: So basically alignment failure is the error here that can happen spuriously at runtime right? 

If the alignment of underlying buffer is ensured (16 bytes) then we can use unsafe pointer cast.

hubcio: dont worry about it - we will have to fix that once miri is in place. (this week probably)

PR #2957, opened 2026-03-17
refactor(go): refactor WithTLS

chengxilo:

N/A

## Rationale

Currently, the TLS options requiring a user to call `WithTLS(true)` in addition to `WithTLSDomain` or `WithTLSCAFile`. This is redundant and error-prone. If a user provides a domain or a CA file, the intent to use TLS is already clear. 

We can use nested option pattern for it, therefore we can create client like this:
```go
cli, err := client.NewIggyClient(
	client.WithTcp(
		tcp.WithServerAddress(connectAddr),
		tcp.WithTLS(
			tcp.WithTLSCAFile(caFile),
			tcp.WithTLSDomain("localhost"),
		),
	),
)
```

## What changed?
1. Refactor WithTLS method.
2. Extract TLS configuration struct, group all TLS-related parameters together

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage

1. Claude Sonnet 4.6 
3. code generation
4. I reviewed the code and tested
5. Sure

### DIFF
--- a/foreign/go/client/tcp/tcp_core.go
+++ b/foreign/go/client/tcp/tcp_core.go
@@ -352,7 +352,7 @@ func (c *IggyTcpClient) connect() error {
 		attempts = uint(c.config.reconnection.maxRetries)
 		interval = c.config.reconnection.interval
 	}
-	// TODO handle tls logic
+
 	var conn net.Conn
 	if err := retry.New(
 		retry.Attempts(attempts),


foreign/go/client/tcp/tcp_core.go
mmodzelewski: @chengxilo would you mind removing this comment, I believe we've got everything for TLS already? :)

chengxilo: Oh sure. Wondering why it's still there😂

chengxilo: solved

mmodzelewski: thanks :)

Issue #2924, opened 2026-03-12
Memory leak (?) in iggy server and consumer offsets skip


lukaszzborek:
### Bug description

I've observed the iggy server's memory usage constantly growing. 
It's not a huge increase, only 10-15 MB for every 2-3 million messages.

I checked it on multiple last edge builds

When the Iggy server hits its memory limit, it uses almost 100% of all CPUs. Usually, I can't log in to the server to restart the service, so I have to do it via the Proxmox UI.

After restarting, memory usage returns to normal. However, two things started happening:
1. cleaners did not respect the message expiration. Segments were deleted just after the segment rotation (the first run of the next cleaner).
2. consumers (pull next offset) are beginning to lose their offsets. Randomly skip random numbers of messages

I have 10 message expiry, with 128mb segment size
Offset lose is only on topic which was on during cpu drain. On new topics created after restart, everythink looks good.

Also i forgot to check memory usaged in system. I only see some chart in Proxmox UI and logs from iggy stats.

I can send local_data, which is saved after a server restart or CPU drain. I cannot attach it here because it's over 100 MB. You can write messages to the topic, and then try to pull the next one. It should skip messages sometimes.
Feel free to ask if you have any questions, or want additional data or something 😅

### Deployment

Pre-built binary / release artifact

### Versions

0.7.2-edge.1, hash: c15100698874c38f88f36a4bb62ba205c3e53db2, built at: 2026-03-10T21:13:24.620646481Z

### Hardware / environment

proxmox, lxc, linux

### Sample code

nuget verions: 0.7.0

```csharp
var iggy = IggyClientFactory.CreateClient(new IggyClientConfigurator()
{
    BaseAddress = "localhost:8090",
    Protocol = Protocol.Tcp,
    AutoLoginSettings = new AutoLoginSettings()
    {
        Enabled = true,
        Username = "otherapp",
        Password = "xxxxxxx"
    }
});

await iggy.ConnectAsync();

var stream = Identifier.String("main");
var topic = Identifier.String("orderbook-topic");
var cg = "cg-orderbook";

await iggy.JoinConsumerGroupAsync(stream, topic, Identifier.String(cg));
var offset = await iggy.GetOffsetAsync(Consumer.Group(cg), stream, topic, 0);
ulong lastOffset = 0L;

var count = 10u;
while (true)
{
    var messages = await iggy.PollMessagesAsync(stream, topic, 0, Consumer.Group(cg), PollingStrategy.Next(), count, true);
[...]

Issue #2715, opened 2026-02-10
Non-deterministic consumer offset jump to "latest" on large streams (~50M records)


jpatidar30:
I am encountering a critical issue where a consumer unexpectedly jumps from an intermediate offset directly to the end of the stream (the latest offset). This behavior is non-deterministic and occurs after ingesting a large volume of data (~50M records).

**Behavior Observed**
The Jump: While consuming, the fetch logic might work perfectly up to a certain point (e.g., offset 5,000,000). On the next poll, instead of receiving 5,000,001, the consumer is suddenly positioned at the end of the stream (50,000,000), effectively skipping 45M records.

Non-Determinism: The "jump point" is not fixed.

A new consumer group/ID might jump at a completely different offset (e.g., at 12M instead of 5M).

Retrying the same consumer sometimes works after 2 retries, and sometimes requires 10+ retries to successfully pass the "glitchy" offset.

Data Integrity: The data does exist in the stream, as repeated retries eventually allow the consumer to read through the records that were previously skipped.

**Environment Information**

**Iggy Version**: Docker image with latest tag

**Storage**:  EBS GP3 (AWS)

**SDK Used**: Rust

**Number of partitions**: 1

spetz:
Hey,
Could you kindly check `edge` image instead, along with the https://crates.io/crates/iggy/0.8.3-edge.1 SDK edge version? We've had tons of improvements since the `latest` release, but also a few breaking changes in transport/storage, we plan to discuss about next release soon, but as of now, it'd best to try out using the edge versions.

hubcio:
@AmritUdasi you can post your opinion here: https://github.com/apache/iggy/discussions/2719

this is older discussion for 0.6.0 release: https://github.com/apache/iggy/discussions/2337

jpatidar30:
Hey @hubcio, We'll be testing it this week

hubcio:
closing this, should be fixed in server `0.8.0` (`latest` tag on dockerhub).

PR #2958, opened 2026-03-17
fix(server): prevent consumer offset skip during concurrent produce+consume

hubcio:
After server restart, MemoryMessageJournal was created via
Default with base_offset=0. The three-tier message routing
used this value as in_memory_floor, which prevented all disk
reads when the journal had data. Consumers using
PollingStrategy::Next with auto_commit permanently skipped
the last few disk messages when new messages arrived in the
journal concurrently.

Root cause: journal.init() existed but was never called in
production code - only in tests. After restart with N
messages on disk, the first journal append set
current_offset = 0 + batch_count - 1 instead of
N + batch_count - 1. slice_by_offset then silently returned
messages from the wrong range (clamping to index 0 when
start_offset < first_offset).

The fix has six layers:

1. Initialize journal base_offset at both bootstrap paths
   (shard/mod.rs and the lazy init_partition_inner path in
   shard/system/partitions.rs). Fix current_offset
   computation to use max end_offset across segments with
   data, not just the active segment. Fix
   should_increment_offset to check any segment has data,
   not just current_offset > 0.

2. Self-heal base_offset in journal.append() on first
   append when messages_count==0, with debug_assert
   validation.

3. Change slice_by_offset to return None when
   start_offset < first_offset instead of silently
   returning data from a higher offset range.

4. Remove Default from MemoryMessageJournal so the bug
   class is structurally impossible. Add explicit
   empty()/at_offset() constructors. Add typed query
   methods (first_offset, last_offset, first_timestamp,
   last_timestamp) on the Journal trait. Delete dead
   Clone impl.

5. Clamp consumer offsets that are ahead of partition
   offset after crash (OOM, SIGKILL) at both bootstrap
   paths. Prevents permanent empty polls when
   auto_commit persisted an offset beyond what was
   flushed to disk.

6. Add consumer offset barrier to time-based expiry
   (delete_expired_segments_for_partition), matching the
   existing size-based barrier. Log a warning when the
   barrier blocks segment deletion, identifying the
   blocking consumer kind, ID, and offset. Fix
   is_expired to treat end_timestamp=0 as non-expired
   (prevents instant deletion of segments with empty
   indexes during bootstrap).

Reference issues: #2715 and  #2924.

### DIFF
--- a/core/server/src/shard/system/segments.rs
+++ b/core/server/src/shard/system/segments.rs
@@ -19,7 +19,7 @@ use crate::configs::cache_indexes::CacheIndexesConfig;
 use crate::shard::IggyShard;
 use crate::streaming::segments::Segment;
 use iggy_common::sharding::IggyNamespace;
-use iggy_common::{IggyError, IggyExpiry, IggyTimestamp, MaxTopicSize};
+use iggy_common::{ConsumerKind, IggyError, IggyExpiry, IggyTimestamp, MaxTopicSize};
 
 impl IggyShard {
     /// Performs all cleanup for a topic's partitions: time-based expiry then size-based trimming.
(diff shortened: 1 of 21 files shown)


core/server/src/shard/system/segments.rs
```diff
@@ -172,6 +182,16 @@ impl IggyShard {
...
+            // Same barrier as delete_expired_segments_for_partition: do not
+            // delete segments containing unconsumed messages.
+            let min_committed = Self::min_committed_offset(
+                &partition.consumer_offsets,
+                &partition.consumer_group_offsets,
+            );
+            if min_committed.is_some_and(|barrier| first.end_offset > barrier) {
+                continue;
+            }
+
```
numinnex, line 212: Do we really want to prevent deleting expired segments that have registered consumers via committed offsets ? I am not sure whether it is an correct behavior + it potentially could lead in some pathological cases, where we have a zombie consumer (who doesn't consume anything for a long time), that prevents the reclamation mechanism from deleting expired segments.

core/server/src/streaming/partitions/journal.rs
```diff
@@ -19,30 +19,43 @@ use crate::streaming::segments::{IggyMessagesBatchMut, IggyMessagesBatchSet};
...
-// TODO: Will have to revisit this Journal abstraction....
-// I don't like that it has to leak impl detail via the `Inner` struct in order to be functional.
-
 #[derive(Default, Debug)]
 pub struct Inner {
-    pub base_offset: u64,
+    /// Offset where the next append will land. After commit(), set to
+    /// current_offset + 1. Used in `append()`: `current_offset = next_offset +
+    /// messages_count - 1`.
+    pub next_offset: u64,
```
numinnex: Why we renamed this from `base_offset` to `next_offset` ? I think the name base_offset was better, as it represents the beginning offset of the journal log.

core/server/src/shard/system/segments.rs
hubcio: i can do this in follow-up. do you have any suggestions? i see a couple of options:
- max_topic_size overrides this barrier
- only barrier against active consumers (lets say those that polled within X minutes, we can make it configurable)
- add logs wrn/err when such segment is being deleted

core/server/src/streaming/partitions/journal.rs
hubcio: i reverted back to the original name.

core/server/src/shard/system/segments.rs
numinnex: I think the option with logs is probably the best way to approach this problem.

hubcio: I added following prints when consumer/consumer group lags. deletion will be still blocked.
image

PR #2964, opened 2026-03-18
feat(go): introduce context.Context to client methods

chengxilo:

N/A

## Rationale

Standardizing the Go SDK to support context.Context across all client methods. This allows users to handle timeouts, cancellations
## What changed?

**HARD BEHAVIOR CHANGE**
-  Auto_Login/Reconnect Logic would be implemented in another PR, currently, the connection would be dead after timeout/io error as reconnection is not implemented in Go SDK yet**
- Almost all method of client would need an extra `ctx context.Context` parameter
- `connect` is now a public method, after `NewIggyClient`, user have to call `Connect()` to actually establish the connection.

The iggcon.Client interface and its TCP implementation have been updated to include context.Context as the first parameter for all API methods. 
Correspondingly, all internal command-handling logic now respects context deadlines and cancellations.

## Local Execution

- Passed 
- Pre-commit hooks ran  

## AI Usage

1. Claude Sonnet 4.6
2. Add context parameters, initial implementation of test (modifed manually)
3. I read the code and test
4. yes


foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,18 +258,32 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+	}
+
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
+	if deadline, ok := ctx.Deadline(); ok {
+		if err := c.conn.SetDeadline(deadline); err != nil {
+			return nil, err
+		}
+		defer func() { _ = c.conn.SetDeadline(time.Time{}) }()
```
hubcio: if a deadline fires mid-read, the deferred `SetDeadline(time.Time{})` clears the deadline, but the connection is now in an unknown protocol state - partial write sent, partial response read. the next `sendAndFetchResponse` call on this connection will read leftover bytes from the timed-out response as its response header, producing silent data corruption or protocol errors.

there's no connection-invalidation logic after a timeout error. after any I/O error (including deadline-induced timeout), the connection should be marked as broken so subsequent calls trigger a reconnect instead of reusing a poisoned connection.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,18 +258,32 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+		return nil, errors.New("nil context")
+	}
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
+	if deadline, ok := ctx.Deadline(); ok {
```
hubcio: for contexts created with `context.WithCancel` (no deadline), `ctx.Deadline()` returns `ok == false`, so `SetDeadline` is never called. this means cancellation during a blocking `read`/`write` is silently ignored - the goroutine blocks until the server responds or TCP times out.

the `ctx.Err()` check at line 273 is pre-flight only and doesn't help once I/O starts. a common fix is to spawn a goroutine watching `ctx.Done()` and calling `c.conn.SetDeadline(time.Now())` to unblock I/O, similar to how `net/http` handles this. alternatively, document that only deadline-based contexts are respected during I/O.

foreign/go/client/tcp/tcp_utilities.go
```diff
@@ -18,13 +18,15 @@
...
+	"context"
+
 	binaryserialization "github.com/apache/iggy/foreign/go/binary_serialization"
 	iggcon "github.com/apache/iggy/foreign/go/contracts"
 	"github.com/apache/iggy/foreign/go/internal/command"
 )
 
-func (c *IggyTcpClient) GetStats() (*iggcon.Stats, error) {
-	buffer, err := c.do(&command.GetStats{})
+func (c *IggyTcpClient) GetStats(ctx context.Context) (*iggcon.Stats, error) {
```
hubcio, line 27: the rust side now serializes `threads_count`, `free_disk_space`, and `total_disk_space` on the wire (in `core/server/src/binary/mapper.rs`), but the go `Stats` struct (`foreign/go/contracts/stats.go`) and its deserializer (`foreign/go/binary_serialization/stats_serializer.go`) don't read these fields - deserialization stops at `KernelVersion`. go consumers of `GetStats()` get silently incomplete responses. since this PR adds context to `GetStats`, the go SDK should also be updated to consume the new stats fields.

see https://github.com/apache/iggy/pull/2917, this can be done in different PR :)

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,18 +258,32 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
-	return c.sendAndFetchResponse(data, cmd.Code())
+	return c.sendAndFetchResponse(ctx, data, cmd.Code())
 }
 
-func (c *IggyTcpClient) sendAndFetchResponse(message []byte, command command.Code) ([]byte, error) {
+func (c *IggyTcpClient) sendAndFetchResponse(ctx context.Context, message []byte, command command.Code) ([]byte, error) {
+	if ctx == nil {
+		return nil, errors.New("nil context")
+	}
+	if err := ctx.Err(); err != nil {
```
hubcio, line 266: this check only fires before `c.mtx.Lock()`. if another goroutine holds the lock doing slow I/O, this goroutine blocks on `Lock()` indefinitely even if its context is already cancelled or deadline has passed. consider re-checking `ctx.Err()` immediately after acquiring the lock.

foreign/go/contracts/client.go
```diff
@@ -17,6 +17,8 @@
 
 package iggcon
 
+import "context"
+
 type Client interface {
 	// Close closes the client and releases all the resources.
 	Close() error
```
hubcio, line 27: every other method in the interface now takes `ctx`, but `Close()` does not - and it performs network I/O (`conn.Close()`). callers can't enforce a timeout on shutdown. worth either adding `Close(ctx context.Context) error` or documenting why this is intentionally excluded.

foreign/go/client/iggy_client.go
```diff
@@ -88,7 +88,7 @@ func NewIggyClient(options ...Option) (iggcon.Client, error) {
 				case <-ctx.Done():
 					return
 				case <-ticker.C:
-					if err := cli.Ping(); err != nil {
+					if err := cli.Ping(context.Background()); err != nil {
```
hubcio: two things here:

1. `context.Background()` means an in-flight `Ping()` won't be cancelled when `Close()` calls `ic.cancel()`. if the connection is hung, this goroutine blocks indefinitely. consider `context.WithTimeout(ctx, heartbeatInterval/2)` instead.

2. `Close()` calls `cancel()` then immediately `Client.Close()`, but there's no `sync.WaitGroup` to wait for this goroutine to exit. the goroutine may be mid-`Ping()` when the connection is closed underneath it, producing a spurious error log.

foreign/go/contracts/client.go
```diff
@@ -26,39 +28,40 @@ type Client interface {
...
 	// Authentication is required, and the permission to manage the streams.
-	CreateStream(name string) (*StreamDetails, error)
+	CreateStream(ctx context.Context, name string) (*StreamDetails, error)
 
 	// UpdateStream update a stream by unique ID or name.
 	// Authentication is required, and the permission to manage the streams.
-	UpdateStream(streamId Identifier, name string) error
+	UpdateStream(ctx context.Context, streamId Identifier, name string) error
 
 	// DeleteStream delete a topic by unique ID or name.
```
hubcio: copy-paste error: `DeleteStream` doc says "delete a topic by unique ID or name" and "permission to manage the topics". should say "stream".

bdd/go/tests/leader_redirection.go
```diff
@@ -122,7 +122,7 @@ func serverTypeFromPort(port uint16) string {
 }
 
 func verifyLeaderInMetadata(client iggcon.Client) (*iggcon.ClusterNode, error) {
-	metadata, err := client.GetClusterMetadata()
+	metadata, err := client.GetClusterMetadata(context.Background())
```
hubcio: this and line 155 hardcode `context.Background()`, while other functions in the same file (lines 324, 342, 437) correctly use the step's `ctx` parameter. should be consistent - use `ctx` everywhere or pass it through from the caller.

hubcio (changes requested):

bdd/go/tests/leader_redirection.go
chengxilo: fixed.

foreign/go/contracts/client.go
chengxilo: Will solve it in #2972

foreign/go/client/iggy_client.go
chengxilo: solved.

foreign/go/client/tcp/tcp_core.go
chengxilo: solved.

foreign/go/contracts/client.go
chengxilo: I think conn.Close() just closes the socket and isn't a blocking network I/O operation. And `Close() error` is a standard Go idiom if I am not mistaken(it implements `io.Closer`), I think it's okay to leave it as-is without additional documentation.

foreign/go/client/tcp/tcp_utilities.go
chengxilo: Address in #3032

foreign/go/client/tcp/tcp_core.go
chengxilo: resolved.

foreign/go/client/tcp/tcp_core.go
chengxilo: resolved

ryankert01 (approved):
Lg, like the idea.

hubcio:
looks like CI is failing.

/author

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,49 +258,102 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+		return nil, err
+	}
+
+	// deadlineMu makes sure that the deadline won't be set to now by the
+	// AfterFunc callback right after we call SetDeadline(time.Time{}) to
+	// clear the deadline after the operation is done.
+	var deadlineMu sync.Mutex
+
+	stop := context.AfterFunc(ctx, func() {
```
hubcio, line 286: the `context.AfterFunc` callback reads `c.conn` without holding `c.mtx`, and `stop()` does not wait for an already-started callback - it can return false while the callback goroutine is still running. once `sendAndFetchResponse` returns and the mutex is released, `connect()` can reassign `c.conn` at line 459, so the callback's read races that write. worse, the callback then calls `SetDeadline(time.Now())` on whatever `c.conn` points to, which can be a fresh healthy connection, poisoning it into an immediate timeout. fix: snapshot `conn := c.conn` under the mutex before registering the callback and have the callback use the snapshot. a `ctx.Done() == nil` fast-path also closes the window for `context.Background()` callers.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,49 +258,102 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+	}
+
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	// deadlineMu makes sure that the deadline won't be set to now by the
```
hubcio: the comment claims `deadlineMu` stops the callback from setting the deadline to now right after `clearDeadline` clears it, but a mutex only gives mutual exclusion, not ordering. if `clearDeadline` acquires `deadlineMu` first it sets the deadline to zero, then the callback runs and sets it to `time.Now()`, leaving the connection with a past deadline. needs a `cleared` flag set under `deadlineMu` that the callback checks, not just the mutex.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,49 +258,102 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
 		return nil, err
 	}
-	return c.sendAndFetchResponse(data, cmd.Code())
+	return c.sendAndFetchResponse(ctx, data, cmd.Code())
 }
 
-func (c *IggyTcpClient) sendAndFetchResponse(message []byte, command command.Code) ([]byte, error) {
+func (c *IggyTcpClient) sendAndFetchResponse(ctx context.Context, message []byte, command command.Code) ([]byte, error) {
+	if ctx == nil {
+		return nil, errors.New("nil context")
```
hubcio: `errors.New("nil context")` is an untyped error - callers can't `errors.Is` it. every other error path here returns typed values (`context.Canceled`, `context.DeadlineExceeded`, `ierror.*`). make this a sentinel like `ierror.ErrNilContext` for consistency at the SDK boundary.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,49 +258,102 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+		c.invalidateConnLocked()
+		return nil, ioError(err)
 	}
 
+	clearDeadline()
 	return buffer, nil
 }
 
+// invalidateConnLocked closes the connection and marks it as disconnected
+func (c *IggyTcpClient) invalidateConnLocked() {
```
hubcio, line 352: `invalidateConnLocked` closes the conn and sets `StateDisconnected` on every i/o error, but nothing on the operation path ever reconnects - `connect()` is only called from `NewIggyTcpClient` and the login-after-redirect path, and the heartbeat goroutine just logs ping failures. so after the first transient i/o error the client is permanently dead: every later call writes to the closed conn and fails forever, and the `reconnection` config (maxRetries/interval) becomes dead code on this path. pre-PR an i/o error left the conn open and state unchanged. either wire up a reconnect trigger here or document this as a hard behavior change.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,49 +258,102 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
 	defer c.mtx.Unlock()
 
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	// deadlineMu makes sure that the deadline won't be set to now by the
+	// AfterFunc callback right after we call SetDeadline(time.Time{}) to
+	// clear the deadline after the operation is done.
+	var deadlineMu sync.Mutex
```
hubcio, line 283: `var deadlineMu sync.Mutex` plus the `context.AfterFunc` registration and the `clearDeadline`/`ioError` closures are allocated on every `sendAndFetchResponse` call - including the hot `SendMessages`/`PollMessages` path and `context.Background()` callers whose `ctx.Done()` is nil and can never fire. a `ctx.Done() == nil` fast-path that skips the deadline machinery entirely avoids the waste, and as noted on the `AfterFunc` line also removes the `c.conn` race for non-cancellable contexts.

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -37,53 +38,54 @@ func (c *IggyTcpClient) LoginUser(username string, password string) (*iggcon.Ide
...
+	shouldRedirect, err := c.HandleLeaderRedirection(ctx)
 	if err != nil {
 		return nil, err
 	}
 	if shouldRedirect {
 		if err = c.connect(); err != nil {
 			return nil, err
 		}
-		return c.LoginUser(username, password)
+		return c.LoginUser(ctx, username, password)
```
hubcio, line 49: also at line 71 in `LoginWithPersonalAccessToken`. the recursion re-passes the same `ctx`, and the first `c.do`'s `AfterFunc` callback can still be live when `connect()` (line 46 / 68) reassigns `c.conn` - same `c.conn` race flagged in `tcp_core.go`, reachable single-threaded here. fixing the callback to use a conn snapshot covers this too.

foreign/go/internal/util/leader_aware.go
```diff
@@ -29,10 +30,10 @@ import (
 
 // CheckAndRedirectToLeader queries the client for cluster metadata and returns
 // an address to redirect to (empty string means no redirection needed).
-func CheckAndRedirectToLeader(c iggcon.Client, currentAddress string, transport iggcon.Protocol) (string, error) {
+func CheckAndRedirectToLeader(ctx context.Context, c iggcon.Client, currentAddress string, transport iggcon.Protocol) (string, error) {
 	log.Println("Checking cluster metadata for leader detection")
 
-	meta, err := c.GetClusterMetadata()
+	meta, err := c.GetClusterMetadata(ctx)
```
hubcio, line 37: if `ctx` is cancelled during this `GetClusterMetadata` call, `sendAndFetchResponse` closes the connection via `invalidateConnLocked` and returns `ctx.Err()` - but the `if err != nil` block just below logs it and returns `("", nil)`, swallowing the cancellation. `HandleLeaderRedirection` then returns `(false, nil)` and `LoginUser` returns `identity, nil`: a successful login on a dead connection. the swallow is intentional for non-clustered servers, so don't propagate all errors blindly - check `ctx.Err()` specifically and return it before the swallow.

foreign/go/client/tcp/tcp_core_test.go
```diff
@@ -0,0 +1,237 @@
...
+	_, err := c.sendAndFetchResponse(ctx, []byte{}, command.Code(0))
+	if err == nil {
+		t.Fatal("expected timeout error, got nil")
+	}
+
+	if !errors.Is(err, context.DeadlineExceeded) {
+		t.Errorf("got %v, want context.DeadlineExceeded", err)
+	}
+	// After a timeout, the connection should be invalidated.
+	if c.state != iggcon.StateDisconnected {
```
hubcio, line 126: this asserts `c.state == StateDisconnected` after a timeout, which codifies the sticky-failure behavior (closed conn, no reconnect path) as expected - if that gets fixed, this test changes. separately, every test here is single-operation: none exercise `connect()` reassigning `c.conn` while an `AfterFunc` callback from a prior op is still live, which is exactly the race this PR introduces. `go test -race` stays green only because of that gap. worth adding a concurrent-reassignment test.

hubcio (changes requested):
the PR doesn't compile. `go build ./...` fails:

```
client/tcp/tcp_segment_management.go:31:17: not enough arguments in call to c.do
```

`DeleteSegments` wasn't migrated - the impl in `tcp_segment_management.go` still calls `c.do(&command.DeleteSegments{...})` without the new `ctx` arg, and the interface method in `contracts/client.go` still lacks `ctx`. that file isn't part of this PR's diff so it can't be commented on inline, but `ctx` needs threading through both the interface signature and the impl, same as every other client method. `go vet` flags the same.

the rest of the findings are inline on the Files changed tab.

chengxilo:
Ok it seems that something happend after merging(). Will start working on this on April 20

foreign/go/client/tcp/tcp_core.go
chengxilo: Yeah I almost forgot about this . I was planning to solve this in another PR. Maybe I can create a PR to implement the reconnect logic first.

foreign/go/client/tcp/tcp_core.go
chengxilo: Done :)

foreign/go/client/tcp/tcp_core.go
chengxilo: Done

foreign/go/client/tcp/tcp_core.go
chengxilo: fast-path provided. Solved.

foreign/go/client/tcp/tcp_core.go
chengxilo: Addressed.

foreign/go/client/tcp/tcp_session_management.go
chengxilo: Done.

foreign/go/client/tcp/tcp_core.go
chengxilo: I tried to implement the logic of reconnection/auto_login on the master branch, it cannot be merged with this easily... So probably I will implmenet it after this branch is merged :) Will tag this as a hard behavior change.

foreign/go/internal/util/leader_aware.go
chengxilo: Done.

foreign/go/client/tcp/tcp_core_test.go
chengxilo: It's difficult to reproduce this in unit test. But I believe the change in code solved the sticky-failure behavior

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -37,53 +38,54 @@ func (c *IggyTcpClient) LoginUser(username string, password string) (*iggcon.Ide
 	}
 
 	identity := binaryserialization.DeserializeLogInResponse(buffer)
-	shouldRedirect, err := c.HandleLeaderRedirection()
+	shouldRedirect, err := c.HandleLeaderRedirection(ctx)
 	if err != nil {
 		return nil, err
 	}
 	if shouldRedirect {
 		if err = c.connect(); err != nil {
```
hubcio: also at line 68.

`c.connect()` calls `net.Dial` which has no context support - if the caller's `ctx` is cancelled mid-redirect, the dial blocks until the OS TCP timeout (~75s). user can't abort it. switch `connect()` to take a `ctx` and use `net.Dialer{}.DialContext(ctx, ...)`.

separately, no `ctx.Err()` recheck between `HandleLeaderRedirection` returning true and calling `connect()`. if ctx cancelled in that window we still tear down + redial pointlessly. cheap guard.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,29 +257,79 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+		defer deadlineMu.Unlock()
+		if !cleared {
+			_ = conn.SetDeadline(time.Now())
+		}
+	})
+	defer stop()
+
+	result, err := c.sendLocked(message, command)
+
+	stop()
```
hubcio: explicit `stop()` at L302 plus `defer stop()` at L298. the `cleared` flag + `deadlineMu` already serialize the AfterFunc callback vs the deadline reset, so the explicit `stop()` is a micro-opt to skip the callback when it'd be a no-op anyway. either drop the explicit call (defer alone is correct) or drop the defer (explicit covers happy path, defer covers panic). nit.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -293,12 +344,19 @@ func (c *IggyTcpClient) sendAndFetchResponse(message []byte, command command.Cod
...
 	if err != nil {
+		c.invalidateConnLocked()
 		return nil, err
 	}
 
 	return buffer, nil
 }
 
+// invalidateConnLocked closes the connection and marks it as disconnected
+func (c *IggyTcpClient) invalidateConnLocked() {
```
hubcio, line 352: `c.conn.Close()` with no nil-guard. no current code path reaches here with `c.conn == nil` (constructor connects before returning, and this is only called after a successful send/read), so it's defensive only - not a live bug. worth adding before future paths (reconnect, partial init) could change that:

```go
if c.conn != nil {
    _ = c.conn.Close()
}
```

nit.

foreign/go/internal/util/leader_aware.go
```diff
@@ -29,11 +30,14 @@ import (
...
 	if err != nil {
+		if ctx.Err() != nil {
```
hubcio: `ctx.Err() != nil` after a network error is racy attribution. if `GetClusterMetadata` returned a genuine network error and ctx then got cancelled between that return and this check, you surface `ctx.Err()` instead of the real cause. check the error directly:

```go
if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
    return "", err
}
```

foreign/go/client/iggy_client.go
```diff
@@ -79,29 +81,35 @@ func NewIggyClient(options ...Option) (iggcon.Client, error) {
 	}
 	ctx, cancel := context.WithCancel(context.Background())
```
hubcio: `context.WithCancel` allocated unconditionally even when `heartbeatInterval <= 0`. the resulting cancel + wg are dead weight - no goroutine to cancel, nothing to wait on. move ctx/cancel creation inside the `if heartbeatInterval > 0` block.

foreign/go/client/iggy_client.go
```diff
@@ -79,29 +81,35 @@ func NewIggyClient(options ...Option) (iggcon.Client, error) {
...
 			ticker := time.NewTicker(heartbeatInterval)
 			defer ticker.Stop()
 			for {
 				select {
 				case <-ctx.Done():
 					return
 				case <-ticker.C:
-					if err := cli.Ping(); err != nil {
+					pingCtx, pingCancel := context.WithTimeout(ctx, heartbeatInterval/2)
+					if err := cli.Ping(pingCtx); err != nil {
```
hubcio: `heartbeatInterval/2` as ping timeout is unguarded. no `WithHeartbeatInterval` setter exists today so only the 5s default is reachable - this is forward-looking. the moment the option becomes user-configurable a small/zero value yields a zero ping timeout and ping always fails. either gate with a minimum:

```go
timeout := heartbeatInterval / 2
if timeout < 100*time.Millisecond {
    timeout = 100 * time.Millisecond
}
```

or document the constraint when the setter lands.

foreign/go/client/tcp/tcp_core_test.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	if !errors.Is(err, context.Canceled) {
+		t.Errorf("got %v, want context.Canceled", err)
+	}
+	// Connection should be invalidated after the I/O error.
+	if c.state != iggcon.StateDisconnected {
+		t.Errorf("expected state %v, got %v", iggcon.StateDisconnected, c.state)
+	}
+}
```
hubcio, line 157: no test exercises the race the local-capture `conn := c.conn` at L287 defends against. `c.mtx` is held across `sendAndFetchResponse` so a concurrent caller can't reassign `c.conn` mid-call - the real concern is the `context.AfterFunc` callback, which runs in its own goroutine **without** holding `c.mtx` and dereferences the captured `conn`. a later `connect()` (redirect path, or future reconnect) can swap `c.conn` while the old callback is still scheduled, and the local capture is what keeps `SetDeadline` from landing on the new conn.

add a test that schedules an AfterFunc, swaps `c.conn` via a fresh `connect()`, then asserts `SetDeadline` only ever touched the original conn. without it the regression is invisible the day someone removes the local-capture line.

foreign/go/client/tcp/tcp_core_test.go
```diff
@@ -0,0 +1,236 @@
...
+	_, err := c.sendAndFetchResponse(ctx, []byte{}, command.Code(0))
+	if err == nil {
+		t.Fatal("expected timeout error, got nil")
+	}
+
+	if !errors.Is(err, context.DeadlineExceeded) {
+		t.Errorf("got %v, want context.DeadlineExceeded", err)
+	}
+	// After a timeout, the connection should be invalidated.
+	if c.state != iggcon.StateDisconnected {
```
hubcio, line 126: also at line 153.

asserting `c.state == StateDisconnected` after timeout/cancel codifies the sticky-failure behavior (no reconnect path) as expected. when the planned reconnect work lands these assertions need to flip. add a `// TODO: revisit after reconnect implementation` so the contract is explicit and the test gets updated, not silently broken.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -256,29 +257,79 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+	}
+
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
 
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
+	if ctx.Done() == nil {
```
hubcio, line 278: fast-path returns from `sendLocked` while holding `c.mtx` (released by the outer `defer c.mtx.Unlock()`). intentional - avoids the `AfterFunc` setup for non-cancellable contexts - but a future reader will look at this `return` and "fix" it by releasing the lock first, breaking serialization. one-line comment would save a debugging session:

```go
// fast path: non-cancellable ctx, skip the deadline mechanism.
// mtx is released by the outer defer.
```

hubcio (changes requested):
overall it looks good, only critical issue I found is in `tcp_session_management.go`, rest are just nits/risks.

foreign/go/internal/util/leader_aware.go
chengxilo: Fixed :)

foreign/go/client/tcp/tcp_core.go
chengxilo: Done. The `_______Locked` already tells developer that the mtx is hold by outer function so it's fine. I add a comment to document that this is for the fast path to improve readability anyway,.

foreign/go/client/tcp/tcp_core.go
chengxilo: Done

foreign/go/client/tcp/tcp_core.go
chengxilo: Yeah, it provide a minimum improvement(I think) but make the code confusing. Removed.

foreign/go/client/tcp/tcp_core_test.go
chengxilo: Done.

foreign/go/client/tcp/tcp_session_management.go
chengxilo: Done. To make conection support context.Context, I made it a public method for client, so user can provide context.Context to the `Connect()` directly.

foreign/go/client/iggy_client.go
chengxilo: Solved.

foreign/go/client/iggy_client.go
chengxilo: I think it's okay to leave it as is, because if the heartbeat interval is set to a small or zero value, the problem is the configuration itself.
Also, I prefer not to adjust the value silently. I think it would be better to discuss and decide whether we should add parameter validation for the heartbeatInterval configuration, returning an error if the value is too small (e.g., < 200ms).

foreign/go/client/tcp/tcp_core_test.go
chengxilo: It's really hard to write a test that can actually capture this race condition. I write a comment to explain the `conn := c.conn` instead.

foreign/go/client/iggy_client.go
hubcio: makes sense. lets leave it as is.

hubcio (approved):

mmodzelewski (approved):

PR #2967, opened 2026-03-18
feat(rust): add sans-IO frame codec and command dispatch table

hubcio:
The binary_protocol crate had wire types but no framing layer.
Server and SDK hand-rolled frame parsing independently.

Add RequestFrame/ResponseFrame sans-IO codec (framing.rs) with
pre-reserved buffers, checked encoded_size() returning Option,
and NonZeroU32 status in encode_error() for type-level safety.

Add CommandMeta dispatch table (dispatch.rs) as single source
of truth for all 47 command codes, names, and VSR operations.
Operation::from_command_code/to_command_code now delegate to it.

Replace frame.rs with framing.rs. Remove glob re-exports from
crate root. Change WireError::Validation to Cow<'static, str>.

Add remaining system responses: ClusterMetadataResponse (cluster
name + nodes with transport endpoints), GetSnapshotResponse (raw
ZIP bytes), GetMeResponse (alias for ClientDetailsResponse).
Promote all consensus types to crate root re-exports. Add
read_u16_le codec helper.

This code is not yet used.

### DIFF
--- a/core/binary_protocol/src/consensus/operation.rs
+++ b/core/binary_protocol/src/consensus/operation.rs
@@ -95,64 +95,48 @@ impl Operation {
     }
 
     /// Bidirectional mapping: `Operation` -> client command code.
+    ///
+    /// Delegates to the dispatch table as the single source of truth.
     #[must_use]
     pub const fn to_command_code(&self) -> Option<u32> {
-        use crate::codes;
         match self {
             Self::Reserved => None,
-            Self::CreateStream => Some(codes::CREATE_STREAM_CODE),
-            Self::UpdateStream => Some(codes::UPDATE_STREAM_CODE),
-            Self::DeleteStream => Some(codes::DELETE_STREAM_CODE),
-            Self::PurgeStream => Some(codes::PURGE_STREAM_CODE),
-            Self::CreateTopic => Some(codes::CREATE_TOPIC_CODE),
-            Self::UpdateTopic => Some(codes::UPDATE_TOPIC_CODE),
-            Self::DeleteTopic => Some(codes::DELETE_TOPIC_CODE),
-            Self::PurgeTopic => Some(codes::PURGE_TOPIC_CODE),
-            Self::CreatePartitions => Some(codes::CREATE_PARTITIONS_CODE),
-            Self::DeletePartitions => Some(codes::DELETE_PARTITIONS_CODE),
-            Self::DeleteSegments => Some(codes::DELETE_SEGMENTS_CODE),
-            Self::CreateConsumerGroup => Some(codes::CREATE_CONSUMER_GROUP_CODE),
-            Self::DeleteConsumerGroup => Some(codes::DELETE_CONSUMER_GROUP_CODE),
-            Self::CreateUser => Some(codes::CREATE_USER_CODE),
-            Self::UpdateUser => Some(codes::UPDATE_USER_CODE),
-            Self::DeleteUser => Some(codes::DELETE_USER_CODE),
-            Self::ChangePassword => Some(codes::CHANGE_PASSWORD_CODE),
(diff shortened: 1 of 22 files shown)


core/binary_protocol/src/consensus/operation.rs
```diff
@@ -95,64 +95,27 @@ impl Operation {
...
-            Self::CreateUser => Some(codes::CREATE_USER_CODE),
-            Self::UpdateUser => Some(codes::UPDATE_USER_CODE),
-            Self::DeleteUser => Some(codes::DELETE_USER_CODE),
-            Self::ChangePassword => Some(codes::CHANGE_PASSWORD_CODE),
-            Self::UpdatePermissions => Some(codes::UPDATE_PERMISSIONS_CODE),
-            Self::CreatePersonalAccessToken => Some(codes::CREATE_PERSONAL_ACCESS_TOKEN_CODE),
-            Self::DeletePersonalAccessToken => Some(codes::DELETE_PERSONAL_ACCESS_TOKEN_CODE),
-            Self::SendMessages => Some(codes::SEND_MESSAGES_CODE),
-            Self::StoreConsumerOffset => Some(codes::STORE_CONSUMER_OFFSET_CODE),
+            _ => match crate::dispatch::lookup_by_operation(*self) {
```
atharvalade: The `_` wildcard here means that if someone adds a new Operation variant but forgets to add it to COMMAND_TABLE, this silently returns None instead of giving a compiler error. The old exhaustive match caught that at compile time. how about listing all variants explicitly (delegating to the dispatch table is fine) so rustc keeps enforcing coverage?

core/binary_protocol/src/responses/system/get_cluster_metadata.rs
```diff
@@ -0,0 +1,230 @@
...
+impl WireDecode for ClusterMetadataResponse {
+    fn decode(buf: &[u8]) -> Result<(Self, usize), WireError> {
+        let name_len = read_u32_le(buf, 0)? as usize;
+        let name = read_str(buf, 4, name_len)?;
+        let mut pos = 4 + name_len;
+
+        let nodes_count = read_u32_le(buf, pos)? as usize;
+        pos += 4;
+
+        let mut nodes = Vec::with_capacity(nodes_count);
```
atharvalade: `Vec::with_capacity(nodes_count)` allocates eagerly before any node is decoded, so a crafted nodes_count of `u32::MAX` tries to reserve ~200 GB and OOMs. I think it'll be a good idea to cap the pre-allocation to something like `nodes_count.min(remaining_buf_len / 18)` where 18 is the minimum encoded node size.

core/binary_protocol/src/consensus/operation.rs
```diff
@@ -95,64 +95,48 @@ impl Operation {
...
-            codes::UPDATE_USER_CODE => Some(Self::UpdateUser),
-            codes::DELETE_USER_CODE => Some(Self::DeleteUser),
-            codes::CHANGE_PASSWORD_CODE => Some(Self::ChangePassword),
-            codes::UPDATE_PERMISSIONS_CODE => Some(Self::UpdatePermissions),
-            codes::CREATE_PERSONAL_ACCESS_TOKEN_CODE => Some(Self::CreatePersonalAccessToken),
-            codes::DELETE_PERSONAL_ACCESS_TOKEN_CODE => Some(Self::DeletePersonalAccessToken),
-            codes::SEND_MESSAGES_CODE => Some(Self::SendMessages),
-            codes::STORE_CONSUMER_OFFSET_CODE => Some(Self::StoreConsumerOffset),
-            _ => None,
+        match crate::dispatch::lookup_command(code) {
```
krishvishal, line 137: The `lookup_command` is searching an array O(n). It can be made to use `match` based O(1).

hubcio: good point. fixed.

PR #2974, opened 2026-03-18
feat(security): generate random JWT secrets when not configured

spetz:
Deploying the HTTP server with hardcoded default JWT secrets
is a security risk — every instance shares the same signing
key. Empty defaults now trigger secure random secret
generation at startup, with a warning logged showing a
redacted preview of the generated value.

When both encoding and decoding secrets are empty, a single
random secret is generated and used for both (symmetric
HMAC). Tokens are invalidated on restart, nudging operators
to set persistent secrets in production.

### DIFF
--- a/core/integration/src/harness/handle/server.rs
+++ b/core/integration/src/harness/handle/server.rs
@@ -410,7 +410,13 @@ impl ServerHandle {
             }
 
             if config_path.exists() {
-                return self.verify_bound_ports(&config_path);
+                match self.verify_bound_ports(&config_path) {
+                    Ok(()) => return Ok(()),
+                    Err(TestBinaryError::InvalidState { message })
+                        if message.starts_with("Failed to read server config at ")
+                            || message.starts_with("Failed to parse server config:") => {}
+                    Err(err) => return Err(err),
+                }
             }
 
             if Instant::now() >= deadline {
--- a/core/integration/tests/data_integrity/mod.rs
+++ b/core/integration/tests/data_integrity/mod.rs
@@ -18,4 +18,5 @@
 
 mod verify_after_server_restart;
 mod verify_consumer_group_partition_assignment;
+mod verify_no_plaintext_credentials_on_disk;
 mod verify_user_login_after_restart;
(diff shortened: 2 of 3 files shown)


core/server/src/http/http_server.rs
```diff
@@ -268,12 +269,30 @@ async fn build_app_state(
         tokens_path = shard.config.system.get_state_tokens_path();
     }
 
-    let jwt_manager = JwtManager::from_config(persister, &tokens_path, &config.jwt);
-    if let Err(e) = jwt_manager {
-        panic!("Failed to initialize JWT manager: {e}");
+    let mut jwt_config = config.jwt.clone();
+    let encoding_empty = jwt_config.encoding_secret.is_empty();
+    let decoding_empty = jwt_config.decoding_secret.is_empty();
+    if encoding_empty || decoding_empty {
```
hubcio: when only one of encoding/decoding is configured and the other is left empty, the `||` guard fires and generates a random secret - but it only fills the empty side. for HMAC (HS256/384/512) encoding and decoding must use the same key, so the server starts fine but every token validation returns `Unauthenticated`.

the both-empty and both-set cases work correctly. only the mixed case is broken - e.g. user sets `encoding_secret = "my_key"` but leaves `decoding_secret = ""`.

fix options:
- reject startup when exactly one is empty (fail fast with a clear error)
- copy the configured secret to the empty slot for HMAC algorithms

core/server/src/http/http_server.rs
```diff
@@ -268,12 +269,30 @@ async fn build_app_state(
...
+        let secret = crypto::generate_secret(32..64);
+        let redacted = secret.chars().take(3).collect::<String>();
+        if encoding_empty {
+            jwt_config.encoding_secret = secret.clone();
```
hubcio, line 282: `secret.clone()` is only needed when both branches fire (both empty). when only encoding is empty, the original `secret` is never used again since the `if decoding_empty` block won't execute. not a real issue since this is startup-only code - just a minor observation.

core/server/src/http/http_server.rs
spetz: Fixed with more warn logs.

Issue #2975, opened 2026-03-19
bug(helm): server podSecurityContext is not applied and UI probes target a 404 route


avirajkhare00:
### Bug description

This looks separate from #2932.

While testing the Helm chart locally on kind, I found two chart bugs:

1. The server deployment does not apply the chart's default podSecurityContext.
2. The UI liveness/readiness probes use /, but the UI returns 404 on that path.

What I tested

kind create cluster --name iggy-chart

helm install iggy ./helm/charts/iggy \
  -n iggy-test \
  --create-namespace \
  --set server.serviceMonitor.enabled=false

I used server.serviceMonitor.enabled=false because a clean local cluster does not have the ServiceMonitor CRD.

What I saw

- The server pod did not get the documented default podSecurityContext.
- The UI pod kept failing health checks because the probe path / returned 404.
- Changing the UI probe path to /auth/sign-in made the UI healthy.

Why this looks like a chart bug

Server deployment:

- helm/charts/iggy/templates/deployment.yaml uses server.podSecurityContext for the server pod.
- helm/charts/iggy/values.yaml defines podSecurityContext only at the root level.

So the server pod ends up without the expected default seccomp profile.

UI deployment:

- helm/charts/iggy/templates/deployment.yaml uses / for UI liveness/readiness probes.
- The running UI returns 404 on /, so the probes fail even when the process is up.

Expected

- The server deployment should use the chart's default podSecurityContext.
- The UI probes should use a route that returns 200.

Actual

- The server pod does not get the expected podSecurityContext.
- The UI pod restarts because the probes hit a 404 route.

Suggested fix

- Make the server deployment use the root podSecurityContext value.
- Change the UI probe path to a valid public route such as /auth/sign-in.

### Deployment

Docker (DockerHub image)

### Versions

server: apache/iggy:0.7.0, helm chart: 0.4.0

### Hardware / environment

Macbook M1 Pro

### Logs

Server error log

  2026-03-19T06:00:25.828867Z  INFO  main     iggy_server: Starting 6 shard(s)

  thread 'shard-0' (15) panicked at core/server/src/bootstrap.rs:189:10:
  called `Result::unwrap()` on an `Err` value: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }

  2026-03-19T06:00:25.833235Z ERROR shard-0  server::configs::sharding: Failed to bind memory Rust(Unsupported)
[...]

Issue #2932, opened 2026-03-14
bug(helm): Setup an ingress other than nginx


OneCricketeer:
### Bug description

The nginx ingress is EoL, and various other options are available, including Gateway API.

**Task** - Decide which option suits the project best. Possibly poll current chart users for quick win on broadest support model. 

### Deployment

Kubernetes

### Versions

0.7.0

hubcio:
hey @OneCricketeer, thanks for raising this.

you're right that `kubernetes/ingress-nginx` (the community controller) hit eol this month. worth noting that f5's `nginx-ingress` is a separate project and still actively maintained, but either way we shouldn't stay coupled to a single implementation.

to be transparent -no one on the core team actively uses helm or has deep k8s expertise, so the helm chart has always been community-driven. that means we'd really need someone from the community to take the lead on this. from what i can tell, **gateway api** seems like the right direction.

would you or anyone else be interested in owning this? we're happy to review and merge, but we'd need someone with hands-on k8s experience to drive the implementation.

also, this feels more like a feature/improvement than a bug - i'll relabel it.

OneCricketeer:
Yes, I too only observed commented and documented examples, and so I thought it was important to at least call out that documented values should suggest alternatives (thus made a bug issue)

avirajkhare00:
The Helm chart is now fully controller-neutral. Users can use any Ingress controller (Traefik, HAProxy, Contour, Kong, etc.) by setting className and appropriate annotations in their values:

```yaml
ingress:
  enabled: true
  className: "<your-ingress-class>"
  annotations: {}
```


What was done:
  • PR #2976 updated the README to use controller-neutral ingress examples
  • The templates already supported className and custom annotations



Remaining (minor):
  • Two commented # kubernetes.io/ingress.class: nginx examples in values.yaml could be made generic, but this is cosmetic

  @OneCricketeer - does this meet your expectations? If so, this issue can be closed.

PR #2976, opened 2026-03-19
fix: Helm chart install guidance and deployment defaults

avirajkhare00:

Closes #

https://github.com/apache/iggy/issues/2975

This PR is mainly for #2975, but it also partially cleans up #2932 by making the README ingress example controller-neutral.

  ## Rationale

  Local `kind` testing showed two real Helm chart bugs: the server deployment was not getting the documented default `podSecurityContext`, and the UI health probes were hitting a route
  that returns `404`.

  ## What changed?

  On a clean Helm install, the server deployment rendered the wrong pod security context because the template looked for `server.podSecurityContext` even though the chart defines
  `podSecurityContext` at the root. The UI deployment also used `/` for liveness and readiness probes, but the web UI returns `404` on that path, so Kubernetes kept restarting the pod.

  The fix makes the server deployment use the root `podSecurityContext` value and changes the UI probe path to `/auth/sign-in`, which returns `200`. I also updated the chart README to
  clarify the `ServiceMonitor` prerequisite, note the local `io_uring` caveat seen during testing, and use more controller-neutral ingress wording.

  ## Local Execution

  - Partially passed.
  - `helm lint ./helm/charts/iggy` passed.
  - `helm install iggy ./helm/charts/iggy -n iggy-test --create-namespace --set server.serviceMonitor.enabled=false` was run on a local `kind` cluster.
  - After the template changes, the server deployment rendered the expected root `podSecurityContext`, and the UI became healthy with `/auth/sign-in` as the probe path.
  - Full local startup still did not fully pass because the server hit a separate `io_uring` / runtime `Invalid argument` failure on `kind`. That looks like a different issue and is not
  fixed in this change.
  - Pre-commit hooks not ran. `prek` was not installed/configured in this checkout, and only the sample git hooks were present.

  ## AI Usage

  1. Tool used: OpenAI Codex / ChatGPT.
  2. Scope: investigation, small Helm template changes, README updates, and drafting the issue/PR text.
  3. Verification: I ran `helm lint`, installed the chart on local `kind`, inspected rendered pod specs, checked pod logs and Kubernetes events, and verified the UI route behavior with
  local HTTP checks.
  4. Yes, I can explain every line of the change if asked.


helm/charts/iggy/templates/deployment.yaml
```diff
@@ -168,11 +168,11 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: /
+              path: /auth/sign-in
               port: http
           readinessProbe:
             httpGet:
-              path: /
+              path: /auth/sign-in
```
atharvalade: `/auth/sign-in` is an application route, not a health endpoint. If this route is renamed or restructured in a future web UI release, the probes will silently break. Could we open a follow-up issue to add a lightweight `/healthz (or /ping)` endpoint to the web UI, similar to what the server already has? That would decouple the probe from the SPA

helm/charts/iggy/README.md
```diff
@@ -46,14 +53,16 @@ helm install iggy ./helm/charts/iggy \
 ```bash
 git clone https://github.com/apache/iggy.git
 cd iggy
-helm install iggy ./helm/charts/iggy
+helm install iggy ./helm/charts/iggy \
+  --set server.serviceMonitor.enabled=false
```
atharvalade: Every install example in this README now includes `--set server.serviceMonitor.enabled=false`. Rather than documenting a workaround in every example, should we change the default in `values.yaml` from `enabled: true to enabled: false?` That way a bare helm install works out of the box, and users with Prometheus Operator can opt in explicitly. The existing `values.yaml` comment already says "Enable this if you're using Prometheus Operator", which implies it should be opt-in.

helm/charts/iggy/templates/deployment.yaml
avirajkhare00: You are right, I have created an issue here: https://github.com/apache/iggy/issues/2977

I will revert back changes for web UI

atharvalade: sounds good! @avirajkhare00

helm/charts/iggy/README.md
avirajkhare00: I have changed `server.serviceMonitor.enabled` default to `false`

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -44,7 +44,7 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "iggy.serviceAccountName" . }}
       securityContext:
-        {{ toYaml .Values.server.podSecurityContext | nindent 8 }}
+        {{ toYaml .Values.podSecurityContext | nindent 8 }}
```
hubcio, line 50: this fixes a real bug - `.Values.server.podSecurityContext` never existed, so the server was silently getting no seccomp profile. the fix is correct.

two things worth considering:

1. any user who discovered the old template and worked around it by setting `server.podSecurityContext` in their custom values will have their override silently ignored after upgrading. a template guard would catch this:
```yaml
{{- if .Values.server.podSecurityContext }}
{{- fail "server.podSecurityContext has been moved to podSecurityContext (root level). Please update your values." }}
{{- end }}
```

2. both the server (here) and UI (line 158) deployments now share the same root-level `podSecurityContext` which sets `seccompProfile.type: Unconfined`. the UI is a SvelteKit frontend - it doesn't need io_uring privileges. same applies to `.Values.securityContext` with `IPC_LOCK` at line 51 vs 162. worth splitting into `server.*` / `ui.*` namespaces in a follow-up.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -98,11 +98,11 @@ spec:
 
           livenessProbe:
             httpGet:
-              path: /
+              path: /ping
```
hubcio, line 110: `/ping` is the right endpoint for health checks - returns a static `"pong"` with no auth or db access.

however, the probes use K8s defaults: `initialDelaySeconds: 0`, `periodSeconds: 10`, `failureThreshold: 3` - that gives the server 30 seconds to respond before the liveness probe kills it. for an io_uring server that needs to initialize ring buffers, restore state from persistent storage, and set up memory-locked regions, 30 seconds can be tight on constrained nodes or when recovering large `local_data` directories.

a `startupProbe` would decouple "is it still starting?" from "is it alive?" - e.g. `startupProbe` with `failureThreshold: 30` and `periodSeconds: 10` gives 5 min startup budget while keeping the liveness probe aggressive once running.

helm/charts/iggy/values.yaml
```diff
@@ -36,7 +36,7 @@ server:
 
   serviceMonitor:
     # -- Enable this if you're using [Prometheus Operator](https://github.com/coreos/prometheus-operator)
-    enabled: true
+    enabled: false
```
hubcio, line 39: the new default is correct - defaulting to `true` created a hard dependency on Prometheus Operator CRDs at install time, which broke fresh installs.

however, this is a breaking change for `helm upgrade`: existing deployments that relied on the default `true` (without pinning it in their values override) will silently lose their `ServiceMonitor` resource on the next upgrade. prometheus stops scraping with no error or warning.

this should be reflected in a chart version bump (currently `0.4.0` in `Chart.yaml`) - at least to `0.5.0` to signal the behavioral change.

helm/charts/iggy/README.md
```diff
@@ -175,13 +196,14 @@ kubectl port-forward svc/iggy-ui 3050:3050
...
-Enable ingress in values:
+Enable ingress in values. Set `className` and any controller-specific annotations to match your
+ingress implementation:
 
 ```yaml
 server:
   ingress:
     enabled: true
-    className: nginx
+    className: ""
```
hubcio: the surrounding text says "Set `className` and any controller-specific annotations to match your ingress implementation" but the example shows `className: ""`. an empty string causes the `ingressClassName` field to be omitted entirely from the rendered manifest (per `ingress.yaml:44-46`), which only works if the cluster has a default IngressClass configured.

either use a placeholder like `className: "<your-ingress-class>"` or add a note that leaving it empty falls back to the cluster's default ingress class.

helm/charts/iggy/templates/deployment.yaml
avirajkhare00: 1. added guradrails
2. Agreed. The UI should not need the server’s `io_uring`-related security settings. I’d prefer to handle that in a follow-up by splitting security settings into `server.*` and `ui.*` values rather than widening this PR further.

helm/charts/iggy/templates/deployment.yaml
avirajkhare00: fixed this, thanks for pointing it out

helm/charts/iggy/values.yaml
avirajkhare00: bumped up chart version

helm/charts/iggy/README.md
avirajkhare00: done. thanks for pointing out

helm/charts/iggy/templates/deployment.yaml
avirajkhare00: shall i create an issue around the same?

hubcio: no, you can just reference this comment. the aim of issue creation rule is to have a trace of `why` given change has to be implemented. in this case the source is my comment under this review.

PR #2983, opened 2026-03-19
feat: add UI /healthz endpoint

### LINKED ISSUE #2977: Add a dedicated /healthz endpoint to the UI for Helm liveness and readiness probes
The current Helm chart cannot safely use a normal web UI route such as `/auth/sign-in` for liveness/readiness probes.

  That route happens to return `200` today, but it is part of the application routing structure, not a stable health contract. If the auth flow or route layout changes in a future UI
  release, Kubernetes probes can start failing even though the UI container itself is healthy.

  We should add a dedicated health endpoint for the web UI and update Helm to use that instead of an SPA route.

  ## Why this is needed

  - Probe paths should be stable and independent from application navigation.
  - Health checks should not depend on the current auth page structure.
  - This avoids future regressions where frontend route changes accidentally break Kubernetes probes.

  ## Proposed direction

  Add a lightweight unauthenticated endpoint for probe usage, for example:

  - `GET /ping` or `GET /healthz` returning `200 OK`
  - no auth required
  - no redirect behavior
  - no dependency on frontend routing
  - no dependency on Iggy server connectivity for the basic probe path

  Optionally, we can also add a richer `GET /health` endpoint for diagnostics, which may include details such as backend reachability, but that should be treated separately from the basic
  probe endpoint.

  ## Follow-up in Helm

  After the endpoint exists, update the web UI liveness/readiness probes in the Helm chart to use the dedicated probe endpoint instead of `/auth/sign-in`.

avirajkhare00:

Closes #2977 

## Rationale

The current Helm chart cannot safely use a normal web UI route such as /auth/sign-in for liveness/readiness probes.

That route happens to return 200 today, but it is part of the application routing structure, not a stable health contract. If the auth flow or route layout changes in a future UI
release, Kubernetes probes can start failing even though the UI container itself is healthy.

We should add a dedicated health endpoint for the web UI and update Helm to use that instead of an SPA route.

## What changed?

Added:

 - `./web/src/routes/healthz/+server.ts`
 -  `./web/src/routes/healthz/route.test.mjs` 
 -  `./web/package.json`

## Local Execution

Passed both via npm test and creating an image locally and pushing it inside k8s cluster and using curl calling that endpoint

`prek is ran`

## AI Usage

1. OpenAI GPT 5.4
2. Editor: VS code
3. generated functions only
4. tested manually
5. yes. i can explain everything


web/src/routes/healthz/+server.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+const STATUS_OK = 'ok';
+
+export const prerender = true;
+
+export const GET = () =>
+  new Response(STATUS_OK, {
+    headers: {
+      'cache-control': 'no-store',
```
hubcio: `cache-control: no-store` with `prerender = true` is contradictory. since this is prerendered, sveltekit generates a static file at build time. the `Response` headers are only used during prerendering - once served as a static file (nginx, adapter-node, etc.), the actual http cache-control header will be whatever the static file server sets, not what's in the handler. this header is effectively dead code in a prerendered context.

either drop the `cache-control` header (since it has no effect), or document that the serving layer must set `cache-control: no-store` for `/healthz` if caching is a concern.

web/src/routes/healthz/route.test.mjs
```diff
@@ -0,0 +1,35 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import { GET, prerender } from './+server.ts';
```
hubcio: this `.mjs` file imports `.ts` directly. `node --test` will fail on this without `--experimental-strip-types` (node 22.6-23.5) or node >= 23.6 where type stripping is unflagged. the `package.json` test script is just `"test": "node --test"` with no flags.

either add `--experimental-strip-types` to the test script, rename the test to `.ts` and use a ts-aware runner, or import from compiled output.

hubcio: test file lives inside `src/routes/`. sveltekit only treats `+`-prefixed files as route files so this won't be served as a page, but co-locating test files in the routes directory is unconventional. if the project already has a convention for test file placement (e.g. `tests/` or `__tests__/`), consider following it.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -181,11 +181,11 @@ spec:
               protocol: TCP
           livenessProbe:
             httpGet:
-              path: /
+              path: /healthz
```
hubcio, line 184: no `startupProbe` for the UI deployment. PR #2976 added one for the server with `failureThreshold: 30` / `periodSeconds: 10` to give 5 min startup budget. the UI likely starts faster so this is minor, but for consistency you might want one here too.

PR #2986, opened 2026-03-20
feat(go): implement binary reader/writer

chengxilo:

N/A

## Rationale

Iggy, when read/write binary, use little endian. When there is a string, it has a prefix provide its length, and the actually string body. So instead of using the `binary.LittleEndian.Uint32(payload[processIDPos : processIDPos+4])` and use `position ++` , why not just wrap everything in a reader/writer to make it simple and clear.

As the Stats deserialize function need to be updated, I think this change would be helpful. https://github.com/apache/iggy/pull/2964#discussion_r2952160233

Here is a example that shows the difference:
Before:
https://github.com/apache/iggy/blob/9c8b5cc6ca682de6f3e7e0074224517e5488dc40/foreign/go/binary_serialization/stats_serializer.go#L53-L90

After
```go
func (s *Stats) UnmarshalBinary(payload []byte) error {
	r := codec.newReader(payload)
	s.ProcessId = r.u32()
	s.CpuUsage = r.f32()
	s.TotalCpuUsage = r.f32()
	s.MemoryUsage = r.u64()
	s.TotalMemory = r.u64()
	s.AvailableMemory = r.u64()
	s.RunTime = r.u64()
	s.StartTime = r.u64()
	s.ReadBytes = r.u64()
	s.WrittenBytes = r.u64()
	s.MessagesSizeBytes = r.u64()
	s.StreamsCount = r.u32()
	s.TopicsCount = r.u32()
	s.PartitionsCount = r.u32()
	s.SegmentsCount = r.u32()
	s.MessagesCount = r.u64()
	s.ClientsCount = r.u32()
	s.ConsumerGroupsCount = r.u32()
	s.Hostname = r.u32LenStr()
	s.OsName = r.u32LenStr()
	s.OsVersion = r.u32LenStr()
	s.KernelVersion = r.u32LenStr()
	return r.Err()
}
```

Example of writer
```
func buildStatsPayload(
	processId uint32,
	cpuUsage, totalCpuUsage float32,
	memoryUsage, totalMemory, availableMemory uint64,
	runTime, startTime uint64,
	readBytes, writtenBytes, messagesSizeBytes uint64,
	streamsCount, topicsCount, partitionsCount, segmentsCount uint32,
	messagesCount uint64,
	clientsCount, consumerGroupsCount uint32,
	hostname, osName, osVersion, kernelVersion, iggyServerVersion string,
	iggyServerSemver *uint32,
	cacheMetrics []CacheMetrics,
	threadsCount uint32,
	freeDiskSpace, totalDiskSpace uint64,
) []byte {
	w := codec.NewWriter()
	w.U32(processId)
	w.F32(cpuUsage)
	w.F32(totalCpuUsage)
	w.U64(memoryUsage)
	w.U64(totalMemory)
	w.U64(availableMemory)
	w.U64(runTime)
	w.U64(startTime)
	w.U64(readBytes)
	w.U64(writtenBytes)
	w.U64(messagesSizeBytes)
	w.U32(streamsCount)
	w.U32(topicsCount)
	w.U32(partitionsCount)
	w.U32(segmentsCount)
	w.U64(messagesCount)
	w.U32(clientsCount)
	w.U32(consumerGroupsCount)
	w.U32LenStr(hostname)
	w.U32LenStr(osName)
	w.U32LenStr(osVersion)
	w.U32LenStr(kernelVersion)
	w.U32LenStr(iggyServerVersion)

	if iggyServerSemver != nil {
		w.U32(*iggyServerSemver)
	}

	w.U32(uint32(len(cacheMetrics)))
	for _, cm := range cacheMetrics {
		w.U32(cm.StreamId)
		w.U32(cm.TopicId)
		w.U32(cm.PartitionId)
		w.U64(cm.Hits)
		w.U64(cm.Misses)
		w.F32(cm.HitRatio)
	}

	if iggyServerSemver != nil {
		w.U32(threadsCount)
		w.U64(freeDiskSpace)
		w.U64(totalDiskSpace)
	}

	return w.Bytes()
}
```

## What changed?

A reader for byte slice is created.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. smart Claude Sonnet 4.6 and my rotten brain
2. entire implementation, manually modified some part to make sure the error output readablity.
3. Reviewed the tests and code, actually modified some logic of reading binary and works well.
4. yes

### DIFF
--- a/.github/actions/python-maturin/post-merge/action.yml
+++ b/.github/actions/python-maturin/post-merge/action.yml
@@ -151,11 +151,13 @@ runs:
         echo "View on PyPI: https://pypi.org/project/apache-iggy/$VERSION/"
       shell: bash
 
-    - name: Publish to PyPI using maturin
+    - name: Install uv
       if: inputs.dry_run == 'false'
-      uses: PyO3/maturin-action@v1
+      uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 # v7.3.1
+
+    - name: Publish to PyPI
+      if: inputs.dry_run == 'false'
+      run: uv publish --check-url https://pypi.org/simple/ --token "$PYPI_API_TOKEN" ${{ inputs.wheels_path }}/*
+      shell: bash
       env:
-        MATURIN_PYPI_TOKEN: ${{ env.PYPI_API_TOKEN }}
-      with:
-        command: upload
-        args: --non-interactive --skip-existing ${{ inputs.wheels_path }}/*
+        PYPI_API_TOKEN: ${{ env.PYPI_API_TOKEN }}
--- a/core/binary_protocol/src/codec.rs
+++ b/core/binary_protocol/src/codec.rs
@@ -75,6 +75,32 @@ pub fn read_u8(buf: &[u8], offset: usize) -> Result<u8, WireError> {
         })
 }
 
+/// Helper to read a `u16` LE from `buf` at `offset`.
+///
+/// # Errors
+/// Returns `WireError::UnexpectedEof` if fewer than 2 bytes remain.
+#[allow(clippy::missing_panics_doc)]
+#[inline]
+pub fn read_u16_le(buf: &[u8], offset: usize) -> Result<u16, WireError> {
+    let end = offset
+        .checked_add(2)
+        .ok_or_else(|| WireError::UnexpectedEof {
+            offset,
+            need: 2,
+            have: buf.len().saturating_sub(offset),
+        })?;
+    let slice = buf
+        .get(offset..end)
+        .ok_or_else(|| WireError::UnexpectedEof {
+            offset,
+            need: 2,
+            have: buf.len().saturating_sub(offset),
+        })?;
+    Ok(u16::from_le_bytes(
+        slice.try_into().expect("slice is exactly 2 bytes"),
+    ))
+}
+
 /// Helper to read a `u32` LE from `buf` at `offset`.
 ///
 /// # Errors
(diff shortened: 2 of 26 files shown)


foreign/go/internal/codec/reader.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	if r.pos+1 > len(r.p) {
+		r.overrun(1)
+		return 0
+	}
+	v := r.p[r.pos]
+	r.pos++
+	return v
+}
+
```
ryankert01, line 61: We probably can add a `u16` for https://github.com/apache/iggy/blob/f4b0af5f7065e559b2ec52829e84dac1cd1b5d7e/foreign/go/contracts/transport_endpoints.go#L66-L69

chengxilo: solved, a method called `u16` is implemented in the reader. tks

foreign/go/internal/codec/reader.go
```diff
@@ -0,0 +1,179 @@
...
+// Str reads exactly n bytes as a string. Use U8LenStr or U32LenStr instead
+// if the data is length-prefixed:
+//
+//	[length: 1 byte][data: N bytes]   → U8LenStr
+//	[length: 4 bytes][data: N bytes]  → U32LenStr
+func (r *Reader) Str(n int) string {
+	if r.err != nil {
+		return ""
+	}
+	if r.pos+n > len(r.p) {
```
hubcio, line 150: `Str(n int)` with a negative `n` bypasses the bounds check here - `r.pos + (-1)` is `-1`, and `-1 > len(r.p)` is false - so it falls through to `strN(-1)`, which panics on `r.p[0:-1]` (negative slice bounds). no current caller passes negative `n`, but since this is an exported method accepting `int`, a defensive guard would be cheap:

```go
if n < 0 {
    r.overrun(n)
    return ""
}
```

foreign/go/internal/codec/writer.go
```diff
@@ -0,0 +1,86 @@
...
+// U32LenStr writes a length-prefixed string where the length is a 4-byte
+// little-endian unsigned integer.
+func (w *Writer) U32LenStr(v string) {
+	w.p = binary.LittleEndian.AppendUint32(w.p, uint32(len(v)))
+	w.Str(v)
+}
+
+// U8LenStr writes a length-prefixed string where the length is a single byte.
+func (w *Writer) U8LenStr(v string) {
+	w.p = append(w.p, uint8(len(v)))
```
hubcio, line 114: `uint8(len(v))` silently wraps for strings longer than 255 bytes - a 256-byte string writes length `0x00` followed by all 256 bytes. the reader would then see length=0 and return `""`, silently corrupting the stream. consider adding a length check or documenting the max size contract.

foreign/go/internal/codec/reader_test.go
```diff
@@ -0,0 +1,275 @@
...
+			_, file, line, _ := runtime.Caller(0)
+			r.F32()
+			return file, line + 1
+		}},
+		{"Str", func(r *Reader) (string, int) {
+			_, file, line, _ := runtime.Caller(0)
+			r.Str(1)
+			return file, line + 1
+		}},
+		{"U32LenStr", func(r *Reader) (string, int) {
```
hubcio: this test case (and `U8LenStr` below) passes an empty buffer, so the overrun always fires on the length-prefix read. the *body*-phase overrun path (prefix succeeds but body is too short) is untested for caller location correctness. e.g. `NewReader(cat(u32le(100), []byte("short")))` then `r.U32LenStr()` would exercise the second `overrun` call in `U32LenStr`.

foreign/go/internal/codec/reader.go
chengxilo: Since the codec is in `/internal`, these methods are only used by us. So it may be acceptable not to validate this case, we should be responsible for ensuring that the input n is correct. I added the overrun check to handle potentially malformed bytes, but validating invalid arguments may not be necessary.

foreign/go/internal/codec/writer.go
chengxilo: Fixed.

foreign/go/internal/codec/reader_test.go
chengxilo: solved

PR #2988, opened 2026-03-20
feat(cpp): Add functions related to consumer groups

slbotbm:

Works towards completion of #2763

## Rationale

Additions to the cpp bindings to make it feature complete.
## What changed?

Add consumer-group creating and deletion functions. Also adds commented out purge_topic and purge_stream but those require the ability to actually send messages to the server, which is not yet implemented. Will be added once they can be tested. Also adds create_partition, delete_partitions, and functions that allow cpp user to directly create identifiers instead of manually defining everything.

## Local Execution

- Passed 
- Pre-commit hooks ran

## AI Usage

1. codex
2. code generation
3. manual planning and inspection, local execution


foreign/cpp/src/client.rs
```diff
@@ -53,22 +53,22 @@ pub fn new_connection(connection_string: String) -> Result<*mut Client, String>
 }
 
 impl Client {
-    pub fn connect(&self) -> Result<(), String> {
+    pub fn login_user(&self, username: String, password: String) -> Result<(), String> {
```
amlel-el-mahrouss, line 56: why do you take `self` but pass `username` and `password` after independently?

foreign/cpp/src/lib.rs
```diff
@@ -81,6 +97,49 @@ mod ffi {
...
+            stream_id: Identifier,
+            topic_id: Identifier,
+            group_id: Identifier,
+        ) -> Result<()>;
+        fn delete_consumer_group(
+            &self,
+            stream_id: Identifier,
+            topic_id: Identifier,
+            group_id: Identifier,
+        ) -> Result<()>;
```
amlel-el-mahrouss, line 131: This could be wrapped in a `impl Group` right?

foreign/cpp/src/topic.rs
```diff
@@ -25,9 +25,9 @@ impl From<RustTopic> for ffi::Topic {
             created_at: topic.created_at.as_micros(),
             name: topic.name,
             size_bytes: topic.size.as_bytes_u64(),
-            message_expiry: topic.message_expiry.into(),
+            message_expiry: u64::from(topic.message_expiry),
```
amlel-el-mahrouss, line 28: What if it's ill formed data?

hubcio: u64::from()/into() are infallible

foreign/cpp/src/client.rs
slbotbm: I did not understand the question. I take username and password from the user to login them in. &self signifies that this method belongs to the Client type.

foreign/cpp/src/lib.rs
slbotbm: What type of implementation are you thinking of? Could you please give a concrete example?

slbotbm: Is it something like:
```rust
type ConsumerGroup

impl ConsumerGroup {
    def create_consumer_group()
    def join_consumer_group()
    ....
}
```
?

amlel-el-mahrouss: Yes exactly this kind of type.

foreign/cpp/src/topic.rs
amlel-el-mahrouss: Got it thx

foreign/cpp/src/client.rs
amlel-el-mahrouss: Nevermind, i dont think my question was a good one anyways, putting username and password in the Client type would be not a good design decision.

foreign/cpp/src/lib.rs
slbotbm: creating a `ConsumerGroup` opaque type would force the end user to handle the clean up for it. I would like to keep the surface area for making a mistake as small as possible, that is why i did not create another opaque type and a implementation based on it. Do you think there are any advantages to creating a `ConsumerGroup` type and its related functions?

amlel-el-mahrouss (approved):
Approved, one note, what about UTF-16?

hubcio (approved):

PR #2990, opened 2026-03-20
feat(ci): add version bump tooling and consistency checks

hubcio:
Bumping versions across 14 components in 6 ecosystems was
entirely manual - editing 30+ files with no validation.

Add bump-version.sh for automated version bumps with
ecosystem-aware translation (Rust -edge.N, Python .devN,
Java -SNAPSHOT), a state machine enforcing valid
transitions, and dry-run support. Supports individual
components, rust-all, and --all targets.

Extend extract-version.sh with --all (version table) and
--check (workspace dep + Python dual-file validation).
Add CI job and pre-commit hook for consistency checks.

Rename python-version-sync.sh to python-sdk-version-sync.sh.

NOTE: bump-version.sh is script NOT supposed to be called
by CI machinery. It's intended to be used ONLY manually by
maintainers.


scripts/bump-version.sh
```diff
@@ -0,0 +1,605 @@
...
+            failed=$((failed + 1))
+            echo -e "${YELLOW}Skipped ${comp} (transition not valid)${NC}"
+        fi
+        echo ""
+    done
+    if [[ $failed -gt 0 ]]; then
+        echo -e "${YELLOW}${failed} component(s) skipped due to invalid transitions.${NC}"
+        echo -e "Use ${CYAN}--set${NC} to handle them individually."
+    fi
+}
```
atharvalade, line 620: `md_bump_multi` counts failures in failed but never returns a non-zero status, so the function always completes “successfully” from the shell’s point of view even when one or more `cmd_bump` calls failed

scripts/bump-version.sh
```diff
@@ -0,0 +1,605 @@
...
+# Remaining args are flags passed to cmd_bump.
+case "$1" in
+    --all)
+        shift
+        cmd_bump_multi "$ALL_COMPONENTS" "$@"
+        exit 0 ;;
+    rust-all)
+        shift
+        cmd_bump_multi "$RUST_COMPONENTS" "$@"
+        exit 0 ;;
```
atharvalade: The `--all` and `rust-all `branches always run exit 0 after `cmd_bump_multi`, so the process exit code is 0 even when `failed > 0`. That makes `rust-all / --all` unsafe for release automation or any script that relies on `$? `(partial bumps look like full success). Suggested direction: have `cmd_bump_multi` return non-zero when failed > 0, and replace the unconditional exit 0 with propagating that status (e.g. `cmd_bump_multi` `...; exit $?` or `exit 1` when any component failed).

PR #2992, opened 2026-03-20
chore(python): bump SDK to 0.7.4-dev1, modernize CI and deps

hubcio:
maturin upload/publish is deprecated (PyO3/maturin#2334)
and will be removed in maturin 2.0. Replace with
uv publish which the maturin maintainers recommend.

Replace pyo3-stub-gen git rev pin with crates.io 0.19.0.
The pin was needed in Sep 2025 when no release supported
pyo3 >= 0.26.0; that's long resolved.

Also update Rust edition to 2024, add missing Python
3.13 classifier, and regenerate stale uv.lock.

### DIFF
--- a/core/binary_protocol/src/codec.rs
+++ b/core/binary_protocol/src/codec.rs
@@ -75,6 +75,32 @@ pub fn read_u8(buf: &[u8], offset: usize) -> Result<u8, WireError> {
         })
 }
 
+/// Helper to read a `u16` LE from `buf` at `offset`.
+///
+/// # Errors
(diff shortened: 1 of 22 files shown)


foreign/python/pyproject.toml
slbotbm: Here, I think it would make sense to chage the authors to maybe "Iggy Committers" and the email to dev@iggy.apache.org since the listed authors are not active anymore

hubcio: good point

PR #2998, opened 2026-03-20
feat(cli): Add context create/delete CLI commands to complete login/logout workflow

### LINKED ISSUE #595: Add login / logout commands to `iggy` cli to reduce credentials related args / flags
Introduce similar solution to `docker login` / `docker logout` - which would allow user to use `iggy` command line tool without a need of putting credentials for each command. (See https://docs.docker.com/engine/reference/commandline/login/ and https://docs.docker.com/engine/reference/commandline/logout/)

atharvalade:

Closes #595

## Rationale

The login/logout commands and context infrastructure (list, use) already existed, but users had to manually edit `~/.iggy/contexts.toml` to create or remove contexts. This PR adds the missing `context create` and `context delete` subcommands so the entire credential workflow can be managed entirely through the CLI, similar to `docker login`/`docker logout` + `docker context`.

## What changed?

Users could not create or delete contexts via the CLI — they had to manually edit TOML files. This made the docker-login-like workflow incomplete.

Added `iggy context create <name> [--transport, --tcp-server-address, --username, ...]` and `iggy context delete <name>` commands with proper validation (duplicate prevention, cannot delete `default`, active context reset on deletion), along with integration tests covering success paths, edge cases, and help text.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Opus 4.6
2. Minimal AI used
3. Verified by building, running all commands locally with edge cases, and running cargo fmt/clippy/license checks
4. Yes

Screenshot 2026-03-20 at 2 08 46 PM
Screenshot 2026-03-20 at 2 08 53 PM
Screenshot 2026-03-20 at 2 09 01 PM
Screenshot 2026-03-20 at 2 09 16 PM
Screenshot 2026-03-20 at 2 09 22 PM
Screenshot 2026-03-20 at 2 09 29 PM
Screenshot 2026-03-20 at 2 09 49 PM
Screenshot 2026-03-20 at 2 09 56 PM

### DIFF
--- a/foreign/go/contracts/users.go
+++ b/foreign/go/contracts/users.go
@@ -60,10 +60,12 @@ func (p *Permissions) MarshalBinary() ([]byte, error) {
 
 	position := 10
 
-	if p.Streams != nil {
+	if len(p.Streams) > 0 {
 		bytes[position] = byte(1)
 		position += 1
 
+		streamsCount := len(p.Streams)
+		currentStream := 1
 		for streamID, stream := range p.Streams {
 			binary.LittleEndian.PutUint32(bytes[position:position+4], uint32(streamID))
 			position += 4
@@ -76,10 +78,12 @@ func (p *Permissions) MarshalBinary() ([]byte, error) {
 			bytes[position+5] = boolToByte(stream.SendMessages)
 			position += 6
 
-			if stream.Topics != nil {
+			if len(stream.Topics) > 0 {
 				bytes[position] = byte(1)
 				position += 1
 
+				topicsCount := len(stream.Topics)
+				currentTopic := 1
 				for topicID, topic := range stream.Topics {
 					binary.LittleEndian.PutUint32(bytes[position:position+4], uint32(topicID))
 					position += 4
@@ -90,16 +94,29 @@ func (p *Permissions) MarshalBinary() ([]byte, error) {
 					bytes[position+3] = boolToByte(topic.SendMessages)
 					position += 4
 
-					bytes[position] = byte(0)
+					if currentTopic < topicsCount {
+						currentTopic++
+						bytes[position] = byte(1)
+					} else {
+						bytes[position] = byte(0)
+					}
 					position += 1
 				}
 			} else {
 				bytes[position] = byte(0)
 				position += 1
 			}
+
+			if currentStream < streamsCount {
+				currentStream++
+				bytes[position] = byte(1)
+			} else {
+				bytes[position] = byte(0)
+			}
+			position += 1
 		}
 	} else {
-		bytes[0] = byte(0)
+		bytes[position] = byte(0)
 	}
 
 	return bytes, nil
--- a/foreign/go/contracts/users_test.go
+++ b/foreign/go/contracts/users_test.go
@@ -1,0 +1,746 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package iggcon
+
+import (
+	"encoding/binary"
+	"testing"
+)
+
+// requireRemaining checks that there are enough bytes remaining in the slice
+func requireRemaining(t *testing.T, data []byte, position, needed int, context string) {
+	t.Helper()
+	if position+needed > len(data) {
+		t.Fatalf("%s: need %d bytes at position %d, but only %d bytes remain (total length: %d)",
+			context, needed, position, len(data)-position, len(data))
+	}
+}
+
+func TestPermissions_MarshalBinary_WithStreamsAndTopics(t *testing.T) {
+	// Test case: Permissions with 2 streams, first stream has 2 topics, second has none
+	permissions := &Permissions{
+		Global: GlobalPermissions{
+			ManageServers: true,
+			ReadServers:   false,
+			ManageUsers:   false,
+			ReadUsers:     true,
+			ManageStreams: true,
+			ReadStreams:   true,
+			ManageTopics:  false,
+			ReadTopics:    false,
+			PollMessages:  true,
+			SendMessages:  false,
+		},
+		Streams: map[int]*StreamPermissions{
+			1: {
+				ManageStream: true,
+				ReadStream:   true,
+				ManageTopics: false,
+				ReadTopics:   false,
+				PollMessages: true,
+				SendMessages: false,
+				Topics: map[int]*TopicPermissions{
+					10: {
+						ManageTopic:  true,
+						ReadTopic:    false,
+						PollMessages: false,
+						SendMessages: true,
+					},
+					20: {
+						ManageTopic:  false,
+						ReadTopic:    false,
+						PollMessages: true,
+						SendMessages: true,
+					},
+				},
+			},
+			2: {
+				ManageStream: false,
+				ReadStream:   true,
+				ManageTopics: true,
+				ReadTopics:   false,
+				PollMessages: false,
+				SendMessages: true,
+				Topics:       nil,
+			},
+		},
+	}
+
+	bytes, err := permissions.MarshalBinary()
+	if err != nil {
+		t.Fatalf("MarshalBinary failed: %v", err)
+	}
+
+	// Verify structure and values
+	position := 0
+
+	// Verify global permissions (10 bytes)
+	requireRemaining(t, bytes, position, 10, "global permissions")
+	expectedGlobal := []byte{1, 0, 0, 1, 1, 1, 0, 0, 1, 0}
+	for i := 0; i < 10; i++ {
+		if bytes[position+i] != expectedGlobal[i] {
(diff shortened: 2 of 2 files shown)


core/cli/src/commands/binary_context/common.rs
```diff
@@ -126,6 +126,69 @@ impl ContextManager {
...
+        if cs.contexts.contains_key(name) {
+            bail!("context '{name}' already exists in {CONTEXTS_FILE_NAME}")
+        }
+
+        let mut new_contexts = cs.contexts.clone();
+        new_contexts.insert(name.to_string(), config);
+
+        self.context_rw.ensure_iggy_home_exists()?;
+        self.context_rw
+            .write_contexts(new_contexts.clone())
```
hubcio, line 154: this is the first production code path that writes credentials (`--password`, `--token`) to `contexts.toml`. `tokio::fs::write` creates files with default permissions (typically `0644` - world-readable). should set `0600` on the file after writing, or use a file-creation API that sets restrictive permissions. same applies to the `write_contexts` call in `delete_context` at line 180.

core/cli/src/commands/binary_context/common.rs
```diff
@@ -126,6 +126,69 @@ impl ContextManager {
         Ok(context_state.contexts.clone())
     }
 
+    pub async fn create_context(&mut self, name: &str, config: ContextConfig) -> Result<()> {
```
hubcio, line 135: no validation on `name` - empty strings, whitespace-only, and control characters are all accepted. an empty string creates a `[""]
` TOML section. the toml crate handles quoting for most special chars and fails cleanly for truly bad ones (e.g. newlines in table keys), so this isn't a corruption risk, but it produces confusing config entries.

a basic check like `name.trim().is_empty()` plus maybe an alphanumeric+hyphen+underscore filter would be a good guard.

also: `create_context("default", ...)` is implicitly blocked because `get_context_state` always seeds a "default" entry, so `contains_key` catches it. but an explicit `if name == DEFAULT_CONTEXT_NAME { bail!(...) }` guard (like `delete_context` has) would be more resilient to future changes.

core/cli/src/args/context.rs
```diff
@@ -50,3 +75,134 @@ pub(crate) struct ContextUseArgs {
...
+pub(crate) struct ContextCreateArgs {
+    /// Name of the context to create
+    #[arg(value_parser = clap::value_parser!(String))]
+    pub(crate) context_name: String,
+
+    /// The transport to use
+    ///
+    /// Valid values are "quic", "http", "tcp" and "ws".
+    #[clap(verbatim_doc_comment, long)]
+    pub(crate) transport: Option<String>,
```
hubcio, line 89: `transport` is a free-form `Option<String>` with no parse-time validation. `--transport foobar` gets silently written to `contexts.toml` and only fails later at connection time with an unrelated error.

a `TransportProtocol` enum already exists in the codebase (`core/common/src/types/configuration/transport.rs`) with strum-based `FromStr`. you could validate here with `clap::value_parser!` or at minimum call `TransportProtocol::from_str` in `create_context` before writing.

this follows the pre-existing `ArgsOptional` pattern, so not a regression - but `context create` is a new user-facing entry point where fail-fast validation matters more.

also heads up: there's a pre-existing mismatch between `TransportProtocol`'s strum serialization (`"ws"`) and `Args::get_server_address` which matches against `"websocket"`. a user creating a context with `--transport ws` would hit this.

core/cli/src/commands/binary_context/common.rs
```diff
@@ -126,6 +126,69 @@ impl ContextManager {
...
+        Ok(())
+    }
+
+    pub async fn delete_context(&mut self, name: &str) -> Result<()> {
+        if name == DEFAULT_CONTEXT_NAME {
+            bail!("cannot delete the '{DEFAULT_CONTEXT_NAME}' context")
+        }
+
+        self.get_context_state().await?;
+        let cs = self.context_state.take().unwrap();
```
hubcio: minor: `delete_context` uses `.take().unwrap()` here while `create_context` uses `.as_ref().unwrap()` at line 131. the `.take()` pattern means `context_state` is `None` if any subsequent operation bails (lines 163, 171, 180). `create_context`'s `.as_ref()` approach is safer since it doesn't consume the state.

in practice this doesn't matter - every `execute_cmd` creates a fresh `ContextManager` and discards it, so no caller ever makes a second call on the same instance. but for API consistency and robustness, matching `create_context`'s pattern would be cleaner.

(the pre-existing `set_active_context_key` at line 100 has the same `.take()` pattern.)

core/cli/src/commands/binary_context/common.rs
```diff
@@ -258,6 +321,18 @@ impl ContextReaderWriter {
         Ok(())
     }
 
+    pub fn ensure_iggy_home_exists(&self) -> Result<()> {
+        if let Some(ref iggy_home) = self.iggy_home
+            && !iggy_home.exists()
+        {
+            std::fs::create_dir_all(iggy_home).context(format!(
```
hubcio: `std::fs::create_dir_all` is a blocking call inside an async context. also `iggy_home.exists()` in the guard is blocking. `tokio::fs::create_dir_all` would be the async-consistent replacement. not a practical issue for a CLI, but inconsistent with the rest of this file which uses `tokio::fs` everywhere.

core/cli/src/commands/binary_context/delete_context.rs
```diff
@@ -0,0 +1,61 @@
...
+    async fn execute_cmd(&mut self, _client: &dyn Client) -> anyhow::Result<(), anyhow::Error> {
+        let mut context_mgr = ContextManager::default();
+
+        context_mgr.delete_context(&self.context_name).await?;
+
+        event!(target: PRINT_TARGET, Level::INFO, "context '{}' deleted successfully", self.context_name);
+
+        Ok(())
+    }
+}
```
hubcio, line 61: nit: `create_context.rs` has a `#[cfg(test)]` module with tests for `explain()`, `login_required()`, and `connection_required()`. `delete_context.rs` is missing the same for consistency.

core/cli/src/commands/binary_context/common.rs
hubcio, line 52: `#[serde(flatten)]` on `ArgsOptional` means unknown fields are silently dropped during deserialization. since `create_context` and `delete_context` do a read-all/modify-one/write-all cycle on `contexts.toml`, running an older CLI binary against a config written by a newer version will strip any new fields from ALL context entries - not just the one being created/deleted. consider adding `#[serde(flatten)] pub extra: HashMap<String, toml::Value>` to `ContextConfig` to preserve unknown keys through round-trips.

core/integration/tests/cli/context/common.rs
```diff
@@ -64,6 +64,10 @@ impl TestIggyContext {
         self.context_manager.read_active_context().await.unwrap()
```
hubcio, line 64: `read_active_context` returns the raw `read_to_string` result without `.trim()`. if a user (or text editor) adds a trailing newline to `.active_context`, the key won't match any context and `get_context_state` bails with a confusing "missing" error. adding `.map(|s| s.trim().to_string())` is cheap insurance.

core/cli/src/commands/binary_context/common.rs
hubcio: `ContextsConfigMap` is `HashMap<String, ContextConfig>`. every write may reorder sections in `contexts.toml`. `BTreeMap` would give stable, alphabetically sorted output. pre-existing, but worth a follow-up.

core/cli/src/commands/binary_context/common.rs
atharvalade: Thanks for catching this -- added `set_owner_only_permissions` that sets 0600 on `contexts.toml` after every write (unix only, no-op on Windows).

core/cli/src/commands/binary_context/common.rs
atharvalade: added `validate_context_name` that rejects empty, whitespace-only, and non-alphanumeric/hyphen/underscore names, plus an explicit if `name == ``DEFAULT_CONTEXT_NAME` guard in create_context.

core/cli/src/args/context.rs
atharvalade: added `validate_transport` in `create_context.rs` that calls `TransportProtocol::from_str `before writing, so invalid values like --transport foobar now fail immediately with a clear message listing valid options.

core/cli/src/commands/binary_context/common.rs
atharvalade: switched `delete_context` to use `.as_ref().unwrap()` with `.clone()` to match `create_context` and avoid leaving `context_state` as `None` if a later operation bails

core/cli/src/commands/binary_context/common.rs
atharvalade: replaced `std::fs::create_dir_all` and `iggy_home.exists()` with t`okio::fs::create_dir_all` and `tokio::fs::try_exists` to stay consistent with the rest of the file.

core/cli/src/commands/binary_context/delete_context.rs
atharvalade: added a matching `#[cfg(test)] `module with tests for `explain(),` `login_required()`, and `connection_required()`

core/cli/src/commands/binary_context/common.rs
atharvalade: added pub extra: `BTreeMap<String, toml::Value> `with `#[serde(flatten)]` to `ContextConfig` so unknown keys from newer CLI versions survive the read-modify-write cycle.

core/integration/tests/cli/context/common.rs
atharvalade: added `.map(|s| s.trim().to_string())` to `read_active_context` so a trailing newline from a text editor won't cause a confusing "missing" error

core/cli/src/commands/binary_context/common.rs
atharvalade: switched `ContextsConfigMap` from `HashMap` to `BTreeMap` so context sections in `contexts.toml` are always written in alphabetical order

Issue #2982, opened 2026-03-19
Go SDK: UpdatePermissions omits mandatory `permissions_len` field when permissions are nil


atharvalade:
Same root cause as the `CreateUser` issue. The Rust wire format for `UpdatePermissions` always includes `permissions_len:u32_le` on the wire regardless of `has_permissions` (see [update_permissions.rs#L37-L39](https://github.com/apache/iggy/blob/master/core/binary_protocol/src/requests/users/update_permissions.rs#L37-L39) and the decoder at [lines 64-65](https://github.com/apache/iggy/blob/master/core/binary_protocol/src/requests/users/update_permissions.rs#L64-L65)).

In the Go SDK (`foreign/go/internal/command/user.go`), the nil-permissions path writes only the flag byte:

```
} else {
    bytes[position] = 0 // missing permissions_len after this
}
```

Note: #2973 partially fixes this function — it adds +1 to the base length to prevent a panic when permissions are nil. But the output is still 4 bytes short of what the server expects.

Fix: Change the base length from len(userIdBytes) + 1 to len(userIdBytes) + 1 + 4, and write 4 zero bytes for permissions_len in the else branch.

Issue #2981, opened 2026-03-19
Go SDK: CreateUser omits mandatory permissions_len field when permissions are nil


atharvalade:
The Rust wire format for `CreateUser` always includes `permissions_len:u32_le` on the wire, even when `has_permissions=0`. The server-side decoder unconditionally reads these 4 bytes (see [create_user.rs#L82-L83](https://github.com/apache/iggy/blob/master/core/binary_protocol/src/requests/users/create_user.rs#L82-L83)), and the Rust encoder always writes them (see [create_user.rs#L61-L62](https://github.com/apache/iggy/blob/master/core/binary_protocol/src/requests/users/create_user.rs#L61-L62)).

In the Go SDK (`foreign/go/internal/command/user.go`), the nil-permissions branch only writes `has_permissions=0` (1 byte) and skips the 4-byte `permissions_len=0`:

```
} else {
    bytes[position] = byte(0) // writes flag only, no permissions_len
}
```

The server will try to read 4 bytes that aren't there, causing a parse failure or reading into the next message.

Fix: Write permissions_len=0 (4 zero bytes) after the flag in the else branch, and add 4 to the base capacity.

PR #3015, opened 2026-03-23
fix(go): correct Permissions serialization and add unit tests

### LINKED ISSUE #2980: Go SDK: CreatePersonalAccessToken writes expiry as uint32 at wrong offset in uint64 field
The binary wire protocol defines the `CreatePersonalAccessToken` format as `[name_len:u8][name:N][expiry:u64_le]`  the expiry is an 8-byte little-endian uint64 (see [create_personal_access_token.rs#L25](https://github.com/apache/iggy/blob/master/core/binary_protocol/src/requests/personal_access_tokens/create_personal_access_token.rs#L25)).

The Go SDK gets this wrong in two ways:

1. The `Expiry` field is declared as `uint32` instead of `uint64`
2. `MarshalBinary()` uses `PutUint32(bytes[len(bytes)-4:], c.Expiry)` which writes 4 bytes at offset `1 + len(name) + 4` instead of writing 8 bytes at offset `1 + len(name)`

This leaves 4 zero bytes between the name and the actual expiry value. The server reads all 8 bytes as a single `u64_le`, so the expiry ends up in the high 32 bits — effectively `value << 32`. Any non-zero expiry (e.g. 3600 for 1 hour) gets interpreted as ~490,000 years, making tokens silently never expire.

**Fix:** In `foreign/go/internal/command/access_token.go`, change the field type to `uint64` and replace the `PutUint32` call with `PutUint64(bytes[1+len(c.Name):], c.Expiry)`.

saie-ch:
## Bugs Fixed:                                                                                                                                     
                                                                                                                                                     
  **Permissions.MarshalBinary()** - Fixed missing continuation flags between stream/topic entries. The serializer wasn't writing 1-byte flags        
  (1=has_next, 0=no_next) after each entry, causing the server to incorrectly parse permission data and potentially grant wrong permissions. Also    
  added `len() > 0` checks to match Rust SDK's behavior of skipping empty maps.                                                                      
                                                                                                                                                     


  **Note:** Does NOT fix #2980, #2981, #2982 (missing 4-byte permissions_len field before permissions data - those are separate issues found during the review of #2973).


foreign/go/contracts/users_test.go
```diff
@@ -0,0 +1,232 @@
...
+
+		if hasNextStream == 0 {
+			break
+		}
+	}
+
+	if streamsFound != 2 {
+		t.Errorf("Expected 2 streams, found %d", streamsFound)
+	}
+}
```
chengxilo, line 623: @saie-ch The test should verify that each attribute of Permissions is encoded correctly. Right now, those checks are being skipped. This only validates the basic frame structure (e.g., continuation flags and the size of each part). It would be better to also read back the encoded content and confirm that it matches the values you originally provided.

ryankert01: +1

foreign/go/contracts/users_test.go
chengxilo, line 48: I'm sorry that I keep asking you to modify your code.

However, your changes actually doesn't address the problem. If in the future I swaped the position of `ManageServers` and `ManageUsers` when encoding, this test won't know I am wrong.


To better desribe the problem, consider you have this struct.

```go
Foo {
    A bool,
    B bool,
    C bool,
}
```

You wrot a function to encode it.

```go
func (f Foo) MarshalBinary() ([]byte, error) {
    return []byte{
        boolToByte(f.A), 
        boolToByte(f.B), 
        boolToByte(f.C),
    }
}
```

You wrote a test and here is your test case:
```go
input: Foo {
    A: true,
    B: false,
    C: true
}
want: []byte {1, 0, 1}
```
And you can get a []byte {1, 0, 1}, you are 100% correct.

However, if I write the function like this:

```go
func (f Foo) MarshalBinary() ([]byte, error) {
    return []byte{
        boolToByte(f.C), 
        boolToByte(f.B), 
        boolToByte(f.A),
    }
}
```

I will still get []byte {1, 0, 1}, I can still pass the test, but actually it's wrong.

So you probably want to make sure every single field are correctly encoded.

Consider  make a bigger test case:
```go
{ 
    input: Foo {
        A: true
    },
    want: []byte{1,0,0}
},
{
    input: Foo {
        B: true
    },
    want: []byte{0,1,0}
},
{
    input: Foo {
        C: true
    },
    want: []byte{0,0,1}
}
```

chengxilo, line 71: same here

chengxilo (changes requested):

foreign/go/contracts/users_test.go
saie-ch: No problem @chengxilo, happy to modify until it's correct. Can you check now?

PR #3019, opened 2026-03-24
ci: add Helm chart validation

### LINKED ISSUE #3007: Add Helm chart CI validation
The chart at `helm/charts/iggy/` has zero CI coverage. The `helm` PR scope exists but no tests run.

Add a `helm` component to the existing change-detection + test pipeline. 2 files, ~50 lines.

`components.yml` - add helm component with `paths: ["helm/**"]`, `tasks: ["validate"]`. No `depends_on` (chart is independent of Rust build). Routes through existing `other_matrix` - no changes to `_detect.yml` or `pre-merge.yml`.

`_test.yml` - add inline step (same pattern as `web-ui`/`ci-workflows`) with:
- Pinned Helm install with SHA256 checksum (matching `yq` pattern in `_detect.yml`)
- `helm lint --strict`
- 6 `helm template` invocations covering all 10 templates and conditional branches:
  - default values (Deployment x2, Service x2, ServiceAccount, Secret)
  - all features enabled (+PVC, HPA, Ingress x2, ServiceMonitor)
  - legacy K8s 1.18 via `--kube-version` (Ingress v1beta1, HPA v2beta2 branches)
  - server only (`ui.enabled=false`)
  - UI only (`server.enabled=false`)
  - existing secret (`existingSecret.name` path vs `createSecret`)

CI cost: 1 job, ~40s, runs only when `helm/**` changes.

Future follow-ups (not this PR):
- kubeconform schema validation
- `values.schema.json`

avirajkhare00:
## Summary
- add a `helm` component to change detection so `helm/**` routes through the existing `other_matrix`
- add pinned Helm install, `helm lint --strict`, and the 6 requested render scenarios in `_test.yml`
- make the legacy ingress and HPA API selection deterministic by Kubernetes version so the 1.18 CI case really covers `networking.k8s.io/v1beta1` and `autoscaling/v2beta2`

Closes #3007

### DIFF
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
@@ -100,6 +100,7 @@ public class AsyncIggyTcpClient {
     private final int port;
     private final Optional<String> username;
     private final Optional<String> password;
+    private final Optional<Duration> connectionTimeout;
     private final Optional<Duration> acquireTimeout;
     private final Optional<Duration> requestTimeout;
     private final Optional<Integer> connectionPoolSize;
@@ -126,7 +127,7 @@ public class AsyncIggyTcpClient {
      * @param port the server port
      */
     public AsyncIggyTcpClient(String host, int port) {
-        this(host, port, null, null, null, null, null, null, false, Optional.empty());
+        this(host, port, null, null, null, null, null, null, null, false, Optional.empty());
     }
 
     @SuppressWarnings("checkstyle:ParameterNumber")
@@ -135,6 +136,7 @@ public class AsyncIggyTcpClient {
             int port,
             String username,
             String password,
+            Duration connectionTimeout,
             Duration acquireTimeout,
             Duration requestTimeout,
             Integer connectionPoolSize,
@@ -145,6 +147,7 @@ public class AsyncIggyTcpClient {
         this.port = port;
         this.username = Optional.ofNullable(username);
         this.password = Optional.ofNullable(password);
+        this.connectionTimeout = Optional.ofNullable(connectionTimeout);
         this.acquireTimeout = Optional.ofNullable(acquireTimeout);
         this.requestTimeout = Optional.ofNullable(requestTimeout);
         this.connectionPoolSize = Optional.ofNullable(connectionPoolSize);
@@ -176,7 +179,7 @@ public class AsyncIggyTcpClient {
         connectionPoolSize.ifPresent(poolConfigBuilder::setMaxConnections);
         acquireTimeout.ifPresent(timeout -> poolConfigBuilder.setAcquireTimeoutMillis(timeout.toMillis()));
         TCPConnectionPoolConfig poolConfig = poolConfigBuilder.build();
-        connection = new AsyncTcpConnection(host, port, enableTls, tlsCertificate, poolConfig);
+        connection = new AsyncTcpConnection(host, port, enableTls, tlsCertificate, poolConfig, connectionTimeout);
         return connection.connect().thenRun(() -> {
             log.debug("Connected to {}:{} | {}", host, port, IggyVersion.getInstance());
             messagesClient = new MessagesTcpClient(connection);
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
@@ -220,29 +220,61 @@ public final class AsyncIggyTcpClientBuilder {
      * @throws IggyInvalidArgumentException if the host is null or empty, or if the port is not positive
      */
     public AsyncIggyTcpClient build() {
-        if (host == null || host.isEmpty()) {
-            throw new IggyInvalidArgumentException("Host cannot be null or empty");
-        }
-        if (port == null || port <= 0) {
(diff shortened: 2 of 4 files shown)


avirajkhare00:
I dug through the hardcoded bits in this PR. The main problem is not the number of scenarios, it is that `_test.yml` is now carrying some chart-owned values that already exist under `helm/charts/iggy/`.

There are two buckets here:

1. Reasonable CI-owned hardcoding
- pinned Helm/kind versions and checksums
- kind cluster config
- ingress-nginx install
- smoke-test hosts / ingress class / namespace / release names

Those are properties of the CI environment and smoke path, so keeping them in the workflow is fine.

2. Chart-owned values that should ideally be reused from `helm/charts/iggy/`
- chart `version` / `appVersion` from `helm/charts/iggy/Chart.yaml`
- default image tags from `helm/charts/iggy/values.yaml`
- render scenario inputs that are really chart fixtures rather than CI wiring

That second bucket is where the PR feels too hardcoded, because it creates a second source of truth in CI.

If you want to keep this structure, I would suggest a cleaner split:
- leave CI/runtime setup in `_test.yml`
- move chart-specific test fixtures into the `helm/charts/iggy/` tree (for example dedicated values files under a small `ci/` or `tests/` subdir), or derive the asserted values from chart files directly

That keeps the Helm chart as the owner of chart data and makes version/tag bumps less noisy.

helm/charts/iggy/templates/hpa.yaml
```diff
@@ -38,24 +38,24 @@ spec:
     - type: Resource
       resource:
         name: cpu
-        {{- if .Capabilities.APIVersions.Has "autoscaling/v2" }}
+        {{ if semverCompare ">=1.23-0" .Capabilities.KubeVersion.GitVersion }}
```
hubcio: switching from `{{-` to `{{` on the if/else/end blocks here introduces blank lines in the rendered output:

```yaml
metrics:
  - type: Resource
    resource:
      name: cpu
      
      target:
        type: Utilization
        averageUtilization: 80
      
```

the blank lines between `name: cpu` and `target:` (and same for memory below) are because the template directives no longer trim whitespace. valid yaml, kubernetes accepts it, but it's sloppy. should use `{{-` trim markers like the original had. same issue at lines 45, 47, 53, 57, 59.

scripts/ci/test-helm.sh
```diff
@@ -0,0 +1,418 @@
...
+    className: ${HELM_SMOKE_INGRESS_CLASS}
+    hosts:
+      - host: ${HELM_SMOKE_SERVER_HOST}
+        paths:
+          - path: /
+            pathType: Prefix
+  env:
+    - name: RUST_LOG
+      value: info
+    - name: IGGY_HTTP_ADDRESS
```
hubcio, line 273: `smoke()` reads `web/package.json` via python3 to get the UI image tag, but `validate()` already has `extract_values_tag` that pulls tags from `values.yaml`. using two different sources of truth for the same value is fragile - if someone updates `values.yaml` but not `web/package.json` (or vice versa), the smoke test will use a different tag than what validate checks. consider reusing `extract_values_tag ui` here instead, or at minimum document why the two intentionally differ.

scripts/ci/test-helm.sh
```diff
@@ -0,0 +1,418 @@
...
+    leftover_resources="$(
+      kubectl -n "$HELM_SMOKE_NAMESPACE" get deployment,service,ingress,secret,serviceaccount \
+        -l "app.kubernetes.io/instance=${HELM_SMOKE_RELEASE}" \
+        -o name 2>/dev/null || true
+    )"
+    if [ -n "$leftover_resources" ]; then
+      cat >&2 <<EOF
+Found leftover resources for a failed Helm smoke install in namespace '${HELM_SMOKE_NAMESPACE}'.
+Run 'scripts/ci/test_helm.sh cleanup-smoke' once, then rerun the smoke test.
+EOF
```
hubcio, line 248: since this script is meant to be run locally by devs (per the README instructions), the `smoke` command should either auto-cleanup on success or the script should accept a `--cleanup` flag. right now a dev who runs `test_helm.sh smoke` locally is left with a dangling kind cluster + namespace + helm release that they have to know to clean up via `cleanup-smoke`. in CI this doesn't matter (ephemeral runners), but locally it leaks resources.

could be as simple as calling `cleanup_smoke` at the end of a successful `smoke()` run, or adding a trap.

scripts/ci/test-helm.sh
hubcio: just to add.. we have `extract_version.sh`
```sh
❯ scripts/extract-version.sh web-ui
0.2.1-edge.1

❯ scripts/extract-version.sh web-ui --tag
web-ui-0.2.1-edge.1
```

scripts/ci/test-helm.sh
```diff
@@ -0,0 +1,418 @@
...
+    print(json.load(f)["version"])
+PY
+  )"
+
+  echo "$ui_image_tag" > "$HELM_SMOKE_REPORT_DIR/ui-image-tag.txt"
+
+  smoke_values_file="$(mktemp "${TMPDIR:-/tmp}/iggy-helm-smoke-values.XXXXXX.yaml")"
+
+  cat > "$smoke_values_file" <<EOF
+server:
```
atharvalade, line 261: Smoke and validate can disagree on which UI image tag is “canonical”: at this revision `values.yaml` has `ui.image.tag: "edge"` while web/package.json has `"0.2.1-edge.1"`. I checked Docker Hub: `apache/iggy-web-ui:0.2.1-edge.1` exists, so smoke is not guaranteed to fail today. The critical angle is future CI breakage if npm version and published image tags drift apart worth aligning with `extract_values_tag` ui, `scripts/extract-version.sh`, or documenting why smoke must override the chart default.

hubcio (dismissed):

PR #3020, opened 2026-03-24
feat(partitions): rework partitions using new message

numinnex:
Refactors partition messaging around Message<H, B>, where the backing storage is part of the type rather than
  being encoded through the old TwoHalves prefix/tail layout.

  Request messages now use a contiguous Owned backing, while response-style paths can use fragmented storage
  directly. This removes the old prefix/tail copy-on-write design entirely, along with the aliasing and uniqueness
  problems it introduced. In particular, storage reads no longer need to reconstruct a request-shaped message from
  a split buffer, which was fragile once the tail could already be shared by cache-backed Frozen values. The new
  design makes the ownership model explicit, keeps the request path simple, and avoids relying on prefix
  detachment semantics to preserve correctness.

### DIFF
--- a/core/integration/tests/connectors/postgres/postgres_source.rs
+++ b/core/integration/tests/connectors/postgres/postgres_source.rs
@@ -260,12 +260,6 @@ async fn delete_after_read_source_removes_rows_after_producing(
             .await;
     }
 
-    let initial_count = fixture.count_rows(&pool).await;
-    assert_eq!(
-        initial_count, TEST_MESSAGE_COUNT as i64,
-        "Expected {TEST_MESSAGE_COUNT} rows before processing"
-    );
-
     let stream_id: Identifier = seeds::names::STREAM.try_into().unwrap();
     let topic_id: Identifier = seeds::names::TOPIC.try_into().unwrap();
     let consumer_id: Identifier = "test_consumer".try_into().unwrap();


core/partitions/src/iggy_partitions.rs
```diff
@@ -502,7 +562,7 @@ where
             // TODO: Propagate send error instead of panicking; requires bus error design.
             consensus
                 .message_bus()
-                .send_to_client(prepare_header.client, generic_reply)
+                .send_to_client(prepare_header.client, reply_buffers)
                 .await
                 .unwrap();
```
hubcio: `send_to_client` does a `HashMap::get` on the client id (`message_bus/src/lib.rs:123`). if `remove_client` was called between prepare and commit (normal TCP disconnect during VSR quorum round-trips), this returns `Err(ClientNotFound)` and `.unwrap()` crashs the shard. the TODO at line 562 acknowledges this. replace with `if let Err(e) = ... { warn!(...); }` - client disconnects happen in normal operation.

core/partitions/src/iggy_partitions.rs
```diff
@@ -750,41 +848,32 @@ where
...
+            .and_then(|writer| writer.as_ref())
+            .cloned()
+            .expect("Sparse index writer not initialized");
 
         let saved = messages_writer
-            .as_ref()
-            .save_frozen_batches(&frozen_batches)
+            .save_frozen_batches(&stripped_batches)
             .await
             .expect("persist: failed to save frozen batches");
```
hubcio: also at line 878 (index write).

`persist_frozen_batches_to_disk` panics on disk write failure. disk full, permission errors, and transient I/O failures are normal runtime conditions that should degrade gracefully, not crash the shard. this function returns `()` but should propagate errors - the caller (`commit_messages`) can then decide whether to retry, skip, or escalate.

core/partitions/src/iggy_partitions.rs
```diff
@@ -847,20 +931,43 @@ where
         )
         .await
         .expect("Failed to create segment storage");
```
hubcio: also at lines 947, 956.

`rotate_segment` panics on filesystem failures (creating segment storage, opening writers). same concern as the persist path - these are runtime operations triggered when a segment fills up. a full disk during rotation would crash the shard instead of pausing writes. should propagate errors.

core/partitions/src/send_messages2.rs
```diff
@@ -0,0 +1,633 @@
...
+        return Err(IggyError::InvalidCommand);
+    }
+
+    let header_offset = std::mem::size_of::<PrepareHeader>();
+    let header_bytes = &mut prefix[header_offset..header_offset + COMMAND_HEADER_SIZE];
+    let mut command = SendMessages2Header::decode(header_bytes)?;
+    command.base_offset = base_offset;
+    command.base_timestamp = base_timestamp;
+    command.encode_into(header_bytes);
+    let message_count = count_messages_blob(&tail)?;
```
hubcio: also at `journal.rs:357` (append), `journal.rs:421` (select_batch_slice), `iggy_partitions.rs:772` (commit loop).

`count_messages_blob` walks every message header in the blob, parsing each one. the same blob is scanned 3-4 times through its lifecycle: stamp, append, commit, and poll. `SendMessages2Header` uses 40 of its 256-byte `COMMAND_HEADER_SIZE` allocation - 216 bytes of padding sit unused. storing `message_count` as a `u32` at offset 40 would eliminate all re-scanning.

core/partitions/src/iggy_partitions.rs
```diff
@@ -689,23 +752,61 @@ where
...
+                        flush_index = Some(index);
+                    }
+                    file_position += batch.header.total_size() as u64;
+                    batch_count += message_count;
+                    frozen.push(owned.into());
+                }
+
+                (
+                    frozen,
+                    flush_index
```
hubcio: if every journal entry fails `decode_prepare_slice` (the `continue` branches at lines 769-777), `flush_index` remains `None` and `.expect()` panics. the guard at line 721 (`messages_count == 0`) prevents this when the journal is truly empty, but memory corruption or `try_merge` failure during the commit loop could trigger the panic despite a nonzero journal. handle `flush_index == None` by returning early or logging a critical warning instead of panicking.

core/partitions/src/iggy_index_writer.rs
```diff
@@ -0,0 +1,83 @@
...
+        (&*file)
+            .write_all_at(indexes, position)
+            .await
+            .0
+            .map_err(|_| IggyError::CannotSaveIndexToSegment)?;
+
+        self.index_size_bytes
+            .fetch_add(len as u64, Ordering::Release);
+
+        self.fsync().await?;
```
hubcio: `MessagesWriter` stores an `fsync: bool` field and conditionally calls `self.fsync()` (messages_writer.rs:90-92). `IggyIndexWriter` has no `fsync` field and unconditionally fsyncs here. when `enforce_fsync` is false, message data writes skip fsync but every index write pays the `sync_all` penalty. add an `fsync: bool` field matching `MessagesWriter`.

core/partitions/src/iggy_partition.rs
```diff
@@ -177,63 +137,56 @@ impl Partition for IggyPartition {
...
         args: PollingArgs,
-    ) -> Result<IggyMessagesBatchSet, IggyError> {
+    ) -> Result<PolledBatches<4096>, IggyError> {
         if !self.should_increment_offset || args.count == 0 {
-            return Ok(IggyMessagesBatchSet::empty());
+            return Ok(PolledBatches::empty());
         }
 
-        let committed_offset = self.offset.load(Ordering::Relaxed);
+        let write_offset = self.dirty_offset.load(Ordering::Relaxed);
```
hubcio: `write_offset` uses `dirty_offset` (prepared but uncommitted). consumers can read data that hasn't reached quorum ack. for single-node this is fine, but under VSR a view change could roll back preared ops that consumers already read. the doc comment on `PartitionOffsets` (types.rs:130) says messages with `offset > commit_offset` are "invisible" - this contradicts the polling behavior. either use `self.offset` (committed) here, or document that this is intentional "read uncommitted" semantics.

core/partitions/src/iggy_partition.rs
```diff
@@ -111,29 +80,25 @@ impl Partition for IggyPartition {
...
+        if batch_messages_count == 0 {
+            return Ok(AppendResult::new(0, 0, 0));
+        }
 
-        let batch_messages_count = batch.count();
-        let batch_messages_size = batch.size();
+        let batch_messages_size =
+            u32::try_from(batch.total_size()).map_err(|_| IggyError::CannotAppendMessage)?;
 
         let last_dirty_offset = if batch_messages_count == 0 {
```
hubcio: this `if batch_messages_count == 0` branch is unreachable- the early return at lines 96-98 already handles that case. the `else` branch at line 106 always executes.

core/common/src/types/send_messages2.rs
```diff
@@ -0,0 +1,633 @@
...
+            id: read_u128(bytes, 8)?,
+            origin_timestamp: read_u64(bytes, 40)?,
+            user_headers: &bytes[payload_end..headers_end],
+            payload: &bytes[payload_start..payload_end],
+            total_size,
+        })
+    }
+}
+
+fn calculate_checksum_parts(header_tail: &[u8], user_headers: &[u8], payload: &[u8]) -> u64 {
```
hubcio, line 713: allocates a `BytesMut` and copies all three slices just to compute a hash. `xxhash3_64::Hasher` implements `std::hash::Hasher` with streaming `write()` + `finish()` - three `write()` calls would be zero-alloc. this is called per message during `from_legacy_request` which is the ingestion hot path (every `SendMessages` request goes through `convert_request_message`).

core/partitions/src/journal.rs
```diff
@@ -335,18 +346,29 @@ where
...
-
         let header = *entry.header();
         let op = header.op;
+        let owned = unsafe { entry.into_inner().try_merge() }
+            .expect("partition journal append expects a unique message buffer");
+        let first_offset_and_timestamp = if header.operation == Operation::SendMessages {
+            decode_prepare_slice(owned.as_slice())
+                .ok()
+                .and_then(|batch| {
+                    batch.message_count().ok().and_then(|count| {
```
hubcio: `batch.message_count()` re-scans the entire blob to count messages (see `count_messages_blob` in send_messages2.rs). this is the second scan of the same blob - the first happened in `stamp_prepare_for_persistence` moments before this `append` call. see comment on send_messages2.rs:512 for the fix (store count in header padding).

core/partitions/src/journal.rs
```diff
@@ -401,3 +411,85 @@ where
...
     }
 }
+
+fn select_batch_slice(
+    batch: &crate::send_messages2::SendMessages2Ref<'_>,
+    query: MessageLookup,
+    already_matched: u32,
+) -> Option<SelectedBatchSlice> {
+    let remaining = query.count().saturating_sub(already_matched);
+    let batch_message_count = batch.message_count().ok()?;
```
hubcio: same repeated `message_count()` blob scan as noted at journal.rs:357 - this one fires on every `select_batch_slice` call during polling. with the count stored in the header, this becomes a constant-time read.

core/partitions/src/iggy_partitions.rs
```diff
@@ -350,6 +398,17 @@ where
...
+                    warn!(?namespace, %error, "on_request: failed to convert SendMessages");
+                    return;
+                }
+            }
+        } else {
+            message
+        };
         let prepare = message.project(consensus);
         pipeline_prepare_common(consensus, prepare, |prepare| self.on_replicate(prepare)).await;
     }
```
hubcio, line 435: re: line 449 - `apply_replicated_operation` is called unconditionally, even when `fence_old_prepare_by_commit` returned `true` at line 434 (which only gates `self.replicate()`, not apply). `append_messages` is not idempotent - it assigns a new `dirty_offset` and fresh `batch_timestamp` on every call. after a VSR view change, the new primary replays from `max_commit+1` and re-executes, producing duplicate partition data with different offsets.

the comment at lines 487-496 acknowledges this and lists two required guards:
1. op-based dedup: check `journal.op_to_storage_offset.contains_key(&header.op)` before appending
2. client reply dedup by `(client_id, request_id)`

the `op_to_storage_offset` BTreeMap already exists - the dedup check is straightforward.

core/partitions/src/journal.rs
```diff
@@ -373,26 +395,14 @@ where
```
hubcio, line 430: re: line 387 - `BTreeMap::insert` at `timestamp_to_op.insert(timestamp, op)` overwrites the previous value. two batches arriving in the same microsecond get the same `base_timestamp` (assigned from `IggyTimestamp::now().as_micros()` at iggy_partition.rs:91). the earlier batch's op mapping is silently lost.

`candidate_start_op` for timestamp queries uses `range(..=timestamp).next_back()` - if the overwritten mapping pointed to an earlier op, a timestamp query could start iteration past it, making the earlier batch unreachable via timestamp-based polling.

offset-based queries are unaffected (keyed by unique offset). fix with composite key `(timestamp, op)` or `BTreeMap<u64, SmallVec<[u64; 1]>>`.

core/common/src/types/send_messages2.rs
```diff
@@ -0,0 +1,633 @@
...
+        Ok(Self {
+            partition_id: read_u64(bytes, 0)?,
+            base_offset: read_u64(bytes, 8)?,
+            base_timestamp: read_u64(bytes, 16)?,
+            origin_timestamp: read_u64(bytes, 24)?,
+            batch_length,
+        })
+    }
+
+    pub fn encode_into(&self, bytes: &mut [u8]) {
```
hubcio, line 84: per-message checksums (line 155) cover the message header tail + user_headers + payload, but `base_offset` and `base_timestamp` in the batch header are not covered by any checksum. a single bit flip in `base_offset` silently corrupts every message's absolute offset (since per-message offsets are delta-encoded from it), and no checksum would detect it.

add a batch-level CRC32C covering the full batch header (bytes 0-39 at minimum) before the first message. particularly important because the batch header is mutated in-place during `stamp_prepare_for_persistence` (line 509-511) - any bug in that stamping path would also go undetected.

core/partitions/src/iggy_partitions.rs
```diff
@@ -847,20 +931,43 @@ where
...
+
+        partition.log.add_persisted_segment(
+            segment,
+            storage,
+            Some(messages_writer),
+            Some(index_writer),
+        );
         partition.stats.increment_segments_count(1);
 
         debug!(?namespace, start_offset, "rotated to new segment");
```
hubcio: re: `send_prepare_ok` at line 976 - the TODO at line 980 acknowledges the journal isn't verified durable before acking. in VSR, a replica must persist the prepare entry before sending PrepareOk - otherwise a power loss after ack but before disk flush means the replica promised durability it can't deliver.

concrete failure: client told "committed" after quorum ack, power kills 2 of 3 replicas before disk flush, data is gone. the fix is either per-prepare fdatasync (correct but expensive) or batched fsync with a flush-before-ack barrier.

core/partitions/src/iggy_partitions.rs
```diff
@@ -67,6 +73,12 @@ pub struct IggyPartitions<C> {
...
 
+fn freeze_client_reply(message: Message<GenericHeader>) -> ClientBuffers {
+    let owned = unsafe { message.into_inner().try_merge() }
+        .expect("client reply expects a mergeable message buffer");
+    vec![owned.into()]
+}
+
 impl<C> IggyPartitions<C> {
     #[must_use]
     pub fn new(shard_id: ShardId, config: PartitionsConfig) -> Self {
```
hubcio, line 88: re: `partitions_mut()` at line 116 - returns `&mut Vec<IggyPartition>` from `&self` via `UnsafeCell`. the `#[allow(clippy::mut_from_ref)]` suppresses the lint but doesn't make it sound - under Rust's aliasing rules, creating multiple `&mut` references to the same data is UB regardless of threading model.

current usage is sequential (traced all call sites - borrows don't overlap across `.await` points), so it works in practice. but the API permits aliasing, and any future code adding a second `get_mut_by_ns` call while a first `&mut` is alive would silently introduce UB.

consider `RefCell` for runtime borrow checking (at least in debug builds) or restructure so mutation requires `&mut self`.

core/partitions/src/journal.rs
```diff
@@ -335,18 +346,29 @@ where
     }
 
     async fn append(&self, entry: Self::Entry) {
-        let first_offset_and_timestamp = Self::message_to_batch(&entry)
-            .and_then(|batch| Some((batch.first_offset()?, batch.first_timestamp()?)));
-
         let header = *entry.header();
         let op = header.op;
+        let owned = unsafe { entry.into_inner().try_merge() }
```
hubcio: also at `iggy_partitions.rs:766` (commit loop).

`try_merge()` reconstitutes `Owned` from `(Prefix, Frozen)` on every append and every commit entry. the fields being read (`base_offset` at byte 264, `base_timestamp` at byte 272) are within the 512-byte prefix - they could be read directly without merging.

in the common case (same control block, refcount==2) the merge is just atomic ops + pointer gymnastics, but when the prefix was COW-detached it triggers a `copy_from_slice` of up to 512 bytes. redesigning the journal to store `(Prefix, Frozen)` pairs and decode from the split representation would eliminate this.

core/common/src/types/send_messages2.rs
```diff
@@ -0,0 +1,633 @@
...
+
+        let mut blob = BytesMut::with_capacity(messages.len());
+        for (index, legacy) in parsed.iter().enumerate() {
+            let id = if legacy.id == 0 {
+                random_id::get_uuid()
+            } else {
+                legacy.id
+            };
+            let offset_delta = u32::try_from(index).map_err(|_| IggyError::InvalidCommand)?;
+            let timestamp_delta = legacy
```
hubcio, line 225: `timestamp_delta` is `u32` representing microseconds delta from `origin_timestamp`. `u32::MAX` = ~4295 seconds = ~71.6 minutes. a batch spanning more than ~72 minutes of client-side `origin_timestamp` is rejected with a generic `IggyError::InvalidCommand` at line 142.

this is a wire format constant - once deployed, it can't be changed without protocol versioning. the error is graceful (not silent), but the limit is undocumented and the error gives zero indication of what went wrong. at minimum document the ~72 minute maximum batch span on `timestamp_delta`, and return a more specific error variant.

core/partitions/src/journal.rs
```diff
@@ -401,3 +411,85 @@ where
...
+    query: MessageLookup,
+    already_matched: u32,
+) -> Option<SelectedBatchSlice> {
+    let remaining = query.count().saturating_sub(already_matched);
+    let batch_message_count = batch.message_count().ok()?;
+    if remaining == 0 || batch_message_count == 0 {
+        return None;
+    }
+
+    if matches!(query, MessageLookup::Timestamp { timestamp, .. } if batch.header.base_timestamp < timestamp)
```
hubcio: the timestamp filter at lines 426-429 skips the entire batch if `batch.header.base_timestamp < timestamp`, then line 444 returns `true` for all remaining messages without per-message timestamp filtering. messages within a kept batch may have `origin_timestamp` (via `timestamp_delta`) earlier than the query timestamp.

for offset-based queries this doesn't matter (offset_delta is monotonic within a batch). but for timestamp queries, this can return messages with timestamps earlier than requested.

core/partitions/src/iggy_partitions.rs
```diff
@@ -688,104 +678,258 @@ where
 
         if is_full || unsaved_messages_count_exceeded || unsaved_messages_size_exceeded {
             // Freeze journal batches.
-            let frozen_batches = {
-                let batches = partition.log.journal_mut().inner.commit();
+            let (frozen_batches, index_bytes, batch_count) = {
+                let entries = partition.log.journal().inner.entries();
```
hubcio: also at lines 745-746.

`entries()` clones a snapshot of the journal here, then `persist_frozen_batches_to_disk().await` yields to the compio executor at line 729. during the yield, the executor can poll another ready future - an incoming prepare triggers `on_replicate` -> `append_messages`, which pushes a new entry to this partition's journal. when persist completes, `commit()` at line 745 drains ALL entries via `std::mem::take` (including the one appended during the await), and `journal_info` is reset to default at line 746. the entry appended during persist was never written to disk, is now gone from the journal, and no error is logged.

this is silent permanent data loss during normal operation under load - no crash required.

fix: replace `entries()` (clone) + `commit()` (drain-all) with a single atomic `drain()` before the async persist. new appends during persist go into a fresh journal and are committed in the next cycle.

core/partitions/src/iggy_partitions.rs
```diff
@@ -688,104 +678,258 @@ where
...
+        partition
+            .stats
+            .increment_size_bytes(journal_info.size.as_bytes_u64());
+        partition
+            .stats
+            .increment_messages_count(u64::from(journal_info.messages_count));
+
+        // 4. Advance committed offset (last, so consumers only see offset after data is durable).
         let committed_offset = journal_info.current_offset;
         partition.offset.store(committed_offset, Ordering::Relaxed);
```
hubcio: the comment at line 772 says "consumers only see offset after data is durable" - but this line runs unconditionally, even when the flush-threshold condition at line 679 is false and the persist block (lines 680-747) is skipped entirely. when thresholds aren't met, data exists only in the in-memory journal but the committed offset tells consumers it's available. crash before the next flush cycle = silent data loss for messages consumers believe they successfully processed.

combined with follower `commit_journal` being a no-op (line 648), data can be "committed" without being durable on any replica.

fix: only advance `partition.offset` after data is persisted, or restructure so the offset tracks what's actually durable.

core/partitions/src/iggy_partitions.rs
```diff
@@ -542,7 +548,7 @@ where
         &self,
```
hubcio, line 619: also at the comment block lines 488-498.

`apply_replicated_operation` unconditionally calls `append_send_messages_to_journal` with no dedup check. the comment at line 491 explicitly says `append_messages` is NOT idempotent and lists two guards required before production traffic:
1. op-based dedup: check `journal.op_to_storage_offset.contains_key(&header.op)` before appending
2. client reply dedup by `(client_id, request_id)`

the `op_to_storage_offset` BTreeMap already exists in journal.rs:114 and is populated on every append - the dedup check is a one-liner. after a VSR view change, the new primary replays from `max_commit+1` and re-executes, producing duplicate partition data with different offsets.

core/partitions/src/iggy_partitions.rs
```diff
@@ -386,7 +439,16 @@ where
         // TODO: Figure out the flow of the partition operations.
         // In metadata layer we assume that when an `on_request` or `on_replicate` is called, it's called from correct shard.
         // I think we need to do the same here, which means that the code from below is unfallable, the partition should always exist by now!
-        self.apply_replicated_operation(&namespace, message).await;
+        if let Err(error) = self.apply_replicated_operation(&namespace, message).await {
```
hubcio, line 482: when `apply_replicated_operation` fails here, `on_replicate` early-returns at line 450 without calling `send_prepare_ok`. but by this point:
- line 432: prepare already forwarded to downstream replicas via `replicate_to_next_in_chain`
- line 436: sequencer already advanced
- line 437: checksum chain already updated

downstream replicas process the prepare and send PrepareOk, but this replica doesn't. result: partial quorum that can never complete - the primary's pipeline entry for this op is stuck forever, blocking the namespace. no timeout, no retry, no NACK.

worse: the sequencer advancement at line 436 means subsequent ops (N+1, N+2...) are accepted while there's a gap at op N, violating the sequential execution invariant.

core/partitions/src/iggy_partitions.rs
```diff
@@ -553,17 +559,18 @@ where
...
                     op = header.op,
                     ?namespace,
                     "on_replicate: send_messages appended to partition journal"
                 );
+                Ok(())
             }
             Operation::StoreConsumerOffset => {
-                let body = message.body_bytes();
-                let body = body.as_ref();
+                let total_size = header.size() as usize;
```
hubcio, line 646: also at lines 574-575.

`body[0]`, `body[1..5]`, `body[5..13]` are unchecked indexing and the `.try_into().unwrap()` calls will panic on malformed input. in a consensus system, a malformed `StoreConsumerOffset` prepare that achieves quorum would crash every replica that processes it (f+1 replicas panic = total cluster crash from a single message).

fix: use `body.get(0..13)` with an early return on `None`, and replace `.unwrap()` with `.map_err()`.

core/partitions/src/iggy_partitions.rs
```diff
@@ -688,104 +678,258 @@ where
...
+                warn!(
+                    client = prepare_header.client,
+                    op = prepare_header.op,
+                    %error,
+                    "on_ack: failed to send reply to client"
+                );
+            }
+        }
+
+        if failed_ns.is_empty() {
```
hubcio: if any namespace fails to commit, `handle_committed_entries` returns `false` here, which causes `on_ack` (line 508) to skip `advance_commit_number` entirely. this blocks global commit advancement for ALL namespaces on the shard - one poisoned partition halts the entire shard's consensus progress. the pipeline fills up after ~8 more ops (PIPELINE_PREPARE_QUEUE_MAX), stalling all partitions.

client replies are still sent inside the loop (lines 836-855), so clients see success while the shard is silently dying.

this undermines the per-namespace independence that `drain_committable_all` was designed to provide.

core/partitions/src/iggy_partitions.rs
```diff
@@ -618,22 +627,22 @@ where
...
-    async fn replicate(&self, message: Message<PrepareHeader>) {
+    async fn replicate(&self, message: Message<PrepareHeader>) -> Message<PrepareHeader> {
         let consensus = self
             .consensus()
             .expect("replicate: consensus not initialized");
-        replicate_to_next_in_chain(consensus, message).await;
+        replicate_to_next_in_chain(consensus, message).await
     }
 
     #[allow(clippy::unused_self, clippy::missing_const_for_fn)]
```
hubcio: followers replicate data via `on_replicate` -> `append_messages` (in-memory journal only) but this function never persists anything. failover to a follower = complete data loss for that replica. combined with the committed-offset-without-durability issue at line 774, the system has no durable state on any replica when flush thresholds aren't met.

core/partitions/src/iggy_partitions.rs
```diff
@@ -322,11 +333,41 @@ impl<C> IggyPartitions<C> {
         )
         .await
         .expect("Failed to create segment storage");
```
hubcio: also at lines 339, 349, 359.

four `.expect()` calls in `init_partition` will crash the shard on any filesystem error (disk full, permission denied, path not found). `rotate_segment` was correctly updated to return `Result` with `map_err` - this function should follow the same pattern.

core/partitions/src/iggy_partitions.rs
```diff
@@ -845,33 +985,63 @@ where
...
+            Some(messages_writer),
+            Some(index_writer),
+        );
         partition.stats.increment_segments_count(1);
 
         debug!(?namespace, start_offset, "rotated to new segment");
+        Ok(())
     }
 
     async fn send_prepare_ok(&self, header: &PrepareHeader) {
```
hubcio, line 1311: `send_prepare_ok` checks `partition.log.journal().inner.entry(header).await.is_some()` which verifies the entry exists in the in-memory journal - not that it's durable on disk. the journal backing is `PartitionJournalMemStorage` (a `Vec` behind `UnsafeCell`). in VSR, a replica sending PrepareOk promises durability. power loss after ack but before the next flush cycle means the replica promised durability it can't deliver.

core/partitions/src/segment.rs
```diff
@@ -0,0 +1,85 @@
...
+use iggy_common::{IggyByteSize, IggyExpiry, IggyTimestamp};
+use std::fmt::Display;
+
+#[derive(Default, Debug, Clone)]
+pub struct Segment {
+    pub sealed: bool,
+    pub start_timestamp: u64,
+    pub end_timestamp: u64,
+    pub max_timestamp: u64,
+    pub current_position: u32,
```
hubcio: `current_position` is `u32` but `max_size` is `IggyByteSize` (u64-backed). `current_position` is incremented in the prepare path (iggy_partition.rs:112) with `+= batch_messages_size` before the commit path checks `is_full()`. multiple prepares can accumulate past 4GB before a commit triggers rotation, causing silent overflow (wraps in release mode, panics in debug).

core/partitions/src/journal.rs
```diff
@@ -383,44 +406,116 @@ where
...
+    let full_body_selected = selection.start == 0 && selection.end == batch.blob().len();
+
+    if full_body_selected {
+        fragments.push(Fragment::slice(
+            prepare.clone(),
+            prepare_header_size,
+            prepare_size,
+        ));
+    } else {
+        let mut rewritten = batch.header;
```
hubcio, line 518: also at lines 508-511.

when a partial batch slice is returned (the `else` branch), `rewritten` copies the full `SendMessages2Header` including `batch_checksum` and `message_count`, then only `batch_length` is overwritten. the serialized partial batch header retains the original full-batch checksum and count, which are incorrect for the sliced data. any consumer that validates checksums on poll responses will reject these fragments.

fix: set `rewritten.message_count = selection.matched_messages` and recompute `batch_checksum` over the new header + sliced blob.

core/partitions/src/iggy_partitions.rs
```diff
@@ -656,136 +703,276 @@ where
...
+            || is_full
+            || unsaved_messages_count_exceeded
+            || unsaved_messages_size_exceeded;
+
+        if !should_persist {
+            return Ok(());
+        }
 
-        if is_full || unsaved_messages_count_exceeded || unsaved_messages_size_exceeded {
+        if is_full
```
hubcio: also at lines 726-728.

lines 716-719 compute `should_persist` and return early at 721-723 when false. this `if` block re-checks the exact same four conditions - it's always true at this point. looks like a refactoring artifact from when the early return was added. the inner `if` can be removed (keep its body).

core/partitions/src/iggy_partitions.rs
numinnex: I've left a `TODO` comment to tackle that with another issue, essentially we would like the `Partition` to be re-entrant, thus never requiring `&mut` access to it.

PR #3023, opened 2026-03-24
feat(consensus): add ClientTable with WAL-backed commit path and view-change safety

### LINKED ISSUE #3022: Clients table implementation from VSR paper, for sending the response
Create an ClientsTable abstraction, that would register async callbacks receiving reponse for particular request id.

krishvishal:
## Summary

Implements the VSR client-table (VR Revisited, Section 4, Figure 2) with a WAL-backed commit path. Both metadata and partitions planes now execute committed ops through the same path on all replicas, ensuring the client table is always populated for view-change correctness.

- **ClientTable** (`client_table.rs`): fixed-size slot array with `HashMap` secondary index for O(1) lookups. Tracks pending requests via `Notify` primitive for async commit notification. Deterministic eviction by lowest commit number.

- **PrepareJournal** (renamed from `MetadataJournal`): reusable append-only WAL for consensus prepare messages, now shared by both planes.

- **WAL-backed commit path**: operations are applied at commit time (not replicate time), eliminating duplicate data after view-change re-execution (IGGY-66). Both `commit_journal` (backup) and `on_ack` (primary) read from the WAL, apply the state machine, flush, advance `commit_min`, and update the client table - per-op, in a single loop.

- **`commit_min` / `commit_max` split**: `commit_max` tracks what the cluster says is committed (advances immediately). `commit_min` tracks what this replica has actually executed (advances sequentially, +1 per op). DVC
messages use `commit_min`. Fence and chain-replication use `commit_min`. Enables future message repair without data loss.

## Additional fixes

- `build_reply_message` uses `prepare_header.op` (not `commit_max`) for deterministic eviction ordering across replicas
- `fence_old_prepare_by_commit` uses `commit_min` so retransmitted prepares needed for gap repair are not fenced out
- `handle_start_view` returns range-based `SendPrepareOk` - caller must verify WAL presence before sending (prevents false acks)
- `drain_committable_prefix` replaces `drain_committable_all` for strict global op ordering (prevents `advance_commit_min` invariant violation)
- Sequencer/checksum updated after WAL append (not before) - failed append no longer leaves consensus state inconsistent
- `commit_reply` asserts: `client_id != 0`, `client_id == header.client`, commit monotonicity, request monotonicity (all hard asserts)
- `advance_commit_min` sequential invariant is a hard assert (not debug-only)
- `checkpoint_if_needed` uses `commit_min` (not `commit_max`) - prevents draining unexecuted WAL entries
- `commit_messages` called per-op (no dedup) - each apply's data flushed before advancing `commit_min`
- `ns_commits` reset in `clear()` - prevents stale view-change state
- `IggyPartitions<C, J>` - partitions plane now generic over journal type

  Closes #3022

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -15,11 +15,21 @@
 # specific language governing permissions and limitations
 # under the License.
 
+[profile.default]
+slow-timeout = { period = "30s", terminate-after = 4 }
+
 [[profile.default.overrides]]
 # This is a solution (or actually a workaround) for the problem that nextest does not support
 # #[serial] macro which shall enforce sequential execution of the test case.
 filter = 'package(integration) and test(cli::system::test_cli_session_scenario::should_be_successful)'
 threads-required = "num-cpus"
 
-[profile.default]
-slow-timeout = { period = "30s", terminate-after = 4 }
+# shared_server tests: the iggy_harness macro prefixes test names with
+# shared_server_ enabling a single generic nextest filter. Bounded to limit
+# resource contention from parallel fixture containers without fully serializing.
+[test-groups.shared-server]
+max-threads = 8
+
+[[profile.default.overrides]]
+filter = 'package(integration) and test(/shared_server_/)'
+test-group = 'shared-server'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6151,6 +6151,7 @@ dependencies = [
  "configs",
  "configs_derive",
  "ctor",
+ "dashmap",
  "figment",
  "futures",
  "harness_derive",
--- a/core/harness_derive/src/attrs.rs
+++ b/core/harness_derive/src/attrs.rs
@@ -48,6 +48,12 @@ impl ClusterNodesValue {
     }
 }
 
+/// Shared server configuration.
+#[derive(Debug, Clone)]
+pub struct SharedServerAttrs {
+    pub key: String,
+}
+
 /// Parsed `#[iggy_harness(...)]` attributes.
 #[derive(Debug, Default)]
 pub struct IggyTestAttrs {
@@ -57,6 +63,7 @@ pub struct IggyTestAttrs {
     pub server: ServerAttrs,
     pub seed_fn: Option<syn::Path>,
     pub cluster_nodes: ClusterNodesValue,
+    pub shared_server: Option<SharedServerAttrs>,
 }
 
 /// MCP configuration attributes.
@@ -82,6 +89,7 @@ impl IggyTestAttrs {
             server: ServerAttrs::default(),
             seed_fn: None,
             cluster_nodes: ClusterNodesValue::None,
+            shared_server: None,
         }
     }
 }
@@ -248,6 +256,9 @@ impl Parse for IggyTestAttrs {
                 AttrItem::ClusterNodes(cluster) => {
                     attrs.cluster_nodes = cluster;
                 }
+                AttrItem::SharedServer(shared) => {
+                    attrs.shared_server = Some(shared);
+                }
             }
         }
 
@@ -264,6 +275,7 @@ enum AttrItem {
     Server(Box<ServerAttrs>),
     Seed(syn::Path),
     ClusterNodes(ClusterNodesValue),
+    SharedServer(SharedServerAttrs),
 }
 
 impl Parse for AttrItem {
@@ -292,6 +304,10 @@ impl Parse for AttrItem {
                 input.parse::<Token![=]>()?;
                 let path: syn::Path = input.parse()?;
                 Ok(AttrItem::Seed(path))
+            }
+            "shared_server" => {
+                let shared = parse_shared_server_attrs(input)?;
+                Ok(AttrItem::SharedServer(shared))
             }
             _ => Err(syn::Error::new(
                 ident.span(),
@@ -471,6 +487,13 @@ fn parse_mcp_attrs(input: ParseStream) -> syn::Result<McpAttrs> {
     }
 
     Ok(mcp)
+}
+
+/// Parse `shared_server = "key"`.
+fn parse_shared_server_attrs(input: ParseStream) -> syn::Result<SharedServerAttrs> {
+    input.parse::<Token![=]>()?;
+    let lit: LitStr = input.parse()?;
+    Ok(SharedServerAttrs { key: lit.value() })
 }
 
 fn parse_connectors_runtime_attrs(input: ParseStream) -> syn::Result<ConnectorsRuntimeAttrs> {
--- a/core/harness_derive/src/codegen.rs
+++ b/core/harness_derive/src/codegen.rs
@@ -138,6 +138,12 @@ pub fn generate_tests(attrs: &IggyTestAttrs, input: &ItemFn) -> syn::Result<Toke
         .collect();
 
     let params = analyze_signature(&input.sig)?;
+
+    // Validate shared_server constraints
+    if attrs.shared_server.is_some() {
+        validate_shared_server(attrs, &params, fn_name)?;
+    }
+
     let matrix_params_list = matrix_params(&params);
     let has_matrix_params = !matrix_params_list.is_empty();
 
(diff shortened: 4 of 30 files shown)


core/consensus/src/clients_table.rs
```diff
@@ -0,0 +1,234 @@
...
+
+/// Per-client entry in the clients table (VR paper Section 4, Figure 2).
+///
+/// Tracks the most recent committed request for a client and caches
+/// the reply so it can be re-sent on duplicate requests.
+#[derive(Debug)]
+pub struct ClientEntry {
+    /// The request number of the most recent committed request.
+    pub request: u64,
+    /// The cached reply for that request (header + body).
```
numinnex: Add an `Notify` primitive, to notify awaiting client that the respones is ready

krishvishal: Done.

core/consensus/src/clients_table.rs
```diff
@@ -0,0 +1,530 @@
...
+/// request pipelining (currently at most one per client).
+///
+/// The `pending` map is local notification state not replicated, not
+/// serialized, not part of the deterministic committed state.
+///
+/// ## TODO
+///   Checkpoint serialization: the slot array is laid out for deterministic
+///   encode/decode to disk.
+#[derive(Debug)]
+pub struct ClientsTable {
```
hubcio: the VR paper (section 4, figure 2) mandates the client-table as replicated state, transferred during view changes. in this PR, only the primary populates the table (`on_request` + `on_ack`). backups never call `commit_reply` because `ack_preflight` rejects non-primaries. `complete_view_change_as_primary()` and `handle_start_view()` don't rebuild or transfer the table.

after any view change, the new primary's client-table is empty. client retransmissions are treated as `New`, causing duplicate execution. for `SendMessages` this means duplicate message appends to the partition log.

image

if this PR is scaffolding, please add a SAFETY comment at `complete_view_change_as_primary` and `handle_start_view` documenting the gap with a tracking issue number.

core/consensus/src/client_table.rs
```diff
@@ -0,0 +1,530 @@
...
+pub struct ClientsTable {
+    /// `None` means the slot is free.
+    /// Deterministic iteration order for eviction and serialization.
+    slots: Vec<Option<ClientEntry>>,
+    /// Secondary index: `client_id` → slot index. Rebuilt on decode.
+    index: HashMap<u128, usize>,
+    /// Pending commit waiters, keyed by `(client_id, request)`.
+    /// Keyed by request number (not just client) to support future pipelining.
+    /// Currently at most one per client.
+    pending: HashMap<ClientRequest, Notify>,
```
hubcio, line 168: `reset_view_change_state()` clears the pipeline and loopback queue but never touches `clients_table.pending`. if the same replica becomes primary again (e.g. replica 0 in view 0, then again in view 3 with 3 replicas since `3 % 3 == 0`), stale pending entries from the old view survive. `check_request` returns `InProgress` for the orphaned `(client_id, request)` key, silently dropping valid client retries.

fix: add `self.clients_table.borrow_mut().pending.clear()` to `reset_view_change_state()`, or expose a `clear_pending()` method on `ClientsTable`.

core/consensus/src/client_table.rs
```diff
@@ -0,0 +1,530 @@
...
+            notify.notify();
+        }
+    }
+
+    /// Evict the client with the oldest commit number.
+    ///
+    /// Iterates the fixed-size slot array (deterministic order), so all replicas
+    /// with the same committed state evict the same client. Ties on commit number
+    /// are broken by slot index (lowest index wins), which is also deterministic.
+    fn evict_oldest(&mut self) {
```
hubcio, line 310: when the 8192-slot table is full, `evict_oldest()` removes the client with the lowest commit number. if that evicted client retransmits a committed request, `check_request` returns `New` (since `index.get(&client_id)` is now `None`), causing re-execution. this violates VSR's at-most-once semantics.

the VR paper addresses this via checkpoint inclusion of the client-table (which isn't implemented yet - acknowledged in the TODO at line 147). worth a comment here noting that eviction can break the dedup invariant for the evicted client.

core/metadata/src/impls/metadata.rs
```diff
@@ -286,8 +286,45 @@ where
...
+            .check_request(client_id, request);
+        match status {
+            RequestStatus::Duplicate(cached_reply) => {
+                debug!(
+                    "on_request: duplicate request={request} for client={client_id}, re-sending cached reply"
+                );
+                consensus
+                    .message_bus()
+                    .send_to_client(client_id, cached_reply.into_generic())
+                    .await
```
hubcio: also at `core/partitions/src/iggy_partitions.rs:366`.

`.unwrap()` on `send_to_client` in the duplicate-reply path - if the client disconnects between sending the request and receiving the cached reply, this panics the shard. the pre-existing `on_ack` paths have the same `.unwrap()` with a `// TODO: Propagate send error` comment. these new call sites should at minimum have the same TODO, or better, use `if let Err(e) = ... { warn!(...) }`.

core/metadata/src/impls/metadata.rs
```diff
@@ -286,8 +286,45 @@ where
 {
     async fn on_request(&self, message: <VsrConsensus<B> as Consensus>::Message<RequestHeader>) {
         let consensus = self.consensus.as_ref().unwrap();
+        let client_id = message.header().client;
+        let request = message.header().request;
+
+        // Duplicate detection via client-table
```
hubcio: also at `core/partitions/src/iggy_partitions.rs:353-390`.

the `check_request` -> match -> `register_pending` block in `on_request` and the `commit_reply` + `send_to_client` block in `on_ack` are nearly identical between the metadata and partitions planes (~25 lines each). extracting into a helper in `plane_helpers.rs` (following the existing `pipeline_prepare_common` / `ack_preflight` pattern) would reduce drift risk - a fix applied in one plane but not the other would be a subtle bug.

core/consensus/src/client_table.rs
```diff
@@ -0,0 +1,530 @@
...
+            return RequestStatus::InProgress;
+        }
+
+        let Some(&slot_idx) = self.index.get(&client_id) else {
+            return RequestStatus::New;
+        };
+        let entry = self.slots[slot_idx].as_ref().expect("index/slot mismatch");
+        let committed_request = entry.reply.header().request;
+
+        if request < committed_request {
```
hubcio, line 213: `request < committed_request` returns `RequestStatus::InProgress`, but these requests are already committed (stale), not in-progress. callers currently handle this correctly (drop), but a future caller could reasonably try to await an `InProgress` result, which would block forever for stale requests. consider a `Stale` variant or rename to `AlreadyHandled`.

core/consensus/src/clients_table.rs
krishvishal: Done. `commit_journal` on both planes walks WAL, applies state machine, updates client table.

core/consensus/src/client_table.rs
krishvishal: Done. `clear_pending()` with wake-before-clear called in reset_view_change_state

core/consensus/src/client_table.rs
krishvishal: Done. Documented in "Known gaps" (message repair + checkpoint serialization)

core/metadata/src/impls/metadata.rs

core/metadata/src/impls/metadata.rs

core/consensus/src/client_table.rs
krishvishal: Done.

core/partitions/src/iggy_partitions.rs
```diff
@@ -389,15 +406,40 @@ impl<C> IggyPartitions<C> {
...
+                plane = "partitions",
+                replica_id = consensus.replica(),
+                client = client_id,
+                request = request,
+                "on_request: pipeline full, dropping request"
+            );
+            return;
+        }
+
+        let Some(_notify) = request_preflight(consensus, client_id, request).await else {
```
hubcio: `request_preflight` calls `register_pending` here, inserting into `client_table.pending`. if `convert_request_message` fails at line 454, the function returns early without removing the pending entry. after that, every retry from the client for this `(client_id, request)` pair hits `check_request` -> `RequestStatus::InProgress` -> silently dropped. the client is stuck until a view change calls `clear_pending()`.

fix: either move `register_pending` after `convert_request_message` succeeds, or remove the pending entry on the error path. the metadata plane doesn't have this bug - there's no fallible step between `register_pending` and `pipeline_prepare_common`.

core/metadata/src/impls/metadata.rs
```diff
@@ -357,7 +380,10 @@ where
 
         // Force a checkpoint if the journal is running low on capacity.
         if let Some(coordinator) = &self.coordinator {
```
hubcio, line 388: `checkpoint_if_needed` here can call `journal.drain()` which clears slot headers up to `commit_min`. then the hash chain integrity check at line 416 calls `journal.handle().previous_header(&header)` which returns `None` for the just-drained entry, silently skipping verification.

fix: move the hash chain verification (lines 416-418) before this checkpoint block so the previous header is still available when checked.

core/consensus/src/impls.rs
```diff
@@ -403,7 +408,7 @@ pub enum Status {
 }
 
 /// Actions to be taken by the caller after processing a VSR event.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug)]
```
hubcio: `VsrAction` previously derived `Clone, PartialEq, Eq` on master. the new `RetransmitPrepares` variant uses `Vec<(PrepareHeader, Vec<u8>)>` which is `Clone`-able, so `#[derive(Debug, Clone)]` should work here. `PartialEq`/`Eq` can't be restored without also adding those derives to `PrepareHeader`.

core/consensus/src/client_table.rs
```diff
@@ -0,0 +1,576 @@
...
+///
+/// Stores the reply for the client's latest committed request. The client ID,
+/// request number, and commit number are all read from `reply.header()`.
+#[derive(Debug)]
+pub struct ClientEntry {
+    /// The cached reply for the client's latest committed request (header + body).
+    pub reply: Message<ReplyHeader>,
+}
+
+/// Result of checking a request against the client table.
```
hubcio, line 114: `RequestStatus` is missing `#[derive(Debug)]`. `Message<ReplyHeader>` in the `Duplicate` variant already implements `Debug`, so the derive is trivially possible and would make test failures more informative.

core/partitions/src/iggy_partitions.rs
krishvishal: Done.

core/metadata/src/impls/metadata.rs
krishvishal: Done.

core/consensus/src/client_table.rs
krishvishal: Done.

core/consensus/src/impls.rs

numinnex (approved):

PR #3026, opened 2026-03-24
feat(java): Support connection timeout

### LINKED ISSUE #3010: [Java SDK] Implement connection timeout in blocking TCP client
### Description

Currently, the Java SDK's blocking `IggyTcpClient` accepts a `connectionTimeout` in its constructor but it is currently unused. It should be implemented to make it work as expected.

### Component

Java

### Proposed solution

Pass the connection timeout to the connect attempt in the `connect` method of the `InternalTcpClient`

### Alternatives considered

N/A

jrhenderson1988:

Closes #3010 

## Rationale

The connection timeout is a value supported by the builders of both the async and blocking TCP builders, but the clients that they build do not properly apply the connection timeout.

## What changed?

Passes through the connection timeout into the `AsyncIggyTcpClient` and the subsequent `AsyncTcpConnection` where it is used in the bootstrap given to the connection pool. The `ChannelOption.CONNECT_TIMEOUT_MILLIS` value is set with the value of the connection timeout in milliseconds, falling back to the 3000ms value that was used previously if it's not set.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

- Claude was used to ask some questions
- Jetbrains full line completion local model was used for autocomplete
- I can explain all code and full test suite was run to verify that everything still works


foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
```diff
@@ -126,7 +127,7 @@ public class AsyncIggyTcpClient {
      * @param port the server port
      */
     public AsyncIggyTcpClient(String host, int port) {
-        this(host, port, null, null, null, null, null, null, false, Optional.empty());
+        this(host, port, null, null, null, null, null, null, null, false, Optional.empty());
```
jrhenderson1988, line 130: I am unsure if this class is considered to be part of the public API and whether changing the constructor would be considered a breaking change. If so, I may need to re-think this.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
```diff
@@ -238,6 +241,7 @@ public AsyncIggyTcpClient build() {
                 username,
                 password,
                 connectionTimeout,
+                acquireTimeout,
```
jrhenderson1988: Before my change, we were passing `connectionTimeout` to the `acquireTimeout` parameter and there was no `connectionTimeout`. I've since added a `connectionTimeout` and made sure that the arguments line up

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
```diff
@@ -229,6 +229,9 @@ public AsyncIggyTcpClient build() {
         if (connectionPoolSize != null && connectionPoolSize <= 0) {
             throw new IggyInvalidArgumentException("Connection pool size cannot by 0 or negative");
         }
+        if (connectionTimeout != null && (connectionTimeout.equals(Duration.ZERO) || connectionTimeout.isNegative())) {
```
ex172000: Since we have int casting in another place, do we want to validate that there's no overflow?

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilderTest.java
```diff
@@ -137,6 +137,70 @@ void shouldThrowExceptionForZeroPort() {
...
+        // Given: Builder with 0 connection pool size
+        AsyncIggyTcpClientBuilder builder = AsyncIggyTcpClient.builder().connectionPoolSize(0);
+
+        // When/Then: Building should throw IggyInvalidArgumentException
+        assertThatThrownBy(builder::build).isInstanceOf(IggyInvalidArgumentException.class);
+    }
+
+    @Test
+    void shouldThrowExceptionForNegativeConnectionPoolSize() {
+        // Given: Builder with 0 connection pool size
```
ex172000: Maybe we need to update the comments here

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
mmodzelewski: At this phase of development, it's OK. Once we hit 1.0 we'll restrain from breaking changes until major release.

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilderTest.java
ex172000: Nvm, this is outdated now

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
jrhenderson1988: Good idea - I've added a check

PR #3032, opened 2026-03-26
fix(go): update stats field and its deserialization logic to align with rust SDK

chengxilo:

Address https://github.com/apache/iggy/pull/2964#discussion_r2952160233

## Rationale

The rust side now serializes threads_count, free_disk_space, and total_disk_space on the wire (in core/server/src/binary/mapper.rs), but the go Stats struct (foreign/go/contracts/stats.go) and its deserializer (foreign/go/binary_serialization/stats_serializer.go) don't read these fields.

(Yeah, I copied the comment) 😋

## What changed?

1. Adds attributes to `Stats`, also the tests.
2. Removed `TcpStats` and update the function related to it as `Stats` is already enough for our logic.
3. Refactor `Stats` deserialization logic with reader introduced in #2986

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Claude Opus 4.6
2. Code generation and review.
3. I checked it with the rust implementation of stats.
4. Yes.


foreign/go/contracts/stats.go
```diff
@@ -17,27 +17,142 @@
...
+	s.OsName = r.U32LenStr()
+	s.OsVersion = r.U32LenStr()
+	s.KernelVersion = r.U32LenStr()
+	s.IggyServerVersion = r.U32LenStr()
+
+	if r.Err() != nil {
+		return r.Err()
+	}
+
+	// iggy_server_semver is optional. If at least 8 bytes remain, the next 4
```
mmodzelewski: Server encoding has been fixed and now the format is stable. If the semver is missing, it will be encoded as 0, so the following check is no longer required.

foreign/go/contracts/stats.go
```diff
@@ -17,27 +17,142 @@
...
+	}
+
+	if cacheCount > 0 {
+		s.CacheMetrics = make([]CacheMetrics, cacheCount)
+		for i := range s.CacheMetrics {
+			r.Obj(cacheMetricsWireSize, &s.CacheMetrics[i])
+		}
+	}
+
+	if r.Err() == nil && r.Remaining() >= 4 {
```
mmodzelewski: these 3 properties will always be there

foreign/go/contracts/stats.go
```diff
@@ -17,27 +17,142 @@
...
 
+import (
+	"fmt"
+
+	"github.com/apache/iggy/foreign/go/internal/codec"
+)
+
+// CacheMetrics holds cache hit/miss statistics for a single partition.
+type CacheMetrics struct {
+	StreamId    uint32  `json:"stream_id"`
```
mmodzelewski, line 28: Will this struct work for HTTP? Metrics are serialized as Map with computed strings as keys.
```
impl CacheMetricsKey {
    pub fn to_string_key(&self) -> String {
        format!("{}-{}-{}", self.stream_id, self.topic_id, self.partition_id)
    }
}
```

chengxilo: Currently Go SDK doesn't support HTTP. the json tags are actually outdated auto-generated code(I guess). Maybe we can address it when we implement the HTTP client.

PR #3036, opened 2026-03-26
fix: stats semver wire alignment using a zero sentinel

### LINKED ISSUE #2979: map_stats skips writing iggy_server_semver when None, risking binary stream misalignment
`core/server/src/binary/mapper.rs:60-62` conditionally writes `iggy_server_semver` only when `Some`:

```
if let Some(semver) = stats.iggy_server_semver {
    bytes.put_u32_le(semver);
}
```

When `None`, zero bytes are written. But all SDK deserializers (Rust, Java, and the new `binary_protocol` crate) unconditionally read 4 bytes here, using 0 as the `None` sentinel. If the server ever sends None, every field after `semver` gets shifted, corrupting the entire stats response.

The same issue exists in `core/binary_protocol/src/responses/system/get_stats.rs:151-153` (semver) and lines 167-171 (where `threads_count`, `free_disk_space`, `total_disk_space` are also conditionally skipped when `semver` is absent).

This is not critical because it is currently unreachable because the server always derives `semver` from `CARGO_PKG_VERSION` via `SEMANTIC_VERSION.get_numeric_version().ok()`, which succeeds for any valid version. But it's a latent correctness issue.

Fix: bytes.put_u32_le(stats.iggy_server_semver.unwrap_or(0)); and always write the tail fields unconditionally.

atharvalade:

Closes #2979

## Rationale

Stats wire encoding could omit `iggy_server_semver` when `None`, while multiple deserializers assume a fixed 4-byte semver slot (`0` sentinel), risking downstream field misalignment.

## What changed?

When semver was `None`, the binary stats encoder could skip semver bytes (and in protocol encoder also skip tail fields), which could shift subsequent fields for decoders expecting fixed layout.  
Now semver is always encoded as 4 bytes using `0` as the sentinel for `None`, and tail fields (`threads_count`, `free_disk_space`, `total_disk_space`) are always encoded.  
The stats decoder now normalizes semver `0` back to `None`, and tests were updated/added to lock this layout and sentinel behavior.

## Local Execution

- Passed
- Pre-commit hooks not ran

## AI Usage

- Opus 4.6
- exploration and identification
- ran `cargo fmt --check`, `cargo clippy`, targeted `cargo test -p iggy_binary_protocol responses::system::get_stats`, and repository CI scripts for trailing whitespace/newline and license headers
- yes all code can be explained


core/binary_protocol/src/responses/system/get_stats.rs
mmodzelewski: this comment and the related if statement below are no longer relevant

atharvalade: yup, removed the stale conditional check and made the decoder read `semver` and tail fields unconditionally to match the encoder

hubcio (approved):

PR #3040, opened 2026-03-27
feat(security): encrypt user headers alongside message payload

spetz:
Server-side encryption protected message payloads and state
log entries but left user-defined headers in plaintext on
disk. Client-side encryption in the Rust and C# SDKs had
the same gap, encrypting only the payload.

Both server and all SDKs with client-side encryption now
encrypt and decrypt user headers using the same encryptor
as the payload. The integration test verifies no plaintext
headers appear on disk when encryption is enabled, and a
new unit test covers the client-side round-trip. The C#
SDK adds TryMapHeaders for graceful handling of encrypted
header bytes during deserialization.

BREAKING

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5290,6 +5290,7 @@ dependencies = [
  "secrecy",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "toml 1.0.7+spec-1.1.0",
--- a/core/cli/Cargo.toml
+++ b/core/cli/Cargo.toml
@@ -67,3 +67,7 @@ toml = { workspace = true }
 tracing = { workspace = true }
 tracing-appender = { workspace = true }
 tracing-subscriber = { workspace = true, default-features = false }
+
+[dev-dependencies]
+tempfile = { workspace = true }
+tokio = { workspace = true, features = ["macros", "rt"] }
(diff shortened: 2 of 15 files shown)


mmodzelewski (approved):

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -380,11 +382,19 @@ private async Task PollMessagesAsync(CancellationToken ct)
                     try
                     {
                         var decryptedPayload = _config.MessageEncryptor.Decrypt(message.Payload);
+
```
lukaszzborek, line 385: i see it's not covered by tests. I think it will be good to add some integration tests which check it. 
It might be single tests for Producer and Consumer client or seperate test. somthink like:
1. create iggyProducer
2. send encrypted message
3. create 'normal' client;
4. poll this message from topic
5. asset that headers was not decrypted
6. manual decrypt header and check if it is valid

and for iggyConsumer
... send encrypted message
4. poll message via IggyConsumer client with decryptor
5. validate if headers was decrypted property

spetz: Good idea, done

hubcio (approved):

hubcio (changes requested):
lgtm

PR #3041, opened 2026-03-27
refactor(server,sdk): replace ServerCommand/BytesSerializable with wire types

hubcio:
The binary protocol layer had three redundant serialization
paths: ServerCommand enum + binary_mapper (hand-rolled
command dispatch), BytesSerializable trait (domain type
serialization), and WireEncode/WireDecode (wire protocol
types). All three did the same job with different APIs.

This PR collapses them into one: wire types flow from
transport through shard dispatch to handlers. The
ServerCommand enum, binary_mapper, 45+ command structs,
and BytesSerializable trait are deleted entirely.

Changes:
- Route wire types through ShardRequestPayload and SDK
  binary impls, eliminating ServerCommand dispatch
- Delete 45+ command structs from iggy_common/commands/,
  replace with thin HTTP DTOs in iggy_common/http/
- Switch WAL (EntryCommand/StateEntry) to WireEncode/
  WireDecode, delete BytesSerializable from WAL path
- Remove all 21 BytesSerializable impls and delete the
  trait itself
- SDK poll/send now use PollMessagesRequest and
  SendMessagesEncoder (zero-copy) instead of
  intermediate BytesSerializable allocations
- Add wire conversion functions and response-to-domain
  conversions in wire_conversions.rs


core/server/src/binary/handlers/topics/create_topic_handler.rs
```diff
@@ -41,31 +38,16 @@ pub async fn handle_create_topic(
...
-        message_expiry: IggyExpiry::from(req.message_expiry),
-        max_topic_size: MaxTopicSize::from(req.max_topic_size),
-        replication_factor: if req.replication_factor == 0 {
-            None
-        } else {
-            Some(req.replication_factor)
-        },
-        name: req.name.to_string(),
-    };
-    command.validate()?;
```
slbotbm, line 64: I think the 1000 partition count check was happening here, but it was removed in the new code.

hubcio: good catch. I'll fix that.

mmodzelewski (approved):

spetz (approved):

Issue #2100, opened 2025-08-12
Implement C++ SDK


hubcio:
Currently, we have non-working skeleton of C++ SDK in `foreign/cpp` directory. The aim of this task is to write minimal C++ SDK which will allow users to send and poll.

Preferably, we want to call Rust via FFI. See python SDK in `foreign/pyton`, where we use `pyo3` for Rust<->Python interop.

For building the project, see #1774, we want to use bazel.

PR #3046, opened 2026-03-29
feat(cpp): add messaging FFI functions for C++ SDK

### LINKED ISSUE #2965: [C++ SDK] implement `bdd/scenarios/basic_messaging.feature` test for C++ SDK
Add BDD-style integration test for the C++ SDK using a Gherkin-compatible framework (e.g. [cucumber-cpp](https://github.com/cucumber/cucumber-cpp) or [Catch2 BDD](https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#bdd-style-test-cases)).

After implementation add it to CI to make sure it runs on every commit. (see how other SDKs run bdd tests and script `scripts/run-bdd-tests.sh`).

seokjin0414:
## Summary

Closes #2965 (Phase 1 of 2).

- Add `get_streams`, `send_messages`, `poll_messages` FFI functions to C++ SDK
- Add shared structs: `Stream`, `IggyMessageToSend`, `IggyMessagePolled`, `PolledMessages`
- Add 6 e2e tests for message round-trip
- All additions are purely additive — no existing FFI changes

Discussed with @slbotbm on Discord — PR split and message struct design confirmed.

## Test plan

- [x] `cargo check` / `cargo clippy` / `cargo fmt` clean
- [x] 6 e2e tests in `tests/message/low_level_e2e.cpp` (Bazel + iggy-server)


slbotbm:
@seokjin0414 Thanks for the PR!

Here is my first round of review:
The CI is failing. Please check which test is failing. Also, the cpp code is likely not linted properly, so please do that as well.

---
Let's unify the Message struct, instead of having separate structs for sending and polling. The unified struct could look like:
```rust
struct Message {
      checksum: u64,
      id_lo: u64,
      id_hi: u64,
      offset: u64,
      timestamp: u64,
      origin_timestamp: u64,
      user_headers_length: u32,
      payload_length: u32,
      reserved: u64,
      payload: Vec<u8>,
      user_headers: Vec<u8>,
  }
```
Define a function using `impl Message` that would allow the user to create a message to send. This has to be exported to the cpp side. If the bdd tests only require the payload to be set, you can just define a function called `from(payload: Vec<u8>)`.

---
You are using `map(Into::into)` in many places. Let's use `from` to be more explicit about what is being converted to what. 

---
```rust
let mut iggy_messages: Vec<IggyMessage> = messages
            .into_iter()
            .map(|m| {
                let id = ((m.id_hi as u128) << 64) | (m.id_lo as u128);
                let payload = Bytes::from(m.payload);
                let msg = if id > 0 {
                    IggyMessage::builder().id(id).payload(payload).build()
                } else {
                    IggyMessage::builder().payload(payload).build()
                };
                msg.map_err(|error| format!("Could not build message: {error}"))
            })
            .collect::<Result<Vec<_>, _>>()?;
```
In the above code, please define `TryFrom<ffi::Message> for IggyMessage` for the conversion and use that.

---
Let's rename `strategy_kind` and `strategy_value` to `partitioning_strategy_kind` and `partitioning_strategy_value` respectively so that there is no ambiguity about its meaning.

slbotbm:
I wrote the above in multiple sittings, so please tell me if I wrote overlapping things.

foreign/cpp/src/identifier.rs
```diff
@@ -62,6 +62,7 @@ impl TryFrom<ffi::Identifier> for RustIdentifier {
     }
 }
 
+#[allow(clippy::wrong_self_convention)]
```
slbotbm, line 68: Why was this added?

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+    auto polled_part1 = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 1,
+                                              "consumer", make_numeric_identifier(1), "offset", 0, 100, false);
+    ASSERT_EQ(polled_part1.count, 0u);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, SendEmptyMessageVector) {
```
slbotbm: Let's rename to SendEmptyMessageVectorThrows

slbotbm: Also edit the desciption to match this

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+    ASSERT_THROW(client->send_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), "partition_id",
+                                       partition_id_bytes(0), std::move(empty_messages)),
+                 std::exception);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, SendMessageWithEmptyPayload) {
```
slbotbm: Rename to `SendMessageWithEmptyPayloadThrows`

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+    ASSERT_THROW(client->send_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), "partition_id",
+                                       partition_id_bytes(0), std::move(messages)),
+                 std::exception);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, SendMessageWithOversizedPayload) {
```
slbotbm: Rename to SendMessageWithOversizedPayloadThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    ASSERT_NO_THROW(client->connect());
+
+    ASSERT_THROW(client->poll_messages(make_numeric_identifier(1), make_numeric_identifier(0), 0, "consumer",
+                                       make_numeric_identifier(1), "offset", 0, 10, false),
+                 std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesWithInvalidStreamId) {
```
slbotbm: rename to PollMessagesWithInvalidStreamIdThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    invalid_id.length = 0;
+
+    ASSERT_THROW(client->poll_messages(invalid_id, make_numeric_identifier(0), 0, "consumer",
+                                       make_numeric_identifier(1), "offset", 0, 10, false),
+                 std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesFromNonExistentStream) {
```
slbotbm: Rename to PollMessagesFromNonExistentStreamThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    ASSERT_NE(client, nullptr);
+
+    ASSERT_THROW(client->poll_messages(make_string_identifier("nonexistent-stream-poll"), make_numeric_identifier(0), 0,
+                                       "consumer", make_numeric_identifier(1), "offset", 0, 10, false),
+                 std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesWithInvalidConsumerKind) {
```
slbotbm: PollMessagesWithInvalidConsumerKindThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+    ASSERT_THROW(client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "invalid",
+                                       make_numeric_identifier(1), "offset", 0, 10, false),
+                 std::exception);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesWithInvalidStrategyKind) {
```
slbotbm: Rename to PollMessagesWithInvalidStrategyKindThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    client->send_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), "partition_id",
+                          partition_id_bytes(0), std::move(messages));
+
+    auto polled = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                        make_numeric_identifier(1), "first", 0, 3, false);
+
+    ASSERT_EQ(polled.count, 3u);
+    ASSERT_EQ(polled.messages.size(), 3u);
+    EXPECT_EQ(polled.messages[0].offset, 0u);
+
```
slbotbm, line 641: Let's verify the message payload as well here.

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    auto polled = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                        make_numeric_identifier(1), "last", 0, 3, false);
+
+    ASSERT_EQ(polled.count, 3u);
+    ASSERT_EQ(polled.messages.size(), 3u);
+    EXPECT_EQ(polled.messages[0].offset, 7u);
+    EXPECT_EQ(polled.messages[2].offset, 9u);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
```
slbotbm, line 679: Here as well -- verify the message payload

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    client->send_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), "partition_id",
+                          partition_id_bytes(0), std::move(messages));
+
+    auto polled1 = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                         make_numeric_identifier(1), "next", 0, 100, false);
+    ASSERT_EQ(polled1.count, 5u);
+
+    auto polled2 = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                         make_numeric_identifier(1), "next", 0, 100, false);
+    ASSERT_EQ(polled2.count, 5u);
```
slbotbm, line 710: Let's check the message payloads and offsets here as well

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+    auto polled2 = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                         make_numeric_identifier(1), "next", 0, 5, true);
+    ASSERT_EQ(polled2.count, 5u);
+    EXPECT_EQ(polled2.messages[0].offset, 5u);
+    EXPECT_EQ(polled2.messages[4].offset, 9u);
+
+    auto polled3 = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                         make_numeric_identifier(1), "next", 0, 5, true);
+    ASSERT_EQ(polled3.count, 0u);
```
slbotbm, line 772: Let's check the paylaod as well

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    ASSERT_EQ(polled.count, 10u);
+    for (std::uint32_t i = 0; i < 5; i++) {
+        std::string expected = "batch1-" + std::to_string(i);
+        std::string actual(polled.messages[i].payload.begin(), polled.messages[i].payload.end());
+        EXPECT_EQ(actual, expected) << "batch1 payload mismatch at index " << i;
+    }
+    for (std::uint32_t i = 0; i < 5; i++) {
+        std::string expected = "batch2-" + std::to_string(i);
+        std::string actual(polled.messages[5 + i].payload.begin(), polled.messages[5 + i].payload.end());
+        EXPECT_EQ(actual, expected) << "batch2 payload mismatch at index " << i;
```
slbotbm, line 856: Let's check the offsets as well.

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    for (std::uint32_t i = 0; i < 5; i++) {
+        EXPECT_EQ(polled.messages[i].id_lo, id_values[i]) << "ID mismatch at index " << i;
+        EXPECT_EQ(polled.messages[i].id_hi, 0u);
+    }
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesAfterStreamDeleted) {
```
slbotbm: Rename to PollMessagesAfterStreamDeletedThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    client->delete_stream(make_numeric_identifier(saved_stream_id));
+
+    ASSERT_THROW(client->poll_messages(make_numeric_identifier(saved_stream_id), make_numeric_identifier(0), 0,
+                                       "consumer", make_numeric_identifier(1), "offset", 0, 10, false),
+                 std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesWithInvalidPartitionId) {
```
slbotbm: Rename to PollMessagesWithInvalidPartitionIdThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+
+    ASSERT_THROW(client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 9999, "consumer",
+                                       make_numeric_identifier(1), "offset", 0, 10, false),
+                 std::exception);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, PollMessagesWithCountZero) {
```
slbotbm: Rename to PollMessagesWithCountZeroThrows

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    }
+    client->send_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), "partition_id",
+                          partition_id_bytes(0), std::move(messages));
+
+    auto polled = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), UINT32_MAX,
+                                        "consumer", make_numeric_identifier(1), "offset", 0, 100, false);
+
+    ASSERT_EQ(polled.count, 5u);
+    ASSERT_EQ(polled.messages.size(), 5u);
+
```
slbotbm, line 998: Let's verify the message payloads as well

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1137 @@
...
+    ASSERT_EQ(expected_offset, 20u);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, SendMessagesLargeBatch) {
+    RecordProperty("description", "Verifies sending a large batch of 1000 messages succeeds and all are retrievable.");
+    const std::string stream_name = "cpp-msg-large-batch";
+    iggy::ffi::Client *client     = login_to_server();
```
slbotbm, line 1110: Let's elete this test since it is not explicitly testing any limits

foreign/cpp/src/identifier.rs
seokjin0414: It was accidentally added during a rebase. Removed in 03193fa27

foreign/cpp/tests/stream/low_level_e2e.cpp
```diff
@@ -293,3 +294,160 @@ TEST(LowLevelE2E_Stream, GetStreamByNumericIdentifierReturnsStreamDetails) {
...
+    ASSERT_NE(client, nullptr);
+
+    client->create_stream(stream_name);
+    auto streams = client->get_streams();
+    ASSERT_GE(streams.size(), 1);
+
+    bool found = false;
+    for (const auto &s : streams) {
+        if (std::string(s.name) == stream_name) {
+            found = true;
```
slbotbm, line 327: Let's check the id, create_at, size_bytes, message_count, topics_count as well

foreign/cpp/tests/stream/low_level_e2e.cpp
```diff
@@ -293,3 +294,160 @@ TEST(LowLevelE2E_Stream, GetStreamByNumericIdentifierReturnsStreamDetails) {
...
+
+    auto streams = client->get_streams();
+    ASSERT_GE(streams.size(), 1u);
+
+    bool found = false;
+    for (const auto &s : streams) {
+        if (std::string(s.name) == stream_name) {
+            found = true;
+            EXPECT_EQ(s.topics_count, 1u);
+            EXPECT_EQ(s.messages_count, 5u);
```
slbotbm, line 369: Let's compare id, create_at, size_bytes, here as well.

slbotbm (changes requested):
@seokjin0414 I requested some small changes. 
Other than this, I noticed
- you were not testing cases for `send_messages` and `poll_messages` when invalid topic ID was passed to them.
- No test for invalid consumer identifier for `poll_messages`

Also, since the consumer group PR was merged, could you add a test for creating a consumer group and polling messages after joining that?

seokjin0414 (commented):
Thank you for the review!

- Added field assertions to `GetStreamsReturnsStreamAfterCreation` (`created_at`, `size_bytes`, `messages_count`, `topics_count`) and field comparison to `GetStreamsConsistentWithGetStream` (`created_at`, `size_bytes`). Note: `id` is excluded from both due to a known cxx `Vec<SharedStruct>` issue where `id` returns 0.
- Added `SendMessagesWithInvalidTopicIdThrows`, `PollMessagesWithInvalidTopicIdThrows`, `PollMessagesWithInvalidConsumerIdThrows`
- Added `join_consumer_group` and `leave_consumer_group` FFI functions
- Added `ConsumerGroupCreateJoinAndPollMessages` test covering the full flow: create → join → verify member count → send → poll with consumer_group kind → verify payloads → leave → verify member count drops to 0

slbotbm (approved):
@seokjin0414 Thanks for the update!

Could you add the following comment in tests/client/low_level_e2e.cpp please:
```
TODO(slbotbm): Add tests for join_consumer_group() and leave_consumer_group()
```
Also, you commented:

> Note: id is excluded from both due to a known cxx Vec<SharedStruct> issue where id returns 0.

Do you have an issue you could link to for me to see?

Otherwise lgtm!

seokjin0414:
@slbotbm 
I investigated further it's not a cxx bug
Iggy uses a slab allocator for stream IDs (`metadata/reader.rs:202`, `vacant_key()`) so the first stream gets id=0
The original `ASSERT_GT(s.id, 0u)` was simply wrong for that reason
I've added `EXPECT_EQ(list_id, single.id)` to `GetStreamsConsistentWithGetStream` instead which compares the two API results without assuming a specific value

amlel-el-mahrouss (approved):
LGTM, Nothing to add.

slbotbm:
Part of #2100

foreign/cpp/Cargo.toml
```diff
@@ -27,6 +27,7 @@ ignored = ["cxx-build"]
 crate-type = ["staticlib"]
 
 [dependencies]
+bytes = "1.11.1"
```
hubcio, line 30: `bytes = "1.11.1"` hard-pins this crate while the rest of the workspace uses `bytes = { workspace = true }` (see `core/sdk/Cargo.toml:36`, `core/common/Cargo.toml:40`). swap to workspace form so dep versions stay unified.

foreign/cpp/src/client.rs
```diff
@@ -127,6 +139,132 @@ impl Client {
...
+    ) -> Result<(), String> {
+        let rust_stream_id = RustIdentifier::try_from(stream_id)
+            .map_err(|error| format!("Could not send messages: {error}"))?;
+        let rust_topic_id = RustIdentifier::try_from(topic_id)
+            .map_err(|error| format!("Could not send messages: {error}"))?;
+
+        let partitioning = match partitioning_kind.as_str() {
+            "balanced" => Partitioning::balanced(),
+            "partition_id" => {
+                if partitioning_value.len() < 4 {
```
hubcio: `partitioning_value.len() < 4` accepts >4 bytes silently and slices `[..4]`. caller passing 8 bytes (mistaking u64) gets the low 4 bytes with no error. tighten to `!= 4` and reject.

foreign/cpp/src/client.rs
```diff
@@ -127,6 +139,132 @@ impl Client {
...
+                    return Err(
+                        "Could not send messages: partition_id requires 4 bytes".to_string()
+                    );
+                }
+                let id = u32::from_le_bytes(partitioning_value[..4].try_into().map_err(|_| {
+                    "Could not send messages: invalid partition_id value".to_string()
+                })?);
+                Partitioning::partition_id(id)
+            }
+            "messages_key" => Partitioning::messages_key(&partitioning_value).map_err(|error| {
```
hubcio: `Partitioning::messages_key(&partitioning_value)` accepts an empty `Vec<u8>` at the FFI boundary; the SDK rejects it later. validate at the FFI boundary or at minimum document the constraint - right now the call shape is ambiguous.

foreign/cpp/src/client.rs
```diff
@@ -127,6 +139,132 @@ impl Client {
...
+        auto_commit: bool,
+    ) -> Result<ffi::PolledMessages, String> {
+        let rust_stream_id = RustIdentifier::try_from(stream_id)
+            .map_err(|error| format!("Could not poll messages: {error}"))?;
+        let rust_topic_id = RustIdentifier::try_from(topic_id)
+            .map_err(|error| format!("Could not poll messages: {error}"))?;
+        let rust_consumer_id = RustIdentifier::try_from(consumer_id)
+            .map_err(|error| format!("Could not poll messages: {error}"))?;
+
+        let consumer = Consumer {
```
hubcio: use `Consumer::new(rust_consumer_id)` / `Consumer::group(rust_consumer_id)` instead of the struct literal. matches the SDK idiom (see `core/sdk/src/clients/client.rs:138,157`) and `core/common/src/types/consumer/consumer_kind.rs:78,86` provides both constructors.

foreign/cpp/src/client.rs
```diff
@@ -127,6 +139,132 @@ impl Client {
...
+            "last" => PollingStrategy::last(),
+            "next" => PollingStrategy::next(),
+            _ => {
+                return Err(format!(
+                    "Could not poll messages: invalid polling strategy: {polling_strategy_kind}"
+                ));
+            }
+        };
+
+        let opt_partition = if partition_id == u32::MAX {
```
hubcio: `partition_id == u32::MAX` is magic number

foreign/cpp/src/lib.rs
```diff
@@ -51,6 +52,36 @@ mod ffi {
...
+    struct Stream {
+        id: u32,
+        created_at: u64,
+        name: String,
+        size_bytes: u64,
+        messages_count: u64,
+        topics_count: u32,
+    }
+
+    struct Message {
```
hubcio: the unified `Message` struct serves both send and poll paths. server-set fields (`checksum`, `offset`, `timestamp`, `origin_timestamp`, `user_headers_length`, `payload_length`, `reserved`) are exposed on the send path where they're silently overridden by the builder. footgun + mixed concerns. the original PR had separate `IggyMessageToSend` / `IggyMessagePolled` - revert that unification, or at minimum add a doc-comment marking which fields are input vs output.

foreign/cpp/src/lib.rs
```diff
@@ -129,6 +161,44 @@ mod ffi {
...
+            partition_id: u32,
+            consumer_kind: String,
+            consumer_id: Identifier,
+            polling_strategy_kind: String,
+            polling_strategy_value: u64,
+            count: u32,
+            auto_commit: bool,
+        ) -> Result<PolledMessages>;
+
+        fn new_message(self: &mut Message, payload: Vec<u8>);
```
hubcio: `fn new_message(self: &mut Message, payload: Vec<u8>)` is awkward - caller default-constructs then mutates. cxx supports return-by-value of POD shared structs, so a free fn `make_message(payload) -> Message` reads better and avoids `*self = Self {...}` silently zeroing fields the caller already set (see `NewMessageOverwritesPreviousState` in `unit_tests.cpp:110` which documents that surprise as a feature).

foreign/cpp/src/messages.rs
```diff
@@ -0,0 +1,84 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::ffi;
+use bytes::Bytes;
+use iggy::prelude::{IggyMessage as RustIggyMessage, PolledMessages as RustPolledMessages};
+
+impl ffi::Message {
+    pub fn new_message(&mut self, payload: Vec<u8>) {
```
hubcio: `new_message` sets `payload_length` but `TryFrom<ffi::Message> for IggyMessage` ignores it (the builder recomputes from `payload.len()` - see `core/common/src/types/message/iggy_message.rs:194`). field is purely cosmetic on the C++ side, reinforcing the mixed-concerns issue with the unified `Message` struct.

foreign/cpp/src/messages.rs
```diff
@@ -0,0 +1,84 @@
...
+        };
+    }
+}
+
+impl From<RustIggyMessage> for ffi::Message {
+    fn from(m: RustIggyMessage) -> Self {
+        let id = m.header.id;
+        ffi::Message {
+            checksum: m.header.checksum,
+            id_lo: id as u64,
```
hubcio: `id_lo: id as u64` and `id_hi: (id >> 64) as u64` are bare truncating casts on a `u128`. correct, but unidiomatic and clippy `cast_possible_truncation` will flag this eventually. prefer `u128::to_le_bytes()` then split, or explicit `(id & u128::from(u64::MAX)) as u64`.

foreign/cpp/src/messages.rs
```diff
@@ -0,0 +1,84 @@
...
+            user_headers_length: m.header.user_headers_length,
+            payload_length: m.header.payload_length,
+            reserved: m.header.reserved,
+            payload: m.payload.to_vec(),
+            user_headers: m.user_headers.map(|h| h.to_vec()).unwrap_or_default(),
+        }
+    }
+}
+
+impl TryFrom<ffi::Message> for RustIggyMessage {
```
hubcio: `TryFrom<ffi::Message> for RustIggyMessage` only wires `id` and `payload`. `m.user_headers` (caller-supplied data, `Vec<u8>`) is silently dropped on send - if the caller sets headers in C++, they vanish with no error. either reject non-empty `user_headers` with an explicit error, or parse the `Vec<u8>` back into the builder's expected `Option<BTreeMap<HeaderKey, HeaderValue>>` (see `core/common/src/types/message/iggy_message.rs:165-186`). the other dropped fields (`checksum`, `timestamp`, etc.) are server-managed so dropping is fine, but `user_headers` is real user data.

foreign/cpp/src/messages.rs
```diff
@@ -0,0 +1,84 @@
...
+    type Error = String;
+
+    fn try_from(m: ffi::Message) -> Result<Self, Self::Error> {
+        let id = ((m.id_hi as u128) << 64) | (m.id_lo as u128);
+        let payload = Bytes::from(m.payload);
```
hubcio, line 63: the `if id > 0 { builder.id(id) } else { builder.payload(...) }` branch is dead. `IggyMessageBuilder::new` does `id: id.unwrap_or(0)` (see `core/common/src/types/message/iggy_message.rs:189`) so `Some(0)` and `None` are equivalent. simplify to always `.id(id).payload(payload)`.

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1266 @@
...
+                                        make_numeric_identifier(1), "offset", 0, 100, false);
+
+    ASSERT_EQ(polled.count, 0u);
+    ASSERT_EQ(polled.messages.size(), 0u);
+
+    client->delete_stream(make_numeric_identifier(stream.id));
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+}
+
+TEST(LowLevelE2E_Message, SendMessagesBeforeLoginThrows) {
```
hubcio, line 117: this test (and `PollMessagesBeforeLoginThrows` at line 479) skip cleanup because they don't create a stream. fine for these specifically. the broader gap: the suite has no fixture for cleanup on early `ASSERT` failure. when an `ASSERT_*` trips mid-test, the `delete_stream` at the end never runs and orphan `cpp-msg-*` streams accumulate across runs, flaking subsequent suites. add a per-suite `TearDown` that deletes leftover `cpp-msg-*` streams. there's already a `TODO(slbotbm)` for a fixture in `test_helpers.hpp:18` - this is what it's for.

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    iggy::ffi::Client *client     = login_to_server();
+    ASSERT_NE(client, nullptr);
+
+    client->create_stream(stream_name);
+    auto stream = client->get_stream(make_string_identifier(stream_name));
+    client->create_topic(make_numeric_identifier(stream.id), "test-topic", 1, "none", 0, "never_expire", 0,
+                         "server_default");
+
+    rust::Vec<std::uint8_t> oversized_payload;
+    for (std::uint32_t i = 0; i < 64000001u; i++) {
```
hubcio: 64MB+1-byte payload via 64M `push_back` calls is slow - `rust::Vec::push_back` does no reserve. either pre-reserve with `oversized_payload.reserve(64000001)` or send exactly `MAX_PAYLOAD_SIZE+1` (whatever constant the SDK exposes) instead of a hardcoded 64M.

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1266 @@
...
+    rust::Vec<iggy::ffi::Message> batch1;
+    for (std::uint32_t i = 0; i < 5; i++) {
+        iggy::ffi::Message msg;
+        msg.new_message(to_payload("batch1-" + std::to_string(i)));
+        batch1.push_back(std::move(msg));
+    }
+    client->send_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), "partition_id",
+                          partition_id_bytes(0), std::move(batch1));
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
hubcio, line 1023: `sleep_for(100ms)` to ensure the two batches land in distinct timestamps is flaky. `IggyTimestamp::now()` is microseconds so 100ms should be safe, but on busy CI the boundary can collapse. assert `batch2_timestamp - batch1_timestamp >= some_threshold` explicitly so the test fails loudly if the gap closes, instead of silently degrading into a tautology.

foreign/cpp/tests/message/low_level_e2e.cpp
```diff
@@ -0,0 +1,1266 @@
...
+                                     make_numeric_identifier(1), "offset", 0, 100, false);
+    ASSERT_EQ(all.count, 10u);
+
+    std::uint64_t batch2_timestamp = all.messages[5].timestamp;
+    ASSERT_GT(batch2_timestamp, all.messages[0].timestamp);
+
+    auto polled = client->poll_messages(make_numeric_identifier(stream.id), make_numeric_identifier(0), 0, "consumer",
+                                        make_numeric_identifier(2), "timestamp", batch2_timestamp, 100, false);
+
+    ASSERT_GE(polled.count, 5u);
```
hubcio, line 1050: `ASSERT_GE(polled.count, 5u)` then loops `polled.messages.size()` indexing `expected = "batch2-" + i`. if the server returns >5 (e.g. boundary inclusion of a batch1 message with equal-or-later timestamp), index 5+ payloads will not be `"batch2-N"` and the comparison fails with a confusing payload-mismatch error rather than a count error. tighten to `ASSERT_EQ(polled.count, 5u)` if the contract guarantees it, or assert each polled payload has prefix `"batch2-"` regardless of count.

foreign/cpp/tests/message/unit_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+    msg.new_message(std::move(payload));
+
+    ASSERT_EQ(msg.payload_length, 3u);
+    ASSERT_EQ(msg.payload.size(), 3u);
+    EXPECT_EQ(msg.payload[0], 0x00);
+    EXPECT_EQ(msg.payload[1], 0x01);
+    EXPECT_EQ(msg.payload[2], 0x00);
+}
+
+TEST(MessageTest, NewMessageOverwritesPreviousState) {
```
hubcio: `NewMessageOverwritesPreviousState` literally tests "second `new_message` call zeroes `id_lo`". this is testing the smell. if `new_message` is reshaped into a constructor (see `lib.rs:191` comment), this test goes away. fine to keep until the API is reshaped, but flagging - it's a test that's locking in awkward behavior.

hubcio (changes requested):
besides below review I noticed:

- `foreign/cpp/src/client.rs:33-54` `new_connection` returns `*mut Client` via `Box::into_raw`. ownership contract is undocumented - who owns it, who calls `delete_connection`, what happens on double-free. add a doc-comment.
- `foreign/cpp/src/client.rs:586` the `TODO` about `logout_user` failing leaving a leak is misleading. `Box::from_raw`/`drop` runs unconditionally before the error is classified, so there's no leak path. either remove the TODO or rephrase what it actually warns about.
- `foreign/cpp/src/lib.rs:29` `enable_all()` works only because the SDK pulls `time`/`net`/`io-util` transitively. fragile - if SDK feature flags change, this breaks. enable explicit features in this crate's `Cargo.toml`.
- `foreign/cpp/src/identifier.rs` not in this PR's diff but recommend audit for the same magic-number pattern (`length: u8` 0-sentinel for invalid, etc).
- roundtrip tests do not assert `polled.partition_id` matches the partition the messages were sent to. that's a free regression detector for the `u32::MAX`-sentinel bug below - add it.

foreign/cpp/Cargo.toml
slbotbm: The cpp sdk is excluded from the workspace, as is the python sdk. That is why this is hard-pinned.

foreign/cpp/src/lib.rs
slbotbm: Then let's revert to IggyMessageToSend / IggyMessagePolled.

foreign/cpp/tests/message/low_level_e2e.cpp
slbotbm: I'll do it in future PR's

foreign/cpp/Cargo.toml
seokjin0414: as @slbotbm noted — cpp crate is workspace-excluded (same as python sdk), so the pin stays.

foreign/cpp/src/client.rs
seokjin0414: tightened to `!= 4` — 8-byte payloads now error explicitly.

foreign/cpp/src/client.rs
seokjin0414: rejected at the ffi boundary now.

foreign/cpp/src/client.rs
seokjin0414: switched to `Consumer::new()` / `Consumer::group()`.

foreign/cpp/src/client.rs
seokjin0414: exposed as `ANY_PARTITION_ID` with a doc-comment.

foreign/cpp/src/lib.rs
seokjin0414: reverted to separate `IggyMessageToSend` / `IggyMessagePolled` per @slbotbm's call. send path only exposes caller-settable fields now.

foreign/cpp/src/lib.rs
seokjin0414: resolved by the split — `make_message(payload) -> IggyMessageToSend` is a free fn, caller no longer default-constructs and mutates.

foreign/cpp/src/messages.rs
seokjin0414: dropped from the send struct. SDK builder recomputes it anyway.

foreign/cpp/src/messages.rs
seokjin0414: swapped for `u128::to_le_bytes()` + split.

foreign/cpp/src/messages.rs
seokjin0414: rejected explicitly when non-empty. parsing bytes back into `BTreeMap<HeaderKey, HeaderValue>` is more than fits here — follow-up PR.

foreign/cpp/src/messages.rs
seokjin0414: removed. confirmed `IggyMessageBuilder` treats `Some(0)` and `None` identically, so the branch was a no-op.

foreign/cpp/tests/message/low_level_e2e.cpp
seokjin0414: leaving as-is — @slbotbm said he'd handle it in a follow-up.

foreign/cpp/tests/message/low_level_e2e.cpp
seokjin0414: tried `reserve_total` but cxx 1.0.194 has it private (caught in CI, reverted in 5d1f1e692). loop now relies on amortised growth — left a comment about it.

foreign/cpp/tests/message/low_level_e2e.cpp
seokjin0414: added an explicit `>= 50ms` gap assert so it fails loudly instead of silently degrading.

foreign/cpp/tests/message/low_level_e2e.cpp
seokjin0414: switched to prefix matching (`batch1-` or `batch2-`) so boundary inclusion doesn't cause confusing payload mismatches.

foreign/cpp/tests/message/unit_tests.cpp
seokjin0414: gone with the api reshape — no more "second call zeros fields" since `make_message` returns by value.

seokjin0414:
@hubcio @slbotbm 

- new_connection ownership: added `# Ownership` / `# Safety` doc sections covering double-free, use-after-free, and threading.
- logout TODO: dropped — `Box::from_raw` runs unconditionally, no leak path. left a one-liner instead.
- `enable_all()` fragility: pinned tokio features in `foreign/cpp/Cargo.toml` (`rt-multi-thread, macros, time, net, io-util`) so it doesn't ride on sdk transitives.
- identifier magic-number audit: out of scope here, will open a separate PR.
- `polled.partition_id` assertions: added to round-trip / specific-partition / no-partition tests — catches the `u32::MAX` sentinel regression.

foreign/cpp/tests/message/low_level_e2e.cpp
slbotbm: @seokjin0414 cxx does have a reserve function, and the following code can utilize it to pre-reserve 64MB + 1 capacity:
```cpp
constexpr std::uint32_t kOversizedPayloadBytes = 64'000'001u;
    rust::Vec<std::uint8_t> oversized_payload;
    oversized_payload.reserve(kOversizedPayloadBytes);
    for (std::uint32_t i = 0; i < kOversizedPayloadBytes; i++) {
        oversized_payload.push_back(0x41);
    }
```

slbotbm: @hubcio there are many more places in the tests where magic numbers such as 64000001 are being used. Instead of applying a one-off fix for this case, I would like to address them later.

foreign/cpp/src/client.rs
slbotbm: Since this is a constant that is used only once, I would rather not define it as a separate constant. The magic number approach seems fine to me since this is only the low level api. Once the high-level api is created, this can be documented explicitly there. What do you think @hubcio ?

mmodzelewski (approved):

numinnex (approved):

hubcio (approved):

spetz (approved):

PR #3049, opened 2026-03-29
feat(simulator): wire Network into tick loop with per-replica outboxes and replica_crash

### LINKED ISSUE #3048: Implement simulator operation -- kill the node
Implement a mechanism that would allow to kill an node inside of the simulated cluster, this will be tricky, because we use in-memory channels for the cross-node communication, so we cannot rely on the normal network that would prevent client from sending messages to dead node.

Look at the related discussion at: https://github.com/apache/iggy/discussions/3017

krishvishal:
## Summary

- Replace shared `MemBus` FIFO queue with per-replica `SimOutbox` instances that stage outbound messages locally
- Wire `PacketSimulator`-backed `Network` into the simulator tick loop for delivery with configurable delay, loss, and partitioning
- Add `replica_crash()` which disables network links, discards outbox, tracks crashed nodes via `HashSet<u8>`

Implements the outbox-only architecture from #3017. `replica_restart` is defered since it requires consensus durability support that does not yet exist.

Closes #3048

### DIFF
--- a/.github/workflows/_common.yml
+++ b/.github/workflows/_common.yml
@@ -46,7 +46,26 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Check Python SDK versions are synchronized
-        run: ./scripts/ci/python-version-sync.sh --check
+        run: ./scripts/ci/python-sdk-version-sync.sh --check
+
+  version-consistency:
+    name: Check version consistency
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup yq
+        run: |
+          if ! command -v yq >/dev/null 2>&1; then
+            YQ_VERSION="v4.47.1"
+            YQ_CHECKSUM="0fb28c6680193c41b364193d0c0fc4a03177aecde51cfc04d506b1517158c2fb"
+            curl -sSL -o /usr/local/bin/yq https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64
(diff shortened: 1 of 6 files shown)


core/simulator/src/lib.rs
```diff
@@ -23,122 +23,206 @@ pub mod packet;
...
+
+        // Phase 1: Deliver ready packets from the network.
+        let packets = self.network.step();
+        for packet in &packets {
+            match packet.to {
+                ProcessId::Replica(id) => {
+                    if !self.crashed.contains(&id)
+                        && let Some(replica) = self.replicas.get(id as usize)
+                    {
+                        Self::dispatch_to_replica(replica, packet.message.clone()).await;
```
numinnex: I think something that we could potentially do there is to create an dedicated `compio` task per dispatch, to simulate multiple peers concurrently sending messages to replica.

core/simulator/src/main.rs
```diff
@@ -16,32 +16,27 @@
...
-
-    pub fn pop(&mut self) -> Option<Message<ReplyHeader>> {
-        self.queue.pop_front()
+use simulator::Simulator;
+use simulator::client::SimClient;
+use simulator::packet::PacketSimulatorOptions;
+
+/// Step the simulator until at least one client reply is received,
+/// or `max_ticks` is reached. Returns all collected replies.
+fn step_until_reply(sim: &mut Simulator, max_ticks: u64) -> Vec<Message<ReplyHeader>> {
```
numinnex, line 28: We could probably move to start using `compio`, rather than default executor from `futures`

spetz (dismissed):

spetz (approved):

PR #3053, opened 2026-03-30
feat(server-ng): create iggy-server-ng binary crate

krishvishal:
- Add new `core/server-ng` crate as the next gen server binary (`iggy-server-ng`)                                                                                 
  - Mirror existing `core/server` structure: Cargo.toml (same deps), Dockerfile, config.toml, .dockerignore, .http file  with `-ng` applied to all names                                   
  - Enable strict clippy lints (`pedantic = "deny"`, `nursery = "warn"`, `enum_glob_use = "deny"`) matching sibling crates


core/server-ng/build.rs
```diff
@@ -0,0 +1,87 @@
...
+            assets_dir.display()
+        );
+        return;
+    }
+
+    println!(
+        "cargo:info=Web UI assets verified at {}",
+        assets_dir.display()
+    );
+}
```
numinnex, line 87: This file is copied from the current `iggy-server` binary crate ?

krishvishal: Yes.

PR #3061, opened 2026-03-30
fix(connectors): check consume() return value in sink runtime

### LINKED ISSUE #2927: bug(connectors): consume() return value discarded in sink runtime
## Description

The connector runtime calls `consume()` on each sink plugin via FFI but discards the return value. If a sink's `consume()` returns an error (e.g., HTTP timeout, database connection failure, serialization error), the runtime ignores it — no logging, no retry, no metric. It proceeds to the next poll cycle as if delivery succeeded.

## Location

`core/connectors/runtime/src/sink.rs`, lines 585-593:

```rust
(consume)(
    plugin_id,
    topic_meta.as_ptr(),
    topic_meta.len(),
    messages_meta.as_ptr(),
    messages_meta.len(),
    messages.as_ptr(),
    messages.len(),
);
```

The FFI function returns an `i32` (0 = success, 1 = error), but the return value is never checked.

## Impact

This affects **all 7 sink connectors** (Elasticsearch, Iceberg, MongoDB, PostgreSQL, Quickwit, Stdout, and the new HTTP sink). Every `Err(...)` returned by any sink's `consume()` implementation is silently swallowed by the runtime.

As a workaround, sink implementations must log errors internally before returning `Err`. The HTTP sink (#2925) documents this limitation and logs all errors at `error!` level inside `consume()`.

## Suggested Fix

Check the return value and log at minimum:

```rust
let result = (consume)(
    plugin_id,
    topic_meta.as_ptr(),
    topic_meta.len(),
    messages_meta.as_ptr(),
    messages_meta.len(),
    messages.as_ptr(),
    messages.len(),
);

if result != 0 {
    error!(
        "Sink plugin {} consume() returned error code {}",
        plugin_id, result
    );
    // Optionally: increment error metric, trigger retry, etc.
}
```

## Related

- Discussed in #2919 (HTTP sink connector proposal)
- Related to #2928 (offset commit ordering)
- HTTP sink PR: #2925

tmchow:
## Summary

The work was to capture the `i32` return from the sink `consume()` FFI callback and propagate failures. 

The runtime at `sink.rs:585-593` was calling `(consume)(...)` as a statement, silently discarding the return code so this seemed like the correct approach

## Changes

- `sink.rs`: Bind the `i32` return from `(consume)(...)`. If non-zero, log at `error!` level with `plugin_id` and return code, then return `Err(RuntimeError::SinkConsumeFailed(plugin_id, result))`.
- `error.rs`: Add `SinkConsumeFailed(u32, i32)` variant to `RuntimeError` with error code `sink_consume_failed`.

The caller at line 311-328 already handles `Err` by logging and returning the error, so no upstream changes are needed. Lmk if this Was incorrect approach though!

## Testing

`cargo check`, `cargo clippy`, `cargo fmt` all pass. The change is a return-value check on an FFI boundary -- the existing test suite covers the surrounding message processing logic.

Fixes #2927

This contribution was developed with AI assistance (Claude Code).


core/connectors/runtime/src/sink.rs
```diff
@@ -592,5 +592,10 @@ async fn process_messages(
         messages.len(),
     );
 
+    if result != 0 {
+        error!("Sink consume failed for plugin {plugin_id} with return code: {result}");
```
spetz, line 596: Maybe let's include more details in the message such as stream, topic, offset, partition ID.

tmchow: cool, lemme look at this and come back with a proposal. tx.

hubcio:
Closing due to inactivity.

PR #3069, opened 2026-04-01
fix(cli): Ensure home directory exists when writing contexts

### LINKED ISSUE #3030: (cli): `context use` can fail on fresh system when `~/.iggy/` directory does not exist
The `context use` command writes to `~/.iggy/.active_context` via `tokio::fs::write`, which fails if the `~/.iggy/` directory hasn't been created yet.

The recently added `context create` command handles this by calling `ensure_iggy_home_exists()` before writing, but the existing `write_active_context()` and `write_contexts()` methods in `ContextReaderWriter` don't do this.

Steps to reproduce:

1. Remove `~/.iggy/` if it exists
2. Run `iggy context use default`
3. Observe a write error because the directory is missing

Fix would be to call `ensure_iggy_home_exists()` inside `write_active_context()` and `write_contexts()` so all write paths handle a fresh system.

Please wait before #2998 merges to work on this.

pauliyobo:
Closes #3030 

## What changed?

When `write_contexts()` or `write_active_context()` are called, they directly attempt to write the file without ensuring that the iggy home directory exists. This causes issues like #3030 
To fix this, `ensure_iggy_home()` is called in both methods prior to file creation.
## Local Execution
Passed, and precommit ran

PS: I had to add writeable to the typos toml configuration since it would kep being flagged. Not sure if this could cause issues when a genuine spelling mistake happens.

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,43 +421,43 @@ Then, run the benchmarking app with the desired options:
 1. Sending (writing) benchmark
 
    ```bash
-   cargo run --bin iggy-bench -r -- -v pinned-producer tcp
+   cargo run --bin iggy-bench -r -- pinned-producer tcp
    ```
 
 2. Polling (reading) benchmark
 
    ```bash
-   cargo run --bin iggy-bench -r -- -v pinned-consumer tcp
+   cargo run --bin iggy-bench -r -- pinned-consumer tcp
    ```
 
 3. Parallel sending and polling benchmark
 
    ```bash
-   cargo run --bin iggy-bench -r -- -v pinned-producer-and-consumer tcp
+   cargo run --bin iggy-bench -r -- pinned-producer-and-consumer tcp
    ```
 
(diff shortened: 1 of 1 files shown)


.typos.toml
```diff
@@ -30,6 +30,8 @@ bimap = "bimap"
 flate = "flate"
 # Valid Rust compression crate name
 flate2 = "flate2"
+# valid rust crate name
+writeable = "writeable"
```
atharvalade: This `writeable = "writeable"` entry is a duplicate.. it already exists below at line 42 (added as part of the ICU4X crate whitelist). Duplicate keys are a TOML spec violation and may cause a parse error depending on the parser. Please remove this addition; the existing entry already suppresses the false positive.

pauliyobo:
Hello
@atharvalade  thanks a lot, fixed the duplicate entry.

PR #3070, opened 2026-04-01
feat(helm): add helm-docs, yamllint, and helmfmt for chart quality

### LINKED ISSUE #2985: auto generate helm readme so that drift between document and code remains at minimum + beautify(fix formatting issues) YAML
Context

> Also, should we clean up comments from values.yaml, otherwise there will always be documentation drift between two

Good point. For this PR, the `values.yaml` comment and `README` are consistent (both describe `serviceMonitor` as opt-in for Prometheus Operator users), so there's no immediate drift. But you're right that maintaining descriptions in both places long-term is fragile. The # -- comments in values.yaml follow the [helm-docs](https://github.com/norwoodj/helm-docs) convention -- the standard approach is to auto-generate the `README` parameter table from those comments so there's a single source of truth. That would be a nice follow-up but shouldn't block this PR. @avirajkhare00

_Originally posted by @atharvalade in https://github.com/apache/iggy/issues/2976#issuecomment-4094371057_

Tool to use: https://github.com/norwoodj/helm-docs

Open to feedback for the same.

avirajkhare00:
## Summary

Adds automated tooling to maintain Helm chart quality and prevent documentation drift.

## Tools Added

| Tool | Purpose | Files |
|------|---------|-------|
| [helm-docs](https://github.com/norwoodj/helm-docs) | Auto-generate README from values.yaml | `values.yaml` → `README.md` |
| [yamllint](https://github.com/adrienverge/yamllint) | YAML formatting validation | `values.yaml`, `Chart.yaml` |
| [helmfmt](https://github.com/digitalstudium/helmfmt) | Helm template formatting | `templates/*.yaml`, `*.tpl` |

### Documentation
- Add `# --` comments to all values in `values.yaml` (single source of truth)
- Create `README.md.gotmpl` template for helm-docs generation
- Add Development section to README with tooling instructions

### Formatting & Linting
- Add `.yamllint.yml` config for Helm chart (ignores `templates/` with Go templating)
- Format all templates with helmfmt (proper Go template indentation)
- Fix trailing whitespace in rendered templates (14 → 0 lines)
- Remove nginx-specific ingress examples (controller-neutral)

### CI Integration
- Add helm-docs, yamllint, helmfmt installation to `setup-helm-tools` action
- Add validation functions to `test-helm.sh validate`
- Exclude auto-generated README from markdown lint

### Pre-commit Hooks
- Add `helmfmt` hook for template formatting
- Add `yamllint-helm` hook for YAML validation
- Add `helm-docs` hook for README generation

## How It Works

1. Document values with `# --` comments in `values.yaml`
2. Run `helm-docs` to generate README (or let pre-commit do it)
3. CI validates:
   - README stays in sync with values.yaml
   - YAML files are properly formatted
   - Templates follow helmfmt style
   - No trailing whitespace in rendered output

## Test Plan

- [x] `helm lint --strict` passes
- [x] `yamllint` passes on values.yaml and Chart.yaml
- [x] `helmfmt --check` passes on templates
- [x] `scripts/ci/test-helm.sh validate` passes
- [x] Rendered templates have 0 trailing whitespace
- [x] Local kind cluster deployment successful
- [x] Server `/ping` returns `pong`
- [x] UI `/healthz` returns `200`

## AI Usage

1. Tool used: Claude
2. Scope: Implementation of helm-docs, yamllint, helmfmt integration, CI updates, and testing
3. Verification: Ran all linters, full local kind cluster test, verified rendered output
4. Yes, I can explain every line of the change

Closes #2985

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "iggy-web-ui",
-  "version": "0.2.1-edge.1",
+  "version": "0.3.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "iggy-web-ui",
-      "version": "0.2.1-edge.1",
+      "version": "0.3.0",
       "dependencies": {
         "@floating-ui/dom": "^1.7.6",
         "@tailwindcss/postcss": "^4.2.1",
--- a/web/package.json
+++ b/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "iggy-web-ui",
-  "version": "0.2.1-edge.1",
+  "version": "0.3.0",
   "private": true,
   "scripts": {
     "dev": "vite dev --port 3050",


.github/actions/utils/setup-helm-tools/action.yml
```diff
@@ -51,17 +71,37 @@ runs:
...
+
+    - name: Install yamllint
+      if: inputs.install-yamllint == 'true'
+      shell: bash
+      run: pip install yamllint
+
+    - name: Install helmfmt
+      if: inputs.install-helmfmt == 'true'
+      shell: bash
+      run: |
```
tungtose, line 108: It would be better if we had `checksum verification` here

.github/actions/utils/setup-helm-tools/action.yml
```diff
@@ -51,17 +71,37 @@ runs:
...
+        set -euo pipefail
+        HELMFMT_VERSION="v0.5.0"
```
tungtose, line 110: Maybe we could get `HELMFMT_VERSION` from inputs instead of hardcoding it here

hubcio: @tungtose not sure: we use this pattern everywhere in the CI. it would be a good followup item.

avirajkhare00: Yes, we can have this as follow up

tungtose: got it

helm/charts/iggy/templates/servicemonitor.yaml
```diff
@@ -22,9 +22,9 @@ kind: ServiceMonitor
 metadata:
   labels:
     {{- include "iggy.labels" . | nindent 4 }}
-    {{- if .Values.server.serviceMonitor.additionalLabels }}
-    {{ toYaml .Values.server.serviceMonitor.additionalLabels }}
-    {{- end }}
+  {{- if .Values.server.serviceMonitor.additionalLabels }}
+    {{- toYaml .Values.server.serviceMonitor.additionalLabels }}
+  {{- end }}
   name: {{ template "iggy.fullname" .  }}
```
atharvalade: `helmfmt` introduced a regression here. The change from `{{ toYaml to {{- toYaml `eats the newline+indent between the `iggy.labels` output and the `additionalLabels`. When a user sets `server.serviceMonitor.additionalLabels`, the rendered `ServiceMonitor` produces invalid YAML (labels get concatenated onto the previous line). I verified this locally -- helm template fails with a YAML parse error. The fix is to use nindent here, e.g.:

```
  {{- if .Values.server.serviceMonitor.additionalLabels }}
    {{- toYaml .Values.server.serviceMonitor.additionalLabels | nindent 4 }}
  {{- end }}
```

helm/charts/iggy/README.md.gotmpl
atharvalade, line 1: The `gotmpl` template drops the `Troubleshooting`, `Testing`, and `Port Forward` sections that were in the previous README. The Troubleshooting content is especially valuable -- it documents specific error messages (Out of memory (os error 12), Invalid argument panic, ServiceMonitor CRD not found) with concrete solutions that users running Iggy on Kubernetes will hit due to `io_uring` requirements. The Testing section with smoke test instructions is also important for contributors. Please add these sections back into the `README.md.gotmpl` template so they're preserved in the generated output.

.github/actions/utils/setup-helm-tools/action.yml
```diff
@@ -51,17 +75,40 @@ runs:
...
         fi
+        if [[ "${{ inputs.install-helm-docs }}" == "true" ]]; then
+          args+=(--install-helm-docs)
+        fi
         scripts/ci/setup-helm-tools.sh "${args[@]}"
+
+    - name: Install yamllint
+      if: inputs.install-yamllint == 'true'
+      shell: bash
+      run: pip install yamllint
```
hubcio: every other tool in this action (helm, kind, helm-docs, helmfmt) is version-pinned with sha256 checksum verification. bare `pip install` is non-reproducible and may break on ubuntu 24.04+ runners due to PEP 668 externally-managed-environment restrictions. suggestion: `pip install yamllint==1.35.1` or use `pipx`.

avirajkhare00: I have pinned version, since we are installing via pip, sha256verification is something i don't know how to do it. anyways pip always fetches from default registry unless we change it.

PR #3071, opened 2026-04-02
feat(partitions): add consensus per partition and extra

numinnex:
Implement `StoreConsumerOffset` and `DeleteConsumerOffset` as replicated operations on partition. Additionally move `partitions` to an consensus per partition model, rather than relying on one instance of consensus for all of the partitions.

### DIFF
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -659,7 +659,7 @@ where
         let previous_commit = self.consensus.commit_max();
         let current_op = {
             let consensus = self.consensus();
-            let current_op = match replicate_preflight(consensus, &header) {
+            match replicate_preflight(consensus, &header) {
                 Ok(current_op) => current_op,
                 Err(reason) => {
                     emit_partition_diag(
@@ -674,35 +674,44 @@ where
                     );
                     return;
                 }
-            };
-
-            if fence_old_prepare_by_commit(consensus, &header) {
-                emit_partition_diag(
-                    tracing::Level::WARN,
-                    &PartitionDiagEvent::new(
-                        ReplicaLogContext::from_consensus(consensus, PlaneKind::Partitions),
-                        "received old prepare, skipping replication",
-                    )
-                    .with_operation(header.operation)
-                    .with_op(header.op),
-                );
-                return;
             }
-
-            current_op
+        };
+        #[allow(clippy::cast_possible_truncation)]
+        let is_old_prepare = {
+            let consensus = self.consensus();
+            fence_old_prepare_by_commit(consensus, &header)
+                || self.log.journal().inner.header_by_op(header.op).is_some()
+        };
+        let message = if is_old_prepare {
+            emit_partition_diag(
+                tracing::Level::WARN,
+                &PartitionDiagEvent::new(
+                    self.diag_ctx(),
+                    "received old prepare, skipping replication",
+                )
+                .with_operation(header.operation)
+                .with_op(header.op),
+            );
+            message
+        } else {
+            let consensus = self.consensus();
+            replicate_to_next_in_chain(consensus, message).await
         };
 
-        debug_assert_eq!(header.op, current_op + 1);
+        if header.op != current_op + 1 {
+            emit_partition_diag(
+                tracing::Level::WARN,
+                &PartitionDiagEvent::new(self.diag_ctx(), "dropping out-of-order prepare (gap)")
+                    .with_operation(header.operation)
+                    .with_op(header.op),
+            );
+            return;
+        }
         {
             let consensus = self.consensus();
             consensus.sequencer().set_sequence(header.op);
             consensus.set_last_prepare_checksum(header.checksum);
         }
-
-        let message = {
-            let consensus = self.consensus();
-            replicate_to_next_in_chain(consensus, message).await
-        };
         let replicated_result = self.apply_replicated_operation(message).await;
 
         let commit = self.consensus.commit_max();
--- a/core/partitions/src/journal.rs
+++ b/core/partitions/src/journal.rs
@@ -229,6 +229,11 @@ where
         }
     }
 
+    pub fn header_by_op(&self, op: u64) -> Option<PrepareHeader> {
+        let headers = unsafe { &*self.headers.get() };
+        headers.iter().find(|header| header.op == op).copied()
+    }
+
     #[allow(dead_code)]
     fn candidate_start_op(&self, query: &MessageLookup) -> Option<u64> {
         match query {
--- a/core/shard/src/lib.rs
+++ b/core/shard/src/lib.rs
@@ -454,6 +454,7 @@ where
 
             let actions = consensus.handle_start_view_change(PlaneKind::Partitions, &header);
             dispatch_vsr_actions::<B, _, MJ>(consensus, None, &actions).await;
+            dispatch_partition_journal_actions(consensus, partition, &actions).await;
             return;
         }
 
@@ -512,6 +513,7 @@ where
 
             let actions = consensus.handle_do_view_change(PlaneKind::Partitions, &header);
             dispatch_vsr_actions::<B, _, MJ>(consensus, None, &actions).await;
+            dispatch_partition_journal_actions(consensus, partition, &actions).await;
             if actions
                 .iter()
                 .any(|action| matches!(action, VsrAction::CommitJournal))
@@ -564,6 +566,7 @@ where
 
             let actions = consensus.handle_start_view(PlaneKind::Partitions, &header);
             dispatch_vsr_actions::<B, _, MJ>(consensus, None, &actions).await;
+            dispatch_partition_journal_actions(consensus, partition, &actions).await;
             return;
         }
 
@@ -655,6 +658,7 @@ where
             let current_commit = consensus.commit_min();
             let actions = consensus.tick(PlaneKind::Partitions, current_op, current_commit);
             dispatch_vsr_actions::<B, _, MJ>(consensus, None, &actions).await;
+            dispatch_partition_journal_actions(consensus, partition, &actions).await;
         }
     }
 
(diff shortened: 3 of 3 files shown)


core/partitions/src/iggy_partitions.rs
```diff
@@ -642,49 +726,36 @@ where
...
-                    }
-                };
-
-                let partition = self
+            Operation::StoreConsumerOffset | Operation::DeleteConsumerOffset => {
+                let (kind, consumer_id, offset) =
+                    Self::parse_staged_consumer_offset_commit(header.operation, &message)?;
+                let write_lock = self
                     .get_by_ns(namespace)
+                    .expect("store_consumer_offset: partition not found for namespace")
```
krishvishal: Return `IggyError` instead of using `expect`.

core/partitions/src/iggy_partition.rs
```diff
@@ -67,6 +120,158 @@ impl IggyPartition {
...
+                if pending.kind == ConsumerKind::ConsumerGroup =>
+            {
+                let group_id = pending.consumer_id;
+                let guard = self.consumer_group_offsets.pin();
+                let key = ConsumerGroupId(
+                    usize::try_from(group_id).expect("u32 group id must fit usize"),
+                );
+                let _ = guard.remove(&key);
+            }
+            _ => {}
```
krishvishal: Instead of silent catch-all. Make it `unreachable!`.

core/partitions/src/offset_storage.rs
```diff
@@ -0,0 +1,56 @@
...
+        .write(true)
+        .create(true)
+        .open(path)
+        .await
+        .map_err(|_| IggyError::CannotOpenConsumerOffsetsFile(path.to_owned()))?;
+    let buf = offset.to_le_bytes();
+    file.write_all_at(buf, 0)
+        .await
+        .0
+        .map_err(|_| IggyError::CannotWriteToFile)?;
```
hubcio, line 45: `persist_offset` writes via `write_all_at` without `sync_data()` or `sync_all()`. the rest of the codebase (`messages_writer.rs:90-91`, `iggy_index_writer.rs:92-93`) conditionally calls `self.fsync().await` when `enforce_fsync` is true. `send_prepare_ok` is called with `Some(true)` at `iggy_partition.rs:1322`, which tells the primary this replica has durably persisted - but the data may only be in the page cache. this is a false durability claim to the consensus layer.

also: `OpenOptions` is missing `.truncate(true)` - safe today (always 8 bytes at offset 0), but fragile if the format changes.

fix: accept `enforce_fsync` from `PartitionsConfig` and call `file.sync_data().await` when enabled.

core/partitions/src/iggy_partition.rs
```diff
@@ -52,11 +75,66 @@ pub struct IggyPartition {
     pub revision_id: u64,
     pub should_increment_offset: bool,
     pub write_lock: Arc<TokioMutex<()>>,
+    consumer_offsets_path: Option<String>,
+    consumer_group_offsets_path: Option<String>,
+    pending_consumer_offset_commits: HashMap<u64, PendingConsumerOffsetCommit>,
```
hubcio, line 82: `pending_consumer_offset_commits` has two compounding problems:

1. **unbounded growth on followers**: entries are inserted during `on_replicate` (lines 203, 215) but only removed via `on_ack` -> `handle_committed_entries` -> `apply_staged_consumer_offset_commit` (line 222). followers never reach that path because `ack_preflight` (`plane_helpers.rs:161`) returns `Err(NotPrimary)`. the HashMap grows monotonically on every follower - ~88 bytes/entry, 1M consumer offset ops = ~88MB leaked.

2. **orphaned on view change**: `reset_view_change_state` clears the pipeline but has no hook into `IggyPartition`. after view change, orphaned entries with reused op numbers could shadow new ones.

this also means followers never apply consumer offset commits to their in-memory state (`consumer_offsets` / `consumer_group_offsets`), so follower reads are stale/missing. the metadata plane has `commit_journal()` called on followers (`metadata.rs:417-419`) but the partition plane has no equivalent follower-side commit path.

fix: (a) followers should apply consumer offset commits directly during `on_replicate` (they don't need staging since they don't send client replies), (b) add a view-change cleanup hook that clears this HashMap.

core/partitions/src/iggy_partition.rs
```diff
@@ -67,11 +145,193 @@ impl IggyPartition {
...
+
+    pub(crate) async fn persist_and_stage_consumer_offset_upsert(
+        &mut self,
+        op: u64,
+        kind: ConsumerKind,
+        consumer_id: u32,
+        offset: u64,
+    ) -> Result<(), IggyError> {
+        let pending = PendingConsumerOffsetCommit::upsert(kind, consumer_id, offset);
+        self.persist_consumer_offset_commit(pending).await?;
```
hubcio, line 212: consumer offsets are persisted to disk here during the prepare phase (before quorum), unlike `SendMessages` which only writes to the in-memory journal during prepare and persists to disk during commit (`commit_messages`).

if the prepare never commits (leader crash, view change), disk has uncommitted values with no rollback mechanism. on restart, the uncommitted offset file is loaded as if it were committed.

recommended fix: move the `persist_consumer_offset_commit` call from `persist_and_stage_*` (prepare phase) to `commit_consumer_offset_entry` (commit phase). during prepare, only stage in-memory. this eliminates the rollback problem entirely and aligns with the SendMessages pattern.

core/partitions/src/iggy_partition.rs
```diff
@@ -272,3 +502,823 @@ impl Partition for IggyPartition {
...
+            }
+
+            current_op
+        };
+
+        let message = {
+            let consensus = self.consensus();
+            replicate_to_next_in_chain(consensus, message).await
+        };
+        if let Err(error) = self.apply_replicated_operation(message).await {
```
hubcio: if `apply_replicated_operation` fails here, the early `return` at line 625 skips `sequencer().set_sequence(header.op)` at line 630 and `send_prepare_ok` at line 640. but `replicate_to_next_in_chain` at line 610 already forwarded the prepare to the next replica BEFORE this point.

result: chain continues, quorum may be reached, op commits cluster-wide - but this replica permanently misses it. `debug_assert_eq!(header.op, current_op + 1)` at line 627 fires in debug, but in release the replica silently diverges. subsequent prepares arrive with op numbers that no longer match the sequencer.

this is a design divergence from the metadata plane, which advances the sequencer unconditionally before the journal append (`metadata.rs:394-396`).

fix: advance the sequencer before `apply_replicated_operation` (matching the metadata plane), or advance it unconditionally regardless of apply outcome. if the operation truly can't be applied, the replica should request state transfer rather than silently diverging.

core/partitions/src/iggy_partition.rs
```diff
@@ -272,3 +502,823 @@ impl Partition for IggyPartition {
...
+        Ok(Some(
+            batch.header.base_offset + u64::from(message_count) - 1,
+        ))
+    }
+
+    fn parse_staged_consumer_offset_commit(
+        operation: Operation,
+        message: &Message<PrepareHeader>,
+    ) -> Result<(ConsumerKind, u32, Option<u64>), IggyError> {
+        let total_size = message.header().size as usize;
```
hubcio: `header.size as usize` is used to slice `message.as_slice()` without bounds validation. if `total_size > message.as_slice().len()` this panics. the call chain is on the replication hot path: `on_replicate` -> `apply_replicated_operation` -> `parse_staged_consumer_offset_commit`. a malformed prepare from the primary (or protocol version mismatch during rolling upgrade) would crash every follower.

note: the subsequent body parsing (lines 1085-1103) correctly uses checked access (`.first()`, `.get(1..5)`, `.get(5..13)`). only this initial slice creation lacks bounds checking.

fix: `message.as_slice().get(std::mem::size_of::<PrepareHeader>()..total_size).ok_or(IggyError::InvalidCommand)?`

core/partitions/src/iggy_partition.rs
```diff
@@ -191,7 +451,10 @@ impl Partition for IggyPartition {
         if args.auto_commit && !fragments.is_empty() {
             let last_offset =
                 last_matching_offset.expect("non-empty poll result must have a last offset");
-            if let Err(err) = self.store_consumer_offset(consumer, last_offset) {
+            if let Err(err) = self
```
hubcio, line 520: `poll_messages` with `auto_commit=true` calls `store_consumer_offset_and_persist` which persists and applies directly without going through consensus replication. this creates three divergent write paths for consumer offsets:

1. replicated via `StoreConsumerOffset` operation (new in this PR) - full prepare/commit cycle
2. local-only persist + apply via auto-commit here - no replication
3. in-memory-only via `Partition::store_consumer_offset` trait method (line 473) - no persistence, no replication

on failover, the new leader has no record of auto-committed offsets since they were never replicated. this is a pre-existing pattern, but now inconsistent since explicit `StoreConsumerOffset` IS replicated. worth documenting whether this is intentional (local optimization for consumer progress) or should be migrated to the consensus path.

numinnex: I think for auto commit polls, at the end of the `command_handler` construct new `Message<Prepare>`, with `StoreConsumerOffset` command and run it through the  `on_message` handler, the same way an `StoreConsumerOffset` client request would. 

Additionally I think we should expand the `StoreConsumerOffset` command to allow for different levels of consistency  such as `no_ack` or `quorum`, something that `kafka` does for `SendMessages`. This way the `auto_commit` from `PollMessages` command, could inherit that aswell.

core/partitions/src/iggy_partition.rs
numinnex: I think the behavior of treating written offset as `committed` by default is fine, as in scenario where we restart a replica after crash we perform state transfer, which would overwrite the currently commited offset for that particular consumer.

PR #3090, opened 2026-04-08
feat(bench): prettier table output

tungtose:
### Motivation
After mass benchmark runs due to `DirectIO` PR, I started dreaming about p99 latencies. Woke up screaming. Doctor prescribed tables for readability

### Changes
- Replace log walls with `comfy_table` output
- Adaptive layout: vertical on narrow terminals, horizontal on wide (≥140 cols)
- Fix ANSI colors (switched from `info!()` to `println!()`)
- Add `terminal_size` dependency (already in the `lockfile`)

### Before
image

#### wide layout
image

#### narrow layout
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1199,6 +1199,7 @@ version = "0.3.0"
 dependencies = [
  "charming",
  "colored",
+ "comfy-table",
  "derive-new",
  "derive_more",
  "human-repr",
@@ -1206,6 +1207,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sysinfo 0.38.4",
+ "terminal_size",
  "tracing",
  "uuid",
 ]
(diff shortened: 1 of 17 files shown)


numinnex:
@spetz @hubcio What do you think about this ? we already have the web-ui dashboard that we use for higher fidelity benchmarking visuals. 

I don't mind this change, but I am not sure whether this should be used only for summary (aggregate results), or for both summary and individual (producer/consumer/consumer_group) results.

spetz:
IMO looks nice, so as long as it doesn't break anything I think it's a good idea :)

spetz (dismissed):

hubcio:
I like this, but i want this behind an argument `--pretty` (top level arg, just after executable name).
we often dump logs from iggy-bench to file, and in this file i want all the lines to be acutal logs.

tungtose:
@hubcio I have updated the args `--pretty`

hubcio (approved):

mmodzelewski (approved):

PR #3092, opened 2026-04-09
feat(consensus): wire up view change protocol in simulator

krishvishal:
## Summary
- Dispatch all VsrAction variants (SVC/DVC/SV/PrepareOk/Commit from tick and incoming message handlers.
- Add CommitMessage heartbeat: primary periodically sends commit point to backups.
- Integrate consensus tick into Simulator::step().
- Add RebuildPipeline action with primary self-ack so quorum be reached after view change.


numinnex (dismissed):

spetz (dismissed):

core/consensus/src/impls.rs
```diff
@@ -1397,6 +1459,58 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+
+        if self.status.get() != Status::Normal {
+            return false;
+        }
+
+        if header.view != self.view.get() {
+            return false;
+        }
+
+        assert_eq!(
```
hubcio: this `assert_eq!` is reachable from untrusted network input: `on_message` -> `MessageBag::Commit` -> `on_commit` -> `handle_commit` -> panic. a single malformed Commit with the wrong `replica` field kills the shard.

`handle_start_view` (line 1392) uses `if != { return }` for the identical check - this should follow the same pattern.

fix: replace with `if header.replica != self.primary_index(header.view) { return false; }`

core/consensus/src/impls.rs
```diff
@@ -1043,6 +1076,35 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
     }
 
+    /// Primary heartbeat: send commit point to all backups so they know
+    /// the primary is alive and can advance their own `commit_max`.
+    fn handle_commit_message_timeout(&self) -> Vec<VsrAction> {
+        if !self.is_primary() || self.status.get() != Status::Normal {
+            return Vec::new();
+        }
+
+        assert_eq!(
```
hubcio: this assert will deterministically panic whenever a behind backup becomes the new primary after a view change.

`complete_view_change_as_primary` calls `advance_commit_max(max_commit)` (line 1537) without touching `commit_min`, then starts `CommitMessage` timeout (line 1555). if the new primary had `commit_min < max_commit` (normal case - it missed some commits), this assert fires on the next tick. no `commit_journal` is called anywhere in the view change path to catch up `commit_min`.

cascading failure: crashed primary triggers another view change, next primary may also be behind, can exhaust all replicas.

the existing test doesn't cover this because all backups are fully caught up before the crash.

fix options: (a) call `commit_journal` in the view change completion path before starting CommitMessage timeout, (b) defer CommitMessage timeout until `commit_min == commit_max`, or (c) replace assert with a guard that skips the heartbeat when `commit_min < commit_max`.

core/shard/src/lib.rs
```diff
@@ -398,29 +595,245 @@ where
...
+            }
+            VsrAction::RebuildPipeline { from_op, to_op } => {
+                let Some(journal) = journal else {
+                    continue;
+                };
+                // Collect headers before borrowing the pipeline to avoid
+                // holding borrow_mut() across journal reads.
+                let mut gap_at = None;
+                let entries: Vec<_> = (*from_op..=*to_op)
+                    .map_while(|op| {
```
hubcio, line 800: when `map_while` stops at a journal gap, the pipeline tail ends at `gap_at - 1`, but `complete_view_change_as_primary` already set the sequencer to `new_op` (impls.rs:1538). on the next client request:

1. `next_sequence()` returns `new_op + 1`
2. `pipeline.push()` asserts `header.op == tail.header.op + 1` (impls.rs:203)
3. `new_op + 1 != gap_at` -> panic

the primary crashes on the first client request after a view change with a journal gap. in practice C2 (the commit_min/commit_max assert) fires first since the heartbeat timeout is shorter, but this is independently a guaranteed crash.

fix: when a gap is encountered, adjust the sequencer to `gap_at - 1` so the next prepare chains correctly. or refuse to enter `Status::Normal` if the journal is incomplete in the uncommitted range.

core/shard/src/router.rs
```diff
@@ -79,6 +97,12 @@ where
                 0
             })
         } else {
+            // TODO: View change messages (StartViewChange, DoViewChange, StartView) return
```
hubcio: Commit messages also return `Operation::Reserved` from `CommitHeader::operation()`, so they fall through here and route to shard 0. fine for single-shard simulator, but in multi-shard production, partition-plane commit heartbeats go to the wrong shard, get namespace-mismatched, and are dropped - causing spurious view changes.

worth adding Commit to the TODO list alongside the view change messages.

core/shard/src/router.rs
```diff
@@ -79,6 +97,12 @@ where
...
+            // Operation::Reserved, so they always land here and route to shard 0. This is
+            // correct only in single-shard deployments. In multi-shard, partition-plane view
+            // change messages must reach the shard owning that consensus group. Fixing this
+            // requires routing by Command2 + consensus namespace (a u64) rather than by
+            // Operation + IggyNamespace, since view change headers don't carry an IggyNamespace.
             0
         };
         let _ = self.senders[target as usize].send(ShardFrame::fire_and_forget(generic));
```
hubcio: now that view change messages (SVC, DVC, SV) and commit heartbeats flow through this path, `let _ =` silently dropping on a full channel has worse consequences than before. dropped heartbeats cause backup timeouts, dropped SVC/DVC delays leader election. VSR retransmission handles eventual recovery, but the feedback loop (dropped heartbeats -> view changes -> more messages -> more drops) is worth keeping in mind.

hubcio (changes requested):

core/consensus/src/impls.rs

We need peer verification to prevent this.

core/consensus/src/impls.rs

core/shard/src/lib.rs
krishvishal: Done.

core/shard/src/router.rs
krishvishal: Done.

core/shard/src/router.rs
krishvishal: Done.

hubcio (approved):

spetz (approved):

PR #3097, opened 2026-04-10
fix(go): correct buffer size for CreateUser and UpdatePermissions serialization

### LINKED ISSUE #2981: Go SDK: CreateUser omits mandatory permissions_len field when permissions are nil

atharvalade:

Closes #2981

## Rationale

The Go SDK's `CreateUser` and `UpdatePermissions` commands had incorrect buffer size calculations for the `has_permissions` flag byte, causing a 1-byte over-allocation in `CreateUser` (non-nil path) and an index-out-of-bounds panic in `UpdatePermissions` (nil path).

## What changed?

Both `CreateUser.MarshalBinary()` and `UpdatePermissions.MarshalBinary()` double-counted the 1-byte `has_permissions` flag in the non-nil branch, while `UpdatePermissions` omitted it entirely from the base allocation (causing a panic when permissions were nil).

The fix accounts for the `has_permissions` byte exactly once: in `CreateUser` it is part of the base capacity `4`, and in `UpdatePermissions` it is now added to the base `length`. The `+1` was removed from the non-nil branch in both methods. Unit tests were added for all four cases (nil/non-nil x CreateUser/UpdatePermissions).

## Local Execution

- Passed
- Pre-commit hooks ran (trailing-whitespace, trailing-newline, license-headers all pass; `go test ./...` all pass; `gofmt` clean; `go vet` clean)

## AI Usage

1. Opus 4.6
2. Minimal AI used
3. Yes, all code can be explained


hubcio (approved):

foreign/go/internal/command/user.go
```diff
@@ -116,10 +116,10 @@ func (u *UpdatePermissions) MarshalBinary() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	length := len(userIdBytes)
+	length := len(userIdBytes) + 1
 
 	if u.Permissions != nil {
-		length += 1 + 4 + u.Permissions.Size()
```
spetz, line 122: This is incorrect, as the byte on the front sets the flag whether permissions are present or not (1 or 0 as the value).

Actually, it should work like this (this byte is required).

```go
length := len(userIdBytes) + 1  // +1 for has_permissions flag
if u.Permissions != nil {
    length += 4 + u.Permissions.Size()
}
```

spetz (approved):

spetz (changes requested):
Please fix one issue (accidentally approved before).

spetz (approved):

hubcio (approved):

PR #3099, opened 2026-04-11
feat(server): add integration test for message deduplication

### LINKED ISSUE #2872: Add integration test for message deduplication

seokjin0414:
## Summary

Closes #2872

- Add integration test for the message deduplication pipeline (7-step scenario)
- Fix server panic when all messages in a batch are duplicates
- Fix partition offset calculation after dedup removes mid-batch messages
- Fix deduplicator not being created for lazily-initialized partitions

### Bug fixes

**Empty batch panic** (`messages.rs`): After `prepare_for_persistence()` removes all duplicate messages, subsequent `.unwrap()` calls on `first_timestamp()`, `last_timestamp()`, `last_offset()` panic. Added empty batch guard.

**Offset calculation** (`messages.rs`): `last_offset` was computed as `current_offset + count - 1`, which doesn't account for offset gaps created by dedup removal. Changed to use `segment.end_offset` (the actual last offset from batch).

**Deduplicator not created** (`partitions.rs`): `init_partition_inner()` hardcoded `None` for the `message_deduplicator` parameter. Added `create_message_deduplicator()` call matching the bootstrap path.

### Integration test scenario

| Step | Description | Validates |
|------|-------------|-----------|
| 1 | Send 10 messages with id=0 (auto UUID) | All pass through with unique IDs |
| 2 | Send 10 messages with explicit IDs 1-10 | Normal dedup registration |
| 3 | Re-send IDs 1-10 with different payload | Duplicates rejected, original payload preserved |
| 4 | Send all-duplicate batch | No server crash, count unchanged |
| 5 | Send mixed batch (IDs 6-15) | Only new IDs 11-15 accepted |
| 6 | Verify offsets | Monotonically increasing after dedup |
| 7 | Wait for TTL expiry, re-send IDs 1-10 | Previously seen IDs accepted again |

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p server -p integration --all-targets -- -D warnings`
- [ ] `cargo test -p integration --test mod -- message_deduplication` (CI)

### DIFF
--- a/core/integration/tests/server/scenarios/message_deduplication_scenario.rs
+++ b/core/integration/tests/server/scenarios/message_deduplication_scenario.rs
@@ -119,43 +119,7 @@ pub async fn run(harness: &TestHarness) {
         "All-duplicate batch must not change count"
     );
 
-    // Step 5: Send mixed batch — IDs 6-15 (6-10 are duplicates, 11-15 are new)
-    let mixed_start = MESSAGES_PER_BATCH as u128 / 2 + 1; // 6
-    let mixed_end = mixed_start + MESSAGES_PER_BATCH as u128 - 1; // 15
-    let mixed_ids: Vec<u128> = (mixed_start..=mixed_end).collect();
-    let new_ids_count = (mixed_end - MESSAGES_PER_BATCH as u128) as usize; // 5
-    let mut mixed_messages = build_messages("mixed", &mixed_ids);
-    client
-        .send_messages(&stream_id, &topic_id, &partitioning, &mut mixed_messages)
-        .await
-        .unwrap();
-
-    let expected_total = (MESSAGES_PER_BATCH * 2) as usize + new_ids_count; // 25
-    let polled = poll_all(&client, &stream_id, &topic_id, &consumer).await;
-    assert_eq!(polled.messages.len(), expected_total);
-
-    for msg in &polled.messages {
-        let payload = std::str::from_utf8(&msg.payload).unwrap();
-        let id = msg.header.id;
-        if auto_ids.contains(&id) {
-            assert!(
-                payload.starts_with("auto-id-"),
-                "Auto-id message payload mismatch: {payload}"
-            );
-        } else if id <= MESSAGES_PER_BATCH as u128 {
-            assert!(
-                payload.starts_with("original-"),
-                "Original message payload mismatch for id {id}: {payload}"
-            );
-        } else {
-            assert!(
-                payload.starts_with("mixed-"),
-                "New message payload mismatch for id {id}: {payload}"
-            );
-        }
-    }
-
-    // Step 6: Verify monotonically increasing offsets
+    // Step 5: Verify monotonically increasing offsets
     for window in polled.messages.windows(2) {
         assert!(
             window[1].header.offset > window[0].header.offset,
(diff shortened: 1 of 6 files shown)


core/integration/tests/server/scenarios/message_deduplication_scenario.rs
```diff
@@ -0,0 +1,221 @@
...
+    ids.iter()
+        .enumerate()
+        .map(|(idx, &id)| {
+            let payload = if id == 0 {
+                format!("{prefix}-auto-{idx}")
+            } else {
+                format!("{prefix}-{id}")
+            };
+            if id != 0 {
+                IggyMessage::builder()
```
tungtose, line 188: the duplication could be reduced with a builder pattern that conditionally `builder.id()`

eg:
```rust
  let mut builder = IggyMessage::builder().payload(Bytes::from(payload));          
  if id != 0 {
      builder = builder.id(id);
  }            
  builder.build().expect("Failed to build message")
```

This keeps a single builder chain and conditionally  `builder.id()` only when needed

core/integration/tests/server/scenarios/message_deduplication_scenario.rs
```diff
@@ -0,0 +1,221 @@
...
+use bytes::Bytes;
+use iggy::prelude::*;
+use integration::harness::TestHarness;
+use std::collections::HashSet;
+use std::time::Duration;
+
+const STREAM_NAME: &str = "dedup-test-stream";
+const TOPIC_NAME: &str = "dedup-test-topic";
+const MESSAGES_PER_BATCH: u32 = 10;
+const DEDUP_TTL_SECS: u64 = 2;
```
tungtose, line 28: If someone change the server TTL to 5s but forget to update `DEDUP_TTL_SECS`, the test will fail because it won't wait long enough for the TTL to expire. So please add a comment here

tungtose (changes requested):

core/integration/tests/server/scenarios/message_deduplication_scenario.rs
seokjin0414: Thanks for the suggestion 
Applied the conditional builder pattern as you described

numinnex:
Thanks a lot for taking the time to put this together. I really appreciate the effort and thought behind it.

That said, I’m going to hold off on merging this for now. We’re currently in the process of redesigning this part of the system for the upcoming clustered version of the server, which should land in roughly the next month.

In the current implementation, we handle deduplication by storing a message_id and checking server-side whether it already exists. In the clustered setup, this approach is being replaced with a different mechanism: clients will provide a monotonically increasing request_id, and deduplication will be based on that (similar in spirit to how TCP handles packet deduplication).

Because of this shift, the changes in this PR would likely be short-lived and would need to be reworked or removed shortly after merging.

seokjin0414:
@numinnex 
Thanks for the heads-up about the redesign.
Happy to revisit once the new request_id-based mechanism lands —let me know if there's anything related I can help with on that.

PR #3101, opened 2026-04-12
feat(server): return base_offset and partition_id from send_messages

sujitn:
The offset assigned during append was computed but discarded. Propagate it through ShardResponse::SendMessages and include it in the binary protocol response.

Wire format change:
  Before: [status:4 LE][length:4 LE (0)]
  After:  [status:4 LE][length:4 LE (12)][base_offset:8 LE][partition_id:4 LE]

existing SDK clients (excluding node) read the length field and skip unknown body bytes. The send_messages wrapper discards the response body and returns Ok(()).

Fixed node SDK to ignore length

Ref: https://github.com/apache/iggy/discussions/3044

## Local Execution
- Passed
- Pre-commit hooks ran 

## AI Usage
1. Which tools? 
Claude

2. Scope of usage? 
SDK validation 

  - 64/64 server unit tests pass (cargo test -p server --lib)
  - 36/36 offset retrieval integration tests pass, these send messages via the SDK
  - 87/87 concurrent produce/consume integration tests pass 
  - 6/6 authentication integration tests pass 
  - cargo clippy -p server --all-targets -- -D warnings - 0 warnings

Yes


core/server/src/binary/handlers/messages/send_messages_handler.rs
```diff
@@ -197,8 +197,11 @@ pub async fn handle_send_messages(
         topic_id: topic.topic_id,
         partition_id,
     };
-    shard.append_messages(partition, batch).await?;
+    let base_offset = shard.append_messages(partition, batch).await?;
 
-    sender.send_empty_ok_response().await?;
+    let mut response = [0u8; 12];
```
tungtose: Can we use explicit types or constants instead of magic numbers?
eg:
```rust
const OFFSET_SIZE: usize = 8;
const PARTITION_SIZE: usize = 4;
const RESPONSE_SIZE: usize = OFFSET_SIZE + PARTITION_SIZE;

let mut response = [0u8; RESPONSE_SIZE];
response[..OFFSET_SIZE].copy_from_slice(&base_offset.to_le_bytes());
response[OFFSET_SIZE..].copy_from_slice(&(partition_id as u32).to_le_bytes());
```

sujitn: done. thanks

core/server/src/binary/handlers/messages/send_messages_handler.rs
```diff
@@ -197,8 +197,15 @@ pub async fn handle_send_messages(
...
 
-    sender.send_empty_ok_response().await?;
+    const OFFSET_SIZE: usize = size_of::<u64>();
+    const PARTITION_SIZE: usize = size_of::<u32>();
+    const RESPONSE_SIZE: usize = OFFSET_SIZE + PARTITION_SIZE;
+
+    let mut response = [0u8; RESPONSE_SIZE];
+    response[..OFFSET_SIZE].copy_from_slice(&base_offset.to_le_bytes());
+    response[OFFSET_SIZE..].copy_from_slice(&(partition_id as u32).to_le_bytes());
+    sender.send_ok_response(&response).await?;
```
hubcio: this is done incorrectly. check crate `binary_protocol` and `core/binary_protocol/src/requests/messages/send_messages.rs`. IMHO you need to add SendMessagesResponse struct to `core/binary_protocol/src/responses/messages/send_messages.rs` (it doesnt yet exist)

in other words, contract for binary protocol schema is in `binary_protocol` crate, not hardcoded in server sources.

hubcio (changes requested):

core/server/src/binary/handlers/messages/send_messages_handler.rs
sujitn: thanks for the review, added a new SendMessagesResponse struct

hubcio:
I'm gonna review this properly once we finalize current release 0.8.0.

atharvalade (approved):

lukaszzborek:
I think we should at least create issues for every sdk to implement that change in api (just after merge)

hubcio:
We decided to postpone this until cluster v0 is merged.
I'll resume discussion at some point in future one of these threads:
https://github.com/apache/iggy/discussions/3081
https://github.com/apache/iggy/discussions/6

Until then, no code should be written besides spikes/PoCs. Of course, design studies can be continued.

PR #3102, opened 2026-04-12
feat(cpp): add functions related to system

slbotbm:

Works toward completion of #2100 

## Rationale

This PR adds functions related to the system, and tests for previously untested functions.

## What changed?

This PR adds:
- The following functions and their tests:
    - get_stats
    - get_me
    - get_client
    - get_clients
    - ping
    - heartbeat_interval
    - snapshot

The places labelled TODO will be completed after merging of #3046.
## Local Execution

- Passed 
- Pre-commit hooks ran

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT) codex
2. Scope of usage? (e.g., autocomplete, generated functions, entire implementation) code generation based on explicit instructions given by me.
3. How did you verify the generated code works correctly? ran the tests, read through and understood the generated code


slbotbm:
This PR is ready for review

foreign/cpp/src/client.rs
```diff
@@ -20,11 +20,112 @@ use iggy::prelude::{
...
+use std::convert::TryFrom;
 use std::str::FromStr;
 use std::sync::Arc;
 
+impl From<RustClientInfo> for ffi::ClientInfo {
+    fn from(client: RustClientInfo) -> Self {
+        ffi::ClientInfo {
+            client_id: client.client_id,
+            // TODO(slbotbm): In high-level client, this should be converted to None.
+            user_id: client.user_id.unwrap_or(u32::MAX),
```
hubcio, line 47: `client.user_id.unwrap_or(u32::MAX)` collapses `None` and a valid id of `u32::MAX` into the same value. backing type is `Option<u32>` (core/common/src/types/client/client_info.rs:33). plumb option through ffi - add `has_user_id: bool` to `ffi::ClientInfo` or split into a shared cxx option type.

foreign/cpp/src/client.rs
```diff
@@ -20,11 +20,112 @@ use iggy::prelude::{
...
+        }
+    }
+}
+
+impl From<RustClientInfoDetails> for ffi::ClientInfoDetails {
+    fn from(client: RustClientInfoDetails) -> Self {
+        ffi::ClientInfoDetails {
+            client_id: client.client_id,
+            // TODO(slbotbm): In high-level client, this should be converted to None.
+            user_id: client.user_id.unwrap_or(u32::MAX),
```
hubcio, line 61: same sentinel collision in `From<RustClientInfoDetails>`. same fix as L40.

foreign/cpp/src/lib.rs
```diff
@@ -75,6 +75,69 @@ mod ffi {
...
 
+    struct ConsumerGroupInfo {
+        stream_id: u32,
+        topic_id: u32,
+        group_id: u32,
+    }
+
+    struct ClientInfo {
+        client_id: u32,
+        user_id: u32,
```
hubcio, line 126: `user_id: u32` in `ffi::ClientInfo` cannot represent an absent value. root cause behind client.rs:40. add `has_user_id: bool` field.

foreign/cpp/src/lib.rs
```diff
@@ -75,6 +75,69 @@ mod ffi {
...
+        address: String,
+        transport: String,
+        consumer_groups_count: u32,
+    }
+
+    struct ClientInfoDetails {
+        client_id: u32,
+        user_id: u32,
```
hubcio, line 135: same shape problem for `ffi::ClientInfoDetails.user_id`.

foreign/cpp/src/lib.rs
```diff
@@ -75,6 +75,69 @@ mod ffi {
...
+        segments_count: u32,
+        messages_count: u64,
+        clients_count: u32,
+        consumer_groups_count: u32,
+        hostname: String,
+        os_name: String,
+        os_version: String,
+        kernel_version: String,
+        iggy_server_version: String,
+        iggy_server_semver: u32,
```
hubcio, line 179: `iggy_server_semver: u32` paired with `unwrap_or(u32::MAX)` sentinel at client.rs:116. backing field is `Option<u32>` (core/common/src/types/stats/mod.rs:73). prior commit #3036 already got burned by a zero-sentinel - don't repeat. add `has_server_semver: bool`.

foreign/cpp/src/client.rs
```diff
@@ -378,6 +479,98 @@ impl Client {
...
+            self.inner
+                .ping()
+                .await
+                .map_err(|error| format!("Could not ping server: {error}"))?;
+            Ok(())
+        })
+    }
+
+    pub fn heartbeat_interval(&self) -> u64 {
+        RUNTIME.block_on(async { self.inner.heartbeat_interval().await.as_micros() })
```
hubcio, line 751: `RUNTIME.block_on` for `heartbeat_interval` reads a config-shaped value. acceptable since the upstream method is async, but worth an inline comment so future readers don't suspect a bug.

foreign/cpp/src/client.rs
```diff
@@ -378,6 +479,98 @@ impl Client {
...
+    }
+
+    pub fn snapshot(
+        &self,
+        snapshot_compression: String,
+        snapshot_types: Vec<String>,
+    ) -> Result<Vec<u8>, String> {
+        let rust_compression = match snapshot_compression.trim() {
+            "" => RustSnapshotCompression::default(),
```
hubcio: empty `snapshot_compression` silently falls back to `RustSnapshotCompression::default()`. either reject empty (consistent with the hpp factory which always emits a non-empty value) or document the `"" == default` contract.

foreign/cpp/src/client.rs
```diff
@@ -378,6 +479,98 @@ impl Client {
...
+            value => RustSnapshotCompression::from_str(value).map_err(|error| {
+                format!("Could not capture snapshot: invalid compression '{value}': {error}")
+            })?,
+        };
+        let rust_snapshot_types = snapshot_types
+            .into_iter()
+            .map(|snapshot_type| {
+                RustSystemSnapshotType::from_str(&snapshot_type).map_err(|error| {
+                    format!(
+                        "Could not capture snapshot: invalid snapshot type '{}': {}",
```
hubcio: format string mixes positional `'{}': {}` with `snapshot_type, error` while the sibling at L550 uses named captures. use `'{snapshot_type}': {error}` for consistency. `cargo clippy -- -W clippy::uninlined_format_args` would catch it.

foreign/cpp/src/client.rs
```diff
@@ -378,6 +479,98 @@ impl Client {
...
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        RUNTIME.block_on(async {
+            let snapshot = self
+                .inner
+                .snapshot(rust_compression, rust_snapshot_types)
+                .await
+                .map_err(|error| format!("Could not capture snapshot: {error}"))?;
+            Ok(snapshot.0)
```
hubcio: `Ok(snapshot.0)` is an opaque tuple-struct index. `let Snapshot(bytes) = snapshot; Ok(bytes)` reads better.

foreign/cpp/src/identifier.rs
```diff
@@ -62,6 +62,7 @@ impl TryFrom<ffi::Identifier> for RustIdentifier {
     }
 }
 
+#[allow(clippy::wrong_self_convention)]
```
hubcio: `#[allow(clippy::wrong_self_convention)]` silences a real signal. `from_string`/`from_numeric` take `&mut self` and mutate in place - the names lie about ownership. rename to `set_string`/`set_numeric` (update lib.rs:210-211 bridge + c++ callers) and drop the suppression.

foreign/cpp/include/iggy.hpp
```diff
@@ -45,6 +45,42 @@ class CompressionAlgorithm final {
     std::string algorithm_;
 };
 
+class SnapshotCompression final {
```
hubcio: `SnapshotCompression`/`SnapshotType` (and the existing `IdKind`/`CompressionAlgorithm`) are identical `std::string` wrapper-with-factory patterns. extract a `StringTag<Tag>` template/CRTP base.

foreign/cpp/include/iggy.hpp
```diff
@@ -45,6 +45,42 @@ class CompressionAlgorithm final {
...
+  public:
+    static SnapshotCompression stored() { return SnapshotCompression("stored"); }
```
hubcio, line 82: factory enforces a fixed enumeration but the FFI signature accepts `String`, which is then re-parsed in `RustSnapshotCompression::from_str` (client.rs:549). either pass an enum tag across FFI or document the parse-trip contract.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+
+    ASSERT_THROW(client->get_stats(), std::exception);
+    ASSERT_NO_THROW(client->connect());
+    ASSERT_THROW(client->get_stats(), std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+    client = nullptr;
+}
+
+// TODO(slbotbm-this PR): add a test to create some streams, topics, partitions, and segments, send messages, and create
```
hubcio: `TODO(slbotbm-this PR)` self-declares as blocking on this PR. resolve here or strip the `-this PR` suffix before merge.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+
+    ASSERT_THROW(client->get_me(), std::exception);
+    ASSERT_NO_THROW(client->connect());
+    ASSERT_THROW(client->get_me(), std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+    client = nullptr;
+}
+
+// TODO(slbotbm-this PR): add additional validation for get_me after merging join_consumer_group PR.
```
hubcio: same `TODO(slbotbm-this PR)` marker. resolve or strip.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+    iggy::ffi::Client *second_client = nullptr;
+    iggy::ffi::Client *third_client  = nullptr;
+
+    iggy::ffi::Stats empty_stats{};
+    ASSERT_NO_THROW({
+        empty_stats = client->get_stats();
+        EXPECT_NE(empty_stats.process_id, 0u);
+        EXPECT_GT(empty_stats.threads_count, 0u);
+        EXPECT_GT(empty_stats.total_memory, 0u);
+        EXPECT_GE(empty_stats.available_memory, 0u);
```
hubcio: `EXPECT_GE(empty_stats.available_memory, 0u)` is a tautology - `available_memory` is `u64`, so `>= 0u` always holds. drop or replace with `EXPECT_LE(available_memory, total_memory)`.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+        EXPECT_GE(empty_stats.total_disk_space, empty_stats.free_disk_space);
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.hostname).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.os_name).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.os_version).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.kernel_version).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.iggy_server_version).empty());
+        EXPECT_TRUE(empty_stats.iggy_server_semver == std::numeric_limits<std::uint32_t>::max() ||
+                    empty_stats.iggy_server_semver > 0u);
+        EXPECT_GE(empty_stats.cache_metrics.size(), 0u);
```
hubcio: `EXPECT_GE(empty_stats.cache_metrics.size(), 0u)` is tautological for `size_t`. drop.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+        EXPECT_EQ(empty_stats.streams_count, 0u);
```
hubcio: absolute counts (`streams_count == 0u`, `topics_count == 0u` here, `consumer_groups_count == 3u` at L264, the full block at L280-285) assume a fresh, single-process server. replace with deltas vs the `empty_stats` baseline already captured at L194.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+// TODO(slbotbm-this PR): add additional validation for get_me after merging join_consumer_group PR.
+TEST(LowLevelE2E_Client, GetMeReturnsCurrentClientDetails) {
+    RecordProperty("description", "Returns the current authenticated client details.");
+    iggy::ffi::Client *client = login_to_server();
+    ASSERT_NE(client, nullptr);
+
+    ASSERT_NO_THROW({
+        const auto me = client->get_me();
+        EXPECT_NE(me.client_id, 0u);
+        EXPECT_NE(me.user_id, std::numeric_limits<std::uint32_t>::max());
```
hubcio: `EXPECT_NE(me.user_id, max())` couples test correctness to the sentinel hack at client.rs:40. also at L438, L449. once the option/none plumbing lands these flip to `EXPECT_TRUE(me.has_user_id)`.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+    ASSERT_NO_THROW(client->connect());
+    ASSERT_THROW(client->get_client(1), std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+    client = nullptr;
+}
+
+TEST(LowLevelE2E_Client, GetClientWithWrongClientIdThrows) {
+    RecordProperty("description", "Rejects querying invalid or non-existent client ids.");
+    const std::uint32_t wrong_client_ids[] = {0u, std::numeric_limits<std::uint32_t>::max()};
```
hubcio: `wrong_client_ids = {0u, max()}` - `max()` IS the sentinel returned when user_id is missing. test conflates `invalid id` with `missing user`. use a different unreachable id (e.g. `last_id_from_get_clients() + 1`).

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+
+TEST(LowLevelE2E_Client, HeartbeatIntervalReturnsDefaultValueForNewConnection) {
+    RecordProperty("description",
+                   "Returns the default heartbeat interval in microseconds for a fresh unauthenticated client.");
+    iggy::ffi::Client *client = nullptr;
+    ASSERT_NO_THROW({ client = iggy::ffi::new_connection(""); });
+    ASSERT_NE(client, nullptr);
+
+    const auto heartbeat_interval = client->heartbeat_interval();
+    EXPECT_EQ(heartbeat_interval, 5'000'000u);
```
hubcio: magic `5'000'000u` (and `10'000'000u` at L757). define `constexpr std::uint64_t kDefaultHeartbeatMicros = 5'000'000ull;` or derive from `IggyDuration::from("5s").as_micros()`.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+    ASSERT_NE(client, nullptr);
+
+    const auto heartbeat_interval = client->heartbeat_interval();
+    EXPECT_EQ(heartbeat_interval, 10'000'000u);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+    client = nullptr;
+}
+
+TEST(LowLevelE2E_Client, SnapshotBeforeLoginThrows) {
```
hubcio, line 797: all four snapshot tests (L763, L782, L797, L811) are negative paths. add a happy-path test with `"deflated"` + `["test"]` asserting the returned `Vec<u8>` is non-empty.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+    RecordProperty("description", "Rejects get_stats before connect, and after connect but before login.");
+    iggy::ffi::Client *client = nullptr;
+    ASSERT_NO_THROW({ client = iggy::ffi::new_connection(""); });
+    ASSERT_NE(client, nullptr);
+
+    ASSERT_THROW(client->get_stats(), std::exception);
+    ASSERT_NO_THROW(client->connect());
+    ASSERT_THROW(client->get_stats(), std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
```
hubcio, line 168: every test manually does `delete_connection(client); client = nullptr;`. wrap in `struct ScopedClient { ~ScopedClient() { iggy::ffi::delete_connection(p); } };` to cut ~30 lines of boilerplate and recover correctness on test-assertion failure - today the connection leaks if an `ASSERT_*` aborts mid-test.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -149,3 +153,671 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+    client = nullptr;
+}
+
+TEST(LowLevelE2E_Client, SnapshotBeforeLoginThrows) {
+    RecordProperty("description", "Rejects snapshot before connect, and after connect but before login.");
+    iggy::ffi::Client *client = nullptr;
+    ASSERT_NO_THROW({ client = iggy::ffi::new_connection(""); });
+    ASSERT_NE(client, nullptr);
+
+    rust::Vec<rust::String> snapshot_types_before_connect;
```
hubcio: `rust::Vec<rust::String> snapshot_types; snapshot_types.push_back(...)` repeats at L774, L787, L802, L816. add a `make_snapshot_types({"test"})` helper in `tests/common/test_helpers.hpp`.

foreign/cpp/build.rs
```diff
@@ -25,5 +25,6 @@ fn main() {
     println!("cargo:rerun-if-changed=src/identifier.rs");
     println!("cargo:rerun-if-changed=src/lib.rs");
     println!("cargo:rerun-if-changed=src/stream.rs");
+    println!("cargo:rerun-if-changed=src/system.rs");
```
hubcio: `cargo:rerun-if-changed=src/system.rs` watches a file that doesn't exist (verified `ls foreign/cpp/src/system.rs`). cargo accepts the entry but it's dead config. delete the line or create the file.

hubcio (changes requested):

foreign/cpp/build.rs
slbotbm: deleted line

foreign/cpp/src/client.rs
slbotbm: added has_user_id

foreign/cpp/src/client.rs
slbotbm: fixed

foreign/cpp/src/lib.rs
slbotbm: fixed

foreign/cpp/src/lib.rs
slbotbm: done

foreign/cpp/src/lib.rs
slbotbm: done

foreign/cpp/src/client.rs
slbotbm: done

foreign/cpp/src/client.rs
slbotbm: client now rejects empty value for snapshot compression

foreign/cpp/src/client.rs
slbotbm: `cargo clippy -- -W clippy::uninlined_format_args` caught a lot more errors than this -- fixed them all

foreign/cpp/src/client.rs
slbotbm: done

foreign/cpp/src/identifier.rs
slbotbm: done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: stripped. I'll fix this in the next PR

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: done

slbotbm:
@hubcio i still need to completely address the review for this pr

maybe /author is better?

hubcio:
@slbotbm I know. `/ready` is me marking the PR ready for review. see `CONTRIBUTING.md` - i added simple triage bot. this PR was created before the bot, so I'm marking manually all existing PRs.

EDIT: you are right! this should be /author...

foreign/cpp/include/iggy.hpp

foreign/cpp/include/iggy.hpp
slbotbm: added a warning + documented what each of enum and its options do.

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: Will be addressed in the next PR. I will create proper clean up for the tests.

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: Done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: now mid-test checks happen with empty stats + delta, and post-cleanup, stats are compared to original baseline.

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: converted to EXPECT_TRUE(me.has_user_id)

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: now tests gets live client ids, and increments 1u if it exists (until a non-existent id is found).

foreign/cpp/src/client.rs
```diff
@@ -604,6 +745,50 @@ impl Client {
...
         })
     }
+
+    pub fn heartbeat_interval(&self) -> u64 {
+        // The upstream client exposes this config-derived value via an async API,
+        // so the synchronous C++ wrapper reads it by blocking on the runtime.
+        RUNTIME.block_on(async { self.inner.heartbeat_interval().await.as_micros() })
+    }
+
+    pub fn snapshot(
```
hubcio, line 754: empty `snapshot_types` vec → server loops zero times in `core/server/src/shard/system/snapshot/mod.rs:71` and returns an empty zip with `Ok(())`. silent no-op. duplicate types collide on the `{snapshot_type}.txt` filename at line 80, so the second entry quietly overwrites the first. reject empty vec at the ffi boundary and dedup before passing to the sdk.

foreign/cpp/include/iggy.hpp
```diff
@@ -26,46 +26,128 @@
...
-    std::string expiry_kind_;
-    std::uint64_t expiry_value_;
+/// Snapshot selector for `Client::snapshot(...)`.
+///
+/// Use this type to choose which snapshot data the server should include.
+///
+/// Internally, each selected value is passed across the Rust FFI as a string.
+/// Values outside the supported set are rejected by the Rust client.
+/// Some Rust-side aliases may also be accepted for compatibility.
+class SnapshotType final : private detail::StringTag<SnapshotType> {
```
hubcio: rust calls this `SystemSnapshotType` (`core/common/src/types/snapshot/mod.rs:36`) and c# preserves the prefix in `foreign/csharp/Iggy_SDK/Enums/SystemSnapshotType.cs`. c++ binding drops `System`. rename to `SystemSnapshotType` while pre-release to stop the divergence.

foreign/cpp/include/iggy.hpp
```diff
@@ -26,46 +26,128 @@
...
+
+}  // namespace detail
+
+/// Compression setting for `Client::create_topic(...)`.
+///
+/// Use this type to choose whether messages in a topic are stored as-is or compressed with gzip.
+///
+/// Internally, this value is passed across the Rust FFI as a string.
+/// Values outside the supported set are rejected by the Rust client.
+/// Some Rust-side aliases may also be accepted for compatibility.
```
hubcio: this `"Some Rust-side aliases may also be accepted for compatibility."` line is repeated on `SnapshotCompression` at line 78 and `SnapshotType` at line 107. only `MaxTopicSize::from_str` accepts genuine token aliases (`"0"` ≡ `"server_default"`). the other three only do `s.to_lowercase()` then exact-match (`core/common/src/types/compression/compression_algorithm.rs:45-51`, `core/common/src/types/snapshot/mod.rs:129-141, 171-181`). drop the alias claim from the three inaccurate docs, keep on `MaxTopicSize` only.

foreign/cpp/src/identifier.rs
```diff
@@ -67,14 +67,14 @@ impl TryFrom<ffi::Identifier> for RustIdentifier {
 // preserves the C++ ABI used by every test and downstream binding.
 #[allow(clippy::wrong_self_convention)]
```
hubcio: follow-up on the earlier rename thread: `from_string` → `set_string` and `from_numeric` → `set_numeric` landed but the `#[allow(clippy::wrong_self_convention)]` here is now dead (`set_*` doesn't trigger that lint) and the comment above about "keeping the names preserves the C++ ABI used by every test and downstream binding" directly contradicts the rename in this same pr. delete both.

foreign/cpp/src/lib.rs
```diff
@@ -114,6 +114,72 @@ mod ffi {
...
+    }
+
+    struct Stats {
+        process_id: u32,
+        cpu_usage: f32,
+        total_cpu_usage: f32,
+        memory_usage: u64,
+        total_memory: u64,
+        available_memory: u64,
+        run_time: u64,
```
hubcio: `run_time` and `start_time` are both bare `u64` but carry different semantics - `client.rs:107-108` fills them as `stats.run_time.as_micros()` (`IggyDuration`, micros since boot) and `stats.start_time.as_micros()` (`IggyTimestamp`, unix-epoch micros). consumers cannot tell them apart without reading the rust source. rename to `run_time_micros` and `start_time_epoch_micros`, or doc-comment the bridge struct.

foreign/cpp/src/client.rs
```diff
@@ -21,15 +21,119 @@ use iggy::prelude::{
...
 /// partition based on the consumer/strategy. Cxx FFI does not support `Option<u32>`, so we
 /// reserve `u32::MAX` as the sentinel for `partition_id`.
 const ANY_PARTITION_ID: u32 = u32::MAX;
+impl From<RustClientInfo> for ffi::ClientInfo {
+    fn from(client: RustClientInfo) -> Self {
+        let has_user_id = client.user_id.is_some();
+        ffi::ClientInfo {
+            client_id: client.client_id,
+            has_user_id,
+            user_id: client.user_id.unwrap_or(0),
```
hubcio: follow-up on the earlier `unwrap_or(u32::MAX)` thread: switching to `unwrap_or(0)` here and at line 60 trades one collision for another - root user has id 0 (`core/common/src/http/users/defaults.rs:26`, applied at `core/server/src/streaming/users/user.rs:96`), so `(has_user_id=true, user_id=0)` for a root login is indistinguishable from `(has_user_id=false, user_id=0)` for an unauthenticated session whenever a consumer reads `user_id` without first checking the bool. wire encoding already uses `u32::MAX` as the none sentinel (`core/server/src/binary/handlers/system/get_me_handler.rs:64`, decoder at `core/common/src/wire_conversions.rs:258-261`). either doc the `has_user_id` contract on the bridge structs (`lib.rs:123-130, 132-140`), or align ffi to the wire sentinel with `unwrap_or(u32::MAX)`.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -150,3 +154,702 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+    ASSERT_NO_THROW(client->connect());
+    ASSERT_THROW(client->get_stats(), std::exception);
+
+    ASSERT_NO_THROW(iggy::ffi::delete_connection(client));
+    client = nullptr;
+}
+
+// TODO(slbotbmR): add a test to create some streams, topics, partitions, and segments, send messages, and create
+// consumer groups and verify it.
+TEST(LowLevelE2E_Client, GetStatsReturnsServerStats) {
```
hubcio, line 178: `tests/common/test_helpers.hpp:18` already has a `TODO(slbotbm): create fixture for setup/teardown` covering the broader cleanup story, but this specific test has two extra fragilities not implicit in that todo: (1) fixed stream/topic/group names (`cpp-get-stats-stream-1` etc) - any failing `ASSERT_NO_THROW` before the inline cleanup at lines 266-272 leaks streams and the next run collides on `create_stream`; (2) `clients_count == empty + 2u` (line 260) compares deltas against `empty_stats` taken seconds earlier on a shared server - any other client connecting in between invalidates the assertion. once the fixture lands, randomize names and switch the count assertions to `EXPECT_GE` so this test stops being order-dependent. add a `TODO` here if the fix isn't landing in this pr.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -150,3 +154,702 @@ TEST(LowLevelE2E_Client, DeleteNullConnectionIsNoop) {
...
+        EXPECT_GT(empty_stats.threads_count, 0u);
+        EXPECT_GT(empty_stats.total_memory, 0u);
+        EXPECT_LE(empty_stats.available_memory, empty_stats.total_memory);
+        EXPECT_GE(empty_stats.total_disk_space, empty_stats.free_disk_space);
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.hostname).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.os_name).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.os_version).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.kernel_version).empty());
+        EXPECT_FALSE(static_cast<std::string>(empty_stats.iggy_server_version).empty());
+        EXPECT_TRUE(!empty_stats.has_server_semver || empty_stats.iggy_server_semver > 0u);
```
hubcio: `!has_server_semver || iggy_server_semver > 0u` is unfalsifiable. wire decoder at `core/binary_protocol/src/responses/system/get_stats.rs:204` sets `Some(raw)` iff `raw != 0`, so `has_server_semver=true` already implies `iggy_server_semver > 0`. use `EXPECT_EQ(empty_stats.has_server_semver, empty_stats.iggy_server_semver != 0u)` or drop the assertion.

foreign/cpp/src/client.rs
```diff
@@ -21,15 +21,119 @@ use iggy::prelude::{
...
+            messages_count: stats.messages_count,
+            clients_count: stats.clients_count,
+            consumer_groups_count: stats.consumer_groups_count,
+            hostname: stats.hostname,
+            os_name: stats.os_name,
+            os_version: stats.os_version,
+            kernel_version: stats.kernel_version,
+            iggy_server_version: stats.iggy_server_version,
+            has_server_semver,
+            iggy_server_semver: stats.iggy_server_semver.unwrap_or(0),
```
hubcio, line 126: same `unwrap_or(0)` shape as the `user_id` follow-up. wire decoder at `core/binary_protocol/src/responses/system/get_stats.rs:204` forbids `Some(0)` from a conforming server so the collision is unreachable today, but the pattern is fragile - if anyone ever changes the wire format to allow `Some(0)`, this silently degrades to the same shape as `None`. doc the invariant on the bridge struct, or align to a sentinel that cannot collide.

foreign/cpp/src/client.rs
```diff
@@ -501,14 +590,12 @@ impl Client {
                 .await
                 .map_err(|error| {
                     format!(
-                        "Could not get consumer group '{}' for topic '{}' on stream '{}': {error}",
-                        rust_group_id, rust_topic_id, rust_stream_id
+                        "Could not get consumer group '{rust_group_id}' for topic '{rust_topic_id}' on stream '{rust_stream_id}': {error}"
```
hubcio, line 594: follow-up on the earlier `clippy::uninlined_format_args` thread: the sweep landed here and at line 627, but missed two callers - `join_consumer_group` at lines 656-657 and `leave_consumer_group` at 741-742 still use the old positional `{}` style with separate args. unify to named captures.

hubcio (changes requested):

foreign/cpp/include/iggy.hpp
```diff
@@ -26,46 +26,128 @@
...
+template <typename Tag>
+class StringTag {
+  protected:
+    explicit StringTag(std::string value) : value_(std::move(value)) {}
+
+    std::string_view value() const { return value_; }
+
+  private:
+    std::string value_;
+};
```
amlel-el-mahrouss, line 49: `StringTag` should be marked `final` and have a destructor to satisfy the C++ ABI.

foreign/cpp/include/iggy.hpp
```diff
@@ -26,46 +26,128 @@
...
+    /// Use bzip2 for a higher compression ratio at the cost of slower processing.
+    static SnapshotCompression bzip2() { return SnapshotCompression("bzip2"); }
+    /// Use Zstandard for fast compression and decompression.
+    static SnapshotCompression zstd() { return SnapshotCompression("zstd"); }
+    /// Use LZMA for high compression, especially for larger files.
+    static SnapshotCompression lzma() { return SnapshotCompression("lzma"); }
+    /// Use XZ, which is similar to LZMA and often faster to decompress.
+    static SnapshotCompression xz() { return SnapshotCompression("xz"); }
+
+    std::string_view snapshot_compression_value() const { return value(); }
```
amlel-el-mahrouss, line 94: Shouldn't this be in a trait type instead?

amlel-el-mahrouss (changes requested):
Requesting the following changes for:
- StringType
- SnapshotCompression

See the comments for more info.

foreign/cpp/src/client.rs
slbotbm: The sdk now rejects empty lists and dedupes when calling rust sdk. Added an e2e test that tests rejection of empty list.

foreign/cpp/include/iggy.hpp

foreign/cpp/include/iggy.hpp
slbotbm: removed wrong documentation and made max topic size documentation more explicit.

foreign/cpp/src/identifier.rs
slbotbm: done

foreign/cpp/src/lib.rs
slbotbm: renamed done

foreign/cpp/src/client.rs
slbotbm: Using u32::MAX

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: Adding a TODO

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: Removed assertion

foreign/cpp/src/client.rs
slbotbm: documented usage of `0`.

foreign/cpp/src/client.rs
slbotbm: done

foreign/cpp/include/iggy.hpp
slbotbm: marking `StringTag` as `final` would make the rest of the public wrapper invalid. The wrapper types have final

Added destructor.

foreign/cpp/include/iggy.hpp
slbotbm: `StringTag` is partly the crtp/trait type for repeated storage behaviour. On the question of also adding getters to it, I think there is not a need currently since it would reduce clarity of the API.

hubcio (approved):

mmodzelewski (approved):

Issue #2928, opened 2026-03-12
bug(connectors): PollingMessages auto-commit commits offsets before sink processing


mlevkov:
## Description

The `AutoCommitWhen::PollingMessages` strategy commits consumer group offsets **before** `consume()` is called on the sink plugin. This means if `consume()` fails (and the failure is already silently discarded per #2927), the messages are permanently lost — the consumer group has already advanced past them.

## Sequence

```
1. consumer.next()     → messages received
2. Offsets committed   → consumer group advances  ← PROBLEM
3. process_messages()  → calls sink consume() via FFI
4. consume() fails     → messages already committed, lost forever
```

## Location

`core/connectors/runtime/src/sink.rs`:

- **Line 421**: Consumer configured with `AutoCommitWhen::PollingMessages`
  ```rust
  .auto_commit(AutoCommit::When(AutoCommitWhen::PollingMessages))
  ```

- **Lines 266-344** (`consume_messages` loop): `consumer.next()` at line 272 triggers the auto-commit, but `process_messages()` doesn't execute until line 311.

## Impact

Combined with #2927 (consume return value discarded), **at-least-once delivery is not achievable** with the current runtime for any sink connector. If `consume()` fails and the process restarts, those messages will never be retried because the offsets have already been committed.

## Suggested Fix

Change auto-commit strategy to `AutoCommitWhen::ConsumerStopped` or `AutoCommit::Disabled`, and commit offsets **after** successful `consume()`:

```rust
// Option A: Commit after processing
.auto_commit(AutoCommit::Disabled)
// ... in the consume loop, after successful consume():
consumer.store_offset(offset).await?;

// Option B: Use AfterPollingMessages if available
.auto_commit(AutoCommit::When(AutoCommitWhen::AfterProcessing))
```

The exact API depends on the Iggy SDK's consumer group offset management capabilities.

## Related

- #2927 (consume return value discarded)
- Discussed in #2919 (HTTP sink connector proposal)
- HTTP sink PR: #2925

kriti-sc:
Could I please be assigned this issue?

kriti-sc:
Related discussion: https://github.com/apache/iggy/discussions/3203

PR #3103, opened 2026-04-13
feat(connectors): add S3 sink connector

### LINKED ISSUE #2956: Amazon S3 Sink Connector
### Summary

A native Rust sink connector that writes messages from Iggy streams to Amazon S3 and S3-compatible stores (MinIO, DigitalOcean Spaces, Backblaze B2, Cloudflare R2). Listed as P0 in the connector ecosystem tracker (#2756). No existing implementation or active PR.

### Why This Matters

Without a native S3 sink, users who want to archive stream data, feed data lake pipelines, or stage for cloud warehouses (Snowflake, Redshift, Athena) must write custom consumers. The Iceberg sink (#2191) writes to S3 internally but requires a catalog server and targets the Iceberg table format, not raw file export. This connector covers the straightforward "stream messages to S3 files" use case that every competing platform (Kafka Connect, Pulsar IO, Flink) already ships.

### Scope

Sink only (Iggy to S3). A source connector (S3 to Iggy) can follow in a separate issue.

### Proposed Config

```toml
type = "sink"
key = "s3"
enabled = true
version = 0
name = "S3 sink"
path = "../../target/release/libiggy_connector_s3_sink"
verbose = false

[[streams]]
stream = "application_logs"
topics = ["api_requests", "errors"]
schema = "json"
batch_length = 1000
poll_interval = "100ms"
consumer_group = "s3_sink"

[plugin_config]
bucket = "my-data-lake"
prefix = "iggy/raw"
region = "us-east-1"
# endpoint = "http://localhost:9000"       # for MinIO / S3-compatible
# access_key_id = "AKIA..."               # omit to use instance profile / env vars
# secret_access_key = "..."               # omit to use instance profile / env vars

# File layout: {stream}, {topic}, {partition}, {date}, {hour}, {timestamp}
path_template = "{stream}/{topic}/{date}/{hour}"

# Rotation: write a new file when this threshold is hit
file_rotation = "size"                     # "size" | "time" | "messages"
max_file_size = "8MiB"
# max_file_age = "5m"
# max_messages_per_file = 10000

output_format = "json_lines"               # "json_lines" | "json_array" | "raw"
# compression = "none"                     # "none" | "gzip" | "zstd"
include_metadata = true
include_headers = true
max_retries = 3
retry_delay = "1s"
```

### Example Output

Writing `api_requests` messages on 2026-03-16 at 14:00 UTC produces:

```
s3://my-data-lake/iggy/raw/application_logs/api_requests/2026-03-16/14/000001.jsonl
```

Each line:

```json
{"offset":42,"timestamp":"2026-03-16T14:02:31Z","stream":"application_logs","topic":"api_requests","partition_id":1,"payload":{"method":"GET","path":"/api/users","status":200}}
```

### Design Questions

**1. `aws-sdk-s3` vs Apache OpenDAL?**

`aws-sdk-s3` is lighter and covers S3 + all S3-compatible stores via custom endpoint. Apache OpenDAL (#1419, with @Xuanwo already engaged) would give native GCS/Azure/HDFS support through one abstraction. Tiered storage (#1419) is a separate feature (Iggy internal persistence) but sharing the same storage layer across both could be a good long-term call. My leaning: start with `aws-sdk-s3`, migrate to OpenDAL later if adopted project-wide.

**2. Buffering: memory vs temp file?**

Messages arrive in small batches. We buffer until rotation triggers, then upload. At 8MiB default file size, in-memory buffering seems fine. Want to confirm this fits the connector runtime's expectations.

**3. Deduplication on crash recovery?**

If the connector crashes before offset commit, messages get re-delivered. Proposal: use deterministic S3 keys based on offset ranges (e.g. `.../{offset_start}-{offset_end}.jsonl`) so re-uploads are idempotent overwrites. No extra state needed.

### Implementation Plan

1. Scaffold `core/connectors/sinks/s3_sink/` following the postgres_sink pattern
2. `S3SinkConfig` struct, `S3Sink` struct implementing `Sink` trait
3. `open()`: create S3 client, HeadBucket connectivity check
4. `consume()`: buffer messages, rotate and upload when threshold hit
5. `close()`: flush remaining buffer
6. Path template rendering, compression, output formatting
7. Unit tests (config, path templates, rotation, formatting)
8. Integration test with MinIO in Docker
9. README + example config

### References

- Connector Ecosystem Tracker: #2756 (P0 priority)
- Tiered Storage / OpenDAL: #1419
- Existing sinks for reference: `core/connectors/sinks/postgres_sink/`, `core/connectors/sinks/iceberg_sink/`
- Sink trait: `core/connectors/sdk/src/lib.rs`

Please share your thoughts @hubcio @kparisa

atharvalade:

Closes #2956 

## Rationale

Iggy lacks a native way to write stream messages to Amazon S3 and S3-compatible stores (MinIO, Cloudflare R2, Backblaze B2, DigitalOcean Spaces). This is a frequently requested capability for data lake ingestion and long-term archival pipelines.

## What changed?

There was no connector for persisting Iggy messages to object storage. Users had to build custom consumers and upload logic to get data into S3.

This PR adds a new `iggy_connector_s3_sink` crate that implements the `Sink` trait. It buffers messages in-memory per stream/topic/partition, rotates files by size or message count, renders S3 keys from a configurable path template (`{stream}/{topic}/{date}/{hour}/...`), and uploads with retry + exponential backoff. Supports `json_lines`, `json_array`, and `raw` output formats with optional Iggy metadata and header embedding. Uses `rust-s3` (already in workspace) with path-style addressing auto-enabled for custom endpoints.

Key implementation details:
- **6 source modules**: `lib.rs` (config + entry point), `client.rs` (S3 client init + bucket verification), `buffer.rs` (in-memory accumulation + rotation logic), `formatter.rs` (JSON/raw output + metadata/header inclusion), `path.rs` (template engine for S3 keys with offset-based filenames), `sink.rs` (Sink trait: open/consume/close lifecycle)
- **36 unit tests** covering config deserialization, buffer rotation, path template rendering, all output formats, credential validation, and edge cases
- **CI integration**: added to `_build_rust_artifacts.yml` and `edge-release.yml` for cdylib plugin builds and release notes
- **Error handling**: warnings logged on invalid config fallbacks, explicit buffer management on upload failure, close() warns if S3 client was never initialized
- **End-to-end tested** locally with MinIO in Docker, Iggy server, CLI producer, and connector runtime — verified messages flow from Iggy stream into S3 bucket as properly formatted JSON

## Local Execution

- Passed
- Pre-commit hooks ran
- Full CI checklist passed locally:
  - `cargo fmt --check` -- pass
  - `cargo clippy --tests -D warnings` -- pass (zero warnings)
  - `cargo test -p iggy_connector_s3_sink` -- 36/36 pass
  - `markdownlint --check` -- pass
  - `trailing-whitespace` -- pass
  - `trailing-newline` -- pass
  - `license-headers` -- pass

## AI Usage

1. Opus 4.6
2. used for scaffolding boilerplate and initial file structure, all logic was reviewed and iterated manually
3. Verified through full local compilation, 36 unit tests, clippy with `-D warnings`, and end-to-end testing with MinIO Docker + Iggy server + CLI producer + connector runtime
4. Yes

---

Here are all the relevant screenshots:

- MinIO Docker container running and accessible at localhost:9000
- MinIO web console showing the created `iggy-test` bucket
- Iggy server started with root credentials configured
- Iggy CLI creating stream `application_logs` and topic `api_requests`
- Iggy CLI sending test messages to the topic
- Connector runtime loading the S3 sink plugin and connecting to MinIO
- Connector runtime consuming messages and uploading to S3
- MinIO console showing the uploaded `.jsonl` file in the correct path structure (`application_logs/api_requests/{date}/{hour}/`)
- Contents of the uploaded file showing properly formatted JSON lines with metadata (offset, timestamp, stream, topic, partition_id, payload)
- All 36 unit tests passing
- `cargo clippy --tests -D warnings` passing with zero warnings

Screenshot 2026-04-13 at 1 37 24 AM
Screenshot 2026-04-13 at 1 36 38 AM
Screenshot 2026-04-13 at 1 36 30 AM
Screenshot 2026-04-13 at 1 36 12 AM
Screenshot 2026-04-13 at 1 35 34 AM
Screenshot 2026-04-13 at 1 28 47 AM
Screenshot 2026-04-13 at 1 28 25 AM
Screenshot 2026-04-13 at 1 28 16 AM
Screenshot 2026-04-13 at 1 28 06 AM
Screenshot 2026-04-13 at 1 27 52 AM


core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+        };
+
+        let delay_str = config.retry_delay.as_deref().unwrap_or(DEFAULT_RETRY_DELAY);
+        let retry_delay = match humantime::Duration::from_str(delay_str) {
+            Ok(d) => d.into(),
+            Err(e) => {
+                tracing::warn!(
+                    "S3 sink ID: {id} invalid retry_delay '{delay_str}': {e}, defaulting to 1s",
+                );
+                std::time::Duration::from_secs(1)
```
slbotbm: Here, if an error occurs, the connector picks the default option, which may not be something that all users want. Would it be better to gate using the default options behind some kind of flag?

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,261 @@
...
+            .as_ref()
+            .ok_or_else(|| Error::InitError("S3 client not initialized".to_string()))?;
+
+        let key = BufferKey {
+            stream: topic_metadata.stream.clone(),
+            topic: topic_metadata.topic.clone(),
+            partition_id: messages_metadata.partition_id,
+        };
+
+        let max_messages = self.config.max_messages_per_file.unwrap_or(u64::MAX);
```
slbotbm: If someone chooses number of messages for file rotation and forgets to define max_messages_per_file, max messages will be set to u64::MAX, which means that processes other than the connector will force the flush to s3, which I believe is not an ideal situation.

core/connectors/sinks/s3_sink/src/client.rs
```diff
@@ -0,0 +1,154 @@
...
+        _ => Err(Error::InvalidConfigValue(
+            "Partially configured credentials. You must provide both access_key_id \
+             and secret_access_key, or omit both."
+                .to_owned(),
+        )),
+    }
+}
+
+pub async fn create_bucket(config: &S3SinkConfig) -> Result<Box<Bucket>, Error> {
+    validate_credentials(config)?;
```
slbotbm: Wouldn't it be better to inline this function? As far as I can see, it is being used in only one place. You could add a comment instead saying that this piece of code validates the creds.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+    pub fn from_str_config(s: &str) -> Result<Self, Error> {
+        match s.to_lowercase().as_str() {
+            "json_lines" | "jsonl" | "jsonlines" => Ok(OutputFormat::JsonLines),
+            "json_array" | "json" => Ok(OutputFormat::JsonArray),
+            "raw" => Ok(OutputFormat::Raw),
+            other => Err(Error::InvalidConfigValue(format!(
+                "Unknown output format: '{other}'. Expected: json_lines, json_array, or raw"
+            ))),
+        }
+    }
```
slbotbm: Wouldn't using From<String> for OutputFormat make more sense here? (though the current version also works just fine)

slbotbm (commented):
I left some comments.
Also, do you plan to support using parquet files as well in the future?

slbotbm:
I also feel data loss due to maximum retries being exceeded should be mentioned in readme.md as a precaution.

atharvalade:
> I left some comments. Also, do you plan to support using parquet files as well in the future?

oh yes absolutely.. parquet support is on the roadmap as a future `output_format` option

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Good point, I'll change these to return hard errors on invalid config

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: I'll add a validation in `open()` that returns an error if file_rotation = "messages" but `max_messages_per_file` is not set. do you think this is a good approach?

core/connectors/sinks/s3_sink/src/client.rs
atharvalade: yeah fair enough

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: I'll implement `TryFrom<&str>` for `OutputFormat` which is more idiomatic

atharvalade:
> I also feel data loss due to maximum retries being exceeded should be mentioned in readme.md as a precaution.

I agree, I'll add that

core/connectors/sinks/s3_sink/src/sink.rs
slbotbm: I think that is a good approach. In addition, how about erroring hard in `let max_messages = self.config.max_messages_per_file.unwrap_or(u64::MAX);` instead of passing `u64::MAX` as a fallback?

core/connectors/sinks/s3_sink/README.md
```diff
@@ -0,0 +1,147 @@
+# Apache Iggy S3 Sink Connector
+
+Writes messages from Iggy streams to Amazon S3 and S3-compatible object stores (MinIO, Cloudflare R2, DigitalOcean Spaces, Backblaze B2).
+
+## Features
+
+- Buffered uploads with configurable file rotation (by size or message count)
+- Multiple output formats: JSON Lines, JSON Array, Raw
+- Configurable path templates with variables for stream, topic, date, hour, partition
+- Deterministic S3 keys based on offset ranges for idempotent crash recovery
```
hubcio: "deterministic S3 keys ... idempotent crash recovery" is false on three grounds:

- `path.rs:64` substitutes `{timestamp}` via `Utc::now().timestamp_millis()` at render time, so retries produce different keys
- `buffer.rs` is in-memory only with no WAL, so a process restart cannot reproduce the same batch boundaries
- runtime auto-commits offsets on poll (#2928) and the FFI return code is discarded (#2927), so replay never happens anyway

either remove the claim, or remove `{timestamp}` from the template and document the in-memory loss path honestly.

core/connectors/sinks/s3_sink/README.md
```diff
@@ -0,0 +1,147 @@
...
+- Optional metadata and header inclusion in output
+- Support for custom endpoints (MinIO, R2) and path-style addressing
+- Retry with exponential backoff on upload failures
```
hubcio: "retry with exponential backoff" - code at `sink.rs:260` is `retry_delay * attempts`, which is linear (1s, 2s, 3s). either update the doc to "linear backoff" or implement `retry_delay * 2u32.pow(attempts - 1)` with jitter. also AFAIR there is backoff in connectors SDK, please check it.

core/connectors/sinks/s3_sink/README.md
```diff
@@ -0,0 +1,147 @@
...
+bucket = "my-bucket"
+region = "auto"
+endpoint = "https://<account-id>.r2.cloudflarestorage.com"
+access_key_id = "..."
+secret_access_key = "..."
+```
+
+## Data Delivery Guarantees
+
+This connector provides **at-least-once** delivery under normal operation. However, **data loss can occur** if all upload retries are exhausted (controlled by `max_retries`). When an upload fails after all retry attempts, the affected messages are dropped and an error is logged. Monitor your connector logs for `failed to upload` errors in production. Increase `max_retries` and `retry_delay` if transient S3 failures are common in your environment.
```
hubcio: this paragraph claims at-least-once and then admits "data loss can occur" in the same sentence - that is self-contradictory. given #2927 + #2928, no sink connector can deliver at-least-once today. the canonical in-tree wording is `http_sink/README.md:790-800` which honestly documents at-most-once and cites both bugs. recommend copying that section verbatim.

core/connectors/sinks/s3_sink/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+byte-unit = { workspace = true }
+chrono = { workspace = true }
+dashmap = { workspace = true }
+humantime = { workspace = true }
+iggy_common = { workspace = true }
```
hubcio, line 42: `iggy_common` is declared under `[dependencies]` but the only usage in this crate is `formatter.rs:233-235` under `#[cfg(test)]`. move it to `[dev-dependencies]` so it does not bloat the cdylib build.

core/connectors/sinks/s3_sink/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_s3_sink"
```
hubcio, line 19: missing `publish = false`. every other connector sink (`postgres_sink`, `delta_sink`, `http_sink`, `elasticsearch_sink`, `stdout_sink`) declares `publish = false` because they are cdylib plugins not meant for crates.io. without it the crate would publish on the next workspace release.

core/connectors/sinks/s3_sink/src/buffer.rs
```diff
@@ -0,0 +1,168 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use crate::FileRotation;
+
+#[derive(Debug)]
+pub struct FileBuffer {
+    entries: Vec<Vec<u8>>,
```
hubcio: `entries: Vec<Vec<u8>>` is one heap alloc per message on append, and `finalize_buffer` (`formatter.rs:128-152`) then does `extend_from_slice` into a fresh `Vec<u8>` which is a second full copy of every byte. for default `batch_length=1000` that is 1000 small allocs plus a 2x peak-memory copy per file.

single contiguous `Vec<u8>` + sidecar `Vec<usize>` for entry boundaries gets this to zero extra copies on finalize and enables future `put_object_stream` use.

core/connectors/sinks/s3_sink/src/client.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+    Ok(())
+}
+
+pub async fn create_bucket(config: &S3SinkConfig) -> Result<Box<Bucket>, Error> {
+    validate_credential_pair(config)?;
+
+    let credentials = match (&config.access_key_id, &config.secret_access_key) {
+        (Some(key), Some(secret)) => {
+            let redacted_key = key.chars().take(3).collect::<String>();
```
hubcio: `key.chars().take(3)` extracts the AWS key category prefix (`AKIA` for IAM user, `ASIA` for STS, `AROA` for role, etc) - it reveals roughly the credential type and zero entropy of the secret material, so it has no diagnostic value either. log `key.len()` or the last 4 chars if you want something useful.

core/connectors/sinks/s3_sink/src/client.rs
```diff
@@ -0,0 +1,154 @@
...
+    let use_path_style = config.path_style.unwrap_or(config.endpoint.is_some());
+    if use_path_style {
+        bucket.set_path_style();
+    }
+
+    Ok(bucket)
+}
+
+pub async fn verify_bucket(bucket: &Bucket) -> Result<(), Error> {
+    bucket.head_object("/").await.map_err(|e| {
```
hubcio: `head_object("/")` is not a bucket connectivity check - any working bucket returns 404 for a key named `/`, and `sink.rs:49-55` then swallows the error as `warn!` and continues. real failures (wrong region, `SignatureDoesNotMatch`, missing bucket, network unreachable) are all masked until the first `put_object` inside the consume loop blows up.

rust-s3 0.37 has no `head_bucket()`. use `bucket.list_page("".into(), None, None, None, Some(1)).await` (requires `s3:ListBucket`) or `bucket.location()` (requires `s3:GetBucketLocation`), and treat the error as fatal in `open()`.

core/connectors/sinks/s3_sink/src/formatter.rs
```diff
@@ -0,0 +1,287 @@
...
+        obj.insert("timestamp".to_string(), Value::String(ts));
+        obj.insert(
+            "stream".to_string(),
+            Value::String(topic_metadata.stream.clone()),
+        );
+        obj.insert(
+            "topic".to_string(),
+            Value::String(topic_metadata.topic.clone()),
+        );
+        obj.insert(
```
hubcio: `topic_metadata.stream.clone()` + `topic_metadata.topic.clone()` are per-message clones inside `format_json_message`. for default `batch_length=1000` that is 2000 string clones per consume call. switching this function to a `derive(Serialize)` struct that holds `&str` references would eliminate this plus the 5 const-key `.to_string()` allocs at `:55-69`.

core/connectors/sinks/s3_sink/src/formatter.rs
```diff
@@ -0,0 +1,287 @@
...
+            "partition_id".to_string(),
+            Value::Number(messages_metadata.partition_id.into()),
+        );
+    }
+
+    if include_headers && let Some(headers) = &message.headers {
+        let mut headers_obj = Map::new();
+        for (key, value) in headers {
+            headers_obj.insert(key.to_string(), Value::String(value.to_string()));
```
hubcio: `value.to_string()` invokes the `Display` impl on `HeaderField<T>` at `core/common/src/types/message/user_headers.rs:343-348`, which writes `"{kind}: {value}"`. emitted JSON becomes `{"content-type":"string: text/plain"}`, and for `HeaderKind::Raw` the value at `user_headers.rs:529` is `format!("{:?}", self.value)` - a debug-formatted `Vec<u8>` like `"raw: [104, 105]"`.

also note `HeaderKey = HeaderField<KeyMarker>` shares the same `Display`, so `key.to_string()` produces `"string: content-type"` as the JSON map key - both halves of the header object are mangled. downstream JSON parsers cannot recover the original header.

fix is a `HeaderKind` match: string -> raw string, raw -> base64, integer/float -> `Value::Number`, bool -> `Value::Bool`.

core/connectors/sinks/s3_sink/src/formatter.rs
```diff
@@ -0,0 +1,287 @@
...
+    let payload_value = payload_to_json_value(&message.payload);
+    obj.insert("payload".to_string(), payload_value);
+
+    match serde_json::to_vec(&Value::Object(obj)) {
+        Ok(bytes) => bytes,
+        Err(e) => {
+            tracing::warn!(
+                "Failed to serialize message at offset {}: {e}",
+                message.offset
+            );
```
hubcio: on `serde_json::to_vec` failure this returns `Vec::new()`, which `sink.rs:103` then appends with the original offset and timestamp. `finalize_buffer` for `JsonLines` emits an empty line for it (`{}\n` -> just `\n`), and for `JsonArray` the `filter_map(... .ok())` at `:139` silently drops it. either way the offset is counted as processed while the message data is gone.

same shape at `:96` for `Raw` (`try_to_bytes().unwrap_or_default()`). this should propagate `Result` and either skip + log + lost-counter or fail the batch.

core/connectors/sinks/s3_sink/src/formatter.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+fn format_raw_message(message: &ConsumedMessage) -> Vec<u8> {
+    message.payload.try_to_bytes().unwrap_or_default()
+}
+
+fn payload_to_json_value(payload: &Payload) -> Value {
+    match payload {
+        Payload::Json(value) => {
+            let bytes = simd_json::to_vec(value).unwrap_or_default();
```
hubcio: `simd_json::to_vec(value).unwrap_or_default()` -> `serde_json::from_slice(&bytes).unwrap_or(Value::Null)` is a full round-trip per JSON message with two silent-corruption unwraps. `sdk/src/lib.rs:50` already re-exports `convert::owned_value_to_serde_json` which is a direct structural converter - elasticsearch_sink, delta_sink and http_sink all use it (`sdk/src/convert.rs:29`). one-line replacement fixes both correctness and the perf round-trip.

core/connectors/sinks/s3_sink/src/formatter.rs
```diff
@@ -0,0 +1,287 @@
...
+            for entry in data {
+                result.extend_from_slice(entry);
+                result.push(b'\n');
+            }
+            result
+        }
+        OutputFormat::JsonArray => {
+            let entries: Vec<Value> = data
+                .iter()
+                .filter_map(|bytes| serde_json::from_slice(bytes).ok())
```
hubcio: JsonArray finalize parses every already-serialized entry to `Value`, collects to `Vec<Value>`, then `serde_json::to_vec` re-serializes the array - triple work. also `filter_map(... .ok())` silently drops any entry that fails to parse.

entries are already valid JSON. write `b"["`, comma-join with `extend_from_slice`, write `b"]"`. pure byte concat, no parse, no silent drop.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+use iggy_connector_sdk::{Error, sink_connector};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::fmt;
+
+pub mod buffer;
+pub mod client;
+pub mod formatter;
+pub mod path;
+pub mod sink;
```
hubcio: `pub mod buffer`, `pub mod client`, etc. expose implementation modules as crate-level public API for a cdylib plugin. `stdout_sink`, `postgres_sink`, `elasticsearch_sink` all keep these helpers private. drop the `pub`.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+
+sink_connector!(S3Sink);
+
+const DEFAULT_MAX_RETRIES: u32 = 3;
+const DEFAULT_RETRY_DELAY: &str = "1s";
+const DEFAULT_MAX_FILE_SIZE: &str = "8MiB";
+const DEFAULT_PATH_TEMPLATE: &str = "{stream}/{topic}/{date}/{hour}";
```
hubcio, line 37: the default `path_template` has no `{partition}` placeholder, and `path.rs:40-44` hardcodes the filename as `{first_offset:06}-{last_offset:06}.{ext}` with no partition_id either. Iggy partition offsets are independent (both start at 0), so two partitions on the same `stream`/`topic`/`hour` write to the identical S3 key and the last writer wins - silent data loss on the shipped default with no misconfiguration required.

fix needs to inject `partition_id` into the FILENAME (not just the template), so users can't bypass it by customizing `path_template`. suggestion: `{partition:05}-{first:020}-{last:020}.{ext}`.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+const DEFAULT_OUTPUT_FORMAT: &str = "json_lines";
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
hubcio: `#[derive(Debug)]` on `S3SinkConfig` plus plain `Option<String>` for `access_key_id` / `secret_access_key` means any `{:?}` log of the config (or of `S3Sink` at `:141`, which transitively derives `Debug` via this field) prints credentials in cleartext.

`postgres_sink/src/lib.rs:25,52` uses `secrecy::SecretString` with `iggy_common::serde_secret::serialize_secret` for exactly this reason. this is a regression vs the in-tree baseline.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+    Raw,
+}
+
+impl TryFrom<&str> for OutputFormat {
+    type Error = Error;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        match s.to_lowercase().as_str() {
+            "json_lines" | "jsonl" | "jsonlines" => Ok(OutputFormat::JsonLines),
+            "json_array" | "json" => Ok(OutputFormat::JsonArray),
```
hubcio: `"json"` aliases to `JsonArray` here, but streaming users defaulting their config to `output_format = "json"` will reasonably expect ndjson (`JsonLines`). drop the ambiguous alias - keep `json_array` / `json_lines` / `jsonl` / `raw` and reject anything else.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+            .retry_delay
+            .as_deref()
+            .unwrap_or(DEFAULT_RETRY_DELAY);
+        self.retry_delay = humantime::Duration::from_str(delay_str)
+            .map(|d| d.into())
+            .map_err(|e| {
+                Error::InvalidConfigValue(format!("Invalid retry_delay '{delay_str}': {e}"))
+            })?;
+
+        if self.config.file_rotation == FileRotation::Messages
```
hubcio: validation gaps here cause real bugs:

- `max_messages_per_file = Some(0)` is accepted; `buffer.rs:59` then has `count >= 0` always true after the first append, so rotation triggers on every message -> one PutObject per message -> S3 PUT throttling at ~3500 RPS per partitioned prefix + cost blowup.
- `max_file_size = "0B"` parses cleanly (`parse_file_size` returns `Ok(0)`); `buffer.rs:58` has the same `>= 0` problem.
- empty `bucket` / empty `region` / empty `path_template` are all accepted.
- `max_file_size` is uncapped on the high end. rust-s3 0.37 `put_object` does not auto-multipart, and S3 rejects single PUT > 5 GiB with `EntityTooLarge`. a user setting `max_file_size = "10GiB"` gets opaque retry-exhaust + silent loss per the `buffer.reset` path at `sink.rs:212-214`.

reject zero, reject empty, cap `<= 5 GiB`.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,369 @@
...
+
+        Ok(())
+    }
+
+    fn max_retries(&self) -> u32 {
+        self.config.max_retries.unwrap_or(DEFAULT_MAX_RETRIES)
+    }
+}
+
+use std::str::FromStr;
```
hubcio: `use std::str::FromStr;` mid-file after the impl block. move to the top with the other use statements.

core/connectors/sinks/s3_sink/src/path.rs
```diff
@@ -0,0 +1,164 @@
...
+    prefix: Option<&str>,
+    template: &str,
+    ctx: &PathContext<'_>,
+    offset_start: u64,
+    offset_end: u64,
+    format: OutputFormat,
+) -> String {
+    let rendered = render_template(template, ctx);
+
+    let filename = format!(
```
hubcio, line 41: `{:06}-{:06}` zero-pads up to 6 digits but does not cap - Iggy offsets are u64. at offset `1_000_000` the rendered name is `"1000000-1001000.jsonl"` (7 chars) and at offset `999900` it is `"999900-999999.jsonl"` (6 chars). lex sort compares byte-by-byte and `'1' < '9'`, so the million-offset file sorts BEFORE the 999k file in any S3 ListObjectsV2 listing.

Athena partition projection, Spark, and `aws s3 ls` all rely on lex order. at default 1000-msg batches and modest throughput, offset 1M is a few minutes in. pad to 20 decimal digits (max u64 = 18446744073709551615) or 16 hex digits.

core/connectors/sinks/s3_sink/src/path.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+        None => format!("{rendered}/{filename}"),
+    }
+}
+
+fn render_template(template: &str, ctx: &PathContext<'_>) -> String {
+    let dt = timestamp_to_datetime(ctx.first_timestamp_micros);
+    let date = dt.format("%Y-%m-%d").to_string();
+    let hour = dt.format("%H").to_string();
+    let now_millis = Utc::now().timestamp_millis().to_string();
```
hubcio: `{timestamp}` substitutes `Utc::now().timestamp_millis()`, which is wall-clock at render time and therefore non-deterministic across retries. combined with the README idempotency claim at line 10, this is the main reason re-uploads do not overwrite-deduplicate.

derive from `ctx.first_timestamp_micros` if a deterministic timestamp is useful, or drop the variable entirely.

core/connectors/sinks/s3_sink/src/path.rs
```diff
@@ -0,0 +1,164 @@
...
+        .replace("{partition}", &ctx.partition_id.to_string())
+        .replace("{date}", &date)
+        .replace("{hour}", &hour)
+        .replace("{timestamp}", &now_millis)
+}
+
+fn timestamp_to_datetime(micros: u64) -> DateTime<Utc> {
+    let secs = (micros / 1_000_000) as i64;
+    let nanos = ((micros % 1_000_000) * 1_000) as u32;
+    DateTime::<Utc>::from_timestamp(secs, nanos).unwrap_or_else(Utc::now)
```
hubcio: `from_timestamp(...).unwrap_or_else(Utc::now)` silently falls back to wall-clock on an out-of-range Iggy timestamp. result: path partitioning silently switches from message-time-based to ingestion-time-based without any signal to the operator. either return an error or use `DateTime::UNIX_EPOCH`.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        let mut buffers = self.buffers.lock().await;
```
hubcio: `self.buffers.lock().await` holds a `tokio::sync::Mutex<HashMap<BufferKey, FileBuffer>>` guard that lives for the entire body of `consume` - through the for-loop at `:94-112`, through `flush_buffer` at `:110`, through `upload_with_retry`, through `bucket.put_object().await` at `:230` and through `tokio::time::sleep(retry_delay * attempts).await` at `:260`. one tokio task is spawned per IggyConsumer (`runtime/sink.rs:211-216`), and all of them route to the same `S3Sink` instance via `INSTANCES` (`sdk/sink.rs:229,241`), so this single mutex serializes uploads across every partition of every stream/topic this sink instance handles.

throughput ceiling = single-uploader latency. with S3 PutObject p50 50ms and p99 200ms, that is 5-20 PUTs/sec/instance regardless of partition count. worst-case mutex hold under failure with default retry_delay=1s + max_retries=3 is 3s blocking every partition.

notably `Cargo.toml:42` already declares `dashmap` in the dependency list (currently unused, sitting in the `cargo-machete.ignored` set). the fix is `DashMap<BufferKey, Arc<Mutex<FileBuffer>>>` - clone the Arc, drop the outer guard before `flush_buffer`. `delta_sink/src/sink.rs:126-128` has the same anti-pattern with a TODO referencing PR #2889 discussion.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+                self.output_format,
+            );
+            buffer.append(formatted, message.offset, message.timestamp);
+
+            if buffer.should_rotate(
+                self.config.file_rotation,
+                self.max_file_size_bytes,
+                max_messages,
+            ) {
+                self.flush_buffer(bucket, &key, buffer).await;
```
hubcio: mid-batch flush failure produces an invisible offset gap in S3. trace: append offsets 0..99 -> `flush_buffer` fails (per `:212-214` resets buffer + logs error + returns) -> loop continues -> append offsets 100..199 -> next flush succeeds with `000100-000199.jsonl`. there is no `000000-000099.jsonl` and nothing in the listing tells a downstream consumer that range existed.

propagate `Err` from `flush_buffer` to `consume` so the runtime sees the failure (which itself only gets honored once #2927 is fixed). as a stopgap, write a `.lost` marker object describing the missing offset range.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+            }
+        }
+
+        let mut state = self.state.lock().await;
+        state.messages_processed += messages.len() as u64;
```
hubcio: `state.messages_processed += messages.len() as u64` runs unconditionally outside the rotate loop, so when a mid-batch flush dropped N messages this counter still claims it processed them. `upload_errors` increments separately at `:209`. result: the close-log at `:155-156` reports more messages processed than actually landed in S3. either decrement on drop or split into `messages_buffered` / `messages_uploaded` / `messages_lost`.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+                    "S3 sink ID: {} failed to upload {} ({} messages lost): {e}",
+                    self.id, s3_key, msg_count
+                );
+                let mut state = self.state.lock().await;
+                state.upload_errors += 1;
+                drop(state);
+                // Reset buffer even on failure to prevent unbounded growth.
+                // Messages are lost but offsets will be re-delivered by the
+                // runtime on next poll since consume() returned Ok.
+                buffer.reset();
```
hubcio: this is the worst data-loss path in the PR. on retry-exhaust the failure branch logs an error and then calls `buffer.reset()` at `:214` to drop the messages, while `consume` at `:126` still returns `Ok(())`. the comment at `:211-213` claims "offsets will be re-delivered by the runtime on next poll since consume() returned Ok" - that is doubly false:

- the runtime auto-commits offsets *before* consume runs (`runtime/sink.rs:421` uses `AutoCommitWhen::PollingMessages`, see #2928), so the offset has already advanced past these messages by the time you return.
- the runtime discards the consume FFI return code (`runtime/sink.rs:585-593`, see #2927), so even returning `Err` wouldn't trigger retry today.

net result: a single transient S3 hiccup that exhausts `max_retries` (default 3) permanently loses every buffered message. the README at line 139 acknowledges this but still claims at-least-once on the same line.

minimum: drop the false comment, propagate `Err` to the runtime, and align the README with `http_sink/README.md:790-800` (at-most-once + cite #2927 / #2928).

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+
+        loop {
+            match bucket.put_object(s3_key, data).await {
+                Ok(response) => {
+                    let status = response.status_code();
+                    if (200..300).contains(&status) {
+                        return Ok(());
+                    }
+                    attempts += 1;
+                    if attempts >= max_retries {
```
hubcio: `attempts` starts at 0 and increments before the `>= max_retries` check, so `max_retries = 3` yields 3 total attempts (2 retries past the initial one). this matches the postgres_sink pattern but the field name is misleading. either rename to `max_attempts` or use `attempts > max_retries` so the field name lines up with semantics.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+        bucket: &s3::Bucket,
+        s3_key: &str,
+        data: &[u8],
+    ) -> Result<(), Error> {
+        let max_retries = self.max_retries();
+        let retry_delay = self.retry_delay;
+        let mut attempts = 0u32;
+
+        loop {
+            match bucket.put_object(s3_key, data).await {
```
hubcio, line 358: the retry loop treats every non-2xx status uniformly. AWS permanent failures - `AccessDenied` (403), `NoSuchBucket` (404), `InvalidBucketName` (400), `MalformedPolicy` - get retried 3 times, wasting `retry_delay * (1 + 2)` = 3s before the final `Err`. classify retriable vs not: 5xx + 408 + 429 + 503 SlowDown -> retry; other 4xx -> fail fast.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,263 @@
...
+                            "S3 PutObject failed after {attempts} attempts for key '{s3_key}': {e}"
+                        )));
+                    }
+                    warn!(
+                        "S3 sink ID: {} PutObject error (attempt {attempts}/{max_retries}): {e}. Retrying...",
+                        self.id
+                    );
+                }
+            }
+            tokio::time::sleep(retry_delay * attempts).await;
```
hubcio: `retry_delay * attempts` is linear backoff (1s, 2s, 3s), but the README at line 13 advertises exponential. either implement `retry_delay * 2u32.pow(attempts - 1)` with jitter, or update the doc to "linear".

separately, `Duration::Mul<u32>` panics on overflow - default config is safe but pathological values (e.g. `retry_delay = "1h"` + large `max_retries`) would panic. `saturating_mul` is cheap defensive practice.

hubcio (changes requested):

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Done, `consume()` now hard-errors if `file_rotation` = "messages" and `max_messages_per_file` is unset; the `u64::MAX` fallback only applies when rotation mode is `Size` (where the value is never checked)

core/connectors/sinks/s3_sink/README.md
atharvalade: Removed claim, documented honestly instead

core/connectors/sinks/s3_sink/README.md
atharvalade: Now uses SDK `exponential_backoff` + `jitter`

core/connectors/sinks/s3_sink/README.md
atharvalade: Replaced with `http_sink's` at-most-once wording, citing #2927/#2928

core/connectors/sinks/s3_sink/Cargo.toml
atharvalade: Moved to `[dev-dependencies]`

core/connectors/sinks/s3_sink/Cargo.toml
atharvalade: Added, matches all other sinks

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Split into `messages_received` / `messages_uploaded` / `messages_lost`

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Removed false re-delivery comment, propagate error, aligned README with known runtime limitations

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Renamed to `max_attempts` for clarity

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Added `is_retriable_status`: only 5xx/408/429 retry, other 4xx fail fast

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Implemented true exponential with jitter, capped at 60s via SDK helpers

core/connectors/sinks/s3_sink/src/buffer.rs
atharvalade: switched to contiguous `Vec<u8>` + sidecar `Vec<usize>` boundaries, zero-copy finalize

core/connectors/sinks/s3_sink/src/client.rs
atharvalade: Now logging `key.len()` only, no prefix chars exposed

core/connectors/sinks/s3_sink/src/client.rs
atharvalade: Replaced with `bucket.list_page(..., Some(1))`, failure is now fatal in `open()`

core/connectors/sinks/s3_sink/src/formatter.rs
atharvalade: Refactored to `#[derive(Serialize)]` struct with `&str` references, zero string clones per message

core/connectors/sinks/s3_sink/src/formatter.rs
atharvalade: Fixed: `HeaderKind` match dispatches correctly, string as-is, raw as base64, integers as `Number`, floats as `Number`, bool as `Bool`

core/connectors/sinks/s3_sink/src/formatter.rs
atharvalade: `format_message` now returns `Result<Vec<u8>, Error>`, propagated to `consume()`

core/connectors/sinks/s3_sink/src/formatter.rs
atharvalade: Replaced with `owned_value_to_serde_json`, single structural conversion

core/connectors/sinks/s3_sink/src/formatter.rs
atharvalade: Pure byte concat now: `[` + comma-joined entries + `]`, no parse/reserialize

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Dropped all `pub`, modules are now crate-private

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Partition ID is now mandatory in the filename: `{partition:05}-{first:020}-{last:020}.{ext}`

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Fields changed to `Option<SecretString>`, custom `Debug` impl prints `[REDACTED]`

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Removed. Only explicit `json_array` / `json_lines` / `jsonl` / `raw` accepted

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Added: reject `max_messages_per_file = 0`, reject `max_file_size = "0B"`, reject empty bucket/region/path_template, cap at 5 GiB

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Moved to top-level imports

core/connectors/sinks/s3_sink/src/path.rs
atharvalade: Padded to 20 decimal digits (covers full u64 range for correct lex sort)

core/connectors/sinks/s3_sink/src/path.rs
atharvalade: Now derived from `ctx.first_timestamp_micros` deterministic across retries

core/connectors/sinks/s3_sink/src/path.rs
atharvalade: Now returns `Err` on out-of-range timestamp, propagated to caller

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Lock is now dropped before `flush_buffer` where possible; flush errors propagate to `consume()`. (`DashMap` conversion deferred pending #2889 pattern stabilization in `delta_sink`.)

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: `flush_buffer` now propagates `Err` to `consume()`, and writes a `.lost` marker object with offset range metadata for gap visibility

core/connectors/sinks/s3_sink/Cargo.toml
atharvalade: Correction to my earlier reply: `iggy_common` remains in `[dependencies]` (not `[dev-dependencies]`) because the header-serialization fix (your comment on `formatter.rs:75`) introduced runtime usage of `HeaderKey`, `HeaderValue`, and `HeaderKind` via `serialize_headers` at `formatter.rs:124-169`. The original premise (test-only) was accurate for v1 but is no longer true after the `HeaderKind` match dispatch was added

atharvalade:
The biggest thing I caught was a credential leak where the derived Debug on S3Sink would dump the full AWS access key and secret key into logs whenever the struct got printed. I replaced that with a manual Debug impl that just shows the bucket name and added a regression test so it never sneaks back in. On the metrics side, the received count was being incremented after processing the batch, which meant if something failed halfway through you could lose track of how many messages actually got dropped. I moved that counter up front and added logic to correctly mark the unprocessed remainder as lost. I also cleaned up the dependency situation since `once_cell` was listed but never used, `simd-json `was only needed in tests but lived in regular deps, and the `cargo-machete` ignore list was papering over all of it. 

The `max_retries` field got renamed to `max_attempts` with a serde alias so existing configs still work, and the README had a contradiction where it said timestamp was wall clock time when the code actually derives it from the first message in the buffer, so I fixed the docs to match reality. For the lost marker file that records data gaps, I routed it through the retry logic so a transient S3 hiccup does not silently eat your gap record. Then I wrote about twenty unit tests covering the pure functions like retry status classification, flush payload extraction, config validation, and the Debug redaction. Finally I stood up a full integration test suite using MinIO in a container that validates the sink actually writes `jsonl` to S3 with the correct path layout and rotates files properly when the message count threshold is hit.

/ready

core/connectors/sinks/s3_sink/src/formatter.rs
```diff
@@ -0,0 +1,375 @@
...
+        .unwrap_or_else(|| "1970-01-01T00:00:00Z".to_string())
+}
+
+/// Finalize buffer entries into the output byte format.
+/// For JsonArray: byte-concatenation (no re-parse). For JsonLines/Raw: newline-separated.
+pub(crate) fn finalize_buffer<'a>(
+    entries: impl Iterator<Item = &'a [u8]>,
+    format: OutputFormat,
+) -> Vec<u8> {
+    match format {
```
ryerraguntla, line 199: OutputFormat::Raw shares the same branch as OutputFormat::JsonLines in finalize_buffer, appending \n after every entry:

```
OutputFormat::JsonLines | OutputFormat::Raw => {
    result.extend_from_slice(entry);
    result.push(b'\n');    // inserted after every raw binary entry
}
```
For text payloads this is benign. For arbitrary binary data (protobufs, images, compressed data), inserting 0x0A after each message payload corrupts the resulting file. A downstream reader cannot distinguish the content byte 0x0A from the separator. Raw entries should be concatenated without any delimiter.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,478 @@
...
+}
+
+#[derive(Debug)]
+struct SinkState {
+    messages_received: u64,
+    messages_uploaded: u64,
+    messages_lost: u64,
+}
+
+impl S3Sink {
```
ryerraguntla, line 223: [nit] S3Sink::new — resolved config fields (max_file_size_bytes, max_messages, output_format, retry_delay) are initialized to zero/defaults in new() and only populated by validate_and_parse_config() in open():

```
max_file_size_bytes: 0,       // → should_rotate fires immediately on first byte
max_messages: u64::MAX,
output_format: OutputFormat::JsonLines,
retry_delay: std::time::Duration::from_secs(1),
```
The SDK contract guarantees open() precedes consume(), so this is not a runtime risk, but the type system does not enforce it. A ResolvedS3SinkConfig struct (constructed only inside open()) would make the invariant structural.

core/connectors/sinks/s3_sink/src/lib.rs
```diff
@@ -0,0 +1,478 @@
...
+                return Err(Error::InvalidConfigValue(
+                    "max_messages_per_file must be greater than 0".to_owned(),
+                ));
+            }
+            self.max_messages = n;
+        }
+
+        Ok(())
+    }
+
```
ryerraguntla, line 321: **[nit]** max_attempts — returns self.config.max_attempts.unwrap_or(DEFAULT_MAX_ATTEMPTS) with DEFAULT_MAX_ATTEMPTS = 3. The semantics in upload_with_retry are "total attempts" (attempt >= max_attempts → fail), so max_attempts=3 gives 3 total attempts (2 retries). Field name max_attempts is consistent with that, but #[serde(alias = "max_retries")] creates an alias that users read as "3 retries" = 4 total. Document which semantics apply.

core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -0,0 +1,657 @@
...
+                        return Err(Error::CannotStoreData(format!(
+                            "S3 PutObject failed after {max_attempts} attempts for key '{s3_key}': {e}"
+                        )));
+                    }
+                    warn!(
+                        "S3 sink ID: {} PutObject error (attempt {attempt}/{max_attempts}): {e}. Retrying...",
+                        self.id
+                    );
+                }
+            }
```
ryerraguntla, line 394: **[nit]**     backoff delay uses attempt - 1:

`let delay = jitter(exponential_backoff(base_delay, attempt - 1, MAX_BACKOFF));`
attempt is incremented before this line (L368 or L380), so attempt >= 1 and no underflow occurs. However, attempt - 1 is converting "attempt count" to a 0-based retry index. exponential_backoff's attempt parameter is 0-based; passing attempt - 1 is correct but non-obvious. A comment or renaming (retry_index) would prevent future regression.

core/connectors/sinks/s3_sink/src/path.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    Ok(key)
+}
+
+fn render_template(template: &str, ctx: &PathContext<'_>) -> Result<String, Error> {
+    let dt = timestamp_to_datetime(ctx.first_timestamp_micros)?;
+    let date = dt.format("%Y-%m-%d").to_string();
+    let hour = dt.format("%H").to_string();
+    let ts_millis = (ctx.first_timestamp_micros / 1_000).to_string();
+
```
ryerraguntla, line 69: warn - render_template — {stream} and {topic} are interpolated verbatim with no sanitization:

.```
replace("{stream}", ctx.stream)
.replace("{topic}", ctx.topic)
```
Iggy stream/topic names may contain characters that produce unexpected S3 key structures (e.g., names with // or non-ASCII). S3 does not interpret .. as path traversal, but unusual names produce keys that are difficult to list or match with glob patterns. Sanitizing to [a-zA-Z0-9_-] or percent-encoding would make keys predictable.

File: core/connectors/sinks/s3_sink/src/client.rs [hunk 39]
```diff
@@ -0,0 +1,168 @@
...
+
+    let use_path_style = config.path_style.unwrap_or(config.endpoint.is_some());
+    if use_path_style {
+        bucket.set_path_style();
+    }
+
+    Ok(bucket)
+}
+
+/// Verify bucket connectivity by listing zero objects. Fatal on failure.
```
ryerraguntla: **[nit]** verify_bucket — uses bucket.list_page(...) to verify connectivity:

```
bucket
    .list_page("".to_string(), None, None, None, Some(1))
    .await
```
This requires s3:ListBucket IAM permission. Write-only IAM policies (common for data-pipeline roles) that grant only s3:PutObject will fail open() with InitError. A HeadBucket call (or a zero-byte PutObject to a probe key) requires only s3:GetBucketLocation or s3:PutObject. Consider bucket.head_object or simply skip verification and let the first put_object call surface the error.

ryerraguntla: @atharvalade - Looks good. The one in formatter.rs with 0x0A will cause data issues. Others are warnings and nits.

core/connectors/sinks/s3_sink/src/formatter.rs
atharvalade: good catch, split `Raw` into its own `finalize_buffer` arm that concatenates entries without any delimiter, added `finalize_raw_no_delimiter` test with embedded 0x0A bytes to prevent regression

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: Added doc comment on the field clarifying "total attempts" semantics: max_attempts = 3 means 1 initial try + 2 retries, and the `max_retries` alias follows the same meaning

core/connectors/sinks/s3_sink/src/sink.rs
atharvalade: Extracted let `retry_index` = attempt - 1 with a comment explaining `exponential_backoff` expects a 0-based index

core/connectors/sinks/s3_sink/src/path.rs
atharvalade: Added `sanitize_key_segment()` that replaces anything outside [a-zA-Z0-9._-] with _, applied to {stream} and {topic} interpolation, with a test covering // and spaces

core/connectors/sinks/s3_sink/src/client.rs
atharvalade: thanks for catching these. I Replaced `list_page` with a zero-byte `put_object` to `.iggy-sink-probe` (best-effort `delete_object` after), so write-only IAM policies that only grant `s3:PutObject `no longer fail on `open()`

core/connectors/sinks/s3_sink/src/lib.rs
atharvalade: resolved fields (`max_file_size_bytes`, `max_messages`, `output_format`, `retry_delay`, `max_attempts`) now live in `ResolvedConfig`, constructed only inside `validate_and_parse_config()` called from `open()`; accessing before open panics with a clear `BUG` message

ryerraguntla:
@atharvalade  - Looks good. Thanks for taking car of it. @hubcio , @numinnex @Standing-Man PR looks good except for few nits. Could one of you review and take the next steps.

ryerraguntla (approved):
approve and need a second review before merge.

ryerraguntla (approved):

hubcio (approved):

numinnex (approved):

spetz (approved):

PR #3108, opened 2026-04-13
feat(consensus): add session registration protocol with combined login+register

krishvishal:
## Summary
   
- Add `Operation::Register` as a VSR-level operation that goes through consensus to create durable client sessions, giving each client a deterministic session number (= commit op number) agreed upon by all replicas.
- Add `session` field to `RequestHeader` with wire-layer validation: register requires `session=0, request=0`; all other operations require `session>0, request>0`.
- Split `ClientTable` into session-aware paths: `check_register()` / `commit_register()` for registration, and `check_request()` / `commit_reply()` with session + strict request monotonicity validation (`NoSession`, `SessionMismatch`, `RequestGap).
- Add `ConsensusSession` (SDK) for client-side session lifecycle and `SessionManager` (server-ng) for transport-to-consensus bridging.
- Add combined `login_register` handler: phase 1 verifies credentials locally, phase 2 submits register through consensus.
- Add `RequestFrame2`/`ResponseFrame2` with `request_id` for request-response correlation and pipelining support.


numinnex:
I've looked through this PR and I wonder how does it differ from the login and login with personal access token ?

krishvishal:
@numinnex I've only implemented login functionality with register but didn't implement login with personal access token. But its easily implementable because register doesn't care about how we logged in.

core/consensus/src/plane_helpers.rs
```diff
@@ -68,6 +75,48 @@ where
...
+            // Synthesize a register reply with the existing session.
+            // The caller can extract session from reply.header().commit.
+            tracing::debug!(
+                client_id,
+                session,
+                "register_preflight: client already registered, ignoring"
+            );
+            None
+        }
+        RequestStatus::InProgress => None,
```
atharvalade, line 104: on `AlreadyRegistered`, the code returns `None` without sending the existing-session reply it says it will synthesize.  This can black-hole duplicate `Register` retries (lost reply/timeout path), leaving clients stuck waiting instead of getting a deterministic session response.

core/metadata/src/impls/metadata.rs
```diff
@@ -311,9 +313,18 @@ where
...
+        // Normal metadata ops use request_preflight (check_request with session).
+        let preflight_ok = if operation == Operation::Register {
+            register_preflight(consensus, client_id).await.is_some()
+        } else {
+            request_preflight(consensus, client_id, session, request)
+                .await
+                .is_some()
         };
+        if !preflight_ok {
+            return;
```
atharvalade, line 326: `register_preflight(...).is_some()` gates progress, and `None` returns early with no response to the client. Together with `register_preflight` behavior, duplicate register requests will be dropped silently

atharvalade (changes requested):

core/consensus/src/plane_helpers.rs
krishvishal: This is done on purpose since SDK is not wired up completely.

core/metadata/src/impls/metadata.rs
krishvishal: Same as above.

numinnex (approved):

spetz (approved):

PR #3135, opened 2026-04-15
fix(server): warn and skip non-numeric filenames in consumer offset directories instead of panicking

### LINKED ISSUE #3132: server: panic on non-numeric filenames in consumer offset directories
When loading consumer offsets at startup, the server panics if any file in the consumer offset directory has a non-numeric filename.

**File:** `core/server/src/streaming/partitions/storage.rs` lines 183-186, 247-252
```rust
let name = dir_entry.file_name().into_string().unwrap();
let consumer_id = name.parse::<u32>().unwrap_or_else(|_| {
    panic!("Invalid consumer ID file with name: '{}'.", name);
});
```

An operator accidentally placing a file (e.g. .DS_Store, editor swap file, backup) in the consumer offsets directory crashes the server on startup. This should be a warning + skip, not a panic.

I think we should log a warning and skip non-numeric filenames instead of panicking.

atharvalade:

Closes #3132

## Rationale

An operator accidentally placing a file (e.g. `.DS_Store`, editor swap file, backup) in the consumer offsets directory crashes the server on startup due to a `panic!` on non-numeric filenames.

## What changed?

Both `load_consumer_offsets` and `load_consumer_group_offsets` panicked on stray files in the consumer offset directories. Several panic and abort paths existed: `into_string().unwrap()` on non-UTF-8 filenames, `dir_entry.unwrap()` on per-entry IO errors, `metadata.is_err() -> break` silently dropping remaining valid offsets, the non-numeric filename panic, and `read_exact`/`File::open` via `?` aborting the entire load on a single truncated or unreadable file.

All panic and abort paths in both functions are now replaced with `warn!` + `continue`, so one bad file never takes down the rest of the load. Note: a genuinely corrupt or renamed offset file that should be numeric is now silently skipped (with a warning), meaning that consumer loses its committed offset and will re-consume on restart. The `warn!` gives observability so operators can detect this.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Opus 4.6
2. Minimal AI used
3. All tests verified locally with `cargo test`, `cargo clippy`, `cargo fmt --check`, and CI scripts

### DIFF
--- a/core/integration/tests/storage/consumer_offsets.rs
+++ b/core/integration/tests/storage/consumer_offsets.rs
@@ -1,0 +1,180 @@
+/* Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use iggy_common::{ConsumerKind, IggyError};
+use server::streaming::partitions::storage::{load_consumer_group_offsets, load_consumer_offsets};
+use std::path::Path;
+use std::sync::atomic::Ordering;
+
+fn write_offset_file(dir: &Path, name: &str, offset: u64) {
+    std::fs::write(dir.join(name), offset.to_le_bytes()).unwrap();
+}
+
+#[test]
+fn load_consumer_offsets_valid_files() {
+    let dir = tempfile::tempdir().unwrap();
+    write_offset_file(dir.path(), "1", 100);
+    write_offset_file(dir.path(), "2", 200);
+    write_offset_file(dir.path(), "3", 300);
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert_eq!(offsets.len(), 3);
+    assert_eq!(offsets[0].consumer_id, 1);
+    assert_eq!(offsets[0].offset.load(Ordering::Relaxed), 100);
+    assert_eq!(offsets[0].kind, ConsumerKind::Consumer);
+    assert_eq!(offsets[1].consumer_id, 2);
+    assert_eq!(offsets[1].offset.load(Ordering::Relaxed), 200);
+    assert_eq!(offsets[2].consumer_id, 3);
+    assert_eq!(offsets[2].offset.load(Ordering::Relaxed), 300);
+}
+
+#[test]
+fn load_consumer_offsets_skips_non_numeric_files() {
+    let dir = tempfile::tempdir().unwrap();
+    write_offset_file(dir.path(), ".DS_Store", 0);
+    write_offset_file(dir.path(), "backup.bak", 0);
+    write_offset_file(dir.path(), "1", 42);
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert_eq!(offsets.len(), 1);
+    assert_eq!(offsets[0].consumer_id, 1);
+    assert_eq!(offsets[0].offset.load(Ordering::Relaxed), 42);
+}
+
+#[test]
+fn load_consumer_offsets_skips_truncated_files() {
+    let dir = tempfile::tempdir().unwrap();
+    std::fs::write(dir.path().join("1"), [0u8; 3]).unwrap();
+    std::fs::write(dir.path().join("2"), []).unwrap();
+    write_offset_file(dir.path(), "3", 500);
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert_eq!(offsets.len(), 1);
+    assert_eq!(offsets[0].consumer_id, 3);
+    assert_eq!(offsets[0].offset.load(Ordering::Relaxed), 500);
+}
+
+#[test]
+fn load_consumer_offsets_empty_dir() {
+    let dir = tempfile::tempdir().unwrap();
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert!(offsets.is_empty());
+}
+
+#[test]
+fn load_consumer_offsets_skips_directories() {
+    let dir = tempfile::tempdir().unwrap();
+    std::fs::create_dir(dir.path().join("123")).unwrap();
+    write_offset_file(dir.path(), "1", 77);
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert_eq!(offsets.len(), 1);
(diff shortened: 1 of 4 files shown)


hubcio:
can you make it a integration test instead of unit? code in core/server will be replaced when we introduce clustering, thus test will be lost.

atharvalade:
> can you make it a integration test instead of unit? code in core/server will be replaced when we introduce clustering, thus test will be lost.

Good call, moved the tests from `core/server` unit tests to `core/integration/tests/storage/consumer_offsets.rs`. They follow the same pattern as `tests/state/` (direct `server::` imports, no full harness needed). This way they survive the `server` ti `server-ng` migration and can be re-run against the new binary.

core/server/src/streaming/partitions/storage.rs
```diff
@@ -181,9 +181,16 @@ pub fn load_consumer_offsets(path: &str) -> Result<Vec<ConsumerOffset>, IggyErro
         }
 
         let name = dir_entry.file_name().into_string().unwrap();
```
hubcio: this fix is incomplete - `into_string().unwrap()` on this line still panics on a non-utf-8 filename, which is perfectly legal on linux filesystems. the pr's whole rationale is an operator dropping a stray file crashing startup, but a stray file with a non-utf-8 name crashes here before ever reaching the parse you patched. handle the `Err` the same way - `to_string_lossy()` or warn + `continue`.

same loop has two more startup-panic paths the pr leaves untouched: `dir_entry.unwrap()` a few lines up panics on a per-entry io error, and `metadata.is_err()` does `break` instead of `continue` - on a metadata error it bails the whole loop and silently drops every remaining offset file, including valid ones.

core/server/src/streaming/partitions/storage.rs
```diff
@@ -181,9 +181,16 @@ pub fn load_consumer_offsets(path: &str) -> Result<Vec<ConsumerOffset>, IggyErro
...
-        let consumer_id = name.parse::<u32>().unwrap_or_else(|_| {
-            panic!("Invalid consumer ID file with name: '{}'.", name);
-        });
+        let consumer_id = match name.parse::<u32>() {
```
hubcio, line 199: behavior tradeoff worth calling out in the pr description: a genuinely corrupt or renamed offset file that should be numeric but isn't is now silently skipped, so that consumer loses its committed offset and re-consumes on restart. the `warn!` gives observability so this is probably acceptable, but it's a crash -> silent-data-skip swap, not a pure robustness win.

core/server/src/streaming/partitions/storage.rs
```diff
@@ -244,12 +251,16 @@ pub fn load_consumer_group_offsets(
 
         let name = dir_entry.file_name().into_string().unwrap();
```
hubcio: same non-utf-8 gap as in `load_consumer_offsets` - `into_string().unwrap()` here still panics on a non-utf-8 filename before the parse you patched is reached. and same as the other loop, `dir_entry.unwrap()` and the `metadata.is_err()` -> `break` (drops all remaining files on a metadata error) are left as startup-panic / data-loss paths.

core/integration/tests/storage/consumer_offsets.rs
```diff
@@ -0,0 +1,145 @@
...
+    write_offset_file(dir.path(), ".DS_Store", 0);
+    write_offset_file(dir.path(), "backup.bak", 0);
+    write_offset_file(dir.path(), "1", 42);
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert_eq!(offsets.len(), 1);
+    assert_eq!(offsets[0].consumer_id, 1);
+    assert_eq!(offsets[0].offset.load(Ordering::Relaxed), 42);
+}
```
hubcio, line 59: missing test case: a numeric-named file with fewer than 8 bytes of content. `load_consumer_offsets` does `read_exact(&mut [0u8; 8])` on each file, so a truncated or partially-written offset file makes the whole function return `Err` and aborts the entire load - same "one bad file kills startup" class this pr targets, and more likely in practice than a non-numeric name (partial write, disk full). worth adding alongside the non-numeric test here.

core/integration/tests/storage/consumer_offsets.rs
```diff
@@ -0,0 +1,145 @@
...
+fn load_consumer_offsets_empty_dir() {
+    let dir = tempfile::tempdir().unwrap();
+
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert!(offsets.is_empty());
+}
+
+#[test]
+fn load_consumer_offsets_skips_directories() {
```
hubcio, line 85: this test doesn't cover anything this pr changes. the directory skip it asserts happens via the `metadata.unwrap().is_dir()` check in `load_consumer_offsets`, which is untouched by this pr - and `123` is a numeric name, so it'd only reach the patched code if that `is_dir` check were removed. fine to keep as a regression test, but it's pinning pre-existing behavior, not this change.

core/integration/tests/storage/consumer_offsets.rs
```diff
@@ -0,0 +1,145 @@
...
+    let offsets = load_consumer_offsets(dir.path().to_str().unwrap()).unwrap();
+
+    assert_eq!(offsets.len(), 1);
+    assert_eq!(offsets[0].consumer_id, 1);
+    assert_eq!(offsets[0].offset.load(Ordering::Relaxed), 77);
+}
+
+#[test]
+fn load_consumer_offsets_nonexistent_dir() {
+    let result = load_consumer_offsets("/tmp/nonexistent_iggy_test_dir_12345");
```
hubcio: also at line 138 in `load_consumer_group_offsets_nonexistent_dir`. both tests hardcode `/tmp/nonexistent_iggy_test_dir_12345` - if that path ever exists (leftover from a crashed run, or a parallel run) the test breaks. create a `tempfile::tempdir()`, grab its path, drop it, then use that path - guaranteed nonexistent and unique.

hubcio (changes requested):
core idea is right - replacing the `panic!` on non-numeric filenames with warn + skip is the correct direction. but as posted this is a partial fix: `into_string().unwrap()` and `dir_entry.unwrap()` in both functions keep the exact startup-panic-on-stray-file behavior #3132 targets, and the regression tests only feed inputs the patch already handles (`.DS_Store`, `backup.bak` - all valid utf-8), so they confirm the patch runs but don't prove the bug class is gone. inline comments below.

/author

core/server/src/streaming/partitions/storage.rs
atharvalade: Fixed, switched to `to_string_lossy()`

core/server/src/streaming/partitions/storage.rs
atharvalade: Good point, added note to PR description

core/server/src/streaming/partitions/storage.rs
atharvalade: Fixed all three in both loops

core/integration/tests/storage/consumer_offsets.rs
atharvalade: Added, truncated files now skip cleanly

core/integration/tests/storage/consumer_offsets.rs
atharvalade: Kept it as a regression guard

core/integration/tests/storage/consumer_offsets.rs
atharvalade: Fixed, using `tempdir` drop pattern now

hubcio (approved):

mmodzelewski (approved):

numinnex (approved):

PR #3138, opened 2026-04-17
feat(consensus): impl consensus group mechanism

numinnex:
Add `consensus_group` field, required to distinguish between different instances of consensus that are stored in `metadata` and `partition`.


core/binary_protocol/src/primitives/partition_assignment.rs
```diff
@@ -0,0 +1,75 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use crate::WireError;
+use crate::codec::{WireDecode, WireEncode, read_u32_le, read_u64_le};
+use bytes::{BufMut, BytesMut};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CreatedPartitionAssignment {
```
hubcio, line 23: IMHO it would be better if this struct was placed in core/binary_protocol/src/primitives - it's not a request per-se.

numinnex: Yeah, I will move it to `primitives`

spetz (approved):

hubcio (approved):

PR #3159, opened 2026-04-23
feat(java): Implement CLI and resource provisioner for pinned producer benchmark

slbotbm:

Closes #

## Rationale

This PR introduces the CLI for Pinned Producer benchmark, and the ResourceProvisioner class to create stream/topics on the server.

## What changed?
```bash
org/apache/iggy/bench
  ├── IggyBench.java
  │   # Main entry point. 
  ├── benchmarks
  │   └── tcp
  │       └── async
  │           └── PinnedProducer.java
  │               # Benchmark runner shell for the pinned producer.
  │               # Holds global/per-benchmark args, delegates stream/topic setup to `ResourceProvisioner`,
  │               # and exposes `provisionResources()`. `run()` is currently empty; will have actual pinned producer actor.
  ├── cli
  │   ├── IggyBenchCommand.java
  │   │   # Root Picocli command for `iggy-bench`.
  │   │   # Defines shared CLI options like message sizing, batching, auth, warmup, sampling, and registers `PinnedProducerCommand` as a subcommand.
  │   └── PinnedProducerCommand.java
  │       # Picocli subcommand for the pinned producer benchmark.
  │       # Translates CLI flags into `GlobalCliArgs` and `PinnedProducerCliArgs`, validates them, provisions resources, then invokes the benchmark.
  ├── exception
  │   └── BenchmarkException.java
  ├── models
  │   ├── cli
  │   │   ├── GlobalCliArgs.java
  │   │   │   # Record holding shared benchmark configuration from the root CLI command.
  │   │   └── PinnedProducerCliArgs.java
  │   │       # Record holding pinned-producer-specific options.
  │   └── provision
  │       └── ProvisionedResources.java
  │           # Record describing the resources created or reused for a run.
  └── provision
      └── ResourceProvisioner.java
          # Creates or reuses benchmark streams/topics in Iggy before a run.
          # Logs in with blocking TCP client credentials, and creates/re-uses streams/topics.
```

## Local Execution

- Passed: `./gradlew :iggy-bench:run --args="pp"`
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
3. How did you verify the generated code works correctly? Ran it, read through and understood it myself.


foreign/java/bench/src/main/java/org/apache/iggy/bench/provision/ResourceProvisioner.java
```diff
@@ -0,0 +1,98 @@
...
+                                1L,
+                                CompressionAlgorithm.None,
+                                BigInteger.valueOf(pinnedProducerCliArgs.messageExpiry()),
+                                BigInteger.valueOf(pinnedProducerCliArgs.maxTopicSize()),
+                                Optional.empty(),
+                                topicNames.get(0));
+            }
+        }
+
+        client.close();
```
atharvalade: `IggyTcpClient` implements `Closeable`. The client is created at L45 but `client.close()` only runs at L95 on the happy path. If any operation between L49-93 throws (e.g., server rejects `createStream`, network error), the TCP connection and its underlying NIO/connection-pool resources leak. For a benchmark tool that may be run repeatedly, this compounds.

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
```diff
@@ -0,0 +1,152 @@
...
+            names = {"--message-batches", "-b"},
+            defaultValue = "1000",
+            description = "Number of message batches.")
+    private int messageBatches = 1000;
+
+    @Option(
+            names = {"--total-data", "-T"},
+            defaultValue = "0",
+            description = "Total data volume in bytes.")
+    private long totalData;
```
atharvalade: Parent command defines `-p` for `--password` (`IggyBenchCommand` L92). Subcommand defines `-p `for `--producers` (PinnedProducerCommand L59). In picocli, subcommand scope shadows parent scope. If a user writes `iggy-bench pp` `-p` `secretpass` intending to set password, picocli will attempt to parse "secretpass" as an integer for `--producers` and throw a confusing type-conversion error. The user would need to know to write `iggy-bench -p secretpass pp` instead.

and also the same for `IggyBenchCommand.java` L92

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Option(
+            names = {"--rate-limit", "-r"},
+            defaultValue = "0",
+            description = "(NOT USED CURRENTLY) Optional total rate limit in bytes per second.")
+    private long rateLimit;
```
atharvalade, line 56: Both the parent `(--total-data)` and this subcommand `(--max-topic-size)` bind `-T`. Since both are long, iggy-bench pp `-T` 5000 quietly sets topic size instead of total data with zero indication anything went wrong. I think it would be better to pick a different short flag for one of them.

atharvalade (changes requested):

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
slbotbm: > Parent command defines -p for --password (IggyBenchCommand L92). Subcommand defines -p for --producers (PinnedProducerCommand L59). In picocli, subcommand scope shadows parent scope. If a user writes iggy-bench pp -p secretpass intending to set password, picocli will attempt to parse "secretpass" as an integer for --producers and throw a confusing type-conversion error. The user would need to know to write iggy-bench -p secretpass pp instead.

This is the intended behaviour. The rust `iggy-bench` also behaves this way.

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
slbotbm: hmm, I've basically replicated the `rust-bench` args in java. If the option needs to be changed, it would be better to change it in the rust bench as well. 

What do you think?

foreign/java/bench/src/main/java/org/apache/iggy/bench/provision/ResourceProvisioner.java
slbotbm: Wrapped the code inside a try-catch block to catch the errors.

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
mmodzelewski: Agreed. Let's keep them consistent, ideally we should be able to copy/paste the args between rust/java/any other benchmark runner that we'll create.

slbotbm:
The latest commits enforce the mutual exclusitivity of --message-batches and --total-data through picocli. This was previously done using hand-rolled code.

foreign/java/bench/src/main/java/org/apache/iggy/bench/provision/ResourceProvisioner.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(ResourceProvisioner.class);
+
+    public ProvisionedResources provisionResources(
+            GlobalCliArgs globalCliArgs, PinnedProducerCliArgs pinnedProducerCliArgs) {
+        var streamNames = new ArrayList<String>();
+        var topicNames = List.of("topic-1");
+
+        try {
+            try (var client = IggyTcpClient.builder()
```
mmodzelewski: both `try` blocks cover the same scope, so we can merge them into one

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
```diff
@@ -0,0 +1,162 @@
...
+        name = "iggy-bench",
+        mixinStandardHelpOptions = true,
+        subcommands = {PinnedProducerCommand.class},
+        description = "Benchmark CLI for the Apache Iggy Java SDK.")
+public final class IggyBenchCommand implements Callable<Integer> {
+    @Option(
+            names = {"--message-size", "-m"},
+            defaultValue = "1000",
+            description = "Message size in bytes.")
+    private int messageSize = 1000;
```
mmodzelewski: Some options are initialized, some are not. Let's keep it consistent. I'd drop the values, as we already have `defaultValue` in the annotation - I suppose it will initialize the variable, right?

foreign/java/bench/src/main/java/org/apache/iggy/bench/IggyBench.java
```diff
@@ -0,0 +1,35 @@
...
+import picocli.CommandLine;
+
+public final class IggyBench {
+
+    private IggyBench() {}
+
+    public static void main(String[] args) {
+        var exitCode = new CommandLine(new IggyBenchCommand()).execute(args);
+        if (exitCode != 0) {
+            System.exit(exitCode);
```
mmodzelewski: I think we can run System.exit for all the codes, no need to exclude `0`

mmodzelewski (changes requested):
@slbotbm Thanks for the PR. It looks good overall, just a few minor comments.

foreign/java/bench/src/main/java/org/apache/iggy/bench/provision/ResourceProvisioner.java
slbotbm: done

foreign/java/bench/src/main/java/org/apache/iggy/bench/IggyBench.java
slbotbm: done

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/IggyBenchCommand.java
slbotbm: yeah it initializes the variable with the default value as long as we use picocli. Applied the change.

mmodzelewski (approved):

spetz (approved):

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/PinnedProducerCommand.java
```diff
@@ -0,0 +1,111 @@
...
+            log.info("Starting the Pinned Producer benchmark...");
+            var benchmark = new TcpAsyncPinnedProducer(globalCliArgs, pinnedProducerCliArgs);
+            benchmark.provisionResources();
+            benchmark.run();
+
+            return ExitCode.OK;
+        } catch (RuntimeException exception) {
+            var message = exception.getMessage() != null ? exception.getMessage() : exception.toString();
+            spec.commandLine().getErr().println(message);
+            return ExitCode.SOFTWARE;
```
atharvalade, line 108: The catch block only prints `getMessage()` from the wrapper `BenchmarkException`, so the actual root cause (connection refused, auth failure, etc.) is completely lost. I think walking the cause chain or at least appending `exception.getCause().getMessage()` would save a lot of debugging headaches here.

Thoughts? @mmodzelewski @spetz

atharvalade (changes requested):

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/PinnedProducerCommand.java
hubcio: IMHO not really that important for benchmark. its not prod code. worth fixing in follow ups.

atharvalade (approved):
everything else lgtm

foreign/java/bench/src/main/java/org/apache/iggy/bench/cli/PinnedProducerCommand.java
mmodzelewski: we can also consider having different verbosity options, like many tools have, but let's take that as an improvement for next commits
cc @slbotbm

PR #3163, opened 2026-04-24
feat(cluster): implement replica bootstrap

numinnex:
Implement `Replica` bootstrap logic. Currently it does not handle cases where replicas are out of sync (needs `State Transfer` to be implemented).


core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+        &mut partition,
+        config,
+        stream_id,
+        topic_id,
+        partition_id,
+        loaded_log,
+    )
+    .await?;
+
+    let current_offset = partition
```
hubcio, line 470: `current_offset` here is derived from `max(end_offset where size > 0)`. `end_offset` comes from the segment metadata index. if the previous run wrote messages but index entries weren't fsynced before a crash, `end_offset` lags actual `messages.log` bytes. next append assigns logical offsets that overlap pre-crash messages already on disk - duplicate-offset corruption visible to all SDK clients. derive `current_offset` from messages-side size + tail-replay, or fail-fast on index/messages length divergence.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+        config
+            .system
+            .get_consumer_group_offsets_path(stream_id, topic_id, partition_id);
+
+    // TODO: decouple consumer offset loading from the `server` crate.
+    let loaded_consumer_offsets = load_consumer_offsets(&consumer_offsets_path).unwrap_or_default();
+    let consumer_offsets = ConsumerOffsets::with_capacity(loaded_consumer_offsets.len());
+    {
+        let guard = consumer_offsets.pin();
+        for offset in loaded_consumer_offsets {
```
hubcio, line 663: also lines 647-650. silent down-clamp of persisted consumer/group offsets to `current_offset`. if recovered `current_offset < persisted consumer offset`, the partition log was truncated/lost relative to consumer state, but the clamp causes the consumer to re-deliver an already-acked range without warning. that breaks at-least-once. fix: detect divergence and either surface an error or require explicit operator opt-in via config.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+    partition_metadata: &Partition,
+    self_replica_id: u8,
+    replica_count: u8,
+    bus: Rc<IggyMessageBus>,
+) -> Result<IggyPartition<Rc<IggyMessageBus>>, ServerNgError> {
+    let stream_id = namespace.stream_id();
+    let topic_id = namespace.topic_id();
+    let partition_id = namespace.partition_id();
+    let stats = Arc::new(PartitionStats::new(topic_stats));
+    let consensus = VsrConsensus::new(
```
hubcio, line 425: per-partition `VsrConsensus::new()` + `init()` boots fresh-zero (commit_min/commit_max/sequencer/view all 0). no `restore_commit_state`, no view restore. today this is vacuous because `MemoryMessageJournal` is volatile, but as soon as a durable partition journal lands, ack-but-not-yet-applied ops are silently lost on restart. either stub a hard-error if the partition journal contains a tail past snapshot, or implement matching restore now so the call shape exists.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+
+    if let Some(active_index) = partition.log.segments().len().checked_sub(1) {
+        let storage = &partition.log.storages()[active_index];
+        if let (Some(messages_reader), Some(index_reader)) = (
+            storage.messages_reader.as_ref(),
+            storage.index_reader.as_ref(),
+        ) {
+            let index_path = index_reader.path();
+            let index_size = std::fs::metadata(&index_path).map_or(0, |metadata| metadata.len());
+            partition.log.messages_writers_mut()[active_index] = Some(Rc::new(
```
hubcio, line 538: `hydrate_partition_log` re-opens active segment files via `OpenOptions::create(true).write(true)` with `file_exists=true` hardcoded at lines 527-541. if the file vanished between `load_segments` and writer init, `metadata().len()` returns 0 and the size atomic resets to 0, but in-memory `segment.end_offset` still holds the stale value - next append writes at file pos 0 with logical offset N+1. file-pos / logical-offset mismatch, segment metadata diverges from file content. open with `create=false` when `file_exists=true` and fail loudly if missing.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+pub async fn bootstrap(
+    config: &ServerNgConfig,
+    current_replica_id: Option<u8>,
+) -> Result<Rc<ServerNgShard>, ServerNgError> {
+    let topology = resolve_tcp_topology(config, current_replica_id)?;
+    let bus = Rc::new(IggyMessageBus::with_config(SHARD_ID, config));
+    let recovered = recover::<ServerNgMuxStateMachine>(Path::new(&config.system.path))
+        .await
+        .map_err(ServerNgError::MetadataRecovery)?;
+    let restored_op = recovered.last_applied_op.unwrap_or_else(|| {
```
hubcio, line 223: `restored_op = snapshot.sequence_number()` when `last_applied_op = None` is correct only because `IggySnapshot::persist` (`core/metadata/src/impls/metadata.rs:91-130`) is atomic (tmp + fsync + rename + dir fsync). that's an implicit contract. add a doc-comment naming the invariant so future snapshot persistence changes don't quietly break this fallback path.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+        bus,
+        LocalPipeline::new(),
+    );
+
+    let last_header = journal
+        .last_op()
+        .and_then(|op| usize::try_from(op).ok())
+        .and_then(|op| journal.header(op).map(|header| *header));
+    if let Some(header) = last_header {
+        consensus.set_view(header.view);
```
hubcio, line 270: `restore_metadata_consensus` calls `consensus.set_view(header.view)` BEFORE `consensus.init()`, but `set_log_view` and `set_last_prepare_checksum` (lines 272-273) happen AFTER `init()`. `init()`'s `is_primary` check uses the already-restored view. on a multi-replica boot at non-zero recovered view, primary-side timers can arm in the brief window before `log_view` and the prepare checksum are set. order all view / log_view / checksum sets before `init()`, or post-restore enter `ViewChange` instead of `Normal` so DVC quorum drives state.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+                topic_stats.push(topic.stats.clone());
+                for partition in &topic.partitions {
+                    namespaces.push((stream.id, topic_id, topic.stats.clone(), partition.clone()));
+                }
+            }
+        }
+        (topic_stats, namespaces)
+    });
+
+    for topic_stats in topic_stats {
```
hubcio, line 335: `zero_out_all` on each topic stat decrements parent stream stats by the topic value (`core/common/src/types/streaming_stats.rs:196-209` does `parent.fetch_sub(prev)`). that assumes `snapshot.stream.size == sum(snapshot.topic.size)`. if the snapshot was filled non-atomically, `fetch_sub` on `AtomicU64` wraps on underflow and the stream counter ends up near `u64::MAX`. cleaner: zero stream stats explicitly via `stream_stats.zero_out_all()` and skip the per-topic decrement.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+        resolve_optional_listener_addr(config.quic.enabled, "quic.address", &config.quic.address)?;
+    if !config.cluster.enabled {
+        return Ok(TcpTopology {
+            // Keep parity with the current server binary and the integration
+            // harness: `--replica-id` may be passed unconditionally, but in
+            // single-node mode there is only replica 0.
+            self_replica_id: SHARD_REPLICA_ID,
+            replica_count: 1,
+            client_listen_addr: default_client_addr,
+            replica_listen_addr: Some(SocketAddr::new(default_client_addr.ip(), 0)),
```
hubcio, line 877: single-node mode (`!config.cluster.enabled`) sets `replica_listen_addr: Some(SocketAddr::new(default_client_addr.ip(), 0))` with `replica_count = 1`. `start_via_replica_io` then unconditionally binds the replica listener (`core/message_bus/src/replica/io.rs:209`). wasted fd, wasted ephemeral port, no consumers. skip the replica listener when `replica_count == 1`.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+        tcp,
+        topology.ws_listen_addr.map(|_| ws),
+        topology.quic_listen_addr.map(|_| quic),
+        topology.tcp_tls_listen_addr.map(|_| tcp_tls),
+        None,
+        shard.bus.config().reconnect_period,
+    )
+    .await
+    .map_err(ServerNgError::StartListeners)?;
+    let Some(bound) = bound else {
```
hubcio, line 1082: `let Some(bound) = bound else { return Ok(()); };` is unreachable. `replica_io::start_on_shard_zero` returns `None` only when `bus.shard_id() != 0` (`core/message_bus/src/replica/io.rs:191-194`), but `SHARD_ID` is hardcoded to 0 at `core/server-ng/src/bootstrap.rs:78` and `IggyMessageBus::with_config(SHARD_ID, ...)` is called at line 214. dead branch hides future SHARD_ID parameterization regressions.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+    /// # Errors
+    ///
+    /// Returns an error if TCP listener bootstrap fails or cluster TCP
+    /// addresses cannot be resolved from config.
+    async fn run(
+        &self,
+        config: &ServerNgConfig,
+        current_replica_id: Option<u8>,
+    ) -> Result<(), ServerNgError> {
+        let topology = resolve_tcp_topology(config, current_replica_id)?;
```
hubcio, line 146: `resolve_tcp_topology` is called here in `RunServerNg::run`, then again at line 213 in `bootstrap()`. `main.rs:69-70` invokes `bootstrap()` then `run()`, with `write_current_config` landing in between. no drift today because `config` isn't reloaded, but any future hot-reload or config mutation between the two calls is a feedback-loop hazard. resolve once, pass topology down.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+        };
+        let request = request.transmute_header(|header, new_header: &mut RequestHeader| {
+            *new_header = header;
+            new_header.client = client_id;
+        });
+        shard.dispatch(request.into_generic());
+    })
+}
+
+fn make_deferred_replica_message_handler(shard_handle: &ServerNgShardHandle) -> MessageHandler {
```
hubcio, line 1237: also `make_deferred_client_request_handler` at line 1103. each call into the deferred handler does `RefCell::borrow()` + `Weak::upgrade()` (Rc clone + drop) per replica/client message on the fd-delegation path. the cycle-break is correct, but the per-message indirection is unnecessary once init is done. swap for `Rc<OnceCell<Rc<ServerNgShard>>>` set once at the end of `build_single_shard`; the closure becomes `cell.get().unwrap()` with no borrow check and no Weak upgrade.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+    }
+
+    if let Some(active_index) = partition.log.segments().len().checked_sub(1) {
+        let storage = &partition.log.storages()[active_index];
+        if let (Some(messages_reader), Some(index_reader)) = (
+            storage.messages_reader.as_ref(),
+            storage.index_reader.as_ref(),
+        ) {
+            let index_path = index_reader.path();
+            let index_size = std::fs::metadata(&index_path).map_or(0, |metadata| metadata.len());
```
hubcio, line 537: `std::fs::metadata(&index_path)` is a blocking syscall called inside `async fn hydrate_partition_log` running on the compio reactor. bootstrap-only so it's a brief reactor stall, but trivial to fix with `compio::fs::metadata`.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+}
+
+async fn build_single_shard(
+    config: &ServerNgConfig,
+    topology: &TcpTopology,
+    metadata: ServerNgMetadata,
+    bus: Rc<IggyMessageBus>,
+) -> Result<Rc<ServerNgShard>, ServerNgError> {
+    let shard_id = ShardId::new(SHARD_ID);
+    let partition_count = metadata.mux_stm.streams().read(|inner| {
```
hubcio, line 293: two separate `streams.read(|inner| ...)` traversals at boot - one to count partitions (286-298), one to collect topic_stats + namespaces (314-326). both walk the same nested structure. cold path, but trivially mergeable into a single closure pass.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+    let default_client_addr = parse_socket_addr("tcp.address", &config.tcp.address)?;
+    let default_ws_addr = resolve_optional_listener_addr(
+        config.websocket.enabled,
+        "websocket.address",
+        &config.websocket.address,
+    )?;
+    let default_quic_addr =
+        resolve_optional_listener_addr(config.quic.enabled, "quic.address", &config.quic.address)?;
+    if !config.cluster.enabled {
+        return Ok(TcpTopology {
```
hubcio, line 870: when `cluster.enabled = false`, `current_replica_id` is silently ignored - `self_replica_id` is hardcoded to `SHARD_REPLICA_ID = 0`. an operator passing `--replica-id 5` accidentally on a single-node deployment gets zero signal. either warn-log when cluster is disabled but the flag was passed, or reject.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1410 @@
...
+
+#[derive(Default)]
+struct BoundClientListeners {
+    tcp: Option<SocketAddr>,
+    tcp_tls: Option<SocketAddr>,
+    ws: Option<SocketAddr>,
+    quic: Option<SocketAddr>,
+}
+
+pub trait RunServerNg {
```
hubcio, line 126: `RunServerNg` trait has exactly one impl (`Rc<ServerNgShard>`) and one caller (`main.rs:26`). single-impl trait used purely for `Rc`-receiver dispatch. a free function `pub async fn run(shard: &Rc<ServerNgShard>, ...)` is cleaner.

core/server-ng/src/server_error.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[error("failed to write current server-ng config at {path}")]
+    CurrentConfigWrite {
+        path: String,
+        #[source]
+        source: std::io::Error,
+    },
+    #[error("failed to initialize server-ng logging")]
+    Logging(#[source] LogError),
+    #[error("failed to recover metadata snapshot and journal")]
+    MetadataRecovery(#[source] RecoveryError),
```
hubcio, line 42: `MetadataRecovery(#[source] RecoveryError)` chain works, but `RecoveryError::StateMachine(IggyError)` (`core/metadata/src/impls/recovery.rs:30-46`) wraps with the prefix `"recovery state machine error: "` on Display, partially burying the underlying `IggyError` context. fine for the CLI, less fine for log scrapers that read top-line Display only.

core/server-ng/src/server_error.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[error("failed to parse {context} socket address '{address}'")]
```
hubcio, line 43: also `transport: &'static str` at lines 79-84. `context` and `transport` are stringly-typed enums in disguise. call sites use literal strings `"tcp.tls"`, `"quic"`, `"tcp.address"`, `"websocket.address"`. typo-prone, no exhaustiveness. replace with small `enum ListenerKind { Tcp, TcpTls, Quic, Ws }` and `enum AddressContext`.

core/server-ng/src/server_error.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[error("cluster node count {count} exceeds supported u8 replica count")]
+    ClusterReplicaCountTooLarge { count: usize },
+    #[error("cluster mode requires --replica-id to identify the current node")]
+    MissingReplicaId,
+    #[error("cluster node for replica {replica_id} is missing tcp_replica port")]
+    ClusterReplicaPortMissing { replica_id: u8 },
+    #[error(
+        "recovered namespace stream {stream_id}, topic {topic_id}, partition {partition_id} exceeds configured limits (max_streams={max_streams}, max_topics={max_topics}, max_partitions={max_partitions})"
+    )]
+    RecoveredNamespaceOutOfBounds {
```
hubcio, line 85: five variants - `PartitionLogLoad` (61-68), `MessagesWriterInit` (87-96), `IndexWriterInit` (97-106), `SegmentIndexesLoad` (107-116), `InitialSegmentStorage` (117-126) - all carry the same `{stream_id, topic_id, partition_id, source}` triplet. factor a `PartitionScope { stream_id, topic_id, partition_id }` field with a single source per variant.

core/configs/src/server_ng_config/server_ng.rs
```diff
@@ -60,6 +63,8 @@ const DEFAULT_CONFIG_PATH: &str = "core/server-ng/config.toml";
 pub struct ServerNgConfig {
     pub consumer_group: ConsumerGroupConfig,
     pub data_maintenance: DataMaintenanceConfig,
+    #[serde(default)]
+    pub extra: ExtraConfig,
```
hubcio, line 67: `[extra]` TOML section name is unprecedented in this codebase. zero matches in legacy `config.toml`, no `pub extra` field anywhere else. the only field inside is `namespace`, so `ExtraConfig` is a single-field wrapper. operator-facing schema decision worth getting right before this lands. rename to `[system.namespace]` (sibling of `system.sharding`, which already lives there) or top-level `[namespace]`; drop the `ExtraConfig` wrapper.

core/configs/src/server_config/server.rs
```diff
@@ -183,9 +183,21 @@ impl ServerConfig {
...
     /// Uses compile-time generated env var mappings for unambiguous resolution.
     pub async fn load() -> Result<ServerConfig, ConfigurationError> {
+        Self::load_with_path(
+            DEFAULT_CONFIG_PATH,
+            include_str!("../../../server/config.toml"),
+        )
+        .await
+    }
+
+    pub async fn load_with_path(
```
hubcio, line 193: also `config_provider_with_default` at line 222. new public API with no external callers. `ServerNgConfig::load` reimplements path/include logic at `core/configs/src/server_ng_config/server_ng.rs:116-131` instead of delegating to these helpers. all 5 grep hits are internal to this file. either route `ServerNgConfig::load` through the new helpers, or revert these methods - right now they pay public-API tax for zero callers.

core/consensus/src/impls.rs
```diff
@@ -662,6 +662,33 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+    ///
+    /// # Panics
+    /// - If `commit_min > commit_max`.
+    /// - If commit progress has already been initialized on this consensus instance.
+    pub fn restore_commit_state(&self, commit_min: u64, commit_max: u64) {
+        assert!(
+            commit_min <= commit_max,
+            "commit_min ({commit_min}) must be <= commit_max ({commit_max})"
+        );
+        assert_eq!(
```
hubcio, line 678: `restore_commit_state` asserts `commit_min == 0 && commit_max == 0` on entry. today fine, but if `VsrConsensus::new` ever changes its initial commit defaults (e.g., for recovery hooks), the asserts break silently in tests but mask corruption in prod. replace with a Builder pattern, or a `restored: Cell<bool>` guard set on first restore so the contract is enforced by construction.

core/metadata/src/impls/recovery.rs
```diff
@@ -104,28 +104,36 @@ pub struct RecoveredMetadata<M> {
...
     let journal_path = metadata_dir.join("journal.wal");
-    let journal = PrepareJournal::open(&journal_path, snapshot.sequence_number()).await?;
+    let journal = PrepareJournal::open(
+        &journal_path,
+        snapshot.as_ref().map_or(0, IggySnapshot::sequence_number),
+    )
+    .await?;
 
     // 4. Replay journal entries after snapshot
     let headers_to_replay = journal.iter_headers_from(replay_from);
```
hubcio, line 142: replay loop `mux_stm.update(entry)?` aborts on first error and `recover()` returns `Err`, bootstrap fails. the journal still contains those entries, so an operator restart hits the same error - node is permanently un-bootable until the journal is manually truncated. there's no skip-and-repair-via-VSR path and no contiguity check on the WAL. compounds with `PrepareJournal` `SLOT_COUNT = 1024` silent eviction. confirm intentional fail-fast for now and document the repair path; longer-term, walk the hash chain and stop at the first break.

core/common/src/sharding/namespace.rs
```diff
@@ -89,4 +130,81 @@ impl IggyNamespace {
...
+        } else {
+            bits_required((max_topics - 1) as u64)
+        };
+        let partition_bits = if max_partitions == 0 {
+            return Err(NamespaceCapacityError::ZeroPartitions);
+        } else {
+            bits_required((max_partitions - 1) as u64)
+        };
+
+        let required_bits = stream_bits
```
hubcio, line 136: `ExceedsU64 { required_bits: u32::MAX }` placeholder is unreachable. `bits_required` returns `u32 ≤ 64` (loop bounded by `u64::BITS` at lines 28-38); `checked_add` of three values each ≤ 64 has max sum 192, cannot overflow `u32`. drop the `checked_add` branch and sum directly - the `required_bits > u64::BITS` check below is the real guard.

hubcio (changes requested):
out of diff findings:

**core/partitions/src/messages_writer.rs:60-63** — `let _ = file.sync_all().await.map_err(...)` swallows IO error on the file_exists open path. EIO at boot indicates dying media, but bootstrap proceeds with a stale view of disk. on master HEAD already, but server-ng makes this path live for every partition. propagate the error.

**core/partitions/src/iggy_index_writer.rs:53** — `let _ = file.sync_all().await;` same pattern, no `map_err` at all. propagate.

**core/partitions/src/messages_writer.rs:137** — `let chunk_vec: Vec<_> = chunk.to_vec();` allocates a Vec per chunk per `save_frozen_batches`. steady-state per-batch alloc on what is now the primary write path. cache a reusable Vec in the writer or refactor the compio iov call to take a borrowed slice.

**core/shard/src/lib.rs:622** — `let namespaces: Vec<_> = planes.1.0.namespaces().copied().collect();` allocates a fresh Vec per inbox-frame iteration. `core/shard/src/router.rs:270` calls `process_loopback` after every dispatched frame in steady state. hoist a `namespaces_buf: Vec<u64>` like the existing `loopback_buf`, or short-circuit when partitions / loopback drained are empty.

**core/shard/src/router.rs:94-103 and 199-208** — `shards_table.shard_for(...).unwrap_or_else(|| { warn!(...); 0 })` silently falls back to shard 0. single-shard server-ng masks this today; multi-shard makes it a silent routing-bug hider. drop the fallback or fail loud.

**core/binary_protocol/src/consensus/message.rs:347** — `transmute_header` does a 256B header copy + 256B memset + `Message::try_from(owned)` re-validate after the closure already wrote a known-valid header. that's a 3rd validation per request (1st transport recv, 2nd `try_into_typed`, 3rd here). reached per-client-message via the new `make_deferred_*_handler` closures at `core/server-ng/src/bootstrap.rs:1086,1116`. add a `transmute_header_unchecked` that skips the trailing TryFrom validation and rely on the closure invariant.

**core/metadata/src/stm/stream.rs:615-676** — `Snapshotable::to_snapshot` reads `round_robin_counter` with `Ordering::Relaxed` during snapshot fill. concurrent producer increments race the snapshot read. single-shard today: not exploitable. document the invariant.

**core/common/src/types/streaming_stats.rs counter ordering** — counter `fetch_add` / `fetch_sub` use `Ordering::AcqRel` even though the counters synchronize no other data. `Relaxed` is correct; `AcqRel` forces unnecessary `ldar` / `stlr` fences on aarch64. per-batch produce path. confirmed via method names like `_inconsistent` and `load_for_snapshot` that no consumer relies on the AcqRel half.

numinnex:
Addressed all of the correctness issues, I've skipped the optimizations nits, as this code will change quite a few times before it's going to be used by main server binary.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1474 @@
...
+    segment: &iggy_common::Segment,
+    storage: &iggy_common::SegmentStorage,
+    indexes: Option<&server::streaming::segments::IggyIndexesMut>,
+) -> Result<(), ServerNgError> {
+    let messages_size_bytes = storage
+        .messages_reader
+        .as_ref()
+        .map_or(0, |reader| u64::from(reader.file_size()));
+    let indexed_size_bytes = indexes.map_or(0, |indexes| u64::from(indexes.messages_size()));
+    if messages_size_bytes == indexed_size_bytes {
```
hubcio: `messages_size_bytes == indexed_size_bytes` is the wrong invariant. `iggy_partition.rs:1228-1244` writes one sparse-index entry per `commit_messages` call (gated by `flush_index.is_none()`), recording start-of-batch position. messages file accumulates all bytes per batch. after any clean shutdown of multi-batch server-ng-written data, `messages_size_bytes > indexed_size_bytes` strictly, so this validator returns `RecoveredSegmentSizeDivergence` on every restart.

note: legacy `core/server/src/shard/system/messages.rs:597,665` writes one index per message, so old-format data still passes, but new server-ng data does not. recovery cannot validate its own output.

fix options: drop the equality check, compute expected size from index + last batch length, or record file size in segment metadata.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1474 @@
...
+        config
+            .system
+            .get_consumer_offsets_path(stream_id, topic_id, partition_id);
+    let consumer_group_offsets_path =
+        config
+            .system
+            .get_consumer_group_offsets_path(stream_id, topic_id, partition_id);
+
+    // TODO: decouple consumer offset loading from the `server` crate.
+    let loaded_consumer_offsets = load_consumer_offsets(&consumer_offsets_path).unwrap_or_default();
```
hubcio: also at line 691.

`load_consumer_offsets` (and the group variant) at `core/server/src/streaming/partitions/storage.rs:165-168,200,208,227-230` returns `IggyError::CannotReadConsumerOffsets` / `CannotReadFile` for any read failure: ENOENT (legitimate first-boot empty), EACCES, EBUSY, EIO. all collapse into `Vec::default()` here. a transient FS failure during boot silently resets consumer offsets, so consumers re-read messages already consumed.

fix: propagate via a new `ServerNgError::ConsumerOffsetsLoad` variant. distinguish "directory missing" (legitimate empty on first boot) from "load failed" if needed.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1474 @@
...
+pub async fn bootstrap(
+    config: &ServerNgConfig,
+    current_replica_id: Option<u8>,
+) -> Result<Rc<ServerNgShard>, ServerNgError> {
+    let topology = resolve_tcp_topology(config, current_replica_id)?;
+    let bus = Rc::new(IggyMessageBus::with_config(SHARD_ID, config));
+    let recovered = recover::<ServerNgMuxStateMachine>(Path::new(&config.system.path))
+        .await
+        .map_err(ServerNgError::MetadataRecovery)?;
+    let restored_op = recovered.last_applied_op.unwrap_or_else(|| {
```
hubcio, line 223: `recovery.rs:139-158` walks the WAL and applies every replayed entry, setting `last_applied_op` to journal tail M. bootstrap then calls `restore_commit_state(M, M)` declaring `commit_min == commit_max == M`. `impls.rs:1110-1130, 1218, 1230` propagate `commit_min` via primary heartbeat and DVC; `view_change_quorum.rs:88-95` `dvc_max_commit` propagates further into peer state via `complete_view_change_as_primary`.

WAL entries past the last quorum-confirmed op are NOT committed; bootstrap declares them so. multi-replica safety violation: a replica with uncommitted-but-replayed ops claims they're committed.

single-node trivially correct (quorum=1). latent landmine the moment cluster ships.

fix: persist a separate commit watermark (superblock) and only restore commit_min up to that watermark. replay-as-applied does not mean replay-as-committed.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1474 @@
...
+        0,
+        bus,
+        LocalPipeline::new(),
+    );
+
+    let last_header = journal
+        .last_op()
+        .and_then(|op| usize::try_from(op).ok())
+        .and_then(|op| journal.header(op).map(|header| *header));
+    if let Some(header) = last_header {
```
hubcio, line 269: `set_log_view(header.view)` uses the view in which the primary appended this entry locally. VSR invariant per the doc-comment at `impls.rs:506-513`: `log_view` is the view in which this replica's log head was quorum-confirmed (won DVC or processed SV from new primary). setting it to a write-time view over-inflates.

`view_change_quorum.rs:76-84` `dvc_select_winner` picks max `(log_view, op)`. a backup with locally-appended-but-uncommitted ops at view V will beat replicas with the genuinely-quorum-confirmed lower log_view, then `complete_view_change_as_primary` at `impls.rs:1546-1601` forces peers to install the contaminated head via SendStartView.

single-node never runs winner selection. multi-replica latent landmine.

fix: persist `log_view_durable` in a superblock; advance only on quorum SV/DVC win.

core/server-ng/src/bootstrap.rs
```diff
@@ -0,0 +1,1474 @@
...
+    let mut consensus = VsrConsensus::new(
+        CLUSTER_ID,
+        self_replica_id,
+        replica_count,
+        0,
+        bus,
+        LocalPipeline::new(),
+    );
+
+    let last_header = journal
```
hubcio, line 265: `last_header` is Some only if `journal.last_op()` AND `journal.header(op)` both Some. after `journal.drain(0..=N)` (used during snapshot+truncate at `prepare_journal.rs:399-405`), slots are cleared but the `last_op` cell is NOT updated by drain. so `journal.last_op() = Some(N)` while `journal.header(N) = None`, `last_header = None` here, `last_prepare_checksum=0`, `log_view=0`. sequencer correctly set to N. next prepare's `parent=0`.

single-node: silent, no peer to detect. on-disk invariant "WAL chains continuously across snapshot boundary" broken, matters for offline tooling and any future joining peer.

fix: persist `last_prepare_checksum` + `log_view` in the snapshot itself, plumb through `restore_metadata_consensus`.

core/common/src/sharding/namespace.rs
```diff
@@ -49,6 +49,47 @@ pub const PARTITION_MASK: u64 = (1u64 << PARTITION_BITS) - 1;
 pub const TOPIC_MASK: u64 = (1u64 << TOPIC_BITS) - 1;
 pub const STREAM_MASK: u64 = (1u64 << STREAM_BITS) - 1;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NamespaceCapacityError {
```
hubcio, line 55: I think we should use derive `Error` from `thiserror` crate.

hubcio (changes requested):
additional findings cite code outside this PR's diff hunks:

`core/journal/src/prepare_journal.rs:143` - open() does not detect WAL exceeding slot capacity

file not in this PR but the recovery rewrite exposes this gap. live append cannot wrap the 1024-slot ring (panic at `prepare_journal.rs:411-443:422-431`; `SnapshotCoordinator::checkpoint_if_needed` at `core/metadata/src/impls/metadata.rs:243-260` with `CHECKPOINT_MARGIN=64` forces snapshot near capacity). live runtime cannot lose ops. but the rebuild loop at lines 143-184 silently overwrites slot `N % SLOT_COUNT` when reading a WAL file containing more entries than the ring can hold (manual recovery, copy from peer, mismatched binary). `iter_headers_from(0)` then returns only the last 1024 slots. `recovery.rs:142` (the recovery caller in this PR) does not detect this. fix: at open, bail when first slot's op > expected `replay_from`, or when slot count exceeds `SLOT_COUNT`. defensive; cheap.

`core/consensus/src/impls.rs:732` - `set_view` should take `&self`

`view: Cell<u32>` (line 504). `set_last_prepare_checksum(&self)` (line 779) and `set_log_view(&self)` (line 788) take `&self` to mutate identical `Cell` fields. `set_view` is the only odd one out at `&mut self`. `Cell` allows mutation through shared refs; `&mut` is not load-bearing. caller `bootstrap.rs:251` has `let mut consensus = VsrConsensus::new(...)` solely because of this method, then later setters work on `&self`. fix: `pub fn set_view(&self, view: u32)`. one-line.

`core/consensus/src/impls.rs:750` - `pipeline_mut` is dead code

`pub const fn pipeline_mut(&mut self) -> &mut RefCell<P>`. `rg pipeline_mut` returns 1 match, the definition itself, with no callers anywhere in `core/`. unused / dead code should be removed. returning `&mut RefCell` is doubly redundant: `RefCell` already provides interior mutability via `borrow_mut`. the author TODO at line 741 already flags the borrow-across-await footgun on the related `pipeline()` accessor; same risk applies here, plus the `&mut` requirement. fix: delete.

core/server-ng/src/bootstrap.rs
numinnex: We don't store the `commit` ranges yet persistently, I've done this with assumption that the cluster for now on bootstrap will be starting either with empty state, or where all replicas in the cluster are in exactly the same state on the shutdown/reboot moment (no state transfer support).

hubcio (approved):

mmodzelewski (approved):

spetz (approved):

PR #3164, opened 2026-04-24
fix(go,csharp): remove duplicate consumer groups in DeserializeClient/MapClient

### LINKED ISSUE #3131: go-sdk: DeserializeClient returns duplicate consumer groups
`DeserializeClient` pre-allocates with `make([]ConsumerGroupInfo, Count)` then uses `append`, resulting in leading zero-value entries followed by the real data. The outer `for position < length` loop can also re-read groups if there are trailing bytes.

**File:** `foreign/go/binary_serialization/binary_response_deserializer.go` lines 539-558
```go
consumerGroups := make([]iggcon.ConsumerGroupInfo, clientInfo.ConsumerGroupsCount)
// ...
for position < length {
    for i := uint32(0); i < clientInfo.ConsumerGroupsCount; i++ {
        // ...
        consumerGroups = append(consumerGroups, consumerGroup)
        position += 12
    }
}
```

Client metadata always contains ConsumerGroupsCount extra zero-valued entries. If trailing bytes exist, groups are duplicated. Same bug exists in C# SDK (BinaryMapper.MapClient).

Gotta use make([], 0, Count) instead of make([], Count) and break after one pass through the inner loop.

atharvalade:

Closes #3131

## Rationale

`DeserializeClient` (Go) and `MapClient` (C#) return duplicate/zero-valued consumer groups due to incorrect slice pre-allocation and a redundant outer loop.

## What changed?

In Go, `make([]ConsumerGroupInfo, Count)` pre-allocated `Count` zero-valued entries, then `append` added real data after them—doubling the slice. The outer `for position < length` loop could also re-read groups if trailing bytes existed.

Fixed both SDKs by using `make([], 0, Count)` (Go) / `new List<>(Count)` (C#) for correct capacity-only allocation, and replaced the outer `while`/`for` loop with a single bounded `for i < Count` pass.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Opus 4.6
2. Minimal AI used
3. Verified via `go build`, `go vet`, `go test`, `dotnet build`, `dotnet test` (100/100 passed), and CI lint scripts

### DIFF
--- a/foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
+++ b/foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
@@ -239,26 +239,22 @@ internal static class BinaryMapper
     internal static ClientResponse MapClient(ReadOnlySpan<byte> payload)
     {
         var (response, position) = MapClientInfo(payload, 0);
-        var consumerGroups = new List<ConsumerGroupInfo>();
-        var length = payload.Length;
+        var consumerGroups = new List<ConsumerGroupInfo>(response.ConsumerGroupsCount);
 
-        while (position < length)
+        for (var i = 0; i < response.ConsumerGroupsCount; i++)
         {
-            for (var i = 0; i < response.ConsumerGroupsCount; i++)
-            {
-                var streamId = BinaryPrimitives.ReadInt32LittleEndian(payload[position..(position + 4)]);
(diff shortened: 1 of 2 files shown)


lukaszzborek (approved):
approve for csharp
for go looks the same 😅, but not sure about how tests looks like. Now there is no tests for that part

numinnex (approved):

ryankert01 (approved):

hubcio (approved):

PR #3165, opened 2026-04-24
fix(go): add bounds checking to DeserializeStreams for payloads > 64KB

### LINKED ISSUE #3130: go-sdk: DeserializeStreams known bug with payloads > 64KB
There is an acknowledged but unfixed deserialization bug for stream lists larger than 2^16 bytes.

**File:** `foreign/go/binary_serialization/binary_response_deserializer.go` lines 81-82
```go
//TODO there's a deserialization bug, investigate this
//it occurs only with payload greater than 2 pow 16
```

Users with many streams (or streams with long names) will get corrupted stream lists. No error is returned.

we need to fix the deserialization logic and add a regression test with a payload > 64KB.

atharvalade:

Closes #3130

## Rationale

`DeserializeStreams` and `DeserializeToStream` had zero bounds checking and no error propagation, causing silent data corruption for stream lists larger than 64KB.

## What changed?

`DeserializeToStream` performed raw slice accesses without validating that the payload contained enough bytes for the 33-byte fixed header or the variable-length name. With large payloads (>64KB), any framing misalignment caused position drift—subsequent streams were deserialized from wrong offsets, silently returning corrupted data with no error.

The fix adds bounds validation before every access in `DeserializeToStream` (returns error on insufficient data), propagates errors through `DeserializeStreams`, and updates the TCP caller. A new test file covers single-stream, multi-stream, truncated header/name, corrupted payload, max-length name, and a 70KB+ regression test that verifies every field of ~1000 streams.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Opus 4.6
2. Minimal AI used
3. All tests ran locally and pass; verified against the Rust wire format

### DIFF
--- a/foreign/go/binary_serialization/binary_response_deserializer_test.go
+++ b/foreign/go/binary_serialization/binary_response_deserializer_test.go
@@ -19,6 +19,7 @@ package binaryserialization
 
 import (
 	"encoding/binary"
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


foreign/go/binary_serialization/binary_response_deserializer_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	if streams[0].Id != 1 || streams[0].Name != "stream-one" {
+		t.Errorf("stream[0] = {Id:%d, Name:%q}, want {Id:1, Name:\"stream-one\"}", streams[0].Id, streams[0].Name)
+	}
+	if streams[1].Id != 2 || streams[1].Name != "s2" {
+		t.Errorf("stream[1] = {Id:%d, Name:%q}, want {Id:2, Name:\"s2\"}", streams[1].Id, streams[1].Name)
+	}
+	if streams[2].Id != 3 || streams[2].Name != "third" {
+		t.Errorf("stream[2] = {Id:%d, Name:%q}, want {Id:3, Name:\"third\"}", streams[2].Id, streams[2].Name)
+	}
```
slbotbm: Here, only the Id and that name are being checked. I think it would be better to check everything.

slbotbm (commented):
one comment, otherwise lgtm

hubcio:
@slbotbm use /author after you are done with review :)

/author

foreign/go/binary_serialization/binary_response_deserializer_test.go
atharvalade: Good catch, now checking all fields

slbotbm (approved):
lgtm

hubcio (approved):

mmodzelewski (approved):

numinnex (approved):

PR #3166, opened 2026-04-24
fix(go): return error instead of panic on S2 decode failure

### LINKED ISSUE #3129: go-sdk: panic on malformed S2-compressed poll response
`DeserializeFetchMessagesResponse` in Go SDK calls `panic()` when S2 decompression fails, instead of returning an error.

```go
payloadSlice, err = s2.Decode(nil, payloadSlice)
if err != nil {
    panic("iggy: failed to decode s2 payload: " + err.Error())
}
```

A truncated or corrupt compressed message from the server crashes the entire Go client process. This is a DoS vector clients should never panic on bad server data.

We should return error instead of calling panic.

atharvalade:

Closes #3129

## Rationale

Library code should never panic on bad server data — a truncated or corrupt S2-compressed message currently crashes the entire Go client process.

## What changed?

`DeserializeFetchMessagesResponse` called `panic()` when `s2.Decode` failed on a malformed compressed payload. Callers had no way to recover.

Replaced `panic(...)` with `return nil, fmt.Errorf(...)` so the error propagates through the existing `error` return value. Added unit tests covering the malformed S2 path, the no-compression path, and the empty-payload path.

## Local Execution

- Passed
- Pre-commit hooks ran (license-headers, trailing-whitespace, trailing-newline all clean on changed files)

## AI Usage

1. Opus 4.6
2. Guided the fix and test scaffolding
3. Ran `go test`, `go vet`, and CI lint scripts locally; all pass


foreign/go/binary_serialization/binary_response_deserializer.go
```diff
@@ -157,7 +157,7 @@ func DeserializeFetchMessagesResponse(payload []byte, compression iggcon.IggyMes
 			}
 			payloadSlice, err = s2.Decode(nil, payloadSlice)
 			if err != nil {
-				panic("iggy: failed to decode s2 payload: " + err.Error())
+				return nil, fmt.Errorf("iggy: failed to decode s2 payload: %w", err)
```
chengxilo: nit
```suggestion
				return nil, fmt.Errorf("failed to decode s2 payload: %w", err)
```

chengxilo (commented):
LGTM

Standing-Man (approved):
Thanks for contribution!

mmodzelewski (approved):

hubcio (approved):

hubcio:
@chengxilo @Standing-Man thank you for doing review!

PR #3178, opened 2026-04-25
fix(connectors): reject duplicate iggy_sink_open to prevent silent data loss

### LINKED ISSUE #3168: `sink_connector!` macro silently discards buffered data on duplicate `iggy_sink_open` with same ID
### Bug description

In `core/connectors/sdk/src/sink.rs`, the `sink_connector!` macro expands `iggy_sink_open` to unconditionally overwrite any existing entry in `INSTANCES`:

```
let mut container = SinkContainer::new(id);
let result = container.open(id, config_ptr, config_len, log_callback, <$type>::new);
INSTANCES.insert(id, container);  // silently overwrites if id already exists
```

If `iggy_sink_open` is called twice with the same `id`, the previous `SinkContainer` is dropped without `close()` being called. For sinks with buffered state (e.g. `DeltaSink`, which buffers records in a `JsonWriter` before flushing to the Delta log), any unflushed records are silently discarded.

The runtime doesn't currently double-open, but the macro is the public SDK contract and this is a latent data loss bug for any sink with real buffered state.

The same pattern exists in `source_connector!` (`core/connectors/sdk/src/source.rs`) and should be fixed there too, though sources currently have no equivalent buffered-state risk.

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

MukundaKatta:
Closes #3168.

The `sink_connector!` macro accepted a second `iggy_sink_open` call without surfacing an error, silently discarding any data buffered against the first session. Reject the duplicate so callers see the failure and can drain explicitly before re-opening. The same guard is applied to `source_connector!` since it has the same overwrite shape.

Test: `core/connectors/sdk/src/sink.rs::tests::guard_duplicate_open_allows_first_call_and_rejects_second` (and the matching one in `source.rs`).


MukundaKatta:
Closing: bad commit, the source files were replaced with path strings instead of the actual content. Apologies for the noise. Will re-open with the correct diff.

PR #3179, opened 2026-04-25
fix(connectors): reject duplicate iggy_sink_open and iggy_source_open

### LINKED ISSUE #3168: `sink_connector!` macro silently discards buffered data on duplicate `iggy_sink_open` with same ID

MukundaKatta:
Closes #3168.

The `sink_connector!` and `source_connector!` macros silently overwrote an existing `INSTANCES` entry when `iggy_sink_open` or `iggy_source_open` was called twice with the same id. Any in-flight buffered data and the prior task were lost without any signal to the caller. Returning -1 on a duplicate id surfaces the mistake so callers drain or close explicitly first.

Verified `cargo test` cannot run locally in this sandbox. CI will validate.


hubcio:
hi, next time before doing work instantly please comment under the issue and wait for green light from maintainer.

MukundaKatta:
Acknowledged, sorry for jumping in directly. I'll comment on issues first and wait for green light next time.

spetz (approved):

numinnex (approved):

PR #3180, opened 2026-04-26
fix(connectors): propagate sink consume FFI failures with batch context

### LINKED ISSUE #2927: bug(connectors): consume() return value discarded in sink runtime

Standing-Man:

Closes #2927

## Rationale

The sink runtime called the FFI consume callback but ignored its returned status code. As a result, sink implementations could return Err, the SDK would convert that failure into a non-zero FFI status, but the runtime would still treat the batch as successfully processed.

This made sink failures hard to observe at the runtime level and could incorrectly increment processed-message metrics. The previous workaround required sinks, such as the HTTP sink, to log errors internally before returning Err, but the runtime still had no structured failure signal for the failed batch.

## What changed?

- Updated `process_messages` to capture the FFI consume return value and return an error when the status is non-zero.
- Added `RuntimeError::SinkConsumeFailed` with batch context:
- Added a regression test proving process_messages returns SinkConsumeFailed when consume fails.
- Updated HTTP sink comments, logs, README, and integration-test documentation to reflect that the runtime now observes non-zero consume status codes instead of ignoring them.

## Local Execution

- Passed / not passed
Passed
Ran

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT)  Codex
   Codex was used to draft the implementation/test/documentation updates. The final changes were reviewed and adjusted by me.
    Based on my understanding of the issue, the implementation has been verified, and all test cases generated by Codex are passing.


Standing-Man:
Hi @hubcio, @spetz, @numinnex and @mmodzelewski, just a gentle ping — this PR is ready on my side. Could you please take a look when convenient?

hubcio:
We will check in upcoming days.

hubcio:
@mlevkov @atharvalade @kriti-sc please review this once you have some free time - you're doing most contributions in connectors.

kriti-sc:
Out of scope for this PR, but Iggy has an error count metric `iggy_connector_errors_total` that is not being incremented for sink failures. Might be worth addressing in this PR as it is already touching the error path.

core/connectors/runtime/src/sink.rs
```diff
@@ -591,6 +591,100 @@ async fn process_messages(
...
+    if status != 0 {
+        return Err(RuntimeError::SinkConsumeFailed {
+            plugin_id,
+            status,
+            stream: topic_metadata.stream.clone(),
+            topic: topic_metadata.topic.clone(),
+            partition_id: messages_metadata.partition_id,
+            current_offset: messages_metadata.current_offset,
+            schema: messages_metadata.schema.to_string(),
+            messages_count: processed_count,
```
kriti-sc: We should aim for consistency here as messages_count & processed_count mean different things in Iggy. The field name should be renamed to `processed_count`.

kriti-sc:
Correct me if I am wrong, but does the sink exit after it throws an error? Exiting the sink if one message batch fails is too aggressive a strategy. This paradigm of data shipping from a streaming system to a target system sees a lot of transient failures.

kriti-sc (changes requested):

core/connectors/runtime/src/sink.rs
Standing-Man: Thanks, I’ve updated the field name to `processed_count` for consistency.

Standing-Man:
> Correct me if I am wrong, but does the sink exit after it throws an error? Exiting the sink if one message batch fails is too aggressive a strategy. This paradigm of data shipping from a streaming system to a target system sees a lot of transient failures.

Thanks for the review — good point. I’ll update the PR so non-zero FFI consume statuses are logged with batch context without terminating the sink task. Failed batches will return a processed_count of 0 and won’t be counted as successfully processed.

/ready

core/connectors/sinks/http_sink/README.md
```diff
@@ -786,7 +786,7 @@ cargo test -p integration --test connectors -- http_sink
 
 ## Delivery Semantics
 
-All retry logic lives inside `consume()`. The connector runtime invokes `consume()` via an FFI callback that returns an `i32` status code. The runtime does not inspect this return value (see `process_messages()` in `runtime/src/sink.rs`), so errors logged by the sink are not propagated to the runtime's retry or alerting mechanisms. Additionally, consumer group offsets are committed before processing ([runtime issue #1](#known-limitations)). This means:
+All retry logic lives inside `consume()`. The connector runtime invokes `consume()` via an FFI callback that returns an `i32` status code. A non-zero return value is treated as a processing error by `process_messages()` in `runtime/src/sink.rs`, stopping the sink task and surfacing the failure to the runtime. Additionally, consumer group offsets are committed before processing ([runtime issue #1](#known-limitations)). This means:
```
hubcio, line 789: this line says non-zero status `stops the sink task and surfaces the failure to the runtime`, but `process_messages()` in `runtime/src/sink.rs` returns `Ok(0)` on non-zero status (line 663) and the consume loop at `sink.rs:312` keeps polling. also self-contradicts line 798 (`failed batch is not retried`) within this same file. pick one story: either propagate `Err` from `process_messages` so `consume_messages` exits (matches this doc), or rewrite to `logs the failure and continues polling; failed batch is dropped`.

core/connectors/sinks/http_sink/src/lib.rs
```diff
@@ -1195,8 +1195,8 @@ impl Sink for HttpSink {
     ///
     /// **Runtime note**: The FFI boundary in `sdk/src/sink.rs` maps `consume()`'s `Result` to
     /// `i32` (0=ok, 1=err), but the runtime's `process_messages()` in `runtime/src/sink.rs`
-    /// discards that return code. All retry logic lives inside this method — returning `Err`
-    /// does not trigger a runtime-level retry.
+    /// treats a non-zero return code as a processing error. All retry logic lives inside this
+    /// method — returning `Err` stops the sink task but does not retry the already-polled batch.
```
hubcio, line 1199: says `returning Err stops the sink task`. false - `runtime/src/sink.rs:663` returns `Ok(0)` on non-zero status and the loop at `sink.rs:312` keeps polling. line 1240 below (`runtime will receive non-zero FFI status`) is technically true, but the runtime does nothing with it beyond logging. match the actual behavior here or change the runtime.

core/connectors/runtime/src/sink.rs
```diff
@@ -649,5 +649,90 @@ async fn process_messages(
         messages.len(),
     );
 
+    if status != 0 {
```
hubcio, line 652: sdk returns `-1` at `sdk/src/sink.rs:141,156,169,180` for abi/decode failures (missing instance, postcard decode fail) and `1` at `sdk/src/sink.rs:233` for sink `consume()` `Err`. this `if status != 0` collapses both into the same log+drop+continue. `-1` from `instance not found` is reachable on plugin lifecycle bugs and means the runtime/sdk abi is broken - retrying is pointless. positive status is data-plane and may continue. either `match status { -1 => Err(...), 1.. => log+continue, 0 => Ok }` or document the `i32` contract on `ConsumeCallback` so the conflation is intentional.

core/connectors/runtime/src/sink.rs
```diff
@@ -649,5 +649,90 @@ async fn process_messages(
...
+            "Sink connector with ID: {plugin_id} failed to consume {processed_count} processed messages \
+             from stream: {}, topic: {}, partition: {}, current offset: {}, schema: {}, status: {}",
+            topic_metadata.stream,
+            topic_metadata.topic,
+            messages_metadata.partition_id,
+            messages_metadata.current_offset,
+            messages_metadata.schema,
+            status,
+        );
+        return Ok(0);
```
hubcio, line 663: missing `metrics.increment_errors(plugin_key, ConnectorType::Sink)` call. `metrics.rs:213` exposes the api; `source.rs:422,436` already calls it for the symmetric source-side failure path. operators currently only see this failure as drift between `messages_consumed` (`sink.rs:339`) and `messages_processed` (`sink.rs:377`). trivial one-liner.

hubcio, line 663: `Ok(0)` here drops the batch silently. `AutoCommit::When(AutoCommitWhen::PollingMessages)` at `sink.rs:467` commits offsets server-side before `consume()` runs (see `consumer.rs:181-185`, `auto_commit_after_polling=true`), so the failed batch is already committed when this return fires. pre-existing at-most-once behavior, but this pr introduces a structured failure point without taking ownership of the loss - no dlq, no retry, no error metric. either propagate `Err` and exit the task (matches the docs at `README.md:789` and `lib.rs:1199`) or add a `messages_failed` counter + backoff and own the at-most-once semantics explicitly.

note: this also affects `mongodb_sink/README.md:135`, which still claims at-least-once. see review body for cross-sink impact.

core/connectors/runtime/src/sink.rs
```diff
@@ -649,5 +649,90 @@ async fn process_messages(
...
+        fn schema(&self) -> Schema {
+            Schema::Raw
+        }
+
+        fn decode(&self, payload: Vec<u8>) -> Result<Payload, Error> {
+            Ok(Payload::Raw(payload))
+        }
+    }
+
+    extern "C" fn failing_consume(
```
hubcio, line 687: `failing_consume` stub takes 6 raw pointer/len args with `_` prefix and never derefs them. acceptable for the narrow status-check path (no deref happens before the return), but brittle if the ffi signature evolves. nit.

core/connectors/runtime/src/sink.rs
```diff
@@ -649,5 +649,90 @@ async fn process_messages(
...
+        _topic_meta_len: usize,
+        _messages_meta_ptr: *const u8,
+        _messages_meta_len: usize,
+        _messages_ptr: *const u8,
+        _messages_len: usize,
+    ) -> i32 {
+        1
+    }
+
+    #[tokio::test]
```
hubcio, line 699: single test covers `status=1` only. no loop-survival test (assert that the next batch reaches the ffi after a failure), no `-1` vs `1` distinction (impossible given the conflation flagged at line 652), no metrics-drift assertion. a 2-batch test where the first fails and the second succeeds would have caught the doc/code contradiction at `README.md:789` and `lib.rs:1199`.

hubcio (changes requested):
PR description claims `Added RuntimeError::SinkConsumeFailed with batch context` and a 'regression test proving process_messages returns SinkConsumeFailed when consume fails'. neither exists - `core/connectors/runtime/src/error.rs` has no `SinkConsumeFailed` variant, and the actual test asserts `processed_count == 0`. either rewrite the description to match the walk-back commit `90e2b51f2` ('log sink consume failures without exiting'), or restore the variant and the err-propagating test the description claims.

`core/connectors/sinks/mongodb_sink/README.md:135` claims **at-least-once** delivery semantics, relying on the idempotent composite `_id` (line 140) for replay safety. with this pr's behavior (`Ok(0)` at `runtime/src/sink.rs:663` after offsets are already committed), failed batches are silently dropped and never re-delivered to the sink - mongodb's idempotency cannot recover what the runtime never re-sends. the doc was already partly stale pre-pr, but this pr sharpens the contradiction without fixing it. http_sink isn't the only sink with delivery-semantics text; please either propagate `Err` (so at-least-once becomes achievable when paired with retries) or update `mongodb_sink/README.md` to say at-most-once.

other 7 sinks (delta, elasticsearch, iceberg, influxdb, postgres, quickwit, stdout) have no delivery-semantics text - they inherit the new runtime behavior implicitly. source-side is a separate ffi path (`HandleCallback` at `runtime/src/source.rs:506`), unrelated to #2927.

and worthy follow up:
the consume loop at `runtime/src/sink.rs:312-393` polls every `poll_interval` (default `5ms` at `sink.rs:453`) with no consecutive-failure threshold. with the new error log per batch, a dead downstream will spam logs at ~200/sec while continuously advancing committed offsets. pre-existing, but worth a follow-up issue.

Standing-Man:
Thanks a lot for the detailed reviews, @hubcio and @kriti-sc! I’m going to release my claim on this issue for now and come back to it after I’ve gained a deeper understanding of the connector internals 😄

hubcio:
thanks for being reasonable. we really appreciate sticking around.

PR #3182, opened 2026-04-27
fix(node): deserialize void response now ignores payload length for all void commands

### LINKED ISSUE #3128: node-sdk: deserializeVoidResponse now ignores payload length for all void commands
## Problem
`deserializeVoidResponse` was changed from:
- `status === 0 && data.length === 0`
to:
- `status === 0`

This applies to all void commands, not just `SendMessages`.

## Why this is critical
This removes a useful protocol safety check.
Malformed or unexpected payloads for void commands are now silently accepted.

## Where
- `foreign/node/src/client/client.utils.ts`

## Expected
Keep strict empty-payload validation for true void commands.
Handle `SendMessages` with a dedicated response deserializer instead of weakening global void validation.

oscarArismendi:

Closes #3128

## Rationale

The issue reported that `deserializeVoidResponse` had been changed from `status === 0 && data.length === 0` to just `status === 0,` removing the protocol safety check for void commands. However, when I inspected the current code, the `data.length === 0 `check was already present. Tracing further up the call chain, I found the real root cause: `handleResponse` was using` r.subarray(8)` to slice the payload, ignoring the length field the server sends. This meant data could include trailing bytes from the next response in the buffer, making the `data.length === 0` check unreliable regardless of whether it was there or not. `SendMessages` was a separate problem, it used `deserializeVoidResponse` even though the server returns a non-empty payload for that command. Instead of weakening the global void check to accommodate it, it now has its own `deserializeStatusResponse` that only checks `status === 0`

## What changed?

`handleResponse` in `client.utils.ts `was using `r.subarray(8)` to slice the payload, ignoring the length field the server sends. This meant data could include trailing bytes from the next response in the buffer.

`SendMessages` was using `deserializeVoidResponse` which checks `data.length === 0` even though the server returns a non-empty payload for that command. Instead of weakening the global void check, it now has its own `deserializeStatusResponse` that only checks `status === 0`.

## Local Execution

Unit (npm run test:unit) — all pass. Two new test suites were added to client.utils.test.ts:

`handleResponse`: proves that with` length=0` and trailing bytes, data is now empty
`deserializeStatusResponse`: proves it returns true with non-empty data (key difference from `deserializeVoidResponse`)
A deserialize block was also added to send-messages.command.test.ts proving SendMessages accepts non-empty server payloads.

<details><summary>Unit test screenshot</summary>
<p>
Screenshot 2026-04-26 at 7 47 26 PM
</p>
</details> 

E2E (npm run test:e2e) — 57/58 pass. The one failure (`getClusterMetadata`) is pre-existing and unrelated, it expects a 2-node cluster which requires` IGGY_CLUSTER_ENABLED=true` and two running server instances, matching how CI runs it. The TLS test is skipped locally for the same infrastructure reason.

<details><summary>E2E test screenshot</summary>
<p>
Screenshot 2026-04-26 at 9 01 36 PM
</p>
</details> 

BDD (`npm run test:bdd`) the main scenario (`Create stream and send messages`) initially failed because stream IDs were hardcoded as `0` in the step definitions instead of using the dynamically assigned IDs from the server. This was fixed by using `this.stream.id `and` this.topic.id` in the step implementations, and making the Given I have no streams in the system step delete any existing streams before asserting a clean state. The scenario now passes end-to-end, including `SendMessages` against a real server. Four cluster/leader redirection scenarios remain undefined, their step definitions don't exist yet in any SDK.

<details><summary>BDD test screenshot</summary>
<p>
Screenshot 2026-04-26 at 9 27 54 PM
</p>
</details> 

## AI Usage

1. Which tools? **Claude (claude.ai/code)**
2. Scope of usage? **Used for codebase exploration, understanding the binary protocol and TCP buffer framing, writing unit tests**
3. How did you verify the generated code works correctly? **Tests were run locally against a real iggy server built from source. Unit tests were written first to prove the bug, then the fix was applied to make them pass. E2E and BDD tests were run against a locally compiled server**
4. Can you explain every line of the code if asked? **Yes**

## Notes
During investigation I initially assumed SendMessages required a dedicated deserializer because the server might return a non-empty response payload, the [CreateMessage type has a payload field](https://github.com/oscarArismendi/iggy/blob/fix%2FdeserializeVoidResponse-now-ignores-payload-length-for-all-void-commands/foreign/node/src/wire/message/message.utils.ts#L40-L47), which led to some confusion and the issue description appear to be pointing in that direction as well. However, that payload is the outgoing message content sent in the request, not the server response.

Both the Rust server source ([send_empty_ok_response() in send_messages_handler.rs](https://github.com/oscarArismendi/iggy/blob/fix%2FdeserializeVoidResponse-now-ignores-payload-length-for-all-void-commands/core/server/src/binary/handlers/messages/send_messages_handler.rs#L202-L203)) and live debug logs (`ONDATA object true 8 false` exactly 8 bytes = 4 status + 4 length, zero payload) confirm the response is always empty. With handleResponse correctly bounding data by length, `SendMessages` would work fine with `deserializeVoidResponse`.

`deserializeStatusResponse` was kept as a defensive measure in case server behaviour changes in the future, but it is not strictly necessary given the current implementation, just let me know if I should removed it.


hubcio:
@T1B0 mind taking a look?

foreign/node/src/bdd/message.ts
```diff
@@ -34,15 +34,15 @@ When(
   async function (
     this: TestWorld,
     msgCount: number,
-    streamId: number,
-    topicId: number,
+    _streamId: number,
```
T1B0: even if you use this.stream.id over argument i don't see why prefixing with "_"

foreign/node/src/wire/message/send-messages.command.ts
```diff
@@ -21,7 +21,7 @@
 import { type Id } from '../identifier.utils.js';
 import { serializeSendMessages, type CreateMessage } from './message.utils.js';
 import type { Partitioning } from './partitioning.utils.js';
-import { deserializeVoidResponse } from '../../client/client.utils.js';
+import {deserializeStatusResponse, deserializeVoidResponse} from '../../client/client.utils.js';
```
T1B0: clean up deserializeVoidResponse from import

foreign/node/src/bdd/stream.ts
```diff
@@ -23,7 +23,10 @@ import { Given, When, Then } from "@cucumber/cucumber";
 import type { TestWorld } from './world.js';
 
 Given('I have no streams in the system', async function (this: TestWorld) {
-  assert.deepEqual([], await this.client.stream.list());
+  const streams = await this.client.stream.list();
```
T1B0: i don't understand the point of this change, it used to assert now it delete ? are you fixing the fact that you can't run bdd test twice without restart server because a fresh state is needed ?

T1B0 (changes requested):
i get the limiting buffer size part it's probably safer this way, and the client test is definitly a nice addition ! i don't understand why this MR touch the bdd test, it seems  unrelated to the fix part or i'm missing something ?

foreign/node/src/bdd/stream.ts
oscarArismendi: Yes, my tests were failing because I needed a fresh state. On second thought, I may have gone overboard by adding the delete step in the Given and changing its behavior, using an `After` hook might be a better approach. I’ve updated it in[ my latest commit](https://github.com/apache/iggy/pull/3182/changes/11990e3dfc27e22a1f62e464e6c0e8830c2fab71#diff-6101ebb91c527c6adc0d2df5c01b26213bccdcb26f277810fbe32472602b31d1R56-R62), let me know what you think

foreign/node/src/bdd/message.ts
oscarArismendi: I prefixed it to indicate to [TypeScript that it’s an unused/ignored variable](https://www.typescriptlang.org/tsconfig/#noUnusedParameters). Because while running the tests and investigating why they were failing, I realized the server assigns these values at runtime. You can verify this by going to the main branch and changing the last test in basic_messaging.feature to use a different streamId, the tests will fail, but in this branch they would pass

foreign/node/src/wire/message/send-messages.command.ts
oscarArismendi: Done

oscarArismendi:
> i get the limiting buffer size part it's probably safer this way, and the client test is definitly a nice addition ! i don't understand why this MR touch the bdd test, it seems unrelated to the fix part or i'm missing something ?

@T1B0 I modified the BDD scenarios since` basic_messaging.feature` helps confirm that the changes didn’t alter behavior. I’ve added context in the other threads, but I can revert them if it’s too much for one MR

foreign/node/src/bdd/stream.ts
```diff
@@ -52,3 +52,11 @@ Then(
     assert.ok(await this.client.stream.delete({streamId}));
   }
 );
+
+After(async function (this: TestWorld) {
```
T1B0: this cleanup should not be part of this MR. at the moment we restart a server with --fresh arg to ensure a clean state. bdd test could create other object not cleaned up by this too (users, tokens ...). i think one way to really fix this for all SDK would be to add a deleting/cleaning BDD test and play it at the end, but it's outside of this issue/MR scope.

foreign/node/src/bdd/message.ts
T1B0: honestly it's not really clear either. we should either keep the original name or remove the arguments. maybe a comment/warning would be more appropriate ?

T1B0 (changes requested):
please keep this MR minimalist toward the issue goal (no bdd cleanup / no argument prefixing).

T1B0:
re-reading this, i think this part "[BDD test] initially failed because stream IDs were hardcoded as 0 in the step definitions instead of using the dynamically assigned IDs from the server." is the base of the misunderstanding: stream/topics ID are not hardcoded, they are defined by bdd test scenario and passed as those arguments you prefixed. at the moment BDD test just expect a fresh server (you can restart your iggy server with --fresh opt to do so) so they'll fail if you run them twice without a fresh restart between each runs. 
I think this concern should be another issue/MR subject and we should see with @hubcio about how it should be addressed - probably across all sdks.

Can we just revert the bdd/ part and LGTM ?

oscarArismendi:
@T1B0  reverted BDD changes

T1B0 (approved):
Nice catch, ty for your contribution! LGTM 🚀

hubcio (approved):

spetz (approved):

PR #3188, opened 2026-04-28
fix(ci): honor staged file scope in lint hooks

### LINKED ISSUE #3181: pre-commit hooks do not consistently honor staged-only scope
### Bug description

1. `markdownlint` passes `--staged`, but `scripts/ci/markdownlint.sh` ignores it
However, `markdownlint.sh` only parses the first argument (`--check` or `--fix`) and ignores any following arguments. As a result, once the hook is triggered by a staged Markdown file, it runs against all `**/*.md` files instead of only staged Markdown files.

2. `shellcheck` passes --staged, but `scripts/ci/shellcheck.sh` ignores it
But `shellcheck.sh` also only parses the first argument and then scans all shell scripts in the repository using find. So staging one shell script can cause every shell script in the repository to be checked.

3. Hook trigger scope and execution scope differ
Because these hooks use pass_filenames: false, pre-commit only uses staged files to decide whether the hook should run. After the hook starts, the script decides what to check. For markdownlint and shellcheck, that currently means full-repo scans.
This is surprising because the config includes --staged, which suggests staged-only behavior.

4. `trailing-whitespace.sh` --fix may fail on macOS
BSD/macOS sed -i requires a backup suffix argument, so this can fail during local pre-commit runs on macOS.

5. .....

### Expected behavior
- `markdownlint` --staged should only check/fix staged Markdown files.
-  `shellcheck` --staged should only check staged shell scripts.
-  `CI/default` full-repository checks should remain available via the existing default or an explicit --all.
-  `trailing-whitespace.sh` --fix should work on both `GNU/Linux` and `macOS`.

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

Standing-Man:

Closes #3181

## Local Execution

- Passed / not passed
Passed
Ran


atharvalade (approved):

hubcio (dismissed):
LGTM.

one follow-up request: the `typos` hook (`.pre-commit-config.yaml:129-135`) has the same scope-mismatch this PR addresses. It runs with `pass_filenames: false` and `args: [-w]`, so any staged text file triggers a full-repo scan. Could you flip it to `pass_filenames: true` here, or open a follow-up issue?

Optional cleanup (separate PR, non-blocking): `trailing-whitespace.sh`, `trailing-newline.sh`, `taplo.sh`, `binary-artifacts.sh` already query staged scope internally via `git diff --cached`, but still use `pass_filenames: false`. Aligning them with the markdownlint/shellcheck pattern would let us drop the internal git-diff branches.

Standing-Man:
> Optional cleanup (separate PR, non-blocking): `trailing-whitespace.sh`, `trailing-newline.sh`, `taplo.sh`, `binary-artifacts.sh` already query staged scope internally via `git diff --cached`, but still use `pass_filenames: false`. Aligning them with the markdownlint/shellcheck pattern would let us drop the internal git-diff branches.

Thanks for the suggestion — I’ve created a follow-up issue #3195 to track this and assigned it to myself.

hubcio:
@Standing-Man have you already joined our discord? if yes, then what's your handle? i'd like to discuss a thing about precommit hooks

hubcio (approved):

mmodzelewski (approved):

numinnex (approved):

PR #3191, opened 2026-04-28
feat(go): add delete segments support

### LINKED ISSUE #3190: go-sdk: Expose delete segments command
### Description

Expose the delete segments command in the Go SDK so clients can manage partition segment cleanup over TCP.

### Component

Go SDK

### Proposed solution

_No response_

### Alternatives considered

_No response_

matanper:
Expose the delete segments command in the Go SDK so clients can manage partition segment cleanup over TCP.

Closes #3190

## Rationale

The Go SDK was missing support for the existing `DeleteSegments` TCP command, so Go clients could not trigger partition segment cleanup through the SDK.

## What changed?

The Go SDK now exposes `DeleteSegments(streamId, topicId, partitionId, segmentsCount)`, adds the TCP command code and binary payload serialization, and includes a serialization test for the command wire format.

## Local Execution

- Passed: `go test ./...` from `foreign/go`
- Passed: live Go SDK smoke test against `apache/iggy:0.8.0` on TCP, including login, stream/topic creation, message send, `DeleteSegments`, topic fetch, and cleanup
- Pre-commit hooks not ran

## AI Usage

1. Cursor
2. Entire implementation
3. Ran against apache/iggy:0.8.0 and tested it live
4. Yes


foreign/go/client/tcp/tcp_segment_management_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	iggcon "github.com/apache/iggy/foreign/go/contracts"
+	"github.com/apache/iggy/foreign/go/internal/command"
+)
+
+func TestIggyTcpClient_DeleteSegments(t *testing.T) {
+	clientConn, serverConn := net.Pipe()
+	defer func() {
+		_ = clientConn.Close()
+	}()
```
chengxilo: My concern is that this specific test doesn't quite align with our current patterns, as most tests requiring a server live in the `/bdd` directory.

Long-term, I think we should move this type of testing to `/bdd` and provide a common scenario for every SDK to implement.

I'm also curious to hear what others think about moving this to BDD. Maybe we can add extra steps in `basic_messaging.feature`?

foreign/go/internal/command/segment.go
```diff
@@ -0,0 +1,42 @@
...
+	PartitionId   uint32            `json:"partitionId"`
+	SegmentsCount uint32            `json:"segmentsCount"`
+}
+
+func (d *DeleteSegments) Code() Code {
+	return DeleteSegmentsCode
+}
+
+func (d *DeleteSegments) MarshalBinary() ([]byte, error) {
+	bytes, _ := iggcon.MarshalIdentifiers(d.StreamId, d.TopicId)
```
chengxilo: This error should not be ignored.

chengxilo (commented):
Overall, a solid PR! I left some comment, regarding the test, I'm curious about you and others' opinion.

foreign/go/internal/command/segment.go
matanper: I did it since it's unreachable in tests, I'll return it

foreign/go/client/tcp/tcp_segment_management_test.go
matanper: I removed the unit test and moved the behavioral coverage into `/bdd`, matching how the rest of the Go TCP suite exercises the SDK against a live server.

chengxilo:
lgtm

spetz (approved):

numinnex (approved):

PR #3192, opened 2026-04-28
feat(message_bus): add QUIC, TCP-TLS, WS, WSS transports for SDK clients

hubcio:
Replica plane stays TCP forever: VSR FIFO + view-change timing,
fd-delegation, writev batching all rely on plaintext between trusted
replicas. SDK-client plane gains four transports alongside TCP:

- QUIC: shard-0 terminal (compio-quic CID demux), 1 bidi stream per
  peer, 0-RTT off + listener defense-in-depth reject.
- TCP-TLS: rustls 1.3, no client auth, 0-RTT off, compio-tls behind
  unified TransportConn::run with bounded close_grace shutdown.
- WS: compio-ws over plaintext TCP; pre-upgrade fd cross-shard
  handover keeps fd-delegation on plain TCP only.
- WSS: WebSocketStream over TlsStream; both handshakes run on the
  per-connection install task.

Shared: TransportListener / TransportConn trait family; WebSocketConfig
+ close_grace threaded through MessageBusConfig and applied uniformly
across TCP-TLS, WS, WSS; bounded safe-shutdown (no select! over
stream.shutdown); single-task pump per WS/WSS using compio-ws
cancel-safe read. Bus auth thin: both planes connect unauthenticated;
server-ng gates via LOGIN_USER / LOGIN_WITH_PAT and future
LOGIN_REPLICA. Ping announces replica_id only; no subprotocol, no
ALPN, no MAC. Per-connection metadata flows via
IggyMessageBus::client_meta; ShardFramePayload setup variants carry
ClientConnMeta end to end.


core/message_bus/src/transports/ws.rs
```diff
@@ -0,0 +1,427 @@
...
+
+/// Decode one consensus `Message<GenericHeader>` from a raw WS Binary
+/// payload. Reused by [`super::wss`] over the TLS-protected channel.
+pub(in crate::transports) fn decode_consensus_frame(
+    body: &[u8],
+) -> Result<Message<GenericHeader>, FrameDecodeError> {
+    if body.len() < iggy_binary_protocol::HEADER_SIZE {
+        return Err(FrameDecodeError::BadHeader);
+    }
+    let total_size = u32::from_le_bytes(
```
atharvalade: `framing.rs` has a `const _: () = { assert!(offset_of!(GenericHeader, size) == 48) }` guard but this function duplicates the same `48..52` magic without one. If `GenericHeader` layout ever shifts, TCP/QUIC get a compile error while WS/WSS silently read the wrong bytes.

core/message_bus/src/transports/ws.rs
```diff
@@ -0,0 +1,427 @@
...
+
+/// Drive the WS connection until shutdown, peer Close, or an
+/// unrecoverable error.
+#[allow(clippy::future_not_send)]
+async fn run_pump(ws: &mut WebSocketStream<TcpStream>, ctx: ActorContext) {
+    let ActorContext {
+        in_tx,
+        rx,
+        shutdown,
+        label,
```
atharvalade, line 190: `run_pump` drops `max_message_size` into `..` and `decode_consensus_frame` hardcodes `framing::MAX_MESSAGE_SIZE`. TCP and QUIC paths honor the per-bus config value, so an operator who lowers `max_message_size` gets enforcement on TCP/QUIC but not WS/WSS.

core/server-ng/src/dedup.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// Mark the handler as starting. Returns `true` if the slot was
+    /// inserted; `false` if an entry for `(client, request)` already
+    /// existed (caller must consult [`Self::lookup`] for the specific
+    /// state).
+    pub fn mark_in_flight(&mut self, client: u128, request: u64) -> bool {
+        let state = self
+            .per_client
+            .entry(client)
+            .or_insert_with(|| PerClient::with_capacity(self.per_client_capacity));
+        if state.find(request).is_some() {
```
atharvalade: `lookup` treats TTL-expired Done entries as `Fresh`, but `mark_in_flight` calls `find()` without a TTL check. A client retrying after TTL expiry sees `Fresh` from lookup then gets `false` from `mark_in_flight` because the physical slot still exists. The retry is silently dropped.

core/message_bus/src/transports/ws.rs
```diff
@@ -0,0 +1,427 @@
...
+    #[allow(clippy::future_not_send)]
+    async fn ws_loopback_round_trip() {
+        let (client_ws, server_ws) = ws_pair().await;
+        let server_conn = WsTransportConn::new_server(server_ws);
+        let client_conn = WsTransportConn::new_client(client_ws);
+
+        let (server_out, server_in, server_shutdown, server_handle) = drive(server_conn);
+        let (client_out, client_in, client_shutdown, client_handle) = drive(client_conn);
+
+        client_out
```
atharvalade, line 409: TCP-TLS's `drive_close` calls `tls.shutdown()` which sends `close_notify`, but WSS just does `ws.close()` + drop. The peer's rustls sees an unexpected EOF on the record layer, which can trigger false-positive alerts in TLS-aware load balancers or WAFs sitting in front.

atharvalade (changes requested):
found these during first round of review... I'll continue to review later. Overall seems good

core/message_bus/src/transports/ws.rs
hubcio: fixed

core/message_bus/src/transports/ws.rs
hubcio: fixed

core/server-ng/src/dedup.rs
hubcio: i removed all dedup related stuff in this PR. it'll be done later, so comment is obsolete.

core/message_bus/src/transports/ws.rs
hubcio: fixed

numinnex (approved):

spetz (approved):

PR #3194, opened 2026-04-29
fix(connectors): replace overloaded InvalidRecord with distinct error variants

### LINKED ISSUE #3176: Error::InvalidRecord is overloaded across unrelated failure modes
In the Iceberg sink's `write_data_with_options` function (`router/mod.rs`), five different failure modes all map to the same `Error::InvalidRecord`:

- Arrow schema mapping failure (line ~224) — config/schema problem
- Record batch deserialization failure (line ~239) — data format problem
- Parquet write failure (line ~243) — I/O problem
- Transaction apply failure (line ~262) — catalog state problem
- Transaction commit failure (line ~271) — catalog I/O problem

This makes it impossible to programmatically distinguish between a schema mismatch (fix your table definition), a corrupt message (skip and continue), and a catalog outage (retry later).

Logs help somewhat, but callers of this function only see `Error::InvalidRecord` regardless of root cause.

**Fix**: Introduce distinct error variants: `Error::SchemaMismatch`, `Error::WriteFailure`,`Error::CatalogError`, etc. Or at minimum use `Error::InitError(detail)` for the cases that already accept a string, and reserve `InvalidRecord` for actual data-level issues.

atharvalade:

Closes #3176

## Rationale

`Error::InvalidRecord` was used for five unrelated failure modes in the Iceberg sink's `write_data` function, making it impossible for callers to distinguish schema mismatches from I/O failures from catalog outages.

## What changed?

The Iceberg sink mapped Arrow schema conversion errors, Parquet write failures, and Iceberg catalog transaction failures all to `Error::InvalidRecord`. Callers could not programmatically decide whether to fix a table definition, skip a corrupt message, or retry a catalog outage.

Three new SDK error variants — `SchemaMismatch(String)`, `WriteFailure(String)`, `CatalogError(String)` — replace the overloaded `InvalidRecord` at the appropriate call sites. `InvalidRecord` is preserved only for the genuine record-batch deserialization error.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

- Opu 4.6
- Writing comments, writing PR Description
- Yes


core/connectors/sdk/src/lib.rs
```diff
@@ -416,4 +416,16 @@ pub enum Error {
     /// failures so that circuit breakers are not tripped by bad data.
     #[error("Permanent HTTP error: {0}")]
     PermanentHttpError(String),
+    /// The source schema could not be mapped to the destination schema.
+    /// Indicates a table definition or configuration problem — not a data issue.
```
slbotbm: The "- not a data issue" part seems irrelevant

core/connectors/sdk/src/lib.rs
```diff
@@ -425,4 +425,16 @@ pub enum Error {
...
+    /// The source schema could not be mapped to the destination schema.
+    /// Indicates a table definition or configuration problem
+    #[error("Schema mismatch: {0}")]
+    SchemaMismatch(String),
+    /// An I/O failure while writing data (e.g. Parquet serialization, file
+    /// writer close). Distinct from record-level validation errors.
+    #[error("Write failure: {0}")]
+    WriteFailure(String),
+    /// A catalog or transaction-level failure (e.g. applying or committing an
+    /// Iceberg transaction). Callers may retry on transient catalog outages.
```
hubcio: doc "callers may retry on transient catalog outages" is misleading. `action.apply()` at `router/mod.rs:213` is in-memory transaction prep - deterministic failures (invalid partition spec, schema validation) cannot be retried. only `tx.commit(catalog)` at `router/mod.rs:222` hits the network. suggest dropping the retry claim, or splitting into `ApplyError` (deterministic) vs `CommitError` (transient-eligible).

hubcio (changes requested):

core/connectors/sdk/src/lib.rs
atharvalade: split `CatalogError` into `TransactionApplyError` and `CatalogCommitError`

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -197,13 +197,13 @@ async fn write_data(
         })?;
         writer.write(batch_data).await.map_err(|err| {
             error!("Error while writing record batch: {}", err);
-            Error::InvalidRecord
+            Error::WriteFailure(err.to_string())
```
hubcio: two problems in this batch loop (lines 193-201).

first: line 196 still maps to `Error::InvalidRecord` (no payload). the PR's stated goal is to replace the overloaded `InvalidRecord`, but this site is left as the unit variant and loses the underlying arrow_json reader error message. change to `Error::InvalidRecordValue(err.to_string())` so the diagnostic survives.

second: an error at line 196 OR 198-201 returns via `?` without calling `writer.close()`. `DataFileWriter` has no `Drop` impl, so any RecordBatches already buffered in the writer are silently dropped, and any partially flushed Parquet bytes that already went to `file_io` become orphaned (you also can't recover the `Vec<DataFile>` metadata without `close()`). combined with `runtime/src/sink.rs:467` `AutoCommitWhen::PollingMessages` (offset commits at poll time, before consume runs), the messages whose batches were buffered are lost on restart. wrap the loop so `writer.close()` runs on early-return (scope guard / explicit close before propagating), or accept the leak and document it.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -197,13 +197,13 @@ async fn write_data(
...
         })?;
     }
 
     let data_files = writer.close().await.map_err(|err| {
         error!("Error while writing data records to Parquet file: {}", err);
-        Error::InvalidRecord
+        Error::WriteFailure(err.to_string())
```
hubcio: `writer.close()` failure here can leave a partially-flushed Parquet file on the object store with no cleanup, because the `Vec<DataFile>` is never produced and no transaction commit happens. the `WriteFailure` name reads as "transient I/O, safe to retry" but retry creates more orphans (each call generates a fresh uuid path at line 119). either add a best-effort cleanup before propagating, or document on the `WriteFailure` variant that it may leave orphaned files and cleanup is consumer-side.

core/connectors/sdk/src/lib.rs
```diff
@@ -425,4 +425,20 @@ pub enum Error {
...
+    #[error("Write failure: {0}")]
+    WriteFailure(String),
+    /// In-memory transaction preparation failed (e.g. invalid partition spec,
+    /// schema validation). These failures are deterministic and not retryable.
+    #[error("Transaction apply error: {0}")]
+    TransactionApplyError(String),
+    /// A catalog commit failed over the network (e.g. REST catalog I/O,
+    /// conflict). Transient failures may be retried.
+    #[error("Catalog commit error: {0}")]
+    CatalogCommitError(String),
```
hubcio, line 448: "transient failures may be retried" is dangerously thin guidance. `iceberg::transaction::Transaction::commit(self, &dyn Catalog)` consumes `self`, so a naive retry isn't even possible without rebuilding the transaction + data files. and even then there's no idempotency token: if the REST catalog applied the commit but the response was lost, a rebuilt retry would add the same data files into a second snapshot -> duplicate rows.

suggest rewriting to something like: "a catalog commit failed. `Transaction` is consumed by `commit()`, so retrying requires rebuilding the transaction from new data files. retry is not idempotent: callers must verify via the catalog whether the original commit was applied before retrying, otherwise data may be duplicated."

core/connectors/sdk/src/lib.rs
```diff
@@ -425,4 +425,20 @@ pub enum Error {
...
+    #[error("Schema mismatch: {0}")]
+    SchemaMismatch(String),
+    /// An I/O failure while writing data (e.g. Parquet serialization, file
+    /// writer close). Distinct from record-level validation errors.
+    #[error("Write failure: {0}")]
+    WriteFailure(String),
+    /// In-memory transaction preparation failed (e.g. invalid partition spec,
+    /// schema validation). These failures are deterministic and not retryable.
+    #[error("Transaction apply error: {0}")]
+    TransactionApplyError(String),
```
hubcio, line 441: "deterministic and not retryable" is too absolute. iceberg's `Error` carries a `retryable()` flag and the apply step interacts with partition spec / schema state that this code doesn't statically classify. better wording: "typically deterministic in the current iceberg version; check `Error::retryable()` on the underlying iceberg error to decide." or just drop the retryability hint and let the caller route based on the inner string.

core/connectors/sdk/src/lib.rs
```diff
@@ -425,4 +425,20 @@ pub enum Error {
     /// failures so that circuit breakers are not tripped by bad data.
     #[error("Permanent HTTP error: {0}")]
     PermanentHttpError(String),
+    /// The source schema could not be mapped to the destination schema.
+    /// Indicates a table definition or configuration problem
```
hubcio: missing terminal period to match `PermanentHttpError` style at line 425. the other three new doc comments already end with periods, just this one.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -225,7 +225,7 @@ async fn write_data(
             table.metadata().uuid(),
             err
         );
-        Error::InvalidRecord
+        Error::CatalogCommitError(err.to_string())
```
hubcio: `err.to_string()` here (and at the four sister sites: lines 181, 200, 206, 219) only renders the top-level `Display` impl and drops the iceberg `Error::source()` chain. underlying parquet / arrow / io causes get swallowed - the new typed variant tells you which stage failed but the payload string is no more informative than before.

the whole stated win of distinct variants is better diagnostics, so it's worth wiring this through. one option: walk `err.source()` manually and join the chain into the string. another: log the chain alongside the existing `error!` line. `format!("{err:#}")` only works for anyhow, not thiserror, so don't reach for that.

hubcio (changes requested):
out-of-scope follow-ups (already tracked, do not block this PR):

- the `AutoCommitWhen::PollingMessages` offset-before-consume data-loss path is already filed as issue #2928, with #2927 covering the paired bug (sink `consume()` return value discarded). PR #3180 fixes #2927; #2928 remains open. with the new typed variants in this PR, every error path through `write_data` lands on that same lost-offset hazard, so worth flagging on #2928 that PR 3194 increases its visibility.
- `core/connectors/sinks/iceberg_sink/src/router/dynamic_router.rs:145-165` runs `write_data` per table; first failure aborts, earlier tables already committed via `tx.commit()`, remaining tables stay unwritten, offset advances once for the whole batch. iceberg `Transaction` is single-table so structural fix needs a separate design.
- no tests added for the new variant mappings. would be good to cover the 5 map_err sites and the early-return-from-batch-loop path.
- same overloading exists in `core/connectors/sources/postgres_source/src/lib.rs` (24 `Error::InvalidRecord` sites mixing sqlx I/O, `row.try_get` column decode, `simd_json` serialization, and type-cast schema-mismatch). issue #3176 was scoped to iceberg, so it's reasonable that this PR is too -- but please file a follow-up issue mirroring #3176 for postgres_source so the same taxonomy split can land there.

the variant split itself is a reasonable refactor and worth landing once the doc claims are softened and line 196 is addressed.

hubcio:
@EdgarModesto23 you might also be interested in this PR because it touches iceberg connector.

EdgarModesto23:
@hubcio Pretty much agree with every change you requested. One thing I want to mention is that the original issue is just asking for distinct error variants, so if we wanted to keep the scope aligned with the issue, maybe we can ask @atharvalade to document the current leaks and address retries and orphans on a different issue. Leave it up to you and the author tho :)

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: Fixed, now uses InvalidRecordValue with chain

Added explicit writer.close() on failure

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: Documented orphaned files on `WriteFailure` variant

core/connectors/sdk/src/lib.rs
atharvalade: Rewrote doc with idempotency warning

core/connectors/sdk/src/lib.rs
atharvalade: Softened, defers to underlying error

core/connectors/sdk/src/lib.rs
atharvalade: Added

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: Added `format_error_chain` walking `source()`

atharvalade:
> @hubcio Pretty much agree with every change you requested. One thing I want to mention is that the original issue is just asking for distinct error variants, so if we wanted to keep the scope aligned with the issue, maybe we can ask @atharvalade to document the current leaks and address retries and orphans on a different issue. Leave it up to you and the author tho :)

Agreed, keeping this PR scoped to the error variant split and doc accuracy. Filed #3314 for the `postgres_source` overloading. Happy to open a separate one for writer cleanup and retry semantics if needed. Can you please check @EdgarModesto23

EdgarModesto23:
Yes please, go ahead and file a separate issue for it. I can take care of it :). The PR looks good to me.

spetz (approved):

numinnex (approved):

hubcio (approved):

PR #3196, opened 2026-04-29
fix(connectors): handle unrecognized types with explicit coverage and diagnostics

### LINKED ISSUE #3175: Unrecognized Postgres types silently coerce to String
The default branch in `extract_column_value_static` attempts to read any unrecognized Postgres type as `Option<String>`:

```rust
_ => {
    let value: Option<String> = row
        .try_get(column_index)
        .map_err(|_| Error::InvalidRecord)?;
    Ok(value
        .map(serde_json::Value::String)
        .unwrap_or(serde_json::Value::Null))
}
```

If the type cannot be deserialized as String (arrays, composite types, ranges, etc.), it returns `Error::InvalidRecord` with no indication of which column or type caused the failure. There are many common Postgres types not covered: arrays (`_INT4`, `_TEXT`), enums, `INTERVAL`, `INET`,`CIDR`, `MACADDR`, `TSVECTOR`, `POINT`, `JSONPATH`, etc.

**Fix**: Log the type name and column name before falling through to the default. Consider using `row.try_get_raw()` and converting to a text representation as a more robust fallback. At minimum: `warn!("Unknown column type '{}' for column '{}', attempting String fallback", type_name, column.name());`

atharvalade:

Closes #3175

## Rationale

Unrecognized Postgres types silently coerced to String, returning opaque `InvalidRecord` on failure with no column or type context.

## What changed?

The default `_` branch in `extract_column_value` attempted to read any unrecognized Postgres type as `Option<String>`, failing silently with `Error::InvalidRecord` and no indication of which column or type caused it.

Added explicit handlers for DATE, TIME, INTERVAL, OID, NAME, BPCHAR, and 13 array types (_BOOL, _INT2, _INT4, _INT8, _FLOAT4, _FLOAT8, _TEXT, _VARCHAR, _CHAR, _BPCHAR, _UUID, _JSON, _JSONB). The default branch now logs a warning with column name and type before attempting String fallback, and returns `InvalidRecordValue` with a diagnostic message on failure.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Opus 4.6
2. Scaffolding, Writing comments, Writing PR Description
3. Verified via cargo check, clippy, fmt, and full test suite (24/24 pass)


core/connectors/sources/postgres_source/src/lib.rs
slbotbm, line 1: The current code matches against wrong literals, which means that everything is coerced into a string.

Current literals (wrong):
```
- _BOOL
- _INT2
- _INT4
- _INT8
- _FLOAT4
- _FLOAT8
- _TEXT
- _VARCHAR
- _CHAR
- _BPCHAR
- _UUID
- _JSON
- _JSONB
```
Correct literals:
```
- BOOL[]
- INT2[]
- INT4[]
- INT8[]
- FLOAT4[]
- FLOAT8[]
- TEXT[]
- VARCHAR[]
- CHAR[]
- UUID[]
- JSON[]
- JSONB[]
```

slbotbm: Also, the code treats BPCHAR as a possible option, but sqlx reports the display name CHAR[], so the BPCHAR-style array match can never fire.

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -995,10 +1021,130 @@ impl PostgresSource {
...
+                Ok(value
+                    .map(|arr| {
+                        serde_json::Value::Array(
+                            arr.into_iter().map(serde_json::Value::Bool).collect(),
+                        )
+                    })
+                    .unwrap_or(serde_json::Value::Null))
+            }
+            "_INT2" => {
+                let value: Option<Vec<i16>> = row
```
slbotbm: For arms decoding `Option<Vec<T>>`, the code assumes every element is non-null, but postgresql can contain null elements. Thus, if a null value is supplied to the current code, sqlx will fail to decode it.

slbotbm: Something like `Option<Vec<Option<T>>>` could mitigate this.

slbotbm (commented):
/author

slbotbm (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -1026,6 +1172,50 @@ fn format_offset_value(value: &str) -> String {
...
+        let total_secs = interval.microseconds / 1_000_000;
+        let remaining_us = (interval.microseconds % 1_000_000).unsigned_abs();
+        let hours = total_secs / 3600;
+        let mins = (total_secs % 3600) / 60;
+        let secs = total_secs % 60;
+        if remaining_us != 0 {
+            parts.push(format!(
+                "{:02}:{:02}:{:02}.{:06}",
+                hours, mins, secs, remaining_us
+            ));
```
slbotbm: This part formats negative numbers incorrectly. For example, the above code would convert `-1.5 seconds` into `00:00:-1.500000`

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -911,7 +913,7 @@ impl PostgresSource {
                     .map(|v| serde_json::Value::from(v as i64))
                     .unwrap_or(serde_json::Value::Null))
             }
-            "INT4" => {
+            "INT4" | "OID" => {
                 let value: Option<i32> = row
                     .try_get(column_index)
                     .map_err(|_| Error::InvalidRecord)?;
```
slbotbm: the code current treats `OID` as i32, but in postgres, that is u32, which can lead batch processing failures. I feel decoding this to an Oid type would be a good approach. 

I also noticed you're decoding individual `OID`'s, but not but not `Vec<OID>`

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -952,14 +954,30 @@ impl PostgresSource {
...
+                let value: Option<NaiveTime> = row
+                    .try_get(column_index)
+                    .map_err(|_| Error::InvalidRecord)?;
+                Ok(value
+                    .map(|t| serde_json::Value::String(t.to_string()))
+                    .unwrap_or(serde_json::Value::Null))
+            }
             "TIMESTAMP" | "TIMESTAMPTZ" => {
                 let value: Option<DateTime<Utc>> = row
                     .try_get(column_index)
```
slbotbm: though not related to your changes, in sqlx, `TIMESTAMP` is `chrono::NaiveDateTime`, while `TIMESTAMPZ` is `chrono::DateTime<Utc>`. If the user passes `TIMESTAMP`, the conversion will fail.

slbotbm (commented):
Separately, the current code also does not handle the following types:
  - TIMETZ
  - TIMETZ[]
  - DATE[]
  - TIME[]
  - TIMESTAMP[]
  - TIMESTAMPTZ[]
  - INTERVAL[]

slbotbm (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: Fixed, all array arms now match the correct sqlx display names

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: Good catch, all array arms now decode as `Option<Vec<Option<T>>>` so null elements within arrays map to JSON null instead of causing a decode failure

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: fixed, the function now takes `unsigned_abs()` of the microseconds field and prepends a `-` sign when negative

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: Agreed, OID is now decoded as i32 on the wire but cast through u32 to u64 before JSON serialization to preserve the full unsigned range

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: Good point, split into separate match arms

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -995,10 +1037,253 @@ impl PostgresSource {
...
+                warn!(
+                    "Column '{column_name}' has unrecognized Postgres type '{type_name}', \
+                     attempting String conversion"
+                );
+                let value: Option<String> = row.try_get(column_index).map_err(|e| {
+                    error!("Failed to read column '{column_name}' (type '{type_name}'): {e}");
+                    Error::InvalidRecordValue(format!(
+                        "column '{column_name}' has unsupported Postgres type '{type_name}'"
+                    ))
+                })?;
```
slbotbm: String is only compatible with text-like types, not types like `INET`, `CIDR`, `MACADDR`, `MONEY`, `BIT/VARBIT`. This fallback will fail hard for those cases.

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -919,6 +921,14 @@ impl PostgresSource {
...
                     .unwrap_or(serde_json::Value::Null))
             }
+            "OID" => {
+                let value: Option<i32> = row
+                    .try_get(column_index)
+                    .map_err(|_| Error::InvalidRecord)?;
+                Ok(value
+                    .map(|v| serde_json::Value::from(v as u32 as u64))
+                    .unwrap_or(serde_json::Value::Null))
+            }
```
slbotbm: For `OID` and `OID[]`, could you write a temporary test and check whether this works? From what I understand, sqlx may reject this since `try_get` checks compatibility before decoding, and fails on mismatched SQL/Rust types. `sqlx::postgres::types::Oid` being the native type for `OID`, `i32` would still be rejected.

slbotbm (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: the default branch now uses `try_get_raw()` + `as_str()` which reads the raw text representation directly from the wire without sqlx type-checking, so it works for any type Postgres can render as text.

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: now decoding via `sqlx::postgres::types::Oid` which is the native sqlx type for OID (wraps u32, implements Decode and `PgHasArrayType`); OID[] also uses `Vec<Option<Oid>>` separately from `INT4[]`

Cargo.lock
hubcio, line 1: why did u do this many changes in Cargo.lock? you're only adding chrono dep to postgres_source.

hubcio (changes requested):

Cargo.lock
atharvalade: Rebased cleanly now, please check

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -1000,22 +1010,54 @@ impl PostgresSource {
...
                     .unwrap_or(serde_json::Value::Null))
             }
-            "VARCHAR" | "TEXT" | "CHAR" => {
+            "VARCHAR" | "TEXT" | "CHAR" | "NAME" | "BPCHAR" | "TIMETZ" => {
                 let value: Option<String> = row
                     .try_get(column_index)
                     .map_err(|_| Error::InvalidRecord)?;
                 Ok(value
                     .map(serde_json::Value::String)
                     .unwrap_or(serde_json::Value::Null))
```
slbotbm: `TIMETZ` does not decode to `String` - it has a dedicated type: `PgTimeTz`. Current code will fail is `TIMETZ` is supplied.

core/connectors/sources/postgres_source/src/lib.rs
slbotbm: Your change assumes that the bytes will be UTF-8 compatible, and thus can be converted to string using `as_str`. This may not always be the case, and the connector will fail in those cases.

slbotbm (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
slbotbm: The fix for this does not seem to be trivial, and a follow up PR could fix this instead.

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: `TIMETZ` now decodes via `sqlx::postgres::types::PgTimeTz` which gives `NaiveTime` + `FixedOffset`, formatted as `"HH:MM:SS+offset"`

atharvalade: the default branch now tries `as_str(`) first and on failure falls back to base64-encoding the raw bytes via `as_bytes()`, so binary-format types like custom composites won't crash the connector

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -1051,17 +1101,342 @@ fn extract_column_value(
...
+        parts.push(format!(
+            "{months} mon{}",
+            if months.abs() != 1 { "s" } else { "" }
+        ));
+    }
+    if interval.days != 0 {
+        parts.push(format!(
+            "{} day{}",
+            interval.days,
+            if interval.days.abs() != 1 { "s" } else { "" }
```
Standing-Man: An overflow occurs if days is `i32::MIN`.

```
postgres=# SELECT make_interval(days => -2147483648);
  make_interval   
------------------
 -2147483648 days
(1 row)
```

Standing-Man (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: switched from `.abs() `to `.unsigned_abs() `which returns `u32` and handles `i32::MIN` without overflow

Standing-Man (approved):
LGTM otherwise. The only issue is that the unknown-type fallback nominally returns a string, but it doesn’t actually go through PostgreSQL’s text conversion, so the output may be unreliable.

slbotbm (approved):
other than alan's final concern, lgtm

hubcio (approved):

krishvishal (approved):

PR #3197, opened 2026-04-29
fix(connectors): Add unwrap_envelope transform and envelope detection to fix source-sink format mismatch

### LINKED ISSUE #3174: No shared contract between source output and sink input format
The Postgres source wraps all row data in a `DatabaseRecord` envelope:

```json
{
  "table_name": "tpch.lineitem",
  "operation_type": "SELECT",
  "timestamp": "2026-04-24T19:57:00Z",
  "data": {"id": 1, "l_orderkey": 123, ...},
  "old_data": null
}
```

The Iceberg sink's Arrow JSON reader expects flat JSON matching the target table schema:

```json
{"id": 1, "l_orderkey": 123, ...}
```

These two formats are fundamentally incompatible. There is no shared message schema spec or envelope definition between sources and sinks. Connecting the Postgres source to the Iceberg sink without the (newly added) `flat_json_output = true` flag produces silent failures because Arrow maps the nested `data` object to the top-level `id` field as null, then fails on non-nullable constraints.

**Fix**: Either define a standard connector message envelope that all sinks know how to unwrap, or have the sink explicitly handle the `DatabaseRecord` format (extract the `data` field before Arrow conversion). The current workaround (`flat_json_output`) is a band-aid that puts the burden on the user to know about the incompatibility.

atharvalade:

Closes #3174

## Rationale

Sources (e.g. Postgres) wrap row data in a `DatabaseRecord` envelope while sinks (e.g. Iceberg) expect flat JSON matching the target table schema — no shared contract exists, producing silent null failures.

## What changed?

The Postgres source emits `{table_name, operation_type, timestamp, data: {...}, old_data}` envelopes, but the Iceberg sink's Arrow JSON reader maps these nested structures to top-level fields as null, silently violating non-nullable constraints.

This adds a reusable `unwrap_envelope` transform to the connector SDK that extracts a nested field (e.g. `data`) and promotes it as the top-level payload, plus explicit envelope detection in the Iceberg sink that errors with an actionable message instead of failing silently.

## Local Execution

- Passed
- Pre-commit hooks ran (fmt, clippy, license-headers, trailing-whitespace, trailing-newline all clean; 119 tests pass across SDK + integration suites)

## AI Usage

1. Opus 4.6
2. used for codebase exploration and following existing transform patterns
3. All 8 new unit tests pass locally, clippy/fmt clean, existing 111 tests unaffected

### DIFF
--- a/core/connectors/sdk/src/transforms/json/unwrap_envelope.rs
+++ b/core/connectors/sdk/src/transforms/json/unwrap_envelope.rs
@@ -1,0 +1,213 @@
+/* Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use crate::{
+    DecodedMessage, Error, Payload, TopicMetadata, transforms::unwrap_envelope::UnwrapEnvelope,
+};
+use simd_json::OwnedValue;
+use tracing::warn;
+
+impl UnwrapEnvelope {
+    pub(crate) fn transform_json(
+        &self,
+        _metadata: &TopicMetadata,
+        mut message: DecodedMessage,
+    ) -> Result<Option<DecodedMessage>, Error> {
+        let Payload::Json(OwnedValue::Object(ref mut map)) = message.payload else {
+            return Ok(Some(message));
+        };
+
+        let Some(inner) = map.remove(self.field.as_str()) else {
+            warn!(
+                "unwrap_envelope: field '{}' not found in payload, passing through unchanged",
+                self.field
+            );
+            return Ok(Some(message));
+        };
+
+        message.payload = Payload::Json(inner);
+        Ok(Some(message))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::transforms::Transform;
+    use crate::transforms::json::test_utils::{
+        create_raw_test_message, create_test_message, create_test_topic_metadata,
+        extract_json_object,
+    };
+    use crate::transforms::unwrap_envelope::{UnwrapEnvelope, UnwrapEnvelopeConfig};
+    use crate::{DecodedMessage, Payload};
+    use simd_json::OwnedValue;
+
+    #[test]
+    fn should_extract_data_field_from_database_record_envelope() {
+        let transform = UnwrapEnvelope::new(UnwrapEnvelopeConfig {
+            field: "data".to_string(),
+        });
+        let msg = create_test_message(
+            r#"{
+                "table_name": "tpch.lineitem",
+                "operation_type": "SELECT",
+                "timestamp": "2026-04-24T19:57:00Z",
+                "data": {"id": 1, "l_orderkey": 123, "l_partkey": 456},
+                "old_data": null
+            }"#,
+        );
+        let result = transform
+            .transform_json(&create_test_topic_metadata(), msg)
+            .unwrap()
+            .unwrap();
+        let json_obj = extract_json_object(&result).unwrap();
+        assert_eq!(json_obj.len(), 3);
+        assert!(json_obj.contains_key("id"));
+        assert!(json_obj.contains_key("l_orderkey"));
+        assert!(json_obj.contains_key("l_partkey"));
+        assert!(!json_obj.contains_key("table_name"));
+        assert!(!json_obj.contains_key("operation_type"));
+    }
+
+    #[test]
+    fn should_handle_missing_field_gracefully() {
+        let transform = UnwrapEnvelope::new(UnwrapEnvelopeConfig {
+            field: "data".to_string(),
+        });
+        let msg = create_test_message(r#"{"table_name": "users", "operation_type": "INSERT"}"#);
+        let result = transform
+            .transform_json(&create_test_topic_metadata(), msg)
+            .unwrap()
+            .unwrap();
+        let json_obj = extract_json_object(&result).unwrap();
+        assert_eq!(json_obj.len(), 2);
+        assert!(json_obj.contains_key("table_name"));
+    }
+
+    #[test]
+    fn should_handle_scalar_inner_value() {
+        let transform = UnwrapEnvelope::new(UnwrapEnvelopeConfig {
+            field: "payload".to_string(),
+        });
+        let msg = create_test_message(r#"{"payload": "just a string", "meta": 1}"#);
+        let result = transform
+            .transform_json(&create_test_topic_metadata(), msg)
+            .unwrap()
+            .unwrap();
+        match &result.payload {
+            Payload::Json(OwnedValue::String(s)) => assert_eq!(s.as_str(), "just a string"),
+            other => panic!("Expected Json(String), got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn should_handle_null_inner_value() {
+        let transform = UnwrapEnvelope::new(UnwrapEnvelopeConfig {
+            field: "old_data".to_string(),
+        });
+        let msg = create_test_message(r#"{"data": {"id": 1}, "old_data": null}"#);
+        let result = transform
+            .transform_json(&create_test_topic_metadata(), msg)
+            .unwrap()
+            .unwrap();
+        match &result.payload {
+            Payload::Json(OwnedValue::Static(simd_json::StaticNode::Null)) => {}
+            other => panic!("Expected Json(Null), got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn should_pass_through_non_json_payload() {
+        let transform = UnwrapEnvelope::new(UnwrapEnvelopeConfig {
+            field: "data".to_string(),
+        });
+        let msg = create_raw_test_message(vec![1, 2, 3, 4]);
+        let result = transform
+            .transform(&create_test_topic_metadata(), msg)
+            .unwrap()
+            .unwrap();
+        if let Payload::Raw(bytes) = &result.payload {
+            assert_eq!(*bytes, vec![1u8, 2, 3, 4]);
+        } else {
+            panic!("Expected Raw payload");
+        }
+    }
+
+    #[test]
+    fn should_pass_through_non_object_json() {
+        let transform = UnwrapEnvelope::new(UnwrapEnvelopeConfig {
+            field: "data".to_string(),
+        });
+        let msg = DecodedMessage {
+            id: None,
+            offset: None,
+            checksum: None,
+            timestamp: None,
+            origin_timestamp: None,
+            headers: None,
+            payload: Payload::Json(OwnedValue::Array(Box::new(vec![
+                OwnedValue::Static(simd_json::StaticNode::I64(1)),
+                OwnedValue::Static(simd_json::StaticNode::I64(2)),
+            ]))),
+        };
+        let result = transform
+            .transform(&create_test_topic_metadata(), msg)
+            .unwrap()
+            .unwrap();
+        match &result.payload {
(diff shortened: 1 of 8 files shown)


hubcio:
@atharvalade please rebase this PR

/author

atharvalade:
> @atharvalade please rebase this PR
> 
> /author

done

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -230,6 +243,13 @@ async fn write_data(
     Ok(())
 }
 
+fn looks_like_envelope(value: &simd_json::OwnedValue) -> bool {
+    let simd_json::OwnedValue::Object(obj) = value else {
+        return false;
+    };
+    obj.contains_key("table_name") && obj.contains_key("data")
+}
```
hubcio: this heuristic checks 2 of the 5 keys on `DatabaseRecord` (`table_name`, `operation_type`, `timestamp`, `data`, `old_data` - see `core/connectors/sources/postgres_source/src/lib.rs:110-116`).

false positives: any legit iceberg table modeling audit logs, catalogs, or CDC metadata that happens to have `table_name` + `data` columns gets the whole batch rejected.

false negatives: Debezium / Kafka-Connect envelopes use `before` / `after` / `op` / `source` / `payload` and slip right through into `JsonArrowReader`, which then writes nulls - the exact bug #3174 is supposed to fix.

also it bakes a postgres-source shape into a generic iceberg sink. options: tighten to the full 4-or-5 key shape, or move detection behind an opt-in `detect_envelope` config flag, or drop sink-side sniffing entirely and rely on `unwrap_envelope` + docs.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -170,6 +170,19 @@ async fn write_data(
         })
         .collect();
 
+    if let Some(first) = msgs.first()
+        && looks_like_envelope(first)
```
hubcio: the sniff only looks at `msgs.first()`. mixed batch where the first message is flat and the rest are envelopes (or vice versa) skips detection entirely - envelope rows then hit `JsonArrowReader` and write nulls, which is the silent-corruption mode #3174 was opened for. envelope-first drops the whole batch including any valid flat rows.

either scan all messages, or document and enforce a batch-homogeneity invariant somewhere upstream.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -170,6 +170,19 @@ async fn write_data(
...
+    {
+        error!(
+            "Incoming JSON appears to be wrapped in a source envelope \
+             (detected 'table_name' + 'data' fields). The Iceberg sink \
+             expects flat JSON matching the target table schema. Add an \
+             'unwrap_envelope' transform with field = \"data\" to your \
+             connector config to extract the inner payload."
+        );
+        return Err(Error::InvalidRecord);
```
hubcio: `Error::InvalidRecord` is a unit variant; the actionable hint ("add an unwrap_envelope transform with field = data") lives only in the `error!` log and is lost the moment the caller receives the error. `Error::InvalidRecordValue(String)` already exists (`sdk/src/lib.rs:389`) and is used in delta_sink and influxdb_source - switch to it and carry the hint in the variant.

this is also a concrete instance of #3176 (overloaded `InvalidRecord`).

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -170,6 +170,19 @@ async fn write_data(
...
 
+    if let Some(first) = msgs.first()
+        && looks_like_envelope(first)
+    {
+        error!(
+            "Incoming JSON appears to be wrapped in a source envelope \
+             (detected 'table_name' + 'data' fields). The Iceberg sink \
+             expects flat JSON matching the target table schema. Add an \
+             'unwrap_envelope' transform with field = \"data\" to your \
+             connector config to extract the inner payload."
```
hubcio: two issues with this error message:

1. it hardcodes `field = "data"`, which only matches the postgres envelope shape. if/when detection broadens to Debezium (`after`) or other shapes, the suggestion is wrong.

2. no connector id / plugin id in the log line, so in a multi-tenant deployment with several iceberg sinks you cannot tell which one rejected the batch.

core/connectors/sdk/src/transforms/json/unwrap_envelope.rs
```diff
@@ -0,0 +1,213 @@
...
+    ) -> Result<Option<DecodedMessage>, Error> {
+        let Payload::Json(OwnedValue::Object(ref mut map)) = message.payload else {
+            return Ok(Some(message));
+        };
+
+        let Some(inner) = map.remove(self.field.as_str()) else {
+            warn!(
+                "unwrap_envelope: field '{}' not found in payload, passing through unchanged",
+                self.field
+            );
```
hubcio, line 40: this is a per-message `warn!` on a hot path. with the default `batch_length = 1000`, a single misconfigured connector floods stdout with 1000 identical warnings per poll cycle, masking real errors.

fix options: downgrade to `debug!`, rate-limit (warn-once per `(stream, topic, config)`), or count occurrences and emit a single summary per batch.

core/connectors/sdk/src/transforms/unwrap_envelope.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Configuration for the UnwrapEnvelope transform.
+///
+/// Extracts a nested JSON field from an envelope object and promotes it
+/// to the top-level payload. For example, given a Postgres source
+/// `DatabaseRecord` with shape `{ table_name, operation_type, timestamp,
+/// data: { ... }, old_data }`, setting `field = "data"` replaces the
+/// entire payload with the contents of `data`.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct UnwrapEnvelopeConfig {
+    pub field: String,
```
hubcio, line 31: `UnwrapEnvelopeConfig.field: String` has no validation. an empty string deserializes fine through `from_config` at `transforms/mod.rs:137-141`, then every message hits the missing-field branch and warn-spams at message rate (compounds the log flood on the json side).

reject empty `field` either via a custom `Deserialize` impl, or inline in `UnwrapEnvelope::new`, or in the `from_config` arm.

core/connectors/sdk/src/transforms/json/unwrap_envelope.rs
```diff
@@ -0,0 +1,213 @@
...
+
+impl UnwrapEnvelope {
+    pub(crate) fn transform_json(
+        &self,
+        _metadata: &TopicMetadata,
+        mut message: DecodedMessage,
+    ) -> Result<Option<DecodedMessage>, Error> {
+        let Payload::Json(OwnedValue::Object(ref mut map)) = message.payload else {
+            return Ok(Some(message));
+        };
```
hubcio, line 34: non-Object JSON (Array, scalar, null) silently passes through with no metric and no log. this is an operational blindspot - if a misconfigured source starts emitting arrays where the sink expects an unwrapped object, there's nothing in the logs that says the transform was a no-op.

at minimum a `debug!` with the actual payload variant, or a counter, so operators can see when the transform isn't doing anything.

core/connectors/sdk/src/transforms/mod.rs
```diff
@@ -131,5 +134,10 @@ pub fn from_config(
                 serde_json::from_value(raw.clone()).map_err(|_| Error::InvalidConfig)?;
             Ok(Arc::new(AvroConvert::new(cfg)))
         }
+        TransformType::UnwrapEnvelope => {
+            let cfg: UnwrapEnvelopeConfig =
+                serde_json::from_value(raw.clone()).map_err(|_| Error::InvalidConfig)?;
```
hubcio: `.map_err(|_| Error::InvalidConfig)` throws away the serde error detail. for a user with a typo in `field` or a wrong type, all they see is `InvalidConfig` - no line number, no field name, no expected-vs-got. this is a pre-existing pattern in every arm of `from_config`, but this PR adds another instance.

worth either fixing all arms in a follow-up, or at least propagating the serde error through `Error::InvalidConfigDetail(String)` or similar.

hubcio (changes requested):
review notes (not posted as line comments because targets are out-of-diff or already tracked):

- the FFI `(consume)` return code is still discarded at `core/connectors/runtime/src/sink.rs:642-650` (`ConsumeCallback` returns `i32`, `sdk/src/sink.rs:28-36`) and `AutoCommit::When(AutoCommitWhen::PollingMessages)` at `core/connectors/runtime/src/sink.rs:467` commits offsets before the sink runs. this PR widens the blast radius: any envelope-detected batch now `return Err(Error::InvalidRecord)` from the sink, gets logged, and the offsets have already advanced. tracked in #2927 and #2928 but this PR adds a brand new fail path triggerable purely by source/sink misconfiguration, so they should land together.

- `core/connectors/sinks/delta_sink/src/sink.rs:106` has the same envelope-vs-flat-JSON mismatch and no detection or unwrap guidance. fixing one sink and leaving the other broken means #3174 is still reachable through the postgres → delta path. either replicate the detection (with the tighter shape suggested below) or, better, document `unwrap_envelope` as the single fix and drop sink-side sniffing entirely.

- neither `core/connectors/sdk/README.md` nor `core/connectors/sinks/iceberg_sink/README.md` mention the new `unwrap_envelope` transform or the source-compatibility requirement for `postgres_source` → iceberg. discoverability is the whole point of this fix; please add at least a short entry to both.

- architecturally, sink-side envelope sniffing is the wrong layer. transform discipline + docs is the real fix. if sink-side detection stays, it should be opt-in via config, not a hardcoded postgres-shaped heuristic.

overall, i'm not fan of this PR.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: Removed entirely, relying on transform and docs now

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: Dropped sink-side detection completely

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: removed

core/connectors/sinks/iceberg_sink/src/router/mod.rs
atharvalade: fixed

core/connectors/sdk/src/transforms/json/unwrap_envelope.rs
atharvalade: Downgraded to `debug!`

core/connectors/sdk/src/transforms/unwrap_envelope.rs
atharvalade: `new()` now rejects empty field with `InvalidConfigValue`

core/connectors/sdk/src/transforms/json/unwrap_envelope.rs
atharvalade: Added `debug! `log for non-Object payload

core/connectors/sdk/src/transforms/mod.rs
atharvalade: Now propagates detail via `InvalidConfigValue(format!(...))`

atharvalade:
removed all sink-side sniffing and shifted to transform discipline + docs (SDK, Iceberg, and Postgres READMEs updated). The FFI return code and AutoCommit issues are pre-existing (#2927/#2928) and out of scope now that this PR no longer adds a new fail path through that code @hubcio

core/connectors/sinks/iceberg_sink/README.md
```diff
@@ -80,3 +80,26 @@ Example:
...
+stream = "postgres_stream"
+topic = "postgres_topic"
+schema = "json"
+
+[transforms.unwrap_envelope]
+enabled = true
+field = "data"
+```
+
+Alternatively, configure the Postgres source with `payload_column` and `payload_format = "json_direct"` to bypass the envelope entirely — see the Postgres source README for details.
```
hubcio: come on, don't leave postgres reference in iceberg readme. go thru all changes in this PR and make sure that all docs (readmes, comments, etc) are aligned with the actual code and placement in the project.

hubcio (changes requested):

core/connectors/sinks/iceberg_sink/README.md
atharvalade: you're right. please check now, made sure everything is good.

mmodzelewski (approved):

hubcio (approved):

PR #3199, opened 2026-04-29
fix(connectors): unify extract_column_value into single free function

### LINKED ISSUE #3172: Two diverging copies of extract_column_value
The Postgres source connector has two nearly identical functions for extracting column values from a `PgRow`:

- `extract_column_value` (method on `PostgresSource`, line ~1314) used by the sequential path
- `extract_column_value_static` (free function, line ~1903) used by the parallel/chunked path

These handle different sets of Postgres types. The static version handles `NUMERIC`, `DATE`, and `BPCHAR`. The method version does not. Any bug fix or new type added to one will silently not apply to the other.

During the benchmark, fixing `NUMERIC` and `DATE` handling in the static version left the method version still broken for those types.

**Fix**: Delete one, keep a single free function, call it from both code paths.

atharvalade:

Closes #3172

## Rationale

Bug fixes or new type support added to one copy of `extract_column_value` silently won't apply to the other, causing divergence between sequential and parallel code paths.

## What changed?

The Postgres source connector had `extract_column_value` as a method on `PostgresSource` (`&self`), making it impossible to call from static/parallel contexts without duplicating the function. Any future parallel/chunked path would need its own copy.

Converted it to a single free function callable from any context. Added missing `DATE` (via `chrono::NaiveDate`) and `BPCHAR` type handling that the parallel copy on the benchmark branch had but this version lacked.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Opus 4.6
2. exploration and implementation guidance
3. Verified via cargo check, clippy -D warnings, fmt --check, all 19 unit tests passing, CI lint scripts passing


lukaszzborek:
@atharvalade 
In issue is `Postgres source connector has two nearly identical functions` and to remove one. But i don't see second one.
I'm jus asking for clarification, if i don't see something 

Also remeber about #3196 where you also add Date

atharvalade:
> @atharvalade In issue is `Postgres source connector has two nearly identical functions` and to remove one. But i don't see second one. I'm jus asking for clarification, if i don't see something
> 
> Also remeber about #3196 where you also add Date

Good catch, the second copy (`extract_column_value_static`) lives on the `feat/pg-to-iceberg-benchmark` branch (not yet merged to master), where the parallel/chunked polling paths were added. This PR preemptively converts the method to a free function so that when those paths land, they can call the same function instead of duplicating it. I'll drop the DATE addition here since #3196 already covers it, and rebase on latest master.

Cargo.lock
hubcio, line 1: no reason to include these changes in this PR.

hubcio (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -933,124 +933,123 @@ impl PostgresSource {
...
+        "NUMERIC" => {
+            let value: Option<String> = row
+                .try_get(column_index)
+                .map_err(|_| Error::InvalidRecord)?;
+            Ok(value
+                .and_then(|s| s.parse::<f64>().ok())
+                .map(serde_json::Value::from)
+                .unwrap_or(serde_json::Value::Null))
+        }
+        "VARCHAR" | "TEXT" | "CHAR" | "BPCHAR" => {
```
hubcio, line 1003: BPCHAR is the only new behavior in this PR but nothing exercises it. codecov flags 0% patch coverage (101 missing lines). a single round-trip test reading a `char(n)` column closes the gap and prevents the same divergence the issue is trying to prevent.

if possible, just make it part of some existing tests instead of making a new tc.

hubcio (changes requested):

core/connectors/sources/postgres_source/src/lib.rs
atharvalade: Added a CHAR(10) column to the existing `PostgresSourceJsonFixture` and verified the BPCHAR value in the raw JSON output of `json_rows_source_produces_messages_to_iggy`

numinnex (approved):

hubcio (approved):

PR #3200, opened 2026-04-29
fix(connectors): log underlying errors instead of discarding them in SDK poll/consume paths

### LINKED ISSUE #3171: Connector SDK silently swallows poll errors
The `handle_messages` function in `core/connectors/sdk/src/source.rs` discards the actual error returned by `source.poll()`:

```rust
let Ok(messages) = messages else {
    error!("Failed to poll messages for source connector with ID: {plugin_id}");
    continue;
};
```

The underlying error is never logged. Any connector failure produces an infinite loop of "Failed to poll messages" with zero context about the root cause. During the Postgres-to-Iceberg benchmark, this cost significant debugging time because the real error (`InvalidRecord` from a NUMERIC column type mismatch) was invisible until a temporary log line was added inside the connector's own `poll()` method.

**Fix**: Log the error: `error!("Failed to poll messages for source connector with ID: {plugin_id}: {err:?}");`

The same pattern should be checked in the sink SDK path.

atharvalade:

Closes #3171

## Rationale

Silent error swallowing in connector SDK made debugging connector failures nearly impossible — only generic "Failed to poll/consume" messages appeared with zero root-cause context.

## What changed?

The source SDK's `handle_messages` loop discarded the `Err` variant from `source.poll()` and `postcard::to_allocvec`, logging only a static string. The sink SDK had the same pattern across all `postcard::from_bytes` deserialization calls, `try_into_payload`, and `sink.consume()` (which wasn't logged at all).

All `let Ok(x) = expr else { ... }` patterns are replaced with `match` to bind and format the underlying error into the existing log lines. No behavioral change — same control flow (continue/return codes) — just makes the root cause visible.

## Local Execution

- Passed / passed
- Pre-commit hooks ran / ran (`cargo fmt --check`, `cargo clippy`, `trailing-whitespace`, `trailing-newline`, `license-headers` all green)

## AI Usage

1. Opus 4.6
2. used to locate the pattern and write the match arms
3. Verified via `cargo clippy`, `cargo fmt --check`, and full CI script suite locally


slbotbm (approved):
lgtm

spetz (approved):

mmodzelewski (approved):

PR #3201, opened 2026-04-30
refactor(ci): rely on pre-commit file arguments

### LINKED ISSUE #3195: Align pre-commit hooks to use passed filenames and remove internal git diff logic
### Bug description

Some pre-commit hooks (`trailing-whitespace.sh`, `trailing-newline.sh`, `taplo.sh`, `binary-artifacts.sh`) currently determine the staged file scope internally using `git diff --cached`, while also setting `pass_filenames: false`.

This differs from other hooks (e.g., markdownlint, shellcheck) that rely on filenames passed directly by pre-commit. Aligning these hooks to the same pattern would simplify the implementation and improve consistency.

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

Standing-Man:

Closes #3195.

## Rationale

Some pre-commit hook scripts still accepted `--staged` and used `git diff --cached` to discover staged files internally.

That duplicated pre-commit's own file selection behavior. Since these hooks now use `pass_filenames`, pre-commit already passes the matched staged files directly to the scripts. Keeping a second staged-file discovery path made the behavior harder to reason about and easier to drift from pre-commit's filtering rules.

## What changed?

- Removed `--staged` handling from file-scoped hook scripts.
- Removed `git diff --cached` based staged-file discovery.
- Kept explicit file arguments as the primary input path for pre-commit.
- Preserved `--ci` behavior for CI / PR-diff based checks.
- Updated `--help` output to document bare local runs with:

**Note**: Running these scripts with --ci outside a CI environment, where neither
`GITHUB_BASE_REF` nor CI is set, now falls back to checking all tracked
files instead of staged files.

Running the scripts with no file arguments also defaults to checking all
tracked files. This is intentional because pre-commit no longer relies on
script-side staged file discovery; it passes matched files directly via
pass_filenames.

## Local Execution

- Passed / not passed
Passed
Ran

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT) Codex
Drafted changes and updated the pr-commit scripts based on human direction.
4. How did you verify the generated code works correctly?
🚧: Not tested locally yet.
5. Can you explain every line of the code if asked?
Yes. I reviewed the changes


scripts/ci/taplo.sh
```diff
@@ -101,8 +92,13 @@ get_files() {
         # Generic CI - compare with HEAD~1
         git diff --name-only --diff-filter=ACM HEAD~1 -- '*.toml'
       else
-        # Fallback to staged files
-        git diff --cached --name-only --diff-filter=ACM -- '*.toml'
+        # Local fallback: check all TOML files
+        find . -name '*.toml' \
```
hubcio: the local fallback here uses `find .` while the sibling scripts (`binary-artifacts.sh`, `trailing-newline.sh`, `trailing-whitespace.sh`) use `git ls-files` for the same fallback. inconsistent. could align to `git ls-files -- '*.toml'` and also collapse the duplication with the `all)` branch below since they're now functionally identical.

scripts/ci/markdownlint.sh
```diff
@@ -34,19 +33,13 @@ while [[ $# -gt 0 ]]; do
       MODE="fix"
       shift
       ;;
-    --staged)
-      FILE_MODE="staged"
-      shift
-      ;;
     --all)
-      FILE_MODE="all"
       shift
```
hubcio: `--all` is now a no-op (just `shift`, no `FILE_MODE` var anymore). help text on line 43 still says `(default)`. flag is harmless but redundant - either drop it or document it as accepted-for-compat.

scripts/ci/shellcheck.sh
```diff
@@ -36,19 +35,13 @@ while [[ $# -gt 0 ]]; do
       MODE="fix"
       shift
       ;;
-    --staged)
-      FILE_MODE="staged"
-      shift
-      ;;
     --all)
-      FILE_MODE="all"
       shift
```
hubcio: same as `markdownlint.sh` - `--all` is now a no-op after `FILE_MODE` removal. help text still claims it's the default. either drop the flag or keep it as a documented compat shim.

hubcio (changes requested):
behavior change worth noting in the commit body: running scripts with --ci outside CI (no GITHUB_BASE_REF, no CI env) now scans the whole repo instead of staged files. not a bug since pre-commit no longer relies on it, but a one-line note in the commit message would help anyone running the scripts manually.

scripts/ci/taplo.sh
Standing-Man: I’ve replaced find with `git ls-files` in the latest commit to collect the files to be checked.

hubcio:
if PR is ready to be reviewed again, please type comment with /ready somewhere in the beginning of any line. also comment/resolve the issues.

/author

scripts/ci/binary-artifacts.sh
```diff
@@ -30,38 +30,28 @@ set -euo pipefail
...
 #
 # Runs in two contexts:
-#   pre-commit hook  -- checks staged files   (--check --staged)
+#   pre-commit hook  -- checks filenames passed by pre-commit (--check files...)
 #   CI workflow      -- checks PR diff files  (--check --ci)
 #
 # Exit codes: 0 = clean, 1 = binary artifacts found or error.
 
-FILE_MODE="staged"
+FILE_MODE="all"
```
hubcio, line 38: default `FILE_MODE` flipped from `staged` to `all`. running the script bare locally (no flag, no files) now scans every tracked file via `file(1)` instead of just `git diff --cached`. fine for pre-commit since it always passes files, but the manual-run behavior change isn't called out in the PR body, and `--help` no longer explains what happens with no args (previously: `(default, for git hooks)`).

same pattern at `taplo.sh:23`, `trailing-newline.sh:23`, `trailing-whitespace.sh:23`. consider adding a `[none]   Check all tracked files (default)` line to each `--help` block.

hubcio (changes requested):
nit outside the diff hunks: scripts/ci/shellcheck.sh:99-100 and 131-132 still say "Use staged or explicitly provided files when present". --staged was removed in this PR, so the "staged or" wording is misleading. drop it.

also, my prior comment about the --ci-outside-CI behavior change still applies - running scripts with --ci when neither GITHUB_BASE_REF nor CI is set now scans the whole repo via git ls-files instead of staged files. a one-line note in the commit body would help anyone running manually.

scripts/ci/binary-artifacts.sh
Standing-Man: Thanks for the feedback. I’ve removed the outdated references to staged files, documented the `--ci` behavior change in the `“What changed?”` section, and updated the related shell scripts with additional `--help` information describing the default full-repo scan behavior.

hubcio (approved):

mmodzelewski (approved):

PR #3204, opened 2026-05-01
fix(common): errors with clippy pedantic (manual_assert)

### LINKED ISSUE #3205: Fix some clippy warnings
Replace `if-then-panic!` patterns with `assert!` macros as suggested by clippy pedantic mode.

Reproduce
```sh
cargo clippy -- -W clippy::manual_assert
```

Proposed fix by clippy

```rs
// Before
if !byte.is_ascii_digit() {
    panic!("Invalid digit in version number");
}

// After
assert!(byte.is_ascii_digit(), "Invalid digit in version number");
```

chammika:

This PR fixes some issues raised by clippy pedantic mode
```sh
cargo clippy -- -W clippy::manual_assert
```

Closes #3205

## Rationale

This was suggested by Clippy as better practice

https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#manual_assert

## What changed?

Fix all `clippy::manual_assert` cases indicated by clippy

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

None


core/common/src/utils/versioning.rs
```diff
@@ -36,23 +36,17 @@ pub struct SemanticVersion {
...
-    }
+    assert!(start < end, "Cannot parse empty range as u32");
 
     let mut result = 0u32;
     let mut i = start;
 
-    if bytes.is_empty() {
-        panic!("Can not parse empty string as u32");
-    }
+    assert!(!bytes.is_empty(), "End index out of bounds");
```
hubcio: the message `"End index out of bounds"` doesn't describe the `!bytes.is_empty()` check - it describes a different condition (matching the assertion at line 96 about `end <= bytes.len()`). this is a diagnostic regression: original message was `"Can not parse empty string as u32"`, which actually matched the check. restore the original message.

core/server/src/bootstrap.rs
```diff
@@ -129,11 +129,10 @@ pub async fn create_directories(config: &SystemConfig) -> Result<(), IggyError>
...
     let mut password = env::var(IGGY_ROOT_PASSWORD_ENV);
-    if (username.is_ok() && password.is_err()) || (username.is_err() && password.is_ok()) {
-        panic!(
-            "When providing the custom root user credentials, both username and password must be set."
-        );
-    }
+    assert!(
+        !((username.is_ok() && password.is_err()) || (username.is_err() && password.is_ok())),
+        "When providing the custom root user credentials, both username and password must be set."
+    );
```
hubcio: double-negated XOR is hard to parse. equivalent and clearer:

```rust
assert_eq!(
    username.is_ok(),
    password.is_ok(),
    "When providing the custom root user credentials, both username and password must be set."
);
```

core/server/src/http/http_server.rs
```diff
@@ -310,9 +310,10 @@ async fn build_app_state(
         Ok(manager) => manager,
         Err(error) => panic!("Failed to initialize JWT manager: {error}"),
     };
-    if jwt_manager.load_revoked_tokens().await.is_err() {
-        panic!("Failed to load revoked access tokens");
-    }
+    assert!(
+        jwt_manager.load_revoked_tokens().await.is_ok(),
+        "Failed to load revoked access tokens"
+    );
```
hubcio: `assert!(... .is_ok(), "...")` drops the underlying error from `load_revoked_tokens()`. the original `if .is_err() { panic!(...) }` had the same issue, so this isn't a regression from this PR, but worth fixing while you're in here:

```rust
if let Err(error) = jwt_manager.load_revoked_tokens().await {
    panic!("Failed to load revoked access tokens: {error}");
}
```

hubcio (changes requested):

Standing-Man (commented):
Replacing `panic` with `assert` in many places is not appropriate. It would be better to return proper error types with clear error messages, rather than using `assert` to validate configuration. cc @hubcio WDYT?

hubcio:
@Standing-Man please list concrete examples

hubcio (approved):

mmodzelewski (approved):

Issue #20, opened 2023-09-10
Add channel notifications for the polled messages in IggyClient


spetz:

spetz:
Completed with #32 - when using the `IggyClient` [flume](https://crates.io/crates/flume) channel receiver can be created by invoking `subscribe_to_polled_messages()` method. Polling the messages in the background by using `start_polling_messages()` method will result in **one** of the following actions:
- Send a message via created channel
- Invoke provided `on_message` closure
- Invoke provided `MessageHandler` trait implementation

PR #3215, opened 2026-05-05
feat(connectors): add Apache Doris sink connector

### LINKED ISSUE #3112: Add Apache Doris connector
### Description

add Apache doris connector

### Component

None

### Proposed solution

_No response_

### Alternatives considered

_No response_

ryankert01:

Closes #3112

## Rationale

Adds an Apache Doris sink so Iggy streams can be written into Doris for analytical querying.

## What changed?

Iggy had no path to land messages in Apache Doris. A new `iggy_connector_doris_sink` crate consumes JSON payloads and writes them via Doris's HTTP Stream Load API (`PUT /api/{db}/{table}/_stream_load`).

The non-obvious bits the connector handles: re-attaching `Authorization` across the FE→BE 307 redirect (which `reqwest` strips by default), parsing the JSON `Status` body to classify success / `Label Already Exists` / transient (`Publish Timeout`, 5xx) / permanent (`Fail`, 4xx, unknown), and emitting a deterministic per-batch label so replays are deduplicated by Doris's label-keep window. v1 is sink-only, JSON-only, HTTP Basic auth only, and assumes pre-created tables — no DDL.

## Local Execution

- Passed
- Pre-commit hooks ran. Pre-push C#/Java hooks skipped (no dotnet/JDK locally; contribution is Rust-only).

## AI Usage

1. Claude Code (Anthropic).
2. Crate scaffolding against `quickwit_sink` / `influxdb_sink`, testcontainer fixture, and iteration on the Stream Load redirect + `Status`-body classification.
3. 14 unit tests + 6 integration tests against a real `apache/doris:doris-all-in-one-2.1.0` container, covering happy path, 1k-row bulk, `max_filter_ratio`, label-replay dedupe, missing-target-table, and `columns` derived expressions; row state verified via the MySQL frontend. help write docs.
4. Yes.

## Some Image!
Screenshot 2026-05-26 at 12 37 08 PM

Screenshot 2026-05-26 at 12 33 21 PM


ryankert01:
There's a performance optimization(may or may not works) that I want to leave as a follow-up PR:
**multi-output-format support**: currently use json, but json seems to be slow in Doris. So I can test out csv & parquet.

hubcio:
we'll check this in upcoming 2-3 days.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+        let total = messages.len();
+        debug!(
+            "Doris sink ID {} received {total} messages for {}.{}",
+            self.id, self.config.database, self.config.table
+        );
+
+        let batch_size = self.batch_size();
+        let mut worst_error: Option<Error> = None;
+
+        for chunk in messages.chunks(batch_size) {
```
hubcio, line 767: if every message in the batch is non-JSON, `json_values` is empty for every chunk, the `continue` at line 488 skips them all, `worst_error` stays `None`, and `consume()` returns `Ok(())`. the runtime treats that as success, the iggy consumer auto-commit advances the offset, and the batch is gone with no signal. fix: track delivered count and return `Error::InvalidPayloadType` when drops > 0 and writes == 0, or simpler -- return the error on the first non-JSON drop.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+const CONNECT_TIMEOUT_SECS: u64 = 5;
+// Doris's FE Stream Load returns 307 to a BE; reqwest's default policy strips
+// the Authorization header on cross-host redirects, so we follow redirects
+// manually. This cap protects against a misbehaving cluster looping.
+const MAX_REDIRECTS: u8 = 5;
+// Per Doris docs, Stream Load labels must be 1..=128 chars matching
+// `[A-Za-z0-9_-]`. These caps keep the worst-case label well under that limit.
+const MAX_LABEL_PREFIX_LEN: usize = 16;
+const MAX_LABEL_NAME_LEN: usize = 16;
+const LABEL_HASH_HEX_LEN: usize = 8;
```
hubcio: `LABEL_HASH_HEX_LEN = 8` truncates blake3 to 32 bits. inside the sanitize-collision class (e.g. `events.v1` and `events_v1`), birthday bound is ~65k. an adversarial multi-tenant namer trivially forces a label collision -- and Doris's server-side label dedupe treats it as `Label Already Exists` success (lines 376-384), silently dropping one tenant's batch. bump to 16 hex chars (64-bit, birthday ~5B). naive bump alone overflows the 128-char Doris label cap though (P=16 + 2*N=32 + 2*H=32 + 5 hyphens + 2 underscores + 10 + 20 + 20 = 137 chars), so `MAX_LABEL_NAME_LEN` needs to drop from 16 to 8 at the same time. with P=16, N=8, H=16 the worst case is 121 chars, under cap. the existing `label_stays_under_doris_128_char_cap` test at lines 657-674 will catch regressions.

core/connectors/sinks/doris_sink/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+version = "0.1.0"
+description = "Iggy is the persistent message streaming platform written in Rust, supporting QUIC, TCP and HTTP transport protocols, capable of processing millions of messages per second."
+edition = "2024"
+license = "Apache-2.0"
+keywords = ["iggy", "messaging", "streaming"]
+categories = ["command-line-utilities", "database", "network-programming"]
+homepage = "https://iggy.apache.org"
+documentation = "https://iggy.apache.org/docs"
+repository = "https://github.com/apache/iggy"
+readme = "../../README.md"
```
hubcio, line 29: missing `publish = false`

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+                Err(e) => {
+                    record_error(
+                        &mut worst_error,
+                        Error::CannotStoreData(format!("Failed to serialize batch for Doris: {e}")),
+                    );
+                    continue;
+                }
+            };
+
+            let first_offset = chunk.first().map(|m| m.offset).unwrap_or(0);
```
hubcio: `chunk.first().map(|m| m.offset).unwrap_or(0)` and `chunk.last()...unwrap_or(0)` are unreachable today -- `chunks(batch_size)` with `batch_size >= 1` (guaranteed by `batch_size().max(1)` at lines 137-139) over a non-empty `messages` (guarded at line 459) never yields an empty chunk. but if any future refactor breaks that invariant, the fabricated `offset = 0` collides with the genuine offset-0 batch label and Doris silently dedupes the wrong batch. replace with `chunk.first().expect("non-empty chunk").offset` -- panic loud, do not silently corrupt.

core/connectors/sinks/doris_sink/README.md
```diff
@@ -0,0 +1,89 @@
...
+- Messages must arrive with `Payload::Json` (i.e. the configured stream schema is `json`). Other payload types are dropped with a `warn!` and the consumer offset advances past them — this is silent data loss for any non-JSON message, so the upstream schema must be guaranteed JSON.
+- The Iggy message JSON shape must match the target table columns. Use the optional `columns` plugin setting if the column order differs from the JSON keys.
+
+## How it works
+
+1. For each batch of messages, the connector serializes the JSON payloads into a JSON array.
+2. It computes a deterministic Stream Load `label` of the form `{label_prefix}-{stream}_{hash8}-{topic}_{hash8}-{partition}-{first_offset}-{last_offset}`.
+   - Each variable-length segment carries a 32-bit blake3 hash of the raw name, so names that sanitize to the same string (e.g. `events.v1` vs `events_v1`) cannot collide.
+   - The total label is bounded under Doris's 128-char cap regardless of input length.
+   - Doris dedupes loads by label inside its `label_keep_max_second` window, so a replayed batch (after restart, retry, etc.) is silently absorbed instead of producing duplicates.
```
hubcio: this and the `label_keep_max_second` guidance at line 81 are misleading. the runtime ignores the sink's `consume()` return value and the iggy consumer commits offset at poll time, so after a crash the next batch picks up past the failed one -- at-most-once, not idempotent replay. the only `replay` that benefits from server-side label dedupe is in-process retry within a single `send_stream_load` call, which this connector does not loop. either fix the runtime to honor the rc and disable poll-time auto-commit for the sink path, or reword this section to `Idempotent retry within a single consume() invocation` and add a Limitations bullet on at-most-once-under-crash.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+        ))
+    })
+}
+
+/// True for errors the runtime must NOT retry (schema mismatch, bad config,
+/// permanent HTTP error, etc.). Used by `consume` to pick the most-severe
+/// error across chunks: a transient error from chunk N must NOT shadow a
+/// permanent error from chunk M, or the runtime would retry forever instead
+/// of routing the batch to its DLQ path.
+fn is_permanent_error(e: &Error) -> bool {
```
hubcio: `is_permanent_error` excludes `HttpRequestFailed`, but the Doris sink raises that variant at three sites with mixed semantics: line 197 (TCP/DNS/timeout -- transient), line 207 (max-redirect-loop -- permanent), line 217 (no `Location` header on 307 -- permanent). blanket-promoting `HttpRequestFailed` to permanent would misclassify real transient outages. split the variant: introduce `Error::HttpProtocolError` for the permanent cases (redirect loop, missing Location) and keep `HttpRequestFailed` for transient. then add `HttpProtocolError` to `is_permanent_error`. moot today because the runtime drops the rc anyway, but lands the wrong semantics if that ever gets fixed.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+                StatusCode::TEMPORARY_REDIRECT | StatusCode::PERMANENT_REDIRECT
+            ) {
+                redirects += 1;
+                if redirects > MAX_REDIRECTS {
+                    return Err(Error::HttpRequestFailed(format!(
+                        "Doris sink ID {} exceeded max redirects ({MAX_REDIRECTS})",
+                        self.id
+                    )));
+                }
+                let Some(location) = response
```
hubcio, line 236: the redirect loop follows `Location` blindly and re-attaches `Authorization` (Basic). `open()` validates only the initial `fe_url` scheme/host at lines 431-441. attack: a compromised or MITM'd FE returns `Location: http://attacker/`. the connector posts the payload + Basic creds to the attacker within `MAX_REDIRECTS = 5`. the trust model is acknowledged in the comment at lines 425-430 but not enforced. before re-attaching `Authorization`, validate that the `Location` scheme is not weaker than the initial scheme (refuse https → http) and the host matches the initial host. if cross-host is required for legitimate FE → BE topology, gate it on an opt-in config flag.

core/integration/tests/connectors/doris/doris_sink.rs
```diff
@@ -0,0 +1,489 @@
...
+        .await
+        .expect("rows");
+    assert_eq!(count, 2, "only the two valid rows should land");
+}
+
+#[iggy_harness(
+    server(connectors_runtime(config_path = "tests/connectors/doris/sink.toml")),
+    seed = seeds::connector_stream
+)]
+async fn given_replayed_label_should_dedupe(harness: &TestHarness, fixture: DorisSinkFixture) {
```
hubcio, line 190: this test proves Doris's server-side label dedupe by issuing a manual reqwest PUT with the same label. it does NOT kill the connector runtime mid-batch and restart, so the README claim `replayed batch (after restart, retry, etc.) is silently absorbed` is not actually under test. add a kill-runtime-mid-batch + restart variant that asserts no duplicate rows.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+/// `{prefix}-{stream_san}_{hash8}-{topic_san}_{hash8}-{partition}-{first}-{last}`.
+///
+/// Exposed as a `pub` free function so integration tests can produce the
+/// exact same label without instantiating a `DorisSink`. The format is
+/// engineered to:
+///   1. Bound the total length under Doris's 128-char label cap regardless
+///      of input name length.
+///   2. Eliminate the silent-collision risk between names that sanitize to
+///      the same string (the hash is over the raw name).
+pub fn build_label(
```
hubcio, line 543: this is the only `pub fn` exported from any sink crate's `lib.rs` (verified across all siblings). the path-dep at `core/integration/Cargo.toml:52` exists solely so integration tests can reuse the label format. that creates a stable public-API contract for a test convenience. options: (a) `#[doc(hidden)] pub` to signal it's not part of the API surface; (b) drop the path-dep and inline the format string in the test fixture, relying on the existing unit tests at lines 599-674 to catch format drift. (b) is cleanest -- the unit tests already cover the contract.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+        let client = self.client.as_ref().ok_or_else(|| {
+            Error::InitError(format!(
+                "Doris sink ID {} called before open() — client not initialized",
+                self.id
+            ))
+        })?;
+        let mut url = self.stream_load_url();
+        let mut redirects = 0u8;
+
+        loop {
```
hubcio, line 195: two things on this redirect loop. (1) the post-redirect request rebuilds every header (`Authorization`, `Expect`, `format`, `strip_outer_array`, `label`, and the conditional `max_filter_ratio` / `columns` / `where`) -- behavior is correct, but no test isolates the invariant. a wiremock test that intercepts the post-redirect PUT and asserts on the expected headers would be cheap insurance against regression. (2) no per-redirect backoff or jitter. `MAX_REDIRECTS = 5` cap fires only on cluster pathology, so this is cold-path and low-priority, but a 10-100ms jitter would avoid burning CPU on a misbehaving FE.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+                        "Doris sink ID {} got {status} with no Location header",
+                        self.id
+                    )));
+                };
+                debug!("Doris sink ID {} following redirect to {location}", self.id);
+                url = location.to_string();
+                continue;
+            }
+
+            let response_text = response.text().await.unwrap_or_default();
```
hubcio: `response.text().await.unwrap_or_default()` on a body-read failure (mid-stream TCP reset, etc.) yields an empty string, which then parses as `PermanentHttpError`. the classification is conservative-correct (errs toward DLQ), but the original `reqwest::Error` diagnostic is silently dropped, making cluster-side issues harder to triage. log the body-read error before discarding -- don't change the classification.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+    auth_header: SecretString,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct DorisSinkConfig {
+    pub fe_url: String,
+    pub database: String,
+    pub table: String,
+    pub username: String,
+    pub password: SecretString,
```
hubcio, line 97: `password: SecretString` is required (no `Option`), but `SecretString::from("")` is infallible. the default `config.toml` ships `password = ""` and the integration fixture uses empty too. add an `open()` warning when password is empty AND scheme is `http://` AND host is non-loopback -- same conditions the existing cleartext-http warning already gates on.

core/integration/tests/connectors/fixtures/doris/container.rs
```diff
@@ -0,0 +1,611 @@
...
+use tokio::time::sleep;
+use tracing::info;
+use uuid::Uuid;
+
+// Global serialization lock for Doris fixtures. The all-in-one image binds
+// the BE HTTP port 1:1 (host:8040 → container:8040) so the FE→BE 307
+// redirect remains reachable; only one container per test process can hold
+// that port at a time. Each fixture holds the guard for its lifetime and
+// drops it on Drop, releasing the slot for the next test.
+static DORIS_LOCK: Lazy<std::sync::Arc<Mutex<()>>> =
```
hubcio: the `tokio::sync::Mutex` global lock here plus `threads-required = "num-cpus"` for `connectors::doris::` tests in `.config/nextest.toml:31-34` plus the 1:1 `host:8040 → container:8040` bind serialize all 6 Doris tests through a single slot. 6 tests x cold container boot (~60-90s) plus ~30s body each ≈ 5-10 minute serial CI wall time. follow-up rather than blocker: amortize one container across tests with per-test databases. also, `tokio::sync::Mutex` does NOT poison on panic (unlike `std::sync::Mutex`) -- a mid-setup panic with a leaked container will fail the next test loudly at the port bind, not hang, but switching to `std::sync::Mutex` would force explicit reset and make the failure mode clearer.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+                        error!("Doris sink ID {} batch failed: {e}", self.id);
+                        record_error(&mut worst_error, e);
+                    }
+                },
+                Err(e) => {
+                    error!("Doris sink ID {} HTTP failed: {e}", self.id);
+                    record_error(&mut worst_error, e);
+                }
+            }
+        }
```
hubcio, line 846: mid-batch partial-commit pathology: chunk N succeeds (Doris commits server-side), chunk N+1 fails, the loop continues, `record_error` records, and `consume()` returns `Err` at the end. today the runtime ignores the `Err` so chunk N+1's data is lost. once the runtime is fixed to honor the rc, replay re-runs the whole batch including chunk N -- server-side label dedupe at lines 376-384 saves it from doubling rows, but only inside `label_keep_max_second`. to make retry semantics deterministic regardless: `break` on first chunk failure, and document `label_keep_max_second > runtime retry budget` as a load-bearing invariant.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+            last_offset,
+        )
+    }
+
+    async fn send_stream_load(
+        &self,
+        label: &str,
+        body: Bytes,
+    ) -> Result<StreamLoadResponse, Error> {
+        let client = self.client.as_ref().ok_or_else(|| {
```
hubcio: dead-defensive check. the SDK lifecycle guarantees `open()` runs before `consume()`, and the runtime guard at `core/connectors/runtime/src/sink.rs:125-127` skips spawn on `init_error`, so this path is unreachable. replace with `debug_assert!(self.client.is_some())` or drop it entirely.

core/integration/tests/connectors/fixtures/doris/container.rs
```diff
@@ -0,0 +1,611 @@
...
+const DEFAULT_BE_REGISTRATION_ATTEMPTS: usize = 180;
+const DEFAULT_BE_REGISTRATION_INTERVAL_MS: u64 = 2000;
+
+// 90 seconds (180 × 500ms) — bulk tests with 1000 rows can take ~30 s on a
+// fresh container, and a Doris instance under memory pressure (e.g. when it
+// is the 3rd or 4th container to spin up serially) takes longer.
+pub const DEFAULT_POLL_ATTEMPTS: usize = 180;
+pub const DEFAULT_POLL_INTERVAL_MS: u64 = 500;
+
+pub const ENV_SINK_FE_URL: &str = "IGGY_CONNECTORS_SINK_DORIS_PLUGIN_CONFIG_FE_URL";
```
hubcio: these `ENV_SINK_*` constants are `pub` but referenced only inside this file (specifically `build_connector_envs` at lines 442-470). drop the `pub`.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+
+fn validate_identifier(name: &str, field: &str, id: u32) -> Result<(), Error> {
+    if name.is_empty() {
+        return Err(Error::InvalidConfigValue(format!(
+            "Doris sink ID {id}: {field} must not be empty"
+        )));
+    }
+    if !name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_') {
+        return Err(Error::InvalidConfigValue(format!(
+            "Doris sink ID {id}: {field} '{name}' must match [A-Za-z0-9_]+ (Doris identifier rules)"
```
hubcio: the error message attributes `[A-Za-z0-9_]+` to `Doris identifier rules` but Doris's actual identifier rules are broader (backtick-quoted identifiers etc.). `[A-Za-z0-9_]+` is iggy's stricter subset to block path traversal in the constructed URL. suggested wording: `must match iggy's stricter subset of Doris identifiers (path-traversal guard)`.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+use tracing::{debug, error, info, warn};
+
+sink_connector!(DorisSink);
+
+const DEFAULT_LABEL_PREFIX: &str = "iggy";
+const DEFAULT_BATCH_SIZE: u32 = 1000;
+const DEFAULT_TIMEOUT_SECS: u64 = 30;
+// A bounded TCP handshake timeout so an unreachable FE fails fast instead of
+// eating the entire request-timeout budget on connect alone.
+const CONNECT_TIMEOUT_SECS: u64 = 5;
```
hubcio: `CONNECT_TIMEOUT_SECS = 5` is hardcoded while `timeout_secs` is configurable. cross-region operators with cold-start FE will trip the 5s connect under their otherwise-30s budget. either expose a `connect_timeout_secs` config field or document the 5s hardcode in the README.

core/connectors/sinks/doris_sink/README.md
```diff
@@ -0,0 +1,89 @@
...
+- **`columns` and `where` are SQL-expression pass-throughs.** Whatever you put in those config fields is forwarded verbatim to Doris's Stream Load and evaluated as a SQL expression. Keep this config trusted.
+
+## Operational guidance
+
+- **`label_keep_max_second`.** Idempotent replay relies on Doris retaining each label for at least as long as it could take the Iggy runtime to redrive a failed batch. The Doris default is 3 days, which is conservative. If you set this lower on the Doris side, make sure your runtime retry budget fits inside the window — once a label expires, a replay re-loads instead of deduping, producing duplicate rows.
+- **Filtered-row alerts.** When Doris reports `number_filtered_rows > 0`, the connector emits a `warn!`. This is your signal that upstream message shapes have drifted from the table schema; alert on it.
+
+## Limitations (todo)
```
hubcio: no sibling sink README uses the `(todo)` suffix on this heading. `http_sink/README.md` uses plain `### Limitations`. rename to `## Limitations` or `## Roadmap`.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+            let mut request = client
+                .request(Method::PUT, &url)
+                .header(header::AUTHORIZATION, self.auth_header.expose_secret())
+                .header(header::EXPECT, "100-continue")
+                .header("format", "json")
+                .header("strip_outer_array", "true")
+                .header("label", label)
+                .body(body.clone());
+
+            if let Some(ratio) = self.config.max_filter_ratio {
```
hubcio: `ratio.to_string()` is rebuilt per batch and per redirect, even though `max_filter_ratio` is config-static (never mutated post-open). same pattern applies to the `columns` and `where_clause` headers at lines 188-193, and to the `Authorization` `HeaderValue` reparsed every PUT at line 178. precompute these once in `open()` as `Option<HeaderValue>` (with `set_sensitive(true)` for `Authorization`). sub-microsecond saves but cheap to fix.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+        // 307s by design.
+        reqwest::Client::builder()
+            .redirect(reqwest::redirect::Policy::none())
+            .timeout(timeout)
+            .connect_timeout(Duration::from_secs(CONNECT_TIMEOUT_SECS))
+            .build()
+            .map_err(|e| Error::InitError(format!("Failed to build Doris HTTP client: {e}")))
+    }
+
+    fn stream_load_url(&self) -> String {
```
hubcio: `stream_load_url()` rebuilds the URL via `format!` on every `send_stream_load` call (line 172). `fe_url`, `database`, `table` are config-static and validated in `open()`. cache the formatted URL string in `open()` after `validate_identifier`.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+                topic_metadata,
+                messages_metadata.partition_id,
+                first_offset,
+                last_offset,
+            );
+
+            match self.send_stream_load(&label, body).await {
+                Ok(response) => match classify_status(&response) {
+                    Ok(()) => {
+                        if response.number_filtered_rows > 0 {
```
hubcio, line 814: `number_filtered_rows > 0` emits `warn!` only. the rows are silently absent from the table -- a semantic data-loss vector. the path is only reachable when the operator opts into `max_filter_ratio > 0` (default Doris would return `Fail` for filtered rows), so warn-only is correct for the opt-in case, but a `filtered_rows_total` counter metric per `{db, table}` would let operators alert on schema drift instead of grepping logs.

hubcio (changes requested):
besides below findings, please try to improve the tests time :( currently, iggy CI takes 11 minutes in github actions CI. if we merge this PR, it'll increase to 19 minutes, which is unacceptable.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Make sense and updated, will looked into others later.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Yeah, we can change to a shared hash, so we can preserve the label length while more robust.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: expose a connect_timeout_secs config field

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in 7b12517. Went with a single 64-bit joint hash over the raw `(stream, topic)` pair as its own label field, rather than the `N=8`/`H=16` swap — same 64-bit collision resistance, but `MAX_LABEL_NAME_LEN` stays at 16 so names remain readable. Worst-case label is now 120 chars (under cap), and `identity_hash` is length-prefixed to prevent boundary-shift aliasing.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in 2219a6d. Reworded both the error message and the comment to frame `[A-Za-z0-9_]+` as iggy's deliberately stricter subset acting as a path-traversal guard, not as Doris's actual identifier rules.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in ed80885 — `consume()` now returns `Error::InvalidPayloadType` on the first non-JSON payload instead of silently skipping.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in 864a458.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in 3b57ba1 — added optional `connect_timeout_secs` (default 5s, independent of `timeout_secs`).

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in ed837ce — switched to `.expect("chunks() never yields an empty chunk")`; a fabricated offset 0 would alias the real offset-0 label and break idempotency, so this should fail loud.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in ed837ce — body-read failures now log a `warn!` before falling back to an empty body.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Kept the per-chunk `continue` (documented in ed837ce, code + README). The runtime commits the consumer offset for the whole poll *before* `consume()` runs (AutoCommit at poll time), so bailing on the first failed chunk wouldn't replay the rest — it'd just drop them. Best-effort across chunks + returning the worst error (which still drives the runtime's error path) strictly dominates here.

core/connectors/sinks/doris_sink/Cargo.toml
ryankert01: Fixed in 2c74198 — added `publish = false`, matching every sibling connector crate.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in 2c74198 — marked `#[doc(hidden)]` with a note that it is `pub` only for the integration harness.

core/integration/tests/connectors/fixtures/doris/container.rs
ryankert01: Fixed in 2c74198 — dropped `pub`; they're only used inside `container.rs` and never re-exported (the integration crate still compiles clean with no dead-code warning).

core/connectors/sinks/doris_sink/README.md
ryankert01: Fixed in 2c74198 — dropped the `(todo)`.

core/connectors/sinks/doris_sink/README.md
ryankert01: Fixed in 2c74198 — reworded to make explicit that label dedup gives retry/redrive idempotency, not crash recovery: the runtime commits the offset at poll time before `consume()`, so a crash mid-load is at-most-once. Also corrected the stale label-format description (single 64-bit joint hash over the raw names, not per-segment 32-bit).

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in dc3d997 — `open()` now emits a `warn!` for an empty password (Doris permits passwordless users, so it stays valid but surfaces the likely misconfig).

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in dc3d997 — the URL is now computed once in `new()` and `stream_load_url()` returns the cached `&str`, so it is no longer re-formatted per batch.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in d7530e5 — but reused `PermanentHttpError` rather than adding a new `HttpProtocolError` variant. The redirect loop and missing-`Location` cases are now `PermanentHttpError` (matching the unparsable-`Location` sites next to them), so `is_permanent_error` no longer treats them as retryable; the genuine `request.send()` transport failure stays the transient `HttpRequestFailed`. A new SDK-wide variant felt like overkill given (a) the runtime does not branch on error variants today, and (b) "permanent HTTP error" already captures the semantics the classifier needs. Happy to promote it to a dedicated variant if you'd rather have the finer-grained type.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: (1) Done in 27c5897 — `redirect_rebuilds_full_request_on_be` intercepts the post-redirect PUT and asserts every header is rebuilt (incl. `Authorization`, which reqwest strips cross-host). (2) Agree the per-redirect jitter is cold-path — it only fires on cluster pathology and is capped at 5 — so leaving it as a follow-up rather than adding backoff machinery to the redirect loop now.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Fixed in e27826e — `max_filter_ratio` is now formatted once in `new()`. Note `Authorization` was already precomputed to a sensitive `HeaderValue` (in the redirect-hardening commit 864a458), and `columns`/`where` are stored as `String` and borrowed directly, so the ratio was the only remaining per-PUT allocation.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Pushing back on dropping this one. The connector is a `cdylib` loaded over the C ABI into the shared runtime; a panic here (which a `debug_assert!` + unwrap would become in a future refactor, or an `expect()` in release) crosses the FFI boundary and aborts the whole runtime, taking sibling connectors down with it. Returning a graceful `Error::InitError` keeps the failure contained to this sink even though the path is unreachable under the current lifecycle. I'd rather keep the one cheap branch as defense-in-depth — happy to revisit if you feel strongly.

core/integration/tests/connectors/doris/doris_sink.rs
ryankert01: Fair — the manual-PUT test proves Doris's server-side dedup but not the full kill-mid-batch + restart replay path the README claims. Tracking as a follow-up; a proper version needs runtime-restart harness support that's larger than this PR.

core/integration/tests/connectors/fixtures/doris/container.rs
ryankert01: Agreed, and noted as a follow-up (as you flagged). The single-slot serialization plus cold container boot is the CI cost; amortizing one container across tests with per-test databases is the right fix. The `std::sync::Mutex`-for-explicit-poisoning point is good too — folding both into the follow-up.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Agreed warn-only is correct for the opt-in `max_filter_ratio > 0` case (default Doris returns `Fail` on filtered rows). A `filtered_rows_total{db,table}` counter for alerting is the right call, but it needs a metrics handle the `Sink` trait doesn't currently expose over the C ABI — tracking as a follow-up.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+/// distinct pairs can't alias into one digest (e.g. `("ab","c")` vs `("a","bc")`).
+fn identity_hash(stream: &str, topic: &str) -> String {
+    let mut hasher = blake3::Hasher::new();
+    hasher.update(&(stream.len() as u64).to_le_bytes());
+    hasher.update(stream.as_bytes());
+    hasher.update(&(topic.len() as u64).to_le_bytes());
+    hasher.update(topic.as_bytes());
+    let hash = hasher.finalize().to_hex();
+    hash.as_str()[..LABEL_HASH_HEX_LEN].to_string()
+}
```
hubcio, line 531: `identity_hash` hashes only (stream, topic), but `label_prefix` is sanitized and truncated to 16 chars and excluded from the hash. two connectors writing to the same db.table from the same iggy stream/topic/partition/offset range whose prefixes collapse to the same 16 chars after sanitize+truncate produce identical labels, which doris's label dedupe (the same mechanism we treat as success at `lib.rs:475-483`) silently absorbs as `Label Already Exists` and the second batch is dropped. concrete collisions: `prod_events_us_east_1` and `prod_events_us_east_2` both truncate to `prod_events_us_e`; `prod.v1` and `prod_v1` both sanitize to `prod_v1`. silent cross-tenant data loss. fix: length-prefix the raw `prefix` bytes into the blake3 input alongside (stream, topic). ~4 LOC, same length-prefix discipline already used for stream/topic. the existing `label_disambiguates_names_that_sanitize_identically` test at `lib.rs:769-789` already establishes this threat model for the stream/topic dimension - close it for prefix too.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+                .header("format", "json")
+                .header("strip_outer_array", "true")
+                .header("label", label)
+                .body(body.clone());
+
+            if let Some(ratio) = self.config.max_filter_ratio {
+                request = request.header("max_filter_ratio", ratio.to_string());
+            }
+            if let Some(columns) = &self.config.columns {
+                request = request.header("columns", columns);
```
hubcio: `columns` and `where` config strings flow verbatim into `request.header(...)`. `reqwest::RequestBuilder::header` defers `HeaderValue::try_from` to `.send()`, so a CR/LF or any non-visible-ASCII byte in either config field makes every batch fail with a cryptic `HttpRequestFailed` error. runtime commits the consumer offset at poll time before `consume()` runs (`AutoCommitWhen::PollingMessages`) and discards `consume()`'s return, so every batch is silently lost for the lifetime of the bad config. http_sink already has the fail-fast pattern at `http_sink/src/lib.rs:1120-1127` (validate at `open()` via `HeaderValue::from_str`, fail with `InvalidConfigValue`). same fix here, optionally cache the resulting `HeaderValue` to skip re-parsing per batch (http_sink does at lib.rs:1129-1142). ~10 LOC.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+                .request(Method::PUT, url.clone())
+                .header(header::AUTHORIZATION, self.auth_header.clone())
+                .header(header::EXPECT, "100-continue")
+                .header("format", "json")
+                .header("strip_outer_array", "true")
+                .header("label", label)
+                .body(body.clone());
+
+            if let Some(ratio) = self.config.max_filter_ratio {
+                request = request.header("max_filter_ratio", ratio.to_string());
```
hubcio: heads-up: comment #20 says this was fixed in e27826ea7 (precomputed `max_filter_ratio_header: Option<String>`), but the later refactor 684e49dc7 ("fold client+base_url into a Connected struct") explicitly removed the field with stated reason "byte-identical output, config-static value". HEAD is back to `ratio.to_string()` per PUT and per redirect hop. the thread on #20 still says "Fixed in e27826e" with no follow-up. either restore the field or drop a note on the #20 thread disclosing the deliberate undo. perf delta is tiny (sub-µs per PUT) - process hygiene rather than perf-critical, but worth resolving so we don't ship with the thread out of sync with HEAD.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+            };
+
+            // `chunks()` never yields an empty slice, so first/last are present.
+            // Use `.expect` over `unwrap_or(0)`: a fabricated offset 0 would alias
+            // the real offset-0 label and break idempotency, so a broken invariant
+            // must fail loud.
+            let first_offset = chunk
+                .first()
+                .map(|m| m.offset)
+                .expect("chunks() never yields an empty chunk");
```
hubcio: `iggy_sink_consume` at `sdk/src/sink.rs:277` is `extern "C"` (not `extern "C-unwind"`) and there's no `catch_unwind` around the `runtime.block_on(sink.consume(...))` call at `sdk/src/sink.rs:225`. a panic across that boundary is UB per the rust nomicon's FFI-and-unwinding rule. the `.expect` is unreachable today (`messages.is_empty()` early-return at lib.rs:572 + `batch_size().max(1)` at lib.rs:179 guarantee non-empty chunks), but one refactor to `chunks_exact` or a filtered iterator brings the panic back and crosses the FFI boundary. cheap fix: `if let Some((first, last)) = chunk.first().zip(chunk.last()) { ... } else { continue; }`. byte-identical today, closes the hazard. ~6 LOC. the proper systemic fix is `catch_unwind` at the SDK FFI dispatch, but that's out of scope.

core/connectors/sinks/doris_sink/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+## How it works
+
+1. For each batch of messages, the connector serializes the JSON payloads into a JSON array.
+2. It computes a deterministic Stream Load `label` of the form `{label_prefix}-{stream_san}-{topic_san}-{hash16}-{partition}-{first_offset}-{last_offset}`.
+   - `hash16` is a single 64-bit blake3 hash computed over the *raw* (un-sanitized), length-prefixed `(stream, topic)` pair. So two identities that sanitize to the same string (e.g. `events.v1` vs `events_v1`) get distinct labels, and no boundary-shift aliasing is possible (`("ab","c")` ≠ `("a","bc")`).
+   - The total label is bounded under Doris's 128-char cap regardless of input length (worst case 120 chars).
+   - Doris dedupes loads by label inside its `label_keep_max_second` window. This makes a **runtime-driven retry/redrive** of the same batch idempotent — the second load is absorbed instead of duplicating rows. Note this is *not* a crash-recovery guarantee: the runtime commits the consumer offset at poll time, before `consume()` runs, so a crash mid-load yields at-most-once delivery (the batch is not replayed). Label dedup protects against in-process retries, not against lost batches.
```
hubcio: this line markets label dedup as a "runtime-driven retry/redrive" idempotency guarantee, but `send_stream_load` (lib.rs:202-323) has no retry loop (only the redirect-follow loop at lib.rs:219), `consume()` runs once per poll, and the runtime at `core/connectors/runtime/src/sink.rs:642` discards `consume()`'s return value. so today the label dedup protects zero production scenarios: not crash-recovery (at-most-once via `AutoCommitWhen::PollingMessages`), not runtime-redrive (doesn't exist), not in-process retry (doesn't exist). README line 91 already says "the runtime currently discards consume()'s return value, so there is no retry or DLQ" - the two lines contradict each other. either land a small `CannotStoreData` retry-with-backoff loop in `send_stream_load` (then dedup becomes load-bearing in v1) or downgrade this prose to "forward-compatible scaffolding for a future runtime-redrive feature".

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+                    return Err(Error::PermanentHttpError(format!(
+                        "Doris sink ID {} got {status} with no Location header",
+                        self.id
+                    )));
+                };
+                // Doris returns an absolute Location; fall back to joining a
+                // relative one. Validate the target before trusting it with
+                // credentials on the next iteration.
+                let target = reqwest::Url::parse(location)
+                    .or_else(|_| url.join(location))
```
hubcio: relative-`Location` fallback resolves against the current url (`/api/{db}/{table}/_stream_load`), not the FE root. `Url::parse("https://fe:8030/api/db/tbl/_stream_load").join("be_endpoint")` yields `https://fe:8030/api/db/tbl/be_endpoint` (sibling, almost certainly 404). dormant in practice - doris always emits absolute `Location` (per `container.rs:21-26`) - but the defensive fallback gives a false sense of safety. cleanest: drop the `or_else`, treat any non-absolute `Location` as `PermanentHttpError("non-absolute Location header")`. matches the documented doris contract.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+            base.scheme(),
+            target.scheme(),
+        )));
+    }
+
+    if let Some(allowed) = allowed_hosts
+        && !allowed.is_empty()
+    {
+        let host = target.host_str().unwrap_or("");
+        if !allowed.iter().any(|h| h.eq_ignore_ascii_case(host)) {
```
hubcio: two gaps in the redirect validation. (1) when `allowed_redirect_hosts` is `None` (default), no host check at all - a compromised FE can 307 to any same-scheme host. README line 81 honestly states this, so it's documented rather than hidden. (2) even when the allowlist IS set, port isn't checked - `validate_redirect` only compares `target.host_str()`. a compromised FE that controls the response can 307 to `https://fe:6379/exfil` (attacker port on an allowlisted host) and creds + payload exfiltrate. the allowlist UX promises host pinning but doesn't pin the endpoint. fix: include port in allowlist comparison (`target.port_or_known_default()`), and tighten the README security-notes wording about the unset-allowlist default.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+            let response_text = response.text().await.unwrap_or_else(|e| {
+                // Don't swallow a body-read failure: log it, then fall back to an
+                // empty body so the status-based handling below still classifies
+                // the outcome (an empty body on a non-2xx becomes permanent).
+                warn!(
+                    "Doris sink ID {} failed to read response body: {e}",
+                    self.id
+                );
+                String::new()
+            });
```
hubcio: on a 2xx response with a body-read failure (mid-stream TCP reset, decompression error, body-read timeout), the `String::new()` fallback falls through `!status.is_success()` (which skips for 2xx) and lands at `parse_stream_load_response(&response_text)`. `serde_json::from_str("")` errors, so the connector returns `PermanentHttpError("Failed to parse Doris stream load response: ...")` for a load that doris almost certainly persisted. today this only mis-logs (runtime ignores rc), but once retry/DLQ wires up this routes successful loads to the DLQ. fix needs a "success-but-counts-unknown" sentinel so the `if response.number_filtered_rows > 0` filtered-rows warning at lib.rs:647-658 doesn't fire on fabricated zero counts. alternative: classify as `CannotStoreData` so a future runtime-retry hits doris's label dedup and discovers the actual outcome.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+    pub database: String,
+    pub table: String,
+    pub username: String,
+    pub password: SecretString,
+    pub label_prefix: Option<String>,
+    pub max_filter_ratio: Option<f64>,
+    pub columns: Option<String>,
+    #[serde(rename = "where")]
+    pub where_clause: Option<String>,
+    pub timeout_secs: Option<u64>,
```
hubcio: `timeout_secs: Option<u64>` and the new `connect_timeout_secs: Option<u64>` diverge from sibling sinks. http_sink (`src/lib.rs:162`) and influxdb_sink (`src/lib.rs:103`) use `timeout: Option<String>` parsed via humantime. elasticsearch_sink uses `timeout_seconds: Option<u64>`. postgres_sink has no request-timeout field. this PR adds a fourth shape (`_secs` suffix vs elasticsearch's `_seconds`). v1 has no config users yet - last chance before the schema fossilizes. rename to `timeout: Option<String>` + `connect_timeout: Option<String>` via humantime_serde, matching the two most-recent web sinks.

core/integration/tests/connectors/fixtures/doris/container.rs
```diff
@@ -0,0 +1,610 @@
...
+use tokio::time::sleep;
+use tracing::info;
+use uuid::Uuid;
+
+// Global serialization lock for Doris fixtures. The all-in-one image binds
+// the BE HTTP port 1:1 (host:8040 → container:8040) so the FE→BE 307
+// redirect remains reachable; only one container per test process can hold
+// that port at a time. Each fixture holds the guard for its lifetime and
+// drops it on Drop, releasing the slot for the next test.
+static DORIS_LOCK: Lazy<std::sync::Arc<Mutex<()>>> =
```
hubcio: `DORIS_LOCK` mutex + the `threads-required = "num-cpus"` override at `.config/nextest.toml:31-34` + the host:8040 1:1 bind at line 177 serialize all 6 doris tests behind a fresh container boot. 6 × ~75s ≈ 7-8min CI - matches the 11→19min regression you flagged. fix: single shared `OnceCell<Arc<DorisContainer>>` per test binary, each test owns a fresh database (`iggy_test_db_<uuid>`). `CREATE DATABASE` ~1s vs ~90s boot. `create_test_database` already uses `raw_sql` for DDL (PREPARE-incompatible with doris MySQL frontend) and `create_table_with_template` already substitutes `{db}` placeholders, so the plumbing is mostly there. one concrete code touch: the missing-target-table test at `core/integration/tests/connectors/doris/doris_sink.rs:405-411` embeds the literal `'iggy_test_db'` in a `WHERE TABLE_SCHEMA = 'iggy_test_db'` clause - needs to be parameterized to per-test-DB. drops total doris CI wall to ~1 boot + 6 × DB-create + workload ≈ 2-3min. saves 5-7min CI.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1217 @@
...
+
+#[derive(Debug)]
+pub struct DorisSink {
+    id: u32,
+    config: DorisSinkConfig,
+    // Precomputed in `new()` and marked sensitive so reqwest keeps it out of any
+    // debug/trace output (and never HPACK-indexes it on HTTP/2).
+    auth_header: header::HeaderValue,
+    // Precomputed in `new()`; depends only on fe_url/database/table.
+    stream_load_url: String,
```
hubcio: consolidated simplify (four reviewers converged on this family). `stream_load_url: String` shadows `Connected.base_url: reqwest::Url` (which carries the parsed form anyway). same family of cleanup: drop the `stream_load_url()` / `label_prefix()` / `batch_size()` one-line getters (production + 2 test callers each), drop the `DorisSink::build_label` wrapper at lib.rs:185-200 (production can call the free fn directly; ~6 unit-test rewrites needed), and resolve `allow_insecure_redirect.unwrap_or(false)` + `allowed_redirect_hosts.as_deref()` once in `open()` onto `Connected` so `validate_redirect`'s signature drops from 5-arg to 3-arg. while you're there, replace the hand-rolled `format!("{}/api/{}/{}/_stream_load", config.fe_url.trim_end_matches('/'), ...)` at lib.rs:134-139 with `Url::parse(fe_url)?.join(...)` - same idiom, no `?query=` brittleness. saves ~15-20 LOC + a stringly-typed shadow.

core/connectors/sinks/doris_sink/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+reqwest = { workspace = true }
+secrecy = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+simd-json = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
+
+[dev-dependencies]
+wiremock = "0.6"
```
hubcio: `wiremock = "0.6"` is hard-coded. please promote it to main `Cargo.toml` and use `workspace = true`. `core/integration/Cargo.toml` also uses it.

hubcio (changes requested):
second pass of review below, you're getting there :)

core/integration/tests/connectors/fixtures/doris/container.rs
ryankert01: I just fixed it! 1 boot at each runner. (but it still needs ~14min)

ryankert01:
/ready

PTAL @hubcio 
Also the ci is failing because infra cancelled the task! Please help press re-run when you have time.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -219,6 +219,26 @@ runs:
...
+        # usually already up). Gated to Linux + integration in scope. A launch
+        # failure is non-fatal: doris tests fall back to self-booting their own
+        # container. BE 8040 is pinned 1:1 because the FE redirects Stream Load
+        # to 127.0.0.1:8040 (hardcoded in the image entrypoint).
+        DORIS_STARTED=0
+        if [[ "$RUNNER_OS" == "Linux" ]] && { [[ -z "$NEXTEST_FILTER" ]] || echo "$NEXTEST_FILTER" | grep -qi 'integration'; }; then
+          echo "::group::Launch shared Doris container (boots during build)"
+          if docker run -d --name iggy-doris-shared \
+              -p 8030 -p 9030 -p 8040:8040 \
+              apache/doris:doris-all-in-one-2.1.0; then
```
hubcio: `apache/doris:doris-all-in-one-2.1.0` here, but `container.rs:56` self-boots `apache/doris:4.0.3-all-slim`. ci ends up exercising a deprecated image while local runs (and the rest of the fixture) use 4.0.3. bump this to `apache/doris:4.0.3-all-slim` so ci matches what's actually tested.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1508 @@
...
+    /// Doris's FE legitimately redirects (307) to a BE on a *different host*, so
+    /// we can't require same-host. Instead we enforce two rules that close the
+    /// credential-exfiltration vector a compromised/MITM'd FE would otherwise have:
+    ///
+    ///   1. No scheme downgrade (`https` -> `http`) unless `allow_insecure_redirect`
+    ///      is set — a downgrade would push Basic-auth creds onto a cleartext hop.
+    ///   2. If `allowed_redirect_hosts` is non-empty, the target must match an
+    ///      entry. A bare-host entry pins only the host; a `host:port` entry pins
+    ///      the exact endpoint, refusing an allowlisted host on an attacker port.
+    fn validate_redirect(&self, target: &reqwest::Url, id: u32) -> Result<(), Error> {
```
hubcio, line 333: `validate_redirect` only catches the https -> non-https downgrade. a `Location: ftp://...` (or `file://`, `gopher://`...) from an `http://` fe slips past: downgrade=false because the base scheme isn't https, and with no allowlist set the host check is skipped too. credentials get re-attached and only reqwest rejects the scheme later, after the request was assembled. add an explicit `target.scheme() in {"http","https"}` check up front.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -0,0 +1,1508 @@
...
+    #[serde(default)]
+    number_loaded_rows: u64,
+    #[serde(rename = "NumberFilteredRows")]
+    #[serde(default)]
+    number_filtered_rows: u64,
+}
+
+impl DorisSink {
+    pub fn new(id: u32, config: DorisSinkConfig) -> Self {
+        let credential = format!("{}:{}", config.username, config.password.expose_secret());
```
hubcio: the password ends up materialized as a plain `String` in `credential` here, which lives on the heap until natural drop and isn't zeroized. the resulting `auth_header` is marked sensitive, but the intermediate string isn't. wrap it in `Zeroizing<String>` (or build the bytes into a `Zeroizing<Vec<u8>>` before base64) so the cleartext doesn't linger.

core/integration/tests/connectors/fixtures/doris/container.rs
```diff
@@ -0,0 +1,657 @@
...
+        expected: i64,
+    ) -> Result<i64, TestBinaryError> {
+        let mut last = 0i64;
+        for _ in 0..DEFAULT_POLL_ATTEMPTS {
+            // Doris is not strongly consistent for very freshly loaded data
+            // until the publish phase completes. Poll until the expected
+            // count is observed (or until we exhaust attempts).
+            if let Ok(c) = self.count_rows(database, table).await {
+                last = c;
+                if c >= expected {
```
hubcio: `wait_for_rows` returns as soon as `c >= expected`, but every call site does `assert_eq!(count, expected)`. if a duplicate ever lands (the label-window expiry replay is documented as reachable) you get an opaque count mismatch instead of a clear signal. either tighten the return to `==` or change the call-site assertions to `>=`.

hubcio (changes requested):
i think after fixing these PR would be ready to merge

.github/actions/rust/pre-merge/action.yml
ryankert01: good catch! (:

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: nice one, fixed

ryankert01:
Just fix the issues mentioned. Thank @hubcio for the review!!!
/ready

.github/actions/rust/pre-merge/action.yml
```diff
@@ -219,6 +219,43 @@ runs:
...
+              -p 8030 -p 9030 -p 8040:8040 \
+              -e SKIP_CHECK_ULIMIT=true \
+              --entrypoint bash apache/doris:4.0.3-all-slim \
+              -c "sed -i 's/-Xmx8192m -Xms8192m/-Xmx2048m -Xms2048m/' /opt/apache-doris/fe/conf/fe.conf && echo 'mem_limit = 4096M' >> /opt/apache-doris/be/conf/be.conf && exec bash /usr/local/bin/entry_point.sh"; then
+            DORIS_STARTED=1
+          else
+            echo "::warning::Failed to start shared Doris container; doris tests will self-boot"
+          fi
+          echo "::endgroup::"
+        fi
```
hubcio: im not a fan of this. why cant we make it work purely from rust code? (via testcontainers) how do you expect it to work for all devs on their machines (just ordinary dev calling `cargo test`) when such elaborate startup machinery is required?

hubcio: if something extra is required from iggy test harness we can implement it. i assume this logic here is needed because you want to run all tests on single doris instance, instead of creating a separate one from scratch for each testcase?

ryankert01: That's so true. Let's me think of a way that we can do it elegantly.

hubcio (approved):

spetz (approved):

PR #3223, opened 2026-05-07
feat(consensus): cache reply payload and notify subscriber on commit

krishvishal:
## Summary

- Cache committed reply bytes per-client in `ClientTable` (`CachedReply`, refcounted `Frozen` buffer, Arc-bump on dedup) and deliver them to in-process subscribers via a oneshot channel embedded in `PipelineEntry. Replaces the prior `Notify`/pending-map indirection; awaiters no longer need a separate cache lookup after wakeup.                                                                                                                                                
- `ClientTable` is metadata-only. Partition plane is at-least-once; consumers handle duplicates.
- New `EvictionHeader` / `EvictionReason` wire frame; typed reasons let the SDK trigger its eviction callback.
- Two-queue pipeline: `prepare_queue` (cap 8) plus a buffered `request_queue` (cap 64). `is_full()` reports only the prepare queue so admission stays separate from drop-on-full backpressure. View-change reset cancels subscribers and clears the request queue; the prepare queue survives for DVC reconciliation.


core/binary_protocol/src/consensus/header.rs
```diff
@@ -303,6 +308,188 @@ impl ConsensusHeader for ReplyHeader {
...
+    /// No session for `client_id`.
+    NoSession = 1,
+    /// Client release < cluster min.
+    ClientReleaseTooLow = 2,
+    /// Client release > cluster max.
+    ClientReleaseTooHigh = 3,
+    /// Invalid operation discriminant.
+    InvalidRequestOperation = 4,
+    /// Body failed state-machine validation.
+    InvalidRequestBody = 5,
```
numinnex, line 337: I am not sure if we want to evict clients on failed validation of the `Command`. Currently our server doesn't do that, instead it just sends an error response. @hubcio  What do you think ?

core/consensus/src/client_table.rs
```diff
@@ -15,207 +15,160 @@
...
+    /// Not seen; proceed with consensus.
     New,
-    /// Exact request already committed, re-send cached reply.
-    Duplicate(Message<ReplyHeader>),
-    /// Request is in the pipeline awaiting commit, drop (client should wait).
-    InProgress,
-    /// Request number is older than the client's latest committed request.
-    /// Already handled in a prior commit cycle, drop silently.
+    /// Exact request already committed; re-send cached reply.
+    Duplicate(CachedReply),
```
numinnex, line 97: We've been talking about it that I am not sure whether our clients table should cache the responses or only function as an deduplicator, that would send an error/empty response on duplicated `request_id`. 

There are potential issues with caching the response.

1) Our most common request path on the server is Send/Poll messages, in the SendMessages those cached reply are very small, on the PollMessages path that is not the case. Additionally we would have duplicated data in the memory, as we are already caching messages in the partition layer.

2) Non idempotent clients: Imagine a scenario in which our client as it's own internal logic stores `stream_id` or id of any other server resources in it's own data strucutres. Client creates stream with id 1, the request gets retried on timeout due to flaky network, the first request that got sent is handled by the server, the response is sent - client reacts to it by storing that Stream with id 1 has been created by storing it in its own DS, then the second request (the retried one) arrives at the server, the server sees that it is a duplicate sends back the cached response clients reacts to it by storing the information that Stream with stream id 1 has been created, essentially duplicating the information.

core/consensus/src/impls.rs
```diff
@@ -84,6 +84,10 @@ impl Sequencer for LocalSequencer {
 /// Maximum number of prepares that can be in-flight in the pipeline.
 pub const PIPELINE_PREPARE_QUEUE_MAX: usize = 8;
 
+/// Max accepted-but-not-yet-prepared requests buffered behind a full
+/// prepare queue. Beyond this, requests drop and the client retries.
+pub const PIPELINE_REQUEST_QUEUE_MAX: usize = 64;
```
numinnex, line 89: I am not sure whether we want to have such queue tho, I think once we reach the PIPELINE_PREPARE_QUEUE_MAX depth, we should simply reject those requests with some error indicating backpressure.

core/consensus/src/impls.rs
```diff
@@ -130,27 +164,33 @@ impl PipelineEntry {
...
 }
 
+/// Two-queue pipeline: in-flight prepares + buffered requests.
 #[derive(Debug)]
 pub struct LocalPipeline {
-    /// Messages being prepared (uncommitted and being replicated).
+    /// Uncommitted prepares; cap [`PIPELINE_PREPARE_QUEUE_MAX`].
     prepare_queue: VecDeque<PipelineEntry>,
+    /// Requests awaiting a prepare slot; cap [`PIPELINE_REQUEST_QUEUE_MAX`].
+    request_queue: VecDeque<RequestEntry>,
```
numinnex, line 193: As mentioned above, not sure about introducing this, especially considering the consensus per partition design, where each partition has it's own `LocalPipeline`.

core/server-ng/src/login_register.rs
```diff
@@ -17,71 +17,74 @@
...
+/// # Failures
+/// Transient (pipeline-full, view-change cancel, primary-not-caught-up,
+/// in-flight register) MUST eventually be absorbed via bounded retry. Until
+/// then they surface as [`LoginRegisterError::Transient`] → `NotEvictable`,
+/// so network can't ship a wire-terminal `Eviction` for a recoverable
+/// failure. SDK read-timeout replays.
+///
+/// Terminal → [`EvictionReason`] in `EvictionHeader`; SDK invokes its
+/// eviction callback and stops.
 pub trait RegisterSubmitter {
```
numinnex: Do we need to have this behind a trait ? Couldn't this go through the `IggyMetadata` path ? and just simply be yet another command that is handled by the `MuxStateMachine` ?

numinnex (changes requested):

core/binary_protocol/src/consensus/header.rs
hubcio: I think it depends on command. if command is malformed (i.e. deserialization fails) then we actually disconnect clients.

core/consensus/src/impls.rs
krishvishal: Request queue is useful for providing fairness to clients. For example metadata goes through only shard-0 and with reject-on-prepare-full, every freed slot triggers a retry race and the fastest clients monopolize them while slower ones starve and worsen the retry storm. The buffer makes admission order-preserving instead of race-based.

core/server-ng/src/login_register.rs
krishvishal: I will clean this up.

core/consensus/src/client_table.rs
krishvishal: 1. This PR splits the planes precisely to avoid that: ClientTable is metadata-only, partition plane  (Send/Poll) is at-least-once with no reply cache. Only metadata ops (Create*/Delete*/Register/ACL) land in ClientTable. Capped at CLIENTS_TABLE_MAX=8192 with oldest-committed eviction; entries are header + small body sized.
2.   VSR paper's client table design mandates at most once. Error-on-duplicate breaks at-most-once: client reads the error as failure, retries with a new request_id, server creates a second stream.

core/server-ng/src/login_register.rs
krishvishal: Pushed a commit addressing this feedback. Dropped the trait and calls `IggyMetadata::submit_register_in_process` directly

numinnex (dismissed):

numinnex (approved):

mmodzelewski (approved):

spetz (approved):

PR #3226, opened 2026-05-08
feat(server): add ObjectStorage trait + CompioFsStorage + InMemoryStorage

jayakasadev:

Part of #3228 — Phase 1a; umbrella stays open through Phase 10.

## Rationale

Adds the \`ObjectStorage\` seam so phases 2+ can migrate persistence subsystems onto a non-fs backend without touching every call site. No-op for fs deployments.

## What changed?

Persistence sites called \`compio::fs\` directly; there was no seam to plug a non-fs backend into.

A new \`ObjectStorage\` trait + two backends (\`CompioFsStorage\` passthrough, \`cfg(test)\` \`InMemoryStorage\`) live in \`core/server/src/streaming/storage/object_store.rs\`. A \`[system.storage]\` config block selects \`kind = "fs"\` (default) vs \`"object"\`; both resolve to fs since the S3 backend lands in #3227. The trait is \`?Send\` because compio's io_uring driver yields non-\`Send\` futures.

## Local Execution

- Passed: \`cargo fmt --all -- --check\`, \`cargo clippy --workspace --no-deps --tests -- -D warnings\`, \`cargo test --workspace --lib\` (11 new tests in \`streaming::storage::object_store::tests\`), \`cargo build --workspace\`, \`cargo machete\`, \`cargo sort --workspace --check\`.
- Pre-commit hooks: \`prek\` not installed locally; the standalone checks above were run.

## AI Usage

1. Claude (Anthropic, Opus 4.7).
2. Generated function — trait, both backends, config, tests drafted by the assistant; reviewed and edited locally for fit with iggy patterns (compio fs helpers, \`err_trail\`, \`static_toml!\`).
3. Verified via \`cargo test --workspace --lib\` and the feasibility spike that informed #3227.
4. Yes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11187,6 +11187,7 @@ dependencies = [
  "anyhow",
  "argon2",
  "async-channel",
+ "async-trait",
  "async_zip",
  "axum",
  "axum-server",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7900,6 +7900,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "md-5"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69b6441f590336821bb897fb28fc622898ccceb1d6cea3fde5ea86b090c4de98"
+dependencies = [
+ "cfg-if",
+ "digest 0.11.2",
+]
+
+[[package]]
 name = "memchr"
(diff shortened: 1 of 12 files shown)


hubcio:
See the reason for closure in [#3228](https://github.com/apache/iggy/issues/3228#issuecomment-4412311425).

PR #3231, opened 2026-05-09
feat(ci): label PR review state via slash commands and lifecycle

hubcio:
Reviewers cannot tell at a glance which open PRs are still in their
queue. Iggy's .asf.yaml requires 2 approvals with stale-dismiss, so
each push wipes prior approvals - the review backlog grows opaque
fast as PR volume rises.

Adopt rust-lang/triagebot's S-waiting-on-{review,author} pattern via
a single GitHub Actions workflow. Comment commands /ready, /author,
and /request-review @user move the labels explicitly; PR lifecycle
events (open, ready_for_review, converted_to_draft, closed) keep
them in sync without manual upkeep. Filter the queue with
`is:open is:pr label:S-waiting-on-review`.

The auth gate is author_association in {COLLABORATOR, OWNER}, which
matches @apache/iggy-committers. MEMBER is excluded deliberately -
it would admit any unrelated apache podling member.

issue_comment.created and pull_request_target are the only triggers;
the workflow never checks out a ref or executes PR-supplied code,
only calls the REST API via actions/github-script. This avoids the
pwn-request RCE class and stays inside the default GITHUB_TOKEN
scope - no PAT, no INFRA Jira ticket, no external host.

CODEOWNERS gains a `* @apache/iggy-committers` wildcard so reviewer
auto-request fires on every PR, not just .github/** changes.

### DIFF
--- a/.github/workflows/pr-triage.yml
+++ b/.github/workflows/pr-triage.yml
@@ -1,0 +1,351 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: PR Triage
+
+# Comment-driven PR triage and lifecycle labels. Inspired by
+# rust-lang/triagebot UX.
+#
+# Comment commands (parsed line-by-line in PR comments):
+#   /request-review @u [@u2 ...]   — request review from one or more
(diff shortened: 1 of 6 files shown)


mmodzelewski (dismissed):

.github/workflows/pr-triage.yml
```diff
@@ -0,0 +1,318 @@
...
+                      params.reviewers = [reviewer];
+                    }
+                    try {
+                      await github.rest.pulls.requestReviewers(params);
+                      core.info(`/request-review: requested ${reviewer}`);
+                    } catch (e) {
+                      core.warning(`/request-review: failed for ${reviewer}: ${e.message}`);
+                    }
+                  }
+                  continue;
```
lukaszzborek: for multi `/request-review` in single comment it should add only first one as reviewer?
Now it's set sawReassign as true, so next line will be ignore from adding next reviewer.

maybe it should be `/request-review @abcd @ddddd` 
or first go through all lines and at end call api to add reviewers, sets labels
or stay as is, all receive notification from comments

hubcio: good catch. went with option 2 + a bit of 1: /request-review now takes multiple handles on one line, e.g. `/request-review @a @b @org/team` and can also repeat across lines. all handles get collected, deduped, and sent in a single requestReviewercall at the end. dropped the sawReassign guard that was eating the extra lines.

updated docs too, thx @lukaszzborek

mmodzelewski (approved):

numinnex (approved):

PR #3233, opened 2026-05-11
feat(server,mcp): add systemd watchdog integration

jpds:

None.

## Rationale

I run my services under systemd - this is a simple integration to have the iggy runtime ping systemd per it's [sd_notify](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html) integration.

This does not have any effect when the systemd unit is not configured as `Type=notify` and without `WatchdogSec=`.

## Local Execution

Validated in a set of VMs with the NixOS integration testing framework:

```
vm-test-run-iggy> server # [   13.822570] iggy-server[937]: 2026-05-11T15:31:32.946941Z  INFO shard-0 server::shard::tasks::periodic::systemd_watchdog: Systemd watchdog enabled, pinging every 15s (timeout: 30s).
```

```
vm-test-run-iggy> mcp # [   14.012098] iggy-mcp[860]: 2026-05-11T15:31:33.267809Z  INFO iggy_mcp::systemd: Systemd watchdog enabled, pinging every 15s (timeout: 30s).
```


core/server/src/shard/tasks/oneshot/config_writer.rs
```diff
@@ -135,5 +135,10 @@ async fn write_config(
 
     info!("Current config written and synced to: {config_path} with all bound addresses",);
 
+    #[cfg(feature = "systemd")]
+    if let Err(e) = sd_notify::notify(&[sd_notify::NotifyState::Ready]) {
```
hubcio: `READY=1` is silently lost on any failure of `write_config()` above this line. the task is registered with `.critical(false)` (around line 32), and `?` propagates errors from `open` / `write_all_at` / `sync_all` (around lines 110-134) before reaching this notify. so any of {runtime dir missing, disk full, AppArmor block, read-only fs, custom path override} causes a healthy, fully-bound server to be killed by systemd's `TimeoutStartSec=` because `READY=1` never fires. `task_registry` will print error though.

fix: move the notify to right after listener-bind completes (the `if tcp_ready && quic_ready && http_ready && websocket_ready { break; }` block earlier in this function), before any fallible disk i/o. readiness is "listeners are bound", not "config file is fsynced". alternative: set `.critical(true)` on the oneshot so the task tears down with an error rather than silently dropping `READY=1`.

the MCP seems fine - `core/ai/mcp/src/main.rs:107` emits `READY=1` immediately after the synchronous `serve(stdio())` returns with no fallible i/o in between.

core/server/src/shard/mod.rs
```diff
@@ -194,6 +199,10 @@ impl IggyShard {
         // Spawn shutdown handler
         compio::runtime::spawn(async move {
             let _ = stop_receiver.recv().await;
+            #[cfg(feature = "systemd")]
+            if shard_for_shutdown.id == 0 {
+                let _ = sd_notify::notify(&[sd_notify::NotifyState::Stopping]);
```
hubcio: raw `sd_notify::notify(...)` inline here for `STOPPING=1`, and again in `core/server/src/shard/tasks/oneshot/config_writer.rs` for `READY=1`. meanwhile `core/ai/mcp/src/systemd.rs` wraps the same calls in a typed module (`notify_ready` / `notify_stopping` / `spawn_watchdog`). same project, two patterns -- a new contributor cannot grep for "where do systemd calls live in iggy" and get one answer.

extract `core/server/src/shard/systemd.rs` mirroring the MCP module shape and route all server-side calls through it. ~30 LoC refactor. fine as a follow-up PR if you want to keep this one small.

core/ai/mcp/src/systemd.rs
```diff
@@ -0,0 +1,50 @@
...
+    };
+
+    let interval = timeout / 2;
+    info!(
+        "Systemd watchdog enabled, pinging every {}s (timeout: {}s).",
+        interval.as_secs(),
+        timeout.as_secs()
+    );
+
+    tokio::spawn(async move {
```
hubcio, line 45: `tokio::spawn` infinite loop, `JoinHandle` dropped, no cancellation. not a true leak -- the process exits on `client_to_shutdown.shutdown().await` in `main.rs` and the task gets reaped -- but it is inconsistent with the server side at `core/server/src/shard/tasks/periodic/systemd_watchdog.rs` which goes through `task_registry.periodic(...)` with cooperative `ShutdownToken` wiring.

easy follow-up: accept a `CancellationToken` (from `tokio_util`) driven by the existing SIGTERM `select!` in `main.rs`, and replace `loop { sleep; ping }` with `select! { _ = token.cancelled() => break, _ = sleep => ping }`. not blocking.

core/server/src/shard/mod.rs
```diff
@@ -194,6 +199,10 @@ impl IggyShard {
...
+            }
             shard_for_shutdown.trigger_shutdown().await;
```
hubcio: `trigger_shutdown()`'s return value is discarded. `task_registry.graceful_shutdown(SHUTDOWN_TIMEOUT)` returns `false` if the timeout fires before all tasks drain (pending fsyncs, segment rotation, replica notifications), but systemd just sees a clean `STOPPING=1` + normal exit and journalctl logs "service stopped successfully" -- even when the 30s shutdown timeout dropped pending writes on the floor.

minor observability gap: forward a `STATUS=shutdown timed out` (or `sd_notify::NotifyState::Errno(...)`) when graceful shutdown returns false. not blocking.

hubcio (changes requested):
overall i think it's solid PR, at first I didn't want to create systemd dependency in iggy itself but your feature approach seems reasonable. i have a couple of follow-up items not tied to a specific line in this diff:

- track an issue to port systemd integration to `core/server-ng` because current `core/server` will be replaced by it (this is for us, we dont want to forget it before release)
- optional follow-up PR, preferably docs only for https://iggy.apache.org/docs/ (repo https://github.com/apache/iggy-website/ ): `iggy-server.service` skeleton with `Type=notify`, recommended `WatchdogSec=`, and hardening directives (`ProtectSystem`, `PrivateTmp`, `NoNewPrivileges`, `CapabilityBoundingSet`). i think it would be nice-to-have.

also: will the compilation for macos with feature `systemd` fail? or is `systemd` crate portable and uses some macos mechanism?

jpds:
> also: will the compilation for macos with feature systemd fail? or is systemd crate portable and uses some macos mechanism?

sd-notify itself is pure Rust, there's no libsystemd linkage, so `--features systemd` would compile fine on macOS; and then just no-op at runtime since no notification socket is set.

I've adjusted the code to address the review comments.

> optional follow-up PR, preferably docs only for https://iggy.apache.org/docs/ (repo https://github.com/apache/iggy-website/ ): iggy-server.service skeleton with Type=notify, recommended WatchdogSec=, and hardening directives (ProtectSystem, PrivateTmp, NoNewPrivileges, CapabilityBoundingSet)

I already made a systemd unit definition over at https://github.com/NixOS/nixpkgs/pull/519080 (if you want to pinch ideas from it).

hubcio (approved):

spetz (approved):

hubcio:
@jpds thanks for contribution. we merged this, but upcoming release will take from 2-3 weeks to 2-3 months, depend on when we finish cluster feature.

PR #3244, opened 2026-05-12
refactor(connectors): isolate per-connector init failures in runtime

mmodzelewski:
Previously, any failure during connector init (path resolution,
dlopen, state load, plugin init, consumer/producer setup) was
propagated as a fatal `RuntimeError` and aborted the entire
runtime. A single bad config or missing plugin file took down
every other healthy connector with it, and the failing connector
was hidden from the API/health view since it never reached the
manager.

Capture per-connector failures inline against the offending
plugin and continue iterating. Connectors that fail before their
FFI container can be loaded are returned from `sink::init` /
`source::init` as `FailedPlugin` entries and surfaced through the
manager as `ConnectorStatus::Error` with the captured message in
`last_error`, so operators can still see and diagnose them.

The `Result` return is preserved on both init paths for symmetry
and future fatal-error handling, but no system-level errors are
currently emitted.


spetz (dismissed):

core/connectors/runtime/src/sink.rs
```diff
@@ -59,102 +71,141 @@ pub async fn init(
...
+    FailedPlugin {
+        id,
+        key: key.to_owned(),
+        name: name.to_owned(),
+        path: config.path.clone(),
+        config_format: config.plugin_config_format,
+        error,
+        enabled: config.enabled,
+    }
 }
```
hubcio: this looks like FailedPlugin::new().

hubcio (changes requested):
overall pr is sound but it's not tested well enough. only one failure mode is tested (the schema-level poll_interval = "not-a-valid-duration", which fails inside setup_sink_consumers). The other branches added by this refactor are not exercised:
- resolve_plugin_path failure (e.g. path = "does-not-exist")
- Container::load failure (e.g. valid path but not a dylib)
- init_sink / init_source failure (plugin returns non-zero)
- FileStateProvider::load failure for sources
- the entire source side has no failure-isolation test

core/connectors/runtime/src/sink.rs
mmodzelewski: refactored the method to `new`

spetz (approved):

numinnex (approved):

hubcio (approved):

PR #3063, opened 2026-03-31
feat(bdd): add shared stream operations scenario for Go SDK

ex172000:
## Which issue does this PR close?

Related to #1986 

## Rationale

Adopt the same pattern as other BDD

## What changed?

Add a shared Gherkin feature file for stream CRUD operations and corresponding `godog` step definitions, beginning the migration of Go BDD tests from embedded Ginkgo scenarios to shared `.feature` files.

## Local Execution

- Passed
- Pre-commit hooks ran (`prek --last-commit` passed)

## AI Usage

1. Which tools? Claude Code
2. Scope of usage? Generate functions under supervision and planning
3. How did you verify the generated code works correctly? Testing and manual verify
4. Can you explain every line of the code if asked? Yes

Issue #1986, opened 2025-07-09
Rewrite golang BDD tests to use scenarios


lukaszzborek:
I noticed that the Golang BDD tests do not use the scenarios from [bdd/scenarios/](https://github.com/apache/iggy/tree/master/bdd/scenarios).
This should be changed. As stated in [Readme.md](https://github.com/apache/iggy/blob/master/bdd/README.md), all SDKs must use the same scenarios and be tested in the same way.

You can use [godog](https://github.com/cucumber/godog) or other framework to test with cucumber

hubcio:
@ex172000 once #3245 is done, what else is there to do?

justinmclean:
#2021 merged back in July 2025 and added a godog BDD suite for the Go SDK. Two later attempts at wider coverage, #3063 and #3245, both closed without merging. So is #2021 enough to call this done, or do we still want more scenario coverage?

PR #3245, opened 2026-05-13
feat(bdd): add shared stream operations scenario for Go SDK

ex172000:
Continues / supersedes #3063, which was auto-closed due to inactivity. The branch was force-pushed (rebase on current `master` plus a new fixup commit), so GitHub does not allow reopening the original PR.

Related to #1986

## What changed?

The four commits on this branch add a shared Gherkin feature file for stream CRUD operations and corresponding `godog` step definitions, then layer in `update`/`delete` for streams across all SDKs. The new commit on top addresses the remaining review feedback from #3063:

- **chengxilo:** `thenStreamDeletedSuccessfully` now verifies via `GetStream` against the server instead of checking a local pointer that `whenDeleteStream` itself set to nil.
- **chengxilo:** anchored all `sc.Step` regexes in `bdd/go/tests/basic_messaging.go` with `^...$` for consistency.
- **chengxilo:** kept the per-scenario `After` cleanup but documented it as best-effort, with a note that a global cleanup script is the better long-term direction.
- **hubcio:** made the stream-delete step explicit in the shared feature file (`When I delete the stream with name "..."`) instead of relying on implicit prior-step state. Updated the Go, Rust, Python, Java, C#, and Node step definitions to match.

## Local Execution

- Branch rebased on current `master`.
- `go vet ./tests/...` and `go test -count=0 ./tests/...` pass for `bdd/go`.
- Pre-commit hooks pass (`golangci-lint` and `csharp-dotnet-format` were skipped locally due to toolchain gaps — `golangci-lint` pinned at v1.64.8 (Go 1.23) cannot lint a Go 1.25 project, and `dotnet` is not on this machine; CI will exercise both properly).

## AI Usage

1. Which tools? Claude Code
2. Scope of usage? Generate functions under supervision and planning
3. How did you verify the generated code works correctly? Build/vet on Go side, manual diff review on the other SDKs


bdd/rust/tests/steps/streams.rs
```diff
@@ -64,3 +64,47 @@ pub async fn then_stream_has_name(world: &mut GlobalContext, expected_name: Stri
...
+        .delete_stream(&identifier)
+        .await
+        .expect("Should be able to delete stream");
+    world.last_stream_id = None;
+}
+
+#[then("the stream should be deleted successfully")]
+pub async fn then_stream_deleted_successfully(world: &mut GlobalContext) {
+    assert!(
+        world.last_stream_id.is_none(),
```
hubcio, line 107: this only checks `world.last_stream_id`, which the when-step at line 101 just set to `None` - it can never fail. for real coverage mirror the go pattern at `bdd/go/tests/basic_messaging.go:250-261`: re-fetch via `client.get_stream(&identifier)` against the deleted id and expect not-found / `None`.

bdd/python/tests/test_basic_messaging.py
```diff
@@ -272,3 +272,43 @@ def verify_last_message_match(context):
...
+        await context.client.delete_stream(name)
+        context.last_stream_id = None
+
+    asyncio.run(_delete())
+
+
+@then("the stream should be deleted successfully")
+def verify_stream_deleted(context):
+    """Verify stream was deleted"""
+    assert context.last_stream_id is None
```
hubcio, line 314: tautological - the when-step three lines above nulled `context.last_stream_id`, and this just asserts it's `None`. add a server roundtrip - `await context.client.get_stream(...)` against the deleted id and expect missing - to match `bdd/go/tests/basic_messaging.go:250-261`.

bdd/java/src/test/java/org/apache/iggy/bdd/BasicMessagingSteps.java
```diff
@@ -209,6 +211,30 @@ public void lastPolledMessageMatchesSent() {
...
+
+    @When("I delete the stream with name {string}")
+    public void deleteStream(String name) {
+        getClient().streams().deleteStream(StreamId.of(name));
+        context.lastStreamId = null;
+    }
+
+    @Then("the stream should be deleted successfully")
+    public void streamDeletedSuccessfully() {
+        assertNull(context.lastStreamId, "Stream should have been deleted");
```
hubcio, line 235: `assertNull(context.lastStreamId)` re-checks a local field the when-step at line 230 just nulled, so it can't fail. mirror `bdd/go/tests/basic_messaging.go:250-261`: call `getClient().streams().getStream(...)` against the deleted id and assert empty `Optional`.

foreign/csharp/Iggy_SDK.Tests.BDD/StepDefinitions/BasicMessagingOperationsSteps.cs
```diff
@@ -196,6 +196,36 @@ public void ThenTheLastPolledMessageShouldMatchTheLastSentMessage()
...
+    public async Task WhenIDeleteTheStream(string name)
+    {
+        await _context.IggyClient.DeleteStreamAsync(Identifier.String(name));
+        _context.CreatedStream = null;
+    }
+
+    [Then(@"the stream should be deleted successfully")]
+    public void ThenTheStreamShouldBeDeletedSuccessfully()
+    {
+        _context.CreatedStream.ShouldBeNull();
```
hubcio, line 227: the when-step at line 221 already nulled `_context.CreatedStream`, so this assertion can't fail. mirror `bdd/go/tests/basic_messaging.go:250-261`: `await _context.IggyClient.GetStreamByIdAsync(...)` against the deleted id and assert null.

foreign/node/src/bdd/stream.ts
```diff
@@ -45,6 +45,41 @@ Then(
...
+  async function (this: TestWorld, name: string) {
+    assert.ok(await this.client.stream.delete({ streamId: name }));
+  }
+);
+
+Then(
+  'the stream should be deleted successfully',
+  async function (this: TestWorld) {
+    // If we reached here without error, the stream was deleted successfully
+    assert.ok(true);
```
hubcio, line 79: `assert.ok(true)` is a no-op - it never fails regardless of server state. the when-step at line 71 already wraps the delete call in `assert.ok(...)`, so this then-step adds zero coverage. re-fetch via `this.client.stream.get({ streamId: name })` and assert it returns null / throws, matching `bdd/go/tests/basic_messaging.go:250-261`.

foreign/python/apache_iggy.pyi
```diff
@@ -259,6 +259,22 @@ class IggyClient:
 
         Returns Option of stream details or a PyRuntimeError on failure.
         """
+    def update_stream(
```
hubcio, line 262: `apache_iggy.pyi` is auto-generated by `pyo3_stub_gen` per the header at line 18. can you confirm this diff was produced by re-running the generator rather than hand-edited? hand edits get clobbered on next regen. the stub shape matches the surrounding methods so it's probably regenerated, just want to be sure.

bdd/scenarios/basic_messaging.feature
```diff
@@ -43,3 +43,9 @@ Feature: Basic Messaging Operations
     And the messages should have sequential offsets from 0 to 9
     And each message should have the expected payload content
     And the last polled message should match the last sent message
+
+    When I update the stream name to "test-stream-updated"
+    Then the stream name should be updated to "test-stream-updated"
+
+    When I delete the stream with name "test-stream-updated"
```
hubcio, line 50: delete-step uses the new name `test-stream-updated`. nothing in the scenario asserts the old name `test-stream` no longer resolves after the rename. minor gap - delete-by-new-name succeeding implies the rename happened, but an explicit `Then stream "test-stream" should not exist` would lock it down.

hubcio (changes requested):
minor extra note that didn't fit a hunk: `bdd/python/tests/test_basic_messaging.py:63-68` - `no_streams_in_system` is `pass`. other 5 SDKs actually call `get_streams()` and assert empty. relies on the `--fresh` server flag instead. asymmetric vs the rest; worth fixing while in the area but not blocking.

also noting a pre-existing gotcha that the new symmetric regex highlights: `bdd/rust/tests/steps/streams.rs:36` uses `^I create a stream with name (.+)$` (no quote strip) so it captures the literal gherkin quotes - rust BDD has been creating a stream actually named `"test-stream"` (with the double quotes) on the server. the new step at line 68 uses `"([^"]*)"` correctly. follow-up.

PR #3250, opened 2026-05-13
feat(csharp): implement rented message polling

lukaszzborek:
## Rationale                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                 
  Polling path allocated fresh byte[] per message payload + another per raw user-headers blob. On high-throughput consumers that's dominant allocation source. Goal: cut allocs by reusing pooled buffers, without breaking existing safe API.                                                                   
                                                                                                                                                                                                                                                                                                                 
##  What changed                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                 
  - Underlying poll now reads server bytes into MemoryPool<byte>.Shared.Rent(...) buffer instead of new byte[len]. Whole batch shares one rented buffer; payloads + raw headers are slices into it.                                                                                                              
  - New rented polling path: IggyConsumer.ReceiveRentedAsync() yields ReceivedRentedMessage. Slices stay alive until caller disposes. Saves 1–2 allocs per message vs old path.                                                                                                                                  
  - Typed variant: IggyConsumer<T>.ReceiveDeserializedAsync() deserializes whole batch eagerly, returns buffer to pool immediately. Caller disposes nothing.                                                                                                                                                     
  - New client method: IIggyClient.PollMessagesRentedAsync(...) → PolledMessagesRental.                                                                                                                                                                                                                          
  - IDeserializer<T>.Deserialize signature: byte[] → ReadOnlyMemory<byte>.                                                                                                                                                                                                                                       
  - Old ReceiveMessagesAsync / MessageResponse API unchanged on surface, but internally also reads into pooled buffer now and materializes byte[] only after mapping — saves 1 allocation per poll (the big receive buffer).                                                                                     
                                                                                                                                                                                                                                                                                                                 
##  What's next                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                 
  - Move encryption/decryption into IggyClient so it stops re-allocating (TODO in IggyConsumer.Rented.cs:158).                                                                                                                                                                                                   
  - Add rented / Memory / Span API on send path.                                                                                                                                                                                                                                                                 
  - Make header mapping lazy after decryption move.                                                                                                                                                                                                                                                              
  - Update docs in iggy website repo.

Benchmark in comment


lukaszzborek:
Some benchmark:

| Method                     | 0.8.0 Mean | Current Mean | Δ Mean   | 0.8.0 Gen0 | Current Gen0 | 0.8.0 Gen1 | Current Gen1 | 0.8.0 Gen2 | Current Gen2 | 0.8.0 Allocated | Current Allocated | Δ Allocated |
|--------------------------- |-----------:|-------------:|---------:|-----------:|-------------:|-----------:|-------------:|-----------:|-------------:|----------------:|------------------:|------------:|
| PollMessagesAsync          |   22.52 ms |     20.26 ms |  -10.03% |   312.5000 |     187.5000 |   312.5000 |      31.2500 |   312.5000 |            - |         3.91 MB |           2.97 MB |    -24.04%  |
| PollMessagesRentedAsync    |  22.52 ms* |     22.56 ms |   +0.18% |  312.5000* |     125.0000 |  312.5000* |            - |  312.5000* |            - |        3.91 MB* |           1.93 MB |    -50.64%  |
| ConsumerReceiveAsync       |   24.15 ms |     22.57 ms |   -6.54% |   312.5000 |     187.5000 |   312.5000 |      31.2500 |   312.5000 |            - |         3.97 MB |           3.02 MB |    -23.93%  |
| ConsumerReceiveRentedAsync |  24.15 ms* |     23.32 ms |   -3.44% |  312.5000* |     125.0000 |  312.5000* |            - |  312.5000* |            - |        3.97 MB* |              2 MB |    -49.62%  |
| TypedConsumerReceiveAsync  |   24.31 ms |     22.57 ms |   -7.16% |   312.5000 |     125.0000 |   312.5000 |      31.2500 |   312.5000 |            - |         4.02 MB |              2 MB |    -50.25%  |

\* Compared against the 0.8.0 non-rented equivalent (`PollMessagesAsync` / `ConsumerReceiveAsync`) as requested.

It was tested with a 1 KB message. Polling messages were performed in batches of 100 until 1,000 processed messages were obtained (deserialising the messages via MessagePack).

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.Rented.cs
```diff
@@ -0,0 +1,249 @@
...
+
+                await StoreOffsetAsync(lastPolledPartitionOffset, partitionId, false, ct);
+            }
+
+            if (anyNewMessages)
+            {
+                _lastPolledOffset.AddOrUpdate(rental.PartitionId, currentOffset, (_, _) => currentOffset);
+            }
+
+            if (_config.PollingStrategy.Kind == MessagePolling.Offset)
```
hubcio, line 224: liveness regression. if every message in the batch is skipped (all `Header.Offset <= lastPolledPartitionOffset`), `currentOffset` stays at `0ul`, then this branch rewinds `PollingStrategy` to `Offset(1)`. next poll fetches near the start, server returns same/overlapping page, all skipped again - infinite loop. the non-rented `PollMessagesAsync` returns early in the `!anyNewMessages` branch (see line 436); rented path is missing the same `return`. fix is either `return` after the auto-commit block when `!anyNewMessages`, or compute next offset from `lastPolledPartitionOffset + 1` instead of `currentOffset + 1`.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -112,7 +118,6 @@ await _client.LeaveConsumerGroupAsync(_config.StreamId, _config.TopicId,
         _consumerErrorEvents.Clear();
         _pollingSemaphore.Dispose();
         _connectionStateSemaphore.Dispose();
-
     }
```
hubcio, line 129: `DisposeAsync` (line 78) never drains `_rentedChannel`. if the consumer is disposed while messages are queued, each queued `ReceivedRentedMessage` still holds a `RentedBatchHandle` ref, so the batch refcount never drops and the pooled TCP buffer is only reclaimed once `SlicedMemoryOwner`'s finalizer runs - non-deterministic. complete the writer and drain/dispose any queued items during disposal.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
```diff
@@ -1236,4 +1279,59 @@ private async Task<bool> RedirectAsync(CancellationToken token)
...
+            }
+
+            ArrayPool<byte>.Shared.Return(_value);
+            if (suppressFinalize)
+            {
+                GC.SuppressFinalize(this);
+            }
+        }
+
+        ~SlicedMemoryOwner()
```
hubcio: soundness hole. `ReadOnlyMemory<byte>` slices (exposed as `RentedMessageResponse.Payload` / `RawUserHeaders`) reference the underlying `byte[]` directly, not the `SlicedMemoryOwner`. if a caller retains a slice but lets the owning `PolledMessagesRental` become unreachable, the finalizer here returns the array to `ArrayPool.Shared`. a subsequent `Rent` can hand the same array to another caller while the original slice still points at it - use-after-return / cross-tenant payload corruption.

fix is to either remove the finalizer (require deterministic dispose, accept the leak risk) or have every public slice retain/ref-count the owner so the array can't be returned while reachable.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -332,85 +333,75 @@ internal static OffsetResponse MapOffsets(ReadOnlySpan<byte> payload)
...
+            return new PolledMessagesRental(payloadOwner)
+            {
+                PartitionId = partitionId,
+                CurrentOffset = currentOffset,
+                Messages = []
+            };
         }
 
-        List<MessageResponse> messages = new();
+        List<RentedMessageResponse> messages = new((int)messagesCount);
```
hubcio: `messagesCount` is read straight from the wire and cast to `int` to pre-size the `List`. a malformed or hostile server response with `messagesCount` close to `uint.MaxValue` casts to a negative `int` and throws `ArgumentOutOfRangeException`, while a value of a few hundred million forces a large `List` allocation before any frame validation. cap by `(length - 16) / PropertiesSize` (the absolute upper bound for messages in the payload) before using it as capacity.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -62,7 +66,9 @@ public IggyConsumer(IIggyClient client, IggyConsumerConfig config, ILoggerFactor
         _config = config;
         _logger = loggerFactory.CreateLogger<IggyConsumer>();
 
-        _channel = Channel.CreateUnbounded<ReceivedMessage>();
+        _channel = Channel.CreateBounded<ReceivedMessage>(new BoundedChannelOptions((int)config.BatchSize * 2));
```
hubcio, line 69: `(int)config.BatchSize * 2` overflows for legal `uint` values. `BatchSize` is `uint` and the builder only rejects `0`, so any value above `int.MaxValue / 2` overflows the multiplication or wraps the cast to negative, and `BoundedChannelOptions` then throws with a misleading message. validate `BatchSize <= int.MaxValue / 2` at the builder / config-bind boundary, or compute the capacity in `checked` arithmetic.

foreign/csharp/Iggy_SDK/IggyClient/IIggyConsumer.cs
```diff
@@ -46,6 +46,18 @@ Task<PolledMessages> PollMessagesAsync(Identifier streamId, Identifier topicId,
...
 
+    /// <summary>
+    ///     Polls messages from a specified topic and partition while renting the payload buffers from a shared pool
+    ///     instead of copying them into byte arrays.
+    /// </summary>
+    /// <remarks>
+    ///     The returned rental must be disposed when the caller is done reading the payload and raw header memory.
+    ///     Payload and raw header slices are invalidated once the rental is disposed.
+    /// </remarks>
+    Task<PolledMessagesRental> PollMessagesRentedAsync(Identifier streamId, Identifier topicId, uint? partitionId,
```
hubcio, line 57: external implementers of `IIggyConsumer` can't satisfy this new required member. it returns `PolledMessagesRental`, whose only constructor (`PolledMessagesRental(IMemoryOwner<byte>)`) is `internal`. any third-party `IIggyConsumer` implementation outside `Apache.Iggy` cannot construct a valid result. either add a public factory / constructor to `PolledMessagesRental`, or move `PollMessagesRentedAsync` off the required public interface (e.g. default-impl that delegates to `PollMessagesAsync` + materialization adapter).

hubcio (changes requested):

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.Rented.cs
lukaszzborek: added same logic like in non-rented api

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
lukaszzborek: good catch, fixed

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
lukaszzborek: That was 50/50 for add finalizer or not when i implementing it. Now is 2:1 for accepting risk of not disposing) so I deleted it

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
lukaszzborek: fixed

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
lukaszzborek: added guard to builder

foreign/csharp/Iggy_SDK/IggyClient/IIggyConsumer.cs
lukaszzborek: changed PolledMessagesRental to have public constructor

hubcio (approved):

numinnex (approved):

PR #3256, opened 2026-05-14
fix(connector): extract dynamic route field as raw JSON string

### LINKED ISSUE #3177: Dynamic router extract_route_field produces JSON-escaped strings
In `iceberg_sink/src/router/dynamic_router.rs`, the `extract_route_field` method used `val.to_string()` on a `simd_json::OwnedValue`:

```rust
.map(|val| val.to_string())
```

For a string value like `tpch.lineitem`, `to_string()` produces the JSON representation `"tpch.lineitem"` (with literal quote characters). This means the route field value passed to `is_valid_namespaced_table` and used as a table lookup key contains surrounding quotes, so it either fails validation or looks up a nonexistent table name.

This means dynamic routing with string-valued route fields has never worked correctly, or has never been tested with a real Iceberg catalog end-to-end.

**Fix** : Use `val.as_str().map(|s| s.to_string())` instead of `val.to_string()`.

This should be covered by an integration test that writes a message with a string route field and verifies it lands in the correct Iceberg table.

Standing-Man:

Closes #3177.

## Local Execution

- Passed / not passed

### DIFF
--- a/core/connectors/sinks/iceberg_sink/src/router/dynamic_router.rs
+++ b/core/connectors/sinks/iceberg_sink/src/router/dynamic_router.rs
@@ -23,6 +23,7 @@ use iceberg::Catalog;
 use iceberg::table::Table;
 use iggy_connector_sdk::{ConsumedMessage, Error, MessagesMetadata, Payload};
 use simd_json::base::ValueAsObject;
+use simd_json::prelude::ValueAsScalar;
 use std::collections::HashMap;
 use tracing::{info, warn};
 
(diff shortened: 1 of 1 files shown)


hubcio:
@EdgarModesto23 mind taking a look at this one? it's iceberg-related. thanks!

EdgarModesto23:
Hello @Standing-Man . Thank you for your contribution! I left a message on the [original issue](https://github.com/apache/iggy/issues/3177) with some concerns I have. TL;DR: I haven't been able to reproduce the issue described. If you'd like to, I prepared a repository that can help you test this end-to-end and see the difference between master and your changes. If you do, please make sure to post your results here! I'll be more than glad to approve your changes if you find it's affecting master as described

Standing-Man:
> Hello @Standing-Man . Thank you for your contribution! I left a message on the [original issue](https://github.com/apache/iggy/issues/3177) with some concerns I have. TL;DR: I haven't been able to reproduce the issue described. If you'd like to, I prepared a repository that can help you test this end-to-end and see the difference between master and your changes. If you do, please make sure to post your results here! I'll be more than glad to approve your changes if you find it's affecting master as described

Hi @EdgarModesto23, Thank you for the review and for preparing the test repository! If you could share some guidance on how to test this issue end-to-end, that would be very helpful for me. I’ll compare the behavior between master and my changes and post the results here.

It would also be very helpful for me if you could provide detailed steps to reproduce this issue. I’ll use them to compare the behavior between master and my changes and post the results here. cc @atharvalade.

PR #3269, opened 2026-05-15
feat(cluster): wire multi-shard cross-shard communication for server-ng

hubcio:
core/server-ng bootstrapped exactly one shard with SHARD_ID=0 and
senders=vec![sender] hardcoded; the multi-shard path was dead code.
Cross-shard primitives copied from legacy core/server also did not
fit core/shard's crossfire model (bounded MTx + try_send-or-drop,
fd-transfer coordinator instead of SO_REUSEPORT).

bootstrap() now spawns N OS threads from sharding.cpu_allocation,
each pinned via nix::sched + hwlocality and running its own compio
runtime + IggyMessageBus + IggyShard for the partitions hashed to
that shard. Cross-thread shutdown rides an Arc<AtomicBool> polled by
a per-shard watchdog, since the bus' Shutdown is !Send and cannot be
triggered from the main thread directly. Partial shard-spawn failure
and shard-thread panic now signal cluster-wide shutdown instead of
hanging; the shutdown watchdog is detached from the bus drain.

ShardFrame becomes a concrete enum (Consensus + Lifecycle); the R
generic is lifted off IggyShard. Named routers (route_metadata /
route_partition / route_consensus_control) replace the duplicated
MessageBag match blocks, and a debug_assert at pump entry catches
receiver-thread mis-binding that the ctor's assert_sender_ordering
cannot see.

ShardMetrics records frame_drops_total (counter, variant+reason
labels), bumped at every inter-shard try_send rejection; without it,
drop-and-recover under VSR retransmit is operationally
indistinguishable from a livelock. The counter is atomic, so it is
safe to bump from !Send compio reactor contexts.

The legacy shard-mapping broadcast subsystem (periodic snapshot
refresh task, three-state MappingSlot table, ReplicaMappingUpdate /
ReplicaMappingClear frames) is retired entirely. Cross-shard replica
routing now flows through the cluster-shared ReplicaOwnerTable: the
owning shard's installer stamps its slot on a successful registry
insert and CAS-clears it on disconnect, so every bus' send_to_replica
slow path reads authoritative state with no broadcast or reconcile
loop. Builder accepts the coordinator at ctor; IggyShard stays
immutable post-construction.

message_bus forward-fn types widen to carry replica/client id, and
send_to_replica routes via the shared owner table so non-zero shards
reconcile against shared state rather than shard 0's private bus.

WAL recovery is serialized across shards: non-zero shards open the
WAL read-only, reject mutating ops (drain, set_snapshot_op) on a
read-only storage, route an invalid WAL header to truncate_or_fail,
and close the read-only fd once recovery completes.

Storage milestones (durable PartitionJournal, durable
(view, commit_op) watermark) and SDK (client_id, request_id)
durability across reconnect remain out of scope; tracked separately.


core/journal/src/prepare_journal.rs
```diff
@@ -387,6 +437,18 @@ impl Journal<FileStorage> for PrepareJournal {
...
+        // Advance the snapshot watermark only AFTER the WAL rewrite is
+        // durable (tmp create -> write -> fsync -> rename -> fsync parent
+        // -> reopen). Advancing earlier would leave `snapshot_op` past
+        // entries still present on disk on any `?` failure above, letting
+        // a future `append()` pass the slot collision check at
+        // `existing.op <= snapshot_op` and silently evict a live entry
+        // from the index. The entry would survive on disk but become
+        // unreachable, stalling `RetransmitPrepares` until view change.
+        if end_op > self.snapshot_op.get() {
+            self.snapshot_op.set(end_op);
```
jayakasadev, line 607: This is a correctness fix, not just a refactor. Previously `snapshot_op` advanced before the compacted WAL was durably renamed into place. A crash between those two points would leave the watermark past entries still in the old file — on the next boot, `append()` would see those ops as below the snapshot watermark, pass the slot-collision check at `existing.op <= snapshot_op`, and silently drop live index entries, making those prepares permanently unreachable to `RetransmitPrepares`.

core/journal/src/prepare_journal.rs
```diff
@@ -143,27 +186,40 @@ impl PrepareJournal {
...
             {
-                // Corrupt or non-prepare entry, truncate here
-                storage.truncate(pos).await?;
+                truncate_or_fail(&storage, pos, "corrupt or non-prepare entry").await?;
                 break;
             }
 
             let entry_size = u64::from(header.size);
 
+            // TODO(hubcio): verify `header.checksum` / `header.checksum_body`
```
jayakasadev, line 290: Is there a tracking issue for this? A body bit-flip that leaves the header structurally valid will pass the scan undetected and be replayed as committed state with no error.

core/message_bus/src/installer/replica.rs
```diff
@@ -266,6 +267,12 @@ pub fn install_replica_conn<C: TransportConn>(
     ) {
         Ok(token) => {
             install_token.set(Some(token));
+            // Stamp this shard as the owner so every shard's
+            // `send_to_replica` slow path can route to us immediately. The
+            // matching CAS-clear runs from
+            // `IggyMessageBus::notify_connection_lost` on either of the
+            // post-loop guards firing.
+            bus.mark_replica_owned(peer_id);
```
jayakasadev: There's a gap between `delegate_replica` returning `Ok(target)` in the coordinator and this line executing on the target shard. During that window, `send_to_replica` for this replica id reads `OWNER_NONE` and drops the frame silently. In practice this is likely safe — the replica has to complete login and session registration before any VSR message is addressed to it, so this stamp should fire well before then. Is that the assumed invariant?

core/server-ng/src/bootstrap.rs
```diff
@@ -42,52 +41,68 @@ use message_bus::transports::tls::{
...
 use std::sync::Arc;
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+use std::thread;
+use std::time::Duration;
 use tracing::{error, info, warn};
 
 const CLUSTER_ID: u128 = 1;
-const SHARD_ID: u16 = 0;
 const SHARD_REPLICA_ID: u8 = 0;
```
jayakasadev, line 88: Confirming my read: this is the single-node fallback — in cluster mode `current_replica_id` from the CLI is used instead, and all shards on the same node share the same replica id because shards are intra-node partition assignments, not VSR replicas. Is that the intended model?

core/configs/src/server_config/sharding.rs
```diff
@@ -21,11 +21,69 @@ use std::str::FromStr;
...
+    /// retransmit. Drops of cross-shard client Reply frames are terminal:
+    /// the client never receives the reply (no in-protocol retransmit).
+    /// Both frame classes share this one channel, so a consensus burst
+    /// can starve client-reply forwards: size against the worst-case sum
+    /// of consensus working set + peak client-reply fan-out per shard
+    /// occurring together; see `DEFAULT_INBOX_CAPACITY` for the
+    /// rationale. Used by `core/server-ng`; the legacy server uses its
+    /// own hard-coded inbox sizing.
+    #[serde(default = "default_inbox_capacity")]
+    pub inbox_capacity: usize,
```
jayakasadev, line 117: Consensus frames and client-reply forwards share this channel with no prioritization, so a VSR retransmit storm can starve client replies — the doc acknowledges this. Is there a tracking issue for separate priority lanes, or is that intentionally out of scope for now?

core/journal/src/prepare_journal.rs
hubcio: yep, exact framing. the failure mode is the slot-collision check at `existing.op <= snapshotpares. caught it while staring at the drain ordering; no existing test reproduced it because all tests crash-injected before rename, not after watermark.

you know, there is so much to do under clustering feature to we kind of have to do such big and possibly unrelated changes.

core/journal/src/prepare_journal.rs
hubcio: no issue yet. for now just todo is enough; we'll scan all TODOs for cluster-related crates (simulator, message bus, journal, etc) and fix them one by one when release comes. for now, we just want to be aware about problems, not necessary fix htem

core/message_bus/src/installer/replica.rs
hubcio: yes, that's the invariant. ordering on the target shard is: `replicas().insert(...)` succeeds (registry is the single source of truth) -> `mark_replica_owned(peer_id)` stamps the owner table -> only then does the dispatch loop drain `rx` and accept inbound VSR. any peer sending VSR to `peer_id` before insert completes is by definition pre-handshake, so `OWNER_NONE` -> drop is correct and observable via `frame_drops_total{reason="owner_none"}`. CAS-clear on disconnect is the symmetric half.

btw, `replica` is aimed to be used only in context of cluster; `shard` should be used only in context of single iggy-server

core/server-ng/src/bootstrap.rs
hubcio: correct. `SHARD_REPLICA_ID = 0` is only used when `current_replica_id` isn't passed (single-node dev / tests). in cluster mode the CLI-supplied `current_replica_id` propagates to every shard on the node, and shards on the same node share the replica id because the VSR replica is a node-level identity, not a per-shard one - shards are intra-node partition assignments hashed via `IggyNamespace`. so per node: one replica id, N shards.

core/configs/src/server_config/sharding.rs
hubcio: intentionally out of scope here. plan is two lanes - one bounded queue for consensus frames (drops recovered by VSR retransmit) and one for client `Reply` (drops terminal, must size for worst-case fan-out). the single-channel design here is the minimum-viable wiring so `frame_drops_total{variant,reason}` lights up under load and we get real numbers to size the split.

actually thansk for this comment, i'll add todo in code.

core/journal/src/prepare_journal.rs
jayakasadev: Good catch. LGTM on the fix.

core/journal/src/prepare_journal.rs
jayakasadev: Makes sense, TODO keeps it visible. CRC32/xxHash over the body at append + verify at scan when you're ready.

core/message_bus/src/installer/replica.rs
jayakasadev: Confirmed — ordering makes the gap safe, and `frame_drops_total{reason="owner_none"}` keeps it observable. LGTM.

core/server-ng/src/bootstrap.rs
jayakasadev: LGTM, model is clear.

core/configs/src/server_config/sharding.rs
jayakasadev: Metrics-first to size the split makes sense. Thanks for the TODO.

jayakasadev (approved):
All threads resolved. LGTM.

krishvishal (commented):
Commit, StartViewChange, DoViewChange, StartView (and Register requests) all carry Operation::Reserved / Operation::Register. Neither is in is_metadata() nor is_partition(), so route_typed falls through to route_consensus_control and hashes header.namespace. Metadata's consensus group is initialized with namespace = 0 (bootstrap.rs:1032), and Murmur3(0u64) >> 16 = 25477, so for every shard_count > 1, the frame is dispatched to a non-zero shard whose IggyMetadata.consensus is None. The receiver hits the fall-through arm and tracing::warn!-drops. VSR retransmit re-routes via the same hash and cluster wedges permanently.

hubcio:
good catch @krishvishal , should be fixed now. I added a sentinel `METADATA_CONSENSUS_NAMESPACE = 1u64 << 63` in `core/common/src/sharding/namespace.rs` (it's above `PACKED_NAMESPACE_MAX` so it can't collide with a real partition namespace), and bootstrap now inits the metadata group with that constant at `bootstrap.rs:1106` instead of 0.

router got an is_vsr_reserved arm in route_typed that short-circuits any frame with `namespace == METADATA_CONSENSUS_NAMESPACE` straight to shard 0 at `router.rs:172`, so metadata consensus traffic never touches the murmur path. per-partition consensus groups still hash through route_consensus_control like before.

I think its good enough for now.

krishvishal (approved):

numinnex (approved):

PR #3272, opened 2026-05-18
feat(simulator): deterministic seed based workload fuzz harness

krishvishal:
- Adds a deterministic fuzz driver. A single PRNG drives every choice (action, target replica, payload bytes); a name-keyed
 `Shadow` predicts server entity state; an `Auditor` matches replies to in-flight requests and tracks per-(client, namespace) bookkeeping.
 - 25 server commands wired through a uniform per-op surface (`sample` / `build_message` / `classify_reply` / `predicted_effect`). The `op_dispatch!` macro generates the dispatch table from the op list so forgetting an op is a compile error.
  - Multi-client support; per-(client, namespace) last-commit cursor reflects that each VSR group keeps its own op counter. Reply lookup runs before any cursor mutation, and the reply's `header.namespace` is cross-checked against the in-flight `request_namespace`.


numinnex (dismissed):

numinnex:
LGTM +1 binding

core/simulator/src/workload/auditor.rs
```diff
@@ -0,0 +1,170 @@
...
+        let Some(entry) = self.in_flight.remove(&key) else {
+            self.stats.replies_unknown += 1;
+            return None;
+        };
+
+        // Reply's namespace must match the namespace the request was
+        // submitted to. A mismatch means the reply landed in the wrong
+        // VSR group's bookkeeping; refuse to apply effects against the
+        // wrong shadow bucket.
+        if entry.request_namespace != header.namespace {
```
hubcio, line 149: ns-mismatch path removes the `in_flight` entry then returns `None`; the caller at `mod.rs:152-156` skips the `in_flight_per_client` decrement on that branch. unreachable today (server echoes request namespace verbatim, request ids are monotonic globally per `SimClient`), but a future routing or dedup bug would manifest as a silent single-client wedge at `CLIENT_REQUEST_QUEUE_MAX=1` instead of a clean `replies_unknown` count. cleanest fix is to return an enum `Match(InFlight) | NsMismatch | Unknown` so the caller decrements on `NsMismatch` without applying effects.

core/simulator/src/workload/ops/change_password.rs
```diff
@@ -0,0 +1,79 @@
...
+pub fn sample(
+    shadow: &mut Shadow,
+    outcome: Outcome,
+    prng: &mut Xoshiro256Plus,
+    _options: &WorkloadOptions,
+) -> Option<Input> {
+    match outcome {
+        Outcome::Success => {
+            let user = shadow.pick_user_name(prng)?;
+            let current_password = format!("pw-{user}");
```
hubcio: `format!("pw-{user}")` is correct only pre-first-commit. once the first `ChangePassword` succeeds server-side it stores `new_password`, but the next `sample()` rebuilds `pw-{user}` again, so a second `ChangePassword` on the same user mismatches the server. dormant under all-Success classification + weight=0 default. fix: shadow tracks per-user current pw; seed `pw-{user}` on `AddUser`, update on `ChangePassword` effect and on `RenameUser`.

core/simulator/src/workload/effect.rs
```diff
@@ -0,0 +1,106 @@
...
+        old: String,
+        new: String,
+    },
+    RenameTopic {
+        stream: String,
+        old: String,
+        new: String,
+    },
+    RenameUser {
+        old: String,
```
hubcio, line 88: `RenameUser { old, new }` carries no pw field; combined with the `pw-{user}` reconstruction at `change_password.rs:52`, a `ChangePassword` issued after `RenameUser` derives the wrong password. add a pw payload to the `RenameUser` effect and propagate it through `shadow.rs:227-231`. bundle with the change_password fix.

core/simulator/src/workload/shadow.rs
```diff
@@ -0,0 +1,313 @@
...
+                self.consumer_group_names.retain(|(s, _, _)| s != &name);
+            }
+            Effect::AddTopic {
+                stream,
+                name,
+                partitions: _,
+            } => {
+                if self.stream_names.contains(&stream) {
+                    self.topic_names.insert((stream, name));
+                }
```
hubcio, line 201: `AddTopic` silently no-ops when the parent stream is gone; same pattern at `AddConsumerGroup` (lines 197-205). `auditor.note_committed` at `mod.rs:174` runs unconditionally, so `commits_per_action` diverges from the shadow under multi-client interleave. same root cause as the multi-client rename collision below. single fix: `Shadow::apply` returns `ApplyResult { applied: bool }` and `note_committed` only fires on `applied = true`.

core/simulator/src/workload/shadow.rs
```diff
@@ -0,0 +1,313 @@
...
+                    self.user_names.insert(new);
+                }
+            }
+        }
+        ApplyResult { sim_commands }
+    }
+
+    fn rename_stream(&mut self, old: &str, new: &str) {
+        if !self.stream_names.shift_remove(old) {
+            return;
```
hubcio: `rename_stream`'s `if !shift_remove(old) { return; }` silently drops the second concurrent rename when two clients picked the same `old`. same pattern at `rename_topic` (lines 271-277). `CLIENT_REQUEST_QUEUE_MAX=1` is per-client; the second client's `pick_*_name` reads the shadow before the first rename's reply lands, so collisions are possible across clients. dormant under default weights but closes via the same `ApplyResult` change as the AddTopic finding above.

core/simulator/src/workload/ids.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub enum IdPermutation {
+    /// `index → index`.
+    Identity,
+    /// `index → u64::MAX - index`.
+    Inversion,
+    /// `index → (index << 32) | rand32(seed ⊕ index)`. Decode discards
+    /// the random low 32 bits. Index must fit in `u32`.
+    Random(u64),
```
hubcio, line 36: doc says `(index << 32) | rand32(seed XOR index)` but code at line 50 uses `seed.wrapping_add(data)`. align the doc, or replace the add with XOR.

core/simulator/src/workload/ids.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub fn encode(&self, data: u64) -> u64 {
+        match self {
+            Self::Identity => data,
+            Self::Inversion => u64::MAX - data,
+            Self::Random(seed) => {
+                debug_assert!(
+                    u32::try_from(data).is_ok(),
+                    "Random id permutation needs index <= u32::MAX, got {data}"
+                );
+                let mut prng = Xoshiro256Plus::seed_from_u64(seed.wrapping_add(data));
```
hubcio, line 50: `data & 0xFFFF_FFFF` silently truncates u32+ indices in release; the `debug_assert!` above only fires in dev. the variant is dead today (`id_permutation: IdPermutation::Identity` at `shadow.rs:73`). either return `Result` or drop the unused variant.

core/simulator/src/workload/shadow.rs
```diff
@@ -0,0 +1,313 @@
...
+        let index = self.next_index;
+        self.next_index += 1;
+        format!("wl-{prefix}-{index:08x}")
+    }
+
+    /// Apply a predicted effect to the shadow. Returns any
+    /// [`SimCommand`](crate::workload::effect::SimCommand)s the driver
+    /// must run against the simulator (e.g. `init_partition`).
+    ///
+    /// Cascades use `IndexSet::retain` for a single-pass O(n) walk that
```
hubcio: comment says "preserves insertion order" but the renames at lines 240/252/267/278/293 do `shift_remove` + `insert`, which appends to the end of the `IndexSet`. determinism is intact (replay over the same op sequence is stable) but readers will conflate the two paths. tighten the wording, or use `shift_insert` at the original index.

core/simulator/src/workload/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+/// pipeline's queue limits.
+pub const CLIENT_REQUEST_QUEUE_MAX: usize = 1;
+
+pub struct Workload {
+    prng: Xoshiro256Plus,
+    pub auditor: ServerAuditor,
+    pub shadow: Shadow,
+    pub options: WorkloadOptions,
+    /// Number of in-flight requests per client.
+    in_flight_per_client: HashMap<u128, usize>,
```
hubcio, line 65: `in_flight_per_client` (and `last_commit_per_client_ns` at `auditor.rs:67`) never reap entries for departed clients. bounded today by the fixed client set at `Simulator::new`; prune on disconnect when client churn lands.

core/simulator/src/workload/auditor.rs
```diff
@@ -0,0 +1,170 @@
...
+        self.commits_per_action[action as usize]
+    }
+}
+
+pub struct ServerAuditor {
+    in_flight: HashMap<(u128, u64), InFlight>,
+    /// Highest `commit` op observed per `(client, namespace)`. Each VSR
+    /// group (one per partition, plus metadata at `namespace = 0`) has its
+    /// own op counter; monotonicity is per-namespace, not per-client.
+    last_commit_per_client_ns: HashMap<(u128, u64), u64>,
```
hubcio: field name `last_commit_per_client_ns` implies strict monotonicity but the code at lines 132-136 only advances a watermark; the relaxation rationale is documented at lines 108-113. rename to `last_commit_watermark_per_client_ns` and tighten the field doc at lines 64-66 so the name matches behavior.

core/simulator/src/client.rs
```diff
@@ -181,8 +450,8 @@ impl SimClient {
         self.build_request_with_namespace(Operation::DeleteConsumerOffset, &payload, namespace)
     }
 
-    /// v2 of `store_consumer_offset` with an `AckLevel` byte. `NoAck` takes
-    /// the primary's fast path (no replication); `Quorum` goes through VSR.
+    /// Store offset with explicit `AckLevel`. `NoAck` takes the primary's
+    /// fast path (no replication); `Quorum` goes through VSR.
     pub fn store_consumer_offset_v2(
```
hubcio: `store_consumer_offset_v2` lacks the `# Panics` section that every other pub fn in this file carries (compare line 100, 122). it panics via `Owned::copy_from_slice` (size assert) and `Message::try_from(...).expect(...)`. add a `# Panics` block.

core/simulator/src/client.rs
```diff
@@ -200,7 +469,7 @@ impl SimClient {
         self.build_request_with_namespace(Operation::StoreConsumerOffset2, &payload, namespace)
     }
 
-    /// v2 of `delete_consumer_offset` carrying an explicit `AckLevel` byte.
+    /// Delete offset with explicit `AckLevel`.
```
hubcio, line 477: same as above: `delete_consumer_offset_v2` is missing a `# Panics` block but panics the same way.

core/simulator/src/workload/actions.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use strum::{EnumCount, EnumIter};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, EnumCount, EnumIter)]
+#[repr(u8)]
+pub enum Action {
+    // DO NOT REORDER (hash baseline depends on these indices).
+    CreateStream,
+    SendMessages,
+    StoreConsumerOffset2,
```
hubcio, line 30: `Action::StoreConsumerOffset2` / `DeleteConsumerOffset2` mirror the `iggy_binary_protocol::Operation` enum, but the corresponding client methods and op files use the `_v2` suffix (`store_consumer_offset_v2`, `ops/store_consumer_offset_v2.rs`). pick one. renaming the Action variants without reordering preserves discriminants, so the locked PRNG/hash baseline stays stable.

core/simulator/src/workload/options.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub weights: ActionWeights,
+    /// Send-batch size = `batch_size_min + prng.range(batch_size_span)`.
+    pub batch_size_min: u32,
+    pub batch_size_span: u32,
+    /// Probability a `StoreConsumerOffset2` request uses `Quorum` vs `NoAck`.
+    pub ack_quorum_ratio: f32,
+    /// Probability a request targets a non-primary replica (exercises the
+    /// redirect / forward path).
+    pub target_non_primary_ratio: f32,
+    /// Probability that a request is intentionally constructed to fail
```
hubcio, line 88: `invalid_request_ratio` is `pub` and self-documented as "Currently unused; reserved". drop it or land the consumer in this PR. same applies to `id_permutation: IdPermutation` at `shadow.rs:55`, which is only ever `Identity`.

core/simulator/src/workload/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+            return None;
+        }
+
+        let action = self.pick_action();
+        let target = self.pick_target_replica();
+
+        let Some((input, outcome)) =
+            ops::sample(action, &mut self.shadow, &mut self.prng, &self.options)
+        else {
+            self.samples_none += 1;
```
hubcio, line 119: `samples_none` is incremented here but never asserted on, and `workload_hash_for_seed` at `lib.rs:980-1006` does not include it. the doc at lines 97-104 calls out the PRNG-shift risk this counter is supposed to catch. either feed it into the determinism hash or add a `debug_assert!(samples_none == 0)` at quiesce.

core/simulator/src/workload/shadow.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Id permutation for fabricated-id paths (`Outcome::ResourceNotFound`).
+    /// Currently unused.
+    pub id_permutation: IdPermutation,
+    next_index: u64,
+}
+
+impl Shadow {
+    #[must_use]
+    pub fn new(namespaces: Vec<IggyNamespace>, id_permutation: IdPermutation) -> Self {
+        let namespaces_live: IndexSet<IggyNamespace> = namespaces.into_iter().collect();
```
hubcio, line 72: `namespaces.into_iter().collect()` into `IndexSet` silently dedups duplicate `IggyNamespace`. user-error class. add `debug_assert_eq!(namespaces.len(), namespaces_live.len())` on construction.

core/simulator/src/workload/ops/send_messages.rs
```diff
@@ -0,0 +1,92 @@
...
+                batch_len,
+                payloads,
+            })
+        }
+    }
+}
+
+#[must_use]
+pub fn build_message(client: &SimClient, input: &Input) -> Message<RequestHeader> {
+    let refs: Vec<&[u8]> = input.payloads.iter().map(Vec::as_slice).collect();
```
hubcio: `Vec<&[u8]>` rebuilt in `build_message` after `sample` already owns `Vec<Vec<u8>>`. pass `&[Vec<u8>]` directly, or have `Input` own `Vec<Bytes>` so the client side at `client.rs:411` can `Bytes::clone` (refcount bump) instead of `copy_from_slice`.

core/simulator/src/workload/shadow.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    fn rename_stream(&mut self, old: &str, new: &str) {
+        if !self.stream_names.shift_remove(old) {
+            return;
+        }
+        self.stream_names.insert(new.to_string());
+        // Rename in (stream, topic) and (stream, topic, group):
+        // collect-then-rebuild keeps the loop borrow simple.
+        let old_topics: Vec<(String, String)> = self
```
hubcio: `rename_stream` calls `shift_remove` inside the topic loop. each `shift_remove` is `O(N)`, total `O(N*M)`. cold path today (`UpdateStream` weight=0 default). single rebuild pass collapses it to `O(N)`.

core/simulator/src/workload/shadow.rs
```diff
@@ -0,0 +1,313 @@
...
+use iggy_common::sharding::IggyNamespace;
+
+use crate::workload::effect::{ApplyResult, Effect};
+use crate::workload::ids::IdPermutation;
+
+pub struct Shadow {
+    pub namespaces_live: IndexSet<IggyNamespace>,
+
+    /// Live streams by name. `CreateStream` inserts; `DeleteStream` removes.
+    pub stream_names: IndexSet<String>,
```
hubcio, line 46: `IndexSet<String>` for entity names grows linearly with creates; `fresh_name` allocates a new `String` each time. bounded today by `tick_budget` per run. flag for `Arc<str>` interning when long-running harness lands.

core/simulator/src/workload/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+             action={:?} classified={classified:?} expected={:?}",
+            entry.action, entry.outcome,
+        );
+
+        let effect = ops::predicted_effect(&entry.input, &entry.outcome);
+        let result = self.shadow.apply(effect);
+
+        self.auditor.note_committed(entry.action);
+
+        if let Some(count) = self.in_flight_per_client.get_mut(&header.client) {
```
hubcio: `count.saturating_sub(1)` will mask any future double-decrement once the auditor `on_reply` consumed-but-rejected hardening lands. switch to `checked_sub(1).expect("in_flight underflow")` so the invariant fails loud.

core/simulator/src/workload/ops/purge_stream.rs
```diff
@@ -0,0 +1,67 @@
...
+    client.purge_stream(&input.stream)
+}
+
+#[must_use]
+pub const fn classify_reply(_reply: &ReplyHeader) -> Outcome {
+    Outcome::Success
+}
+
+#[must_use]
+pub const fn predicted_effect(_input: &Input, outcome: Outcome) -> Effect {
```
hubcio, line 62: `predicted_effect = Effect::None` is acceptable today because `shadow.sends_committed` is keyed by `IggyNamespace` (packed stream/topic/partition) and the purge request carries only a name-based `WireIdentifier`. the shadow has no name -> ns reverse index, so it cannot zero the right keys. add a `TODO` so this surfaces when that reverse index lands and the offset clamp at `store_consumer_offset.rs:63-64` starts validating against post-purge state.

core/simulator/src/workload/ops/purge_topic.rs
```diff
@@ -0,0 +1,69 @@
...
+pub fn build_message(client: &SimClient, input: &Input) -> Message<RequestHeader> {
+    client.purge_topic(&input.stream, &input.topic)
+}
+
+#[must_use]
+pub const fn classify_reply(_reply: &ReplyHeader) -> Outcome {
+    Outcome::Success
+}
+
+#[must_use]
```
hubcio, line 64: same as `purge_stream.rs:62`. add a matching `TODO`.

hubcio (changes requested):

krishvishal:
@hubcio I've addressed review comments.

hubcio (approved):

numinnex (approved):

PR #3284, opened 2026-05-20
ci: add Miri UB detector for binary_protocol and consensus

krishvishal:
## Summary

  - Adds a `miri` task to the pre-merge action, scoped to `iggy_binary_protocol` and `consensus` (the unsafe-heavy crates that don't link compio's io_uring driver). Pinned
  `nightly-2026-04-21`, `-Zmiri-tree-borrows -Zmiri-strict-provenance`, separate cache key.
  - New `rust-miri` component triggers on `core/binary_protocol/**` and `core/consensus/**`. Mirrored in `justfile` as `just miri` for local runs.


core/journal/src/prepare_journal.rs
```diff
@@ -557,6 +568,36 @@ mod tests {
...
+
+        {
+            let journal = PrepareJournal::open(&path, 0).await.unwrap();
+            journal.append(make_prepare(1, 64)).await.unwrap();
+            journal.append(make_prepare(2, 128)).await.unwrap();
+            journal.storage.fsync().await.unwrap();
+        }
+
+        // Entry 2 at offset HEADER_SIZE+64=320; command byte at +60 within header.
+        let entry_2_offset = (HEADER_SIZE + 64) as u64;
```
hubcio, line 586: hard-coded `+ 60` for the command byte inside `PrepareHeader` is fragile. a field reorder of `size`/`view`/`release` silently corrupts the wrong byte and this test still passes (writes garbage to an unrelated field instead of clobbering the discriminant, no truncation triggers).

replace with `std::mem::offset_of!(PrepareHeader, command)`. same pattern is already used in `core/binary_protocol/src/consensus/message.rs` for `SIZE_OFF`/`COMMAND_OFF`/etc.

justfile
```diff
@@ -48,6 +48,16 @@ nextest: build
 nextests TEST: build
   cargo nextest run --nocapture -- {{TEST}}
 
+# Run Miri (UB detector) on the unsafe-heavy crates that don't pull
+# tokio/compio. Mirrors the `miri` task in CI. Requires a nightly toolchain
+# with the `miri` component:
+#
+#   rustup toolchain install nightly --component miri
```
hubcio: `cargo +nightly miri ...` is unpinned, but CI pins `nightly-2026-04-21` in `.github/actions/rust/pre-merge/action.yml`. local miri runs will drift from CI on the next nightly bump and either mask or surface failures that don't reproduce in CI.

either pin to `+nightly-2026-04-21` here, or add a comment that the date must stay in sync with `action.yml`.

core/common/src/types/send_messages2.rs
```diff
@@ -753,3 +774,49 @@ fn read_u128(bytes: &[u8], offset: usize) -> Result<u128, IggyError> {
...
+        }
+    }
+
+    #[cfg(debug_assertions)]
+    #[test]
+    #[should_panic(expected = "must be at least 16-byte aligned")]
+    fn decode_prepare_slice_debug_asserts_on_misaligned_input() {
+        // `Vec<u8>` requests align=1; glibc returns 16-aligned bases so
+        // `&buf[1..]` is reliably misaligned (16k + 1 mod 16 = 1).
+        let buf: Vec<u8> = vec![0u8; std::mem::size_of::<PrepareHeader>() + 1];
```
hubcio, line 814: `(16k + 1 mod 16 = 1)` is ambiguous - `16k` reads as 16384 (kilobytes) rather than `16 * k`. rephrase: `base addr is a multiple of 16, so base+1 has offset 1 mod 16`.

core/binary_protocol/src/consensus/iobuf.rs
```diff
@@ -450,11 +500,19 @@ fn extent_ref_count<const ALIGN: usize>(extent: &Extent<ALIGN>) -> usize {
     unsafe { extent.ctrlb.as_ref().ref_count.load(Ordering::Acquire) }
 }
 
+/// Decrement refcount; on last reference, drop the box and free the `AVec`.
+///
+/// # Safety
+///
+/// * `ctrlb`: live `ControlBlock`; caller owns one refcount share.
```
hubcio, line 507: minor: `assert!(old > 0, ...)` fires *after* `fetch_sub` has already wrapped on UAF, so by the time the assert trips, the refcount is corrupted. the abort is still the right move (beats a delayed double-free), but the comment could note that detection is post-corruption rather than pre, so a future reader doesn't expect this to catch the underflow before it happens.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -303,6 +307,29 @@ runs:
           --port 8090 --wait-secs 180
```
hubcio, line 307: `nightly-2026-04-21` is hardcoded and `bump quarterly` lives only in a comment - nothing enforces the cadence. consider a workflow-level env var (`MIRI_NIGHTLY`) so the date is in one place, or a scheduled reminder issue. out of scope for this PR but worth a follow-up.

hubcio (changes requested):

core/journal/src/prepare_journal.rs
krishvishal: Fixed.

justfile
krishvishal: Pinned the version and added a comment in `action.yml`

core/common/src/types/send_messages2.rs
krishvishal: Done.

core/binary_protocol/src/consensus/iobuf.rs
krishvishal: Added comment with clarification.

hubcio (approved):

numinnex (approved):

Issue #3258, opened 2026-05-14
Improve pre-push hook granularity and CI maintenance checks


Standing-Man:
## Background

The current pre-merge CI already has component detection for language-specific jobs, but several checks still run broadly:

- If the files being pushed do not include files relevant to a language/tool, that language/tool hook should not run.
- `Check PR Title` does not rerun when a PR title is edited.
- common checks in `_common.yml` always run, even when the changed files cannot affect that check.
- `scripts/ci/license-headers.sh` still has a Docker fallback for `addlicense`.

We should tighten CI execution so PRs run the checks that are relevant to the changed files, while preserving dependency-driven coverage for core changes.

## Goals

- [ ] ~~Make pre-push hooks run only when the pushed changes include files relevant to that hook.~~
- [x] Ensure PR title validation reruns after a PR title is edited.
- [x] Remove Docker from the license-header check path and evaluate replacing `addlicense` with HawkEye.

### Component

None

hubcio:
I think for point 3 you can ignore it, check what @krishvishal did couple of months ago in https://github.com/apache/iggy/pull/3095

Standing-Man:
This has been completed. Closing this issue.

PR #3301, opened 2026-05-22
ci: run PR title check from dedicated workflow

Standing-Man:

- Closes part of #3258.

## What changed?

Move the semantic PR title validation out of common pre-merge checks so editing a PR title can retrigger only the lightweight title CI.

## Local Execution

- Passed / not passed
Passed
Ran

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -1,0 +1,100 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: PR Title
+on:
+  pull_request:
+    branches: [master]
+    types: [opened, edited, reopened, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  pr-title:
+    name: Check PR Title
+    if: github.event.action != 'edited' || github.event.changes.title
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,100 @@
...
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  pr-title:
+    name: Check PR Title
```
hubcio: moving the title check out of `_common.yml` breaks the transitive dependency through `finalize_pr`. branch protection in `.asf.yaml` currently lists only `finalize_pr` as required, so this check is now advisory rather than merge-blocking. live API confirms the registered check-run name is `"Check PR Title"` (top-level workflows do not get the `workflow / job` prefix).

if keeping this as a required gate is intended, two options:

- add `contexts: - "Check PR Title"` to `.asf.yaml` under `protected_branches.master.required_status_checks`.
- rename this job to `name: PR Title` and add `"PR Title"` to contexts. workflow name and check name then align, single source of truth. cleaner.

.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,100 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: PR Title
+on:
+  pull_request:
+    branches: [master]
+    types: [opened, edited, reopened, synchronize]
```
hubcio, line 22: `synchronize` cannot change the PR title (it fires on new commits, no `changes` payload). redundant for the current advisory setup. keeping it is reasonable forward-compat though - the `amannn/action-semantic-pull-request` README recommends `synchronize` for any workflow that may become a required check, since required checks must run against the head SHA.

.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,100 @@
...
+
+permissions:
+  contents: read
+  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
```
hubcio: `cancel-in-progress: true` is safe while this check is advisory. if it is later promoted to a required status check, note that `cancelled` does not satisfy required status - rapid edits could leave a stale cancelled check on the PR. consider `cancel-in-progress: false` at that point.

hubcio (changes requested):

.github/workflows/pr-title.yml
Standing-Man: Thanks a lot for the review. I’ve added the PR title check to the required status checks for merging PRs and made this CI a required check. Since the PR title validation consumes very little CI resource, I think keeping it always enabled is reasonable.

hubcio (approved):

spetz (approved):

PR #3318, opened 2026-05-23
feat(python): Configure ruff properly and conform to modern python coding standards

slbotbm:

Closes #

## Rationale

Ruff needs to be configured properly for supply chain hardening + code needs to conform to modern python standards.

## What changed?

- Configured ruff to detect non-idiomatic python code and also increase security.
- deleted requirements.txt since that has been depreciated. Added pylock.toml, standardized in PEP 751.
- removed unused dependencies. 
- To conform to modern python guidelines, the rust source files had to be edited so that the generated stubs produced correct types. There were also places that were producing mangled stubs, and for those, `PyStubType` was used to produce correct stubs.

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
2. Scope of usage? research and stub correction
3. How did you verify the generated code works correctly? executed locally.


foreign/python/apache_iggy.pyi
```diff
@@ -16,28 +16,43 @@
...
+    "AutoCommitAfter",
+    "AutoCommitWhen",
+    "IggyClient",
+    "IggyConsumer",
+    "PollingStrategy",
+    "ReceiveMessage",
+    "SendMessage",
+    "StreamDetails",
+    "TopicDetails",
+]
```
hubcio, line 38: `__all__` is missing ReceiveMessageIterator, but the runtime module exports it. this makes from apache_iggy import * and direct typed imports disagree with runtime behavior. add it to the stub or stop exporting it from the pyO3 module.

hubcio: tbh i'm not 100% sure if this is really problem, please investigate @slbotbm

slbotbm: The python side exposes `ReceiveMessageIterator` as `AsyncIterator[ReceiveMessage]`, effectively hiding it behind `IggyConsumer.iter_messages()` in python. `__all__` only lists the classes directly exposed to the user, and thus, I do not think we need to export it.

hubcio (approved):

spetz (approved):

PR #3319, opened 2026-05-24
ci(python): add interpreter version sync check

### LINKED ISSUE #2776: [Python SDK] Create `sync-python-version.sh` to keep python versions in sync across folders and docker images
### Description

Create a bash script similar to `sync-rust-version.sh` that runs in CI and pre-commit hooks and checks that the python versions being used are all in sync. Currently we use Python 3.10.

### Component

Python SDK

### Proposed solution

_No response_

### Alternatives considered

_No response_

WaterWhisperer:

Closes #2776

## Rationale

Create a bash script that runs in CI and pre-commit hooks and checks that the python versions being used are all in sync.

## What changed?

The Python SDK declares its supported interpreter version in `foreign/python/pyproject.toml`, but related files can drift independently.

A new `sync-python-version.sh` check uses that `requires-python` value as the source of truth, verifies the matching docker/README/CI/BDD/examples references, and checks that wheel builds include the selected interpreter. The check is wired into common CI and pre-commit.

## Local Execution

- Passed:
  - `./scripts/ci/sync-python-version.sh --check`
  - `./scripts/ci/sync-python-version.sh --fix`
  - `./scripts/ci/shellcheck.sh --check scripts/ci/sync-python-version.sh`
  - `typos`
- Pre-commit hooks ran

## AI Usage

1. Tools used: Codex
2. Scope: repository inspection, implementation drafting, verification command selection.
3. Verification: reviewed the generated diff and ran the local checks listed above.
4. Yes


scripts/ci/sync-python-version.sh
slbotbm, line 1: This script is not working properly since it failed to detect that `.github/workflows/_build_python_wheels.yml` uses python 3.11, while it should be using 3.10.

slbotbm: It is only reading lines containing `--interpreter`, but not line like `python-version: "3.11"`

slbotbm: `iggy/foreign/python/.devcontainer/Dockerfile` is also not covered by this.

scripts/ci/sync-python-version.sh
```diff
@@ -0,0 +1,222 @@
...
+    "^requires-python = \">=${PYTHON_VERSION_REGEX}\".*$" \
+    "\\1${PYTHON_VERSION}\\2" \
+    "examples Python requirement"
+
+ensure_line \
+    ".github/actions/python-maturin/pre-merge/action.yml" \
+    "^([[:space:]]*python-version: \")[0-9]+\\.[0-9]+(\".*)$" \
+    "^[[:space:]]*python-version: \"${PYTHON_VERSION_REGEX}\".*$" \
+    "\\1${PYTHON_VERSION}\\2" \
+    "maturin action Python version"
```
slbotbm, line 358: Please add detection for `pylock.toml` to this as well as the pre-commit hooks.

slbotbm (changes requested):

scripts/ci/sync-python-version.sh
WaterWhisperer: Thanks, I'll update.

scripts/ci/sync-python-version.sh
WaterWhisperer: I don't see any `pylock.toml` files in the current tree, but I'll add optional detection for it as well.

slbotbm (approved):
lgtm from my side, but an issue remains: if we upgrade the python version to 3.11 and forget to remove code publishing 3.10 wheels, the script will still pass. (I am not really sure that is in the scope of this PR -- maybe @hubcio can comment?)

WaterWhisperer:
> lgtm from my side, but an issue remains: if we upgrade the python version to 3.11 and forget to remove code publishing 3.10 wheels, the script will still pass. (I am not really sure that is in the scope of this PR -- maybe @hubcio can comment?)

Thanks, I’ll wait for @hubcio’s thoughts.

scripts/ci/sync-python-version.sh
```diff
@@ -0,0 +1,301 @@
...
+        return
+    fi
+
+    if [ "$missing" -eq 0 ]; then
+        echo -e "${GREEN}✓${NC} $file: wheel builds include $target_interpreter"
+        return
+    fi
+
+    if [ "$MODE" = "fix" ]; then
+        sed -i -E "/--interpreter/ { /python${PYTHON_VERSION_REGEX}/! s/--interpreter/--interpreter ${target_interpreter}/; }" "$file"
```
hubcio: `ensure_wheel_interpreters` is additive only. on a `requires-python` bump (say 3.10 -> 3.11) the existing `--interpreter python3.10 python3.11 python3.12 python3.13` line already contains the new target as a substring, so `missing=0`, the fix is a no-op, and the stale `python3.10` entry stays in the wheel matrix. consider asserting the exact interpreter set instead of checking for substring presence.

scripts/ci/sync-python-version.sh
```diff
@@ -0,0 +1,301 @@
...
+        sed -i -E "s|$current_pattern|$replacement|" "$file"
+        FIXED_CHECKS=$((FIXED_CHECKS + 1))
+        echo -e "${GREEN}Fixed${NC} $file: $description"
+    else
+        echo -e "${RED}✗${NC} $file: $description is not $PYTHON_VERSION"
+        FAILED=1
+    fi
+}
+
+ensure_lock_python_requirement() {
```
hubcio, line 207: this function plus the `PYLOCK_FILES` loop at 271-281 is dead code - no `pylock.toml` exists under `foreign/python/`, `bdd/python/`, or `examples/python/`. the active drift surface is `uv.lock`, which is handled by the separate `uv-lock-check.sh` hook. suggest dropping the function, the loop, and the three `pylock\.toml` paths from the `files:` regex in `.pre-commit-config.yaml`.

scripts/ci/sync-python-version.sh
```diff
@@ -0,0 +1,301 @@
...
+    local file="$1"
+    local target_interpreter="python${PYTHON_VERSION}"
+    local missing=0
+    local total=0
+
+    TOTAL_CHECKS=$((TOTAL_CHECKS + 1))
+
+    while IFS= read -r line; do
+        total=$((total + 1))
+        if [[ "$line" != *"$target_interpreter"* ]]; then
```
hubcio: substring match here has no word boundary, and same applies to the sed regex at line 197. not currently exploitable since `PYTHON_VERSION` at line 72 captures `[0-9]+\.[0-9]+` only and `python3.10` won't realistically collide with `python3.100`, but anchoring with something like `python${PYTHON_VERSION_REGEX}([^0-9]|$)` would tighten future safety.

scripts/ci/sync-python-version.sh
```diff
@@ -0,0 +1,301 @@
...
+    local total=0
+
+    TOTAL_CHECKS=$((TOTAL_CHECKS + 1))
+
+    while IFS= read -r line; do
+        total=$((total + 1))
+        if [[ ! "$line" =~ $expected_pattern ]]; then
+            mismatched=1
+        fi
+    done < <(grep -E "$current_pattern" "$file" || true)
```
hubcio, line 112: `grep -E ... || true` masks ENOENT as `could not find $description`. same pattern at line 183. grep still prints `No such file or directory` to stderr so the real cause is adjacent, but a `[ -f "$file" ]` precondition would emit a clean missing-file error instead of the misleading one.

hubcio (changes requested):
besides below comments, `foreign/python/pyproject.toml` declares classifiers like `Programming Language :: Python :: 3.10..3.13` that the new sync script does not validate. on a `requires-python` bump those rows can drift, shipping misleading metadata to PyPI. worth adding an `ensure_classifiers` check now that source-of-truth tooling exists.

scripts/ci/sync-python-version.sh
slbotbm: the pylock files will be added in #3318. Leaving at least the pylock part as it is seems prudent.

hubcio: alright, then we'll merge #3318 first. @WaterWhisperer ignore this comment.

WaterWhisperer:
@hubcio thanks for the guidance

hubcio (approved):

spetz (approved):

numinnex (approved):

PR #3320, opened 2026-05-25
feat(cpp): Create fixture for clean up and add purge_* functions

slbotbm:

part of #2100 

## Rationale

Add a fixture for cpp to clean up resources left after tests. Also add `purge_topic` and `purge_stream` functions.

## What changed?

- Adds a fixture class, and utilizes it in the tests
- Adds `purge_topic` and `purge_stream` functions and their tests
## Local Execution

- Passed  
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
3. How did you verify the generated code works correctly? ran the tests, manually reviewed the code


foreign/cpp/tests/common/test_helpers.hpp
```diff
@@ -68,3 +64,120 @@ inline rust::Vec<rust::String> make_snapshot_types(std::initializer_list<const c
...
+        }
+    }
+
+    void CleanupStreams() {
+        if (tracked_stream_names_.empty() && tracked_stream_ids_.empty()) {
+            return;
+        }
+
+        iggy::ffi::Client *cleanup_client = nullptr;
+        ASSERT_NO_THROW({ cleanup_client = iggy::ffi::new_connection(""); });
```
hubcio: if any of these `ASSERT_NO_THROW` cleanup-client setup steps fail (connect/login), gtest's fatal assert does a `return;` out of `CleanupStreams()` before the `clear()` at the end - so the tracked stream names/ids stay in the vectors and the streams never get deleted server-side. on a long-lived shared server those leaked streams pile up across runs and can collide. do the delete loop / `clear()` before the throwing asserts, or use `EXPECT_*` here.

foreign/cpp/tests/common/test_helpers.hpp
```diff
@@ -68,3 +64,120 @@ inline rust::Vec<rust::String> make_snapshot_types(std::initializer_list<const c
...
+        tracked_stream_names_.erase(
+            std::remove(tracked_stream_names_.begin(), tracked_stream_names_.end(), stream_name),
+            tracked_stream_names_.end());
+    }
+    void ForgetTrackedStream(const std::uint32_t stream_id) {
+        tracked_stream_ids_.erase(std::remove(tracked_stream_ids_.begin(), tracked_stream_ids_.end(), stream_id),
+                                  tracked_stream_ids_.end());
+    }
+
+    void DeleteClient(iggy::ffi::Client *&client) {
```
hubcio, line 131: `delete_connection` frees the client (`Box::from_raw`) and only after that can return an error if `logout_user` fails. the `ASSERT_NO_THROW` runs before `ForgetClient(client)` and the null-out, so a throw does a fatal `return;` that leaves the already-freed non-null pointer in `clients_`. teardown's `CleanupClients()` then calls `delete_connection` on it again -> use-after-free + double-free. low probability (needs logout to fail mid-teardown) but real. forget/null the pointer before the assert, or use `EXPECT_NO_THROW` and forget unconditionally.

foreign/cpp/tests/common/test_helpers.hpp
```diff
@@ -68,3 +64,120 @@ inline rust::Vec<rust::String> make_snapshot_types(std::initializer_list<const c
     }
     return snapshot_types;
 }
+class E2ETestFixture : public ::testing::Test {
+  public:
+    ~E2ETestFixture() { Cleanup(); }
```
hubcio: `Cleanup()` runs gtest `ASSERT_*` from the destructor. it won't terminate (fatal assert is just a `return` and the FFI throw is caught by `ASSERT_NO_THROW`), so it's safe, but failures recorded after the test already ended are effectively useless. since `TearDown()` already calls `Cleanup()`, the dtor path could be non-asserting and let `TearDown` own the asserting cleanup.

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -626,58 +628,97 @@ TEST(LowLevelE2E_Client, GetClientsReturnsActiveClientSessions) {
...
-            found_match = true;
-            EXPECT_EQ(second_entry.has_user_id, first_entry.has_user_id);
-            EXPECT_EQ(second_entry.user_id, first_entry.user_id);
-            EXPECT_EQ(static_cast<std::string>(second_entry.address), static_cast<std::string>(first_entry.address));
-            EXPECT_EQ(static_cast<std::string>(second_entry.transport),
-                      static_cast<std::string>(first_entry.transport));
-            EXPECT_EQ(second_entry.consumer_groups_count, first_entry.consumer_groups_count);
-            break;
+    bool found_first_me_in_first_clients = false;
+    for (const auto &entry : first_clients) {
```
hubcio: these four blocks (654, 670, 686, 702) are the same scan-the-vec-for-a-client-id-then-5x-`EXPECT_EQ` loop with only the haystack and needle swapped. a small lambda like `expect_entry_matches(clients, me)` called 4x collapses ~50 lines with identical semantics.

foreign/cpp/Cargo.toml
```diff
@@ -31,8 +31,6 @@ bytes = "1.11.1"
 cxx = "1.0.194"
 iggy = { path = "../../core/sdk" }
 iggy_common = { path = "../../core/common" }
-# Explicitly enable the runtime + I/O drivers required by `Runtime::enable_all()` in lib.rs.
-# Listing the features here insulates this crate from upstream SDK feature changes.
 tokio = { version = "1.52.3", features = ["rt-multi-thread", "macros", "time", "net", "io-util"] }
```
hubcio, line 36: the comment explaining why the tokio features are pinned here (insulating this crate from upstream SDK feature changes for `Runtime::enable_all()`) got dropped. the feature list itself is unchanged so this is just lost intent - worth restoring the one-liner.

hubcio (changes requested):

foreign/cpp/tests/common/test_helpers.hpp
slbotbm: Used `EXPECT_NE` and `EXPECT_NO_THROW`

foreign/cpp/tests/common/test_helpers.hpp
slbotbm: Used `EXPECT_NO_THROW`

foreign/cpp/tests/common/test_helpers.hpp
slbotbm: done

foreign/cpp/tests/client/low_level_e2e.cpp
slbotbm: done

foreign/cpp/Cargo.toml
slbotbm: done

hubcio (approved):

krishvishal (approved):

PR #3321, opened 2026-05-25
feat(connectors): add agent docs, per-batch observability, atomic state

spetz:
Connector work lacked a concise agent-facing reference, and
runtime behavior around drops, shutdown, and state writes was
hard to reason about or debug. Per-batch latency visibility
was missing, some failure paths were under-accounted, hot
paths still paid avoidable metric lookup costs, and source
state persistence was not crash-atomic.

Add a repo-level AGENTS entry point plus `.claude` connector
skills anchored to concrete symbols so authoring rules survive
refactors. Extend the runtime with opt-in benchmark events,
always-on stage histograms, aligned sink/source stage labels,
JSON or text log selection, cached metric labels and counters
on hot paths, explicit filtered-message accounting, safer
source shutdown, and crash-atomic state saves with durable
parent-directory sync.

This keeps observability available without forcing benchmark
logs by default, improves failure-path correctness, and gives
future connector work one documented model to follow.


core/connectors/runtime/src/source.rs
```diff
@@ -616,22 +694,27 @@ pub(crate) extern "C" fn handle_produced_messages(
...
-                    if let Err(send_error) = sender.send(messages) {
-                        error!(
-                            "Failed to send messages for source connector with ID: {plugin_id}. Channel closed: {send_error}"
-                        );
-                    }
-                }
-                Err(err) => {
+        // Entry missing = SOURCE_SENDERS already cleaned up during shutdown - benign race,
+        // expected at close. No metric (operator would conflate it with real failures).
+        let Some(entry) = SOURCE_SENDERS.get(&plugin_id) else {
```
hubcio, line 724: compound bug between this silent-return and `manager/source.rs::stop_connector` ordering.

`stop_connector` (manager/source.rs:151) calls `source::cleanup_sender(plugin_id)` BEFORE the FFI close at :159. between those two points the SDK polling task is still alive (it only stops once `iggy_source_close` fires its internal shutdown watch), so the plugin keeps invoking this callback against an already-empty `SOURCE_SENDERS`. each such callback drops a polled batch AND its embedded `state` with no metric, no log. this is the EXPECTED path on every stop / restart, not just shutdown.

the 5s `tokio::time::timeout` on `handle` at manager/source.rs:154 does NOT abort the task either - per tokio docs, timeout just drops the wrapper future; the underlying `JoinHandle` detaches and the task continues running. if `file.save()` is mid-fsync past 5s, `start_connector` then constructs a fresh `FileStateProvider` on the same path with its own per-instance `save_lock` - the old task's late save can race tmp+rename with the new one and overwrite newer state with stale.

fix has two parts:
1. in `stop_connector`, close FFI first, then await tasks, then `cleanup_sender` - the plugin contract is no callbacks after `iggy_source_close` returns.
2. after the timeout, `handle.abort(); let _ = handle.await;` to prevent leaked tasks colliding with the next `start_connector`.

at minimum, drop a `tracing::trace!(plugin_id, "dropping produced batch: sender already cleaned up")` here so the drops are debuggable. the comment's argument against `warn!`/`error!` (operators conflating with real failures) is sound, but `trace!` is free.

core/connectors/runtime/src/metrics.rs
```diff
@@ -38,6 +40,94 @@ pub enum ConnectorType {
...
 
+#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelSet)]
+pub struct StageLabels {
+    pub connector_key: String,
+    pub connector_type: ConnectorType,
+    pub stage: Stage,
+}
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq, EncodeLabelValue)]
+pub enum Stage {
```
hubcio, line 66: `prometheus-client-derive-encode` derives `EncodeLabelValue` via `stringify!(ident)`, which emits the variant name verbatim - so this enum produces PascalCase label values: `stage="Prepare"`, `"Ffi"`, `"IggySend"`, `"StateSave"`, `"Total"`. prometheus naming convention (https://prometheus.io/docs/practices/naming/) is snake_case for label values.

`ConnectorType` PascalCase is pre-existing and worth leaving alone, but `Stage` is new in this PR - fix now before any dashboards start querying `stage="IggySend"` and lock the casing in. rename to `Prepare, Ffi, Decode, IggySend, StateSave, Total` -> hand-impl `EncodeLabelValue` writing `"prepare"`/`"ffi"`/`"decode"`/`"iggy_send"`/`"state_save"`/`"total"`, or rename variants to `IggySend` -> a wrapper that serialises snake_case.

core/connectors/runtime/src/source.rs
```diff
@@ -407,72 +427,114 @@ pub(crate) async fn source_forwarding_loop(
...
                 "Failed to process {count} messages by source connector with ID: {plugin_id} before sending them to stream: {}, topic: {}.",
                 producer.stream(),
                 producer.topic()
             );
             error!("{error_msg}");
-            context
-                .metrics
-                .increment_errors(&plugin_key, ConnectorType::Source);
+            context.metrics.inc_errors_with_labels(&labels.counter);
             context.sources.set_error(&plugin_key, &error_msg).await;
```
hubcio: one bad message permanently breaks the connector status.

`process_messages` propagates the first transform `Err` via `?` at line 659, this branch then calls `set_error` which flips the connector to `ConnectorStatus::Error`. `set_error` never auto-clears - the only paths that clear the error are `update_status(Running|Stopped)` in manager/source.rs:90-92. so a single malformed payload through `avro_convert` / `flatbuffer_convert` / `proto_convert` (all of which can return `Err` from `transform.transform()`) leaves the connector stuck in ERROR forever, while the loop happily continues processing subsequent batches via `continue` below.

fix: do per-message filtering inside `process_messages` itself - mirror the encoder.encode handling at lines 668-674 (log + `inc_errors_with_labels` + `continue`). don't promote a single-message failure to a connector-wide status.

core/connectors/runtime/src/state.rs
```diff
@@ -42,72 +40,364 @@ pub enum StateStorage {
...
-        self.file.lock().await.replace(file);
-        if buffer.is_empty() {
-            info!("State file is empty: {}", self.path);
-            Ok(None)
-        } else {
-            info!("Loaded state file: {}", self.path);
-            Ok(Some(ConnectorState(buffer)))
+        // Orphan tmps are reclaimed by the next save's truncate; deleting
+        // here would race a still-draining save across provider instances.
+        if let Some(parent) = self.path.parent()
```
hubcio: this parent pre-check is racy and has a worse failure mode than just removing it.

between the `fs::metadata(parent)` here and the `fs::read(&self.path)` at line 71, the parent dir could be removed. `fs::read` then returns `NotFound`, which the match arm at line 80 maps to `Ok(None)` - which the runtime interprets as "no state, start fresh". result: a missing-parent transient (e.g. mount flap, operator rm -rf race) silently rewinds the source's offset and replays everything from zero. the pre-check exists to distinguish that case but loses to TOCTOU.

also: `save()` has no equivalent pre-check, so the asymmetry is real.

fix: drop the pre-check entirely. after `fs::read` returns `NotFound`, re-stat the parent to decide between `Ok(None)` (parent exists, file simply missing) and `Err(CannotOpenStateFile)` (parent gone). closes the TOCTOU because both decisions are taken from the same post-failure observation.

.claude/skills/connector-runtime/SKILL.md
```diff
@@ -0,0 +1,286 @@
...
+
+2. **Stage-duration histograms** - `iggy_connector_stage_duration_seconds{connector_key, connector_type, stage}`. **Always on** regardless of the flag. Observed at the same per-stage points the text event uses. Bucket boundaries: `50µs, 100µs, 250µs, 500µs, 1ms, 5ms, 10ms, 50ms, 100ms, 250ms, 500ms, 1s, 5s`. Read via `/metrics` endpoint (requires `[http.metrics] enabled = true` in runtime config). Lets operators do `histogram_quantile(0.95, ...)` queries in Prometheus.
+
+When wiring new stages in the runtime, mirror the existing pattern: capture
+`Instant::now()` before the stage, capture `.elapsed()` after, call
+`metrics.observe_stage_with_labels(&labels.stage_x, elapsed)` using the
+**pre-built labels cache** (`SinkLabels` / `SourceLabels` from `metrics.rs`).
+The cache is built once per consumer/producer task at spawn time and threaded
+through (`Arc<SinkLabels>` for sink, `Arc<SourceLabels>` for source). This
+avoids per-batch `String::to_owned` for the label keys - cost stays at ~10 ns
```
hubcio: the `~10 ns per stage (Instant + atomic bucket increment)` claim is off by roughly an order of magnitude.

`observe_stage_with_labels` calls `Family::get_or_create(labels).observe(...)`. in `prometheus-client-0.24.1` (pinned in Cargo.lock), `get_or_create` does a `parking_lot::RwLock::read` + `HashMap::get` with `SipHash-1-3` over the full `StageLabels { connector_key: String, connector_type, stage }` per call. then `Histogram::observe` walks the 13-bucket array and bumps an atomic counter + atomic sum (the histogram itself sits behind an internal `Mutex`). realistic per-call cost is ~80-200 ns, not ~10 ns. "Instant + atomic bucket increment" undersells what `get_or_create` actually does.

this is still a win versus the pre-PR `&str` API which did `String::to_owned` per call (extra alloc on top of the same hash-probe cost). worth re-wording the doc to something like: `~80-200 ns/stage; avoids the per-batch String allocation of the old &str-based API`.

core/connectors/runtime/src/source.rs
```diff
@@ -407,72 +427,114 @@ pub(crate) async fn source_forwarding_loop(
...
+            benchmark::emit_source_event(
+                &plugin_key,
+                &topic_metadata.stream,
+                &topic_metadata.topic,
+                count,
+                sent_count,
+                benchmark::as_micros(decode_elapsed),
+                benchmark::as_micros(prepare_elapsed),
+                benchmark::as_micros(iggy_send_elapsed),
+                benchmark::as_micros(state_save_elapsed),
```
hubcio: `state_save_us` is emitted unconditionally with `as_micros(state_save_elapsed)`, but `state_save_elapsed` stays at `Duration::ZERO` when the batch has no state (the assignment at line 513 is inside `if let Some(state)`). every no-state batch therefore emits `state_save_us=0`. aggregating those zero samples with real save durations pulls down p50/p95/p99 in dashboards / benchmark traces.

fix options: (a) skip emitting `state_save_us` when state was absent (split the event or use `Option<u64>`), or (b) emit only when `state_save_elapsed > 0`.

core/connectors/runtime/src/log.rs
```diff
@@ -31,38 +31,58 @@ use tracing_subscriber::EnvFilter;
...
+        }
+        (false, LogFormat::Json) => {
+            tracing_subscriber::registry()
+                .with(env_filter())
+                .with(tracing_subscriber::fmt::layer().json())
+                .init();
+        }
+    }
 
     if telemetry_config.enabled {
```
hubcio, line 65: the `Logging initialized (format: ..., ...)` line is only emitted when `telemetry_config.enabled` is true. operators flipping `IGGY_CONNECTORS_LOGGING_FORMAT=json` with telemetry off get no startup confirmation that the format took effect.

the current message wording references "telemetry enabled, service name: ..." so it can't just be hoisted as-is. emit a stripped-down `info!("Logging initialized (format: {})", logging_config.format)` outside the conditional, and keep the telemetry-flavored line where it is.

core/connectors/runtime/src/source.rs
```diff
@@ -407,72 +427,114 @@ pub(crate) async fn source_forwarding_loop(
...
-        match &state_storage {
-            StateStorage::File(file) => {
-                if let Err(error) = file.save(state).await {
-                    let error_msg = format!(
-                        "Failed to save state for source connector with ID: {plugin_id}. {error}"
-                    );
-                    error!("{error_msg}");
-                    context.sources.set_error(&plugin_key, &error_msg).await;
-                    continue;
+        let state_save_start = Instant::now();
```
hubcio: `Instant::now()` sampled here, but `state_save_elapsed` is only assigned inside the `if let Some(state)` block at line 498. when the batch has no state, this clock read is dead work. move `let state_save_start = Instant::now();` inside the `Some` arm. trivial - ~20 ns per stateless batch, but free to fix.

core/connectors/runtime/src/source.rs
```diff
@@ -50,7 +52,13 @@ use crate::{
 use iggy_connector_sdk::api::ConnectorStatus;
 use tokio::task::JoinHandle;
 
-pub static SOURCE_SENDERS: Lazy<DashMap<u32, Sender<ProducedMessages>>> = Lazy::new(DashMap::new);
+pub struct SourceSenderEntry {
```
hubcio: `SourceSenderEntry` and its fields don't need to be `pub`. the runtime is a binary crate (no `[lib]`), grep confirms `SourceSenderEntry` / `SOURCE_SENDERS` are referenced only within this file. tighten to `pub(crate)` for both the struct and its fields, plus `SOURCE_SENDERS` and `cleanup_sender` at lines 61 and 63.

core/connectors/runtime/src/metrics.rs
```diff
@@ -38,6 +40,94 @@ pub enum ConnectorType {
...
+#[derive(Debug, Clone)]
+pub struct SinkLabels {
+    pub counter: ConnectorLabels,
+    pub stage_prepare: StageLabels,
+    pub stage_ffi: StageLabels,
+    pub stage_total: StageLabels,
+}
+
+impl SinkLabels {
+    pub fn new(plugin_key: &str) -> Self {
```
hubcio, line 108: `SourceLabels::new` (lines 109-129) already uses a `let stage = |s: Stage| StageLabels { connector_key: key.clone(), connector_type: ConnectorType::Source, stage: s };` closure to keep the body short. mirror the same pattern here - saves about a dozen lines and removes the repeated `connector_type: ConnectorType::Sink, stage: ...` boilerplate.

core/connectors/sdk/src/api.rs
```diff
@@ -112,6 +112,7 @@ pub struct ConnectorStats {
     pub messages_consumed: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub messages_processed: Option<u64>,
+    pub messages_filtered: u64,
```
hubcio: consider `#[serde(default)] pub messages_filtered: u64`. no `foreign/` SDK currently consumes `ConnectorStats` (grep returns 0 hits), so the impact today is zero - but adding `#[serde(default)]` future-proofs the field for any non-Rust deserializer that lands later. matches the leniency pattern the sibling `Option<u64>` fields already have via `skip_serializing_if`.

hubcio (changes requested):
a few non-blocking items that fall outside the changed hunks, recorded here so they don't get lost:

- `core/connectors/runtime/src/configs/runtime.rs::Display for ConnectorsRuntimeConfig` (around the `write!` near line 366) was not updated when `logging: LoggingConfig` was added at line 145 - operators dumping the config won't see the new format setting.
- `core/connectors/runtime/src/configs/connectors.rs::Display for SinkConfig`/`SourceConfig` (around line 313 / 332) similarly omits the new `verbose` and `benchmark` fields.
- `core/connectors/runtime/src/sink.rs:42-44` imports only `Instant`, but `Duration` is reached via fully-qualified path at line 552. fold into `use std::time::{Duration, Instant};`.
- `core/connectors/runtime/example_config/config.toml` has no `[logging]` section and none of `example_config/connectors/*.toml` shows `benchmark = true`. operators learn the new knobs only from source. add one example each.
- the state save now does +1 dir-fsync per save vs the previous in-place code. pre-PR was 1 `sync_all`, post-PR is 1 `sync_data` + 1 parent-dir `sync_all`. on NVMe that's ~50-200 us extra per save - worth a one-liner in the PR body / CHANGELOG so operators don't blame the upgrade for state-save latency.

on a separate axis - there's a compound shutdown / restart bug between `manager/source.rs::stop_connector` and the silent-return added to `source.rs::handle_produced_messages`. the inline comment on `source.rs` covers it.

.claude/skills/connectors-overview/SKILL.md
```diff
@@ -0,0 +1,236 @@
...
+   - Source: `connector_type, connector_key, stream, topic, batch_size, sent_count, decode_us, prepare_us, iggy_send_us, state_save_us, total_us`
+   - Filter live: `RUST_LOG=iggy_connectors::benchmark=info`
+
+2. **Prometheus histograms** (always on, regardless of flag) - aggregated for proper p50/p95/p99 queries:
+   - Metric: `iggy_connector_stage_duration_seconds{connector_key, connector_type, stage}`
+   - `stage`: `Prepare`, `Ffi` (sink), `Decode`, `IggySend`, `StateSave` (source), `Total` (both)
+   - Buckets: 50µs - 5s, geometric (see `runtime/src/metrics.rs::STAGE_BUCKETS_SECONDS`)
+   - Query example: `histogram_quantile(0.95, sum(rate(iggy_connector_stage_duration_seconds_bucket{stage="Ffi"}[5m])) by (le, connector_key))`
+
+The flag only gates the verbose text event; the histogram observation is always on. Cost when no observers attached: ~50 ns per batch (5 stages × `Instant::now()` + atomic bucket bump). Negligible at every realistic throughput.
```
hubcio: two issues here. mechanism description is wrong - `Histogram::observe` in `prometheus-client-0.24.1` takes `self.inner.write()` (parking_lot rwlock write guard) then linear-scans up to 14 buckets, not an atomic. `Family::get_or_create` also hashes the full `StageLabels` (incl. the `String connector_key`) under a `parking_lot::RwLock::read()` per call. and the number is off by an order of magnitude - realistic per-stage cost is ~100-180 ns (sip-hash on the string + hashmap probe + two rwlock acquires + 14-bucket scan + `Instant::now()`). per batch: source ~600 ns - 1.1 us (5 stages), sink ~360-660 ns (3 stages). cost is paid every batch regardless of `/metrics` scrape state.

suggested rewrite: `~500 ns - 1 us per batch (5 stages * (Instant::now() + String hash + 2 parking_lot lock acquires + 14-bucket linear scan)). negligible below ~10k batches/sec.`

.claude/skills/connector-runtime/SKILL.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Versioning
+
+- Configs carry `version: u64`. Local provider auto-increments if absent; HTTP provider trusts the server.
+- Restart on version change uses `restart_guard`.
+
+## State storage (`state.rs`)
+
+- File path: `{state_path}/source_{connector_key}.state`.
+- `save()` is crash-atomic: bytes are written to `{path}.tmp`, fsynced, then `rename`d into place. POSIX guarantees the rename, so a crash leaves either the previous good state or the new one - never a truncated file. On Unix, the parent directory is fsynced after the rename as a best-effort to make the rename itself durable. A `Mutex<()>` serializes concurrent saves on the same provider.
```
hubcio: first sentence conflates rename atomicity with crash-durability. POSIX `rename(2)` guarantees atomic namespace-link replacement (no observer sees the file as missing), but says nothing about whether the rename itself survives a power loss - that comes from the `sync_data()` on tmp at `state.rs:119` plus the post-rename `sync_parent_dir` at `state.rs:140`, which the next sentence already correctly mentions. suggest rewriting the first sentence as: ``rename(2)` is atomic in the namespace; the preceding `sync_data` on the tmp file plus the post-rename parent-dir fsync are what make a crash leave either the old or new content - never truncated.`

.claude/skills/connector-runtime/SKILL.md
```diff
@@ -0,0 +1,286 @@
...
+
+When extending the runtime with new per-batch logging, follow the same pattern - default to `debug!`, upgrade to `info!` when `verbose` is set. Don't introduce a third level. Plugin authors are encouraged to mirror this pattern internally for their own diagnostics.
+
+## Config provider rules
+
+### Local provider (`configs/connectors/local_provider.rs`)
+
+- Reads `*.toml` from `config_dir`. Skips hidden files (`.`) and `Cargo.toml`.
+- Filename convention: `{key}_{type}[_v{N}].toml` (e.g., `postgres_sink_v2.toml`).
+- One TOML deserialized into either `SinkConfig` or `SourceConfig`. Grouped by `key`; highest `version` wins unless `.active_versions.toml` (leading dot - it's a hidden file by convention) overrides. Built by `LocalConnectorsConfigProvider::active_versions_file_path` as `{config_dir}/.active_versions.toml`.
```
hubcio: reads contradictory on first pass - the previous bullet says the discovery walk skips hidden files, then this bullet describes `.active_versions.toml` as if it's found by the same walk. it's actually two code paths: walk skips dotfiles at `local_provider.rs:185`; `.active_versions.toml` is read via direct path through `active_versions_file_path` + `load_active_versions` at `:281-305`. one extra sentence resolves it - `the discovery walk skips dot-files; .active_versions.toml is read via direct path, not via the walk.`

.claude/skills/connector-sink/SKILL.md
```diff
@@ -0,0 +1,370 @@
...
+# + your client crate (reqwest, sqlx, mongodb, ...)
+```
+
+Run `cargo sort --no-format --workspace` after edits.
+
+## src/lib.rs skeleton
+
+Code reads top to bottom like a book. Public types first, then `impl`, then private helpers. Don't make the reader jump.
+
+```rust
```
hubcio: the embedded ~150-line `lib.rs` skeleton (this through ~line 205) dominates the skill. progressive-disclosure pattern says SKILL.md should be the hub; templates load on demand. suggest moving the skeleton to a sibling file like `.claude/skills/connector-sink/template.rs.md` and referencing it from SKILL.md behind an `if creating a new sink, read TEMPLATE.md` branch. reduces every-trigger load cost; the template is needed only when authoring.

.claude/skills/connector-source/SKILL.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Cargo.toml skeleton
+
+Identical to a sink's `Cargo.toml` (see `connector-sink` skill) - only the crate name suffix differs (`iggy_connector_<name>_source`) and the upstream client dep changes. Same `[lib] crate-type = ["cdylib", "lib"]`, same `cargo-machete` ignored list.
+
+## src/lib.rs skeleton
+
+Code reads top to bottom like a book. Public types first, `impl Source` afterward, helpers below.
+
+```rust
```
hubcio: same as connector-sink - ~157-line skeleton from here through ~line 184 dominates the file. move to a sibling `template.rs.md` and reference it conditionally. SKILL.md as hub, template on demand.

.claude/skills/connector-transform/SKILL.md
```diff
@@ -0,0 +1,263 @@
...
+- Returns `Ok(Some(msg))` to pass through (possibly modified), `Ok(None)` to **filter** (drop silently), `Err(e)` for actual failures.
+- `&TopicMetadata` is for routing-aware logic. Most transforms ignore it.
+
+## Steps to add a new transform
+
+### 1. Create the file
+
+`sdk/src/transforms/my_transform.rs`. Apache 2.0 license header. Code reads top to bottom - config struct, then transform struct, then `impl`, then helpers.
+
+```rust
```
hubcio, line 76: ~50-line skeleton through line 103. smaller than sink/source but same shape. consider moving to a sibling template file too, so the trio (sink/source/transform) all follow the same pattern.

.claude/skills/connector-sink/SKILL.md
```diff
@@ -0,0 +1,370 @@
...
+
+Always read the relevant exemplar end-to-end before writing - the patterns are intentionally repetitive across plugins.
+
+## Cargo.toml skeleton
+
+```toml
+# Apache 2.0 header (copy verbatim from any existing sink Cargo.toml)
+[package]
+name = "iggy_connector_<name>_sink"
+version = "0.4.1-edge.1"          # match the version most other sinks use; check by grepping `version =` in sibling Cargo.toml files
```
hubcio: hard-coded `version = "0.4.1-edge.1"` will rot on the next bump. better to instruct: `match the version in a sibling sink crate's Cargo.toml; do not hard-code`. otherwise every release this line drifts.

.claude/skills/connectors-overview/SKILL.md
```diff
@@ -0,0 +1,236 @@
+---
+name: connectors-overview
+description: Entry point and index for the Apache Iggy Connectors subsystem (`core/connectors/`). Load this first when working anywhere under `core/connectors/` - runtime, SDK, sinks, sources, or transforms. Routes to the focused per-area skills.
+---
+
+# Apache Iggy Connectors - Overview
+
+## What sinks and sources are
```
hubcio: `what sinks and sources are` plus the ASCII flow diagram below is tutorial-style for an audience that already knows what a sink is. trim to one sentence per definition, keep the file map and architecture pointers, drop the explanatory prose.

.claude/skills/connectors-overview/SKILL.md
```diff
@@ -0,0 +1,236 @@
+---
+name: connectors-overview
+description: Entry point and index for the Apache Iggy Connectors subsystem (`core/connectors/`). Load this first when working anywhere under `core/connectors/` - runtime, SDK, sinks, sources, or transforms. Routes to the focused per-area skills.
```
hubcio, line 3: a few descriptions in this skill set all start with `Apache Iggy connector(s)...` (overview, sink, source, runtime). fuzzy-match auto-router may pick the wrong skill for queries phrased around a single area. only `connector-sdk` and `connector-runtime` carry explicit `NOT for X` disambiguators. adding a one-line `use this when... NOT for ...` tail to each description in the overlapping skills would harden routing.

.claude/skills/connector-sink/SKILL.md
```diff
@@ -0,0 +1,370 @@
+---
+name: connector-sink
+description: Author a new Apache Iggy connector sink plugin under `core/connectors/sinks/<name>_sink/`. Sinks consume messages from Apache Iggy streams and push them to an external system (DB, search engine, HTTP endpoint, object store). Load when creating, modifying, or reviewing a sink crate.
```
hubcio: description overlaps with the other connector skills and lacks an explicit `NOT for X` disambiguator. would help the auto-router pick the right skill for queries like `modify how the sink consumes` vs `how the runtime polls iggy`. suggest adding a tail like `use for sink plugin authoring; NOT for runtime internals (see connector-runtime)`.

.claude/skills/connector-source/SKILL.md
```diff
@@ -0,0 +1,346 @@
+---
+name: connector-source
+description: Author a new Apache Iggy connector source plugin under `core/connectors/sources/<name>_source/`. Sources poll an external system (DB, API, queue) and produce messages into Apache Iggy streams. Load when creating, modifying, or reviewing a source crate.
```
hubcio: same description-overlap risk as the sink. suggest adding a tail disambiguator like `use for source plugin authoring; NOT for runtime internals (see connector-runtime)`.

.claude/skills/connector-transform/SKILL.md
```diff
@@ -0,0 +1,263 @@
+---
+name: connector-transform
+description: Add a new transform under `core/connectors/sdk/src/transforms/`. Transforms mutate, filter, or convert messages between source/encode and decode/sink. Use when adding field-level transforms (add/delete/update/filter), format conversions, envelope unwrapping, or similar config-driven message manipulations for Apache Iggy connectors.
```
hubcio: same description-overlap risk. add a tail disambiguator (e.g. `use for transform authoring; NOT for runtime internals or sink/source loops`) so the auto-router doesn't pick this for runtime/plugin queries.

.claude/skills/connector-testing/SKILL.md
```diff
@@ -0,0 +1,336 @@
+---
+name: connector-testing
+description: Write tests for Apache Iggy connectors - sinks, sources, transforms, SDK, runtime. Covers BDD unit-test naming, the canonical source state round-trip tests, and the real-infra integration test layout under `core/integration/tests/connectors/` with `testcontainers-modules` + `#[iggy_harness]`. Load when writing or reviewing tests in `core/connectors/` or `core/integration/tests/connectors/`.
```
hubcio: same description-overlap risk. add a tail disambiguator (e.g. `use for connector test authoring; NOT for runtime internals or non-connector test patterns`).

AGENTS.md
```diff
@@ -0,0 +1,355 @@
...
+| `markdownlint-cli` (`markdownlint`)           | Markdown linter                          | pre-commit, CI                                   |
+| `shellcheck`                                  | Shell linter                             | pre-commit, CI                                   |
+| `typos`                                       | Spelling                                 | pre-commit                                       |
+| `cargo about` + `license-checker-rseidelsohn` | Third-party license enumeration          | `scripts/ci/third-party-licenses.sh`             |
+| Apache RAT / `license-headers.sh`             | Apache 2.0 source-header check           | pre-commit, CI                                   |
+| `uv`                                          | Python lockfile / env                    | `scripts/ci/uv-lock-check.sh`                    |
+| `ruff`                                        | Python formatter + linter                | pre-commit (`foreign/python`, `bdd`, `examples`) |
+| `golangci-lint`                               | Go linter                                | pre-push (`foreign/go`)                          |
+
+## Core principles
```
hubcio: `core principles` block here through line 223 duplicates content already present in `.claude/skills/connectors-overview/SKILL.md` (LLM-slop, imports-at-top, `tokio::sync::Mutex`, top-down code, zero-clone, forward-compat config - lines 60-107 of the overview skill cover the same rules). AGENTS.md gets loaded every session; everything that's also in a skill costs tokens twice. similar story for lines 225-246 (connectors-specific invariants overlapping `connector-runtime` SKILL.md) and the 18-row tooling table at 147-171 which is reference, not preamble.

suggest pulling duplicated blocks out into their owning skill or a dedicated tooling/commands skill; target AGENTS.md <= ~200 lines holding only the architecture map, gotchas, canonical lint/test commands, and commit conventions.

hubcio (changes requested):
this part is more focused on markdowns review that you're adding in this PR. will be a long one, so brace yourself 🦀

- CLAUDE.md / AGENTS.md direction is best left as discussion. AGENTS.md is the converging cross-tool standard (codex, cursor, aider, windsurf all read it); inverting would optimize for one ecosystem at the cost of cross-tool discoverability. cheap middle ground: add a `CLAUDE.md` symlink pointing to `AGENTS.md` so tools that hardcode the claude filename find the same content. zero churn.

- STOP-and-ask sections per skill: each SKILL.md has scattered "don't do X" / "gotchas" / "hard rules", but no consolidated 3-5 bullet list of triggers that should always surface to the user before action (e.g. bumping SDK version, changing FFI signature, renaming the default consumer group, touching `state.rs` save path, changing wire-format `#[repr(C)]` layout). a tight top-of-file STOP section would be high signal for autonomous agents.

- community pointer from skills (optional): AGENTS.md lists discord + github discussions in the support section at the bottom; individual skills don't. a one-liner at the end of each SKILL.md pointing at AGENTS.md instead of duplicating urls would be enough.

- universal-rule duplication: beyond AGENTS.md vs skills, per-area skills also restate rules (license header, verification order `fmt`->`sort`->`clippy`->`test`, `tokio::sync::Mutex`, zero-clone, forward-compat config). pick one owner per topic - repo-wide in AGENTS.md, connector-wide in `connectors-overview`, per-area skills link not restate. follow-up.

- cross-skill content duplication: benchmark + metrics block in `connector-runtime` (lines 75-114, 213-230) approximately equals `connectors-overview` (lines 165-196). `SecretString` pattern in sink (240-267), source (228-246), overview (155-163). `verbose` flag re-explained in 4 places. canonicalize and link. follow-up.

- no TOC in long files: all 7 SKILL.md plus AGENTS.md are >100 lines (largest: connector-sink 370, AGENTS.md 355, connector-source 346, connector-testing 336). short TOC at top helps partial-read previews show scope.

- no CI / pre-commit governance for `.claude/skills/`: this PR introduces the first AI-agent tree in the repo. nothing references `.claude/` in `.gitignore`, `.pre-commit-config.yaml`, or `.github/workflows/`. follow-up worth scoping: validate YAML frontmatter on every `SKILL.md` (non-empty `name:` kebab-case <=64 chars, non-empty `description:` <=1024 chars), resolve markdown links inside `.claude/skills/**/*.md`, optionally grep-check that `path::symbol` references in skill files resolve to identifiers under `core/`. this catches both frontmatter drift and rename drift in one pass.

all above are follow-ups, not blockers.

BTW: in order to prepare this analysis, i used these resources:
  - https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices — Anthropic skill authoring
  best practices
  - https://code.claude.com/docs/en/skills — Claude Code Docs: Extend Claude with skills
  - https://claude.com/blog/how-claude-code-works-in-large-codebases-best-practices-and-where-to-start — Anthropic
   blog: large codebases
  - https://www.groff.dev/blog/implementing-claude-md-agent-skills — Groff: implementing CLAUDE.md + agent skills
  - https://alexop.dev/posts/claude-code-customization-guide-claudemd-skills-subagents/ — alexop.dev:
  customization guide
  - https://www.anthropic.com/engineering/equipping-agents-for-the-real-world-with-agent-skills — Anthropic
  engineering: equipping agents
  - https://github.com/anthropics/skills — anthropics/skills reference repo

core/connectors/runtime/src/manager/source.rs
```diff
@@ -148,18 +148,30 @@ impl SourceManager {
...
+            {
+                warn!(
+                    "Timed out waiting for source task to finish (plugin_id: {plugin_id}); aborting to prevent stale collisions with the next start."
+                );
+                handle.abort();
+                let _ = handle.await;
+            }
+        }
+
+        source::cleanup_sender(plugin_id);
```
hubcio: shutdown order regression. `cleanup_sender` runs last, so `handler_task` blocks on `recv_async()` (`source.rs:393`) until the 5s timeout fires and `handle.abort()` cancels it. `iggy_source_close` (`sdk/src/source.rs:121-128`) already `runtime.block_on`s the polling task before returning, so the "in-flight callbacks still find their entry" comment is no longer true. abort can land mid `file.save()` rename/fdatasync. fix: move `cleanup_sender` between the close call and the await loop so the receiver closes cleanly.

core/connectors/runtime/src/state.rs
```diff
@@ -42,72 +40,367 @@ pub enum StateStorage {
...
+            error!(
+                "Cannot rename {} -> {}. {error}.",
+                self.tmp_path.display(),
+                self.path.display()
+            );
+            cleanup_tmp(&self.tmp_path).await;
+            return Err(Error::CannotWriteStateFile);
+        }
+
+        sync_parent_dir(&self.path).await;
```
hubcio: `sync_parent_dir` failures only get a `warn!`, save returns Ok regardless. on EIO from parent fsync (fsyncgate semantics: dirty bit cleared, next fsync returns success on stale data) the rename dentry can revert on power loss, state file vanishes, `load()` maps NotFound to `Ok(None)` and the source silently restarts from scratch. either propagate the error (return Result from `sync_parent_dir`, convert to `CannotWriteStateFile`) or drop the call entirely. half-measure is worst-of-both.

core/connectors/runtime/src/sink.rs
```diff
@@ -540,67 +592,72 @@ async fn process_messages(
             current_message = transform.transform(topic_metadata, message)?;
         }
 
-        // The transform may return no message based on some conditions
+        // Filter contract: transform returning Ok(None) is an intentional drop.
         let Some(message) = current_message else {
+            metrics.inc_messages_filtered_with_labels(&labels.counter, 1);
```
hubcio: also at lines 605, 613, 621, 629, 637, 645, 656 and `source.rs:672`. each call goes through `Family::get_or_create(&labels)` which takes a `parking_lot::RwLock` read + HashMap lookup per message. for sieve transforms (`Ok(None)` per message) or error storms this fires on every iteration. accumulate a `u64` in the loop, single `inc_by(n)` after.

core/connectors/runtime/src/source.rs
```diff
@@ -407,72 +427,105 @@ pub(crate) async fn source_forwarding_loop(
...
+            match &state_storage {
+                StateStorage::File(file) => {
+                    if let Err(error) = file.save(state).await {
+                        let error_msg = format!(
+                            "Failed to save state for source connector with ID: {plugin_id}. {error}"
+                        );
+                        error!("{error_msg}");
+                        context.metrics.inc_errors_with_labels(&labels.counter);
+                        context.sources.set_error(&plugin_key, &error_msg).await;
+                        continue;
```
hubcio: also at line 461. `continue` on `producer.send` / `file.save` failure jumps past `total_elapsed` (line 511), `observe_stage_with_labels(&labels.stage_total, ...)` (line 512), and `benchmark::emit_source_event` (line 516). failed batches under-count the total histogram and miss the benchmark emit, contradicting the PR's observability goal. move total observation + benchmark emit to a tail block (or guarantee via Drop guard).

core/connectors/runtime/src/sink.rs
```diff
@@ -354,18 +375,20 @@ pub(crate) async fn consume_messages(
...
             &transforms,
             &decoder,
+            metrics,
+            labels,
         )
         .await
         {
-            Ok(count) => count,
+            Ok(result) => result,
             Err(error) => {
```
hubcio: symmetric gap to the source-side `continue` issue: `process_messages` failure here returns `Err` from `consume_messages`, jumping past `observe_stage_with_labels(&labels.stage_total, elapsed)` at line 405 and `benchmark::emit_sink_event` at line 421. failed sink batches under-count the total histogram and miss the benchmark emit. same fix - tail block / Drop guard.

core/connectors/runtime/src/benchmark.rs
```diff
@@ -0,0 +1,255 @@
...
+        connector_type = "source",
+        connector_key = connector_key,
+        stream = stream,
+        topic = topic,
+        batch_size = batch_size,
+        sent_count = sent_count,
+        decode_us = decode_us,
+        prepare_us = prepare_us,
+        iggy_send_us = iggy_send_us,
+        state_save_us = ?state_save_us,
```
hubcio: `state_save_us = ?state_save_us` Debug-formats the `Option<u64>`. under the new `LogFormat::Json` path the tracing-subscriber JSON layer serializes record_debug values as quoted strings (`"state_save_us":"Some(89)"` / `"None"`), so numeric consumers can't parse it. the test at lines 229 and 253 hardcodes the literal `"Some(89)"` / `"None"` strings, proving the current shape. emit two fields (bool present + `u64`) or always emit `u64` with sentinel.

core/connectors/runtime/src/sink.rs
```diff
@@ -540,67 +592,72 @@ async fn process_messages(
             current_message = transform.transform(topic_metadata, message)?;
```
hubcio: transform error here uses `?` and kills the whole consume loop, while the source side at `source.rs:651-664` catches the Err, bumps `errors`, drops the offending message, and continues. same `Transform` trait, same lifecycle, divergent semantics. pre-PR both used `?`; the PR changed source to drop-and-continue but left sink. pick one model (recommend source's per-message drop) and align both sides.

core/connectors/runtime/src/sink.rs
```diff
@@ -374,8 +397,12 @@ pub(crate) async fn consume_messages(
             }
         };
 
-        metrics.increment_messages_processed(plugin_key, processed_count as u64);
+        metrics.inc_messages_processed_with_labels(&labels.counter, processed_count as u64);
         let elapsed = start.elapsed();
+        let prepare_elapsed = elapsed.saturating_sub(ffi_elapsed);
```
hubcio: sink `Stage::Prepare` is derived as `elapsed.saturating_sub(ffi_elapsed)` and covers decode + transform + serialize lumped together. source explicitly splits `Stage::Decode` (line 430) from `Stage::Prepare` (line 445). same Prometheus `stage="prepare"` label means different things across `connector_type`, so dashboards built on it interpret differently per side. either split sink prepare into decode + serialize stages or rename source's prepare to match the lumped scope.

core/connectors/runtime/src/metrics.rs
```diff
@@ -38,6 +40,106 @@ pub enum ConnectorType {
...
+            Stage::IggySend => "iggy_send",
+            Stage::StateSave => "state_save",
+            Stage::Total => "total",
+        }
+    }
+}
+
+impl EncodeLabelValue for Stage {
+    fn encode(&self, encoder: &mut LabelValueEncoder) -> Result<(), std::fmt::Error> {
+        self.as_label().encode(encoder)
```
hubcio, line 90: label case is inconsistent within `iggy_connector_stage_duration_seconds`. `ConnectorType` uses `#[derive(EncodeLabelValue)]` which emits `stringify!(Ident)` (`"Source"` / `"Sink"`, PascalCase). `Stage` manually emits `"prepare"` / `"ffi"` / `"iggy_send"` / `"state_save"` (snake_case). end result is one metric with `connector_type="Source"` + `stage="prepare"` mixing cases. align both to snake_case (write a manual `EncodeLabelValue` for `ConnectorType` too, returning `"source"`/`"sink"`) or move `Stage` to PascalCase. either way pick one.

core/connectors/runtime/README.md
```diff
@@ -204,8 +236,15 @@ The runtime exposes Prometheus-compatible metrics via the `/metrics` endpoint wh
...
 - `iggy_connector_messages_consumed_total`: Messages consumed from Iggy (sink)
 - `iggy_connector_messages_processed_total`: Messages processed and sent to sink plugin
+- `iggy_connector_messages_filtered_total`: Messages intentionally dropped by transforms returning `Ok(None)`
 - `iggy_connector_errors_total`: Errors encountered
 
+### Stage Duration Histograms (labeled with `connector_key`, `connector_type`, `stage`)
+
+- `iggy_connector_stage_duration_seconds`: Per-batch processing stage duration
+
+Stages: `Prepare`, `Ffi`, `Decode`, `IggySend`, `StateSave`, `Total`. Histograms are always exposed (independent of `benchmark`); buckets cover 50us to 5s.
```
hubcio: docs say `Stages: Prepare, Ffi, Decode, IggySend, StateSave, Total` (PascalCase) but the code emits lowercase snake_case via `Stage::as_label`. operators writing PromQL from the readme will get zero matches. update the listing to `prepare, ffi, decode, iggy_send, state_save, total` (or change the encoder, see the metrics.rs comment).

core/connectors/sdk/src/api.rs
```diff
@@ -112,6 +112,9 @@ pub struct ConnectorStats {
     pub messages_consumed: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub messages_processed: Option<u64>,
+    #[serde(default)]
+    pub messages_filtered: u64,
```
hubcio: `messages_filtered: u64` is bare with `#[serde(default)]` while peer counters (`messages_produced` / `sent` / `consumed` / `processed`) are `Option<u64>` with `skip_serializing_if = "Option::is_none"`. mirrors the pre-existing shape of `errors: u64` so it's consistent with itself but inconsistent with the rest. consider `Option<u64>` so consumers don't see a `0` for connectors where the field is semantically N/A.

core/connectors/runtime/src/metrics.rs
```diff
@@ -38,6 +40,106 @@ pub enum ConnectorType {
...
+    }
+}
+
+const STAGE_BUCKETS_SECONDS: [f64; 13] = [
+    50e-6, 100e-6, 250e-6, 500e-6, 1e-3, 5e-3, 10e-3, 50e-3, 100e-3, 250e-3, 500e-3, 1.0, 5.0,
+];
```
hubcio, line 96: 13 explicit buckets + the implicit `+Inf` + `_sum` + `_count` is ~16 series per (`connector_key`, `stage`) pair, multiplied by 3 sink stages or 5 source stages, multiplied by operator-controlled `connector_key`. fine for typical fleets but blows up if `connector_key` is unbounded (per-user keys etc.). document the cardinality constraint in the connectors readme so operators size their TSDB appropriately.

core/connectors/runtime/src/state.rs
```diff
@@ -42,72 +40,367 @@ pub enum StateStorage {
...
+        assert_eq!(
+            tmp_path_for(&p),
+            PathBuf::from("/tmp/state/source_x.state.tmp")
+        );
+    }
+
+    #[test]
+    fn given_path_without_filename_when_tmp_path_built_should_yield_bare_tmp() {
+        let p = PathBuf::from("/");
+        assert_eq!(tmp_path_for(&p).file_name().unwrap(), ".tmp");
```
hubcio, line 364: `tmp_path_for(PathBuf::from("/"))` yields `/.tmp`. not reachable via the runtime's path-construction pattern (`format!("{state_path}/source_{key}.state")`) but a public direct-API footgun: handing the provider a root-only or directory-only path lands the tmp somewhere unexpected. cheap guard: `debug_assert!(path.file_name().is_some())` in `FileStateProvider::new`.

core/connectors/runtime/src/log.rs
```diff
@@ -31,38 +31,60 @@ use tracing_subscriber::EnvFilter;
...
+
+    match (telemetry_config.enabled, logging_config.format) {
+        (true, LogFormat::Text) => {
+            let (logger_provider, tracer_provider) = init_telemetry(telemetry_config, version);
+            let tracer = tracer_provider.tracer(telemetry_config.service_name.clone());
+            global::set_tracer_provider(tracer_provider);
+            global::set_text_map_propagator(TraceContextPropagator::new());
+            tracing_subscriber::registry()
+                .with(env_filter())
+                .with(tracing_subscriber::fmt::layer())
```
hubcio: the 4-arm `match (telemetry_config.enabled, logging_config.format)` is roughly 80% duplicated. build the fmt layer once as `Box<dyn Layer<Registry> + Send + Sync>` (via `Layer::boxed()`), then a 2-arm match on telemetry. saves about 30 lines without losing clarity.

core/connectors/runtime/src/configs/runtime.rs
```diff
@@ -142,6 +142,89 @@ pub struct ConnectorsRuntimeConfig {
...
+}
+
+impl Display for LogFormat {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            LogFormat::Text => write!(f, "text"),
+            LogFormat::Json => write!(f, "json"),
+        }
+    }
+}
```
hubcio: `LogFormat`'s manual `Display` + `FromStr` duplicates what `#[serde(rename_all = "lowercase")]` already encodes. swap to `#[derive(strum::Display, strum::EnumString)]` with `#[strum(serialize_all = "lowercase", ascii_case_insensitive)]` (the `ascii_case_insensitive` preserves the current `to_ascii_lowercase()` behavior in `FromStr`). repo already pulls strum in (`state.rs:23`). saves about 15 lines.

core/connectors/runtime/src/source.rs
```diff
@@ -50,9 +52,15 @@ use crate::{
 use iggy_connector_sdk::api::ConnectorStatus;
 use tokio::task::JoinHandle;
 
-pub static SOURCE_SENDERS: Lazy<DashMap<u32, Sender<ProducedMessages>>> = Lazy::new(DashMap::new);
+pub(crate) struct SourceSenderEntry {
+    pub(crate) sender: Sender<ProducedMessages>,
+    pub(crate) metrics: Arc<crate::metrics::Metrics>,
```
hubcio: `SourceSenderEntry.metrics: Arc<Metrics>` is only used to call `inc_errors_with_labels(&labels.counter)` from the FFI callback. pre-extract the owned `Counter` via `Family::get_or_create_owned(&labels)` at insert time (Counter is `Arc<AtomicU64>` per prometheus-client 0.24, cheap clone), cache it on the entry, drop the `Arc<Metrics>` field. FFI hot path becomes a single relaxed atomic inc, no RwLock, no HashMap lookup. same pattern fits `SinkLabels.counter`.

core/connectors/runtime/src/state.rs
```diff
@@ -42,72 +40,367 @@ pub enum StateStorage {
...
+        // fdatasync: skip mtime/atime sync. Parent dir is synced separately.
+        if let Err(error) = tmp.sync_data().await {
+            error!(
+                "Cannot sync temp state file: {}. {error}.",
+                self.tmp_path.display()
+            );
+            drop(tmp);
+            cleanup_tmp(&self.tmp_path).await;
+            return Err(Error::CannotWriteStateFile);
+        }
```
hubcio: the three error blocks for `write_all`, `sync_data`, `rename` all run `drop(tmp); cleanup_tmp(...).await; return Err`. collapse the first two (which share the tmp file lifecycle) into a private `async fn write_tmp(state: &[u8]) -> Result<(), Error>` that owns the file + cleanup on its own failures; keep the explicit `rename` + `cleanup_tmp` on rename failure as-is. cuts ~10 lines without changing the cleanup graph.

hubcio (changes requested):

hubcio (approved):

numinnex (approved):

PR #3322, opened 2026-05-26
ci: broaden dependabot coverage, add cooldown and PR caps

hubcio:
Was Rust workspace + GitHub Actions only, leaving 5 foreign SDKs,
17 Dockerfiles, and the C++ Bazel module on manual updates - stale
transitive deps and base image CVEs went unsurfaced.

Adds uv, npm, gomod, nuget, gradle, bazel, and docker. Cargo entry
also picks up /foreign/cpp and /foreign/python (workspace-excluded
crates with their own Cargo.lock). uv over pip - repo uses uv.lock
natively. nuget covers foreign + examples csharp; docker covers
production, bdd, foreign.

bdd/python/requirements.txt removed; pyproject+uv.lock is canonical
and pins had drifted (pytest>=7 vs >=9.0.3, orphan maturin).

Tunes existing entries: cooldown.default-days 7 (skip yanked
first-day releases), open-pull-requests-limit 3 (headroom so
ungrouped majors don't starve grouped minor-patch on one slot),
schedule.day stagger Mon-Fri (avoid Monday CI + reviewer herd),
per-ecosystem labels.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,14 +18,26 @@
 version: 2
 updates:
   - package-ecosystem: "cargo"
-    directory: "/"
+    directories:
+      - "/"
+      - "/foreign/cpp"
+      - "/foreign/python"
+      # TODO(hubcio): uncomment once #3235 is merged
+      # - "/foreign/php"
     schedule:
       interval: "weekly"
+      day: "monday"
     rebase-strategy: "disabled"
+    open-pull-requests-limit: 3
     commit-message:
       prefix: "chore(deps): "
     reviewers:
       - "apache/iggy-committers"
+    labels:
+      - "dependencies"
+      - "rust"
+    cooldown:
+      default-days: 7
     groups:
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -38,11 +50,219 @@ updates:
...
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+
+  - package-ecosystem: "nuget"
+    directories:
+      - "/foreign/csharp"
+      - "/examples/csharp"
```
slbotbm, line 159: bdd for csharp?

slbotbm (changes requested):

.github/dependabot.yml
hubcio: there's no actual csharp files in this directory
image

krishvishal (approved):

spetz (approved):

Issue #3315, opened 2026-05-23
refactor(server-ng): remove use server::* imports from core/server-ng


jaya-traversal:
\`core/server-ng\` imports several utilities from \`core/server\`, the legacy binary it is meant to replace. This creates a compile-time dependency that prevents server-ng from ever building independently.

The imports fall into five clusters across \`main.rs\`, \`bootstrap.rs\`, and \`server_error.rs\`:

| File | Import |
|------|--------|
| \`main.rs\` | \`server::bootstrap::create_shard_executor\` |
| \`main.rs\` | \`server::diagnostics::print_*_info\` (3 fns) |
| \`main.rs\` | \`server::log::logger::Logging\` |
| \`bootstrap.rs\` | \`server::bootstrap::create_directories\` |
| \`bootstrap.rs\` | \`server::log::logger::Logging\` |
| \`bootstrap.rs\` | \`server::streaming::partitions::storage::{load_consumer_group_offsets, load_consumer_offsets}\` |
| \`bootstrap.rs\` | \`server::streaming::segments::storage::create_segment_storage\` |
| \`bootstrap.rs\` | \`server::bootstrap::load_segments\` |
| \`server_error.rs\` | \`server::server_error::LogError\` |

Each cluster is independently removable with no behaviour change. The storage and partition log clusters depend on #3298 landing first, since that PR moves the relevant types out of \`core/server\`.

hubcio:
@numinnex what we're doing with this one?

numinnex:
Hi, the dependency on the `server` crate in `server-ng` is temporarely there due to code reuse of some core mechanism such as segment loading or consumer offset loading. We wanted to get rid of them together when we replace `server` binary with the `server-ng` one, but if you have an idea how to improve state of the affairs in a quick way, we are down to merge it.

jayakasadev:
Thanks for the context, @numinnex.

A PR is coming shortly that cleans up `main.rs` entirely — it moves `create_shard_executor` and the io_uring diagnostics into a new `core/server-runtime` crate shared by both `core/server` and `core/server-ng`. No behaviour change.

`bootstrap.rs` and `server_error.rs` still have remaining `use server::*` imports. The storage/partition loading ones (`load_segments`, `create_segment_storage`, `load_consumer_offsets`) need #3298 to land first since that PR moves those types out of `core/server`. The rest (`create_directories`, `Logging`, `LogError`, `ShardAllocator`) can follow in separate PRs after Phase 1 merges.

[...]

PR #3331, opened 2026-05-28
chore(server-ng): extract compio executor and io_uring diagnostics into shared crate

jayakasadev:

Part of #3315 — Phase 1.

## What changed?

`core/server-ng/src/main.rs` imported `create_shard_executor` and three io_uring diagnostic functions directly from `core/server`. This created a compile-time dependency on the binary being replaced.

A new `core/server-runtime` crate now owns these two things:
- `create_shard_executor` — builds the compio io_uring executor used to spin up shard threads
- `diagnostics` — `print_invalid_io_uring_args_info`, `print_locked_memory_limit_info`, `print_io_uring_permission_info`

Both `core/server` and `core/server-ng` depend on `server-runtime`. No behaviour change. `main.rs` is left with zero `use server::*` imports after this lands.

The remaining clusters in `bootstrap.rs` and `server_error.rs` are follow-on work. The storage/partition loading imports depend on #3298 landing first.

## Local Execution

- `cargo clippy --workspace --no-deps --tests -- -D warnings`
- `cargo test --workspace --lib`
- `cargo build --workspace`
- `cargo machete`
- `cargo sort --workspace --check`

## AI Usage

1. Claude (Anthropic, Sonnet 4.6).
2. Generated function — none; code was written locally. Claude assisted with scoping and PR description.
3. Verified via the cargo checks above.
4. Yes.


hubcio (changes requested):
instead of creating a new crate, please put the implementation in `server_common` crate and use it in both `core/server-ng` and legacy `core/server`

jayakasadev (commented):
SGTM

core/server_common/src/executor.rs
```diff
@@ -0,0 +1,54 @@
...
+/// Falling back to default flags would silently degrade shard performance -
+/// do not add a retry with reduced flags here.
+pub fn create_shard_executor() -> Result<Runtime, std::io::Error> {
+    // TODO: The event interval tick, could be configured based on the fact
+    // How many clients we expect to have connected.
+    // This roughly estimates the number of tasks we will create.
+    let mut proactor = compio::driver::ProactorBuilder::new();
+
+    proactor
+        .capacity(4096)
```
hubcio: this hardcodes `.capacity(4096)` and drops the `IGGY_SHARD_RUNTIME_CAPACITY` env read that the old `create_shard_executor` had (it was `std::env::var("IGGY_SHARD_RUNTIME_CAPACITY").ok().and_then(|v| v.parse::<u32>().ok()).unwrap_or(4096)`). the var is still registered in the env provider and the integration harness still sets it to `256`, so the knob is now silently dead. every io_uring ring in both binaries reserves 4096 entries again, which is exactly the memlock regression the old comment warned about - multi-node tests under an ~8mb memlock budget hit ENOMEM at ring setup and panic. this is the behaviour change the PR says doesn't exist. fix: restore the env read with `unwrap_or(4096)` here, and bring back the explanatory comment.

hubcio (changes requested):
besides that one finding:
- `core/configs/src/configs_impl/typed_env_provider.rs:57-58`: the comment there still says `IGGY_SHARD_RUNTIME_CAPACITY` is read by `server::bootstrap::create_shard_executor`. that's now doubly stale - the reader moved to `server_common` and (per the executor change) no longer reads it at all. update the pointer once the env read is restored

core/server_common/src/executor.rs
jayakasadev: that's my mistake. correcting

jayakasadev:
On coverage: even with new tests on `format_limit` + `parse_kernel_version` overflow, the 3 eprintln printers stay uncovered

/ready

hubcio (dismissed):

krishvishal (dismissed):

hubcio:
@jayakasadev please fix typo

hubcio (approved):

krishvishal (approved):

PR #3367, opened 2026-05-29
ci(rust): split nextest into 3 partitions instead of 2

hubcio:
After the Doris connector merge both test-1 and test-2 jobs
ran ~15 minutes each. Add a third hash partition to spread
the load and cut wall-clock time per node.


hubcio:
Ffr now closing this because ~15 min vs ~13 min doesn't seem like that big gain. i'll reopen when there's a need.

PR #3368, opened 2026-05-29
feat(php): add SDK examples component

### LINKED ISSUE #3311: PHP SDK: add examples and examples-php component

Every other SDK has examples wired through repository component detection, but the PHP SDK does not yet have `examples/php/` examples or an `examples-php` component entry. This is fine as a follow-up and not a merge blocker, but should be tracked so it does not get lost.

## Scope

- Add `examples/php/` examples following the shape and coverage expectations of the other SDK example directories.
- Add an `examples-php` component entry in `.github/config/components.yml`.
- Wire the component paths so PHP example changes trigger the expected CI tasks.
- Ensure the examples component depends on the PHP SDK and any required Rust server / CI infrastructure components.

## References

- `examples/python/`, `examples/node/`, `examples/go/`, `examples/java/`, `examples/csharp/`, `examples/rust/`
- `.github/config/components.yml`

Related: #3302
Initial PHP SDK work: #3235

countradooku:
## Summary

- add PHP SDK examples under `examples/php` for getting-started and basic producer/consumer flows
- add an `examples-php` component entry with PHP SDK, Rust SDK, Rust server, and CI infrastructure dependencies
- wire PHP examples into `_test_examples.yml` and `scripts/run-examples-from-readme.sh`

Closes #3311.

## Validation

- `php -l` for all PHP example files
- `bash -n scripts/run-examples-from-readme.sh`
- `git diff --check`
- `./scripts/ci/markdownlint.sh examples/php/README.md`
- `./scripts/ci/license-headers.sh --check`
- `cargo build --locked --bin iggy-server`
- `cargo build` in `foreign/php`
- pre-push hook: cargo clippy passed; Go lint suites passed; Java/C# skipped as no relevant files changed

### DIFF
--- a/examples/php/README.md
+++ b/examples/php/README.md
@@ -34,5 +34,6 @@ php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_p
 php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" basic/consumer.php
 ```
 
-The examples use `IGGY_CONNECTION_STRING` when it is set. Otherwise they connect
-to `iggy+tcp://iggy:iggy@127.0.0.1:8090`.
+The examples use `IGGY_CONNECTION_STRING` when it is set. Otherwise they build
+`iggy+tcp://iggy:iggy@127.0.0.1:8090` from `IGGY_HOST`, `IGGY_PORT`,
+`IGGY_USERNAME`, and `IGGY_PASSWORD`, which can be set individually.
--- a/examples/php/README.md
+++ b/examples/php/README.md
@@ -1,0 +1,38 @@
+# Iggy PHP Examples
+
+This directory contains sample applications that show how to use the Apache
+Iggy PHP SDK extension.
+
+## Running Examples
+
+Start the server from the repository root:
+
+```bash
+cargo run --bin iggy-server -- --fresh --with-default-root-credentials
+```
+
+From `examples/php`, build the PHP extension and point PHP at it:
+
+```bash
+(cd ../../foreign/php && cargo build)
+export PHP_IGGY_EXTENSION="$(pwd)/../../foreign/php/target/debug/libiggy_php.so"
+```
+
+On macOS, use `../../foreign/php/target/debug/libiggy_php.dylib` instead.
+
+### Getting Started
+
+```bash
+php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" getting-started/producer.php
+php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" getting-started/consumer.php
+```
+
+### Basic Usage
+
+```bash
+php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" basic/producer.php
+php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" basic/consumer.php
+```
+
+The examples use `IGGY_CONNECTION_STRING` when it is set. Otherwise they connect
+to `iggy+tcp://iggy:iggy@127.0.0.1:8090`.
--- a/examples/php/basic/producer.php
+++ b/examples/php/basic/producer.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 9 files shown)


scripts/run-examples-from-readme.sh
```diff
@@ -284,6 +284,49 @@ run_python_examples() {
...
+        local extension_candidate
+        for extension_candidate in \
+            target/debug/libiggy_php.so \
+            target/debug/libiggy_php.dylib \
+            target/debug/iggy_php.dll \
+            foreign/php/target/debug/libiggy_php.so \
+            foreign/php/target/debug/libiggy_php.dylib \
+            foreign/php/target/debug/iggy_php.dll; do
+            if [ -f "${extension_candidate}" ]; then
+                PHP_IGGY_EXTENSION="${extension_candidate}"
```
hubcio: this sets `PHP_IGGY_EXTENSION` to a repo-root-relative path, but `run_language_examples` cds into `examples/php` before the readme command evals `php -d extension="${PHP_IGGY_EXTENSION:-...}"`. the var is now set so the `:-` fallback is skipped, and the relative path resolves against `examples/php/` -> file missing -> extension fails to load -> `Iggy\Client` undefined -> fatal. the `[ ! -f ]` check above runs while still at repo root so it passes and masks it. CI is fine since the workflow exports an absolute `realpath`, but `./scripts/run-examples-from-readme.sh --language php` run locally hits it. fix: `PHP_IGGY_EXTENSION="$(pwd)/${extension_candidate}"`. also the `iggy_php.dll` candidates are dead here (no windows runner) and can be dropped.

examples/php/src/common.php
```diff
@@ -0,0 +1,96 @@
...
+function print_polled_messages(Client $client, string $stream, string $topic, int $count): void
+{
+    $messages = $client->pollMessages($stream, $topic, 0, PollingStrategy::first(), $count, true);
+
+    if (count($messages) < $count) {
+        throw new RuntimeException("Expected {$count} messages, received " . count($messages));
+    }
+
+    foreach ($messages as $message) {
+        if (!$message instanceof ReceiveMessage) {
```
hubcio: `pollMessages()` always returns an array of `ReceiveMessage`, so this guard can never be false and the `continue` is unreachable. drop it and iterate directly.

examples/php/src/common.php
```diff
@@ -0,0 +1,96 @@
...
+}
+
+function ensure_stream_and_topic(Client $client, string $stream, string $topic): void
+{
+    if ($client->getStream($stream) === null) {
+        $client->createStream($stream);
+    }
+
+    if ($client->getTopic($stream, $topic) === null) {
+        $client->createTopic($stream, $topic, 1, null, null, null, null);
```
hubcio: the 4 trailing `null`s match the stub defaults, so `createTopic($stream, $topic, 1)` is equivalent and cleaner. the python example skips them too.

scripts/run-examples-from-readme.sh
```diff
@@ -284,6 +284,49 @@ run_python_examples() {
...
+    }
+
+    run_language_examples \
+        "PHP" \
+        "examples/php" \
+        "README.md" \
+        "^php " \
+        "" \
+        "" \
+        0 \
```
hubcio, line 324: non-blocking and the other runners pass `0` too, so fine to leave - but `0` means no per-command timeout, and `connect()` retries unbounded by default. if the server dies after startup a stalled connect rides all the way to the 60-min job timeout instead of failing fast. python passes `10`. could be a small repo-wide follow-up.

examples/php/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+### Basic Usage
+
+```bash
+php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" basic/producer.php
+php -d extension="${PHP_IGGY_EXTENSION:-../../foreign/php/target/debug/libiggy_php.so}" basic/consumer.php
+```
+
+The examples use `IGGY_CONNECTION_STRING` when it is set. Otherwise they connect
```
hubcio: the examples also fall back to `IGGY_HOST`, `IGGY_PORT`, `IGGY_USERNAME` and `IGGY_PASSWORD` when `IGGY_CONNECTION_STRING` is unset, but only the connection string is documented here. worth a line so people know the individual vars work too.

examples/php/basic/producer.php
```diff
@@ -0,0 +1,37 @@
...
+declare(strict_types=1);
+
+require_once __DIR__ . '/../src/common.php';
+
+$stream = 'php-basic-stream';
+$topic = 'php-basic-topic';
+$client = iggy_client();
+
+ensure_stream_and_topic($client, $stream, $topic);
+send_payloads($client, $stream, $topic, [
```
hubcio: small consistency thing - the getting-started producer builds payloads via `example_payloads()` while this one uses an inline array. could call `example_payloads('basic', 3)` here so both follow one style.

hubcio (changes requested):

hubcio (approved):

spetz (approved):

PR #3373, opened 2026-05-29
feat(sdk): support sending raw requests for custom commands

spetz:
The SDK only exposed the commands it explicitly wraps, with no public
way to issue an arbitrary one. This left no room for extending a
deployment with custom server commands without modifying the SDK.

Expose two inherent methods on IggyClient: send_binary_request(code,
payload) for the binary transports (TCP, QUIC, WebSocket) and
send_http_request(method, path, body) for HTTP, each returning the
raw response bytes. Both reuse the live authenticated connection,
framing, and reconnect logic. Calling the binary method on HTTP or
the reverse returns FeatureUnavailable.

These methods stay off the Client trait to avoid burdening every
transport with calls meaningless to it. The HttpMethod enum lives in
iggy_common so the trait carries no concrete HTTP client dependency.


core/sdk/src/clients/client.rs
```diff
@@ -182,6 +184,31 @@ impl IggyClient {
     pub async fn get_connection_info(&self) -> ConnectionInfo {
         self.client.read().await.get_connection_info().await
     }
+
+    /// Send a raw binary command (`code` + serialized `payload`) and return the
+    /// raw response. Binary transports only; HTTP yields `FeatureUnavailable`.
+    pub async fn send_binary_request(&self, code: u32, payload: Bytes) -> Result<Bytes, IggyError> {
```
hubcio, line 211: heads up for the `vsr`/server-ng path: this forwards any `code` straight into `send_raw_with_response` with no validation. under `--features vsr`, passing `LOGIN_REGISTER_CODE`/`LOGIN_REGISTER_WITH_PAT_CODE` on an already-logged-in client hits the Register branch of `encode_request_header`, which calls `register_request_id()` -> that asserts `!register_consumed` and `!is_bound()`, both false after login, so it panics. the panic fires while holding the `consensus_session` `std::sync::Mutex` (websocket encodes inline so it unwinds the caller; tcp/quic encode inside `tokio::spawn` so it first surfaces as `TcpError`/`QuicError`), which poisons the mutex - so every later `reset_vsr_session`/`bind`/`disconnect` hits `.lock().expect("...poisoned")` and re-panics, leaving the client unusable on all binary transports. classic framing is unaffected (no consensus session there). raw `LOGOUT_USER_CODE` is the same class: no panic, but it skips `reset_vsr_session`/state reset so the sdk keeps a dead session. simplest fix is to reject login/logout control codes here (there's an `is_login_register_code` helper already), and/or have `register_request_id`/`encode_request_header` return `Err` instead of asserting now that this is a public caller-controlled entry point.

core/sdk/src/clients/client.rs
```diff
@@ -182,6 +184,31 @@ impl IggyClient {
     pub async fn get_connection_info(&self) -> ConnectionInfo {
         self.client.read().await.get_connection_info().await
     }
+
+    /// Send a raw binary command (`code` + serialized `payload`) and return the
```
hubcio: the doc only mentions HTTP yielding `FeatureUnavailable`, but under `--features vsr` this can also fail at encode time: a genuinely-custom/unknown code is rejected by `operation_for_code` with `InvalidCommand` (a known-but-replicated code with no mapping gets `UnknownReplicatedCommand`). so the "send custom commands without modifying the sdk" goal only holds on the classic protocol - under server-ng the encoder is closed-world and only carries codes already in the dispatch table. worth spelling that restriction out here.

core/sdk/src/clients/client.rs
```diff
@@ -182,6 +184,31 @@ impl IggyClient {
...
+    /// raw response. Binary transports only; HTTP yields `FeatureUnavailable`.
+    pub async fn send_binary_request(&self, code: u32, payload: Bytes) -> Result<Bytes, IggyError> {
+        match &*self.client.read().await {
+            ClientWrapper::Tcp(client) => client.send_raw_with_response(code, payload).await,
+            ClientWrapper::Quic(client) => client.send_raw_with_response(code, payload).await,
+            ClientWrapper::WebSocket(client) => client.send_raw_with_response(code, payload).await,
+            ClientWrapper::Http(_) | ClientWrapper::Iggy(_) => Err(IggyError::FeatureUnavailable),
```
hubcio, line 219: minor consistency: `Iggy(_)` is lumped with `Http(_)` into `FeatureUnavailable`, but `ClientWrapper::Iggy` wraps a nested `IggyClient` that has these same inherent methods, so it could just recurse (`Iggy(client) => client.send_binary_request(...).await`). in practice `Iggy` is never constructed today (the builder and `from_connection_string` only produce concrete transports), so this is dead-path tidiness rather than a live bug. note `send_http_request` hides `Iggy` under the `_` arm just below, so a fix would need an explicit arm there too.

core/integration/tests/sdk/raw.rs
```diff
@@ -0,0 +1,90 @@
...
+#[iggy_harness(test_client_transport = [Tcp, Quic, Http, WebSocket])]
+async fn given_authenticated_client_when_sending_raw_request_should_round_trip(
+    harness: &TestHarness,
+) {
+    let client = harness.root_client().await.unwrap();
+    assert_raw_round_trip(&client).await;
+}
+
+#[cfg(feature = "vsr")]
+#[iggy_harness(test_client_transport = [Tcp, WebSocket])]
```
hubcio, line 36: the `vsr` arm only ever sends `PING_CODE` and `GET_STATS_CODE`, both non-replicated, so it never exercises a replicated/custom code or the login/logout-code path under vsr. a case that sends a login/register code on a bound client (should return a clean `Err`, not panic) plus an unknown-code reject would actually cover the raw path on server-ng.

core/common/src/types/http.rs
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+use strum::{Display, EnumString, IntoStaticStr};
+
+/// HTTP request method, kept dependency-free so `iggy_common` carries no
+/// concrete HTTP client type.
+#[derive(Clone, Copy, Debug, Default, Display, PartialEq, Eq, EnumString, IntoStaticStr)]
```
hubcio, line 23: `Display`, `EnumString` (FromStr) and `Default` aren't used anywhere in-repo right now - only `IntoStaticStr` via `<&str>::from` is. fine to keep them since this is prelude-exported public api, just flagging they're currently unexercised. `#[default] Get` only matters to downstream callers; a one-line "defaults to GET" doc would remove any surprise.

core/common/src/types/http.rs
```diff
@@ -0,0 +1,40 @@
...
+pub enum HttpMethod {
```
hubcio, line 24: follow-up, not for this pr: there's already an `HttpMethod` in `core/connectors/sinks/http_sink/src/lib.rs` (serde-based, `#[default] Post`, no `Options`). two enums now coexist in different crates. not trivially mergeable (this one is intentionally dependency-free, that one is serde config), but worth a dedup later if `iggy_common` ever gains serde.

hubcio (changes requested):

core/common/src/types/http.rs
spetz: Good point for the future refactoring.

hubcio (approved):

numinnex (approved):

PR #3374, opened 2026-05-29
fix(server): distribute partitions evenly in cooperative rebalancing

### LINKED ISSUE #3334: Bug: cooperative rebalancing only moves 1 partition to new consumers
### Bug description

Think I found a bug in `rebalance_cooperative()`. When new consumers join, each idle member receives at most 1 partition per over-assigned source because the idle member is removed from the eligible list after each assignment. Excess partitions have nowhere to go.

  **Reproduce:** 1 consumer with 4 partitions. Add a second consumer.

  **Expected:** 2 partitions each
  
  **Actual:** 3 and 1

  **Root cause** in `core/server/src/metadata/consumer_group.rs`:

  ```rust
  for partition_id in revocable.into_iter().take(excess_count) {
      let idle_id = idle_slab_ids.remove(0);  // removes after 1 assignment
      // ...
  }
  
  First over-assigned member transfers 1 partition, idle member gets removed from list, remaining excess partitions have nowhere to go.

  Fix: Two-pass approach. First collect all excess partitions, then distribute round-robin:
  
  // Pass 1: Collect all excess
  let mut all_excess: Vec<(PartitionId, usize)> = Vec::new();
  for &mid in &member_ids {
      // ... calculate excess ...
      for partition_id in revocable.into_iter().take(excess_count) {
          all_excess.push((partition_id, mid));
      }
  }

  // Pass 2: Distribute round-robin to idle members
  for (i, (partition_id, source_mid)) in all_excess.into_iter().enumerate() {
      let idle_id = idle_slab_ids[i % idle_slab_ids.len()];
      // ... mark revocation ...
  }                                                                                                                      

  ```

**Happy to provide PR**

### Deployment

Compiled from source

### Versions

v0.8.1-edge.1

### Hardware / environment

Bare metal, Ubuntu 22.04, 32-core/128GB    

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

geertbrit:
## Summary
- Fix cooperative rebalancing to distribute partitions evenly across idle members
- Add 6 new integration tests verifying balanced distribution in various scenarios

## Problem
When multiple consumers joined a consumer group, the first idle member received most excess partitions while others starved. Example: 16 partitions held by 1 member + 15 new members → uneven distribution.

## Solution
Two-pass distribution approach:
1. **Pass 1:** Collect ALL excess partitions from over-assigned members
2. **Pass 2:** Distribute round-robin using `idle_slab_ids[i % idle_slab_ids.len()]`

Example: 16 partitions held by 1 member, 15 idle members join
- Before: first idle member gets many, others get 0-1
- After: each of 16 members gets exactly 1 partition

## Test plan
- [x] All 95 existing consumer group partition assignment tests pass
- [x] 6 new tests added covering balanced distribution scenarios:
  - 16 partitions → 16 consumers (exact benchmark scenario)
  - 12 partitions → 4 concurrent joins
  - 10 partitions → 4 members (uneven ratio with remainder)
  - 16 partitions with 2 over-assigned → 8 total members
  - 24 partitions → 8 members (large scale)
  - Member churn (leave + rejoin)

Closes #3334

### DIFF
--- a/core/integration/tests/data_integrity/verify_consumer_group_partition_assignment.rs
+++ b/core/integration/tests/data_integrity/verify_consumer_group_partition_assignment.rs
@@ -2263,6 +2263,51 @@ async fn join_cg(client: &IggyClient) {
         .unwrap();
 }
 
+fn assert_balanced_partition_distribution(cg: &ConsumerGroupDetails, expected_total: u32) {
+    let member_count = cg.members.len() as u32;
+    assert!(member_count > 0, "No members in consumer group");
+
+    let total: u32 = cg.members.iter().map(|m| m.partitions_count).sum();
+    assert_eq!(
+        total, expected_total,
+        "Total partitions mismatch. Expected {expected_total}, got {total}. Members: {:?}",
+        cg.members
+    );
+
+    let fair_share = expected_total / member_count;
+    let remainder = expected_total % member_count;
+
+    let min_expected = fair_share;
+    let max_expected = if remainder > 0 {
+        fair_share + 1
+    } else {
+        fair_share
+    };
+
+    for member in &cg.members {
+        assert!(
+            member.partitions_count >= min_expected && member.partitions_count <= max_expected,
+            "Member {} has {} partitions, expected between {} and {}. \
+             Distribution is unbalanced! Members: {:?}",
+            member.id,
+            member.partitions_count,
+            min_expected,
+            max_expected,
+            cg.members
+        );
+    }
+
+    let counts: Vec<u32> = cg.members.iter().map(|m| m.partitions_count).collect();
+    let min_count = *counts.iter().min().unwrap();
+    let max_count = *counts.iter().max().unwrap();
+    assert!(
+        max_count - min_count <= 1,
+        "Partition distribution variance too high: min={min_count}, max={max_count}. \
+         Difference should be at most 1. Members: {:?}",
+        cg.members
+    );
+}
+
 #[iggy_harness(test_client_transport = [Tcp, WebSocket, Quic], server(
     heartbeat.enabled = true,
     heartbeat.interval = "2s",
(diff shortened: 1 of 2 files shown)


core/server/src/metadata/consumer_group.rs
```diff
@@ -137,14 +137,21 @@ impl ConsumerGroupMeta {
...
+        // Pass 1: Collect excess partitions from over-assigned members
         let member_ids: Vec<usize> = self.members.iter().map(|(id, _)| id).collect();
         let mut members_with_remainder = remainder;
+        let mut all_excess: Vec<(PartitionId, usize)> = Vec::new();
 
         for &mid in &member_ids {
-            if idle_slab_ids.is_empty() {
-                break;
-            }
             let effective_count = self
```
hubcio: the per-member `pending` set gets built twice for each over-assigned member - once here in the `effective_count` closure, then again below in the `revocable` closure (around line 186). same `HashSet<PartitionId>` from the same `pending_revocations`. build it once before the `max_allowed` branch and reuse for both, saves a few lines and one alloc per over-assigned member. pre-existing, just surfaced by the restructure.

core/integration/tests/data_integrity/verify_consumer_group_partition_assignment.rs
```diff
@@ -2263,6 +2263,51 @@ async fn join_cg(client: &IggyClient) {
...
+             Distribution is unbalanced! Members: {:?}",
+            member.id,
+            member.partitions_count,
+            min_expected,
+            max_expected,
+            cg.members
+        );
+    }
+
+    let counts: Vec<u32> = cg.members.iter().map(|m| m.partitions_count).collect();
```
hubcio: this variance check (lines 2300-2308) is redundant. the per-member range assert just above (2287-2298) already bounds every member to `[fair_share, fair_share + (remainder>0?1:0)]`, a span of at most 1, so `max_count - min_count <= 1` always holds. can drop the whole block plus the `counts` vec.

core/integration/tests/data_integrity/verify_consumer_group_partition_assignment.rs
```diff
@@ -3267,3 +3312,536 @@ async fn should_not_complete_other_members_revocations_on_leave(harness: &TestHa
...
+#[iggy_harness(test_client_transport = [Tcp, WebSocket, Quic], server(
+    heartbeat.enabled = true,
+    heartbeat.interval = "2s",
+    tcp.socket.override_defaults = true,
+    tcp.socket.nodelay = true
+))]
+async fn should_distribute_16_partitions_evenly_across_16_consumers(harness: &TestHarness) {
+    let root_client = harness.root_client().await.unwrap();
+
+    root_client.create_stream(STREAM_NAME).await.unwrap();
```
hubcio: all 6 new tests inline this same ~21-line create stream + topic + consumer group block, only the partition count differs (16/12/10/16/24/16). also at 3438, 3525, 3610, 3691, 3766. the existing `setup_stream_topic_cg` helper hardcodes `PARTITIONS_COUNT` so it can't be reused here. a `setup_stream_topic_cg_with_partitions(client, n)` with the existing helper delegating to it would cut ~120 lines, and helper-based setup is already the convention in this file.

hubcio (dismissed):
overall very well structured PR, everything looks solid. below comments are just possible simplifications.

hubcio:
@geertbrit let us know if you want to fix these comments that I posted at `May 29` or just ignore.

geertbrit:
@hubcio pushed fixes

hubcio (approved):

hubcio:
@geertbrit looks like there is one failing testcase, can you check? https://github.com/apache/iggy/actions/runs/26759334334/job/78869626276?pr=3374

geertbrit:
@hubico Thanks for flagging. I investigated and this is a pre-existing flaky test: t also fails on master. The test (expiry_respects_consumer_offset) uses a simple Consumer, not a ConsumerGroup, so it's unrelated to the rebalancing changes in this PR.

The root cause is a race condition: the test sends 300 messages one-at-a-time with a 2s expiry and 100ms cleaner interval, so early messages expire before the send loop completes.

hubcio (approved):

numinnex (approved):

spetz (approved):

Issue #2940, opened 2026-03-16
Discussion: how should sinks handle partial external writes and replay-safe progress?


avirajkhare00:
While reviewing the MongoDB sink follow-up, I ended up at a broader connector-runtime question that likely applies beyond MongoDB.

Some sinks can partially commit a batch in the external system and still return an error. That creates a hard tradeoff between:
- not reporting full success when only part of the batch was written
- not getting stuck replaying already-written messages forever
- keeping progress and restart behavior understandable

A few design questions seem worth discussing more generally:
- should sinks be able to report partial success to the runtime
- should runtime progress be able to advance partially
- should deterministic external write identity be a standard requirement for sinks
- when is duplicate detection enough, and when do we need a stronger contract

I am not proposing a concrete runtime change in this issue yet.

I have a small PR separately to align the MongoDB sink README and test naming with the current merged behavior. This issue is about the larger contract and design question behind partial writes and replay-safe progress.

Issue #2892, opened 2026-03-08
Create Pareto Source Connector Test Suite


amuldotexe:
We already have a decent test harness and a growing set of source connector tests.

But one thing that seems to come up again and again in source connector work is not the crate layout. It is the behavior.

The same questions keep showing up:
- when do we move the source forward?
- what happens if sending to Iggy fails?
- what happens after restart?
- if a source marks or deletes records after reading them, when exactly does that happen?

We have already seen parts of this in a few places:
- #2579 added Postgres sink/source integration tests
- #2739 and the MongoDB source work ended up being mostly about these behavior questions

This makes me think it may be useful to have a small shared source test suite on top of the current harness.

By Pareto I just mean: not a huge framework, just a small set of tests that catch most of the common mistakes.

A practical first version could check things like:
- progress does not move forward before send to Iggy succeeds
- a failed send is offered again on the next poll
- restart resumes from committed state
- mark/delete-after-read behavior happens only after commit
- the replay/duplicate window is at least explicit

This would not replace connector-specific tests. It would just give new source connectors a small common baseline.

## Done when

- there is a reusable source-side test helper or suite on top of the current harness
- at least one existing source connector is wired into it
- there is a short note showing how a new source connector plugs into it

arunsingh:
@hubcio , I would like to work on #2892.

I plan to keep the first PR small: add a reusable source-side integration test helper on top of the existing harness, wire the existing random source into it, and add short docs showing how a future source connector can plug in.

I will avoid changing connector FFI or source trait semantics in the first PR. If this scope looks right, could you please assign the issue to me?

hubcio:
@arunsingh assigned, good luck! don't hesitate to ask us about anything related on discord.

PR #3285, opened 2026-05-20
feat(connectors): add mongodb source connector

Luishfs:
## Which issue does this PR close?

Closes #3042 

## What changed?
Added MongoDB as a source connector

## Local Execution

- Passed
Video

https://github.com/user-attachments/assets/a5db01b0-bec3-4e65-9617-5af753679fe0

 showing results ( with the master branch updated, since last video was a month ago )

- Pre-commit hooks passed
last commit with prek
Screenshot from 2026-05-20 08-25-13

## AI Usage
1. Claude
2. Claude for general rust questions & MongoDB insertion bash i used in the video
3. Didn't use generated code
4. Yes ( just might need a bit of time since i pushed it just now =P )

Issue #2753, opened 2026-02-17
Connector Ecosystem Roadmap: tracking 120+ connector targets across databases, cloud, IoT, and observability


amuldotexe:
# Connector Ecosystem Roadmap: tracking 120+ connector targets across databases, cloud, IoT, and observability

## Overview

Apache Iggy's connector ecosystem is growing. This issue tracks the full landscape of connectors -- what's done, what's in progress, what's planned, and what's proposed next -- to help contributors find high-impact work and avoid duplication.

**Current state**: 5 native Rust connectors (sinks + sources) + 2 Java connectors + 2 codecs in progress.

**Competitive context**: Kafka Connect has 200+ connectors, Redpanda Connect has 200+ (via Benthos), Pulsar IO has 40+ built-in. Iggy's Rust plugin model (cdylib + C FFI) offers performance and resource advantages but needs breadth to compete.

---

## Status: Done

| Connector | Type | Issue | Notes |
|-----------|------|-------|-------|
| [PostgreSQL](https://www.postgresql.org/) | Sink + Source (CDC) | #1849 | Native Rust. WAL-based CDC for source. ~2,400 LOC |
| [Elasticsearch](https://www.elastic.co/elasticsearch) | Sink + Source | #1851 | Native Rust. Bulk API sink, scroll/search source. ~1,500 LOC |
| [Apache Iceberg](https://iceberg.apache.org/) | Sink | #1850 | Native Rust. 7k GitHub stars, used by Airbnb, Apple, LinkedIn |
| [Quickwit](https://quickwit.io/) | Sink | -- | Native Rust. Cloud-native search engine. ~260 LOC |
| Stdout | Sink | -- | Reference implementation for connector developers. ~155 LOC |
| [Apache Pinot](https://pinot.apache.org/) | Sink | -- | Java connector via Kafka compatibility layer |

**Connector runtime infrastructure** (shipped):
- HTTP-based configuration provider (#2388)
- Separated connector/runtime config (#2318)
- E2E tests: Iceberg sink (#2595), Elasticsearch sink (#2592), Quickwit sink (#2594)

---

## Status: In Progress (Open Issues)

| Connector | Type | Issue | Assignee | Notes |
|-----------|------|-------|----------|-------|
| [InfluxDB](https://www.influxdata.com/) | Sink + Source | #2700 | @ryerraguntla | 31k GitHub stars. #1 time-series DB. IoT standard |
[...]

PR #3377, opened 2026-05-29
test(connectors): add random source liveness helper

arunsingh:
## Summary

Part of #2892.

Adds a small random source liveness helper and wires the random source smoke test through it.

Changes:

- Add `connectors::random_source_liveness` for bounded random source message polling.
- Replace random source polling with the liveness helper.
- Drop random restart coverage because random has no stable id/cursor to prove resume.

The broader behavior suite should start from sources with observable cursors/ids, such as postgres or influxdb, with #2940 scope kept separate for failed-send semantics.

## Local validation

- `cargo fmt --all --check`
- `git diff --check`

Local blockers:

- `cargo test -p integration --no-run` is blocked on Windows before reaching these tests by existing Unix-only `message_bus` imports.
- WSL validation is blocked because `cargo` is not installed in WSL.


core/integration/tests/connectors/source_suite.md
hubcio, line 1: remove this file; source_suite.rs code should be self-documenting

hubcio (changes requested):

core/integration/tests/connectors/source_suite.md
arunsingh: I have removed `source_suite.md`; the helper now stands on the Rust code only.

hubcio:
overall change is OK, but what about the rest of connectors? you marked that #2892 will be closed, yet this issue it only touching the random source connector which is essentially a pure-test connector type.

arunsingh:
> overall change is OK, but what about the rest of connectors? you marked that #2892 will be closed, yet this issue it only touching the random source connector which is essentially a pure-test connector type.

@hubcio I have changed this to `Part of #2892`; this PR won't close the issue, and real source connectors can be wired in follow-ups. Let me know if you want this merged as the first step or want one real source connector wired before merge,  likely, wire one Docker-backed source connector, probably Postgres or Elasticsearch, into source_suite.

core/integration/tests/connectors/random/random_source.rs
```diff
@@ -17,79 +17,33 @@
...
-        messages.current_offset > 0,
-        "Current offset should be greater than 0"
-    );
 }
 
 #[iggy_harness(
     server(connectors_runtime(config_path = "tests/connectors/random/source.toml")),
     seed = seeds::connector_stream
 )]
 async fn state_persists_across_connector_restart(harness: &mut TestHarness) {
```
hubcio: this test can't fail. it polls two different fresh consumers (`source_suite_before_restart`, then `source_suite_after_restart`) against a source that emits forever, and only checks `!after.is_empty()`. a fresh consumer gets new messages whether or not connector state survived the restart, so a state-loss regression still passes green - which is worse than no test under this name.

the old `current_offset` check was the partition head too (it's documented as the partition's offset, not the consumer's), so it never proved cursor survival either. the pattern to copy is the postgres source restart test from #2579: one consumer across the restart, seed a second batch with strictly-greater ids, assert every post-restart id is greater than the pre-restart max. random can't supply that discriminator, so either move this test to postgres or drop it and keep random as a liveness smoke (and rename it).

core/integration/tests/connectors/random/random_source.rs
```diff
@@ -17,79 +17,33 @@
...
-            &Consumer::new(consumer_id),
-            &PollingStrategy::next(),
-            10,
-            true,
-        )
-        .await
-        .expect("Failed to poll messages");
-
+    let messages = source_suite::assert_source_produces_messages(harness).await;
     assert!(
```
hubcio: redundant - `assert_source_produces_messages` already asserts non-empty before returning. drop this, or call `poll_until_min_messages` directly if you want the explicit check here.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,101 @@
...
+                    None,
+                    &Consumer::new(consumer_id.clone()),
+                    &PollingStrategy::next(),
+                    config.poll_batch,
+                    true,
+                )
+                .await
+                .expect("poll source messages");
+
+            if polled.messages.len() >= config.min_messages {
```
hubcio: this returns a single poll iteration's messages, but with `auto_commit=true` above each poll commits its batch. when `min_messages` is larger than one server batch, the earlier batches are committed and then dropped here - never accumulated - so the loop can spin to the timeout while silently discarding everything it read. latent now since the default `min_messages` is 1, but it bites the moment a real source is wired with a higher count. accumulate into a `Vec` across iterations like the postgres and influxdb source loops do.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,101 @@
...
+ * under the License.
+ */
+
+use std::time::Duration;
+
+use iggy_common::{Consumer, Identifier, IggyMessage, MessageClient, PollingStrategy};
+use integration::harness::{TestHarness, seeds};
+use tokio::time::{sleep, timeout};
+
+pub(crate) struct SourceSuiteConfig {
```
hubcio: as-is this can only assert liveness - there's no notion of message identity, exact count, ordering, or the connector cursor, so it can't host the checks #2892 is about (restart-resume needs an id discriminator, replay/dup needs `== N`). if this is meant to grow into the suite, the shape wants to be a fixture/trait the runner is generic over (a source-under-test that seeds n stamped records and states its expected set), not a free-function poll. sketch is in the review summary.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub(crate) consumer_name: &'static str,
+    pub(crate) min_messages: usize,
+    pub(crate) poll_batch: u32,
+    pub(crate) warmup: Duration,
```
hubcio: this hardcodes its own retry budget (`warmup`/`retry_interval`/`timeout`). worth noting there isn't a single shared constant to reuse yet - postgres and elasticsearch each carry their own `POLL_ATTEMPTS=100`/`POLL_INTERVAL_MS=50` copy and influxdb just hardcodes `0..100`/`from_millis(100)`. a helper whose point is to absorb duplication should unify all of those into one place rather than add a fourth timing style.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,101 @@
...
+            timeout: Duration::from_secs(5),
+        }
+    }
+}
+
+pub(crate) async fn poll_until_min_messages(
+    harness: &TestHarness,
+    config: &SourceSuiteConfig,
+) -> Vec<IggyMessage> {
+    sleep(config.warmup).await;
```
hubcio: this fixed 1s sleep is dead time - it's before and outside the `timeout` below, so worst case is warmup + timeout. the harness already gates runtime readiness with a health-poll on startup, and the retry loop below covers not-ready, which is why the existing postgres/elasticsearch/influxdb source loops poll immediately with no warmup. drop it.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,101 @@
...
+                    &stream_id,
+                    &topic_id,
+                    None,
+                    &Consumer::new(consumer_id.clone()),
+                    &PollingStrategy::next(),
+                    config.poll_batch,
+                    true,
+                )
+                .await
+                .expect("poll source messages");
```
hubcio: `.expect` here panics the test on any transient poll error instead of retrying. the postgres and elasticsearch source loops use `if let Ok(...)` and keep looping until their timeout, which is the more robust pattern. (the influxdb source loops already use `.expect` too, so they share this fragility - worth fixing there rather than copying it here.)

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,101 @@
...
+                return polled.messages;
+            }
+
+            sleep(config.retry_interval).await;
+        }
+    };
+
+    timeout(config.timeout, poll)
+        .await
+        .expect("source suite timed out waiting for messages")
```
hubcio: on timeout this doesn't say which consumer or config timed out - include `consumer_name` so a failure points at the right test.

hubcio (changes requested):
thanks for PR, I finally had some time to properly review it.

two blockers inline (the restart test can't actually fail, and `poll_until_min_messages` drops messages when `min_messages` is above one batch), then a note on direction.

on direction: #2892 is really about behaviour - progress-before-send, retry-on-fail, restart-resume, mark/delete-after-commit, replay window. a poll-until-non-empty helper proves liveness, which already works everywhere; it doesn't give a seam for any of those, and two structural things stop it growing into the suite as-is.

first, the random source can't host 4 of the 5 checks - no seed, the produced `ProducedMessage` goes out with envelope `id: None` (only a random, non-monotonic `Uuid` lives inside the json payload as `Record.id`), no consumer-visible cursor (just an in-process `messages_produced` counter that's already unit-tested in the crate), and no external system to mark/delete. so it can only ever be a liveness smoke. the sources that already have an observable, gated cursor are postgres and influxdb - those are the honest first wires for a real behaviour test, and the postgres source/restart tests from #2579 are the pattern to copy.

second, heads up that behaviours 1 and 2 aren't guaranteed by the SDK today: the plugin advances its cursor inside `poll()` before returning (random does `messages_produced +=`, postgres marks/deletes rows and bumps tracking offsets), and the send path is one-way - `SendCallback` returns `()`, and the runtime's failed-send branch only logs/metrics/sets-error, nothing re-offers the batch. so on a failed send the un-sent batch is dropped (at-most-once on the runtime->iggy hop), and #2892 item 2 can't pass against any source without an SDK/FFI change. this is the same contract question already open in #2940 (partial writes / replay-safe progress / deterministic write identity) - worth linking the suite scope to that and rescoping, rather than writing a test that asserts behaviour the code doesn't have. the one invariant the runtime does enforce is `on-disk state only advances on send success` (the state save sits entirely in the success arm), which is testable now.

suggested shape if you want this to become the real suite: a small `SourceUnderTest` trait the runner is generic over (config_toml, seed(n) with a monotonic stamped id, expected()), plus one accumulating `poll_until_n` that replaces the inline loops and unifies the retry budget - today that's two independent per-module copies of `POLL_ATTEMPTS`/`POLL_INTERVAL_MS` (postgres + elasticsearch) plus hardcoded literals in influxdb, so there's no single shared constant to point at yet. that's basically the existing postgres fixture made generic. with the connector roadmap (#2753) and new sources still landing (mongodb #3285, meilisearch #3404), a real per-source baseline pays off. random then stays as an explicit liveness smoke and the restart check lives on postgres.

arunsingh:
Updated this to liveness-only: random restart is dropped, polling now accumulates/retries, and postgres/influxdb behavior coverage can follow with #2940 scope in mind.

core/integration/tests/connectors/random/random_source.rs
arunsingh: thanks @hubcio for detailed review, I appreciate your efforts, I have dropped the random restart test; random now stays as liveness smoke only.

core/integration/tests/connectors/random/random_source.rs
arunsingh: Removed the duplicate assertion and call the helper directly.

core/integration/tests/connectors/source_suite.rs
arunsingh: Changed polling to accumulate batches until min_messages is reached.

core/integration/tests/connectors/source_suite.rs
arunsingh: I narrowed this PR to liveness only; the SourceUnderTest behavior suite belongs in the postgres/influxdb follow-up.

core/integration/tests/connectors/source_suite.rs
arunsingh: Removed warmup; I’ll keep cross-source retry-budget unification for the behavior-suite follow-up.

core/integration/tests/connectors/source_suite.rs
arunsingh: Removed the fixed warmup; polling starts immediately and retries inside the timeout.

core/integration/tests/connectors/source_suite.rs
arunsingh: Poll errors now retry within the timeout budget.

core/integration/tests/connectors/source_suite.rs
arunsingh: Timeout failures now include the consumer name.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,97 @@
...
+ * under the License.
+ */
+
+use std::time::Duration;
+
+use iggy_common::{Consumer, Identifier, IggyMessage, MessageClient, PollingStrategy};
+use integration::harness::{TestHarness, seeds};
+use tokio::time::{sleep, timeout};
+
+pub(crate) struct SourceSuiteConfig {
```
hubcio: `SourceSuiteConfig` carries 5 fields plus a `Default` impl, but the only caller goes through `::default()` with `min_messages=1` and never overrides anything, and `poll_until_min_messages` is `pub(crate)` with no other caller. that's parameterization with nothing to parameterize yet. non-blocking, but worth collapsing: inline the poll loop into `assert_source_produces_messages` with hardcoded constants (min 1, batch 100, 100ms retry, 5s timeout). bonus, it makes the multi-batch accumulate path unreachable by construction instead of unreachable by luck. when the real postgres/influxdb suite lands it'll want an observable-cursor abstraction, not this counter poller, so there's little to reuse here anyway.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub(crate) min_messages: usize,
+    pub(crate) poll_batch: u32,
+    pub(crate) retry_interval: Duration,
+    pub(crate) timeout: Duration,
+}
+
+impl Default for SourceSuiteConfig {
+    fn default() -> Self {
+        Self {
+            consumer_name: "source_suite_consumer",
```
hubcio: the `source_suite` naming (this file, the struct, the `source_suite_consumer` id, the `source suite timed out` panic text) over-promises - the actual behaviour suite was deferred to the postgres/influxdb follow-up, and this is a liveness smoke for the one pure-test source. renaming to something like `random_source_liveness` would make the name match the scope. follow-up, not blocking.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    timeout(config.timeout, poll).await.unwrap_or_else(|_| {
+        panic!(
+            "source suite timed out after {:?} waiting for {} messages for consumer {}",
+            config.timeout, config.min_messages, config.consumer_name
+        )
+    })
+}
+
+pub(crate) async fn assert_source_produces_messages(harness: &TestHarness) -> Vec<IggyMessage> {
```
hubcio: `-> Vec<IggyMessage>` is unused - the sole caller (`random_source_produces_messages`) discards the result. make it `-> ()` unless a content-checking caller is coming soon.

core/integration/tests/connectors/source_suite.rs
```diff
@@ -0,0 +1,97 @@
...
+    let messages = poll_until_min_messages(harness, &SourceSuiteConfig::default()).await;
+    assert!(
```
hubcio: this `assert!(!messages.is_empty())` can't fire - `poll_until_min_messages` only returns when `len() >= min_messages` (default 1) or panics on timeout, so a successful return already guarantees non-empty. drop it. (if you inline the loop here as suggested above it turns into a real guard, since the threshold constant would live alongside.)

hubcio (dismissed):

hubcio:
I approved, above comments are just nits. feel free to fix them now or in future PRs. let me know :)

core/integration/tests/connectors/source_suite.rs
arunsingh: Collapsed this into fixed random liveness polling and removed the unused config helper.

core/integration/tests/connectors/source_suite.rs
arunsingh: Renamed the helper to random_source_liveness so the name matches the narrowed scope.

core/integration/tests/connectors/source_suite.rs
arunsingh: Changed the helper to return ().

core/integration/tests/connectors/source_suite.rs
arunsingh: Removed the unreachable assert; the non-empty check now lives in the poll guard.

arunsingh:
Addressed the nits in 08d22879 and kept the PR scoped to random source liveness.

hubcio (approved):

spetz (approved):

PR #3379, opened 2026-05-30
feat(go): implement golang sdk logger

### LINKED ISSUE #3364: Implement Golang SDK logger
### Description

The Go SDK currently uses `log.Printf` for internal logging, which writes directly to stdout. This is problematic.

### Component

Golang SDK 🐹

### Proposed solution

(I am not a expert in logger so I believe you can come up with a better solution):
- Define a minimal logger interface:
```go
type Logger interface {
     Debug()
     Info()
     Error()
     Warning()
     ....
}
```

- Accept it as a client option: `WithLogger(logger Logger)`
- The default logger should be silent by default, but it still need to have enough feature implemented that user can use by simply modify the log level/filter, and the output stream should also be configurable.
- Replace all `log.Printf` calls with c.logger.Printf, mirror the logger with rust sdk.

References

Any exellent Golang opensource project. Many of them have a logger in it :)

slash-init:
# Which issue does this PR address?

Closes #3364 

## Rationale

The SDK previously logged internal events using the standard library `log.Printf`, giving users no way to control the output. They couldn't turn it off, change the log level, or redirect it. This PR introduces configurable, structured logging via the standard library's `log/slog`, allowing users to fully control observability.

## What changed?

-   Replaced internal `log.Printf` calls with structured `*slog.Logger` logging.
    
-   Added a `WithLogger(*slog.Logger)` client option for custom logger configuration.
    
-   Logging is now disabled by default using `slog.DiscardHandler`; users must explicitly provide a logger via `WithLogger()` to enable SDK diagnostic logs.

## Local Execution

Passed

- Ran `go vet ./...` - no issues
- Ran `gofmt` - all files formatted correctly
- Ran `golangci-lint` - no errors or warnings
- Ran `go build ./...` - builds successfully
- Ran `go test ./contracts ./client/tcp ./internal/util` - all tests pass

## AI Usage

Tool: Claude Sonnet 4.6

Scope: Assisted with migration from `log.Printf` to `slog`. All generated code was reviewed, integrated and tested manually.

Verification: All code compiles and tests pass locally.


foreign/go/client/tcp/tcp_core.go
```diff
@@ -71,6 +72,8 @@ type config struct {
 	reconnection tcpClientReconnectionConfig
 	// noDelay disable Nagle's algorithm for the TCP connection
 	noDelay bool
+	// logger is the logger used for internal diagnostics
+	logger iggcon.Logger
```
chengxilo:
```suggestion
```

You already include logger as part of `IggyTcpClient ` at R54. This one should be removed.

chengxilo:
I was doing code review but realized that you are still working on it since github tell me to refresh the PR. However, I do found some problem. I'm sorry when I wrote the misleading proposal since it's just a small silly example.

I did some small research on it and realized that there is something new about go log system and it would probably help: https://go.dev/blog/slog. I think it can be used directly to replace the `iggcon.Logger` interface. we just need to use  `*slog.Logger` as the logger interface, perhaps.

slash-init:
> I was doing code review but realized that you are still working on it since github tell me to refresh the PR. However, I do found some problem. I'm sorry when I wrote the misleading proposal since it's just a small silly example.
> 
> I did some small research on it and realized that there is something new about go log system and it would probably help: https://go.dev/blog/slog. I think it can be used directly to replace the `iggcon.Logger` interface. we just need to use `*slog.Logger` as the logger interface, perhaps.

Thanks for the clarification. I introduced the interface based on the original proposal, but I agree that using `*slog.Logger` directly would keep things simpler and align with the standard library.
I'll update the implementation to use `slog` directly unless you have a different approach in mind.

chengxilo:
By the way, test is missing. So if you don't mind please provide unit test for this PR.

foreign/go/contracts/logger.go
```diff
@@ -0,0 +1,39 @@
...
+	})
+	return slog.New(handler)
+}
+
+func NewStderrLogger(level slog.Level) *slog.Logger {
+	return NewLogger(level, os.Stderr)
+}
+
+func NopLogger() *slog.Logger {
+	return slog.New(slog.NewTextHandler(io.Discard, nil))
```
hubcio: `NopLogger()` isn't actually silent. `slog.NewTextHandler(io.Discard, nil)` passes nil options, so the handler's min level defaults to info - info/warn/error still get fully formatted and then written to `io.Discard`, only debug is skipped. measured on go 1.25 that's ~430-530 ns/op (and 1 alloc once an `error` value is passed, which is exactly the heartbeat warn) vs ~3ns for a true noop. the `WithLogger` doc says output is "silently discarded" and the PR claims zero overhead, which only holds for debug.

fix is one line: `return slog.New(slog.DiscardHandler)`. it reports `Enabled() == false` at every level so nothing is formatted, 0 alloc across the board, and it's already in the stdlib (go.mod and toolchain are on 1.25). all the current call sites are cold (heartbeat, login, redirect) so there's no runtime regression today, but the claim and the contract should match the code.

foreign/go/contracts/logger.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func NewLogger(level slog.Level, writer io.Writer) *slog.Logger {
+	handler := slog.NewTextHandler(writer, &slog.HandlerOptions{
+		Level: level,
+	})
+	return slog.New(handler)
+}
+
+func NewStderrLogger(level slog.Level) *slog.Logger {
```
hubcio: `NewStderrLogger` has no non-test callers - the only thing that calls it is its own test (`TestNewStderrLogger_RespectsLevel`), and `NewLogger(level, os.Stderr)` already covers what it does. golangci's `unused` won't flag it since it's exported, so it just sits as dead public surface. worth dropping it (and its test).

foreign/go/contracts/logger.go
```diff
@@ -0,0 +1,39 @@
...
+
+package iggcon
+
+import (
+	"io"
+	"log/slog"
+	"os"
+)
+
+func NewLogger(level slog.Level, writer io.Writer) *slog.Logger {
```
hubcio: two small things on the exported helpers: `NewLogger(level, writer)` flips the usual slog arg order (`slog.NewTextHandler(writer, opts)` is writer-first), so `NewLogger(writer, level)` would read more naturally. also none of `NewLogger`/`NewStderrLogger`/`NopLogger` have doc comments - not a CI failure with the current lint config, but they're exported.

foreign/go/internal/util/leader_aware.go
```diff
@@ -31,25 +31,36 @@ import (
 
 // CheckAndRedirectToLeader queries the client for cluster metadata and returns
 // an address to redirect to (empty string means no redirection needed).
-func CheckAndRedirectToLeader(ctx context.Context, c iggcon.Client, currentAddress string, transport iggcon.Protocol) (string, error) {
-	log.Println("Checking cluster metadata for leader detection")
+func CheckAndRedirectToLeader(ctx context.Context, c iggcon.Client, currentAddress string, transport iggcon.Protocol, logger *slog.Logger) (string, error) {
+	logger.Debug("Checking cluster metadata for leader detection")
```
hubcio, line 35: `logger` is dereferenced here unconditionally as the first statement, and again in `processClusterMetadata`. adding the param turned this exported helper into one with an undocumented non-nil precondition. it's safe through the real path today since both `NewIggyClient` and `NewIggyTcpClient` default a nil logger to `NopLogger()`, but a struct-literal `IggyTcpClient` (e.g. `newTestClient` in the tcp tests) leaves `logger` nil and would panic the moment the leader path runs. cheapest guard is to default nil to `NopLogger()` at the top of `CheckAndRedirectToLeader`, or document the precondition.

foreign/go/client/iggy_client.go
```diff
@@ -72,15 +82,24 @@ func NewIggyClient(options ...Option) (iggcon.Client, error) {
 		opt(&opts)
 	}
 
+	logger := opts.logger
+	if logger == nil {
+		logger = iggcon.NopLogger()
+	}
+
+	// Prepend the logger option so transport inherits it.
+	tcpOpts := append([]tcp.Option{tcp.WithLogger(logger)}, opts.tcpOptions...)
```
hubcio: this prepends `tcp.WithLogger(logger)` then appends the user's `opts.tcpOptions`, so if someone passes both `WithLogger(a)` and `WithTcp(tcp.WithLogger(b))`, the transport ends up logging through `b` (last write wins) while `ic.logger` at line 102 keeps `a` - and the heartbeat at line 132 logs through `a`. so the client and its transport can silently use two different loggers.

it's structural, not a precedence thing: the transport's `logger` field is unexported so the client can't read it back, and reordering prepend/append won't fix the split. either route the heartbeat through one governing logger so a single logger covers everything, or document that `WithTcp(tcp.WithLogger)` doesn't govern the client-level heartbeat and `WithLogger` is the unified entry. don't just swap the append order - that would clobber an explicit transport logger.

foreign/go/client/iggy_client.go
```diff
@@ -110,7 +129,7 @@ func (ic *IggyClient) Connect(ctx context.Context) error {
 				case <-ticker.C:
 					pingCtx, pingCancel := context.WithTimeout(lifetimeCtx, ic.heartbeatInterval/2)
 					if err := ic.Ping(pingCtx); err != nil {
-						log.Printf("[WARN] heartbeat failed: %v", err)
+						ic.logger.Warn("heartbeat failed", "err", err)
```
hubcio: key here is `"err"` but `leader_aware.go` uses `"error"` for the same kind of field. pick one (slog idiom is `"error"`) so the structured keys stay consistent across the SDK.

hubcio (changes requested):

foreign/go/client/iggy_client.go
slash-init: I went with the documentation approach for now and clarified the behavior of `WithLogger(...)` vs `tcp.WithLogger(...)`. `WithLogger(...)` remains the client-level logger used by heartbeat logging and is forwarded to the transport as a default, while an explicit `tcp.WithLogger(...)` only overrides the transport logger.

foreign/go/client/iggy_client.go
slash-init: Pushed the requested changes.
* `NopLogger` uses `slog.DiscardHandler` (zero formatting, zero alloc)
* dropped `NewStderrLogger` and its test
* flipped `NewLogger` to `(writer, level)` and added doc comments
* added a nil guard on the logger at the top of `CheckAndRedirectToLeader`
* documented the `WithLogger` / `tcp.WithLogger` split on `WithTcp`, `WithLogger`, and `tcp.WithLogger`
* changed the heartbeat key to `"error"` for consistency with `leader_aware.go`

foreign/go/client/tcp/tcp_core.go
```diff
@@ -203,8 +217,14 @@ func NewIggyTcpClient(options ...Option) *IggyTcpClient {
 		}
 	}
 
+	logger := opts.logger
+	if logger == nil {
+		logger = iggcon.NopLogger()
+	}
```
chengxilo: This if statement can be removed. We can just move the default value into the GetDefaultOptions() at R46.

foreign/go/client/tcp/tcp_core.go
chengxilo, line 47:
```suggestion
		config: defaultTcpClientConfig(),
		logger: iggcon.NopLogger() // Or just slog.New(slog.DiscardHandler) should be better in my opinion
```

foreign/go/client/iggy_client.go
```diff
@@ -72,15 +86,26 @@ func NewIggyClient(options ...Option) (iggcon.Client, error) {
 		opt(&opts)
 	}
 
+	logger := opts.logger
+	if logger == nil {
+		logger = iggcon.NopLogger()
+	}
```
chengxilo: same here, move it to GetDefaultOptions(). We set NopLogger() in default options, and let the 
```go
	for _, opt := range options {
		opt(&opts)
	}
```
to overwrite it.

foreign/go/contracts/logger.go
```diff
@@ -0,0 +1,36 @@
...
+	handler := slog.NewTextHandler(writer, &slog.HandlerOptions{
+		Level: level,
+	})
+	return slog.New(handler)
+}
+
+// NopLogger returns a logger that silently discards all output.
+func NopLogger() *slog.Logger {
+	return slog.New(slog.DiscardHandler)
+}
```
chengxilo: Maybe we can delete `logger.go` and `logger_test.go`. If we use `*slog.Logger`, we can just inline `slog.New(slog.DiscardHandler)` for default logger. And for user who want to see the logs, they can do 
```go
client.NewIggyClient(client.WithLogger(slog.New(slog.NewTextHandler(
    os.Stdout, 
    &slog.HandlerOptions{
        Level: slog.LevelDebug,
     }))))
```
So the `NewLogger` won't be actually used.

The NopLogger makes sense, but I think it's also fine to just inline 
```go
slog.New(slog.DiscardHandler)
```

So we can remove both of them, and the corresponding tests.

foreign/go/internal/util/leader_aware.go
```diff
@@ -31,25 +31,39 @@ import (
 
 // CheckAndRedirectToLeader queries the client for cluster metadata and returns
 // an address to redirect to (empty string means no redirection needed).
-func CheckAndRedirectToLeader(ctx context.Context, c iggcon.Client, currentAddress string, transport iggcon.Protocol) (string, error) {
-	log.Println("Checking cluster metadata for leader detection")
+func CheckAndRedirectToLeader(ctx context.Context, c iggcon.Client, currentAddress string, transport iggcon.Protocol, logger *slog.Logger) (string, error) {
+	if logger == nil {
+		logger = iggcon.NopLogger()
+	}
```
chengxilo: Regarding here, if you change the `GetDefaultOptions()`, it should be safe to remove the nil check.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -203,8 +217,14 @@ func NewIggyTcpClient(options ...Option) *IggyTcpClient {
...
+
 	return &IggyTcpClient{
 		config:                 opts.config,
+		logger:                 logger,
```
chengxilo, line 213:
```suggestion
		logger:                 opts.logger,
```

chengxilo (changes requested):
I think both logger.go and logger_test.go can be removed.  But it doesn't mean we don't need test. Actually, I think it would be great if we can test wither the `WithLogger` works.

`NopLogger()` can be replaced with `slog.New(slog.DiscardHandler)` directly, very straight forward and easy to understand.

foreign/go/client/iggy_client.go
chengxilo: Another approach would be remove `tcp.WithLogger`, and change the signature of `NewIggyTcpClient` to
`func NewIggyTcpClient(logger *slog.Logger, options ...Option) *IggyTcpClient`. logger actually is not a optional parameter, we always use `NewIggyClient` to create TCP client, and we always provide a logger to it. This is actually better than document it. (in my opinion)

slash-init:
> I think both logger.go and logger_test.go can be removed. But it doesn't mean we don't need test. Actually, I think it would be great if we can test wither the `WithLogger` works.
> 
> `NopLogger()` can be replaced with `slog.New(slog.DiscardHandler)` directly, very straight forward and easy to understand.

Makes sense. If we remove `logger.go` and `logger_test.go`, where would you prefer the replacement tests to live? I was thinking of moving them closer to the actual usage and testing `WithLogger(...)` behavior in the client/TCP tests instead, but wanted to check if that's what you had in mind.

chengxilo:
> > I think both logger.go and logger_test.go can be removed. But it doesn't mean we don't need test. Actually, I think it would be great if we can test wither the `WithLogger` works.
> > `NopLogger()` can be replaced with `slog.New(slog.DiscardHandler)` directly, very straight forward and easy to understand.
> 
> Makes sense. If we remove `logger.go` and `logger_test.go`, where would you prefer the replacement tests to live? I was thinking of moving them closer to the actual usage and testing `WithLogger(...)` behavior in the client/TCP tests instead, but wanted to check if that's what you had in mind.

Yeah, great choice, under the `foreign/go/client` and `foreign/go/client/tcp` is appropriate.

slash-init:
I removed the custom logger helpers and moved the default logger into `GetDefaultOptions()`. Per Chengxilo's suggestion, I kept the logger parameter mandatory and removed `tcp.WithLogger`, so the TCP client now receives the logger directly and the client and transport always use the same logger. I also added tests covering the logger behavior in both the client and TCP packages as well.

foreign/go/client/iggy_client.go
```diff
@@ -40,6 +41,7 @@ func GetDefaultOptions() Options {
 	return Options{
 		protocol:          iggcon.Tcp,
 		tcpOptions:        nil,
+		logger:            slog.New(slog.DiscardHandler),
```
chengxilo: **Behavioral change**: the old `log.Printf` wrote heartbeat failures and leader redirection events to stderr by default. With `slog.DiscardHandler`, operators get zero visibility unless they explicitly call `WithLogger`. The docstring documents this as intentional, but it reverses the old default-visible behavior.

Worth considering whether `slog.Default()` (writes to stderr at Info level, matching old behavior) would be a better default for discoverability.

foreign/go/client/iggy_client.go
```diff
@@ -54,8 +56,19 @@ func WithTcp(tcpOpts ...tcp.Option) Option {
...
+// WithLogger sets the logger for the Iggy client and its underlying transport.
+// This logger is used by the heartbeat and forwarded to the transport as a
+// default. When no logger is provided, all internal log output is silently
+// discarded.
+func WithLogger(logger *slog.Logger) Option {
+	return func(opts *Options) {
+		opts.logger = logger
+	}
+}
+
```
chengxilo, line 70:
```suggestion
func WithLogger(logger *slog.Logger) Option {
     return func(opts *Options) {
         if logger == nil {
             return
         }
         opts.logger = logger
     }
}
```
`WithLogger(nil)` will stores a nil `*slog.Logger`, overwriting the `DiscardHandler` default from `GetDefaultOptions`. This nil propagates to both `IggyClient.logger` and `IggyTcpClient.logger`.

foreign/go/client/iggy_client_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"strings"
+	"testing"
+)
+
+func TestGetDefaultOptions_ProvidesLogger(t *testing.T) {
+	opts := GetDefaultOptions()
+	if opts.logger == nil {
+		t.Fatal("expected default options to include a non-nil logger")
+	}
+}
```
chengxilo:
```suggestion
```

unnecessary

foreign/go/client/iggy_client_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	cli, err := NewIggyClient(WithLogger(logger))
+	if err != nil {
+		t.Fatalf("unexpected error creating client: %v", err)
+	}
+
+	ic, ok := cli.(*IggyClient)
+	if !ok {
+		t.Fatal("expected *IggyClient from NewIggyClient")
+	}
```
chengxilo:
```suggestion
	ic := cli.(*IggyClient)
```
it will always be a IggyClient so the check is not needed.

foreign/go/client/iggy_client_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+)
+
+func TestGetDefaultOptions_ProvidesLogger(t *testing.T) {
+	opts := GetDefaultOptions()
+	if opts.logger == nil {
+		t.Fatal("expected default options to include a non-nil logger")
+	}
+}
+
```
chengxilo, line 37: As we will provide a nil check, I recommend test it to prevent regression.

```suggestion


func TestWithLogger_Nil(t *testing.T) {
	cli, err := NewIggyClient(WithLogger(nil))
	if err != nil {
		t.Fatalf("unexpected error creating client: %v", err)
	}

	ic := cli.(*IggyClient)
	if ic.logger.Handler() != slog.DiscardHandler {
		t.Errorf("expected slog.DiscardHandler, get %v", ic.logger.Handler())
	}
}
```

chengxilo (commented):
Just a few changes needed. Mainly the nil  guard

foreign/go/client/iggy_client.go
slash-init: > `WithLogger(nil)` will stores a nil `*slog.Logger`, overwriting the `DiscardHandler` default from `GetDefaultOptions`. This nil propagates to both `IggyClient.logger` and `IggyTcpClient.logger`.

Thanks for pointing this out. I've made the requested changes.

slash-init:
I've added the nil guard for `WithLogger(nil)`, updated the logger tests, and removed the unnecessary default-options test.

foreign/go/client/iggy_client_test.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+func TestWithLogger_Nil(t *testing.T) {
+	cli, err := NewIggyClient(WithLogger(nil))
+	if err != nil {
+		t.Fatalf("unexpected error creating client: %v", err)
+	}
+	ic := cli.(*IggyClient)
+	if ic.logger.Handler() != slog.DiscardHandler {
+		t.Errorf("expected slog.DiscardHandler, get %v", ic.logger.Handler())
```
chengxilo:
```suggestion
		t.Errorf("expected slog.DiscardHandler, got %v", ic.logger.Handler())
```

Nit, typo, introduced by me tho.

chengxilo:
Just minor typo issue. lgtm now. 

We still need to add the logs in the tcp client to align with rust sdk. But it can be done in another PR (Or if you prefer, we can do it here too). We can open an issue when this is done.

slash-init:
> Just minor typo issue. lgtm now.
> 
> We still need to add the logs in the tcp client to align with rust sdk. But it can be done in another PR (Or if you prefer, we can do it here too). We can open an issue when this is done.

Thanks. I missed that typo too.
I guess let's keep it for another PR since the conversation history here is getting a bit too long haha.

numinnex (dismissed):

thepatrykk (dismissed):

foreign/go/client/tcp/tcp_core.go
```diff
@@ -195,7 +197,7 @@ func WithTLSValidateCertificate(validate bool) TLSOption {
 
 // NewIggyTcpClient creates a new Iggy TCP client with the given options.
 // warning: don't use this function directly, use iggycli.NewIggyClient with iggycli.WithTcp instead.
-func NewIggyTcpClient(options ...Option) *IggyTcpClient {
+func NewIggyTcpClient(logger *slog.Logger, options ...Option) *IggyTcpClient {
```
hubcio, line 200: `NewIggyTcpClient` takes `logger *slog.Logger` as a required positional arg but stores it raw with no nil-guard. calling `NewIggyTcpClient(nil)` (or a bare `IggyTcpClient{}` literal) leaves the field nil, and the first login runs `HandleLeaderRedirection` -> `CheckAndRedirectToLeader`, whose first statement is `logger.Debug(...)` - nil pointer deref panic, no recover. the documented `NewIggyClient` + `WithLogger` path is safe (default `DiscardHandler` plus the nil-guard in `WithLogger`), so this only bites direct callers of this exported ctor - but it is exported, and the new required logger arg makes `nil` the natural no-logging value.

cheapest fix is one chokepoint here: `if logger == nil { logger = slog.New(slog.DiscardHandler) }`. that covers every `logger.Debug/Info/Warn` in `leader_aware.go` at once (several of them deref unconditionally, so guarding only the first call site would not be enough). keep the `WithLogger` guard too - it protects the separate `ic.logger` heartbeat field, which this ctor never touches.

separately, the signature change from `(options ...Option)` is a compile break for any external code calling this directly - fine since it is doc-discouraged and pre-1.0, but worth a changelog note.

foreign/go/internal/util/leader_aware.go
```diff
@@ -82,14 +96,23 @@ func processClusterMetadata(metadata *iggcon.ClusterMetadata, currentAddress str
 	}
 
 	leaderAddress := net.JoinHostPort(leader.IP, strconv.Itoa(int(leaderPort)))
-	log.Printf("Found leader node: %s at %s (using %s transport)\n", leader.Name, leaderAddress, transport)
+	logger.Info(
```
hubcio: `logger.Info("Found leader node", ...)` fires before the `isSameAddress` check below, so it logs at info on every redirect check even when you are already on the leader and nothing changes. the steady-state already-leader case probably wants debug here, matching the `Debug` line at the bottom of the function.

foreign/go/internal/util/leader_aware.go
```diff
@@ -82,14 +96,23 @@ func processClusterMetadata(metadata *iggcon.ClusterMetadata, currentAddress str
...
+		"Found leader node",
+		"leader", leader.Name,
+		"address", leaderAddress,
```
hubcio: `leaderAddress` is logged under key `"address"` here but as `"leader_address"` a few lines down (line 110), same value two keys. pick one so the structured keys stay consistent.

foreign/go/internal/util/leader_aware.go
```diff
@@ -82,14 +96,23 @@ func processClusterMetadata(metadata *iggcon.ClusterMetadata, currentAddress str
...
+		logger.Info(
+			"Current connection is not the leader, redirecting",
+			"current_address", currentAddress,
+			"leader_address", leaderAddress,
+		)
 		return leaderAddress, nil
 	}
 
-	log.Printf("Already connected to leader at %s\n", currentAddress)
+	logger.Debug("Already connected to leader at", "current_address", currentAddress)
```
hubcio: message reads `"Already connected to leader at"` with the address now moved to the `current_address` attr, so it renders with a dangling "at". drop the trailing word - `"Already connected to leader"`. the redirect log just above was reworded fine, this one got missed.

foreign/go/client/iggy_client.go
```diff
@@ -54,8 +56,22 @@ func WithTcp(tcpOpts ...tcp.Option) Option {
 	}
 }
 
+// WithLogger sets the logger for the Iggy client and its underlying transport.
+// This logger is used by the heartbeat and forwarded to the transport as a
```
hubcio: doc says the logger is "forwarded to the transport as a default", but there is no `tcp.WithLogger` override anymore (it was removed in this branch), so "as a default" implies an override path that no longer exists. the logger is just passed straight to the transport - reword to drop "as a default".

hubcio (changes requested):
couple of notes that don't map to a single changed line:

- the PR description still talks about a custom `Logger` interface plus `DefaultLogger` / `NoopLogger` implementations, but the code that ships here is stdlib `log/slog` + `slog.DiscardHandler` (the custom helpers were removed mid-PR). worth updating the description so it matches what merges.

- the default logger is now `slog.DiscardHandler` (silent). before this PR these messages went to stderr unconditionally via stdlib `log`. that's the intended behavior per the issue, but it is a behavior change - a dying connection's heartbeat warn now logs nothing unless the user passes `WithLogger`. worth a line in the changelog / release notes so operators know to opt in.

overall, everything looks solid. fix these comments and I think we can merge. good job @slash-init

slash-init:
@hubcio Thanks for the feedback! I'll update the PR description and make those changes.

hubcio (approved):

spetz (approved):

numinnex (approved):

PR #3380, opened 2026-05-30
feat(php): fill Packagist metadata

### LINKED ISSUE #3303: PHP SDK: fill out Packagist metadata

`foreign/php/composer.json` currently has the minimum package fields, but it is too thin for Packagist publication as-is. Before publishing the PHP package, fill out the remaining package metadata so the package is discoverable and has the expected Composer shape.

## Scope

- Add an appropriate `autoload` section for the PHP SDK layout.
- Add relevant `keywords` for Apache Iggy / streaming / messaging / PHP extension bindings.
- Add `authors` metadata.
- Add a `homepage` URL.
- Re-check any other Packagist-facing metadata needed before publication.

This should remain separate from the initial experimental cut, but should not be forgotten before a Packagist publish path is enabled.

Related: #3302
Initial PHP SDK work: #3235

countradooku:
## Summary

- fill out `foreign/php/composer.json` Packagist metadata: homepage, keywords, authors, support links, and an explicit empty autoload section
- switch the package type to `php-ext` and add PIE `php-ext` metadata for the `iggy_php` extension
- add archive exclusions so Composer package archives stay scoped to the PHP extension source and omit dev/test/vendor output

Closes #3303.

## Validation

- `php -r 'json_decode(file_get_contents("foreign/php/composer.json"), true, flags: JSON_THROW_ON_ERROR);'`
- `composer validate --strict foreign/php/composer.json`
- `composer archive --format=zip` from `foreign/php` and inspected the archive contents
- `PATH=/tmp/iggy-addlicense-bin:$PATH prek run --files foreign/php/composer.json`
- pre-push hook: `cargo clippy` passed; Go/Java/C# hooks skipped as no relevant files changed

## Notes

The archive excludes use `composer.json` instead of root `.gitattributes` so Apache source-release archives created with `git archive` are not affected.


foreign/php/composer.json
```diff
@@ -1,14 +1,95 @@
...
+        "extension-name": "iggy_php",
+        "support-zts": false,
+        "support-nts": true,
+        "configure-options": [],
+        "download-url-method": [
+            "pre-packaged-binary",
+            "composer-default"
+        ]
+    },
+    "archive": {
```
hubcio: these excludes are package-root-relative (the `foreign/php/` dir), so the leading-slash repo-root paths like `/core`, `/bdd`, `/web`, `/.github`, `/foreign/csharp` never match anything here - only `/tests`, `/phpunit.xml.dist`, `/Dockerfile.test`, `/docker-compose.test.yml`, `/target`, `/vendor` actually exist under this dir. separately, `archive.exclude` only affects the `composer archive` cli. when someone `composer require`s this, packagist serves the github zipball and PIE's `composer-default` builds from the github source zip - both honor `.gitattributes export-ignore`, not `archive.exclude`. so once publishing is wired this whole block gets ignored. a scoped `foreign/php/.gitattributes` with `export-ignore` does what you want here - keeps the root `git archive` source release untouched (your stated reason for avoiding root `.gitattributes`) while actually filtering the real download path.

foreign/php/composer.json
```diff
@@ -1,14 +1,95 @@
 {
     "name": "apache/iggy-php",
     "description": "PHP extension bindings for Apache Iggy.",
+    "type": "php-ext",
```
hubcio, line 4: flagging for the eventual publish path (out of scope here): public packagist can't register a package whose `composer.json` lives in a repo subdirectory - it references the source repo root and won't zip subtrees [packagist#472](https://github.com/composer/packagist/issues/472). there's no root `composer.json` or split tooling, so `apache/iggy-php` isn't installable from this monorepo as-is.

foreign/php/composer.json
```diff
@@ -1,14 +1,95 @@
...
     "require-dev": {
         "phpunit/phpunit": "^10.5"
     },
+    "autoload": {},
+    "php-ext": {
+        "extension-name": "iggy_php",
+        "support-zts": false,
+        "support-nts": true,
+        "configure-options": [],
+        "download-url-method": [
```
hubcio, line 40: `pre-packaged-binary` makes PIE look for github release assets named like `php_iggy_php-{ver}_php{X.Y}-{arch}-{os}-{libc}-nts.zip`. nothing produces those yet, so PIE silently falls through to `composer-default` and builds from source. harmless - that's exactly why the fallback is second - just noting this entry is aspirational until a release-asset workflow exists.

foreign/php/composer.json
```diff
@@ -1,14 +1,95 @@
...
+        "issues": "https://github.com/apache/iggy/issues",
+        "source": "https://github.com/apache/iggy/tree/master/foreign/php"
+    },
     "require": {
         "php": ">=8.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^10.5"
     },
+    "autoload": {},
```
hubcio: empty `autoload` is valid but does nothing for a native extension - there's no php source to autoload, the module ships as a compiled `.so`. cleaner to drop the key entirely.

hubcio (changes requested):

foreign/php/composer.json
```diff
@@ -1,14 +1,46 @@
...
+    "authors": [
+        {
+            "name": "Apache Iggy (Incubating) Contributors",
+            "homepage": "https://iggy.apache.org"
+        }
+    ],
+    "support": {
+        "docs": "https://iggy.apache.org/docs/",
+        "issues": "https://github.com/apache/iggy/issues",
+        "source": "https://github.com/apache/iggy/tree/master/foreign/php"
```
slbotbm, line 27: What do you think about adding a discord link as well?

slbotbm (changes requested):

foreign/php/composer.json
countradooku: Sure

hubcio (approved):

spetz (approved):

numinnex (approved):

PR #3382, opened 2026-05-31
feat(php): map Iggy errors to typed exceptions

### LINKED ISSUE #3294: Map PHP SDK errors to typed exception classes
The PHP SDK currently converts all Rust/Iggy errors into a base PHP Exception through PhpException::default(error.to_string()). That forces retry and circuit-breaker code to inspect message strings.\n\nFollow-up: expose a small Iggy\Exception hierarchy and map at least top-level error categories such as connection, authentication, not found, and transient/server errors to typed PHP exceptions.

WaterWhisperer:

Closes #3294 

## Rationale

The PHP SDK exposed Iggy failures as generic PHP exceptions, so callers could not reliably handle common error categories without parsing exception messages.

## What changed?

Maps `IggyError` to typed PHP exceptions under `Iggy\Exception`.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. AI tools used: Codex
2. Scope of usage: explanation, validation commands
3. Verification: validated with commands
4. Yes

### DIFF
--- a/foreign/php/src/error.rs
+++ b/foreign/php/src/error.rs
@@ -16,8 +16,156 @@
  * under the License.
  */
 
-use ext_php_rs::exception::PhpException;
+use ext_php_rs::{error::Error, exception::PhpException, php_class, zend::ce};
+use iggy::prelude::IggyError;
 
-pub(crate) fn to_php_exception(error: impl std::fmt::Display) -> PhpException {
-    PhpException::default(error.to_string())
+#[php_class]
+#[php(name = "Iggy\\Exception\\IggyException")]
+#[php(extends(ce = ce::exception, stub = "\\Exception"))]
+#[derive(Default)]
+pub struct IggyException;
+
+#[php_class]
+#[php(name = "Iggy\\Exception\\ConnectionException")]
+#[php(extends(IggyException))]
+#[derive(Default)]
+pub struct ConnectionException;
+
+#[php_class]
+#[php(name = "Iggy\\Exception\\AuthenticationException")]
+#[php(extends(IggyException))]
+#[derive(Default)]
+pub struct AuthenticationException;
+
+#[php_class]
+#[php(name = "Iggy\\Exception\\NotFoundException")]
+#[php(extends(IggyException))]
+#[derive(Default)]
+pub struct NotFoundException;
+
+#[php_class]
+#[php(name = "Iggy\\Exception\\TransientException")]
+#[php(extends(IggyException))]
+#[derive(Default)]
+pub struct TransientException;
+
+pub(crate) trait IntoPhpException {
+    fn into_php_exception(self) -> PhpException;
+}
+
+pub(crate) fn to_php_exception(error: impl IntoPhpException) -> PhpException {
+    error.into_php_exception()
+}
+
+impl IntoPhpException for IggyError {
+    fn into_php_exception(self) -> PhpException {
+        let message = self.to_string();
+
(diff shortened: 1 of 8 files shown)


foreign/php/src/error.rs
```diff
@@ -16,8 +16,156 @@
...
+            PhpExceptionKind::Base => PhpException::from_class::<IggyException>(message),
+        }
+    }
+}
+
+impl IntoPhpException for Error {
+    fn into_php_exception(self) -> PhpException {
+        PhpException::from_class::<IggyException>(self.to_string())
+    }
+}
```
hubcio, line 88: this re-wraps every callback failure into a fresh `IggyException` built from `self.to_string()`. for `Error::Exception` - a PHP exception thrown inside the `consume_messages` callback, which `try_call` pulls out of the executor via `take_exception()` - the original exception object is lost (its class, message, stack trace) and the message turns into a `{:?}` debug dump of the `ZendObject` instead of the user's message. so a caller that throws to abort consumption ends up catching a generic `IggyException` with an opaque message. this already happened before the PR (the old `to_php_exception` stringified the error too), but the typed hierarchy makes faithful propagation the expectation now. the fix would be to special-case `Error::Exception` and rethrow the original object via `PhpException::with_object` instead of re-wrapping.

foreign/php/src/lib.rs
```diff
@@ -38,6 +42,11 @@ use crate::topic::TopicDetails;
 #[php_module]
 pub fn get_module(module: ModuleBuilder) -> ModuleBuilder {
     module
+        .class::<IggyException>()
+        .class::<ConnectionException>()
+        .class::<AuthenticationException>()
+        .class::<NotFoundException>()
+        .class::<TransientException>()
```
hubcio, line 51: registration order matters here: `IggyException` has to be registered before its subclasses, otherwise the child's parent lookup hits the `expect(...)` in `ClassMetadata::ce()` and panics during module startup (the parent `ClassEntry` only gets set by its own `register()`). it's correct as written, but nothing enforces it at compile or run time, so a short comment pinning the parent-first ordering would stop a future reorder (say, alphabetizing) from silently breaking extension load.

foreign/php/src/error.rs
```diff
@@ -16,8 +16,156 @@
...
+
+impl IntoPhpException for String {
+    fn into_php_exception(self) -> PhpException {
+        PhpException::from_class::<IggyException>(self)
+    }
+}
```
hubcio, line 94: validation errors raised as `Err("...".into())` (e.g. `consumer_group` / `non_zero_duration_micros` in client.rs, `partition_id_or_current_after_poll` in consumer.rs, `timestamp_seconds` in receive_message.rs) go through ext-php-rs's `From<String>` / `From<&str> for PhpException`, which build a base `\Exception`, not `Iggy\Exception\IggyException`. they never reach this `IntoPhpException for String` path, so `catch (Iggy\Exception\IggyException)` won't catch them. either route those sites through `to_php_exception(...)` to stay consistent with the rest of the SDK, or document that local input validation throws a plain `\Exception`.

hubcio (changes requested):

foreign/php/src/lib.rs
WaterWhisperer: Thanks for the review, I will update as soon as possible

hubcio (approved):

spetz (approved):

PR #3399, opened 2026-06-01
fix(metadata): fix integration tests

numinnex:
This PR fixes all of the metadata operations except purge stream/topic as those are awaiting the partition reconciliation loop PR.


core/metadata/src/impls/metadata.rs
```diff
@@ -1087,6 +1088,105 @@ where
...
+                } else {
+                    RegisterSubmitError::NotPrimary
+                },
+            );
+        }
+
+        // Dedup / session / eviction. `false` = absorbed (duplicate cached
+        // reply already resent, or evicted, or gap). Surface as Canceled so
+        // the home shard stays silent and the SDK replays.
+        if !request_preflight(consensus, &self.client_table, client_id, session, request).await {
```
hubcio, line 1142: heads up - this `request_preflight` path makes a pre-existing reply-routing bug newly reachable for replicated client requests.

on a `Duplicate` (cached reply) or eviction (`NoSession` / `SessionTooLow`), preflight resends via `send_to_client(client_id, ...)` in `metadata_helpers.rs` (~lines 83-90, 95, 103). but `client_id` here is the vsr consensus id (the random uuid the sdk mints), and `send_to_client` routes by `client_id_owning_shard` = the top 16 bits, which only carry home-shard bits for transport ids. a vsr uuid's top bits are random, so the resend misroutes to a garbage shard -> no registry slot -> dropped. bites single-shard too (top16 of a random uuid is ~never 0).

effect: client loses the first reply, the sdk transparently retries the same `(client, request)`, the server sees `Duplicate`, resends -> misroutes again -> client hangs/livelocks until the session times out. eviction is the same - the client never learns it was evicted.

the fresh-commit path already does this right (home shard writes the committed reply via `send_to_client(transport_client_id, ...)`). the dedup/eviction resends need the same treatment - return the `Duplicate`/`Evicted` outcome to the home shard and let it resend the cached body by transport id, since shard 0 can't reconstruct the vsr->transport mapping.

not exercised by the tier-1 tests this pr enables (single-shard tcp happy path, `data_integrity` gated off), so it doesn't break the suite - but it ships in the server path. minimum bar: a TODO at the resend sites + bound the reply-wait so a retry fails fast instead of leaving an orphaned pump parked on `rx.recv()`. full fix before vsr client traffic goes past tier-1.

hubcio (changes requested):

core/metadata/src/impls/metadata.rs
numinnex: Added 3 TODOs, will address some of those in the next PR that suppose to do some cleanup on the most recent work.

hubcio (approved):

mmodzelewski (approved):

PR #3400, opened 2026-06-01
ci: auto-assign issue authors who volunteer to contribute

hubcio:
The bug/feature issue forms ask reporters to tick "I'm willing to
submit a pull request", but nothing acted on it, so volunteers were
never put on the issue. Extend the issue-labeler workflow (renamed
to issue-labeler-assigner) with an add-only step that assigns the
issue to its author when the box is checked. The assignee comes
from the trusted event payload, never the body, so the parsed text
only gates a boolean and cannot inject an assignee.

Also surface /author in the PR triage welcome comment, which only
listed /ready and /request-review, and drop the stale PR size-limit
section from CONTRIBUTING.md.

### DIFF
--- a/.github/workflows/issue-labeler-assigner.yml
+++ b/.github/workflows/issue-labeler-assigner.yml
@@ -15,20 +15,29 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: Issue Labeler
+name: Issue Labeler / Assigner
 
-# Maps the "Affected area / component" dropdown from the bug/feature issue
-# forms (.github/ISSUE_TEMPLATE/) onto real repo labels, so reporters who
-# lack triage permission still get their issue categorized at creation.
+# Two add-only triage steps that run when a reporter opens or edits an issue:
 #
-# Add-only by design: it never removes labels. Reporters rarely deselect,
-# and removing would fight maintainers who labeled by hand. addLabels is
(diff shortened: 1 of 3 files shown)


mmodzelewski (approved):

spetz (approved):

.github/workflows/issue-labeler-assigner.yml
```diff
@@ -103,3 +112,32 @@ jobs:
...
+        with:
+          script: |
+            // The "Contribution" checkbox in both issue forms renders as a
+            // task-list line. Checked, it reads `- [x] I'm willing to submit a
+            // pull request ...`; the tail differs per template (bug: "to fix
+            // this bug", feature: "to implement this feature"), so match the
+            // shared prefix only.
+            const body = context.payload.issue.body || '';
+            const volunteered =
+              /(^|\n)\s*-\s*\[[xX]\]\s+I'm willing to submit a pull request/.test(body);
```
Standing-Man, line 127: Hi @hubcio, Could we use a simpler command(`take` or `assgin`) to trigger this? Also, if the current assignee no longer wants to work on the issue, could they use `untake` to relinquish it?

hubcio: not sure about that. lets discuss that on discord. there are issues that are deliberately not done in months because we don't want them to be done YET.

PR #3404, opened 2026-06-01
feat(connectors): add Meilisearch source connector

countradooku:
## Summary
- add a Meilisearch source connector using the official `meilisearch-sdk`
- wire the source into the connectors workspace and source documentation
- split Meilisearch integration fixtures into `container.rs`, `sink.rs`, and `source.rs`, then add source runtime coverage

## Validation
- `cargo check -p iggy_connector_meilisearch_source`
- `cargo test -p iggy_connector_meilisearch_source`
- `cargo check -p integration`
- pre-push `cargo clippy` hook passed


core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+    retry_delay: Duration,
+    max_retry_delay: Duration,
+    max_open_retries: u32,
+    client: Option<Client>,
+    state: Mutex<State>,
+}
+
+impl MeilisearchSink {
+    pub fn new(id: u32, config: MeilisearchSinkConfig) -> Self {
+        let primary_key = config
```
slbotbm: this function makes a lot of assumptions about the config, some of which are encoded as `const`s and some of which do not have `const`s (`create_index_if_not_exists`, `include_metadata`, `wait_for_tasks`). I think it would be prudent to make these as `const`s as well, and document the default options as a rustdoc or in the readme for the user.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+        _ => false,
+    }
+}
+
+fn map_sdk_error(error: MeilisearchSdkError) -> Error {
+    match error {
+        MeilisearchSdkError::Meilisearch(meilisearch_error) => {
+            if meilisearch_error.error_type == MeilisearchErrorType::Internal {
+                Error::HttpRequestFailed(meilisearch_error.to_string())
+            } else {
```
slbotbm: A from trait can be used here to conform with idiomatic rust

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+        };
+        self.wait_for_task(client, task, false).await?;
+        Ok(documents.len())
+    }
+
+    async fn wait_for_task(
+        &self,
+        client: &Client,
+        task: TaskInfo,
+        allow_index_already_exists: bool,
```
slbotbm: This function is supposed to be a generic helper function, but it is mixing feature-specific code with generic code by including the `allow_index_already_exists` flag. It would be better to let `create_index` handle the `IndexAlreadyExists` part.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+#[derive(Debug)]
+pub struct MeilisearchSink {
+    id: u32,
+    config: MeilisearchSinkConfig,
+    primary_key: String,
+    document_action: MeilisearchDocumentAction,
+    create_index_if_not_exists: bool,
+    include_metadata: bool,
+    batch_size: usize,
+    timeout: Duration,
```
slbotbm, line 116: This code carries both the raw config and a second normalized config. This creates two sources of truth inside the same type.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+
+fn normalize_host(raw: &str) -> Result<String, Error> {
+    let url = Url::parse(raw)
+        .map_err(|error| Error::Connection(format!("Invalid Meilisearch URL: {error}")))?;
+    let mut host = url.to_string();
+    while host.ends_with('/') {
+        host.pop();
+    }
+    Ok(host)
+}
```
slbotbm, line 795: The `normalize_host` in the sink and the source are different. Any reason for that?

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+        let task = self
+            .retry_sdk_operation("create index", || {
+                client.create_index(&self.config.index, Some(&self.primary_key))
+            })
+            .await?;
+        self.wait_for_task(client, task, true).await?;
+
+        info!("Created Meilisearch index '{}'", self.config.index);
+        Ok(())
+    }
```
slbotbm, line 270: The readme says that the `wait_for_tasks` only controls whether `consume()` waits for meilisearch tasks, but this function uses `wait_for_tasks` and immediately returns if the flag is false (the `false` case configuration is broken). This lets `open()` succeed before the index exists, so the first `consume()` can race and fail with `IndexNotFound`

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+                    }),
+                }
+            }
+            Payload::Text(text) => json!({
+                "text": text,
+                "data_type": "text",
+            }),
+            _ => {
+                warn!("Unsupported payload format: {}", messages_metadata.schema);
+                return None;
```
slbotbm: The current approach logs a warning before simply filtering out non-json/text messages. If the batch becomes empty, `consume` returns `Ok(())` which is a silent data-loss path.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+            .or_insert_with(|| Value::String(generated_id.clone()));
+
+        if self.include_metadata {
+            object
+                .entry(DEFAULT_PRIMARY_KEY.to_string())
+                .or_insert_with(|| Value::String(generated_id));
+            object.insert("iggy_message_id".to_string(), Value::String(id.to_string()));
+            object.insert("iggy_offset".to_string(), Value::from(offset));
+            object.insert(
+                "iggy_stream".to_string(),
```
slbotbm: These inserts unconditionally overwrite user fields. If those fields already exist, the document gets mutated on ingestion. Using `entry(...).or_insert(...)` would be a more prudent approach. Logging a warning would also be good.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,838 @@
...
+) -> String {
+    format!(
+        "{}_{}_{}_{}_{}",
+        sanitize_identifier_component(&topic_metadata.stream),
+        sanitize_identifier_component(&topic_metadata.topic),
+        messages_metadata.partition_id,
+        message.offset,
+        message.id
+    )
+}
```
slbotbm, line 748: `sanitize_identifier_component` would convert both `orders.stream` and `orders/stream` to `orders_stream`.  If both also have the same topic after sanitization, same partition, same offset, and same message ID, they produce the same Meilisearch document ID, resulting in loss of data.

slbotbm (changes requested):
review for the sink

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+
+#[derive(Debug)]
+pub struct MeilisearchSource {
+    id: u32,
+    config: MeilisearchSourceConfig,
+    client: Option<Client>,
+    batch_size: usize,
+    polling_interval: Duration,
+    include_metadata: bool,
+    state: Mutex<State>,
```
slbotbm: `MeilisearchSourceConfig` duplicates some options like the sink did.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    documents_produced: usize,
+    poll_count: usize,
+}
+
+impl MeilisearchSource {
+    pub fn new(id: u32, config: MeilisearchSourceConfig, state: Option<ConnectorState>) -> Self {
+        let batch_size = config.batch_size.unwrap_or(DEFAULT_BATCH_SIZE).max(1);
+        let polling_interval =
+            parse_duration(config.polling_interval.as_deref(), DEFAULT_POLLING_INTERVAL);
+        let include_metadata = config.include_metadata.unwrap_or(false);
```
slbotbm: The `false` can become a global const.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+            state.next_offset
+        };
+        let filter_expression = self.filter_expression()?;
+        let sort_refs = self.sort_refs();
+        let index = client.index(&self.config.index);
+        let mut query = index.search();
+        query
+            .with_query(self.config.query.as_deref().unwrap_or_default())
+            .with_offset(offset)
+            .with_limit(self.batch_size);
```
slbotbm: This code saves `next_offset` and advances it by `messages.len()` next time a poll happens. This is only stable if the result set is immutable between polls. Inserting / deleting documents at offsets already polled will lead to re-ordering of documents, and thus to silent skips or re-delivery of documents on the connector's side.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+            self.id, self.config.url, self.config.index
+        );
+
+        let client = self.create_client()?;
+        self.check_connectivity(&client).await?;
+        self.client = Some(client);
+
+        info!(
+            "Successfully opened Meilisearch source connector with ID: {}",
+            self.id
```
slbotbm, line 410: Here, the existence of index is never checked. Cases like a typo in the name, a dropped index, or permissions that hide the index will look like a healthy connector since `IndexNotFound` is converted to `Vec::new()` instead of erroring hard (line 160).

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    pub index: String,
+    #[serde(serialize_with = "iggy_common::serde_secret::serialize_optional_secret")]
+    pub api_key: Option<SecretString>,
+    pub query: Option<String>,
+    pub filter: Option<Value>,
+    pub sort: Option<Vec<String>>,
+    pub batch_size: Option<usize>,
+    pub polling_interval: Option<String>,
+    pub timeout: Option<String>,
+    pub include_metadata: Option<bool>,
```
slbotbm: `timeout` is defined in the config but never used in the connector. Also never referenced in the readme.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+fn normalize_host(host: &str) -> Result<String, Error> {
+    let trimmed = host.trim();
+    if trimmed.is_empty() {
+        return Err(Error::Connection(
+            "Invalid Meilisearch URL: host cannot be empty".to_string(),
+        ));
+    }
+
+    let with_scheme = if host.starts_with("http://") || host.starts_with("https://") {
+        trimmed.to_string()
```
slbotbm, line 460: the host string is trimmed but not used when checking with `starts_with`.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+                .collect::<Result<Vec<_>, _>>()
+                .map(|filters| {
+                    if filters.is_empty() {
+                        None
+                    } else {
+                        Some(filters.join(" AND "))
+                    }
+                }),
+            _ => Err(Error::InvalidConfigValue(
+                "Meilisearch filter must be a string or an array of strings".to_string(),
```
slbotbm: This code can only process flat array of strings, and joins everything together with `AND`. Thus, valid filters like `[[ "genres = abc", "genres = def" ], "director = 'xyz'"]` cannot be here.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+        Ok(())
+    }
+}
+
+fn parse_duration(value: Option<&str>, default: &str) -> Duration {
+    value
+        .and_then(|value| humantime::Duration::from_str(value).ok())
+        .unwrap_or_else(|| humantime::Duration::from_str(default).expect("valid default duration"))
+        .into()
+}
```
slbotbm: This would silently use the default value in case the user-supplied value is wrong. Logging this default behaviour while the connector is in use would be a better alternative.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+}
+
+#[async_trait]
+impl Source for MeilisearchSource {
+    async fn open(&mut self) -> Result<(), Error> {
+        info!(
+            "Opening Meilisearch source connector with ID: {} for URL: {}, index: {}",
+            self.id, self.config.url, self.config.index
+        );
```
slbotbm, line 400: The sink santizes URLs befoer logging, but the source does not. This can lead to leaking of credentials in logs.

core/connectors/sources/meilisearch_source/src/lib.rs
slbotbm: `map_sdk_error` should probably be used here.

slbotbm (commented):
review for the source

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+
+        let client = self
+            .client
+            .as_ref()
+            .ok_or_else(|| Error::Connection("Meilisearch client not initialized".to_string()))?;
+
+        let messages_count = messages.len();
+        let mut documents = Vec::with_capacity(messages.len());
+        for message in messages {
+            if let Some(document) =
```
ryerraguntla: (? on prepare_document, unsupported payload → Err(InvalidRecordValue) at :307-312) → propagates to consume() Err at :626 → runtime/src/sink.rs:740 calls FFI consume and discards return value entirely. consume_messages's Err branch (sink.rs:426-431) only fires for postcard serialize failures, never plugin logic errors. Combined w/ AutoCommit::When(PollingMessages)
  (sdk/src/clients/consumer.rs:182-184,723-725) committing offset at poll time, BEFORE delivery: bad-payload docs in batch silently, permanently dropped. Task survives, no redelivery, no crash. Fix:
  drop-and-continue per bad message inside consume() (mirror runtime's own decode/transform skip pattern, sink.rs:584-592)

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+        let primary_key = config
+            .primary_key
+            .filter(|value| !value.trim().is_empty())
+            .unwrap_or_else(|| DEFAULT_PRIMARY_KEY.to_string());
+        let document_action = config.document_action.unwrap_or_default();
+        let create_index_if_not_exists = config
+            .create_index_if_not_exists
+            .unwrap_or(DEFAULT_CREATE_INDEX_IF_NOT_EXISTS);
+        let include_metadata = config.include_metadata.unwrap_or(DEFAULT_INCLUDE_METADATA);
+        let batch_size = config.batch_size.unwrap_or(DEFAULT_BATCH_SIZE).max(1);
```
ryerraguntla, line 137: 137-148,150-178,295-312  check_connectivity/get_primary_key/search_documents/create_client zero tokio::time::timeout, zero retry, no timeout config field at all. Sink
  wraps every SDK call via retry_sdk_operation's tokio::time::timeout (sink lib.rs:486-539). Unreachable/hung Meilisearch → open() hangs forever, poll() hangs forever. Fix: add timeout config + wrap, mirror sink.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+        match value {
+            Value::Object(_) => value,
+            other => {
+                let mut object = Map::new();
+                object.insert("value".to_string(), other);
+                Value::Object(object)
+            }
+        }
+    }
+
```
ryerraguntla, line 377: 377-415 index_documents/index_document_chunk accumulate accepted across chunks, discarded on early ? return — partial-success count lost. Fix: return partial count alongside Err.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+                    warn!(
+                        "Meilisearch get index '{}' failed (attempt {attempt}/{max_attempts}): {error}. Retrying in {delay:?}...",
+                        self.config.index
+                    );
+                    tokio::time::sleep(delay).await;
+                }
+            }
+        }
+    }
+
```
ryerraguntla, line 254: create_index retries via retry_sdk_operation, but wait_for_index_creation_task (:417-484) poll loop doesn't — single transient error during wait permanently fails
  open() despite max_open_retries existing for this. Fix: wrap wait-loop's per-poll call in same retry.

core/integration/tests/connectors/fixtures/meilisearch/container.rs
```diff
@@ -0,0 +1,339 @@
...
+    error: Option<serde_json::Value>,
+}
+
+pub struct MeilisearchContainer {
+    #[allow(dead_code)]
+    container: ContainerAsync<GenericImage>,
+    pub base_url: String,
+}
+
+impl MeilisearchContainer {
```
ryerraguntla, line 65: MeilisearchContainer::start() missing .with_container_name(unique_container_name("meilisearch")). Breaks iggy-test-*  cleanup sweep. Fix: add it.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,604 @@
...
+
+    async fn search_documents(&self, client: &Client) -> Result<Vec<ProducedMessage>, Error> {
+        let last_primary_key = {
+            let state = self.state.lock().await;
+            state.last_primary_key.clone()
+        };
+        let primary_key = self.primary_key.as_deref().ok_or_else(|| {
+            Error::Connection("Meilisearch primary key is not initialized".to_string())
+        })?;
+        let filter_expression = self.filter_expression()?;
```
ryerraguntla, line 219: 189-191,206 cursor pagination sorts/filters via > on primary key; integration test pins getmeili/meilisearch:v1.13, which lacks > on string attrs (v1.15+ only). String PK → every poll past page 1 errors via map_sdk_error, poll() Err forever — permanent error loop. Fix: verify against pinned version / require numeric PK / bump version + test string PK.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,978 @@
...
+        state.errors_count += 1;
+    }
+}
+
+#[async_trait]
+impl Sink for MeilisearchSink {
+    async fn open(&mut self) -> Result<(), Error> {
+        info!(
+            "Opening Meilisearch sink connector with ID: {} for URL: {}, index: {}",
+            self.id,
```
ryerraguntla, line 618: line : 553 sanitize_url_for_log called on raw un-normalized config.url BEFORE normalize_host/create_client. Url::parse (:677-689) requires scheme; schemeless URL w/ creds  (user:pass@host:port) → parse fails → fallback returns raw string unredacted, logged via info!. Fix: normalize (add default scheme) before sanitizing.

countradooku:
Superseded by the split connector PRs: #3497 for the Meilisearch sink and #3498 for the Meilisearch source.

PR #3407, opened 2026-06-02
fix(server): correct VSR client session routing and lifecycle

numinnex:
- Drop replicated sessions cluster-wide on disconnect. A lost connection now routes a synthetic `Logout` to the metadata owner, so the replicated session slot is released everywhere instead of lingering until peer
  eviction.

- Mutate the SessionManager only after the register commits. Login/register no longer flips the connection to `Authenticated` before submit (with rollback on failure); it submits, awaits the commit (post-quorum),
  then transitions `Connected → Bound` in one step. A transient submit failure leaves the connection untouched and the SDK replays.

- Split the`bootstrap.rs` into focused modules — `wire`, `pat`, `responses`, `auth`, `dispatch` — leaving `bootstrap.rs` with cluster boot/recovery/startup only. The request handlers, owner-forwarders
  (`submit_*_on_owner`), wire-response builders, and credential verification now live next to each other.


core/server-ng/src/dispatch.rs
```diff
@@ -0,0 +1,913 @@
...
+                // replicas drop the slot too. Without this, an
+                // already-committed `Register` lingers on every backup
+                // until they evict the client themselves -- a
+                // local-vs-cluster divergence observable across view
+                // changes. `submit_logout_in_process` requires shard 0
+                // consensus, which this handler always runs on.
+                shard_for_disconnect
+                    .plane
+                    .metadata()
+                    .remove_client_session(vsr_client_id);
```
hubcio: on an abrupt disconnect of a shard-0-homed client this removes the local clienttable slot synchronously, before the detached `submit_logout_in_process` task runs. by the time that task executes `get_session` already returns none, so it short-circuits and returns the current commit without ever proposing a `Logout` - no logout is replicated and peer replicas keep the orphan session slot until they evict it themselves. that's the exact divergence this pr set out to remove, and it hits every quic / tcp-tls client since those terminate on shard 0. the explicit-logout path (`handle_logout_request`) gets the order right: submit first, drop the connection after. fix: don't call `remove_client_session` before the submit - let the committed `Logout` apply drop the slot on shard 0 too (shard 0 is itself a replica), and only drop the `SessionManager` connection synchronously here.

core/server-ng/src/responses.rs
```diff
@@ -0,0 +1,752 @@
...
+    request_header: &RequestHeader,
+    committed: Message<GenericHeader>,
+    raw_token: Option<String>,
+) -> Result<Message<GenericHeader>, IggyError> {
+    let Some(raw) = raw_token else {
+        return Ok(committed);
+    };
+    let header_len = std::mem::size_of::<ReplyHeader>();
+    let committed_header =
+        bytemuck::checked::try_from_bytes::<ReplyHeader>(&committed.as_slice()[..header_len])
```
hubcio, line 689: this casts the committed bytes to `ReplyHeader` unconditionally, but `submit_request_in_process` hands back an `EvictionHeader`-backed message on the evict outcome. for a `CreatePersonalAccessToken` whose session got evicted between bind and request, `raw_token` is `Some`, so we land here on the eviction frame. the checked cast still passes (the eviction byte pattern is a valid `ReplyHeader` - `command` = eviction is a valid discriminant, `operation` falls on zeroed bytes = reserved), so `validate()` never runs. result: the eviction is swallowed (client gets a normal reply and never learns its session died) and we ship a raw pat whose hash never committed - a token that authenticates against nothing. gate on `committed.header().command == Command2::Reply` via the typed `header()` and pass the eviction through untouched otherwise.

core/server-ng/src/dispatch.rs
```diff
@@ -0,0 +1,913 @@
...
+                let shard = Rc::clone(&shard_for_disconnect);
+                compio::runtime::spawn(async move {
+                    // request_id 0 + session 0: synthetic disconnect
+                    // logout; `submit_logout_in_process` short-circuits to
+                    // current commit when no session entry exists, so a
+                    // race with the explicit logout path is a no-op.
+                    if let Err(error) = shard
+                        .plane
+                        .metadata()
+                        .submit_logout_in_process(vsr_client_id, 0, 0)
```
hubcio: the synthetic disconnect logout goes through with session 0, and the logout apply does an unconditional `remove_client(header.client)` with no session-match check. if the same vsr client id ever re-registered before a late in-flight disconnect-logout committed, the stale logout would wipe the fresh session. not reachable today because the sdk mints a fresh random client id on every reconnect (`reset_vsr_session`), so the stale logout targets a dead key - but a custom client reusing an id would hit it. cheap guard: make the logout apply a no-op when the current session != the logged-out session.

core/shard/src/lib.rs
```diff
@@ -542,6 +583,61 @@ where
...
+            }
+        }
+        // Drop the local handle so `recv` returns `Err` once every shard's
+        // reply sender is dropped (defensive; we also bound by count).
+        drop(reply_tx);
+
+        let mut clients = Vec::new();
+        let mut received = 0usize;
+        while received < expected {
+            match compio::time::timeout(LIST_CLIENTS_GATHER_TIMEOUT, reply_rx.recv()).await {
```
hubcio: the timeout wraps each individual `recv()`, not the whole gather, so worst-case wall time is `expected * 3s`, not the 3s the constant implies. it's a cold admin-only read and a wedged shard still caps at one timeout via the break, so impact is small, but a single deadline across the loop would match the intent of a per-gather budget.

core/metadata/src/impls/metadata.rs
```diff
@@ -1136,11 +1140,27 @@ where
...
-            return Err(RegisterSubmitError::Canceled);
+        // Dedup / session / eviction. shard 0 cannot route by the VSR
+        // consensus `client_id` (its top bits are random, not home-shard
+        // routing), so a Replay/Evict is returned to the home shard as the
+        // reply -- `handle_client_request` writes it to the originating socket
+        // by transport id, exactly like a fresh commit. Drop surfaces as
+        // Canceled so the home shard stays silent and the SDK replays.
+        match request_preflight(consensus, &self.client_table, client_id, session, request) {
+            PreflightOutcome::Dispatch => {}
+            PreflightOutcome::Replay(reply) => {
```
hubcio, line 1156: the replay arm copies the cached reply into a fresh owned buffer and re-runs the checked decode + validate on already-committed bytes, where the old wire path handed the same frozen buffer back by refcount. it's the duplicate-retry path only (cold, metadata control plane), it's structurally forced by the `Message<GenericHeader>` owned return type, and it replaces a send that was misrouted by the vsr client id and never reached the client anyway - so net it's a correctness fix, not a regression. only worth a frozen-backed return variant if this ever shows up hot.

core/server-ng/src/dispatch.rs
```diff
@@ -0,0 +1,913 @@
...
+                );
+                return;
+            }
+        }
+    } else {
+        None
+    };
+    let request = request.transmute_header(|header, new_header: &mut RequestHeader| {
+        *new_header = header;
+        new_header.client = transport_client_id;
```
hubcio: dead store - `bound` is always `Some` past the unbound early-return above, so the `client` field set here is overwritten a few lines down. the namespace write next to it is live; just this one line can go.

core/server-ng/src/dispatch.rs
```diff
@@ -0,0 +1,913 @@
...
+
+    let header = *request.header();
+    if header.operation == Operation::NonReplicated {
+        // Auth bypass guard: only `PING` and `GET_CLUSTER_METADATA` are
+        // legitimately pre-auth (liveness probe + connection bootstrap
+        // metadata). Every other non-replicated code (`GET_STREAM*`,
+        // `GET_TOPIC*`, `GET_STATS`, `POLL_MESSAGES`) reads live state and
+        // MUST go through Register first, since server-ng has no
+        // per-resource authz layer.
+        let nr_code = u32::from_le_bytes(
```
hubcio: this decodes the same 4-byte slice as the site further down but with `unwrap_or_default()` vs `unwrap()`. the slice is provably 4 bytes so the default arm is dead either way - not an auth-bypass (a default 0 isn't ping, it falls through to the drop), just an inconsistency. matching the `unwrap()` below keeps it uniform.

core/server-ng/src/pat.rs
```diff
@@ -0,0 +1,128 @@
...
+
+/// Mints a fresh PAT and returns the raw token plus its hex-encoded SHA-256
+/// hash (64 bytes ASCII). Only the hash is replicated; the raw token is
+/// returned to the minting client by the home shard (it cannot be reproduced
+/// by the deterministic `apply` running on every replica).
+fn mint_pat_raw_and_hash() -> (String, [u8; 64]) {
+    let (raw, hash) = iggy_common::PersonalAccessToken::mint_raw_and_hash();
+    let bytes = hash.as_bytes();
+    let mut out = [0u8; 64];
+    let len = bytes.len().min(64);
```
hubcio: the `.min(64)` clamp guards against a short hash, but `calculate_256` is blake3 hex - always exactly 64 chars - so the clamp can never truncate. if it somehow did, it would silently ship a wrong-length hash. `copy_from_slice` (panic on mismatch) or a `debug_assert_eq!(len, 64)` states the invariant instead of hiding it.

core/server-ng/src/dispatch.rs
```diff
@@ -0,0 +1,913 @@
...
+        GET_CLIENT_CODE => {
+            // No reverse map from the wire u32 id to a u128 transport id /
+            // home shard (the u32 is just the seq tail), so gather all and
+            // filter -- same fan-out as `get_clients`.
+            let target = GetClientRequest::decode_from(request_body(&request))
+                .ok()
+                .map(|req| req.client_id);
+            let infos = shard.list_all_clients().await;
+            #[allow(clippy::cast_possible_truncation)]
+            let found = target.and_then(|id| infos.iter().find(|info| info.client_id as u32 == id));
```
hubcio, line 538: `get_client` matches with `info.client_id as u32 == id` and takes the first hit, which is order-nondeterministic if two u128 transport ids collide in their low 32 bits. needs 2^32 concurrent clients to actually happen and it's a diagnostic-only read, so mostly a note - the u32 on the wire is the real limit here.

core/server-ng/src/session_manager.rs
```diff
@@ -60,9 +61,28 @@ pub enum ConnectionState {
...
+/// Flattened view of one connected client for the `get_me` /
+/// `get_clients` reads.
+///
+/// Holds everything those responses need from a single per-shard source,
+/// so the caller no longer also consults the message bus. `consumer_groups`
+/// membership is not modelled yet (see the TODO in the `get_clients`
+/// response builder); it ships empty until the partition-reconciliation /
+/// consumer-group-rebalancing work lands.
+#[derive(Debug, Clone)]
+pub struct ClientRecord {
```
hubcio: `ClientRecord` is field-for-field identical to `shard::ConnectedClientInfo`, and the two `*_to_response` helpers are byte-identical. server-ng already depends on the shard crate, so `SessionManager` could yield `ConnectedClientInfo` directly and drop `ClientRecord` plus one helper and the manual field map in the list-clients handler - three near-identical artifacts collapse into one.

core/server-ng/src/dispatch.rs
```diff
@@ -0,0 +1,913 @@
...
+) -> RequestHandler {
+    let shard = Rc::clone(shard);
+    let sessions = Rc::clone(sessions);
+    let queues: ClientRequestQueues = Rc::new(RefCell::new(HashMap::new()));
+    let active: ActiveClientRequests = Rc::new(RefCell::new(HashSet::new()));
+    let sessions_for_disconnect = Rc::clone(&sessions);
+    let shard_for_disconnect = Rc::clone(&shard);
+    shard
+        .bus
+        .set_client_connection_lost_fn(Rc::new(move |client_id| {
```
hubcio, line 78: this disconnect closure and the peer one further down duplicate the same remove + detached-submit + identical warn body. a shared `submit_disconnect_logout(shard, vsr_client_id)` helper would dedup it (`submit_logout_on_owner` already branches on shard 0 internally). note it only removes the duplication - it doesn't by itself fix the shard-0 short-circuit flagged above.

hubcio (changes requested):
a few findings landed on unchanged context lines, so they couldn't be attached inline - noting them here:

- `login_register.rs:47` - `InvalidClientId` is never constructed in production (only the enum def, the `TryFrom` arm and a test). the client_id==0 path it documents is rejected upstream by the header validation and would hit the `assert!(client_id != 0)` first. reads like deliberate non_exhaustive scaffolding, but if it isn't pinning a future contract it can go.

- `login_register.rs:104` - the doc on `TryFrom<&LoginRegisterError> for EvictionReason` says 'no production caller yet', but `surface_login_failure` calls it via `.is_ok()` on the live login/register branches. just drop the stale claim.

- `login_register.rs:97` - that `TryFrom` mapping is only consumed via `.is_ok()` today; a `fn is_terminal(&LoginRegisterError) -> bool` would do the same and save the mapping plus the `NotEvictable` type. documented as scaffolding for the eviction-frame dispatcher though, so fine to keep if that's landing soon.

- `session_manager.rs:250` - `get_connection` / `connection_for_client` / `bound_count` / `connection_count` have no production callers (tests only), and `connection_for_client`'s 'for routing consensus replies' doc is obsolete now that the home shard routes by transport id. safe to delete - keep the `client_to_connection` map itself, `bind_session` still needs it. pre-existing, not introduced here.

hubcio (approved):

krishvishal (approved):

PR #3408, opened 2026-06-02
ci(php): add SDK coverage report

### LINKED ISSUE #3307: PHP SDK: add coverage, test reports, and Codecov support

The PHP pre-merge test path currently builds the extension with a plain `cargo build --manifest-path foreign/php/Cargo.toml`, runs PHPUnit without producing a JUnit report, does not upload PHP test/coverage artifacts, and `_test.yml` has no PHP Codecov upload step. As a result, PHP changes do not feed the same CI reporting/dashboard path as Python and Node.

## Scope

Coverage instrumentation:

- Install / make available `cargo llvm-cov` for the PHP pre-merge action, following the Python action pattern.
- Wrap the test extension build in `.github/actions/php/pre-merge/action.yml` with `cargo llvm-cov` environment setup at the current `cargo build --manifest-path foreign/php/Cargo.toml` point.
- Add a `cargo llvm-cov report --lcov` step after the PHP SDK tests run and before the Iggy server stop step.
- Emit `reports/php-coverage.lcov` or another stable LCOV path for PHP coverage.
- Adjust source paths if needed so Codecov attributes Rust extension coverage under `foreign/php/src/`.

Test reports and artifacts:

- Pass a JUnit output path to PHPUnit, for example:

```sh
./scripts/test.sh --log-junit ../../reports/php-junit.xml
```

  or wire the equivalent through `foreign/php/scripts/test.sh` / the PHP action.
- Ensure the reports directory exists before test/report generation.
- Upload the PHP JUnit XML and PHP LCOV file via `actions/upload-artifact`.

Codecov wiring:

- Add a PHP Codecov upload call in `.github/workflows/_test.yml`, matching the existing Python / Node shape:
  - `files: reports/php-coverage.lcov`
  - `flags: php`
  - `disable_search: true`
  - preserve the existing Codecov token / PR override conventions.
- Also modify `.github/workflows/coverage-baseline.yml`, which runs post-merge, so PHP coverage establishes the baseline dashboard data.

## Files of interest

- `.github/actions/php/pre-merge/action.yml`
- `foreign/php/scripts/test.sh`
- `.github/workflows/_test.yml`
- `.github/workflows/coverage-baseline.yml`

## Reference

The Python pre-merge action already wraps the equivalent Rust extension build with `cargo llvm-cov`, emits `reports/python-coverage.lcov`, uploads JUnit/coverage artifacts, and feeds Codecov. The PHP action should follow the same shape so PHP changes are visible in CI reporting and coverage dashboards.

Merged scope from duplicate issue: #3309
Related: #3302
Initial PHP SDK work: #3235

WaterWhisperer:

Closes #3307 

## Rationale

PHP SDK tests should feed the same JUnit, artifact, and Codecov reporting path as the other SDKs

## What changed?

The PHP pre-merge test path built the Rust extension with a plain cargo build, ran PHPUnit without JUnit output, and didn't upload PHP coverage to Codecov.

The test build now uses `cargo-llvm-cov` instrumentation, PHPUnit writes a stable JUnit report, and the PHP action emits an lcov report with source paths normalized under `foreign/php/src/`

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. tools: Codex
2. scope: inspect the issue, compare existing CI patterns, run local verification
3. verify: with PHP 8.3 Docker execution, PHPUnit, lcov source-path checks
4. Yes


.github/actions/php/pre-merge/action.yml
```diff
@@ -180,11 +195,48 @@ runs:
...
+      shell: bash
+      run: |
+        mkdir -p reports
+        cd foreign/php
+        source <(cargo llvm-cov show-env --sh)
+        export CARGO_TARGET_DIR=$CARGO_LLVM_COV_TARGET_DIR
+
+        cargo llvm-cov report --lcov \
+          --ignore-filename-regex='(\.cargo/|/rustc/|/core/)' \
+          --output-path ../../reports/php-coverage.lcov
```
hubcio, line 210: no non-emptiness check on the generated lcov - the step ends with a cosmetic `wc -l` echo and never fails on a degenerate report. combined with `fail_ci_if_error: false` on the codecov upload and `if-no-files-found: ignore` on the artifact, a stale or mis-pathed `.so`, or an `LLVM_PROFILE_FILE` / target-dir mismatch between the two separate `cargo llvm-cov show-env` blocks, could silently upload near-0% coverage on a green build. worth a small gate that fails if the lcov has zero `SF:`/`DA:` lines. note the python recipe has the same gap, so ideally fix it for both rather than php-only.

.github/actions/php/pre-merge/action.yml
```diff
@@ -180,11 +195,48 @@ runs:
         IGGY_PORT: 8090
         IGGY_TLS_CONNECTION_STRING: iggy+tcp://iggy:iggy@127.0.0.1:8090?tls=true&tls_domain=localhost&tls_ca_file=${{ github.workspace }}/core/certs/iggy_ca_cert.pem
         IGGY_TLS_PLAINTEXT_ADDRESS: 127.0.0.1:8090
-      run: ./scripts/test.sh tests/TlsTest.php
+      run: ./scripts/test.sh --log-junit ../../reports/php-tls-junit.xml tests/TlsTest.php
+
+    - name: Generate PHP coverage report
+      if: inputs.task == 'test'
```
hubcio: this step and the tls test above both use a bare `if: inputs.task == 'test'`, which github implicitly treats as `success() && ...`. since `test.sh` is `set -euo pipefail`, a tls-only failure (after the plain run already passed) aborts the composite and skips this step - the plain-run profraw is already on disk but the lcov is never written, so that coverage is thrown away. only bites an already-failing build, but `if: always() && inputs.task == 'test'` here would keep the plain-run coverage.

.github/actions/php/pre-merge/action.yml
```diff
@@ -180,11 +195,48 @@ runs:
...
+      shell: bash
+      run: |
+        mkdir -p reports
```
hubcio: redundant - `../../reports` is already created by the plain test step above, and this step only runs if that one succeeded. can drop this `mkdir`.

hubcio (changes requested):
the pr adds php coverage uploads (`flags: php` in `_test.yml` and `coverage-baseline.yml`) and even rewrites the coverage-baseline header to "all 7 languages", but `codecov.yml` is never updated. php is missing from both `flag_management.individual_flags` and `component_management.individual_components`, where all 6 other sdks (rust/java/csharp/python/node/go) are registered.

effect: the php flag falls through to default_rules so carryforward still works (no data loss), but it has no path scoping and there's no "PHP SDK" row in the codecov pr comment - the components table only renders registered components. so the coverage uploads but stays invisible.

add php to both blocks, mirroring the siblings:
```yaml
# flag_management.individual_flags:
  - name: php
    paths:
      - foreign/php/
# component_management.individual_components:
  - component_id: php
    name: "PHP SDK"
    paths:
      - foreign/php/
```
(`codecov.yml` is outside this pr's current file set, so this is the one consumer-side change still needed for full parity.)

.github/actions/php/pre-merge/action.yml
WaterWhisperer: thanks

hubcio (dismissed):

numinnex (dismissed):

spetz (dismissed):

hubcio:
@WaterWhisperer can you fix CI? looks like after github rebase the deps are not sorted.

WaterWhisperer:
> @WaterWhisperer can you fix CI? looks like after github rebase the deps are not sorted.

I checked the failed job. It ran out of disk while restoring the Rust cache before `cargo-sort` ran, so I disabled cache restore for the sort task and verified locally. I'm not sure if this is correct to fix it, so I can drop the change if you'd rather just rerun the failed job.

hubcio:
it's fine, i wanted to do it for most of sanity / lightweight jobs @WaterWhisperer

hubcio (approved):

spetz (approved):

PR #3410, opened 2026-06-03
test(php): add BDD suite wiring

### LINKED ISSUE #3310: PHP SDK: add BDD scenarios and bdd-php component

Every other SDK has BDD coverage wired through repository component detection, but the PHP SDK does not yet have `bdd/php/` scenarios or a `bdd-php` component entry. This is fine as a follow-up and not a merge blocker, but should be tracked so it does not get lost.

## Scope

- Add a `bdd/php/` test suite following the shape of the other SDK BDD directories.
- Add a `bdd-php` component entry in `.github/config/components.yml`.
- Wire the component paths so PHP BDD changes trigger the expected CI tasks.
- Ensure the PHP SDK component dependencies include the right Rust SDK/server/CI infrastructure dependencies for BDD execution.

## References

- `bdd/python/`, `bdd/node/`, `bdd/go/`, `bdd/java/`, `bdd/csharp/`, `bdd/rust/`
- `.github/config/components.yml`

Related: #3302
Initial PHP SDK work: #3235

countradooku:
## Summary

- add `bdd/php` with a PHPUnit implementation of the shared basic messaging BDD scenario
- wire `php-bdd` into the shared Docker Compose BDD runner and `scripts/run-bdd-tests.sh`
- add `bdd-php` component detection and include PHP BDD paths in `sdk-php`

Closes #3310.

## Verification

- `php -l bdd/php/tests/BasicMessagingFeatureTest.php`
- parsed `bdd/php/phpunit.xml.dist` with PHP `SimpleXMLElement`
- `bash -n scripts/run-bdd-tests.sh`
- `git diff --check`
- `docker compose -f bdd/docker-compose.yml config`
- component YAML validation for `bdd-php` task/dependencies
- PHP BDD container against local Iggy server: `OK (1 test, 38 assertions)`

## Notes

A full local `./scripts/run-bdd-tests.sh php` run on macOS failed before PHP tests because the shared BDD server image copied local `target/debug` Mach-O binaries into a Linux container. CI builds Linux binaries before invoking the BDD runner, so the shared path should be valid there.


licenserc.toml
```diff
@@ -76,6 +76,7 @@ excludes = [
     "**/.env",
     "**/.gitkeep",
     "**/META-INF/services/**",
+    "bdd/php/tests/BasicMessagingFeatureTest.php",
```
hubcio: this excludes the test from license-header enforcement permanently. root cause is header style: `licenserc.toml` maps `php` to `DOUBLESLASH_STYLE` (line 88) so hawkeye wants `//` headers, but this file uses a `/* */` block header. every other tracked `.php` (e.g. `foreign/php/tests/bootstrap.php`) uses `//`. instead of disabling the gate on a brand-new source file, switch the header to `//` and drop this exclude - hawkeye passes that way.

.github/actions/php/pre-merge/action.yml
```diff
@@ -58,7 +58,7 @@ runs:
     - name: Setup Rust with cache
       uses: ./.github/actions/utils/setup-rust-with-cache
       with:
-        shared-key: dev
+        shared-key: php
```
hubcio: this switches the rust cache `shared-key` from `dev` to `php`, but the job is read-only (`save-cache: false` right below) and nothing in the repo writes a `php` key - the only cache writers live in `coverage-baseline.yml` and they save under the default `dev`. net effect: the php pre-merge job now misses the rust cache every run and does a full cold build, where `dev` used to be warm.

we dont need php cache, its fast enough. drop this change.

bdd/php/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+COPY foreign/php/Cargo.toml ./foreign/php/
+COPY foreign/php/composer.json ./foreign/php/
+COPY foreign/php/phpunit.xml.dist ./foreign/php/
+COPY foreign/php/README.md foreign/php/LICENSE foreign/php/NOTICE ./foreign/php/
+COPY foreign/php/.cargo/ ./foreign/php/.cargo/
+COPY foreign/php/src/ ./foreign/php/src/
+COPY foreign/php/tests/ ./foreign/php/tests/
+COPY foreign/php/scripts/ ./foreign/php/scripts/
+
+COPY bdd/php/ ./bdd/php/
```
hubcio, line 72: the image never copies `bdd/scenarios`, so the feature file only exists via the compose volume mount. the `BDD_FEATURE_FILE` default and the in-test fallback path both point at locations that don't exist inside the image, so a bare `docker run` of this image can't find the feature. every other bdd image bakes the scenarios in (rust/python/java via `COPY . .`, go/node/csharp explicitly). add `COPY bdd/scenarios ./bdd/scenarios` for parity.

bdd/php/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+WORKDIR /workspace/foreign/php
+
+RUN cargo install cargo-php --locked --version 0.1.21 \
```
hubcio: `cargo install cargo-php` sits after all the `COPY` layers, but it has no dependency on the copied workspace. any source edit busts the cache here and rebuilds cargo-php from source on every local/cached build. move this `RUN` above the `COPY` lines (right after the apt-get) so it caches independently.

bdd/docker-compose.yml
```diff
@@ -221,6 +221,25 @@ services:
     networks:
       - iggy-bdd-network
 
+  php-bdd:
```
hubcio, line 224: there's no `php-bdd` entry in `docker-compose.coverage.yml`, so `run-bdd-tests.sh --coverage php` (and the `all` path) runs this uninstrumented service and silently produces no coverage report. every other sdk has a coverage override.

add one for php or note that php bdd coverage is intentionally skipped. (and create gh issue for this)
or implement it now.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,278 @@
...
+                $this->client,
+                $this->lastStreamName,
+                $this->lastTopicName !== null ? [$this->lastTopicName] : [],
+            );
+        }
+
+        parent::tearDown();
+    }
+
+    private function scenarioSteps(): array
```
hubcio: this hand-rolls a gherkin parser plus a big `if`/`preg_match` dispatch chain. every other sdk uses a real cucumber binding (cucumber-rs, pytest-bdd, godog, cucumber-js, cucumber-jvm, Reqnroll); the php equivalent is Behat and it isn't pulled in. the runtime parse here is also mostly redundant - `runStep` already hardcodes every step regex, and the scenario title + feature path are hardcoded, so the parse buys very little. adopting Behat (or at least driving off the parsed steps properly) would replace this whole block and fix the silent-skip issues flagged below at once.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,278 @@
...
+        $steps = [];
+        $insideTargetScenario = false;
+        foreach ($lines as $line) {
+            $line = trim($line);
+            if ($line === '') {
+                continue;
+            }
+
+            if (str_starts_with($line, 'Scenario:')) {
+                $insideTargetScenario = $line === 'Scenario: Create stream and send messages';
```
hubcio: the parser only collects steps when the scenario title is exactly `Create stream and send messages`, and nothing asserts how many scenarios ran. the moment a second scenario is added to the shared feature, php silently skips it while rust (`World::run`) and python (`scenarios()`) run all of them. silent false-green against a feature file that's meant to be shared across sdks.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,278 @@
...
+            }
+
+            if (preg_match('/^(Given|When|Then|And) (.+)$/', $line, $matches) === 1) {
+                $steps[] = $matches[2];
+            }
+        }
+
+        assert_true($steps !== [], 'no BDD steps were loaded from the basic messaging scenario');
+
+        return [
```
hubcio: the `Background` steps are hardcoded and prepended manually here. the parser never reads the `Background` block (it skips everything before the `Scenario:` line), so if a background step is added or changed in the shared feature, php won't pick it up and the test stays green. drive the background off the parsed feature instead of hardcoding it.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,278 @@
...
+            if (str_starts_with($line, 'Scenario:')) {
+                $insideTargetScenario = $line === 'Scenario: Create stream and send messages';
+                continue;
+            }
+
+            if (!$insideTargetScenario) {
+                continue;
+            }
+
+            if (preg_match('/^(Given|When|Then|And) (.+)$/', $line, $matches) === 1) {
```
hubcio: this regex only matches `Given|When|Then|And`, but gherkin also allows `But` (and `*`). such a step returns `preg_match` 0 and gets dropped during parsing, so it never reaches the `self::fail` for unhandled steps - it just silently disappears. no `But` in the feature today, but worth widening the alternation.

bdd/php/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    ca-certificates \
+    clang \
+    composer \
+    curl \
+    git \
+    libclang-dev \
+    libssl-dev \
+    php-cli \
```
hubcio: php is unpinned here, so on `trixie` you get php 8.4 while the ci action pins `php8.3`. the bdd run and the sdk tests end up on different php minors, which can drift on a base-image bump. pin `php8.3-*` to match ci.

.github/config/components.yml
```diff
@@ -307,6 +307,18 @@ components:
...
+  bdd-php:
+    depends_on:
+      - "rust-server"
+      - "rust-sdk"
+      - "sdk-php"
+      - "bdd-infrastructure"
+      - "ci-infrastructure"
+    paths:
+      - "bdd/php/**"
+      - "bdd/scenarios/**"
```
hubcio: `bdd/scenarios/**` here means a `leader_redirection.feature` edit triggers the php bdd job even though php only runs `basic_messaging`, so it's wasted ci. mirrors the python component so not new, but worth narrowing if you want the trigger to be accurate.

scripts/ci/license-headers.sh
```diff
@@ -185,7 +185,10 @@ find_duplicate_license_headers() {
   local path
 
   : > "$output_file"
-  mapfile -t LICENSE_EXCLUDES < <(load_license_excludes)
+  LICENSE_EXCLUDES=()
```
hubcio: this mapfile->while-read swap is behavior-identical on bash 4 (it only helps bash 3.2 / older macos) and it's unrelated to adding the php bdd suite.

dont do that in this PR.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,278 @@
...
+            return;
+        }
+
+        if (preg_match('/^I poll messages from stream (\d+), topic (\d+), partition (\d+) starting from offset (\d+)$/', $step, $matches) === 1) {
+            $this->lastPolledMessages = $this->requireClient()->pollMessages(
+                (int) $matches[1],
+                (int) $matches[2],
+                (int) $matches[3],
+                PollingStrategy::offset((int) $matches[4]),
+                10,
```
hubcio: poll count is a hardcoded `10` while `lastSentMessageCount` is captured a few steps earlier and unused. happens to match the scenario so no live bug, just a magic number - derive it from the sent count instead.

bdd/php/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+
+WORKDIR /workspace/foreign/php
+
+RUN cargo install cargo-php --locked --version 0.1.21 \
+    && rm -rf /usr/local/cargo/registry /usr/local/cargo/git
+RUN composer install --no-interaction --prefer-dist
+RUN cargo php install --yes
+RUN chmod +x ./scripts/test.sh
+
+CMD ["./scripts/test.sh", "--configuration", "/workspace/bdd/php/phpunit.xml.dist", "/workspace/bdd/php/tests"]
```
hubcio: the command passes both `--configuration .../phpunit.xml.dist` and a positional `.../bdd/php/tests`. phpunit uses the positional path and ignores the `<testsuites>` block in the config, so that block is dead (harmless - both resolve to the same dir). drop the positional arg (here and in the compose `command`) to keep one source of truth.

bdd/php/Dockerfile
```diff
@@ -0,0 +1,66 @@
...
+ENV PHP=/usr/bin/php
+ENV PHP_CONFIG=/usr/bin/php-config
+ENV IGGY_HOST=iggy-server
+ENV IGGY_PORT=8090
+ENV BDD_FEATURE_FILE=/app/features/basic_messaging.feature
+
+WORKDIR /workspace
+
+COPY Cargo.toml Cargo.lock ./
+COPY core/ ./core/
```
hubcio: `COPY core/` pulls the whole workspace in to build a cdylib that only needs the `core/sdk` closure (`sdk` + `common` + `binary_protocol`). a change to any unrelated core crate busts this layer. you can narrow it to those three crates plus the root `Cargo.toml`/`Cargo.lock` (the root manifest has to stay for workspace resolution) - build-time only, no correctness impact.

hubcio (changes requested):

bdd/php/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+
+ENV PATH=/root/.cargo/bin:$PATH
+ENV PHP=/usr/bin/php8.3
+ENV PHP_CONFIG=/usr/bin/php-config8.3
+ENV IGGY_HOST=iggy-server
+ENV IGGY_PORT=8090
+ENV BDD_FEATURE_FILE=/app/features/basic_messaging.feature
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
+    | sh -s -- -y --profile minimal --default-toolchain 1.95.0 \
```
hubcio: the `--default-toolchain 1.95.0` rustup form here doesn't match the rest of the repo. master pins 1.96.0 in `rust-toolchain.toml`, and since the COPY block below never copies that file, this 1.95 is what actually builds the ext and the sdk crates. the other bdd images use `ARG RUST_VERSION` + `FROM rust:${RUST_VERSION}`, and `scripts/ci/sync-rustc-version.sh` only knows how to rewrite those two forms - it skips the rustup line entirely, so this pin is invisible to the version-sync gate and will silently drift on the next toolchain bump. switch to the `ARG RUST_VERSION` + `FROM rust:` shape the siblings use.

bdd/docker-compose.coverage.yml
```diff
@@ -45,6 +45,24 @@ services:
...
         --cov --cov-report=xml:/reports/python-bdd-coverage.xml
 
+  php-bdd:
+    volumes:
+      - ../reports:/reports
+    command:
+      - bash
+      - -lc
+      - >-
+        cargo install cargo-llvm-cov --locked
```
hubcio: this coverage path can't actually produce a report as-is. the image installs rust with `--profile minimal` (Dockerfile:45) and never adds `llvm-tools-preview`, but `cargo llvm-cov` needs it - rust-bdd adds the component explicitly and php doesn't, so `cargo llvm-cov show-env` and `report` here will fail. separately, `cargo install cargo-llvm-cov` runs at container runtime over the network on every coverage run - bake it into the image instead.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,324 @@
...
+            assert_same($matches[1], $topic->name);
+
+            return;
+        }
+
+        if (preg_match('/^the topic should have (\d+) partitions$/', $step, $matches) === 1) {
+            $topic = $this->requireClient()->getTopic($this->requireStreamId(), $this->requireTopicName());
+            assert_not_null($topic);
+            assert_same((int) $matches[1], $topic->partitions_count);
+            assert_same((int) $matches[1], $this->lastTopicPartitions);
```
hubcio: this assert is redundant - line 216 already checks `topic->partitions_count` from the server response against the expected count, so re-checking the locally-captured `lastTopicPartitions` against the same value adds nothing. you can drop both this line and the `lastTopicPartitions` field.

bdd/php/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+COPY foreign/php/tests/ ./foreign/php/tests/
+COPY foreign/php/scripts/ ./foreign/php/scripts/
+
+COPY bdd/php/ ./bdd/php/
+COPY bdd/scenarios/ ./bdd/scenarios/
+
+WORKDIR /workspace/foreign/php
+
+RUN composer install --no-interaction --prefer-dist
+RUN cargo php install --yes
```
hubcio, line 78: `cargo php install` recompiles the sdk closure (binary_protocol + common + sdk) on every `up --build`, and line 47 wipes the cargo registry cache so there's no dep cache to reuse across builds. a cache mount here would avoid the repeated cold rebuild. test-only so not urgent.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,324 @@
...
+                    $scenarios[$currentScenario] = $currentSteps;
+                }
+
+                $currentScenario = trim(substr($line, strlen('Scenario:')));
+                $currentSteps = [];
+                $section = 'scenario';
+                continue;
+            }
+
+            if (preg_match('/^(Given|When|Then|And|But|\*) (.+)$/', $line, $matches) !== 1) {
```
hubcio: step lines that appear before the first `Background:` or `Scenario:` get silently dropped here - `section` is still null so neither branch appends them, and they never reach the `self::fail` for unknown steps. the feature has none today, but a misplaced step in the shared file would vanish instead of failing. cheap to guard: fail when `section` is null and the line matches the step regex.

hubcio (changes requested):

bdd/php/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+ENV CARGO_HOME=/usr/local/cargo
+ENV RUSTUP_HOME=/usr/local/rustup
+ENV PATH=/usr/local/cargo/bin:$PATH
+ENV PHP=/usr/bin/php8.3
+ENV PHP_CONFIG=/usr/bin/php-config8.3
+ENV IGGY_HOST=iggy-server
+ENV IGGY_PORT=8090
+ENV BDD_FEATURE_FILE=/app/features/basic_messaging.feature
```
hubcio: this `ENV` still points at `/app/features/...`, which only exists via the compose mount - so a bare `docker run` of this image throws `feature file not found` even though the scenarios are now baked in at `/workspace/bdd/scenarios`. the env override preempts the in-test fallback path. point it at `/workspace/bdd/scenarios/basic_messaging.feature` (compose already sets its own value) or drop the `ENV` entirely and let the fallback resolve.

scripts/ci/sync-rustc-version.sh
```diff
@@ -101,7 +101,7 @@ for dockerfile in $DOCKERFILES; do
         EXPECTED_VERSION="$RUST_VERSION_SHORT"
     elif grep -qE "FROM[[:space:]].*\brust:[0-9]" "$dockerfile" 2>/dev/null; then
         SOURCE="from"
-        CURRENT_VERSION=$(grep -E "FROM[[:space:]].*\brust:[0-9]" "$dockerfile" | head -1 | sed -nE 's/.*\brust:([0-9]+\.[0-9]+(\.[0-9]+)?).*/\1/p')
+        CURRENT_VERSION=$(grep -E "FROM[[:space:]].*\brust:[0-9]" "$dockerfile" | head -1 | sed -nE 's/.*rust:([0-9]+\.[0-9]+(\.[0-9]+)?).*/\1/p')
```
hubcio: also at line 129.

why drop `\b` here? the php Dockerfile uses the `ARG RUST_VERSION` form, so neither of these sed lines ever runs for it - this change is unrelated to the PR. it also loosens fix mode: `s/(rust:)[0-9]+.../g` now rewrites any image name that merely ends in `rust:`. revert, or if there's a concrete reason (BSD sed on macOS not supporting `\b`?), say so in the commit.

bdd/php/tests/BasicMessagingFeatureTest.php
```diff
@@ -0,0 +1,326 @@
...
+                    $scenarios[$currentScenario] = $currentSteps;
+                }
+
+                $currentScenario = trim(substr($line, strlen('Scenario:')));
+                $currentSteps = [];
+                $section = 'scenario';
+                continue;
+            }
+
+            if (preg_match($stepPattern, $line, $matches) !== 1) {
```
hubcio, line 107: structural gherkin keywords fall through this regex silently - `Scenario Outline:`, `Rule:`, `Examples:` and `|` table rows all get dropped, and outline steps would attach to the previous scenario since `section` stays `'scenario'`. the shared feature file is meant to evolve across sdks, so failing fast on lines that start with these known structural keywords is cheap and keeps this honest.

hubcio (changes requested):
`bdd/README.md` isn't updated - php is missing from the structure tree and the usage examples while every other sdk is listed.

hubcio (approved):

spetz (approved):

PR #3425, opened 2026-06-05
feat(cluster): authenticate replica peers with PSK + BLAKE3 handshake

hubcio:
The server-ng replica port (tcp_replica) was plaintext: any TCP peer
that learned the cluster id could inject VSR frames or register as an
arbitrary replica. mTLS does not fit - the replica conn is a dup'd
plaintext fd round-robined across shards, state rustls cannot carry.

Authenticate with a pre-shared cluster key and a 3-message mutual
BLAKE3 keyed-MAC handshake (ReplicaHello / ReplicaChallenge /
ReplicaFinish) over the reserved GenericHeader bytes: no new typed
header, the stream stays a dupable plaintext fd. These are dedicated
Command2 variants, not a Ping/Pong squat (left free for a future VSR
liveness ping). ReplicaChallenge carries a status field, so a reject is
the same frame with a nonzero status and the finish is identified by
command, not position. The MAC proves PSK possession (cluster
membership, not per-replica identity - the registry still trusts the
announced id, so keep the port on a trusted boundary). The PSK is never
serialized to disk.

Configured under [cluster.auth] (enabled, shared_secret). Enabling it,
and adding the consensus-plane commands, is a coordinated-restart change
(cluster_id is derived from the cluster name).


core/message_bus/src/replica/listener.rs
```diff
@@ -210,30 +238,215 @@ pub async fn run(
...
+        return reject(
+            stream,
+            our_cluster,
+            self_id,
+            peer_id,
+            HandshakeStatus::DirectionalRule,
+            nackable,
+        )
+        .await;
+    }
```
numinnex, line 318: This is true for initial cluster bootstrapping, but for already existing clusters that expand their replica set enforcing this rule could be problematic, as rather having the new node initiate the challenge request, all of the other nodes in the cluster would have to do it.

numinnex (changes requested):

numinnex:
Ok, the idea is sound and looks good, there is one piece that I think is worth mentioning, `replica_id` generation. 

Currently we are supplying the replica_id through an CLI argument, it's fine for an "seed" cluster values (e.g perfect information first bootstrap), but falls flat in cases where we would like to have dynamic cluster counts.

One way to address this isssue would be to keep the initial seed, but on the first bootstrap (when the metadata log is empty, so the cluster is fresh), write that seed configuration into the metadata log with op=0. Make the `replica_id` be generated at runtime, rather than supplied (the one comming from CLI arg would be used as an index into the `replicas` set, so the initial seed configuration has granular control over all of the nodes required for determinism). 
After the initial configuration is written to the log and resolved, we can go through the `challenge` protocol with those nodes). I omitted one important detail which is `ReplicaJoin` command, as this is the command that will assign the `replica_id` (it has to be written to the log and replicated).

numinnex (approved):

spetz (approved):

PR #3429, opened 2026-06-06
feat(sdk): support client-level and per-request timeouts (default: not timeout -> 30 s)

### LINKED ISSUE #3419: Support per-request timeout to TCP/QUIC/WebSocket/Http clients
## Summary

I found this problem when I try to implement a new scenario in /bdd for rust sdk, which define the behavior of sdk when the network/server is down, and realized that the client would stuck forever which doesn't seem appropriate?

The Rust SDK (and other SDKs inheriting the same pattern) has no per-request timeout on TCP/QUIC/WebSocket transports. When the server becomes unreachable after a connection is established, in-flight requests hang indefinitely inside `send_raw_with_response` because `stream.read()` has no deadline.

## Reproduce

```rust
use iggy::prelude::*;
use std::process::Command;
use std::time::Duration;

#[tokio::main]
async fn main() {
    // Start server
    Command::new("docker")
        .args(["run", "-d", "--name", "iggy-server",
               "--security-opt", "seccomp=unconfined",
               "-p", "8090:8090", "apache/iggy:latest"])
        .status().unwrap();
    loop {
        if tokio::net::TcpStream::connect("127.0.0.1:8090").await.is_ok() { break; }
        tokio::time::sleep(Duration::from_secs(1)).await;
    }

    let client = IggyClientBuilder::new()
        .with_tcp()
        .with_server_address("127.0.0.1:8090".parse().unwrap())
        .build().unwrap();
    client.connect().await.unwrap();

    // Kill server after connected
    Command::new("docker").args(["stop", "iggy-server"]).status().unwrap();

    // This hangs forever -- need Ctrl+C to exit
    let _ = client.login_user("iggy", "iggy").await;
}
```

## Root cause

`send_raw` itself has no read/write deadline on the TCP stream -- if the connection is alive but the server is unresponsive (e.g. paused process), `stream.read()` blocks until the OS TCP keepalive timeout.

## Comparison with Kafka (AI generated, didn't verify manually)

| Capability | Kafka (librdkafka) | Iggy |
|---|---|---|
| Per-request timeout | `request.timeout.ms` = 30s | None |
| Total delivery timeout | `message.timeout.ms` = 300s | None |
| Reconnect backoff | Exponential 100ms → 10s | Fixed 1s |
| Reconnect bound | Bounded by delivery timeout | Unlimited by default |

## Proposal

This could be exposed via the builder API:

Use client level timeout (default: 30 seconds per-request)

```rust
   let client = IggyClientBuilder::new()
          .with_tcp()
          .with_server_address("127.0.0.1:8090".to_string())
          .with_request_timeout(IggyDuration::from_str("5s")?)
          .build()?;
```

Use the per-request level is

```rust
client.ping().with_timeout(IggyDuration::from_str("10ms").unwrap())
```

The per-request level should override the client level timeout.

chengxilo:

Closes #3419

## Rationale

Clients had no deadline on individual request, so a stalled server could block a caller forever.

## What changed?

`send_raw` would be blocked forever if server doesn't response. I added a request_timeout field (IggyDuration, default 30 s) to TcpClientConfig, QuicClientConfig, and WebSocketClientConfig. Each client's `send_raw` now wraps its I/O in `tokio::time::timeout` 

Use client level timeout
```rust
   let client = IggyClientBuilder::new()
          .with_tcp()
          .with_server_address("127.0.0.1:8090".to_string())
          .with_request_timeout(IggyDuration::from_str("5s")?)
          .build()?;
```

Use the per-request level is 
```rust 
client.ping().with_timeout(IggyDuration::from_str("10ms").unwrap())
```

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Claude Opus 4.6
2. implementation
3. manually reviewed
4. ~Yes~ Can I?

## Change Log

By default, Iggy Rust SDK would have a 30s timeout for each request.


core/common/src/types/configuration/quic_config/quic_client_config.rs
```diff
@@ -64,6 +66,7 @@ impl Default for QuicClientConfig {
             server_name: "localhost".to_string(),
             auto_login: AutoLogin::Disabled,
             heartbeat_interval: IggyDuration::from_str("5s").unwrap(),
+            request_timeout: IggyDuration::from_str("300s").unwrap(),
```
hubcio: 300s is IMHO overkill. make it 10s or so.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -663,111 +662,118 @@ impl TcpClient {
...
                 }
-            }
 
-            error!("Cannot send data. Client is not connected.");
-            Err(IggyError::NotConnected)
+                error!("Cannot send data. Client is not connected.");
+                Err(IggyError::NotConnected)
+            };
+
+            tokio::time::timeout(request_timeout.get_duration(), io)
```
hubcio: on timeout the `io` future is dropped while it still holds the stream lock, mid-write or after-write-before-read. tcp and websocket use one persistent stream with no request id, so responses are matched purely by arrival order. `RequestTimeout` isn't in the `send_raw_with_response` reconnect set, so the stream is kept and reused: the next `send_raw` reads the previous request's late response, or if the timeout hit mid-write, a partial frame is left on the wire. result is a permanent response/frame desync until the process restarts. quic avoids this because it opens a fresh `open_bi` stream per request.

adding `RequestTimeout` to the reconnect set alone isn't enough - the `!reconnection.enabled` and (under vsr) `auto_login == Disabled` early returns fire before `disconnect()`, which is the only thing that resets the stream. simplest robust fix: on timeout, reset the stream (`*stream.lock().await = None`) inside `send_raw` before returning, independent of the reconnect path.

core/sdk/src/websocket/websocket_client.rs
```diff
@@ -651,129 +651,136 @@ impl WebSocketClient {
...
+
+                let mut response_buffer = vec![0u8; length];
+                stream.read(&mut response_buffer).await?;
+
+                trace!("Received {NAME} response payload, size: {} bytes", length);
+                Ok(Bytes::from(response_buffer))
+            }
+        };
+
+        tokio::time::timeout(self.config.request_timeout.get_duration(), io)
```
hubcio: same desync as the tcp path - the timeout drops `io` while it holds the stream lock, no request id on the wire, `RequestTimeout` not in the reconnect set, so the stream is reused and every later response is shifted by one. worse here because this `io` runs inline with no `tokio::spawn`, so it's also exposed to caller cancellation, not just the internal timeout. needs the same fix: reset the stream on timeout before returning.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -663,111 +662,118 @@ impl TcpClient {
...
-                if !payload.is_empty() {
+            let io = async {
+                let mut stream = stream.lock().await;
+                if let Some(stream) = stream.as_mut() {
+                    #[cfg(feature = "vsr")]
+                    let (request_header, request_size) = {
+                        let mut consensus_session = consensus_session
+                            .lock()
+                            .expect("consensus session mutex poisoned");
+                        crate::vsr::encode_request_header(&mut consensus_session, code, &payload)?
```
hubcio: under the `vsr` feature this advances the shared session request counter before the write happens. if the timeout fires after the id is taken but before the request lands, the counter has moved but the server never saw the request - the next replicated request uses id n+1, the primary sees a gap and silently drops it (and everything after), bricking replicated writes on that session. the counter lives in the shared `ConsensusSession`, not the stream, so quic is not immune to this one even though it dodges the frame desync. same teardown fixes it: reset the vsr session on timeout (the disconnect path's `reset_vsr_session`) inside `send_raw`, since the early returns skip it.

core/sdk/src/quic/quic_client.rs
```diff
@@ -642,78 +642,86 @@ impl QuicClient {
...
+                        .await;
+                }
 
-            error!("Cannot send data. Client is not connected.");
-            Err(IggyError::NotConnected)
+                error!("Cannot send data. Client is not connected.");
+                Err(IggyError::NotConnected)
+            };
+
+            tokio::time::timeout(request_timeout.get_duration(), io)
```
hubcio: `IggyDuration::from_str` maps `"0"`, `"unlimited"`, `"disabled"` and `"none"` to `Duration::ZERO`, and `tokio::time::timeout(ZERO, io)` elapses on the first pending poll - so setting request_timeout to any of those makes every request fail instantly instead of disabling the timeout. same at the tcp and websocket sites. guard with `is_zero()` (already on `IggyDuration`) and skip the timeout wrap when it's zero.

core/common/src/types/configuration/tcp_config/tcp_client_config.rs
```diff
@@ -73,6 +76,7 @@ impl From<ConnectionString<TcpConnectionStringOptions>> for TcpClientConfig {
             tls_validate_certificate: true,
             reconnection: connection_string.options().reconnection().to_owned(),
             heartbeat_interval: connection_string.options().heartbeat_interval(),
+            request_timeout: IggyDuration::from_str("300s").unwrap(),
```
hubcio: this hardcodes 300s instead of reading from the connection string like the field right above it (`heartbeat_interval` uses `connection_string.options()`). same in the quic and websocket `From<ConnectionString>` impls. so request_timeout is settable via the builder and cli but silently pinned to 300s for anyone configuring through a connection string. either parse it from the options or document that it's not supported there.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -663,111 +662,118 @@ impl TcpClient {
         }
 
         let stream = self.stream.clone();
+        let request_timeout = self.config.request_timeout;
         #[cfg(feature = "vsr")]
         let consensus_session = self.consensus_session.clone();
         // SAFETY: we run code holding the `stream` lock in a task so we can't be cancelled while holding the lock.
```
hubcio: this comment is no longer accurate - the new `tokio::time::timeout` below drops the lock-holding `io` from inside the task, which is exactly the cancellation it says can't happen. spawn only shields against caller cancellation. worth correcting when the timeout teardown lands. same at the quic site.

core/sdk/src/client_provider.rs
```diff
@@ -120,6 +120,7 @@ impl ClientProviderConfig {
                     keep_alive_interval: args.quic_keep_alive_interval,
                     max_idle_timeout: args.quic_max_idle_timeout,
                     validate_certificate: args.quic_validate_certificate,
+                    request_timeout: IggyDuration::from_str(&args.request_timeout).unwrap(),
```
hubcio: `from_str(&args.request_timeout).unwrap()` panics on a bad `--request-timeout` value since it's a free-form string with no validation. also at lines 152 and 186. matches the existing pattern for the other duration args here so it's not new, but it adds another panic point on user input - a clap `value_parser` on the arg would reject bad values cleanly instead.

core/common/src/types/configuration/tcp_config/tcp_client_config.rs
```diff
@@ -54,6 +56,7 @@ impl Default for TcpClientConfig {
             tls_ca_file: None,
             tls_validate_certificate: true,
             heartbeat_interval: IggyDuration::from_str("5s").unwrap(),
+            request_timeout: IggyDuration::from_str("300s").unwrap(),
```
hubcio: heads up this is a behavior change - before this, requests had no deadline, now they default to 300s. anything that legitimately takes longer than 300s on a connection now returns `RequestTimeout` (and trips the desync above). worth a changelog / release-note callout so users who relied on unbounded requests aren't surprised.

core/common/src/types/configuration/tcp_config/tcp_client_config.rs
```diff
@@ -41,6 +41,8 @@ pub struct TcpClientConfig {
     pub reconnection: TcpClientReconnectionConfig,
     /// Interval of heartbeats sent by the client
     pub heartbeat_interval: IggyDuration,
+    /// Per-request timeout for send/receive operations.
+    pub request_timeout: IggyDuration,
```
hubcio, line 45: adding a pub field here is a breaking change for anyone constructing this struct with a literal without `..Default::default()`, since it isn't `#[non_exhaustive]`. fine for now while the crate is pre-1.0, but if you plan to keep adding config fields it's worth marking the three client config structs `#[non_exhaustive]` once and steering people to the builders.

hubcio (changes requested):

core/common/src/types/configuration/quic_config/quic_client_config.rs
chengxilo: makes sense, 300 seconds feels the same as waiting infinitely. I checked Kafka's configuration and they use 30 seconds （reference: https://kafka.apache.org/43/configuration/producer-configs/) A little bit more time by default may benefit the client, since it can survive in some brief network problem.

core/sdk/src/client_provider.rs
chengxilo: Done, also verify other duration args now.

core/common/src/types/configuration/tcp_config/tcp_client_config.rs
chengxilo: I would like to keep this for another PR. It's a breaking change which worth a seperate PR imo.

chengxilo:
After merging there is something need to be changed.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -681,152 +674,147 @@ impl TcpClient {
...
-                    );
-                    return TcpClient::handle_response(status, length, stream).await;
+                error!("Cannot send data. Client is not connected.");
+                Err(IggyError::NotConnected)
+            };
+
+            if request_timeout.is_zero() {
+                io.await
+            } else {
+                match tokio::time::timeout(request_timeout.get_duration(), io).await {
```
hubcio: the timeout wraps the whole `io` future including `stream.lock().await`, so the lock-acquire wait counts against the request budget. on master the read deadline was set after lock+write+flush, timing only the response read. since the heartbeat ping shares this same per-transport stream mutex, a request queued behind a stalled in-flight one can hit `RequestTimeout` on lock-wait alone and then tear down the shared connection, cascading to every queued request. acquire the stream lock outside the timed region and time only the io. same in `quic_client.rs` and `websocket_client.rs`.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -681,152 +674,147 @@ impl TcpClient {
...
-            Err(IggyError::NotConnected)
         })
         .await
         .map_err(|e| {
             error!("Task execution failed during TCP request: {}", e);
             IggyError::TcpError
         })?;
 
-        if matches!(result, Err(IggyError::Disconnected)) {
+        if matches!(
```
hubcio, line 831: this teardown nulls the stream and sets `Disconnected` directly, bypassing `disconnect()`, so `publish_event(DiagnosticEvent::Disconnected)` never fires on a timeout - diagnostic subscribers miss every timeout disconnect. routing the timeout teardown through `disconnect()` would emit the event from one shared path. same gap on websocket (`set_state` in the timeout arm) and quic (no `set_state` at all).

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -681,152 +674,147 @@ impl TcpClient {
...
+            };
+
+            if request_timeout.is_zero() {
+                io.await
+            } else {
+                match tokio::time::timeout(request_timeout.get_duration(), io).await {
+                    Ok(result) => result,
+                    Err(_) => {
+                        // Reset to prevent response desync on the shared stream.
+                        *stream.lock().await = None;
```
hubcio: this inner `*stream.lock().await = None` is redundant with the outer `self.stream.lock().await.take()` that already runs for `RequestTimeout(_)`. only the `consensus_session` reset is unique to this block. dropping the inner null and letting the single outer teardown handle stream + state removes the split-lock window and the duplicate work.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -681,152 +674,147 @@ impl TcpClient {
...
+                        #[cfg(feature = "vsr")]
+                        {
+                            *consensus_session
```
hubcio: resetting `consensus_session` mints a fresh client_id. replicated metadata ops are deduped server-side on `(client_id, session, request)`, so a mutation retried after a timed-out-but-committed op runs under the new client_id, misses dedup, and can double-apply. send_messages/partition writes are unaffected (deduped by message id). preserve the client_id across the timeout-reconnect, or suppress transparent retry of replicated mutations after a session reset. vsr-gated so not a blocker for the non-vsr build, but worth fixing before vsr ships.

core/sdk/src/quic/quic_client.rs
```diff
@@ -642,78 +642,99 @@ impl QuicClient {
...
-            error!("Cannot send data. Client is not connected.");
-            Err(IggyError::NotConnected)
+                error!("Cannot send data. Client is not connected.");
+                Err(IggyError::NotConnected)
+            };
+
+            if request_timeout.is_zero() {
+                io.await
+            } else {
+                match tokio::time::timeout(request_timeout.get_duration(), io).await {
```
hubcio: unlike tcp/ws, the quic timeout path resets `consensus_session` (vsr) but never calls `set_state(Disconnected)` or closes the connection - state stays `Connected`. `RequestTimeout` is also absent from the quic retry list, so a quic client never auto-reconnects after a timeout (tcp/ws recover because their next call sees `NotConnected`). under a persistent server stall the consumer just re-opens `open_bi()` and re-times-out every `request_timeout`, surfacing the error each cycle without recovering. fix: `set_state(Disconnected)` here too, or add `RequestTimeout` to the retry list.

core/sdk/src/client_provider.rs
```diff
@@ -120,6 +123,7 @@ impl ClientProviderConfig {
                     keep_alive_interval: args.quic_keep_alive_interval,
                     max_idle_timeout: args.quic_max_idle_timeout,
                     validate_certificate: args.quic_validate_certificate,
+                    request_timeout: parse_duration(&args.request_timeout)?,
                 }));
             }
             TransportProtocol::Http => {
```
hubcio, line 129: the http arm builds `HttpClientConfig` without `request_timeout` and sets no reqwest timeout, so `--request-timeout` is silently dropped when `--transport http` is used, while it applies to the other three transports. wire it into the reqwest client or reject the combination explicitly.

core/common/src/types/args/mod.rs
```diff
@@ -218,6 +218,13 @@ pub struct ArgsOptional {
     #[arg(long)]
     #[serde(skip_serializing_if = "Option::is_none")]
     pub websocket_reconnection_interval: Option<String>,
+
+    /// The optional per-request timeout for send/receive operations
```
hubcio: worth a note here that `request_timeout` values `"0"`, `"unlimited"`, `"disabled"`, `"none"` all map to no timeout (infinite wait), since the `is_zero()` fast-path skips the timeout entirely. `"0"` reading as infinite rather than instant is the opposite of what most people expect.

hubcio (changes requested):
a few findings that don't map onto changed lines:

- the consumer poll loop error classifier (`consumer.rs`) only treats `Disconnected | Unauthenticated | StaleClient` as connection errors, so a `RequestTimeout` (and `NotConnected`) falls through as a raw error with `can_poll` left true and no backoff, on every transport. this is the real root of the quic no-recovery behavior noted in the quic comment. consumer.rs isn't touched by this PR so flagging it here.
- the PR description says default 300s but the code is 30s everywhere (a later commit reduced it). update the description.
- no test for the `is_zero()` no-timeout branch, nor for the vsr session-reset-on-timeout path.
- the timeout is rust-sdk only; java/.net/python/go/c++/node don't get it - worth filing parity issues later on

core/sdk/src/client_provider.rs
chengxilo: Done

core/common/src/types/args/mod.rs
chengxilo: Done

core/sdk/src/tcp/tcp_client.rs
chengxilo: Addressed by reset and reuse same client_id.

chengxilo:
I thought I can make the commit history more clean so I used rebase + force push. But it doesn't end up well so I have to note it here. The changes after the second code review start from [fix(sdk): address code review feedback for per-request timeout](https://github.com/apache/iggy/pull/3429/commits/2e22773426b40800d6f414fe3f39333090086de4).

I added a per-request level timeout in [feat(sdk): provide per-request timeout to override the client level one](https://github.com/apache/iggy/pull/3429/commits/97cd054e6ea208090ae26b32e78c3657e20b1316). I think we should also provide it so the caller can have a better control over their request.

The way we can use the per-request level is 
```rust 
client.ping().with_timeout(IggyDuration::from_str("10ms").unwrap())
```

Issue #3430, opened 2026-06-07
bug(connectors): `FlatBuffer` encoder ignores schemas and silently produces incompatible payloads


Standing-Man:
### Description

The connectors `FlatBuffer` encoding path currently reports successful encoding
but does not produce payloads compatible with the configured FlatBuffers schema.

This affects `JSON`, `Text`, `Raw`, `Proto`, and `Avro` to `FlatBuffer` encoding.

#### 1. The encoder ignores the configured FlatBuffers schema
`FlatBufferEncoderConfig` exposes:

- `schema_path`
- `root_table_name`
- `include_paths`
- `preserve_unknown_fields`

However, none of these fields affect encoding.

`encode_json_to_flatbuffer()` creates an `ad-hoc` root vector/string instead of
encoding the configured root table. Therefore, external consumers generated
from the configured `.fbs` schema cannot read the produced payload.

#### 2. JSON field types and structure are lost during encoding

For JSON objects, every non-string field is serialized into a string before
being placed in the buffer.

This loses FlatBuffers field semantics for:

- integers and floating-point values
- booleans
- nested objects
- arrays
- enums, structs, tables, and unions

For example, `{"active": true, "age": 30}` is encoded using the strings
`"true"` and `"30"` rather than schema-defined boolean and integer fields.

The current buffer also stores key/value offset values inside a generic
`vector<u32>` instead of building the configured FlatBuffers table.

### Expected behavior
- The encoder must load and validate the configured `.fbs` schema.
- `JSON` fields must be encoded according to the configured root table and field
types.
- .....

### Contribution

- [x] I'm willing to submit a pull request to fix this bug

PR #3431, opened 2026-06-07
fix(connectors): prevent silent failures in FlatBuffer conversions

Standing-Man:

Relates to #3430

## Rationale

Prevent `FlatBufferConvert` from silently passing through messages when the
configured conversion is unsupported.

`FlatBufferConvert` previously used a catch-all match branch that returned the
original payload unchanged. For example, configuring `Text -> FlatBuffer` returned success while leaving the
message as `Payload::Text`.

## What changed?

- Validate that the payload variant matches the configured source format.
- Return an explicit configuration error for unsupported conversions.
- Preserve pass-through behavior when source and target formats are identical.
- Apply JSON field mappings only in `FlatBufferStreamEncoder`.
- Add regression tests covering:
  - unsupported conversion errors
  - mismatched payload and source format errors
  - same-format pass-through
  - applying field mappings exactly once

## Behavior changes

- Unsupported conversion pairs, such as `json -> text` and
  `text -> flatbuffer`, now fail during transform configuration loading instead
  of silently passing through the original payload.

- Messages whose payload variant does not match the configured `source_format`
  are now rejected with an error describing the expected payload format. Some
  mismatched payloads that the downstream encoder previously accepted will no
  longer be converted.

- Field mappings are now applied only by the FlatBuffer encoder during
  `json -> flatbuffer` conversion. As a result, same-format conversions such as
  `json -> json` no longer rename fields through `field_mappings`.

## Local Execution

- Passed / not passed
Passed
Ran


core/connectors/sdk/src/transforms/flatbuffer_convert.rs
```diff
@@ -170,9 +148,11 @@ impl Transform for FlatBufferConvert {
                 let encoder = FlatBufferStreamEncoder::default();
                 encoder.convert_format(message.payload, Schema::Raw)?
             }
-            _ => {
-                // For unsupported conversions, pass through unchanged
-                message.payload
+            (source, target) if source == target => message.payload,
+            (source, target) => {
+                return Err(Error::InvalidConfigValue(format!(
```
hubcio: this unsupported-pair error only depends on `source_format` and `target_format`, both static config, but it's checked per message. a misconfigured pair like `text -> flatbuffer` won't fail at startup - the runtime just logs, drops the message, and continues, so every message on the stream is silently dropped forever. that's the same silent-failure class this pr kills, moved from passthrough to drop. make `FlatBufferConvert::new` return `Result` and validate the pair there. `from_config` already returns `Result` and threads `?` for `FilterFields::new` / `UnwrapEnvelope::new`, so the plumbing's there. once validated in `new`, this arm is unreachable and can go.

core/connectors/sdk/src/transforms/flatbuffer_convert.rs
```diff
@@ -125,12 +105,10 @@ impl Transform for FlatBufferConvert {
         _metadata: &TopicMetadata,
         mut message: DecodedMessage,
     ) -> Result<Option<DecodedMessage>, Error> {
-        // Apply field mappings if configured
-        if let Some(field_mappings) = &self.config.field_mappings {
-            message.payload = self.apply_field_mappings(message.payload, field_mappings)?;
+        if !self.payload_matches_source_format(&message.payload) {
```
hubcio: `InvalidPayloadType` is a unit variant - display is just "invalid payload type", no expected-vs-actual. when it fires the operator can't tell if they mislabeled `source_format` or an upstream decoder produced an unexpected variant. the sibling unsupported-conversion error embeds `{source} -> {target}`, so the two failure modes here are asymmetric. cheap fix: `Error::InvalidConfigValue(format!("expected {} payload", self.config.source_format))` since `Schema` already implements `Display`.

core/connectors/sdk/src/transforms/flatbuffer_convert.rs
```diff
@@ -185,3 +165,101 @@ impl Default for FlatBufferConvert {
...
+        let result = converter.transform(&metadata(), message(Payload::Text("value".to_string())));
+
+        assert!(matches!(
+            result,
+            Err(Error::InvalidConfigValue(message))
+                if message == "unsupported FlatBuffer conversion: text -> flatbuffer"
+        ));
+    }
+
+    #[test]
```
hubcio, line 221: `payload_matches_source_format` enumerates all 6 schema/payload pairs but this test only hits `json` source with a `text` payload. the other 5 guards never see a mismatched payload, so a future edit that drops or mis-maps one arm (say avro -> raw) passes every test. exhaustiveness of this guard is the whole point of the fix - worth a table-driven test over each schema variant with a deliberately wrong payload, all asserting `InvalidPayloadType`.

core/connectors/sdk/src/transforms/flatbuffer_convert.rs
```diff
@@ -185,3 +165,101 @@ impl Default for FlatBufferConvert {
...
+            source_format: Schema::Text,
+            target_format: Schema::FlatBuffer,
+            ..FlatBufferConvertConfig::default()
+        });
+
+        let result = converter.transform(&metadata(), message(Payload::Text("value".to_string())));
+
+        assert!(matches!(
+            result,
+            Err(Error::InvalidConfigValue(message))
```
hubcio: asserting the full string `"unsupported FlatBuffer conversion: text -> flatbuffer"` couples the test to the exact prose, the strum lowercasing, and the arrow format - any wording tweak breaks it for no behavioral reason. loosen to `Err(Error::InvalidConfigValue(_))` and, if the message matters, assert it contains the source/target tokens instead of full equality.

core/connectors/sdk/src/transforms/flatbuffer_convert.rs
```diff
@@ -185,3 +165,101 @@ impl Default for FlatBufferConvert {
         Self::new(FlatBufferConvertConfig::default())
     }
 }
+
+#[cfg(test)]
```
hubcio, line 186: the default config is `flatbuffer -> json` (the most common real setup) yet none of the three `FlatBuffer ->` arms have a happy-path test - only `json -> flatbuffer` and the flatbuffer same==same passthrough are covered. a regression in the decoder wiring or the `convert_format` text/raw arms wouldn't be caught. worth one deterministic test each: `-> text` produces base64 `Payload::Text`, `-> raw` returns the bytes unchanged, `-> json` produces the `flatbuffer_size`/`raw_data_base64` shape.

hubcio (changes requested):
core fix is sound - removing the silent `_ => message.payload` catch-all, adding the payload/source-format guard, and dropping the double-application of field_mappings are all the right direction.

two things that don't land on changed lines so noting them here:

`flatbuffer_convert.rs:137-141` - the `FlatBuffer -> Json` arm still does `if let Payload::FlatBuffer(bytes) = message.payload { ... } else { return Err(InvalidPayloadType) }`. the new front guard already proves the payload is flatbuffer whenever `source_format == FlatBuffer`, so that `else` is unreachable. it's also inconsistent with the sibling `-> Text` / `-> Raw` arms that just trust the guard. collapse to `let Payload::FlatBuffer(bytes) = message.payload else { return Err(Error::InvalidPayloadType); };` - one line, still fail-closed.

behavior changes worth calling out in the pr description (all intended, no code change): unsupported pairs the old catch-all silently passed through (`json -> text`, `text -> json`, etc) now hard-error; the new guard rejects mismatched payloads the old encoder would have coped with; and `json -> json` with field_mappings no longer renames keys (the old pre-conversion step did).

core/connectors/sdk/src/transforms/flatbuffer_convert.rs
Standing-Man: Agreed. I’ll make this consistent with the `unsupported-conversion error` by including the expected payload format in the error message.

core/connectors/sdk/src/transforms/flatbuffer_convert.rs
Standing-Man: Good catch. I’ll move the validation of these static configuration pairs into `FlatBufferConvert::new` so that invalid configurations fail fast during initialization rather than being checked on every message at runtime.

hubcio (approved):

numinnex (approved):

spetz (approved):

PR #3432, opened 2026-06-07
test(python): Add tests for python sdk

### LINKED ISSUE #3317: Expand Python SDK test coverage
### Description

Today, the python test suite covers basic happy paths for connectivity, stream/topic creation, polling, TLS, and a few error cases, but it does not exercise the Python binding layer broadly or systematically. There is also no property-based testing, and thus invariants around payload conversion, polling strategy construction, auto-commit configuration, and connection-string parsing are not being stress-tested across a wide input space.

We should strengthen the Python SDK test suite so it can catch binding regressions, input-handling bugs, and edge cases earlier.

### Component

Python SDK

### Proposed solution

1. Add property-based testing support to foreign/python test dependencies.
    - Add hypothesis to the testing extras in foreign/python/pyproject.toml.
2. Add broader tests for Python-exposed API behavior.
    - Cover valid and invalid inputs, boundary values, Unicode payloads, binary payloads, empty payloads, large payloads, and malformed connection strings.
3. Add property-based tests for binding invariants.
4. Expand the existing suite instead of only adding more happy-path tests.
  
This work will be done in multiple PR's instead of a single one since the change is large.

### Alternatives considered

_No response_

slbotbm:

Closes #3317
Relates to #

## Rationale

The python sdk only had a few happy path tests. This PR adds extensive testing for the sdk. Also explicitly maps errors, replacing unwraps.
## What changed?

- added extensive tests, divided into files. Division was done to separate concerns. 
- removed all usage of unwrap in the the rust code.
- moved code of iterator.rs into consumer.rs.  No need for an extra file for a small amount of code. 
## Local Execution

- Passed 
- Pre-commit hooks 

## AI Usage

1. Which tools? codex
2. Scope of usage? code generation, investigation
3. How did you verify the generated code works correctly? read through and reasoned about the code, executed on my machine


foreign/python/tests/test_consumer_group.py
```diff
@@ -0,0 +1,1318 @@
...
+        )
+
+        await iggy_client.send_messages(
+            stream_name,
+            topic_name,
+            partition_id,
+            [Message(message) for message in test_messages],
+        )
+
+        async for message in consumer.iter_messages():
```
hubcio: this loop only ends by counting up to `len(test_messages)`. a live `IggyConsumer` stream never returns `None` - on an empty poll the sdk re-arms and parks, there's no `Ready(None)` path - so if the server ever delivers fewer messages than expected (lost write, rebalance, slow poll) the loop awaits forever. and there's no per-test timeout armed: pytest-timeout is declared but no `timeout=` is set in pyproject and CI passes no `--timeout`, so the only backstop is the 60-min job timeout. that turns a quick assertion failure into a long opaque hang. same pattern at lines 402, 446, 495, 591, 609, 661, 736, 754, 810, 884, 941, 991 and 1018. either wrap each bounded loop in `asyncio.timeout(...)`, or set a suite-wide `timeout=` in `[tool.pytest.ini_options]`.

foreign/python/tests/test_consumer_group.py
```diff
@@ -0,0 +1,1318 @@
...
+            partition_id,
+            PollingStrategy.Next(),
+            1,
+            auto_commit=AutoCommit.Disabled(),
+            poll_interval=timedelta(milliseconds=25),
+            allow_replay=allow_replay,
+        )
+
+        if allow_replay:
+            async for message in replay_consumer.iter_messages():
```
hubcio: the `allow_replay=True` branch reads one message with no timeout, while the `else` branch just below guards with `asyncio.wait_for(..., timeout=0.2)`. the assert expects the rewind to `offsets[-2]` to have taken effect. if that rewind silently failed - which is exactly the regression this test is meant to catch - the `Next` consumer has nothing past `offsets[-1]` and this loop parks forever instead of failing the assert. wrap it in the same `wait_for` guard as the else branch so a broken rewind fails loudly.

foreign/python/tests/test_stream.py
```diff
@@ -0,0 +1,192 @@
...
+            "   ",
+            "a" * 255,
+            ("é" * 126) + "abc",
+            ("한" * 84) + "abc",
+            ("漢" * 84) + "abc",
+            ("あ" * 84) + "abc",
+            ("😀" * 62) + "abcdefg",
+        ],
+    )
+    async def test_create_and_get_stream(
```
hubcio, line 50: this test is parametrized with fixed literal stream names and nothing deletes the streams afterwards (the conftest TODO notes cleanup is blocked on `delete_stream`). iggy persists streams and rejects duplicates - `test_duplicate_stream_creation` below proves it - so a second run against a non-fresh server makes all these cases fail with `StreamNameAlreadyExists`. every other stream-creating test here already uses `unique_name()`. prefix these literals with it too (truncate to <=255 bytes to keep the boundary-length case).

foreign/python/tests/test_stream.py
```diff
@@ -0,0 +1,192 @@
...
+        self, iggy_client: IggyClient, stream_name: str
+    ):
+        """Test stream creation and retrieval."""
+        await iggy_client.create_stream(stream_name)
+
+        stream = await iggy_client.get_stream(stream_name)
+        assert stream is not None
+        assert stream.name == stream_name
+        assert stream.id >= 0
```
hubcio: `stream.id >= 0` is always true - `id()` returns a `u32`, so this can never fail and validates nothing. use `> 0` (iggy ids start at 1) or drop the line. same one at `test_topic.py:66`.

foreign/python/tests/test_topic.py
```diff
@@ -0,0 +1,579 @@
...
+            "   ",
+            "a" * 255,
+            ("é" * 126) + "abc",
+            ("한" * 84) + "abc",
+            ("漢" * 84) + "abc",
+            ("あ" * 84) + "abc",
+            ("😀" * 62) + "abcdefg",
+        ],
+    )
+    async def test_create_and_get_topic(
```
hubcio, line 52: no teardown here - each parametrized case creates a stream and topic and never removes them, so server state grows on every run. unique names keep it correct, it's just unbounded accumulation against a long-lived server. fine to leave until `delete_stream`/`delete_topic` are bound, but worth a cleanup fixture once they are.

foreign/python/tests/test_message_operations.py
```diff
@@ -0,0 +1,778 @@
...
+        polled_messages = await iggy_client.poll_messages(
+            stream=stream_name,
+            topic=topic_name,
+            partition_id=partition_id,
+            polling_strategy=PollingStrategy.First(),
+            count=10,
+            auto_commit=True,
+        )
+
+        assert len(polled_messages) >= len(test_messages)
```
hubcio: `>=` tolerates over-delivery - a fresh unique topic with 3 sent messages should return exactly 3, so a duplicate-publish regression would slip past this. and the `if i < len(polled_messages)` guard below is dead, since you just asserted `len >= 3` and `i` is in `range(3)`. use the full-list equality you already use elsewhere in this file: `assert [m.payload().decode() for m in polled_messages] == test_messages`. same at lines 101-106.

foreign/python/tests/test_consumer_group.py
```diff
@@ -0,0 +1,1318 @@
...
+    @pytest.mark.asyncio
+    async def test_consumer_group_metadata(self, iggy_client: IggyClient, unique_name):
+        """Test that metadata information can be read about the consumer group."""
+        consumer_name = unique_name()
+        stream_name = unique_name()
+        topic_name = unique_name()
+        partition_id = 0
+
+        await iggy_client.create_stream(stream_name)
+        await iggy_client.create_topic(
```
hubcio, line 49: this create_stream + create_topic + consumer_group setup is copy-pasted across ~24 tests in this file, each repeating the same `poll_interval` and `auto_commit` kwargs. worth pulling into a fixture/helper that returns a configured consumer and parametrizing only what varies (polling strategy, batch size, auto_commit, allow_replay).

foreign/python/src/consumer.rs
```diff
@@ -214,6 +217,37 @@ impl IggyConsumer {
...
+                Ok(message
+                    .map(|m| ReceiveMessage {
+                        inner: m.message,
+                        partition_id: m.partition_id,
+                    })
+                    .map_err(|e| {
+                        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(format!("{e:?}"))
+                    })?)
+            } else {
+                Err(PyStopIteration::new_err("No more messages"))
```
hubcio: `__anext__` is an async iterator, so on exhaustion it should raise `StopAsyncIteration`, not `StopIteration` - a `StopIteration` escaping a coroutine becomes a `RuntimeError` under PEP 479, and both the docstring and the `.pyi` stub already promise `StopAsyncIteration`. it's latent today since the sdk consumer stream never yields `None` for a live consumer, so this branch isn't actually reachable - but if that ever changes it surfaces as a confusing `RuntimeError`. swap to `PyStopAsyncIteration::new_err(...)`. (this block was moved verbatim from the old `iterator.rs`, not introduced here.)

foreign/python/src/identifier.rs
```diff
@@ -30,20 +33,33 @@ pub(crate) enum PyIdentifier {
...
-    fn from(py_identifier: PyIdentifier) -> Self {
+impl TryFrom<PyIdentifier> for Identifier {
+    type Error = PyErr;
+
+    fn try_from(py_identifier: PyIdentifier) -> Result<Self, Self::Error> {
         match py_identifier {
-            PyIdentifier::String(s) => Identifier::from_str(&s).unwrap(),
-            PyIdentifier::Int(i) => Identifier::numeric(i).unwrap(),
+            PyIdentifier::String(s) => Identifier::from_str(&s)
+                .map_err(|e| PyErr::new::<PyValueError, _>(format!("{e:?}"))),
```
hubcio: `{e:?}` formats with Debug, so python sees the variant name (`InvalidIdentifier`) instead of the clean Display message (`Invalid identifier`). `{e}` reads better for a user-facing exception. one heads-up: a few of the new tests assert on the Debug variant-name substrings (`StreamNameAlreadyExists`, `TooManyPartitions`), so switching to `{e}` means updating those asserts too. it also matches the existing `{e:?}` convention in client.rs, so it's a wash unless you sweep every site. same pattern at `send_message.rs:67,74` and `client.rs:63`. minor either way.

hubcio (changes requested):
a couple notes that don't land on lines this PR changed, so leaving them here:

`consume_messages` in `src/consumer.rs` still has four `unwrap()` calls (around lines 170, 192, 207, 213) that this PR didn't touch. the `call_method0("wait").unwrap()` plus the join-handle `unwrap()` form a reachable panic path - if the passed shutdown event object's `wait` isn't callable it panics across the FFI boundary instead of raising a clean `TypeError`, since pyo3 doesn't enforce the `asyncio.Event | None` stub annotation. so the description's "removed all usage of unwrap in the rust code" isn't quite right - client.rs, identifier.rs and send_message.rs are clean now, but consumer.rs isn't. either map these too or soften the claim.

foreign/python/tests/test_consumer_group.py
slbotbm: Added per-test timeout of 30s as well as 5s for the tests that check `len(test_messages)`

foreign/python/tests/test_topic.py
slbotbm: TODO already in place

foreign/python/tests/test_stream.py
slbotbm: removed check

foreign/python/tests/test_stream.py
slbotbm: rewrote unique_name so that it can accept minimum and maximum bytes range, and a prefix as well. set reasonable defaults for these.

foreign/python/src/consumer.rs
slbotbm: Done

foreign/python/tests/test_consumer_group.py
slbotbm: Will create a helper function in the future test fixture. Added a TODO

foreign/python/tests/test_consumer_group.py
slbotbm: Used `asyncio.wait_for` instead of `asyncio.timeout` since `timeout` is only supported in/after python 3.11~

foreign/python/src/identifier.rs
slbotbm: Done

slbotbm:
Also removed the remaining unwraps from consumer.rs

foreign/python/tests/test_consumer_group.py
hubcio: what if we dropped 3.10 support?

slbotbm: Then we can use `asyncio.timeout()`. 3.10's EOL is 31th October 2026, so I think we should wait until maybe october, and then start using 3.11

foreign/python/tests/test_message_operations.py
slbotbm: DOne

foreign/python/tests/test_consumer_group.py
slbotbm: Done

spetz (approved):

hubcio (approved):

numinnex (approved):

thepatrykk (approved):

PR #3436, opened 2026-06-08
feat(server-ng): add partition reconciliation loop

krishvishal:
## What

  A level-triggered partition reconciliation loop, one task per shard, that converges each shard's local partition set to the committed Streams metadata.

  - Wakes on a post-commit notifier broadcast by shard 0 for partition-shaping ops, a periodic safety tick (default 1s, capped 30s, configurable), and an initial startup pass.
  - Each pass: build missing owned partitions, seed routing rows for peer-owned namespaces, tear down ghosts removed from the committed target.

  ## How

  **Single-writer via the pump funnel.** The reconciler runs off-pump. It stages `ReconcileOp::{InsertOwned, InsertRouted, ConfirmRemove, RemoveRouted}` and the pump applies them single-writer.

  **Two-phase tombstone teardown.** The reconciler fences writes synchronously (tombstone + `shards_table` row removal), awaits the disk delete, then enqueues `ConfirmRemove`. A failed delete stays tombstoned and backed off; the in-memory partition is never dropped while its data still exists.

  **Epoch / fast-skip model.** A monotonic `Streams::revision` (bumped in the STM on every partition-shaping commit) lets a converged pass skip the O(N) diff when nothing changed. A per-partition `created_revision` distinguishes a delete +
  recreate that reused the same slab-key namespace from the original incarnation, triggering a rebuild instead of serving stale segments.

  **Failure handling.** Per-cause exponential backoff (separate budgets for build vs delete) and recovery for a permanently-wedged tombstone after a failed disk delete.

### DIFF
--- a/core/configs/src/server_config/validators.rs
+++ b/core/configs/src/server_config/validators.rs
@@ -443,8 +443,10 @@ impl Validatable<ConfigurationError> for ShardingConfig {
         let reconcile = self.reconcile_periodic_interval.get_duration();
         if reconcile.is_zero() {
             eprintln!(
-                "Invalid sharding configuration: reconcile_periodic_interval must be > 0 (a zero \
-                 cadence busy-loops the partition reconciler on every shard)"
+                "Invalid sharding configuration: reconcile_periodic_interval resolves to zero. \
+                 Note that \"0\", \"none\", \"unlimited\", and \"disabled\" all parse to zero. The \
+                 periodic reconcile tick is a safety net for dropped commit-wakes and cannot be \
+                 turned off; set a positive duration (default \"1s\", max {RECONCILE_PERIODIC_INTERVAL_MAX:?})."
             );
             return Err(ConfigurationError::InvalidConfigurationValue);
         }
--- a/core/configs/src/server_config/validators.rs
+++ b/core/configs/src/server_config/validators.rs
@@ -22,8 +22,8 @@ use super::server::{
 };
 use super::server::{MemoryPoolConfig, PersonalAccessTokenConfig, ServerConfig};
 use super::sharding::{
-    CpuAllocation, INBOX_CAPACITY_MAX, SHUTDOWN_DRAIN_TIMEOUT_MAX, SHUTDOWN_POLL_INTERVAL_MAX,
-    ShardingConfig,
+    CpuAllocation, INBOX_CAPACITY_MAX, RECONCILE_PERIODIC_INTERVAL_MAX, SHUTDOWN_DRAIN_TIMEOUT_MAX,
+    SHUTDOWN_POLL_INTERVAL_MAX, ShardingConfig,
 };
 use super::system::SegmentConfig;
 use super::system::{CompressionConfig, LoggingConfig, PartitionConfig};
@@ -436,6 +436,23 @@ impl Validatable<ConfigurationError> for ShardingConfig {
                  shutdown_drain_timeout {:?} (a poll cadence coarser than the drain budget makes \
                  the shutdown flag effectively unobservable)",
                 poll, drain
+            );
+            return Err(ConfigurationError::InvalidConfigurationValue);
+        }
+
+        let reconcile = self.reconcile_periodic_interval.get_duration();
+        if reconcile.is_zero() {
+            eprintln!(
+                "Invalid sharding configuration: reconcile_periodic_interval must be > 0 (a zero \
+                 cadence busy-loops the partition reconciler on every shard)"
+            );
+            return Err(ConfigurationError::InvalidConfigurationValue);
+        }
+        if reconcile > RECONCILE_PERIODIC_INTERVAL_MAX {
+            eprintln!(
+                "Invalid sharding configuration: reconcile_periodic_interval {:?} exceeds the \
+                 {:?} cap (a long tick makes post-failure convergence latency operator-visible)",
+                reconcile, RECONCILE_PERIODIC_INTERVAL_MAX
             );
             return Err(ConfigurationError::InvalidConfigurationValue);
         }
--- a/core/message_bus/src/lib.rs
+++ b/core/message_bus/src/lib.rs
@@ -507,6 +507,14 @@ pub trait MessageBus {
     /// Panics on a second install on the same bus, same one-shot
     /// `OnceCell` invariant as [`Self::set_replica_forward_fn`].
     fn set_client_forward_fn(&self, f: ClientForwardFn);
+
+    /// Register a detached `compio::runtime::spawn` handle so
+    /// [`shutdown`](IggyMessageBus::shutdown) can await it before the
+    /// runtime drops. Production [`IggyMessageBus`] pushes onto a
+    /// `RefCell<Vec<JoinHandle>>` drained on shutdown; test stubs can
+    /// leave the no-op default (their tasks are short-lived + don't
+    /// need orderly drain).
+    fn track_background(&self, _handle: JoinHandle<()>) {}
 }
 
 /// Production message bus backed by real TCP connections.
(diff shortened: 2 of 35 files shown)


core/configs/src/server_config/validators.rs
```diff
@@ -440,6 +440,23 @@ impl Validatable<ConfigurationError> for ShardingConfig {
             return Err(ConfigurationError::InvalidConfigurationValue);
         }
 
+        let reconcile = self.reconcile_periodic_interval.get_duration();
+        if reconcile.is_zero() {
+            eprintln!(
+                "Invalid sharding configuration: reconcile_periodic_interval must be > 0 (a zero \
+                 cadence busy-loops the partition reconciler on every shard)"
```
hubcio: rejecting a zero `reconcile_periodic_interval` is right - `run_reconciler` does `compio::time::sleep(periodic)` every loop, so zero busy-loops. problem is `IggyDuration::from_str` maps `"0"`, `"unlimited"`, `"disabled"` and `"none"` all to zero, and the shipped config.toml already teaches those words as off/no-limit (`max_size = "unlimited"`, `message_expiry = "none"`). so an operator writing `reconcile_periodic_interval = "none"` to mean "disable the periodic tick, rely on commit-wakes" gets a hard load-time reject whose message blames a busy-loop. either name the sentinel->zero trap in the message, or treat those sentinels as "use default cadence" instead of rejecting.

maybe "disabled" or "wake_on_commit"?

core/partitions/src/iggy_partition.rs
```diff
@@ -669,6 +669,29 @@ impl<B> IggyPartition<B>
 where
     B: MessageBus,
 {
+    /// Fsync every writer before reconciler-driven teardown drops the
+    /// partition + unlinks its on-disk hierarchy. Errors logged + swallowed.
+    pub async fn drain_and_shutdown(&self) {
```
hubcio: the only caller of this is the ConfirmRemove path in `apply_reconcile_ops`, which runs after `tear_down_owned_partition` already did `delete_partitions_from_disk` -> `remove_dir_all`. so by the time `drain_and_shutdown` fsyncs, the files are already unlinked - it's flushing dirty pages of inodes about to be reclaimed, and the doc here saying fsync runs before the unlink is inverted for that path. the detached spawn over `drained` only exists to host this fsync. no data-loss since the data is being deleted anyway, but `drain_and_shutdown` + the drained Vec + the spawn can all go - just drop the partition inline at ConfirmRemove, the compio file close is non-blocking via io_uring and no fsync is wanted there.

core/server_common/src/sharding/partition_location.rs
```diff
@@ -17,18 +17,33 @@
...
 use super::{LocalIdx, ShardId};
 
-/// Location of a partition: which shard owns it and its local index within that shard.
+/// `local_idx = None` on non-owning rows so peer-shard inserts can't
+/// clobber the owner's real index via unconditional papaya insert.
 #[derive(Debug, Clone, Copy)]
 pub struct PartitionLocation {
     pub shard_id: ShardId,
-    pub local_idx: LocalIdx,
+    pub local_idx: Option<LocalIdx>,
```
hubcio: `local_idx` has no readers - only `owned()`/`routed()` write it, nothing reads it back (`epoch` is the field actually consumed, via `epoch_for`). the `Option<LocalIdx>` is wrapping a field nobody consults. either wire a reader or drop `local_idx`.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -0,0 +1,1437 @@
...
+        .enqueue_reconcile_op(ReconcileOp::ConfirmRemove { namespace: ns });
+    ctx.record_success(ns, FailureCause::Delete);
+    counters.removed_local += 1;
+}
+
+/// Committed `(namespace, created_revision)` pairs. The epoch lets the
+/// additions pass detect a stale local incarnation after slab-key reuse
+/// without an `Arc<TopicStats>` clone per partition; stats are fetched
+/// lazily in [`fetch_topic_stats`] only for namespaces actually built.
+fn snapshot_target_namespaces(ctx: &ReconcilerCtx) -> Vec<(IggyNamespace, u64)> {
```
hubcio, line 570: this allocates a Vec over every committed partition on each non-skipped pass, and `reconcile_removals` builds full namespace sets too. the revision fast-skip covers the steady state, but under sustained partition churn the skip never arms and every pass goes O(committed partitions), re-paid on the periodic tick. fine for now, just flagging the scaling ceiling for large clusters.

core/server-ng/src/bootstrap.rs
```diff
@@ -2340,6 +2160,60 @@ fn socket_addr_from_parts(
...
+                }
+            }
+        }
+    })
+}
+
+/// Filter at broadcast site, keeping unrelated ops off the SDK reply path.
+/// Includes both public and `*WithAssignments` variants for
+/// defense-in-depth. Any new partition-shape op must be added here.
+const fn operation_triggers_partition_reconcile(op: Operation) -> bool {
```
hubcio, line 2240: the bare `CreateTopic` and `CreatePartitions` arms are dead - commits only ever carry the `*WithAssignments` variants (the allocator stamps consensus_group_ids before commit), so `fire_commit_notifier` never sees the plain ones. harmless as defense-in-depth, but they can go if you want it tight.

perhaps mark them with todo to remove after core/server is deleted?

core/message_bus/src/lib.rs
```diff
@@ -507,6 +507,14 @@ pub trait MessageBus {
...
     /// `OnceCell` invariant as [`Self::set_replica_forward_fn`].
     fn set_client_forward_fn(&self, f: ClientForwardFn);
+
+    /// Register a detached `compio::runtime::spawn` handle so
+    /// [`shutdown`](IggyMessageBus::shutdown) can await it before the
+    /// runtime drops. Production [`IggyMessageBus`] pushes onto a
+    /// `RefCell<Vec<JoinHandle>>` drained on shutdown; test stubs can
+    /// leave the no-op default (their tasks are short-lived + don't
+    /// need orderly drain).
+    fn track_background(&self, _handle: JoinHandle<()>) {}
```
hubcio: the default no-op body means any `MessageBus` impl that forgets to override silently drops detached handles, so shutdown won't await them. prod is fine (always the real `IggyMessageBus` through the `Rc<T>` forward), only the test/`()` stub leans on the default. making it a required method removes the leak-by-omission footgun.

core/shard/Cargo.toml
```diff
@@ -22,6 +22,13 @@ edition = "2024"
 license = "Apache-2.0"
 publish = false
 
+[features]
+# Exposes simulator-only test hooks (`IggyShard::init_partition`).
+# Production builds must NOT enable this: it bypasses the reconciler's
```
hubcio: "production builds must NOT enable this" overstates the isolation. under resolver=2 a `cargo build --workspace` unifies features across the graph, and the simulator crate depends on `shard` with this feature as a normal dep, so it gets compiled into the same `shard` unit server-ng links. benign (the gated `init_partition` has no prod callers and the shipped binary is iggy-server), but the comment implies an isolation that workspace builds don't actually give.

hubcio (changes requested):
overall all these comments are just nits. gj krishna, despite the size change is very reasonable.

core/configs/src/server_config/validators.rs
krishvishal: Done.

core/partitions/src/iggy_partition.rs
krishvishal: Done.

core/server_common/src/sharding/partition_location.rs
krishvishal: Done.

core/server-ng/src/partition_reconciler.rs
krishvishal: Done.

core/server-ng/src/bootstrap.rs
krishvishal: Done.

core/message_bus/src/lib.rs
krishvishal: Done.

core/shard/Cargo.toml
krishvishal: Done.

numinnex (approved):

hubcio (approved):

PR #3442, opened 2026-06-08
perf(go): pool the request-path wire payload

### LINKED ISSUE #3441: feat(go-sdk): high per-RPC allocation pressure on TCP poll path
### Description

The Go client's TCP poll path allocates a fresh `[]byte` for every RPC, with multiple sub-allocations re-encoding immutable values. Under sustained load this produces a large allocation/GC tax even when the actual message payloads are small.

#### Evidence

pprof showed consumer running at ~230 events/s with ~400-byte payloads showed:

- **~78.6 GB allocated over 60 s** (~1.3 GB/s) across **~591 M allocations** (~9.85 M/s)
- That works out to **~5.8 MB and ~43,000 allocations per event** against ~400-byte payloads
- Heap *in use*: ~60 MB — i.e. nearly all of the above is churn → heavy GC
- CPU profile is dominated by `internal/runtime/syscall.Syscall6` (33%) and `runtime.futex` (8.5%) — poll RPCs + scheduler/GC overhead, not business logic

#### Dominant allocators (alloc_space / alloc_objects)

From `client/tcp` and `contracts`:

- `tcp.(*IggyTcpClient).PollMessages` → `createPayload` + `command.PollMessages.MarshalBinary` — **~38% of bytes / ~49% of objects**. A fresh request buffer is marshaled on every poll RPC, and `createPayload` allocates a second buffer to prepend the 4-byte length header and 4-byte command code.
- `contracts.Identifier.MarshalBinary` — **~3.4 GB / ~45 M allocs** over 60 s. The same immutable stream/topic/consumer/partition IDs are re-encoded into a fresh slice on every poll.
- `tcp.(*IggyTcpClient).read` — **~5.6 GB / ~68 M objects**. Per-RPC read buffers (both the 8-byte status header and the response body) are freshly allocated.

The data here is small; the cost is *fixed overhead per RPC*. At sustained polling rates this becomes the bottleneck.

### Affected area / component

Go SDK

#### 1. Cache constant `Identifier` wire bytes + zero-alloc encoder path

`Identifier` values are constructed via `NewIdentifier` and never change. Pre-encode the wire form (`Kind | Length | Value`) once at construction and add:

- `Identifier.MarshalledSize() int` — known up-front, lets callers size buffers without trial encodes.
- `Identifier.AppendBinary([]byte) ([]byte, error)` — already present on `Identifier`; make the fast path use the cached bytes.

Mirror this on `command.PollMessages` so the full request body can be encoded into a caller-provided buffer.

`MarshalBinary` keeps its current signature and output for backward compatibility.

#### 2. Pool the request wire-payload buffer in `client/tcp`

Add a `sync.Pool` of request buffers. `IggyTcpClient.do` builds the full wire payload (length header + command code + body) directly into a pooled buffer via the new `AppendBinary` path, eliminating both the `MarshalBinary` allocation and the `createPayload` allocation.

A small `readInto` helper lets the 8-byte response status header read into a stack-local array, removing one more allocation per RPC.

Commands that don't implement the new encoder interface (e.g. `SendMessages`) fall through to `MarshalBinary` with no behaviour change.

### Alternatives considered

_No response_

### Contribution

### Good first issue

matanper:

Closes #3441 

## Rationale

Reduces per-RPC allocation pressure on the Go client's TCP poll path

## What changed?

| Path | Before | After |
| --- | --- | --- |
| All commands: wire-frame assembly (length + code + body) | `createPayload` allocates a fresh `[]byte` per RPC | Reused `sync.Pool`-backed buffer (0 allocs) |
| All commands: 8-byte response status header | `make([]byte, 8)` per RPC | Reused `respHeader [8]byte` field on `IggyTcpClient` |
| `PollMessages`: body encode | `MarshalBinary` allocates the output slice | `AppendBinary` writes into the pooled buffer (0 allocs) |
| `PollMessages`: `Identifier` wire bytes | Re-marshaled per RPC × 4 identifiers (stream, topic, consumer, partition) | Cached at `NewIdentifier`; appended from the cache |
| Other commands: body encode | `MarshalBinary` allocates the output slice | Unchanged — `MarshalBinary` is still called, output is appended into the pooled buffer |
| Other commands: `Identifier` wire bytes | Re-marshaled per RPC via `Identifier.MarshalBinary` | Unchanged — `Identifier.MarshalBinary` still allocates per call; the cache is only consumed via `AppendBinary`, which only `PollMessages` calls in this PR. Migrating other commands' `MarshalBinary` implementations to `AppendBinary` would unlock this for them too. |

## Local Execution

Passed

## AI Usage

1. Claude 
2. entire implementation
3. run it locally on iggy server
4. yes

/ready


foreign/go/client/tcp/tcp_core.go
```diff
@@ -250,16 +284,64 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+	if a, ok := cmd.(appender); ok {
+		bodySize := a.MarshalledSize()
+		total := 8 + bodySize
+		if cap(buf) < total {
+			buf = make([]byte, 8, total)
+		} else {
+			buf = buf[:8]
+		}
+		// total wire message length = body + 4-byte command code
+		binary.LittleEndian.PutUint32(buf[0:4], uint32(bodySize+4))
```
hubcio: the 4-byte length prefix here is written from `MarshalledSize()` (a predicted size) before `AppendBinary` fills the body, and the two are never reconciled. if a command's `MarshalledSize()` ever disagrees with the bytes `AppendBinary` actually writes, the prefix mis-sizes the frame and the single persistent tcp stream desyncs permanently - no req-id, no resync, every later request misframed. the old `createPayload` derived the prefix from the realized body length so it couldn't drift.

today only `PollMessages` takes this branch and it's correct, but the guarantee is now test-enforced instead of structural. cheap fix - realize the body first, then set the prefix from the actual length:

```go
out, err := a.AppendBinary(buf)
if err != nil {
    return nil, err
}
binary.LittleEndian.PutUint32(out[0:4], uint32(len(out)-4))
return out, nil
```

that's `len(out)-4` (code + body), still little-endian. makes `MarshalledSize` a pure capacity hint - if it's wrong you get a realloc, never a corrupt frame.

foreign/go/contracts/identifier.go
```diff
@@ -89,6 +100,14 @@ func (id Identifier) String() (string, error) {
 	return string(id.Value), nil
 }
 
+// MarshalledSize returns the byte length of the AppendBinary encoding.
+func (id Identifier) MarshalledSize() int {
+	if len(id.encoded) > 0 {
+		return len(id.encoded)
+	}
+	return 2 + id.Length
```
hubcio: this fallback returns `2 + id.Length` but `AppendBinary` below writes `2 + len(id.Value)`. they only agree when `Length == len(Value)`. `NewIdentifier` always ties them, but the fields are exported so a hand-built `iggcon.Identifier{Kind, Length, Value}` with a mismatched `Length` is publicly constructible, and `PollMessages` feeds it straight to the encode path with no validation. pre-PR that input was harmless (wire header and body both keyed off `Length`); now the predicted size and the realized append diverge and corrupt the stream.

make this `2 + len(id.Value)` so it can't disagree with `AppendBinary`, and route `MarshalBinary` through `AppendBinary` too (`out := make([]byte, 0, MarshalledSize()); return AppendBinary(out)`) so there's a single source of truth. byte-identical output, same alloc count.

foreign/go/benchmarks/poll_encode_benchmark_test.go
```diff
@@ -0,0 +1,104 @@
...
+func BenchmarkIdentifierMarshalBinary(b *testing.B) {
+	id, _ := iggcon.NewIdentifier("test_stream_id")
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		out, err := id.MarshalBinary()
+		if err != nil {
+			b.Fatal(err)
+		}
+		_ = out
```
hubcio: `_ = out` lets the compiler eliminate the allocation - this benchmark reports 0 allocs even though the comment above calls it the "still-allocating" path. assign to a package-level sink (`var sink []byte` then `sink = out`) so the result escapes and the real cost (1 alloc) shows up. the `PollMessages` pair above doesn't have this problem because its body slice genuinely escapes.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -312,34 +394,28 @@ func (c *IggyTcpClient) sendAndFetchResponse(ctx context.Context, message []byte
...
-	if _, err := c.write(payload); err != nil {
+func (c *IggyTcpClient) sendLocked(wirePayload []byte) ([]byte, error) {
+	if _, err := c.write(wirePayload); err != nil {
 		c.invalidateConnLocked()
 		return nil, err
 	}
 
-	readBytes, buffer, err := c.read(ResponseInitialBytesLength)
-	if err != nil {
+	var header [ResponseInitialBytesLength]byte
```
hubcio: small thing - the commit message says this reads into a stack-local buffer to drop an allocation, but `header[:]` is passed to `c.conn.Read` (an interface call) so escape analysis moves it to the heap anyway (`go build -gcflags=-m` shows `moved to heap: header`). no regression since the old `read(8)` also heap-allocated 8 bytes - just worth fixing the wording so it doesn't read as a win it isn't.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -250,16 +284,64 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
...
+	buf = buf[:8]
+	binary.LittleEndian.PutUint32(buf[0:4], uint32(len(body)+4))
+	binary.LittleEndian.PutUint32(buf[4:8], uint32(cmd.Code()))
+	buf = append(buf, body...)
+	return buf, nil
+}
+
+// sendAndFetchResponse wraps a bare command body with the wire header and
+// sends it. Kept for callers/tests that build payloads from a body + code.
 func (c *IggyTcpClient) sendAndFetchResponse(ctx context.Context, message []byte, command command.Code) ([]byte, error) {
```
hubcio: after this PR `do()` routes everything through `sendWireAndFetchResponse`, so `sendAndFetchResponse` has no production callers left - only the ctx tests use it. could drop it and have those tests call `sendWireAndFetchResponse(ctx, createPayload(...))` directly, keeping `createPayload` as the test oracle. one less method to maintain.

hubcio (changes requested):

foreign/go/client/tcp/tcp_core.go
matanper: `MarshalledSize` is only a hint for the pool buffer, and it seems the `PollMessages` is 58 bytes while the pool returns buffer with initial size of 256.
In practice hint never actually fired, so removing it for simplicity and getting rid of the drift risk

foreign/go/contracts/identifier.go
matanper: removing `MarshalledSize`

foreign/go/client/tcp/tcp_core.go
matanper: I can add response header field on `IggyTcpClient`, this will resolve the `make([]byte, 8)` per RPC

foreign/go/client/tcp/tcp_core.go
matanper: done

foreign/go/benchmarks/poll_encode_benchmark_test.go
matanper: done

hubcio:
alright, from my side this change reasonable and it's merge-able. lets wait for @changhc for couple of days - he's our main golang guy, and nowadays he does most of go sdk work.

thanks for contribution @matanper

foreign/go/client/tcp/tcp_core.go
```diff
@@ -356,6 +414,8 @@ func (c *IggyTcpClient) invalidateConnLocked() {
 	c.state = iggcon.StateDisconnected
 }
 
+// createPayload builds wire-format request bytes; the hot path uses
+// encodeWireRequest with a pooled buffer instead.
 func createPayload(message []byte, command command.Code) []byte {
```
chengxilo: If we want to keep it. We probably should mark it as deprecated. In my opinion, it won't hurt to remove it after code review tho.

foreign/go/contracts/identifier.go
chengxilo, line 113: `Identifier.MarshalBinary` and `AppendBinary` have different internal logic: MarshalBinary reads from Kind/Length/Value fields, while AppendBinary reads from the encoded cache when available. If someone mutates the exported fields after construction, they produce different bytes.

Since `PollMessages.MarshalBinary` already delegates to `AppendBinary,` I'd suggest doing the same here for consistency:

```go
func (id Identifier) MarshalBinary() ([]byte, error) {
    return id.AppendBinary(nil)
}
```

As a follow-up, it might be worth making `Kind/Length/Value` unexported with read-only accessors, i.e. `Kind()/Length()/Value()`. This can make sure we remove the inconsisitency.

Additionally, I'm not sure if it worth fixing, but currently everytime when we create an Identifier, we do the encoding. I'm thinking if we can make it lazy encoding, only encode it when  `MarshalBinary` or `AppendBinary ` is called. And, with current approach, I also don't understand why we still have the logic for encoding in both functions. Isn't it enough to just return `encoded`? For `Identifier{}`, we can just return empty bytes slice.

chengxilo (commented):
To be honest it's not in my expertise. But I tried my best to find some issue that might worth addressing.

foreign/go/client/tcp/tcp_core.go
matanper: Agree, I'll remove it and fix tests

foreign/go/contracts/identifier.go
matanper: I delegated `Identifier.MarshalBinary` to `AppendBinary`

I'd be happy to also unexport `Kind/Length/Value`, after doing that `NewIdentifier` becomes single entry point and `encoded` the source of truth, so we can drop the fallback in `AppendBinary`. The only downside is breaking the public API - let me know if you want it in this PR or in separate one with a changelog / deprecation note

foreign/go/client/tcp/tcp_core_test.go
```diff
@@ -27,7 +27,6 @@ import (
 
 	iggcon "github.com/apache/iggy/foreign/go/contracts"
 	ierror "github.com/apache/iggy/foreign/go/errors"
-	"github.com/apache/iggy/foreign/go/internal/command"
 )
```
chengxilo, line 34:
```suggestion
var emptyWireReq = []byte{4, 0, 0, 0, 0, 0, 0, 0}  
```
nit: use a constant to replace hardcoded literal  `[]byte{4, 0, 0, 0, 0, 0, 0, 0} `

chengxilo (commented):
LGTM. Regarding unexporting Value/Length/Kind, I think the trade-off is worth it. Since it's a public API, allowing users to modify these fields without the binary encoding changing is dangerous.

hubcio:
@matanper once PR is ready for review please always flip the label using
/ready
(put it somewhere in the top level comment body beginning line)
this will change label to `S-waiting-for-review` :) and it's info for maintainers to start checking again.

hubcio (approved):

spetz (approved):

PR #3448, opened 2026-06-10
feat(server-ng): add expired personal access token cleaner

krishvishal:
## Summary

  - **Leader-gated.** The cleaner is spawned on shard 0 of every node but a pass acts only when `is_caught_up_primary()` holds. The delete is a replicated mutation: the primary proposes once, every replica applies the commit. Independent per-node cleanup would race redundant proposals, and only the primary can commit.
  - **Sessionless consensus submit.** `submit_delete_personal_access_token_in_process` is the first server-originated replicated metadata write. With no client session it cannot use the preflight-gated client path, so it mirrors `submit_logout_in_process`: skip `request_preflight`, use the reserved internal `client` id `0`. `get_session(0)` is `None`, so the commit path skips `commit_reply` and its `assert!(client_id != 0)`; the preflight and register asserts are never reached. Delete is idempotent, so dropping dedup is safe.
  - **Expiry-ordered index.** Expired tokens are found by walking an expiry-sorted `BTreeSet` and stopping at the first not-yet-expired entry, so an idle tick is O(log n) instead of a full per-user scan. 
   - **Shutdown.** Purely periodic (`compio::time::timeout`); no metadata-commit wake needed. Stop is observed between tokens; an in-flight submit is not cancel-safe, so the current delete finishes rather than being cut, bounding
    shutdown to one submit.
  - **Rename** `RegisterSubmitError` -> `MetadataSubmitError`: the error now covers register, logout, and PAT-delete submits.


core/server-ng/src/personal_access_token_cleaner.rs
```diff
@@ -0,0 +1,115 @@
...
+        if stop.try_recv().is_ok() {
+            stop_requested = true;
+            break;
+        }
+        // `name` came from a stored PAT, originally a validated `WireName`,
+        // so reconstruction is infallible.
+        let wire_name = WireName::new(name.as_ref())
+            .expect("stored PAT name was validated as a WireName at creation");
+        match metadata
+            .submit_delete_personal_access_token_in_process(user_id, wire_name)
```
hubcio, line 122: the cleaner reads the expired-token snapshot once (line 70), then awaits one replicated delete per token in this loop. the delete is keyed only by `(user_id, name)` - `DeletePersonalAccessTokenRequest::apply` removes by name with no expiry recheck. so if a client deletes and recreates a token with the same name (fresh future expiry) while the cleaner is awaiting an earlier token's commit, the cleaner's later delete for that name purges the freshly recreated, still-valid token. it commits on all replicas and the client is never told. the first/only token in a pass is safe (its op is sequenced before any await yields), but tokens 2..N in a multi-token pass are exposed. the cleaner is enabled by default at a 60s interval in config.toml, so this is reachable on shipped defaults. legacy core/server avoided this by deleting by token_hash (a recreate gets a new hash) and running synchronously within one tick. fix: make the cleaner-originated delete a no-op when `pat.expiry_at > prepare.timestamp` (apply already receives the prepare timestamp), keeping client-initiated delete unconditional; or delete by token_hash like the legacy cleaner.

core/metadata/src/stm/user.rs
```diff
@@ -444,6 +476,13 @@ impl StateHandler for CreatePersonalAccessTokenRequest {
             PersonalAccessToken::raw(self.user_id, self.name.as_ref(), token_hash_str, expiry_at);
         let token_hash = Arc::clone(&pat.token);
         user_tokens.insert(name_arc, pat);
+        if let Some(expiry_at) = expiry_at {
+            state.personal_access_token_expiry_index.insert((
+                expiry_at.as_micros(),
+                self.user_id,
+                Arc::from(self.name.as_str()),
```
hubcio: `Arc::from(self.name.as_str())` here allocates a fresh `Arc<str>` for the same name that's already allocated as `name_arc` (line 457) and allocated again for the token index (line 488), plus once more inside `PersonalAccessToken::raw`. build one `Arc<str>` and `Arc::clone` it for each consumer instead. cold path (pat create) so minor, but easy win.

core/server-ng/src/personal_access_token_cleaner.rs
```diff
@@ -0,0 +1,115 @@
...
+        .users()
+        .read(|users| users.expired_personal_access_tokens(now));
+
+    if expired.is_empty() {
+        return false;
+    }
+
+    let mut removed = 0u32;
+    let mut stop_requested = false;
+    for (user_id, name) in expired {
```
hubcio: `expired_personal_access_tokens` collects every due token into one vec with no per-pass cap, and each one is a separate awaited consensus round in this loop. a large burst (first tick after downtime, or many tokens sharing an expiry) keeps the shard-0 metadata pipeline busy for the whole pass. cap the batch per pass (take first N) and continue on the next tick.

core/metadata/src/impls/metadata.rs
```diff
@@ -1131,12 +1140,116 @@ where
...
+            ..RequestHeader::default()
+        };
+        let prepare = build_prepare_message(
+            consensus,
+            &header,
+            Operation::DeletePersonalAccessToken,
+            &body,
+        );
+
+        consensus.verify_pipeline();
```
hubcio: this dispatch tail (verify_pipeline -> snapshot view/commit_min -> pipeline_message_with_subscriber -> on_replicate -> loopback drain -> receiver.await) is near-identical to the same block in `submit_register_in_process`, `submit_logout_in_process`, and `submit_request_in_process`. worth extracting an `async fn dispatch_and_await(prepare)` and having each caller build its own prepare + map the result. note this fn intentionally drops the `has_message_from_client`/InProgress guard the others have, so keep that difference at the call site.

core/metadata/src/impls/metadata.rs
```diff
@@ -1692,7 +1805,8 @@ where
     // stamps wall-clock once here so every replica's `StateHandler::apply`
     // reads the same `created_at`. A `0` stamp would persist a 1970-01-01
     // `created_at` on every CreateStream/CreateTopic/CreatePartitions, since
-    // submit_command_in_process bypasses `Project::project` and calls this
+    // submit_delete_personal_access_token_in_process bypasses
```
hubcio: this comment now names `submit_delete_personal_access_token_in_process`, but the rationale it gives is about stamping `created_at` for CreateStream/Topic/Partitions, which reach `build_prepare_message` via `prepare_request` - not via the pat-delete fn, whose apply ignores the timestamp. `build_prepare_message` has 3 callers, so either word this generically ("in-process callers that bypass Project::project") or list them.

core/server-ng/src/personal_access_token_cleaner.rs
```diff
@@ -0,0 +1,115 @@
...
+    let expired = metadata
+        .mux_stm
+        .users()
+        .read(|users| users.expired_personal_access_tokens(now));
+
+    if expired.is_empty() {
+        return false;
+    }
+
+    let mut removed = 0u32;
```
hubcio: `removed` is a `u32`; `usize` matches the collection-length idiom. trivial.

core/server-ng/src/personal_access_token_cleaner.rs
```diff
@@ -0,0 +1,115 @@
...
+        // Observe stop between tokens: an in-flight submit is not
+        // cancel-safe, so let the current delete finish rather than cut it.
+        // Bounds shutdown delay to one submit, not the whole batch.
+        if stop.try_recv().is_ok() {
+            stop_requested = true;
+            break;
+        }
+        // `name` came from a stored PAT, originally a validated `WireName`,
+        // so reconstruction is infallible.
+        let wire_name = WireName::new(name.as_ref())
```
hubcio: `WireName::new(...).expect(...)` is infallible for names validated at creation, but a corrupted snapshot carrying an invalid name would panic the cleaner here rather than skipping the entry. minor / out of scope, just flagging.

hubcio (changes requested):

hubcio (approved):

numinnex (approved):

PR #3453, opened 2026-06-11
feat(connectors): add SurrealDB sink connector

countradooku:
## Summary

Adds a SurrealDB sink connector for writing Iggy messages into SurrealDB over the HTTP API via `reqwest`.

The connector supports deterministic record IDs, bulk `INSERT IGNORE` writes for idempotent replay, configurable batch sizing, root/namespace/database/no-auth modes, optional table and offset-index definition, payload modes (`auto`, `json`, `text`, `base64`/`binary`), metadata/header/checksum/origin timestamp fields, retry/backoff handling, and runtime metrics logging.

This PR does not depend on the SurrealDB Rust SDK. The `v3.1.4` reference is only the Docker image tag used by the integration-test fixture, so the earlier SurrealDB SDK BUSL licensing concern is not applicable to this connector.

It also wires the connector into workspace membership, connector docs, binary artifact builds, edge-release output, version bump scripts, and Docker-backed integration test scaffolding.

## Tests

- `cargo fmt --all`
- `cargo test -p iggy_connector_surrealdb_sink`
- `cargo clippy -p iggy_connector_surrealdb_sink --all-targets -- -D warnings`
- `cargo test -p integration --no-run`
- `cargo build -p iggy_connector_surrealdb_sink`
- `cargo test -p integration -- connectors::surrealdb::surrealdb_sink`
- `cargo sort --no-format --workspace --check`
- `./scripts/ci/taplo.sh --check`
- `git diff --check`

`./scripts/ci/markdownlint.sh` could not be run locally because `markdownlint` is not installed in this shell.


ryerraguntla:
Please check the pre-checks failure

countradooku:
Sure

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+                self.id, self.table
+            );
+        } else {
+            debug!(
+                "SurrealDB sink ID: {} wrote {successful_inserts} messages to table '{}'",
+                self.id, self.table
+            );
+        }
+
+        if let Some(error) = last_error {
```
ryerraguntla, line 660: returning Err here propagates out of process_messages, through the runtime's process_messages (core/connectors/runtime/src/sink.rs:426-430, return Err(error)), and out of consume_messages. spawn_consume_tasks (sink.rs:288-297) just logs the error, bumps a metric, and lets the task end, no retry or restart. the consumer is built with .auto_commit(AutoCommit::When(AutoCommitWhen::PollingMessages)) (sink.rs:522),  documented as "the offset is stored on the server when the messages are received" (core/sdk/src/clients/consumer.rs:69-70), i.e. before process_messages runs. so by the time this Err returns, the offset for this batch is already committed, the consumer task is dead, and the batch is gone for good. one batch that exhausts insert_records_with_retry (lib.rs:487-494) permanently drops messages and silently stops the sink.
Fix:  consider not letting a per-batch failure end the task, or have the runtime restart the consumer on error.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+
+    if chars.any(|ch| !(ch == '_' || ch.is_ascii_alphanumeric())) {
+        return Err(Error::InvalidConfigValue(format!(
+            "SurrealDB {field} must contain only ASCII letters, digits, and underscores"
+        )));
+    }
+
+    Ok(())
+}
+
```
ryerraguntla, line 1041: string-matches the lowercased Display of surrealdb::Error against a fixed substring list. surrealdb_types::QueryError::TransactionConflict is documented "the operation can be retried" and surrealdb-core's own
  kvs::Error::is_retryable() returns true for it, but its rendered message is "Transaction conflict: ... This transaction can be retried", which matches none of the substrings above. so a transaction conflict is
  classified as permanent, insert_records_with_retry (lib.rs:461-509) won't retry it, returns CannotStoreData on the first attempt (lib.rs:487-494), and per comment 1 that becomes a dropped batch plus a dead
  consumer task. match on the structured QueryError/surrealdb::Error variants instead of Display text.

core/connectors/sinks/surrealdb_sink/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+iggy_common = { workspace = true }
+iggy_connector_sdk = { workspace = true }
+secrecy = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+surrealdb = { workspace = true }
```
ryerraguntla: pulls in the full surrealdb client, roughly 30 transitive crates (storage engine backends, full rustls/tokio-tungstenite stack, etc.) for what this sink only needs as a websocket query client. worth checking if
  surrealdb's feature matrix can be trimmed to a protocol-only subset, both for build size and for the license surface in comment 10.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+        || message.contains("network")
+        || message.contains("websocket")
+        || message.contains("channel")
+        || message.contains("broken pipe")
+        || message.contains("reset by peer")
+        || message.contains("temporarily unavailable")
+        || message.contains("service unavailable")
+}
+
+fn redact_endpoint(endpoint: &str) -> String {
```
ryerraguntla: this redacts a user:pass@host shape, but connect() (lib.rs:290-303) passes self.config.endpoint straight into Surreal::new::<Ws>/Surreal::new::<Wss>, which expects plain host:port, not a URL with embedded
  credentials. credentials come from the separate username/password config fields via signin_if_configured (lib.rs:305-354), never from endpoint. so the @-split branch only fires for an endpoint that would also
  fail to connect, real credentials never land here. either confirm some supported endpoint format puts credentials here, or drop the branch.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+    topic_metadata: &TopicMetadata,
+    messages_metadata: &MessagesMetadata,
+    message_id: u128,
+) -> String {
+    let mut id = String::with_capacity(
+        topic_metadata.stream.len() * 2 + topic_metadata.topic.len() * 2 + 48,
+    );
+    id.push('s');
+    push_hex_component(&mut id, topic_metadata.stream.as_bytes());
+    id.push_str("_t");
```
ryerraguntla, line 999: push_hex_component(&mut id, topic_metadata.stream.as_bytes());
  ...
  push_hex_component(&mut id, topic_metadata.topic.as_bytes());

  topic_metadata.stream/topic are constant for the whole consumer, but build_record_id recomputes their hex encoding for every record via build_record (lib.rs:511-525). for large batches
  (large_batch_processed_in_chunks pushes 50) that's an identical loop+allocation per message. precompute the hex-encoded stream/topic once per process_messages call (or cache on TopicMetadata) and reuse.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+    push_hex_component(&mut id, topic_metadata.topic.as_bytes());
+    id.push_str("_p");
+    id.push_str(&messages_metadata.partition_id.to_string());
+    id.push_str("_m");
+    id.push_str(&format!("{message_id:032x}"));
```
ryerraguntla: message_id is u128, :032x always produces exactly 32 hex chars (128 bits), so the output is correct, but format! allocates a temporary String just to push into id, when id is already sized with with_capacity for
  this (lib.rs:688-690). use write!(&mut id, "{message_id:032x}") with std::fmt::Write to write directly into the existing buffer and skip the extra allocation.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+        info!(
+            "SurrealDB sink ID: {} processed {} messages with {} errors",
+            self.id, messages_processed, insertion_errors
+        );
+        Ok(())
+    }
+}
+
+impl SurrealDbSink {
+    async fn connect(&self) -> Result<SurrealDbClient, Error> {
```
ryerraguntla: called once from open() (lib.rs:241), the client is stored in self.client and reused for the sink's lifetime. insert_records_with_retry (lib.rs:461-509) retries the query on this same client but never calls
  connect() again. if the websocket connection itself drops, that's exactly one of the cases is_transient_error matches on ("connection"/"network"/"websocket"/"broken pipe"), so every retry reuses the dead client
  and keeps retrying until max_retries, then the batch is dropped per comment 1. consider reconnecting when a connection-class transient error is hit.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1093 @@
...
+        client: &SurrealDbClient,
+        records: Vec<Value>,
+    ) -> BatchInsertOutcome {
+        let mut attempts = 0u32;
+        let query = build_insert_query(&self.table);
+        let record_count = records.len() as u64;
+        let variables = json!({ "records": records });
+
+        loop {
+            let result = client
```
ryerraguntla: variables (the json!({ "records": records }) built at lib.rs:469) gets cloned on every retry iteration, cloning the whole records payload each time, and variables isn't used after the loop. take it by value on
  the final use, or restructure so the clone only happens when another retry is actually about to occur.

core/integration/tests/connectors/surrealdb/surrealdb_sink.rs
```diff
@@ -0,0 +1,352 @@
...
+        assert_eq!(record["iggy_offset"], Value::from(idx as u64));
+        assert_eq!(record["payload"], serde_json::json!({"idx": idx}));
+    }
+}
+
+#[iggy_harness(
+    server(connectors_runtime(config_path = "tests/connectors/surrealdb/sink.toml")),
+    seed = seeds::connector_stream
+)]
+async fn duplicate_record_id_is_idempotent_replay_not_overwrite(
```
ryerraguntla, line 250: lines 250-354 ---   duplicate_record_id_is_idempotent_replay_not_overwrite preseeds a record at the deterministic id for message id 2, sends ids 1-3, and checks the preseeded record stays untouched (INSERT IGNORE,
  build_insert_query lib.rs:599-601) while ids 1 and 3 land. good coverage for the idempotent-replay path. there's no test for the failure path in comments 1/2 though, insert_records_with_retry hitting a
  non-transient error, process_messages returning Err, and what that does to the consumer (currently: silent permanent drop). worth adding once the retry/error classification above is fixed, so this doesn't
  regress unnoticed again.

ryerraguntla:
/author - could you please check the pre-checks failures

core/connectors/sinks/surrealdb_sink/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+iggy_common = { workspace = true }
+iggy_connector_sdk = { workspace = true }
+secrecy = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
```
ryerraguntla, line 44: surrealdb = { workspace = true } pulls surrealdb-core as a transitive dependency. surrealdb-core is licensed under Business Source License 1.1 (BUSL-1.1), an OSI-non-approved license incompatible with Apache 2.0 redistribution. The PR author explicitly flags this: **`./scripts/ci/third-party-licenses.sh --validate --manifest`** core/connectors/sinks/surrealdb_sink/Cargo.toml reports BUSL-1.1 failures. The workspace-level CI passes only because that step runs against core/server/Cargo.toml and core/cli/Cargo.toml, which do not include this crate. The crates.io page for surrealdb v3.1.4 lists the license as "non-standard" and the README confirms surrealdb-core is BUSL-1.1. This cannot be distributed in an Apache Software Foundation project.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1264 @@
...
+        client
+            .query(query)
+            .await
+            .map_err(|e| Error::InitError(format!("Failed to define SurrealDB table: {e}")))?
+            .check()
+            .map_err(|e| Error::InitError(format!("Failed to define SurrealDB table: {e}")))?;
+
+        Ok(())
+    }
+
```
ryerraguntla, line 465: process_messages — every insertion failure is logged and swallowed, function always returns Ok(()):

```
if let Some(error) = outcome.error {
    self.insertion_errors.fetch_add(batch.len() as u64, Ordering::Relaxed);
    error!("Failed to insert SurrealDB batch ...: {error}");
}
// no early return, no Err propagation
Ok(())
```
consume() at L258 propagates this directly. The runtime receives Ok(()) and advances the consumer offset, permanently losing all messages from failed batches with no possibility of redelivery. Other connectors in the repo (Meilisearch, S3, Elasticsearch) return Err from consume() to trigger runtime retry.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1264 @@
...
+    fn new(topic_metadata: &TopicMetadata) -> Self {
+        let mut stream = String::with_capacity(topic_metadata.stream.len() * 2);
+        push_hex_component(&mut stream, topic_metadata.stream.as_bytes());
+        let mut topic = String::with_capacity(topic_metadata.topic.len() * 2);
+        push_hex_component(&mut topic, topic_metadata.topic.as_bytes());
+
+        Self { stream, topic }
+    }
+}
+
```
ryerraguntla, line 988: build_record_id — the deterministic ID uses message_id (u128) not the message offset:

```
id.push_str("_m");
let _ = write!(&mut id, "{message_id:032x}");
```
INSERT IGNORE INTO {table} $records silently deduplicates on matching IDs. If a producer sends messages with id = 0 (the default when IggyMessage::builder().build() is used without setting .id()), every such message in the same stream/topic/partition maps to the same record ID. Only the first insert succeeds; all subsequent messages with id = 0 are silently dropped by SurrealDB with no error returned to the connector. The offset is written as a field but is not part of the record ID. Replace or augment with message.offset, which is unique per partition.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1264 @@
...
+            include_origin_timestamp: true,
+            auto_define_table: false,
+            define_indexes: false,
+            verbose: false,
+            messages_processed: AtomicU64::new(0),
+            insertion_errors: AtomicU64::new(0),
+        }
+        .with_config_defaults()
+    }
+
```
ryerraguntla: with_config_defaults — max_retries clamped to .max(1):

`self.max_retries = self.config.max_retries.unwrap_or(DEFAULT_MAX_RETRIES).max(1);`
Setting max_retries = 0 is silently raised to 1. Document the minimum-1 behavior or reject 0 with an error. Severity:

core/integration/tests/connectors/fixtures/surrealdb/container.rs
```diff
@@ -0,0 +1,187 @@
...
+            .use_db(DEFAULT_DATABASE)
+            .await
+            .map_err(|e| TestBinaryError::FixtureSetup {
+                fixture_type: "SurrealDbContainer".to_string(),
+                message: format!("Failed to select namespace/database: {e}"),
+            })?;
+
+        Ok(client)
+    }
+
```
ryerraguntla, line 300: wait_until_ready — polls by attempting a full create_client() (WebSocket connect + sign in + use_ns/use_db + health check) on every attempt:

```
for _ in 0..SURREALDB_BOOT_ATTEMPTS {
    if let Ok(client) = self.create_client().await
        && client.health().await.is_ok()
    {
        return Ok(());
    }
    sleep(Duration::from_millis(SURREALDB_BOOT_INTERVAL_MS)).await;
}
```
With SURREALDB_BOOT_ATTEMPTS = 120 and SURREALDB_BOOT_INTERVAL_MS = 250, the max wait is 30s. Connection errors during the boot window are swallowed via if let Ok. This is acceptable in test code, but swallowing all errors including non-transient auth failures means a misconfigured test fixture fails with "SurrealDB did not become ready" rather than the actual error.

ryerraguntla:
@countradooku - Please check the Cargo.toml comment on licensing and distribution in the apache redistribution. Please check with any other packages which are distributed with apache license. Unresolved this could be a show stopper. otherwise looks good with more of nits.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+            successful_inserts += outcome.inserted_count;
+
+            if let Some(error) = outcome.error {
+                self.insertion_errors
+                    .fetch_add(batch.len() as u64, Ordering::Relaxed);
+                error!(
+                    "Failed to insert SurrealDB batch for connector ID: {}, table: {}, error: {error}",
+                    self.id, self.table
+                );
+                last_error = Some(error);
```
ryerraguntla: last_error = Some(error) overwrites on every failing batch. Only final batch error returned to caller. Batches 1..N-1 errors silently dropped. Runtime commits offsets on   Ok(()) return — earlier data loss not signaled back. **Fix:** return first error immediately, or collect compound error.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+
+fn push_hex_component(out: &mut String, bytes: &[u8]) {
+    const HEX: &[u8; 16] = b"0123456789abcdef";
+
+    for byte in bytes {
+        out.push(HEX[(byte >> 4) as usize] as char);
+        out.push(HEX[(byte & 0x0f) as usize] as char);
+    }
+}
+
```
ryerraguntla, line 1018: validate_identifier — called for table in open() (always). Called for namespace/database only inside ensure_namespace_database(), which runs only when auto_define_table=true (default: false). In
  default path, arbitrary namespace/database values flow into Surreal-NS/Surreal-DB HTTP headers and /signin JSON body unvalidated. **Fix:** call validate_identifier("namespace", ...) and  validate_identifier("database", ...) unconditionally in open(), same pattern as Doris sink.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+            }
+        }
+    }
+}
+
+impl SurrealDbSink {
+    pub fn new(id: u32, config: SurrealDbSinkConfig) -> Self {
+        let table = config.table.clone();
+        let base_url = build_base_url(&config.endpoint, config.use_tls.unwrap_or(false));
+        let auth_scope = AuthScope::from_config(config.auth_scope.as_deref());
```
ryerraguntla: AuthScope::from_config — unknown value silently defaults to AuthScope::Root (highest privilege) with only a warn!. Operator typo "roo" → root access granted. **Fix**: return 
  Err(Error::InvalidConfigValue) on unknown scope, or default to AuthScope::None.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+        self.process_messages(topic_metadata, &messages_metadata, &messages)
+            .await
+    }
+
+    async fn close(&mut self) -> Result<(), Error> {
+        info!("Closing SurrealDB sink connector with ID: {}", self.id);
+        self.client.get_mut().take();
+
+        let messages_processed = self.messages_processed.load(Ordering::Relaxed);
+        let insertion_errors = self.insertion_errors.load(Ordering::Relaxed);
```
ryerraguntla, line 335: insertion_errors — on build failure in insert_batch, process_messages calls fetch_add(batch.len()). But insert_batch early-returns on first build_record failure — remaining messages never  attempted. Counter inflated by batch.len() - 1 phantom errors. **Fix**: count 1 error for the single build failure; or count only unprocessed messages.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+    }
+
+    async fn get_client(&self) -> Result<SurrealDbClient, Error> {
+        self.client
+            .lock()
+            .await
+            .clone()
+            .ok_or_else(|| Error::InitError("SurrealDB sink is not connected".to_string()))
+    }
+
```
ryerraguntla, line 372: N concurrent consume() tasks hitting connection error simultaneously each call connect_and_select() (network + signin + optional DDL) in parallel. Last writer wins; N-1 connections  orphaned. Under sustained failure: thundering herd. **Fix**: atomic reconnecting flag + try-approach so losing threads skip reconnect and retry with existing client.

ryerraguntla:
@countradooku Except for the process_messages where error reporting could mess up, all others are minor warnings, nits - good to haves and cleanups.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+    is_transaction_conflict(error)
+        || is_connection_error(error)
+        || is_timeout_or_service_error(error)
+}
+
+fn is_transaction_conflict(error: &SurrealDbRequestError) -> bool {
+    let message = error.to_string().to_ascii_lowercase();
+    message.contains("transaction conflict") || message.contains("transaction can be retried")
+}
+
```
ryerraguntla, line 1126: is_connection_error — message.contains("websocket") and message.contains("channel") are dead. Connector uses HTTP-only reqwest::Client; these strings never appear in reqwest errors. **Fix**: remove both
  arms.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+    fn build_payload_document(&self, payload: &Payload) -> Result<PayloadDocument, Error> {
+        match self.payload_format {
+            PayloadFormat::Auto => build_auto_payload_document(payload),
+            PayloadFormat::Json => build_json_payload_document(payload),
+            PayloadFormat::Text => build_text_payload_document(payload),
+            PayloadFormat::Base64 => build_base64_payload_document(payload),
+        }
+    }
+}
+
```
ryerraguntla, line 880: serde_json::to_string(records) allocates full-batch JSON, then format!("INSERT IGNORE INTO {table} {records}...") allocates again. **Fix**: serde_json::to_writer into Vec<u8> with 
  manual prefix/suffix.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+    }
+
+    async fn execute_sql_without_scope(
+        &self,
+        client: &SurrealDbClient,
+        query: &str,
+    ) -> Result<Vec<SurrealSqlStatement>, SurrealDbRequestError> {
+        self.execute_sql_request(client, query, None).await
+    }
+
```
ryerraguntla, line 519: .body(query.to_string()) where query: &str converts &str -> String -> Body. But callers have owned String (from build_insert_query) that was borrowed down to &str. Round-trip wastes one allocation per batch attempt. **Fix**: propagate String or pass Body::from(owned_string) directly.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1487 @@
...
+    }
+}
+
+fn build_record_id(
+    record_id_prefix: &RecordIdPrefix,
+    messages_metadata: &MessagesMetadata,
+    message_id: u128,
+    offset: u64,
+) -> String {
+    let mut id =
```
ryerraguntla, line 995: String::with_capacity(stream.len() + topic.len() + 70). Actual tail: 's'(1) + '_t'(2) + '_p'(2) + u32::MAX(10) + '_o'(2) + u64::MAX(20) + '_m'(2) + u128-hex(32) = 71. One-char  underestimate at max values; one realloc. Fix: +72.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+    }
+
+    let message = error.to_string().to_ascii_lowercase();
+    message.contains("timeout")
+        || message.contains("timed out")
+        || message.contains("temporarily unavailable")
+        || message.contains("service unavailable")
+}
+
+fn redact_endpoint(endpoint: &str) -> String {
```
ryerraguntla: redact_endpoint is identity function Returns endpoint.to_string() verbatim. Name promises credential redaction. Caller at lib.rs:285 logs result at INFO. Endpoint http://user:pass@host:8000 logs credentials in plaintext. **Fix:** replace with  &self.base_url — it's already cleaned by build_base_url at construction time

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+
+    async fn get_client(&self) -> Result<SurrealDbClient, Error> {
+        self.client
+            .lock()
+            .await
+            .clone()
+            .ok_or_else(|| Error::InitError("SurrealDB sink is not connected".to_string()))
+    }
+
+    async fn reconnect(&self) -> Result<(), Error> {
```
ryerraguntla: lib.rs:353-375 + lib.rs:703-712 — concurrent reconnect burns max_retries against dead client  CAS loser in reconnect() returns Ok(()) immediately while winner's reconnect is in flight. Loser loops, calls get_client(), gets old disconnected client, fails again, increments attempts. With default  max_retries=3 and slow reconnect (signin + health + DDL), loser exhausts all 3 attempts and returns permanent Error::CannotStoreData while winner succeeds moments later. Runtime drops the batch. **Fix:** after CAS
  miss, sleep retry_delay and skip the attempt counter increment, OR use a tokio::sync::watch channel instead of AtomicBool.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+                1
+            }
+            Some(max_retries) => max_retries,
+            None => DEFAULT_MAX_RETRIES,
+        };
+        self.include_metadata = self.config.include_metadata.unwrap_or(true);
+        self.include_headers = self.config.include_headers.unwrap_or(true);
+        self.include_checksum = self.config.include_checksum.unwrap_or(true);
+        self.include_origin_timestamp = self.config.include_origin_timestamp.unwrap_or(true);
+        self.auto_define_table = self.config.auto_define_table.unwrap_or(false);
```
ryerraguntla: lib.rs:258-259 + lib.rs:330-336 — define_indexes=true + auto_define_table=false silent no-op  define_indexes flag is only consumed inside ensure_table() (lib.rs:436), which is only called when auto_define_table=true (lib.rs:330). Operator sets define_indexes=true, auto_define_table=false, expects index,  gets none, queries are slow. No warning logged at open(). README:65 says "Requires auto_define_table" but code doesn't enforce or warn. **Fix:** if self.define_indexes && !self.auto_define_table { warn!(...) } in
  open().

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+                    return BatchInsertOutcome {
+                        inserted_count: record_count,
+                        error_count: 0,
+                        error: None,
+                    };
+                }
+                Err(error) => {
+                    attempts += 1;
+                    let transient = is_transient_error(&error);
+                    if !transient || attempts >= self.max_retries {
```
ryerraguntla, line 750: max_retries field semantics: total attempts, not retry count
  attempts starts at 0, increments after each failure, loop exits when attempts >= self.max_retries. max_retries=3 → 3 total attempts (2 retries after first). max_retries=1 → 1 attempt, 0 retries. Warning at
  lib.rs:244-249 says "Using 1 attempt" confirming total-attempts semantics, but the public TOML field name says "retries." Any operator familiar with standard retry configs sets max_retries=1 expecting one retry
  and gets zero. **Fix:** rename to max_attempts in both config struct and connector field, or add explicit doc clarification.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+            return request;
+        };
+        let Some(password) = self.config.password.as_ref() else {
+            return request;
+        };
+
+        request.basic_auth(username, Some(password.expose_secret()))
+    }
+
+    async fn process_messages(
```
ryerraguntla, line 589: messages_processed overcounts on runtime retry On partial batch failure (sub-batch N fails, N-1 succeeded): counter incremented by N-1 at line 597. Runtime retries full consume(). Second run succeeds: counter incremented again by all batches. Final count =  (N-1) + N per restart-and-retry cycle. Metric permanently inflated. Not a correctness issue (INSERT IGNORE makes writes idempotent) but monitoring is broken. **Fix:** reset successful_inserts=0 at the start of each  retry, or document metric as "cumulative optimistic count."

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+        || is_connection_error(error)
+        || is_timeout_or_service_error(error)
+}
+
+fn is_transaction_conflict(error: &SurrealDbRequestError) -> bool {
+    let message = error.to_string().to_ascii_lowercase();
+    message.contains("transaction conflict") || message.contains("transaction can be retried")
+}
+
+fn is_connection_error(error: &SurrealDbRequestError) -> bool {
```
ryerraguntla, line 1127: is_connection_error false positive on Query variant For SurrealDbRequestError::Query(message) and HttpStatus { body }, the function calls error.to_string().to_ascii_lowercase() and checks contains("connection"). Any SurrealDB error mentioning "connection" in text  (e.g., "Connection pool size exceeded", schema errors) triggers reconnect() at line 703-704, wasting one retry slot on a non-network error. **Fix:** gate the string match to SurrealDbRequestError::Request only; for
  other variants rely on HTTP status codes only.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+    }
+}
+
+impl SurrealDbSink {
+    async fn connect_and_select(&self) -> Result<SurrealDbClient, Error> {
+        let client = self.connect().await?;
+        self.signin_if_configured(&client).await?;
+        self.health_check(&client).await?;
+
+        if self.auto_define_table {
```
ryerraguntla, line 350: auto_define_table=true + auth_scope != root fails with opaque InitError
  ensure_namespace_database() issues DEFINE NAMESPACE DDL via execute_sql_without_scope but apply_auth (lib.rs:516) adds whichever scoped credentials were configured. SurrealDB rejects DEFINE NAMESPACE from  non-root credentials → Error::InitError("Failed to define SurrealDB namespace/database: HTTP status 403: ..."). No config-time check. **Fix:** validate at open(): if auto_define_table && auth_scope != 
  AuthScope::Root, return Err(InvalidConfigValue(...)) with clear message.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+        messages_metadata: &MessagesMetadata,
+    ) -> BatchInsertOutcome {
+        if messages.is_empty() {
+            return BatchInsertOutcome {
+                inserted_count: 0,
+                error_count: 0,
+                error: None,
+            };
+        }
+
```
ryerraguntla, line 681: error_count: 1 for record build failure regardless of batch size
  When build_record fails on ANY message in a batch, insert_batch returns immediately with error_count: 1. For a batch of 500, insertion_errors counter gets +1 not +500. Semantically wrong — the entire batch is
  rejected. **Fix:** return error_count: messages.len() as u64.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+    if chars.any(|ch| !(ch == '_' || ch.is_ascii_alphanumeric())) {
+        return Err(Error::InvalidConfigValue(format!(
+            "SurrealDB {field} must contain only ASCII letters, digits, and underscores"
+        )));
+    }
+
+    Ok(())
+}
+
+fn build_base_url(endpoint: &str, use_tls: bool) -> String {
```
ryerraguntla, line 1087: build_base_url ignores use_tls when endpoint has explicit scheme; accepts paths  endpoint="http://host:8000" + use_tls=true → scheme stays http://, TLS silently ignored. endpoint="http://host:8000/extra/path" → all API calls go to /extra/path/sql, /extra/path/health → 404. **Fix:** in open(),
  parse endpoint URL; reject non-empty path components (Error::InvalidConfigValue); warn if use_tls=true conflicts with explicit http:// scheme.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum PayloadFormat {
+    Auto,
+    Json,
+    Text,
+    Base64,
+}
+
+impl PayloadFormat {
+    fn from_config(value: Option<&str>) -> Self {
```
ryerraguntla: PayloadFormat::from_config unknown values silently default to Auto with no open()-time recheck Unknown payload_format string in config → warn + PayloadFormat::Auto at new() time. No re-validation in open(). Inconsistent with AuthScope which re-validates in open() and returns Err. Operator typo in  payload_format silently becomes "auto" behavior. **Fix:** add open()-time validation of payload_format or document the permissive-default behavior clearly.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1580 @@
...
+        if self.auto_define_table {
+            self.ensure_namespace_database(&client).await?;
+            self.ensure_table(&client).await?;
+        }
+
+        Ok(client)
+    }
+
+    async fn connect(&self) -> Result<SurrealDbClient, Error> {
+        HttpClient::builder()
```
ryerraguntla, line 359: connect() declared async fn with no .await
  HttpClient::builder().build() is synchronous. Unnecessary async frame wastes stack and scheduler context. **Fix:** remove async.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1713 @@
...
+            debug!(
+                "SurrealDB sink ID: {} wrote {successful_inserts} messages to table '{}'",
+                self.id, self.table
+            );
+        }
+
+        Ok(())
+    }
+
+    async fn insert_batch(
```
ryerraguntla, line 667: insert_batch — one malformed record (e.g. bad JSON payload) discards the whole already-built batch (up to batch_size=1000 msgs), error_count reported as 1 not batch size. lib.rs:578-622 process_messages then stops at first failing chunk, never attempts later chunks. Combined with the runtime bug above: one bad payload in a 1000-msg batch can silently vaporize hundreds of messages  with a metrics undercount. **Fix:** don't let one bad record nuke the batch; count full chunk as affected, or process remaining chunks. Also please see other sink connectors approach and implement in the same pattern.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1713 @@
...
+
+        if self.include_checksum {
+            record.insert(
+                "iggy_checksum".to_string(),
+                Value::String(message.checksum.to_string()),
+            );
+        }
+
+        if self.include_origin_timestamp {
+            record.insert(
```
ryerraguntla, line 848: lib.rs:786-791 vs lib.rs:779-784 — iggy_origin_timestamp (u64) stored as raw Value::Number, while iggy_checksum (also u64) is deliberately stringified specifically to dodge JSON-f64 precision loss
  above 2^53. iggy_offset/iggy_timestamp (lib.rs:766-771) have the same gap. Not an active bug yet (microsecond timestamps won't cross 2^53 for a long time), but inconsistent with the precision-loss-avoidance
  rationale already applied one field over. **Fix:** stringify all u64 metadata fields consistently.

hubcio:
how much time do the newly added tests take, on your machine?

countradooku:
On my machine, the focused SurrealDB integration test filter:
cargo test -p integration -- connectors::surrealdb::surrealdb_sink
runs 4 tests in 1.94s test time, 11.57s wall-clock including incremental compile.
The SurrealDB sink unit tests:
cargo test -p iggy_connector_surrealdb_sink
run 35 tests in 0.02s test time, 0.61s wall-clock.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+                        error: None,
+                    };
+                }
+                Err(error) => {
+                    let transient = is_transient_error(&error);
+
+                    if transient && is_connection_error(&error) {
+                        match self.reconnect().await {
+                            Ok(true) => {}
+                            Ok(false) => continue,
```
hubcio, line 763: the reconnect race-loser path skips the retry accounting. when two `consume` tasks hit a connection error at the same time, the one that loses the `reconnecting` CAS sleeps `retry_delay`, returns `Ok(false)`, and the caller does `continue` here (the arm under `Ok(true) => {}`). that jumps straight back to the top of the loop, skipping `attempts += 1` and the `attempts >= max_retries` guard below it. so a single batch can re-issue the full failing insert more than `max_retries` times - the cap is silently ignored.

this is reachable: `consume(&self)` is shared (one `SinkContainer` per connector id) and the runtime spawns one task per (stream, topic) consumer, all driving the same instance concurrently, so any config with >=2 stream/topic pairs can hit it. the 4 integration tests are all single-stream/single-topic so they never exercise it.

each doomed spin re-runs `execute_sql`, which is bounded by `query_timeout` (default 30s), so on a hung db the retry rate is roughly one insert per timeout rather than a tight loop - but it still ignores `max_retries` and the consumer's poll loop stalls for the duration. fix: `attempts += 1` before the `continue` so a lost race counts as a transient attempt and falls through to the guard.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+}
+
+fn build_json_payload_document(payload: &Payload) -> Result<PayloadDocument, Error> {
+    match payload {
+        Payload::Json(value) => Ok(PayloadDocument {
+            value: owned_value_to_serde_json(value),
+            encoding: ENCODING_JSON,
+        }),
+        _ => {
+            let bytes = payload.try_to_bytes()?;
```
hubcio: also at 915 and 927.

`build_payload_document` calls `payload.try_to_bytes()` which clones the inner `Vec`/`String` for `Payload::Raw`/`FlatBuffer`/`Avro`. `consume()` already owns `messages: Vec<ConsumedMessage>` but threads them by reference, so the zero-copy `try_into_vec(self)` in the sdk goes unused. the json path (`Payload::Json` under auto/json format) routes through `owned_value_to_serde_json` and avoids this, so the clone only bites raw/flatbuffer/avro and the format-mismatch fallbacks. fixing it needs an ownership refactor of the consume chain.

core/connectors/sinks/surrealdb_sink/README.md
```diff
@@ -0,0 +1,91 @@
+# SurrealDB Sink Connector
+
+Writes Apache Iggy stream messages into SurrealDB over the HTTP API.
+
+The sink writes one SurrealQL bulk `INSERT IGNORE` per connector batch. Each
+record uses a deterministic SurrealDB record id derived from stream, topic,
+partition, offset and Iggy message id, so replayed batches are idempotent and
+existing records are left untouched.
```
hubcio, line 8: this idempotency claim is true for replay, but the readme doesn't mention delivery semantics on failure. the runtime discards the plugin's `consume` return code and commits the offset at poll time, before the write, so a batch that fails permanently (db down past `max_retries`) is committed-but-never-written and never redelivered - silent at-most-once loss. worth a one-line caveat that delivery is at-most-once on persistent sink failure. (the discard + commit-before-write live in the shared runtime, not this pr, so that part is a separate follow-up.)

core/connectors/sinks/surrealdb_sink/README.md
```diff
@@ -0,0 +1,91 @@
...
+| `namespace` | required | SurrealDB namespace selected during `open()`. |
+| `database` | required | SurrealDB database selected during `open()`. |
+| `table` | required | Target table. Must be a safe SurrealQL identifier. |
+| `username` / `password` | none | Optional credentials. |
+| `auth_scope` | `root` | `root`, `namespace`, `database`, or `none`. |
+| `use_tls` | `false` | Uses `https://` when true and `endpoint` has no scheme, `http://` otherwise. |
+| `auto_define_table` | `false` | Runs `DEFINE TABLE IF NOT EXISTS <table> SCHEMALESS`. |
+| `define_indexes` | `false` | Defines an offset index on stream/topic/partition/offset. Requires `auto_define_table`. |
+| `batch_size` | `1000` | Maximum number of records per SurrealDB request. |
+| `payload_format` | `auto` | `auto`, `json`, `text`, or `base64`. |
```
hubcio: the table omits the `binary` alias - it's accepted as a synonym for `base64` (the parser maps `binary -> base64`, and the validation error message lists it too). add it here for parity.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+                    .process_messages(
+                        &test_topic_metadata(),
+                        &test_messages_metadata(),
+                        &[message],
+                    )
+                    .await;
+
+                assert!(
+                    matches!(result, Err(Error::InvalidRecordValue(_))),
+                    "batch failures should be returned to the runtime"
```
hubcio: this comment says batch failures are returned to the runtime, but the runtime discards the FFI return code - it calls the consume callback as a bare statement and returns Ok unconditionally. the Err is only observed here in the plugin's own test, never by the runtime. the comment is misleading.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+    if chars.any(|ch| !(ch == '_' || ch.is_ascii_alphanumeric())) {
+        return Err(Error::InvalidConfigValue(format!(
+            "SurrealDB {field} must contain only ASCII letters, digits, and underscores"
+        )));
+    }
+
+    Ok(())
+}
+
+fn validate_endpoint_config(endpoint: &str, use_tls: bool) -> Result<(), Error> {
```
hubcio, line 1042: `validate_endpoint_config` accepts a `user:pass@host` endpoint (it rejects path/query/fragment but not userinfo), and then `build_base_url` silently strips the credentials at lines 1073-1074 with no warning. either reject userinfo fail-closed or warn that embedded creds are ignored.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+                    "Failed to insert SurrealDB batch of {} messages for connector ID: {}, table: {}, error: {batch_error}",
+                    batch.len(),
+                    self.id,
+                    self.table
+                );
+                last_error = Some(batch_error);
+            }
+        }
+
+        self.messages_processed
```
hubcio, line 645: `messages_processed` adds the full `record_count` on a successful insert, but with `INSERT IGNORE` some of those rows are duplicates the db skipped. so the metric counts ignored duplicates as processed - minor semantic mismatch.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+            return BatchInsertOutcome {
+                inserted_count: 0,
+                error_count: record_error_count,
+                error: last_record_error,
+            };
+        }
+
+        let mut outcome = self.insert_records_with_retry(records).await;
+        outcome.error_count += record_error_count;
+        if outcome.error.is_none() {
```
hubcio: when the db insert returns an error it takes precedence over `last_record_error`, so a db error masks a permanent record-build error in the propagated `Error` value (the `error_count` is still summed, only the error variant is lost). a permanent build failure could then get retried as if it were transient.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+        let batch_size = config
+            .batch_size
+            .unwrap_or(DEFAULT_BATCH_SIZE as u32)
+            .max(1) as usize;
+        let query_timeout = parse_duration(config.query_timeout.as_deref(), DEFAULT_QUERY_TIMEOUT);
+        let retry_delay = parse_duration(config.retry_delay.as_deref(), DEFAULT_RETRY_DELAY);
+        let max_retry_delay =
+            parse_duration(config.max_retry_delay.as_deref(), DEFAULT_MAX_RETRY_DELAY);
+
+        SurrealDbSink {
```
hubcio, line 241: `new()` hardcodes defaults in the struct literal, then `with_config_defaults()` immediately overwrites every one of those fields from config. you could compute the values once and build the struct directly, dropping `with_config_defaults` - around 15 lines.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+const DEFAULT_RETRY_DELAY: &str = "100ms";
+const DEFAULT_MAX_RETRY_DELAY: &str = "5s";
+const ENCODING_BASE64: &str = "base64";
+const ENCODING_JSON: &str = "json";
+const ENCODING_TEXT: &str = "text";
+
+type SurrealDbClient = HttpClient;
+
+#[derive(Debug)]
+pub struct SurrealDbSink {
```
hubcio, line 52: the struct stores the whole `config` plus mirrored scalars (`table`, `batch_size`, `include_*`). keeping both invites drift - pick one (cache the scalars or read from config), not both.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1785 @@
...
+            Some(value) if value.eq_ignore_ascii_case("none") => Ok(AuthScope::None),
+            Some(value) if value.eq_ignore_ascii_case("root") => Ok(AuthScope::Root),
+            Some(value) => Err(Error::InvalidConfigValue(format!(
+                "SurrealDB auth_scope must be one of root, namespace, database, or none: {value}"
+            ))),
+            None => Ok(AuthScope::Root),
+        }
+    }
+
+    fn from_config(value: Option<&str>) -> Self {
```
hubcio: also 156-161.

`AuthScope::from_config` / `PayloadFormat::from_config` are the lenient warn-and-default variants, but `open()` re-parses both strictly (lines 283-284) before anything reads them, so the lenient values are always overwritten. store the raw strings and parse once in `open()`, dropping both functions.

core/integration/tests/connectors/fixtures/surrealdb/sink.rs
```diff
@@ -0,0 +1,288 @@
...
+
+        if let Some(batch_size) = self.batch_size {
+            envs.insert(ENV_SINK_BATCH_SIZE.to_string(), batch_size.to_string());
+        }
+
+        envs
+    }
+}
+
+pub struct SurrealDbSinkJsonFixture {
```
hubcio: the three fixture flavors (json/raw/batch) differ only in schema (json vs raw) and batch_size (none vs some(10)) - payload_format is `auto` in all three. could be one parameterized fixture instead of three Deref structs.

hubcio (changes requested):
a few non-line notes:

- the description says this uses the surrealdb rust sdk 3.1.4 and flags BUSL-1.1 license failures from surrealdb sdk crates, but the code uses raw reqwest and the lockfile adds no surrealdb crate. `3.1.4` is the test-container image tag, not a dep, so the BUSL concern is moot. please fix the description and drop that item.
- the jwt default crypto-provider install (under `core/server/src/http/jwt` + the a2a_jwt tests) and the cpp-bazel cargo registry retries look unrelated to the surrealdb sink. if the jwt change is a rustls feature-unification fix triggered by linking the sink into the all-crates build, please say so; otherwise split these out.
- no adversarial round-trip test for the query builder (quotes/brackets in payloads and identifiers). it's safe by construction - identifiers are validated and bodies go through serde_json escaping - but a test would lock that in.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1845 @@
...
+                inserted_count: 0,
+                error_count: record_error_count,
+                error: last_record_error,
+            };
+        }
+
+        let mut outcome = self.insert_records_with_retry(records).await;
+        outcome.error_count += record_error_count;
+        let db_error = outcome.error.take();
+        outcome.error = last_record_error.or(db_error);
```
ryankert01: `Option::or` keeps the receiver when `Some`, so a per-record build error always beats the DB result. If a payload fails to decode **and** SurrealDB is down, the retryable `db_error` is dropped and the log/returned error show `InvalidRecordValue` instead of the outage. `db_error.or(last_record_error)` would surface a real write failure when one occurred. (The runtime ignores `consume()`'s return, so this is a log/metric issue, not redelivery.)

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1845 @@
...
+        )
+    {
+        return true;
+    }
+
+    let message = error.to_string().to_ascii_lowercase();
+    message.contains("timeout")
+        || message.contains("timed out")
+        || message.contains("temporarily unavailable")
+        || message.contains("service unavailable")
```
ryankert01: This fallback substring-matches the full error `Display`. For `HttpStatus`, `Display` includes the response body, so a permanent `400`/`422` whose body contains "timeout"/"service unavailable" is retried (same for a non-transient `Query`). The typed arms above already catch the real transient cases, and the SDK's `is_transient_status` (retry.rs:190) covers the status set. Drop the string fallback and match on the typed `reqwest::Error`/`StatusCode`.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1845 @@
...
+    async fn ensure_table(&self, client: &SurrealDbClient) -> Result<(), Error> {
+        let table = &self.table;
+        let mut query = format!("DEFINE TABLE IF NOT EXISTS {table} SCHEMALESS;");
+
+        if self.define_indexes {
+            let offset_index = format!("{table}_iggy_offset_idx");
+            validate_identifier("index", &offset_index)?;
+            query.push_str(&format!(
+                " DEFINE INDEX IF NOT EXISTS {offset_index} ON TABLE {table} FIELDS iggy_stream, iggy_topic, iggy_partition_id, iggy_offset;"
+            ));
```
ryankert01, line 456: Index fields `iggy_stream/iggy_topic/iggy_partition_id/iggy_offset` are only written when `include_metadata=true`. With `define_indexes=true, include_metadata=false` the index covers fields no record ever sets. Reject `define_indexes && !include_metadata` in `open()`, or document the dependency.

ryankert01 (commented):
Some comments

core/connectors/sinks/surrealdb_sink/config.toml
```diff
@@ -0,0 +1,58 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+type = "sink"
+key = "surrealdb"
+enabled = true
+version = 0
+name = "SurrealDB sink"
```
ryerraguntla, line 22: surrealdb_sink/config.toml:23 + README.md:22 — plugin path missing ../../ prefix. Every other sink uses ../../target/release/.... Fix: align both.

ryerraguntla:
Apart from the one mentioned, looks good to me.

core/connectors/sinks/surrealdb_sink/src/lib.rs
```diff
@@ -0,0 +1,1875 @@
...
+
+                assert!(matches!(result, Err(Error::InvalidRecordValue(_))));
+            });
+
+        assert_eq!(sink.messages_processed.load(Ordering::Relaxed), 0);
+        assert_eq!(sink.insertion_errors.load(Ordering::Relaxed), 2);
+    }
+
+    #[test]
+    fn given_malformed_record_in_batch_should_still_attempt_valid_records() {
```
ryerraguntla, line 1824: This test given_malformed_record_in_batch_should_still_attempt_valid_records broken. No payload_format = Json, no open(). Auto base64-encodes Raw; failure is InitError("not connected"), not mixed valid/invalid. **Fix**: set PayloadFormat::Json, mock/open client, assert 1 insert + 1 skip.

hubcio (approved):

thepatrykk (approved):

spetz (approved):

PR #3454, opened 2026-06-11
feat(server-ng): reject incompatible clients at login via protocol version

hubcio:
Version mismatches between SDK and server surfaced as runtime decode
failures instead of a clear error at login. The VSR login-register
body now starts with a required ClientVersionInfo prefix: the packed
iggy_binary_protocol crate semver plus sdk name and version. The
server gates on a [min, current-minor] range before touching
credentials (patch never changes the wire) and rejects with a typed
Eviction frame carrying the accepted window; a body without a
decodable prefix is rejected with a dedicated MalformedLogin reason
instead of the version window. The response advertises the server's
protocol and build version; the language-neutral wire spec lives in
the version module docs.

The SDK decodes eviction and reply headers by byte offset instead of
struct casts: response read buffers are not 16-aligned, and offset
reads let the client sanity-check the protocol window. The new
EvictionReason variants and the EvictionHeader protocol window are
appended into reserved space, leaving existing consensus frames and
discriminants byte-compatible.

The SDK validates password and PAT token lengths before encoding,
with the same bounds and errors the servers already enforce, so an
oversized secret cannot desync the u8 length prefix; server-ng
mirrors the legacy username/password bounds before lookup or
hashing. SDK identity is recorded per connection in the
SessionManager; get_clients wire exposure is a follow-up. The
legacy server is untouched.

### DIFF
--- a/core/binary_protocol/src/consensus/header.rs
+++ b/core/binary_protocol/src/consensus/header.rs
@@ -372,6 +372,12 @@ pub enum EvictionReason {
     /// Client missed heartbeats past the configured threshold; the server
     /// evicted the session (and dropped it from any consumer groups).
     StaleClient = 13,
+    /// Client protocol version outside the server's accepted range; the
+    /// header carries `server_protocol_version{,_min}` so the SDK can
+    /// report the exact window.
+    IncompatibleProtocol = 14,
+    /// Login body without a decodable `ClientVersionInfo` prefix.
+    MalformedLogin = 15,
 }
 
 // EvictionHeader - primary -> client (session-terminal, no body)
@@ -393,7 +399,10 @@ pub struct EvictionHeader {
     pub reserved_frame: [u8; 66],
 
     pub client: u128,
-    pub reserved: [u8; 111],
+    /// Accepted protocol window on `IncompatibleProtocol`; zero otherwise.
+    pub server_protocol_version: u32,
+    pub server_protocol_version_min: u32,
+    pub reserved: [u8; 103],
     pub reason: EvictionReason,
 }
 const _: () = {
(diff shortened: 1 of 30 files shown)


core/binary_protocol/src/consensus/header.rs
```diff
@@ -138,10 +137,9 @@ pub struct RequestHeader {
     pub cluster: u128,
     pub size: u32,
     pub view: u32,
-    pub release: u32,
```
krishvishal, line 141: `release: u32` field is removed from all consensus headers. This makes version gate in login only. This creates problems in mixed-version cluster during a rolling upgrade.

The correct behavior would be to revive release field and assert that its non-zero in every frame.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -170,10 +168,9 @@ impl Default for RequestHeader {
             cluster: 0,
             size: 0,
             view: 0,
-            release: 0,
```
krishvishal, line 173: same here.

core/binary_protocol/src/consensus/header.rs
hubcio: good call - dropped the release removal entirely. this commit no longer touches the VSR headers; `release` stays as-is on every frame. the only wire change left is the eviction frame carrying the protocol window, carved additively from its reserved bytes - no renumber, no offset shift. rolling upgrade / per-frame release belongs with the framework work, separate future PR.

krishvishal (dismissed):

mmodzelewski (dismissed):

spetz (approved):

numinnex (approved):

krishvishal (approved):

thepatrykk (approved):

PR #3463, opened 2026-06-12
feat(php): add consumer message iterator

### LINKED ISSUE #3462: Add PHP iterator API for consumer messages
PHP currently exposes consumer message consumption only through `Iggy\Consumer::consumeMessages(callable, int )`, which forces callers into callback-style processing and requires a finite limit.\n\nPython exposes `IggyConsumer.iter_messages()`, which composes naturally with language iteration constructs. The PHP SDK should offer an analogous iterator-style API so users can consume messages with `foreach`, generators, or fibers while preserving the existing callback method.\n\nExpected shape:\n- Add an idiomatic PHP iterator API on `Iggy\Consumer`, for example `iterMessages()`.\n- Yield `Iggy\ReceiveMessage` instances.\n- Preserve existing `consumeMessages()` behavior and compatibility.\n- Document the new API and cover it with PHP SDK tests.

countradooku:
## Summary

Adds `Iggy\Consumer::iterMessages()` for PHP consumers, returning an `Iggy\MessageIterator` that implements PHP's `Iterator` interface and yields `Iggy\ReceiveMessage` instances for `foreach` consumption.

The existing callback-based `consumeMessages(callable, int $limit)` path remains unchanged. The README and PHP stubs document the new API, and the PHP SDK tests now cover ordered iterator consumption.

Closes #3462

## Validation

- `cargo fmt --all`
- `cargo fmt --manifest-path foreign/php/Cargo.toml -- --check`
- `php -l foreign/php/iggy-php.stubs.php && php -l foreign/php/tests/IggySdkTest.php`
- `docker run php:8.3-cli-bookworm cargo check --manifest-path foreign/php/Cargo.toml`
- pre-commit hooks during commit
- pre-push hooks during push, including `cargo clippy`


foreign/php/src/message_iterator.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub(crate) fn new(inner: Arc<Mutex<RustIggyConsumer>>) -> Self {
+        Self {
+            inner,
+            current: None,
+            key: 0,
+            started: false,
+        }
+    }
+
+    fn poll_next(&mut self) -> PhpResult {
```
hubcio, line 49: this async block is identical to `next_message()` over in `consumer.rs` - same `inner.lock().await`, `inner.next().await`, and match arms building `ReceiveMessage`. worth pulling out a small free fn like `poll_one(inner: &Arc<Mutex<RustIggyConsumer>>) -> PhpResult<Option<ReceiveMessage>>` and calling it from both: here `self.current = poll_one(&self.inner)?`, there just `return poll_one(...)`. removes the only real duplication, about 14 lines.

foreign/php/src/message_iterator.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn rewind(&mut self) -> PhpResult {
+        if !self.started {
+            self.started = true;
+            self.poll_next()?;
+        }
+
+        Ok(())
+    }
+
+    pub fn valid(&self) -> bool {
```
hubcio, line 96: worth a one-line note here on why the iterator is open-ended. the underlying sdk consumer stream never returns `None` for a live consumer - on an empty batch it re-arms the poll and parks, so `valid()` stays true and the `None => Ok(None)` arm in `poll_next` is dead during normal operation. net effect is a plain `foreach` over `iterMessages()` blocks forever unless the caller `break`s. readme covers the break, but the reason it never ends isn't obvious from this file.

foreign/php/src/message_iterator.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    pub fn next(&mut self) -> PhpResult {
+        if self.current.is_some() {
+            self.key += 1;
+        }
+
+        self.poll_next()
+    }
+
+    pub fn rewind(&mut self) -> PhpResult {
```
hubcio, line 87: no-op rewind after the first call is the right choice for a forward-only server cursor - you can't seek back, and re-polling here would skip the buffered `current`. only side effect is that re-`foreach`ing the same iterator object resumes instead of restarting. fine as-is, maybe a short doc line that the iterator is single-pass so nobody expects a second loop to replay from the start.

foreign/php/src/consumer.rs
```diff
@@ -128,6 +127,11 @@ impl IggyConsumer {
 
         Ok(consumed)
     }
+
+    /// Returns an iterator over messages for use with foreach.
+    pub fn iter_messages(&self) -> MessageIterator {
```
hubcio, line 132: `iter_messages()` clones the same `Arc<Mutex<RustIggyConsumer>>` the callback path uses, so the iterator isn't an independent cursor. two iterators, or a `foreach` plus a `consumeMessages()` call on the same consumer, split the stream between them - whoever locks first drains the shared buffer and advances the offset. not unsafe, but a php dev will likely expect `iterMessages()` to be its own cursor, so worth a doc note on the shared-consumer semantics.

foreign/php/README.md
```diff
@@ -190,6 +199,8 @@ iggy+tcp://iggy:iggy@127.0.0.1:8090?tls=true&tls_domain=localhost&tls_ca_file=/p
   polled. Pass an explicit partition id before the first poll.
 - `consumeMessages()` requires an explicit finite limit. It does not run forever
   by default.
+- `iterMessages()` returns a PHP `Iterator` and can be used with `foreach`.
```
hubcio, line 202: the `AutoCommit::when()` caveat just below applies to `iterMessages()` too, not only the callback path. with `AutoCommit::interval`/`when` an offset can be stored as the message is polled, before the `foreach` body runs or breaks, so the same at-least-once gap exists. worth folding iterMessages into that note so it's documented for both paths.

hubcio (changes requested):

hubcio (approved):

spetz (approved):

Issue #3460, opened 2026-06-12
Align Go SDK TCP client logs with Rust SDK


slash-init:
### Description

The Go SDK's TCP client logging is not fully aligned with the Rust SDK.
A comparison of the TCP client implementations shows that the Rust SDK logs additional connection lifecycle, reconnection, authentication, request/response, TLS, and leader redirection events that are either missing or inconsistent in the Go SDK.
Aligning these logs would provide a more consistent debugging and observability experience across SDKs while making it easier to troubleshoot similar issues regardless of the language being used.

### Affected area / component

Go SDK

### Proposed solution

Review the Rust TCP client's logging behavior and align the Go SDK where equivalent functionality already exists.

### Contribution

- [x] I'm willing to submit a pull request to implement this feature

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

PR #3464, opened 2026-06-12
feat(go): align TCP client logging with Rust SDK

slash-init:

Relates to #3460 

## Rationale

The Go SDK's TCP client has fewer diagnostic logs than the Rust SDK, making it harder to trace request/response flow, connection lifecycle events, authentication, TLS configuration issues, and leader redirection behavior.

## What changed?

Added missing TCP client logs for requests, responses, connections, authentication, TLS errors, and leader redirection.
The logs were added to match the Rust SDK more closely and make debugging easier.

## Local Execution

* Passed
* Pre-commit hooks ran

## AI Usage

1. Claude Opus 4.6
2. Helped compare logs and review changes.
3. Verified by reviewing the changes and running local checks.
4. Yes.


chengxilo (changes requested):
nit: Just a minor stype recommendation
For example:
```go
    c.logger.Info("Retrying to connect to server...", "retry_count", n+1, "max_retries", attempts, "error", err)
```
can be changed to 
```go
c.logger.Info("Retrying to connect to server...",
  slog.Uint64("retry_count", uint64(n+1)),
  slog.Uint64("max_retries", uint64(attempts)), 
  slog.Any("error", err))
```

For detailed reason, check this reference:
https://www.dash0.com/guides/logging-in-go-with-slog#adding-contextual-attributes-to-your-logs

To be honest I think the previous approach may be more clear. But if we have a lot of KVs, we probably need this strong but ugly approach. 

I don't know why Go doesn't provide function like slog.Uint() for convenience. 🤦‍♂️

slash-init:
> nit: Just a minor stype recommendation For example:
> 
> ```go
>     c.logger.Info("Retrying to connect to server...", "retry_count", n+1, "max_retries", attempts, "error", err)
> ```
> 
> can be changed to
> 
> ```go
> c.logger.Info("Retrying to connect to server...",
>   slog.Uint64("retry_count", uint64(n+1)),
>   slog.Uint64("max_retries", uint64(attempts)), 
>   slog.Any("error", err))
> ```
> 
> For detailed reason, check this reference: https://www.dash0.com/guides/logging-in-go-with-slog#adding-contextual-attributes-to-your-logs
> 
> To be honest I think the previous approach may be more clear. But if we have a lot of KVs, we probably need this strong but ugly approach.
> 
> I don't know why Go doesn't provide function like slog.Uint() for convenience. 🤦‍♂️

Thanks. I went with slog attributes.

Also, one thing I noticed is that Rust's SDK has a connection state guard in the `send_raw()` function before attempting any network operations. The Go SDK doesn't have one in `sendWireAndFetchResponse()`. Unless I'm missing something, it looks like calling an operation before `Connect()` succeeds could eventually reach `c.conn.Write(...)` while `c.conn` is still nil and result in a panic rather than `ErrNotConnected`.
Do we need one here too?

chengxilo:
> > nit: Just a minor stype recommendation For example:
> > ```go
> >     c.logger.Info("Retrying to connect to server...", "retry_count", n+1, "max_retries", attempts, "error", err)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > can be changed to
> > ```go
> > c.logger.Info("Retrying to connect to server...",
> >   slog.Uint64("retry_count", uint64(n+1)),
> >   slog.Uint64("max_retries", uint64(attempts)), 
> >   slog.Any("error", err))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For detailed reason, check this reference: https://www.dash0.com/guides/logging-in-go-with-slog#adding-contextual-attributes-to-your-logs
> > To be honest I think the previous approach may be more clear. But if we have a lot of KVs, we probably need this strong but ugly approach.
> > I don't know why Go doesn't provide function like slog.Uint() for convenience. 🤦‍♂️
> 
> Thanks. I went with slog attributes.
> 
> Also, one thing I noticed is that Rust's SDK has a connection state guard in the `send_raw()` function before attempting any network operations. The Go SDK doesn't have one in `sendWireAndFetchResponse()`. Unless I'm missing something, it looks like calling an operation before `Connect()` succeeds could eventually reach `c.conn.Write(...)` while `c.conn` is still nil and result in a panic rather than `ErrNotConnected`. Do we need one here too?

Makes sense to me, just go ahead and add it : )

slash-init:
Done adding the state guard.
I also added nil guards in `disconnect()` and `shutdown()` since those can be called while a connection attempt is still in progress. In that case, `c.conn` may not have been assigned yet, which could lead to a nil pointer dereference panic.

chengxilo (approved):
lgtm

atharvalade (dismissed):
lgtm only one nit: `foreign/go/client/tcp/tcp_core.go` line 409-410

Wouldn't it read better to log the error before `invalidateConnLocked()`? That way structured log output captures connection state at time-of-failure, not post-teardown. Minor thing though, both orderings work.

slash-init:
@atharvalade Thanks for the feedback! You're right, I've made the change.

atharvalade (approved):
lgtm

ryankert01 (approved):
lg

hubcio (approved):

PR #3471, opened 2026-06-14
feat(csharp): enhance message handling with rented payloads

lukaszzborek:
Rationale

#3250 introduced a rented (span/memory) API on the polling path to cut per-message allocations. This PR does the same for the sending path: let producers serialize and encrypt straight into pooled/rented buffers instead of allocating a byte[] per payload. Previously ISerializer<T>.Serialize returned a fresh byte[] per message (the dominant allocation on typed producers), and there was no way to build a batch whose payloads share one buffer.

The rented send API

- RentedMessageBatch / RentedMessageBatchBuilder — serialize every payload into a single buffer rented from ArrayPool<byte>, send the batch, then dispose. Each Message.Payload is a zero-copy slice into that shared buffer. Memory is valid only until Dispose(); afterwards Messages and any Payload.Span throw ObjectDisposedException (via RentedMemoryManager) rather than reading recycled pool memory.
- IggyPublisher.SendAsync(RentedMessageBatch) (background publisher) takes ownership of the batch and releases its buffer after the flush — caller must not dispose it. For a direct send, await then dispose.
- Single-message overload SendMessagesAsync(Identifier, Identifier, Partitioning, Message, CancellationToken) on publisher and client. The TCP client serializes it from a stack-allocated span and copies the payload into the (pooled) wire buffer synchronously before the returned task completes, so caller-owned memory can be released once it completes.
- SystemTextJsonSerializer<T> — writes JSON straight into the publisher's buffer via a pooled Utf8JsonWriter, no intermediate byte[]. Recommended ISerializer<T>, opted into by setting Serializer (it is required).

Enabling API changes (breaking)

These are the signature changes that make span/memory sending possible:

- Message.Payload is now ReadOnlyMemory<byte> (was byte[]). Constructors still accept byte[] via implicit conversion, but reading message.Payload as a byte[] no longer compiles. New Message.Encrypted flag.
- ISerializer<T>.Serialize changed from byte[] Serialize(T) to void Serialize(T, IBufferWriter<byte>).
- IMessageEncryptor.Encrypt/Decrypt changed from byte[] (byte[]) to byte[] (ReadOnlySpan<byte>), so encryption runs directly over pooled/rented memory.

Background processor

Reworked to carry the rented buffers through async sending without copying:

- Internal SendUnit abstraction: ReadyUnit for already-materialized messages (owning an optional rented batch), TypedUnit<T> for values serialized lazily at flush time.
- Each flush serializes all units into one shared PooledBufferWriter, then a second pass materializes the wire Message list (the buffer can move while growing, so slices are stable only after every unit has serialized). A byte-size guard (BackgroundMaxBatchBytes) bounds the batch; a unit is never split; each unit is disposed once after its flush.
- A throwing serializer/encryptor no longer kills the loop: the offending unit is rolled back out, dropped, and reported through SubscribeOnBackgroundError with its original values; the rest of the batch ships.
- Supporting types: PooledBufferWriter, RentedMemoryManager, PublisherEncryption (in-place, idempotent — a message already marked Encrypted is not re-encrypted on retry). Consumer decrypt path updated to the span-based encryptor.

Tests

SendUnitTests, RentedMessageBatchTests, PooledBufferWriterTests, PublisherEncryptionTests, typed publisher unit + integration tests (IggyTypedPublisherTests), SendMessagesRentedTests.


lukaszzborek:
## Benchmark: 0.8.0 vs Current

Allocation + GC pressure on the send path. Batch methods shown below.

> **Single-message methods omitted.** They allocate too little per op to trigger a measurable Gen0 collection — BenchmarkDotNet reports GC counts per 1,000 ops, and single-send allocations round to <0.5/1,000 (blank). Their allocations still dropped **~49% on average** (e.g. `TypedSendSingle` 3.39 KB → 1.08 KB, −68%).

| Method | 0.8.0 Gen0 | Current Gen0 | 0.8.0 Gen1 | Current Gen1 | 0.8.0 Gen2 | Current Gen2 | 0.8.0 Allocated | Current Allocated | Δ Allocated |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| RawSendBatch | 33.2031 | 5.8594 | 33.2031 | - | 33.2031 | - | 223.51 KB | 115.37 KB | -48.38% |
| PublisherSendBatch | 33.2031 | 5.8594 | 33.2031 | - | 33.2031 | - | 223.66 KB | 115.61 KB | -48.31% |
| TypedSendBatch | 33.2031 | - | 33.2031 | - | 33.2031 | - | 220.76 KB | 26.05 KB | -88.20% |
| RawSendBatchWithHeaders | 35.1563 | 5.8594 | 35.1563 | - | 35.1563 | - | 258.09 KB | 115.37 KB | -55.30% |
| PublisherSendBatchWithHeaders | 35.1563 | 5.8594 | 35.1563 | - | 35.1563 | - | 258.23 KB | 115.63 KB | -55.22% |
| TypedSendBatchWithHeaders | 35.1563 | - | 35.1563 | - | 35.1563 | - | 255.33 KB | 26.05 KB | -89.80% |
| RawRentedSendBatch | 33.2031* | 0.9766 | 33.2031* | - | 33.2031* | - | 223.51 KB* | 25.77 KB | -88.47%* |
| PublisherRentedSendBatch | 33.2031* | 0.9766 | 33.2031* | - | 33.2031* | - | 223.66 KB* | 26.02 KB | -88.37%* |
| BackgroundPublisherSendBatch | 31.2500 | - | 31.2500 | - | 31.2500 | - | 224.92 KB | 116.66 KB | -48.13% |
| BackgroundTypedSendBatch | 31.2500 | - | 31.2500 | - | 31.2500 | - | 221.90 KB | 20.05 KB | -90.96% |
| BackgroundPublisherRentedSendBatch | 31.2500* | - | 31.2500* | - | 31.2500* | - | 224.92 KB* | 26.24 KB | -88.33%* |

`*` = Rented method compared against the 0.8.0 non-rented equivalent (`RawSendBatch` / `PublisherSendBatch` / `BackgroundPublisherSendBatch`).

```
// * Legends *
  Gen0        : GC Generation 0 collects per 1000 operations
  Gen1        : GC Generation 1 collects per 1000 operations
  Gen2        : GC Generation 2 collects per 1000 operations
  Allocated   : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  Δ Allocated : Relative change in allocated memory (Current vs 0.8.0)
```

**Highlights:** 0.8.0 batches triggered full Gen0+Gen1+Gen2 collections on every batch; Current eliminates Gen1/Gen2 churn entirely and cuts Gen0 by ~6× (Rented: ~34×). Typed and Rented paths are now near-allocation-free (−88% to −91%).

hubcio (approved):

numinnex (approved):

spetz (approved):

PR #3487, opened 2026-06-15
fix(integration): fix partitions,permissions integration tests

numinnex:
Fixes integration tests for `partitions` (excluding those that use `consumer_group`) & permissions auth tests.


core/journal/src/file_storage.rs
```diff
@@ -89,19 +89,33 @@ impl FileStorage {
...
     ///
     /// # Errors
     /// Returns an I/O error if the write fails.
-    #[allow(clippy::cast_possible_truncation)]
-    pub async fn write_append<B: IoBuf>(&self, buf: B) -> io::Result<usize> {
-        let len = buf.buf_len();
+    pub async fn write_append<B: IoBuf>(&self, buf: B) -> io::Result<u64> {
+        let len = buf.buf_len() as u64;
+        let offset = self.write_offset.get();
+        self.write_offset.set(offset + len);
```
hubcio, line 119: `write_append` advances `write_offset` here before the write `.await`, but never rolls it back if `write_all_at` fails. on a write error the `?` returns yet the cursor stays advanced. `on_replicate` only logs and returns on append failure (it doesn't crash or poison), so the inflated cursor survives on the long-lived `FileStorage`. the next append - e.g. a vsr retransmit of the op - then writes past a zero hole, fsyncs and acks. on reopen the recovery scan hits the zero bytes at the hole, treats it as a corrupt/non-prepare entry and truncates from there, silently dropping the later op that was already committed and fsync'd. since this is the metadata consensus WAL, that's loss of committed metadata (streams/users/PATs) plus cluster divergence, from a single write fault. the `recovery truncates the gap` note above assumes a reclaimable tail, but the hole sits under a committed op. fix: roll back `write_offset` to `offset` on the error path, or write a skippable filler record so the scan steps over instead of truncating.

core/server-ng/src/responses.rs
```diff
@@ -744,3 +746,121 @@ pub(crate) fn current_metadata_commit(shard: &Rc<ServerNgShard>) -> u64 {
...
+                u32::from_le_bytes(header[36..40].try_into().expect("4-byte slice")) as usize;
+
+            let sections_start = cursor + STORED_MESSAGE_HEADER_SIZE;
+            let sections_end = sections_start + payload_length + user_headers_length;
+            if sections_end > batch_end {
+                return Err(IggyError::InvalidCommand);
+            }
+
+            let offset = batch.base_offset + u64::from(offset_delta);
+            let timestamp = batch.base_timestamp + u64::from(timestamp_delta);
```
hubcio: this mixes two clocks. `base_timestamp` is the broker wall-clock stamped once per batch (`IggyTimestamp::now()` in `iggy_partition` before persist), but `timestamp_delta` is a per-message delta against the producer origin timestamp, not the broker base. so `base_timestamp + timestamp_delta` gives a meaningless per-message `timestamp` for any batch with more than one distinct origin timestamp. the `origin_timestamp` recon on the next line is correct. fix: `let timestamp = batch.base_timestamp;` (drop the delta) - that's the flat broker append time, same as the currently-dead `owned_message` reader. the round-trip test passes only because its messages share an origin timestamp (delta = 0) and it never asserts the timestamp field.

core/shard/src/lib.rs
```diff
@@ -985,6 +1156,54 @@ where
...
+    /// materialised (post-`CreateTopic` convergence window: the metadata
+    /// commit precedes the reconciler pass that builds the local replica).
+    ///
+    /// Returns `Some(message)` when the frame should be processed normally:
+    /// non-partition operation, namespace materialised, or namespace
+    /// tombstoned (the plane's own tombstone guard handles the drop).
+    /// Returns `None` when the frame was parked (or dropped on overflow);
+    /// [`Self::apply_reconcile_ops`] re-dispatches parked frames once the
+    /// matching `ReconcileOp::InsertOwned` lands.
+    fn park_if_unmaterialised<H>(
```
hubcio, line 1213: parked frames are only ever drained on `ReconcileOp::InsertOwned`. if a topic is created (frames parked) then deleted before the reconciler materialises it, the namespace never reaches `InsertOwned`, so `pending_partition_frames[ns]` is never reclaimed - it leaks until process exit. bounded at 128/ns but unbounded across many create-delete-raced namespaces. drop the pending entry on tombstone / remove-routed / confirm-remove too.

core/partitions/src/iggy_partition.rs
```diff
@@ -965,10 +1246,37 @@ where
...
+            // - below the sequencer: a duplicate delivery (parked-frame
+            //   redispatch, retransmit echo) of an op this primary already
+            //   sequenced. Apply is keyed by `header.op` and the primary
+            //   never advances its sequencer post-apply, so proceeding is
+            //   idempotent-safe; log loudly for diagnosis.
+            // - above the sequencer: journaling an op the sequencer has not
+            //   assigned yet means the next local assignment would collide
+            //   with it. Not recoverable in place; crash in release too
+            //   rather than corrupt op assignment silently.
+            assert!(
```
hubcio: this is a release `assert!` in a library on a should-never-happen path. `header.op > current_op` looks unreachable today (the `replicate_preflight` view fences run first, one primary per view, and the chain ring stops before the primary), so prefer `debug_assert!` plus a graceful log-and-return here - apply is keyed by `header.op` and idempotent. separately: the `header.op < current_op` idempotent-apply branch added just below is only reachable when a prior apply left the sequencer ahead of the journal, which is exactly the `write_append` cursor bug - so this branch masks that symptom rather than fixing it. the metadata plane keeps the tighter `debug_assert_eq` for the same invariant; better to fix the journal cursor and keep both planes strict than to relax the partition side.

core/server-ng/src/responses.rs
```diff
@@ -744,3 +746,121 @@ pub(crate) fn current_metadata_commit(shard: &Rc<ServerNgShard>) -> u64 {
...
+///
+/// Body layout: `[partition_id:4][current_offset:8][count:4][messages...]`.
+pub(crate) fn build_polled_messages_body(
+    partition_id: u32,
+    current_offset: u64,
+    fragments: PollFragments,
+) -> Result<Bytes, IggyError> {
+    // Batches may arrive split across fragments (rewritten command header +
+    // sliced blob); concatenate into one stream before walking batches.
+    let mut stream: Vec<u8> = Vec::new();
```
hubcio, line 760: this copies the whole poll payload twice on every poll: once concatenating all fragments into `stream`, then again re-encoding message-by-message into `messages` (plus the final `body` alloc). the fragments are refcounted `Frozen` slices, so the concat throws away the zero-copy. since only the 16-byte header differs from the stored format, you can walk the fragments in place and emit `[rewritten header][payload+user_headers slice]` as chained `Bytes` without the two intermediate buffers.

core/partitions/src/iggy_partition.rs
```diff
@@ -1870,3 +2200,59 @@ where
...
+    count: u32,
+    matched: &mut u32,
+    fragments: &mut PollFragments<4096>,
+    last_matching_offset: &mut Option<u64>,
+) -> usize {
+    let bytes: &[u8] = chunk;
+    let mut cursor = 0usize;
+
+    while *matched < count && cursor + COMMAND_HEADER_SIZE <= bytes.len() {
+        let Ok(batch) = decode_batch_slice(&bytes[cursor..]) else {
```
hubcio, line 2234: disk poll path: `decode_batch_slice` validates the full batch checksum (an xxhash pass over the whole blob) here, then `build_polled_messages_body` walks the same bytes again header-only to re-encode. that's a double parse plus a hash the response builder doesn't need. could have the disk walk emit already-rewritten wire messages, or a header-only decode variant that only checksums when integrity is actually wanted. disk-tier only (polls below the resident range), so not steady-state, but real on backlog replay.

core/partitions/src/iggy_partition.rs
```diff
@@ -674,6 +687,269 @@ where
...
+
+    /// Read one chunk for a disk poll, retrying transient IO failures.
+    async fn read_chunk_with_retry(
+        &self,
+        file: &compio::fs::File,
+        position: u64,
+        len: usize,
+    ) -> Option<Frozen<4096>> {
+        for attempt in 0..3u8 {
+            let buffer = Owned::<4096>::zeroed(len);
```
hubcio, line 899: `Owned::<4096>::zeroed(len)` zeroes up to the full `DISK_POLL_CHUNK` (1 MiB) and then `read_exact_at` immediately overwrites all of it - the memset is wasted since the read fills the whole range before anything reads it. allocate uninitialised, or reuse a single chunk buffer across the segment loop instead of a fresh zeroed one per chunk.

core/partitions/src/iggy_partition.rs
```diff
@@ -674,6 +687,269 @@ where
...
+                );
+                if consumed == 0 {
+                    if (len as u64) >= persisted - position {
+                        // The whole remainder fit and still no complete
+                        // batch decoded: corrupt tail; stop.
+                        break;
+                    }
+                    // A single batch larger than the chunk: grow and
+                    // re-read instead of spinning.
+                    chunk_len = chunk_len.saturating_mul(4);
```
hubcio, line 851: when a batch is bigger than the chunk this regrows by 4x and re-reads from `position`, discarding the bytes just read. the batch header already carries `total_size`, so size the next read to the actual batch length once the header decodes, instead of a blind 4x grow-and-reread.

core/partitions/src/iggy_partition.rs
```diff
@@ -674,6 +687,269 @@ where
...
+            Some((fragments, last_matching_offset, matched))
+        }
+    }
+
+    /// Open a segment file for a disk poll, retrying transient IO failures
+    /// (fd pressure under heavy parallel load) so one failed syscall does
+    /// not silently collapse the poll into an empty result.
+    async fn open_segment_with_retry(&self, path: &str) -> Option<compio::fs::File> {
+        for attempt in 0..3u8 {
+            match compio::fs::File::open(path).await {
```
hubcio, line 872: disk poll opens a fresh file per segment per poll (no fd cache) and re-resolves plus clones `partition_dir()` once per segment. a from-offset-0 replay over many small segments is then one open and one path alloc per segment, every poll. sealed segments are immutable so their fds could be cached, and `partition_dir()` can be hoisted out of the segment loop.

core/server-ng/src/dispatch.rs
```diff
@@ -606,6 +763,293 @@ async fn send_non_replicated_bytes(
...
+///
+/// Covers the post-`CreateTopic` convergence window where the metadata
+/// commit has returned to the client but the per-shard reconcilers have
+/// not yet seeded routing rows / materialised partitions.
+#[allow(clippy::future_not_send)]
+/// Ack a partition op that cannot be routed (unresolved or never-
+/// materialised namespace) with an empty Reply. The SDK connection
+/// processes replies in lockstep, so a silent drop wedges every
+/// subsequent request on that connection.
+async fn send_empty_partition_reply(
```
hubcio, line 901: the `Wait (bounded)...` doc block and the `#[allow(clippy::future_not_send)]` here actually describe `wait_for_partition_routable` below, not `send_empty_partition_reply`. the doc is attached to the wrong fn and the allow is off-target - harmless today since the lint isn't denied, but it won't suppress anything on the fn it was meant for. move both down onto `wait_for_partition_routable`.

core/server_common/src/send_messages2.rs
```diff
@@ -560,6 +564,12 @@ pub fn convert_request_message(
     SendMessages2Owned::from_legacy_request(namespace, body)?.encode_request(request_header)
 }
 
+/// Decode one stored batch slice (`[256B command header][blob]`) -- the
+/// persisted segment-file record format. Validates the batch checksum.
+pub fn decode_batch_slice(body: &[u8]) -> Result<SendMessages2Ref<'_>, IggyError> {
```
hubcio, line 550: `decode_batch_slice` is a verbatim one-line alias of the private `decode_request_slice` with the same signature. either inline it at the single caller or just make `decode_request_slice` pub - the persisted segment record is the same layout as the request slice, so the extra name doesn't earn its keep.

core/partitions/src/journal.rs
```diff
@@ -474,7 +481,7 @@ impl QueryableJournal<PartitionJournalMemStorage> for PartitionJournal<Partition
     }
 }
 
-fn select_batch_slice(
+pub fn select_batch_slice(
```
hubcio, line 484: `select_batch_slice`, `SelectedBatchSlice` (line 53) and `oldest_resident_offset` (264) - plus `segment_indexes` in `log.rs:213` - are all `pub` but every caller is inside the `partitions` crate, so `pub(crate)` fits. `decode_batch_slice` is the one that genuinely needs `pub` (cross-crate). `count` (line 45) is borderline since it's on a pub type.

core/server-ng/src/bootstrap.rs
```diff
@@ -864,6 +865,27 @@ async fn shard_main(
...
+    // client request hangs until the SDK read timeout.
+    let (consensus_tick_stop_tx, consensus_tick_stop_rx) = channel::<()>(1);
+    let tick_shard = Rc::clone(&shard);
+    let consensus_tick_handle = compio::runtime::spawn(async move {
+        const CONSENSUS_TICK_INTERVAL: std::time::Duration = std::time::Duration::from_millis(10);
+        loop {
+            if consensus_tick_stop_rx.try_recv().is_ok() {
+                break;
+            }
+            tick_shard.tick_metadata().await;
```
hubcio, line 896: this tick wakes every 10ms per shard forever and `tick_partitions` builds a fresh `namespaces()` Vec every tick even when there's no armed timeout to service - idle cost scales with partition count. fine for now, but a follow-up could rearm event-driven / skip when nothing is pending (matches the tigerbeetle timer-lifecycle TODO in consensus).

core/server-ng/src/dispatch.rs
```diff
@@ -606,6 +763,293 @@ async fn send_non_replicated_bytes(
...
+
+    if shard.shards_table().shard_for(namespace).is_some() {
+        return true;
+    }
+    let deadline = std::time::Instant::now() + BUDGET;
+    while shard.shards_table().shard_for(namespace).is_none() {
+        if std::time::Instant::now() >= deadline {
+            return false;
+        }
+        compio::time::sleep(ATTEMPT_DELAY).await;
```
hubcio, line 942: the routable probe fires a full cross-shard read roundtrip every 50ms during the post-create-topic convergence window. only hits that window so it's minor, but a single-flight probe per namespace or a routability signal from the reconciler would avoid the spam when many clients write to a just-created topic at once.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -719,7 +755,18 @@ impl TcpClient {
                     let body_size = response_size - iggy_binary_protocol::HEADER_SIZE;
                     let body = if body_size > 0 {
                         let mut body = BytesMut::with_capacity(body_size);
-                        stream.read_buf(&mut body, body_size).await.map_err(|error| {
+                        let body_read = tokio::time::timeout(
+                            RESPONSE_READ_TIMEOUT,
```
hubcio: the header read and body read each get their own full `RESPONSE_READ_TIMEOUT`, so a reply that delivers a header then stalls can wait up to 2x the timeout total instead of 30s. use a single shared deadline (`Instant::now() + RESPONSE_READ_TIMEOUT`) across both reads.

core/server-ng/src/dispatch.rs
```diff
@@ -606,6 +763,293 @@ async fn send_non_replicated_bytes(
...
+    }
+    // The local row is seeded by THIS shard's reconciler; the owner
+    // materialises the partition on its own pass. Probe with a cheap read
+    // until the owner answers, so the write below cannot be dropped by the
```
hubcio: the comment says the write `cannot be dropped by the owner's guard`, but the partition can de-materialise between this probe returning `Some` and the dispatch below - it's just benign because the downstream park/tombstone path re-checks and the client retries. worth softening the comment so it doesn't read as a hard guarantee.

hubcio (dismissed):
a few findings land on code this PR didn't change, so they're here rather than as line comments:

- `core/sdk/src/websocket/websocket_client.rs` (around 700-716): the WS vsr `send_raw` reads the response header and body with no timeout while holding the `stream` lock, which is the exact lockstep wedge the TCP client just got fixed (`RESPONSE_READ_TIMEOUT` plus drop-on-timeout). if the server loses a reply the WS client hangs forever. the new `sdk/messages.rs` test runs over WebSocket too, so this path is now exercised. it's actually worse than the old TCP case - the WS read runs in the caller's task, not a spawned one, so there's no post-task stream drop. mirror the TCP timeout plus stream-drop on WS (and check QUIC).

- `core/server-ng/src/responses.rs:294` and `:597`: the `POLL_MESSAGES` arm in `build_non_replicated_response` and the `NonReplicatedResponse::EmptyPolledMessages` variant look dead now - `handle_poll_messages` has its own explicit dispatch arm, so the fallback never reaches POLL. ~15 lines plus a never-served empty-poll path can go.

- `core/server_common/src/send_messages2.rs:465`: `owned_message` is `#[allow(dead_code)]` with zero callers, and it happens to be the one place with the correct flat-`base_timestamp` reconstruction. either wire it into `build_polled_messages_body` (which would also fix the timestamp bug) or delete it.

hubcio (approved):

krishvishal (approved):

PR #3497, opened 2026-06-15
feat(connectors): add Meilisearch sink connector

### LINKED ISSUE #3495: Add Meilisearch sink connector
## Summary
Add a Meilisearch sink connector that consumes Iggy messages and indexes them into a Meilisearch index through the official Rust SDK.

## Motivation
Users should be able to route Iggy stream data into Meilisearch for full-text search and retrieval without writing a custom connector.

## Scope
- Add the sink connector crate and sample runtime config.
- Document sink configuration and payload behavior.
- Add Docker-backed integration coverage for indexing JSON messages.

countradooku:
## Summary
- add the Meilisearch sink connector using the official `meilisearch-sdk`
- document the sink configuration and payload mapping behavior
- add Docker-backed integration coverage for indexing JSON messages into Meilisearch

## Motivation
This splits the sink portion out of #3404 so the Meilisearch sink can be reviewed and merged independently from the source connector.

Closes #3495.

## Validation
- `cargo test -p iggy_connector_meilisearch_sink`
- `cargo clippy -p iggy_connector_meilisearch_sink --all-targets -- -D warnings`
- `env CARGO_BIN_EXE_iggy-server=... CARGO_BIN_EXE_iggy-connectors=... cargo test -p integration --test mod -- connectors::meilisearch::meilisearch_sink`
- `git diff --check`


Cargo.toml
ryerraguntla, line 156: dashmap and once_cell listed in [dependencies], never imported in src/lib.rs. [package.metadata.cargo-machete] ignored = ["dashmap", "once_cell"] suppresses the machete warning. Same issue as PR 3498 source connector - copy-paste from template without cleanup. Remove both from [dependencies] and the ignored list.

ryerraguntla: Please see the comment about dashmap and applicable for onecell.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+                Ok(())
+            }
+            None if self.config.create_index_if_not_exists => self.create_index(client).await,
+            None => Err(Error::InitError(format!(
+                "Meilisearch index '{}' does not exist and create_index_if_not_exists=false",
+                self.config.index
+            ))),
+        }
+    }
+
```
ryerraguntla, line 309: fn get_index_if_exists — hand-rolled retry loop that calls client.get_index(...) with no tokio::time::timeout wrapper:

**match client.get_index(&self.config.index).await {**
Every other retried operation goes through retry_sdk_operation_with_attempts which applies self.config.timeout per attempt. This one does not. A hung HTTP connection at open() time blocks the entire open call indefinitely. Route through retry_sdk_open_operation (the special case for IndexNotFound can be modeled as Ok(None) inside the closure by catching that arm before passing to the retry wrapper).

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+        ))
+    }
+
+    async fn wait_for_task_status(
+        &self,
+        client: &Client,
+        task: TaskInfo,
+        max_attempts: u32,
+    ) -> Result<Task, Error> {
+        let task_uid = task.get_task_uid();
```
ryerraguntla, line 605: elapsed_time only accounts for task_poll_interval sleeps, not the time spent inside retry_sdk_operation_with_attempts("get task status", ...):
`let mut elapsed_time = Duration::ZERO;
while self.config.task_timeout > elapsed_time {
    let status = self.retry_sdk_operation_with_attempts(...).await?;
    // ... network I/O not counted ...
    tokio::time::sleep(self.config.task_poll_interval).await;
    elapsed_time += self.config.task_poll_interval;
}`
With task_timeout = 30s, task_poll_interval = 100ms, and each network round-trip taking e.g. 1s, the actual wall time before declaring timeout is 30s / 100ms × 1.1s ≈ 5.5 minutes. Replace with tokio::time::Instant::now() and check actual elapsed wall time.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+                        "data_type": "raw",
+                        "data_encoding": ENCODING_BASE64,
+                    }),
+                }
+            }
+            Payload::Text(text) => json!({
+                "text": text,
+                "data_type": "text",
+            }),
+            _ => {
```
ryerraguntla, line 379: README/code contract mismatch:

prepare_document returns Err(Error::InvalidRecordValue(...)) for unsupported payload schemas:
`_ => {
    return Err(Error::InvalidRecordValue(format!(
        "Unsupported payload format for Meilisearch sink: {}",
        messages_metadata.schema
    )));
}`
The README states: "Unsupported payload schemas fail the batch instead of being silently dropped." The code silently drops them (with a warn, no batch failure). Either change the code to propagate the error (fail the batch) or fix the README. The current behavior is the one existing sinks apply to malformed records, so the code is likely intentional - the README is wrong.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+            .ok_or_else(|| Error::Connection("Meilisearch client not initialized".to_string()))?;
+
+        let messages_count = messages.len();
+        let mut documents = Vec::with_capacity(messages.len());
+        let mut invalid_records = 0usize;
+        for message in messages {
+            match self.prepare_document(topic_metadata, &messages_metadata, message) {
+                Ok(Some(document)) => documents.push(document),
+                Ok(None) => {}
+                Err(Error::InvalidRecordValue(reason)) => {
```
ryerraguntla, line 821: catches that variant and drops the message with a warn:
```
Err(Error::InvalidRecordValue(reason)) => {
    invalid_records += 1;
    warn!("Dropping invalid Meilisearch sink record...");
}
```
The README states: "Unsupported payload schemas fail the batch instead of being silently dropped." The code silently drops them (with a warn, no batch failure). Either change the code to propagate the error (fail the batch) or fix the README. The current behavior is the one existing sinks apply to malformed records, so the code is likely intentional - the README is wrong.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+            _ => {
+                return Err(Error::InvalidRecordValue(format!(
+                    "Unsupported payload format for Meilisearch sink: {}",
+                    messages_metadata.schema
+                )));
+            }
+        };
+
+        let object = document
+            .as_object_mut()
```
ryerraguntla: l```
et object = document
    .as_object_mut()
    .expect("document_from_json_value always returns an object");
```
The invariant holds (document_from_json_value always returns Value::Object), but expect() on a non-I/O invariant panics in production. Use unwrap_or_else(|| unreachable!(...)) or restructure document_from_json_value to return Map<String, Value> directly so the call site doesn't need a cast.

File: core/connectors/sinks/meilisearch_sink/src/lib.rs [hunk 6]
```diff
@@ -0,0 +1,1074 @@
...
+    ]);
+    let encoded = serde_json::to_vec(&components)
+        .map(|bytes| general_purpose::URL_SAFE_NO_PAD.encode(bytes))
+        .expect("generated ID components are always JSON-serializable");
+    format!("iggy_{encoded}")
+}
+
+fn insert_metadata_field(object: &mut Map<String, Value>, field: &str, value: Value) {
+    if object.contains_key(field) {
+        warn!(
```
ryerraguntla: [nit]
With include_metadata = true and a high-throughput stream where user payloads consistently include one of the 9 metadata field names (e.g. iggy_offset), this emits a warn! per message per matching field. Downgrade to debug!.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+        &self,
+        operation: &str,
+        max_attempts: u32,
+        mut operation_fn: Op,
+    ) -> Result<T, Error>
+    where
+        Op: FnMut() -> Fut,
+        Fut: Future<Output = Result<T, MeilisearchSdkError>>,
+    {
+        let max_attempts = max_attempts.max(1);
```
ryerraguntla: redundant .max(1) inside retry_sdk_operation_with_attempts when the config From conversion already clamps (L145, L149-152). Dead code.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1074 @@
...
+        }
+        MeilisearchSdkError::MeilisearchCommunication(communication_error) => {
+            communication_error.status_code == 429 || communication_error.status_code >= 500
+        }
+        MeilisearchSdkError::HttpError(_) | MeilisearchSdkError::Timeout => true,
+        _ => false,
+    }
+}
+
+struct SinkSdkError(MeilisearchSdkError);
```
ryerraguntla: [nit] SinkSdkError — private newtypes implementing From only to be called via map_sdk_error(e) = Error::from(SinkSdkError::from(e)). The wrapper adds indirection with no callsite benefit; either inline the match directly in map_sdk_error (as the source connector does) or drop map_sdk_error and call SinkSdkError(e).into() directly.

ryerraguntla:
overall correct approach - JSON-object document model with generated PK is sound, wait_for_tasks + partial-error accounting is thoughtful. test coverage is present but there is 
coverage gap — only happy-path (2 JSON messages indexed, iggy_id present). No test for: document_action = update semantics, include_metadata = false, raw-payload base64 fallback, or wait_for_tasks = false fire-and-forget path. Not a merge-blocker given this is the initial connector, but tracking it.  Along with coverage, please take care of the above mentioned issues.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+        info!("Meilisearch sink connector with ID: {} is closed.", self.id);
+        Ok(())
+    }
+}
+
+fn generated_document_id(
+    topic_metadata: &TopicMetadata,
+    messages_metadata: &MessagesMetadata,
+    message: &ConsumedMessage,
+) -> String {
```
ryerraguntla, line 907: 733-742 — json!([..., message.id]) where message.id: u128 (ConsumedMessage sdk/src/lib.rs:361). serde_json 1.0.150 returns Err(NumberOutOfRange) for u128 > u64::MAX;
  json! macro calls .expect() → panic. All tests use id: 42 — bug undetected. iggy_message_id at line 317-319 correctly uses .to_string(). **Fix: use message.id.to_string() in the json! array too.**

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+        Fut: Future<Output = Result<T, MeilisearchSdkError>>,
+    {
+        let mut attempt = 0u32;
+
+        loop {
+            let result = tokio::time::timeout(self.config.timeout, operation_fn()).await;
+            match result {
+                Ok(Ok(value)) => return Ok(value),
+                Ok(Err(error)) => {
+                    attempt += 1;
```
ryerraguntla: **566,582 + README.md:20,23** — max_retries=3 (.max(1)) gives 3 total attempts (2 retries after first). README says "Maximum transient retry attempts" implying 3 retries = 4 total. Log **line :576** prints attempt {attempt}/{max_attempts} reinforcing the mismatch. Same bug in max_open_retries. **Fix: rename to max_attempts or rewrite guard to attempt <= max_retries.**

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+                state.documents_enqueued += accepted;
+                if self.config.wait_for_tasks {
+                    state.documents_indexed += accepted;
+                }
+                info!(
+                    "Accepted {} of {} messages into Meilisearch index '{}'",
+                    accepted, messages_count, self.config.index
+                );
+                Ok(())
+            }
```
ryerraguntla, line 851: **697-706** — consume() error arm increments state.documents_indexed += partial_error.accepted (line 700-701) when wait_for_tasks=true. But PartialIndexError at line 416-419
  fires when wait_for_task fails AFTER submit — Meilisearch accepted the task but terminal state is unknown. Counting unconfirmed docs as indexed is wrong. **Fix: in error arm, only increment documents_enqueued, not
  documents_indexed.**

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+    async fn wait_for_task_status(
+        &self,
+        client: &Client,
+        task: TaskInfo,
+        max_attempts: u32,
+    ) -> Result<Task, Error> {
+        let task_uid = task.get_task_uid();
+        let started = Instant::now();
+
+        loop {
```
ryerraguntla, line 608: **492-516** — wait_for_task_status checks started.elapsed() >= task_timeout before each poll (line 492), but retry_sdk_operation_with_attempts (line 496) is uncancellable and
  can itself take timeout × max_retries = 30s × 3 = 90s per iteration. Worst case total: task_timeout + timeout × max_retries = 120s against stated 30s task_timeout. **Fix: pass remaining = 
  task_timeout.saturating_sub(started.elapsed()) as an outer tokio::time::timeout around each retry_sdk_operation_with_attempts call.**

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+                    messages_metadata.schema
+                )));
+            }
+        };
+
+        document
+            .entry(self.config.primary_key.clone())
+            .or_insert_with(|| Value::String(generated_id.clone()));
+
+        if self.config.include_metadata {
```
ryerraguntla, line 401: **312-314** — when primary_key != "iggy_id" and include_metadata=true, document gets BOTH self.config.primary_key set to generated_id (line 308) AND hardcoded  DEFAULT_PRIMARY_KEY ("iggy_id") set to same generated_id (line 313). Two fields, same value, undocumented. README line 39-40 says "metadata fields only inserted when absent" — gives no hint iggy_id is  unconditional. Test preserves_existing_configured_primary_key (line 965-984) asserts and thus cements this undocumented behavior. **Fix: document in README as explicit guarantee, or gate line 312-314 on
  self.config.primary_key != DEFAULT_PRIMARY_KEY.**

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1075 @@
...
+            .retry_sdk_open_operation("create index", || {
+                client.create_index(&self.config.index, Some(&self.config.primary_key))
+            })
+            .await?;
+        self.wait_for_index_creation_task(client, task).await?;
+
+        info!("Created Meilisearch index '{}'", self.config.index);
+        Ok(())
+    }
+
```
ryerraguntla, line 337: **[nit]** **257-356** — prepare_document return type Result<Option<Value>, Error> but code only returns Ok(Some(...)) or Err(...). Ok(None) path unreachable. Matching arm at line 665
  (Ok(None) => {}) is dead. **Fix: change return type to Result<Value, Error>, update consume() to match.**

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+    if !url.username().is_empty() {
+        let _ = url.set_username("");
+    }
+    if url.password().is_some() {
+        let _ = url.set_password(None);
+    }
+    url.to_string()
+}
+
```
ryerraguntla, line 989: normalize_host preserves URL path component; http://host/prefix passes through unchanged; Meilisearch SDK appends routes (/indexes/...) to this base → produces 
  http://host/prefix/indexes/... instead of http://host/indexes/...; silent misconfiguration with zero error. Integration test L1058-1059 input "user:pass@localhost:7700/indexes" → output 
  "http://localhost:7700/indexes" confirms path is kept. Fix: strip to scheme+host+port only; reject or warn on non-empty path.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+        .map_err(|error| PartialIndexError { accepted: 0, error })?;
+        self.wait_for_task(client, task)
+            .await
+            .map_err(|error| PartialIndexError {
+                accepted: documents.len(),
+                error,
+            })?;
+        Ok(documents.len())
+    }
+
```
ryerraguntla, line 539: 422-428 + README — wait_for_tasks=false path: wait_for_task returns Ok(()) immediately without spawning watcher; consume() returns Ok; runtime commits consumer offset; async Meilisearch task 
  failures (disk full, schema conflict, quota) are permanently invisible; no retry, no log, no counter increment. At-least-once guarantee is broken for this mode. Call chain: consume → index_document_chunk → 
  wait_for_task (immediate Ok) → accepted count returned → offset advanced by runtime. Fix: document explicitly as "fire-and-forget, no delivery guarantee" with wait_for_tasks=false, or add background failure 
  logging.

core/connectors/sinks/meilisearch_sink/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+publish = false
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+iggy_common = { workspace = true }
+iggy_connector_sdk = { workspace = true }
```
ryerraguntla, line 39: with  Cargo.lock — direct reqwest = "0.13.4" dep alongside meilisearch-sdk which pulls reqwest = "0.12.28"; cdylib links two reqwest versions and two TLS stacks (rustls versions differ); only
  use is use reqwest::Url at lib.rs:37. Fix: drop reqwest dep, add url = { workspace = true }, change import to use url::Url.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+                }
+                info!(
+                    "Accepted {} of {} messages into Meilisearch index '{}'",
+                    accepted, messages_count, self.config.index
+                );
+                Ok(())
+            }
+            Err(partial_error) => {
+                let mut state = self.state.lock().await;
+                state.documents_enqueued += partial_error.accepted;
```
ryerraguntla, line 854: state.errors_count += 1 for entire chunk failure regardless of batch size; inconsistent with invalid_records at L677 which increments per-document; a 1000-doc batch failure reports 
  errors_count=1; operator metrics untrustworthy. Fix: += chunk.len() (or document semantics explicitly as "batch-level error count").

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+            }
+        }
+        if invalid_records > 0 {
+            self.record_errors(invalid_records).await;
+        }
+
+        if documents.is_empty() {
+            return Ok(());
+        }
+
```
ryerraguntla, line 838: documents_indexed only incremented inside if self.config.wait_for_tasks; stays 0 for entire connector lifetime when wait_for_tasks=false; close() logs 0 documents indexed misleading operators who assume nothing was indexed. Fix: rename to documents_confirmed or add a (N/A: wait_for_tasks=false) annotation in close log.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+    }
+
+    fn prepare_document(
+        &self,
+        topic_metadata: &TopicMetadata,
+        messages_metadata: &MessagesMetadata,
+        message: ConsumedMessage,
+    ) -> Result<Value, Error> {
+        let generated_id = generated_document_id(topic_metadata, messages_metadata, &message);
+        let ConsumedMessage {
```
ryerraguntla, line 344: generated_document_id() called eagerly per message before entry().or_insert_with() check at L305; 4 heap allocs (json! Value, to_vec Vec<u8>, base64 String, format String) wasted per 
  message when document already carries the primary key field. Fix: move call into or_insert_with(|| Value::String(generated_document_id(...))) closure.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+            .unwrap_or(DEFAULT_CREATE_INDEX_IF_NOT_EXISTS);
+        let include_metadata = config.include_metadata.unwrap_or(DEFAULT_INCLUDE_METADATA);
+        let batch_size = config.batch_size.unwrap_or(DEFAULT_BATCH_SIZE).max(1);
+        let timeout = parse_duration(config.timeout.as_deref(), DEFAULT_TIMEOUT);
+        let wait_for_tasks = config.wait_for_tasks.unwrap_or(DEFAULT_WAIT_FOR_TASKS);
+        let task_timeout = parse_duration(config.task_timeout.as_deref(), DEFAULT_TASK_TIMEOUT);
+        let task_poll_interval = parse_duration(
+            config.task_poll_interval.as_deref(),
+            DEFAULT_TASK_POLL_INTERVAL,
+        );
```
ryerraguntla, line 147: max_open_retries governs two semantically unrelated retry budgets: (1) SDK call retries in retry_sdk_open_operation (health check, get-index, create-index), and (2) 
  per-iteration get_task poll retry budget inside wait_for_task_status when called from wait_for_index_creation_task at L437; same field, two meanings; undocumented in README and config. Fix: document dual role 
  explicitly or split into max_open_call_retries and max_task_poll_retries.

core/connectors/sinks/meilisearch_sink/config.toml
```diff
@@ -0,0 +1,43 @@
...
+
+[[streams]]
+stream = "test_stream"
+topics = ["test_topic"]
+schema = "json"
+batch_length = 100
+poll_interval = "5ms"
+consumer_group = "meilisearch_sink_cg"
+
+[plugin_config]
```
ryerraguntla, line 34: example file sets batch_size = 100; DEFAULT_BATCH_SIZE = 1000 at lib.rs:50 and README states "Defaults to 1000"; operators copy config.toml as starting point and get non-default behavior silently. Fix: align example to 1000, or add comment explaining why example uses a smaller value.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+        let started = Instant::now();
+
+        loop {
+            let remaining = self.config.task_timeout.saturating_sub(started.elapsed());
+            if remaining.is_zero() {
+                break;
+            }
+
+            let status = match tokio::time::timeout(
+                remaining,
```
ryerraguntla, line 615: **[nit]** task.clone() per poll iteration; task_uid already extracted; clone dominated by 100ms poll interval.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+    }
+
+    let with_scheme = if trimmed.starts_with("http://") || trimmed.starts_with("https://") {
+        trimmed.to_string()
+    } else {
+        format!("http://{trimmed}")
+    };
+    let url = Url::parse(&with_scheme)
+        .map_err(|error| Error::Connection(format!("Invalid Meilisearch URL: {error}")))?;
+    let mut host = url.to_string();
```
ryerraguntla: **[nit]** normalize_host while-pop loop → trim_end_matches('/') cleaner.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+                    messages_metadata.schema
+                )));
+            }
+        };
+
+        document
+            .entry(self.config.primary_key.clone())
+            .or_insert_with(|| Value::String(generated_id.clone()));
+
+        if self.config.include_metadata {
```
ryerraguntla, line 401: **[nit]** self.config.primary_key.clone() per document for entry() key; Map::entry takes String by value; minor but avoidable with get+insert.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+            if self.config.primary_key != DEFAULT_PRIMARY_KEY {
+                document
+                    .entry(DEFAULT_PRIMARY_KEY.to_string())
+                    .or_insert_with(|| Value::String(generated_id));
+            }
+            insert_metadata_field(
```
ryerraguntla: **[nit]** insert_metadata_field calls field.to_string() x9-11 per doc; unavoidable with Map<String, Value>.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+        Err(Error::HttpRequestFailed(
+            "Meilisearch task did not reach a terminal state".to_string(),
+        ))
+    }
+
+    async fn wait_for_task_status(
+        &self,
+        client: &Client,
+        task: TaskInfo,
+        max_retries: u32,
```
ryerraguntla: **[nit]** wait_for_task_status parameter named max_retries controls per-call retry budget inside inner retry fn, not poll iteration count; rename to max_get_task_retries.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+}
+
+fn is_transient_sdk_error(error: &MeilisearchSdkError) -> bool {
+    match error {
+        MeilisearchSdkError::Meilisearch(meilisearch_error) => {
+            meilisearch_error.error_type == MeilisearchErrorType::Internal
+        }
+        MeilisearchSdkError::MeilisearchCommunication(communication_error) => {
+            communication_error.status_code == 429 || communication_error.status_code >= 500
+        }
```
ryerraguntla, line 1084: **[nit]** MeilisearchCommunication with status_code == 0 (pre-HTTP transport failure) treated as non-transient; add || e.status_code == 0 to transient check.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+                    "Accepted {} of {} messages into Meilisearch index '{}'",
+                    accepted, messages_count, self.config.index
+                );
+                Ok(())
+            }
+            Err(partial_error) => {
+                let mut state = self.state.lock().await;
+                state.documents_enqueued += partial_error.accepted;
+                state.errors_count += 1;
+                drop(state);
```
ryerraguntla, line 859: **[nit]** no error! log before return Err(partial_error.error) in consume(); runtime logs it, but per-connector visibility lost.

core/integration/tests/connectors/meilisearch/meilisearch_sink.rs
```diff
@@ -0,0 +1,81 @@
...
+use iggy::prelude::{IggyMessage, Partitioning};
+use iggy_common::{Identifier, MessageClient};
+use integration::harness::seeds;
+use integration::iggy_harness;
+
+#[iggy_harness(
+    server(connectors_runtime(config_path = "tests/connectors/meilisearch/sink.toml")),
+    seed = seeds::connector_stream
+)]
+async fn meilisearch_sink_indexes_json_messages(
```
ryerraguntla: **[nit]** test name meilisearch_sink_indexes_json_messages violates given_X_when_Y_should_Z convention.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1078 @@
...
+        operation_fn: Op,
+    ) -> Result<T, Error>
+    where
+        Op: FnMut() -> Fut,
+        Fut: Future<Output = Result<T, MeilisearchSdkError>>,
+    {
+        self.retry_sdk_operation_with_retries(operation, self.config.max_open_retries, operation_fn)
+            .await
+    }
+
```
ryerraguntla, line 670: **[nit]** retries counter shared between SDK error path and timeout path; undocumented; behavior bounded but surprising.

ryerraguntla:
@countradooku - few blockers, many nits. One thing which is encouraged is to have more test coverage.   - Integration coverage: no tests for Payload::Raw, Payload::Text, wait_for_tasks=false, custom primary_key, create_index_if_not_exists=false error path. (intro)

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1154 @@
...
+    id: u128,
+) -> String {
+    let components = json!([
+        topic_metadata.stream.as_str(),
+        topic_metadata.topic.as_str(),
+        messages_metadata.partition_id,
+        offset,
+        id.to_string()
+    ]);
+    let encoded = serde_json::to_vec(&components)
```
ryerraguntla, line 930: Line 791 -  .expect("generated ID components are always JSON-serializable") in generated_document_id_from_parts() — called from prepare_document() → consume(). Panics in production if serde_json::to_vec ever fails. Project rule: no panics in libraries. **Fix**: propagate as Error::Serialization via ?, return Result<String, Error> from generated_document_id_from_parts.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1154 @@
...
+                .await
+            }
+        }
+        .map_err(|error| PartialIndexError {
+            accepted: 0,
+            failed: documents.len(),
+            error,
+        })?;
+        self.wait_for_task(client, task)
+            .await
```
ryerraguntla, line 531: 437-443 — index_document_chunk() sets PartialIndexError { accepted: documents.len(), failed: documents.len() } when wait_for_task fails. Both fields = same count. In consume(): 
  documents_enqueued += partial_error.accepted AND errors_count += partial_error.failed — every document counted twice. Stats diverge from reality. **Fix**: accepted = documents.len() (submission succeeded), failed =   0 (outcome unknown, not a failure of submission).

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1154 @@
...
+        mut operation_fn: Op,
+    ) -> Result<T, Error>
+    where
+        Op: FnMut() -> Fut,
+        Fut: Future<Output = Result<T, MeilisearchSdkError>>,
+    {
+        // One retry budget covers both SDK errors and per-attempt timeouts.
+        let mut retries = 0u32;
+
+        loop {
```
ryerraguntla, line 684: 590-628 — retry backoff sleep(delay).await between attempts not bounded by any outer deadline in consume() / open() call paths. With max_retries=3, timeout=30s, max_retry_delay=5s: worst-case 
  wall time = 4×30s + 3×5s = 105s per operation. Neither consume() nor open() wrap retry_sdk_operation_with_retries in an outer timeout. Runtime's sink loop can stall >100s silently. **Fix**: apply min(delay,  budget_remaining) or wrap each retry_sdk_operation call in an outer deadline.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1154 @@
...
+
+    fn create_client(&self) -> Result<Client, Error> {
+        let url = normalize_host(&self.config.url)?;
+        let api_key = self.config.api_key.as_ref().map(|key| key.expose_secret());
+        Client::new(url, api_key).map_err(|error| {
+            Error::Connection(format!("Failed to create Meilisearch client: {error}"))
+        })
+    }
+
+    async fn check_connectivity(&self, client: &Client) -> Result<(), Error> {
```
ryerraguntla, line 216: check_connectivity() returns Err(Error::Connection(...)) immediately on health.status != "available" (e.g. "not_responding" during Meilisearch startup). retry_sdk_open_operation only   retries MeilisearchSdkError; it does not retry a successful HTTP response with non-ready status. Connector fails open() on a transiently degraded node without retrying. **Fix**: wrap status check in a retry loop up  to max_open_retries.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1154 @@
...
+            headers,
+            payload,
+        } = message;
+
+        let mut document = match payload {
+            Payload::Json(value) => {
+                Self::document_from_json_value(owned_value_to_serde_json(&value))
+            }
+            Payload::Raw(bytes) => {
+                let mut bytes_copy = bytes.clone();
```
ryerraguntla, line 359: let mut bytes_copy = bytes.clone() for Payload::Raw is unnecessary. ConsumedMessage is moved into prepare_document(), so bytes is already owned. simd_json requires &mut [u8]. **Fix**: let 
  Payload::Raw(mut bytes) = payload, pass &mut bytes directly, use bytes for base64 fallback. Eliminates full payload clone per raw message.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1196 @@
...
+            origin_timestamp,
+            headers,
+            payload,
+        } = message;
+
+        let mut document = match payload {
+            Payload::Json(value) => {
+                Self::document_from_json_value(owned_value_to_serde_json(&value))
+            }
+            Payload::Raw(mut bytes) => {
```
ryerraguntla: simd_json::from_slice(&mut bytes) mutates input buffer in-place during parse; on Err, general_purpose::STANDARD.encode(&bytes) encodes the garbled bytes, not the original. **Fix:**  clone before parse, encode clone on failure.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1196 @@
...
+                })
+                .await
+            }
+        }
+        .map_err(|error| PartialIndexError {
+            accepted: 0,
+            failed: documents.len(),
+            error,
+        })?;
+        self.wait_for_task(client, task)
```
ryerraguntla, line 530: task-wait failure returns PartialIndexError { accepted: documents.len(), failed: 0 }; consume() at line 765 increments documents_enqueued by those "accepted" and returns Err;  runtime retries; with document_action=update, all previously-submitted docs get a second partial update applied. **Fix:** swap to accepted: 0, failed: documents.len() on task-wait failure.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1196 @@
...
+
+    fn create_client(&self) -> Result<Client, Error> {
+        let url = normalize_host(&self.config.url)?;
+        let api_key = self.config.api_key.as_ref().map(|key| key.expose_secret());
+        Client::new(url, api_key).map_err(|error| {
+            Error::Connection(format!("Failed to create Meilisearch client: {error}"))
+        })
+    }
+
+    async fn check_connectivity(&self, client: &Client) -> Result<(), Error> {
```
ryerraguntla, line 216: check_connectivity outer loop retries max_open_retries+1 times on non-"available" health status; each call to retry_sdk_open_operation internally retries max_open_retries times on  transient errors; total health calls = (max_open_retries+1)^2 = 36 at default max_open_retries=5, bounded by config.timeout wall clock but not visible to operators. **Fix:** flatten to single retry loop or document
  multiplicative behavior.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1196 @@
...
+                    partial_error.accepted,
+                    partial_error.failed,
+                    partial_error.error
+                );
+                Err(partial_error.error)
+            }
+        }
+    }
+
+    async fn close(&mut self) -> Result<(), Error> {
```
ryerraguntla, line 873: close() logs "documents confirmed: 0" when wait_for_tasks=false; in-flight fire-and-forget tasks are abandoned silently; no warning that indexing may be incomplete. **Fix:** add warn! at  close() if wait_for_tasks=false. Also: documents_confirmed at lib.rs:752-756 stays 0 forever under fire-and-forget - operator sees misleading 0 in logs.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1196 @@
...
+            wait_for_tasks: None,
+            task_timeout: None,
+            task_poll_interval: None,
+            max_retries: None,
+            retry_delay: None,
+            max_retry_delay: None,
+            max_open_retries: None,
+        }
+    }
+
```
ryerraguntla, line 1176: integration test - only happy-path JSON covered; missing: Payload::Raw non-JSON binary, Payload::Text, include_metadata=false, wait_for_tasks=false, existing index with mismatched primary key,
  document_action=update, generated ID uniqueness. The simd_json mutation bug (lib.rs 292-306) is not caught by the existing raw_payloads_are_base64_encoded_when_not_json test at lib.rs:1154-1167 because vec![0,1,2,3]
  fails JSON parse immediately before simd_json mutates anything. Fix: add test with near-valid JSON bytes (e.g. b"[1,2,3") to prove base64 fidelity of original.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1261 @@
...
+            Value::Object(object) => object,
+            other => {
+                let mut object = Map::new();
+                object.insert("value".to_string(), other);
+                object
+            }
+        }
+    }
+
+    async fn index_documents(
```
ryerraguntla, line 472: batch partial failure with document_action=update: when chunk N fails, chunks 0..N-1 are already indexed; runtime retries entire batch; update action applies partial updates a second  time to already-succeeded chunks. replace action is idempotent; update is not. No warn at open() when document_action=update. **Fix:** add warn! at open() when action is Update, documenting at-least-once +  non-idempotent.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1261 @@
...
+
+    async fn record_errors(&self, errors: usize) {
+        let mut state = self.state.lock().await;
+        state.errors_count += errors;
+    }
+}
+
+#[async_trait]
+impl Sink for MeilisearchSink {
+    async fn open(&mut self) -> Result<(), Error> {
```
ryerraguntla, line 754: open() has no warn when wait_for_tasks=false. close() now warns, but operators see that too late. **Fix:** add warn! early in open() when !self.config.wait_for_tasks.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1261 @@
...
+                        "Meilisearch health check timed out after {:?} (retry {}/{}). Retrying in {:?}...",
+                        self.config.timeout, retries, self.config.max_open_retries, delay
+                    );
+                    sleep(delay).await;
+                }
+            }
+        }
+    }
+
+    async fn ensure_index_exists(&self, client: &Client) -> Result<(), Error> {
```
ryerraguntla, line 283: ensure_index_exists on Some(_) only logs "already exists"; does not validate existing index's actual primary key against self.config.primary_key. A misconfigured primary_key on an existing index fails silently at first consume() with a permanent Meilisearch error, not at open(). **Fix: r**ead index.primary_key from the returned Index and warn! (or Err) on mismatch.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1261 @@
...
+        let include_metadata = config.include_metadata.unwrap_or(DEFAULT_INCLUDE_METADATA);
+        let batch_size = config.batch_size.unwrap_or(DEFAULT_BATCH_SIZE).max(1);
+        let timeout = parse_duration(config.timeout.as_deref(), DEFAULT_TIMEOUT);
+        let wait_for_tasks = config.wait_for_tasks.unwrap_or(DEFAULT_WAIT_FOR_TASKS);
+        let task_timeout = parse_duration(config.task_timeout.as_deref(), DEFAULT_TASK_TIMEOUT);
+        let task_poll_interval = parse_duration(
+            config.task_poll_interval.as_deref(),
+            DEFAULT_TASK_POLL_INTERVAL,
+        );
+        let max_retries = config.max_retries.unwrap_or(DEFAULT_MAX_RETRIES);
```
ryerraguntla, line 148: retry_delay > max_retry_delay not guarded. exponential_backoff clamps to max, so behavior is technically correct, but operator who sets retry_delay="10s", max_retry_delay="1s" gets silent misconfiguration. **Fix:** swap + warn! (cf. http_sink pattern).

core/integration/tests/connectors/fixtures/meilisearch/container.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    fn wait_for_documents(
+        &self,
+        expected_count: usize,
+    ) -> impl std::future::Future<Output = Result<Vec<serde_json::Value>, TestBinaryError>> + Send
+    {
+        async move {
+            let mut last_count = 0usize;
+            for _ in 0..POLL_ATTEMPTS {
+                if let Ok(documents) = self.list_documents(TEST_INDEX).await {
```
ryerraguntla: wait_for_documents hardcodes TEST_INDEX instead of forwarding the index_name param to list_documents.

ryerraguntla:
@countradooku Looks good except for the few warnings around back indexing and open calls. After the above requested changes, good to go for second reviewer.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+            .map_err(|error| PartialIndexError {
+                accepted: 0,
+                failed: documents.len(),
+                error,
+            })?;
+        Ok(documents.len())
+    }
+
+    async fn wait_for_task(&self, client: &Client, task: TaskInfo) -> Result<(), Error> {
+        if !self.config.wait_for_tasks {
```
ryerraguntla, line 541: lib.rs:521-523 + 818-829 — wait_for_tasks=false returns Ok on submit; async Meilisearch task failures (disk full, schema conflict) invisible, offset committed, silent loss. **Fix**: none required — documented README + warn open/close. Non-blocking.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+                            Value::String(ENCODING_BASE64.to_string()),
+                        ),
+                    ]),
+                }
+            }
+            Payload::Text(text) => Map::from_iter([
+                ("text".to_string(), Value::String(text)),
+                ("data_type".to_string(), Value::String("text".to_string())),
+            ]),
+            _ => {
```
ryerraguntla, line 379: lib.rs:797-812 + 368-373 — unsupported schema (Avro/Protobuf/etc.) dropped, errors_count++, offset commits → permanent silent loss for whole topic if schema unsupported. Fix: matches runtime per-record drop convention; document explicitly / consider DLQ.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1305 @@
...
+        }
+        if invalid_records > 0 {
+            self.record_errors(invalid_records).await;
+        }
+
+        if documents.is_empty() {
+            return Ok(());
+        }
+
+        match self.index_documents(client, documents).await {
```
ryerraguntla, line 839: 818-846 — counter accounting: partial-fail undercounts un-attempted trailing chunks; redelivery double-counts prior chunks; documents_confirmed not bumped for confirmed-durable earlier chunks in Err arm. Metrics-only. **Fix**: label best-effort or account remaining. Keep the correct metrics .

core/connectors/sinks/meilisearch_sink/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_meilisearch_sink"
+version = "0.4.0"
```
ryerraguntla: version 0.4.0, siblings all 0.4.1-edge.1. Version-sync CI flag. Fix: bump to 0.4.1-edge.1.  Same as Source connector

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+#[async_trait]
+impl Sink for MeilisearchSink {
+    async fn open(&mut self) -> Result<(), Error> {
+        info!(
+            "Opening Meilisearch sink connector with ID: {} for URL: {}, index: {}",
+            self.id,
+            sanitize_url_for_log(&self.config.url),
+            self.config.index
+        );
+        if self.config.document_action == MeilisearchDocumentAction::Update {
```
ryerraguntla, line 762: lib.rs:749-754 + README.md:47-61 promise "runtime retries are at-least-once" / idempotent redelivery. But  runtime/src/sink.rs:739-747 calls (consume)(...) as a bare statement — the FFI i32 return is discarded — and process_messages then returns Ok unconditionally. The return Err at sink.rs:426 fires only on postcard serialization failure, never on the plugin's consume Err. Offsets auto-commit via AutoCommitWhen::PollingMessages (sink.rs:522).  This  means  an index failure is swallowed, its offset commits on the next poll, no redelivery.  **Fix**: Couple of options -
 rewrite open() docstring + README to at-most-once-on-error; 
 state failed/dropped batches are NOT redelivered; 
 carry the same caveat siblings ship (s3_sink/README.md, http_sink/README.md). 

Or gate the offset commit on the sink consume return in the runtime

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+            .await
+            .map_err(|error| PartialIndexError {
+                accepted: 0,
+                failed: documents.len(),
+                error,
+            })?;
+        Ok(documents.len())
+    }
+
+    async fn wait_for_task(&self, client: &Client, task: TaskInfo) -> Result<(), Error> {
```
ryerraguntla, line 540: lib.rs:528-531 returns Ok on task submit before the Meilisearch task reaches terminal; async task failure is never observed or counted,  **Fix**: keep default true (config.toml:43); document that false = fire-and-forget with no durability, on top of at-most-once-on-error.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+
+fn normalize_host(raw: &str) -> Result<String, Error> {
+    let trimmed = raw.trim();
+    if trimmed.is_empty() {
+        return Err(Error::Connection(
+            "Invalid Meilisearch URL: host cannot be empty".to_string(),
+        ));
+    }
+
+    let with_scheme = if trimmed.starts_with("http://") || trimmed.starts_with("https://") {
```
ryerraguntla, line 998: lib.rs:960-964 normalize_host prepends http:// to a bare host; operator writing meili.prod:7700 intending TLS gets the Bearer api_key (ms_lib.rs:199-200) sent cleartext.  **Fix**: when api_key is set and the resolved scheme is http on a non-loopback host, warn! or require an explicit http://.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+        .map(|bytes| general_purpose::URL_SAFE_NO_PAD.encode(bytes))
+        .map_err(|error| {
+            Error::Serialization(format!(
+                "Failed to serialize generated document ID: {error}"
+            ))
+        })?;
+    Ok(format!("iggy_{encoded}"))
+}
+
+fn insert_metadata_field(object: &mut Map<String, Value>, field: &str, value: Value) {
```
ryerraguntla: lib.rs:927-935 preserves existing fields, so a payload carrying iggy_offset/iggy_stream shadows the real values (lib.rs:1205-1220) — audit fields lie. Documented (README:42-45) but a silent integrity trap. **Fix:** force-overwrite reserved iggy_* provenance fields, or namespace them under one non-collidable object (e.g. _iggy).

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+            checksum,
+            timestamp,
+            origin_timestamp,
+            headers,
+            payload,
+        } = message;
+
+        let mut document = match payload {
+            Payload::Json(value) => {
+                Self::document_from_json_value(owned_value_to_serde_json(&value))
```
ryerraguntla: lib.rs:345,350 owned_value_to_serde_json(&value) deep-clones the whole parsed doc (every string + key re-allocated via convert.rs) then drops the original. Per-message. **Fix:** keep the simd_json OwnedValue (it implements Serialize), insert metadata into it, hand to the SDK directly; or add a consuming From<OwnedValue> that moves.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+                    )?,
+                };
+                document.insert(DEFAULT_PRIMARY_KEY.to_string(), Value::String(id));
+            }
+            insert_metadata_field(
+                &mut document,
+                "iggy_message_id",
+                Value::String(message_id.to_string()),
+            );
+            insert_metadata_field(&mut document, "iggy_offset", Value::from(offset));
```
ryerraguntla: lib.rs:411,427,428,432 emit offset/checksum/timestamp as raw JSON numbers; Meilisearch/JS store as f64 → values >2^53 corrupt at rest (checksum spans full u64). message_id is correctly stringified (:409),**fix:** perf warns stringifying in the loop adds allocs. Fix: stringify checksum only (widest range, used for integrity), document offset/timestamp.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+
+        let mut generated_id = None;
+        if !document.contains_key(self.config.primary_key.as_str()) {
+            let value = generated_document_id_from_parts(
+                topic_metadata,
+                messages_metadata,
+                offset,
+                message_id,
+            )?;
+            document.insert(
```
ryerraguntla: lib.rs:386-388 Value::String(value.clone()) + retained generated_id is a per-message dead clone when primary_key == iggy_id (the default). (perf P.3, intro) **Fix**: insert by move; set generated_id only when primary_key != DEFAULT_PRIMARY_KEY.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+            _ => {
+                return Err(Error::InvalidRecordValue(format!(
+                    "Unsupported payload format for Meilisearch sink: {}",
+                    messages_metadata.schema
+                )));
+            }
+        };
+
+        let mut generated_id = None;
+        if !document.contains_key(self.config.primary_key.as_str()) {
```
ryerraguntla, line 388: lib.rs:377 skips stable-id injection when the payload already carries the configured PK; a non-unique user PK silently makes two messages collide under add_or_replace. (storage S.2 / distsys D.6) **Fix**: document that idempotency holds only when the PK is absent or unique.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+#[serde(rename_all = "snake_case")]
+pub enum MeilisearchDocumentAction {
+    #[default]
+    Replace,
+    Update,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct MeilisearchSinkConfig {
+    pub url: String,
```
ryerraguntla, line 81: lib.rs:82,163 — empty index reaches client.get_index("") → opaque error at open() (fails fast, not silent). **Fix**: reject blank index in the From impl, symmetric to normalize_host.

core/connectors/sinks/meilisearch_sink/README.md
```diff
@@ -0,0 +1,61 @@
+# Meilisearch Sink Connector
+
+A sink connector that consumes messages from Iggy streams and writes them to a
+Meilisearch index through the official Rust SDK.
+
+## Configuration
+
```
ryerraguntla, line 7: [niit] README.md:6-23 is bullet-only; every sibling ships a [plugin_config] block (surrealdb/quickwit/postgres). **Fix**: add one.

core/connectors/sinks/meilisearch_sink/config.toml
```diff
@@ -0,0 +1,43 @@
...
+path = "../../target/release/libiggy_connector_meilisearch_sink"
+verbose = false
+
+[[streams]]
+stream = "test_stream"
+topics = ["test_topic"]
+schema = "json"
+batch_length = 100
+poll_interval = "5ms"
+consumer_group = "meilisearch_sink_cg"
```
ryerraguntla: config.toml:32 /fixture.rs:82 use meilisearch_sink_cg; siblings drop the _cg. **Fix**: meilisearch_sink.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -0,0 +1,1316 @@
...
+
+    fn create_client(&self) -> Result<Client, Error> {
+        let url = normalize_host(&self.config.url)?;
+        let api_key = self.config.api_key.as_ref().map(|key| key.expose_secret());
+        Client::new(url, api_key).map_err(|error| {
+            Error::Connection(format!("Failed to create Meilisearch client: {error}"))
+        })
+    }
+
+    async fn check_connectivity(&self, client: &Client) -> Result<(), Error> {
```
ryerraguntla, line 216: lib.rs:205-269 total = (max_open_retries+1) × timeout + backoff. Bounded, terminates, but misleadingly named. **Fix**: document or add a total open budget.

core/connectors/sinks/meilisearch_sink/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_meilisearch_sink"
+version = "0.4.1-edge.1"
+description = "Iggy Meilisearch sink connector"
+edition = "2024"
+license = "Apache-2.0"
```
ryerraguntla, line 23: Cargo.toml:24 keywords drop "sink"; bytes.clone() (:348) and per-message IggyTimestamp::now() (:437) are pre-surfaced sibling patterns, low value; confirm third-party-licenses.sh resolves the new meilisearch-sdk tree + verify its TLS backend is rustls.

ryerraguntla:
Mostly minor and nits. Almost  there.

ryerraguntla:
Few comments/Nits : Please compare the config params like batch size and timeout have the same default values like other connectors. Regarding , explain what the timeout value means. Is it for every retry or cumulative retries timeout value. Give an example in the readme.md file.

Other than this observation, I approve the changes and good for merge.

ryerraguntla:
Need a second reviewer before merging.

ryerraguntla (approved):
Other than the nits mentioned https://github.com/apache/iggy/pull/3497/#issuecomment-5111948025. I approve.

slbotbm:
Agree with @ryerraguntla 

@countradooku Could you resolve the conflicts?

foreign/cpp/tests/e2e/client.cpp
slbotbm, line 1: Please inline the code wherever it is called. I'd rather not have random namespaces with ad-hoc functions since it makes maintainability of the code hard.

slbotbm (changes requested):

foreign/cpp/tests/e2e/client.cpp
hubcio, line 1: why is this file even changed? this is meilisearch sink PR, not cpp sdk change

hubcio (changes requested):

hubcio (approved):

slbotbm (approved):

PR #3498, opened 2026-06-15
feat(connectors): add Meilisearch source connector

### LINKED ISSUE #3496: Add Meilisearch source connector
## Summary
Add a Meilisearch source connector that polls documents from a Meilisearch index and produces JSON messages into Iggy.

## Motivation
Users should be able to ingest Meilisearch documents into Iggy streams with stable cursor-based polling and persisted connector state.

## Scope
- Add the source connector crate and sample runtime config.
- Document source configuration, retry behavior, and primary-key cursor requirements.
- Add Docker-backed integration coverage for producing indexed documents.

countradooku:
## Summary
- add the Meilisearch source connector using the official `meilisearch-sdk`
- document source configuration, retry behavior, and primary-key cursor requirements
- add Docker-backed integration coverage for producing indexed Meilisearch documents into Iggy

## Motivation
This splits the source portion out of #3404 so the Meilisearch source can be reviewed and merged independently from the sink connector.

Closes #3496.

## Validation
- `cargo test -p iggy_connector_meilisearch_source`
- `cargo clippy -p iggy_connector_meilisearch_source --all-targets -- -D warnings`
- `env CARGO_BIN_EXE_iggy-server=... CARGO_BIN_EXE_iggy-connectors=... cargo test -p integration --test mod -- connectors::meilisearch::meilisearch_source`
- `git diff --check`


Cargo.toml
ryerraguntla, line 157: dashmap and once_cell listed in [dependencies] but never imported in src/lib.rs. [package.metadata.cargo-machete] ignored silences machete to hide this.
Both crates must be removed from [dependencies] and from the ignored list (other sources carry them because they actually use DashMap for in-flight dedup or once_cell for lazy statics; this one does neither).

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    include_metadata: bool,
+    timeout: Duration,
+    max_retries: u32,
+    retry_delay: Duration,
+    max_retry_delay: Duration,
+    max_open_retries: u32,
+}
+
+impl From<MeilisearchSourceConfig> for ResolvedMeilisearchSourceConfig {
+    fn from(config: MeilisearchSourceConfig) -> Self {
```
ryerraguntla, line 100: max_retries and max_open_retries clamped to .max(1) at config resolution:

```
let max_retries = config.max_retries.unwrap_or(DEFAULT_MAX_RETRIES).max(1);
let max_open_retries = config.max_open_retries.unwrap_or(DEFAULT_MAX_OPEN_RETRIES).max(1);
```
then clamped again inside the loop at L346:

`let max_attempts = max_attempts.max(1);
`
double guard means 0 is unreachable and the inner guard is dead. Remove one.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    async fn retry_sdk_operation_with_attempts<T, Fut, Op>(
+        &self,
+        operation: &str,
+        max_attempts: u32,
+        mut operation_fn: Op,
+    ) -> Result<T, Error>
+    where
+        Op: FnMut() -> Fut,
+        Fut: Future<Output = Result<T, MeilisearchSdkError>>,
+    {
```
ryerraguntla, line 452: semantics — max_retries = 3 → max_attempts = 3 → 3 total attempts (2 retries) because attempt is incremented before the attempt < max_attempts check. User reads max_retries = 3 as "retry 3 times" = 4 total attempts. Consistent with postgres_source naming, but the config field and README both say "retries"; the log prints attempt 3/3 for what is in fact the last attempt.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        .transpose()
+}
+
+fn primary_key_filter_literal(value: &Value) -> Result<String, Error> {
+    match value {
+        Value::Number(_) => serde_json::to_string(value).map_err(|error| {
+            Error::Serialization(format!(
+                "Failed to serialize Meilisearch primary key: {error}"
+            ))
+        }),
```
ryerraguntla: primary_key_filter_literal rejects non-numeric primary keys at poll time, not at open():

```
_ => Err(Error::InvalidConfigValue(
    "Meilisearch source primary key values must be numbers with getmeili/meilisearch:v1.13"
        .to_string(),
))
```
String PKs (e.g. UUIDs) are common in Meilisearch. The failure is silent until the first non-empty poll batch; the operator has no warning at startup. The error message hardcodes a Docker image tag that will drift as the connector's pinned image is upgraded. Move validation into open() via a probe document or document the limitation prominently in the open() log.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+                    }
+                    let delay = jitter(exponential_backoff(
+                        self.config.retry_delay,
+                        attempt,
+                        self.config.max_retry_delay,
+                    ));
+                    warn!(
+                        "Meilisearch {operation} failed (attempt {attempt}/{max_attempts}): {error}. Retrying in {delay:?}..."
+                    );
+                    tokio::time::sleep(delay).await;
```
ryerraguntla: [nit] sleep imported via use tokio::{sync::Mutex, time::sleep} on L35 but used via full path tokio::time::sleep(delay).await inside retry_sdk_operation_with_attempts. Inconsistent; use the imported alias.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+                    }
+                    let delay = jitter(exponential_backoff(
+                        self.config.retry_delay,
+                        attempt,
+                        self.config.max_retry_delay,
+                    ));
+                    warn!(
+                        "Meilisearch {operation} timed out after {:?} (attempt {attempt}/{max_attempts}). Retrying in {delay:?}...",
+                        self.config.timeout
+                    );
```
ryerraguntla, line 490: [nit] sleep imported via use tokio::{sync::Mutex, time::sleep} on L35 but used via full path tokio::time::sleep(delay).await inside retry_sdk_operation_with_attempts. Inconsistent; use the imported alias.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        };
+
+        Ok(ProducedMessages {
+            schema: Schema::Json,
+            messages,
+            state: persisted_state,
+        })
+    }
+
+    async fn close(&mut self) -> Result<(), Error> {
```
ryerraguntla, line 549: explicit drop(state) in close():

```
let state = self.state.lock().await;
info!(...);
drop(state);
self.client = None;
```
replace with a block scope; drop() is valid but a scope expresses the intent more clearly.

core/integration/tests/connectors/fixtures/meilisearch/container.rs
```diff
@@ -0,0 +1,341 @@
...
+                .http_client()
+                .post(&url)
+                .query(&[("primaryKey", "id")])
+                .json(&documents)
+                .send()
+                .await
+                .map_err(|e| TestBinaryError::InvalidState {
+                    message: format!("Failed to index Meilisearch documents: {e}"),
+                })?;
+
```
ryerraguntla, line 309: returns Ok(()) on HTTP 202 without waiting for the Meilisearch indexing task to finish:

```
if response.status().is_success() {
    return Ok(());
}
```
Meilisearch indexing is async; 202 means enqueued, not applied. When the task fails (e.g. schema conflict, disk full), the fixture returns Ok(()) and wait_for_documents times out with "Expected 2 documents, found 0" - hiding the real failure cause. Pattern from create_source_index is correct: call parse_task_response and then wait_for_task. Apply the same to index_documents.

ryerraguntla:
overall correct direction - cursor-by-primary-key is the right shape for Meilisearch, test coverage is present but coverage gap — only happy-path covered. No test for: (a) state restored across restart → cursor starts from last_primary_key, not from zero; (b) second poll produces zero duplicates. refer The elasticsearch_source.rs has state_persists_across_connector_restart as a pattern. For a cursor-based source, duplicate-freedom is the core invariant; it should be integration-tested.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    primary_key: &str,
+    last_primary_key: Option<&Value>,
+) -> Result<Option<String>, Error> {
+    last_primary_key
+        .map(|value| {
+            primary_key_filter_literal(value).map(|literal| format!("{primary_key} > {literal}"))
+        })
+        .transpose()
+}
+
```
ryerraguntla, line 693: 540-551 — primary_key_filter_literal accepts Value::Number(_) which includes f64. serde_json 1.0.150 without arbitrary_precision uses f64 for all numbers. Integer PKs > 2^53 lose precision
  silently: cursor advances to wrong value, documents near precision boundary permanently skipped with no error. Additionally, float-typed PKs (e.g. 1.5) produce id > 1.5 filters, which Meilisearch accepts but
  whose ordering semantics differ from integers. **Fix: guard with value.is_i64() || value.is_u64(); reject floats.**

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        }
+        MeilisearchSdkError::HttpError(_) | MeilisearchSdkError::Timeout => true,
+        _ => false,
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
ryerraguntla, line 843: 606-738 — Missing 3 of 4 mandatory canonical state tests per connector-testing SKILL: given_no_state_should_start_fresh, given_invalid_state_should_start_fresh,
  state_should_be_serializable_and_deserializable. Only given_persisted_state_should_restore_last_primary_key (line 635) present. State.last_primary_key: Option<Value> msgpack round-trip needs explicit
  verification — Value::Number to msgpack to Value differs from Value::String. Reference: random_source/src/lib.rs:217-275. Note: rmp_serde available transitively via iggy_connector_sdk, no new dep needed.

core/integration/tests/connectors/fixtures/meilisearch/container.rs
```diff
@@ -0,0 +1,332 @@
...
+}
+
+pub struct MeilisearchContainer {
+    #[allow(dead_code)]
+    container: ContainerAsync<GenericImage>,
+    pub base_url: String,
+}
+
+impl MeilisearchContainer {
+    pub async fn start() -> Result<Self, TestBinaryError> {
```
ryerraguntla, line 67: 67 — unique_network = format!("iggy-meilisearch-{}", Uuid::new_v4()) created per test, never removed. testcontainers drops container on MeilisearchContainer drop but not user-created
  networks. just clean-test-containers sweeps iggy-test-* containers, not networks. Linux Docker bridge limit ~32; repeated CI runs exhaust it. Fix: remove with_network(unique_network) entirely — single-container
  test gains nothing from a custom network; testcontainers creates an isolated network automatically.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+            query.with_filter(filter);
+        }
+
+        let results = self
+            .retry_sdk_operation("search documents", || {
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
```
ryerraguntla, line 356: 239-252 (interaction with runtime) — ALC gap: search_documents advances in-memory state.last_primary_key at line 247 before runtime persists state. If the runtime's Iggy producer fails to send
  the batch, it does not save ConnectorState. Next poll() reads the already-advanced in-memory cursor → documents in the failed batch permanently skipped, breaking at-least-once guarantee. Source trait provides
  no send-success callback for cursor revert. **Fix: document as known limitation. Deeper fix requires architectural Source trait change (acknowledgment callback).**

core/connectors/sources/meilisearch_source/README.md
```diff
@@ -0,0 +1,46 @@
...
+- `max_retry_delay`: Maximum retry delay. Defaults to `5s`.
+- `max_open_retries`: Maximum transient retry attempts during `open()`. Defaults to `5`.
+
+## Behavior
+
+The connector requires the source index to define a primary key. Each poll sends
+a `/search` request sorted by that primary key and stores the last emitted
+primary-key value in connector state. This avoids offset pagination skips when
+documents are inserted or deleted between polls.
+
```
ryerraguntla, line 48: 29 — States primary key "must be numeric, filterable, and sortable in Meilisearch" but gives no instruction on HOW. Configuring filterableAttributes and sortableAttributes requires a PATCH
  to /indexes/{uid}/settings (fixture does this at container.rs:147-165 but README omits it). User pointing connector at existing index will get silent poll-time search errors. **Fix: add example PATCH call or link
  to Meilisearch docs.**

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+const DEFAULT_MAX_OPEN_RETRIES: u32 = 5;
+const PRIMARY_KEY_SORT_DIRECTION: &str = "asc";
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct MeilisearchSourceConfig {
+    pub url: String,
+    pub index: String,
+    #[serde(serialize_with = "iggy_common::serde_secret::serialize_optional_secret")]
+    pub api_key: Option<SecretString>,
+    pub query: Option<String>,
```
ryerraguntla, line 59: 58 — filter: Option<Value> accepts any JSON type; type errors (e.g. filter: {"foo": "bar"}) pass From<> conversion and first open(), then surface as Error::InvalidConfigValue on first poll.
  **Fix: validate filter shape in open() rather than per-poll in filter_expression(). (raised: ecosystem; one agent, but verifiable — filter_expression() at line 255 is called every search_documents with no earlier  gate)**

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+    async fn open(&mut self) -> Result<(), Error> {
+        let sanitized_url = sanitize_url_for_logging(&self.config.url);
+        info!(
+            "Opening Meilisearch source connector with ID: {} for URL: {}, index: {}",
+            self.id, sanitized_url, self.config.index
+        );
+
+        let client = self.create_client()?;
+        self.check_connectivity(&client).await?;
+        let primary_key = self.get_primary_key(&client).await?;
```
ryerraguntla, line 510: [nit] line 400- Info is better than warn,

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+            let state = self.state.lock().await;
+            state.last_primary_key.clone()
+        };
+        let primary_key = self.primary_key.as_deref().ok_or_else(|| {
+            Error::Connection("Meilisearch primary key is not initialized".to_string())
+        })?;
+        let filter_expression = self.filter_expression()?;
+        let cursor_filter = cursor_filter_expression(primary_key, last_primary_key.as_ref())?;
+        let combined_filter = combine_filter_expressions(filter_expression, cursor_filter);
+        let sort = format!("{primary_key}:{PRIMARY_KEY_SORT_DIRECTION}");
```
ryerraguntla: - [nit] 219 — format!("{primary_key}:asc") allocates String every poll; primary_key is immutable after open(). Fix: build once in open(), store in MeilisearchSource.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        let mut state = self.state.lock().await;
+        if let Some(primary_key) = last_document_primary_key {
+            state.last_primary_key = Some(primary_key);
+        }
+        state.documents_produced += messages.len();
+        state.poll_count += 1;
+
+        Ok(messages)
+    }
+
```
ryerraguntla, line 374: [nit] 255-267 — filter_expression() recomputes filter String (including recursive join) every poll. Config immutable after construction. Fix: compute in open(), cache as Option<String> in  ResolvedMeilisearchSourceConfig.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,738 @@
...
+        Ok(())
+    }
+
+    async fn poll(&self) -> Result<ProducedMessages, Error> {
+        sleep(self.config.polling_interval).await;
+        let client = self
+            .client
+            .as_ref()
+            .ok_or_else(|| Error::Connection("Meilisearch client not initialized".to_string()))?;
+        let messages = self.search_documents(client).await?;
```
ryerraguntla, line 536: - [nit] 421-424 — serialize_state (msgpack) runs every poll including empty polls. **Fix: skip if cursor unchanged (last_document_primary_key.is_none() && messages.is_empty()).**

core/integration/tests/connectors/fixtures/meilisearch/container.rs
```diff
@@ -0,0 +1,332 @@
...
+    {
+        async move {
+            let url = format!(
+                "{}/indexes/{}/documents",
+                self.container().base_url,
+                index_name
+            );
+            let response = self
+                .http_client()
+                .get(&url)
```
ryerraguntla, line 234: [nit] 234 — list_documents hardcodes limit=100; tests inserting > 100 docs would see wait_for_documents complete early and the connector might emit fewer messages than expected.

ryerraguntla:
@countradooku - Looks very good. There are no blockers/criticals , only warnings and nits. After this fix, it should be good. @hubcio - After these are taken care, could one of the committers review and merge

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+            query.with_filter(filter);
+        }
+
+        let results = self
+            .retry_sdk_operation("search documents", || {
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
```
ryerraguntla, line 356: document_primary_key called only on documents.last(); middle docs with missing/non-integer PK are serialized and emitted; cursor advances past them; those docs are permanently excluded from future polls (id > last_pk filter never reaches them). Silent, unrecoverable data loss.
  Fix: validate every document's PK before calling documents_to_messages, not after. One approach: documents.iter().map(|d| document_primary_key(d, primary_key)).collect::<Result<Vec<_>, _>>()? before building messages

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+
+        self.client = None;
+        info!(
+            "Meilisearch source connector with ID: {} is closed.",
+            self.id
+        );
+        Ok(())
+    }
+}
+
```
ryerraguntla, line 566: normalize_host strips trailing slashes via string loop only; path components preserved; http://host:7700/v1 passes through unchanged; Meilisearch SDK prepends this as base, making all API routes double-prefixed (/v1/indexes/name/search becomes /v1/v1/indexes/name/search). Breaks any reverse-proxy deployment where URL has a path. Fix: parse with Url::parse, extract scheme + authority only, reject non-empty path components.

core/connectors/sources/meilisearch_source/config.toml
```diff
@@ -0,0 +1,45 @@
...
+topic = "test_topic"
+schema = "json"
+batch_length = 100
+linger_time = "5ms"
+
+[plugin_config]
+url = "http://localhost:7700"
+index = "iggy_messages"
+query = ""
+batch_size = 100
```
ryerraguntla, line 40: polling_interval = "100ms" contradicts DEFAULT_POLLING_INTERVAL = "5s" (lib.rs:43) and README "Defaults to 5s". Operators copy-pasting this example will hit Meilisearch at 50x the 
  intended rate.
  Fix: change to "5s".

core/integration/tests/connectors/meilisearch/meilisearch_source.rs
```diff
@@ -0,0 +1,211 @@
...
+const TEST_MESSAGE_COUNT: usize = 2;
+const POLL_ATTEMPTS: usize = 100;
+const POLL_INTERVAL_MS: u64 = 50;
+const SOURCE_INDEX: &str = "iggy_messages";
+
+#[iggy_harness(
+    server(connectors_runtime(config_path = "tests/connectors/meilisearch/source.toml")),
+    seed = seeds::connector_stream
+)]
+async fn meilisearch_source_produces_index_documents(
```
ryerraguntla: Also on line 94, both integration test names violate given_X_when_Y_should_Z convention mandated by CLAUDE.md. meilisearch_source_produces_index_documents and 
  meilisearch_source_state_persists_across_connector_restart have no given_ prefix or should_ assertion clause.
  Fix: rename per convention.

core/integration/tests/connectors/meilisearch/meilisearch_source.rs
```diff
@@ -0,0 +1,211 @@
...
+        duplicate_poll.messages.is_empty(),
+        "second poll should not receive duplicate Meilisearch documents"
+    );
+}
+
+#[iggy_harness(
+    server(connectors_runtime(config_path = "tests/connectors/meilisearch/source.toml")),
+    seed = seeds::connector_stream
+)]
+async fn meilisearch_source_state_persists_across_connector_restart(
```
ryerraguntla: integration test names violate given_X_when_Y_should_Z convention mandated by CLAUDE.md. meilisearch_source_produces_index_documents and 
  meilisearch_source_state_persists_across_connector_restart have no given_ prefix or should_ assertion clause.
  Fix: rename per convention.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+        Ok(())
+    }
+
+    async fn poll(&self) -> Result<ProducedMessages, Error> {
+        sleep(self.config.polling_interval).await;
+        let client = self
+            .client
+            .as_ref()
+            .ok_or_else(|| Error::Connection("Meilisearch client not initialized".to_string()))?;
+        let (messages, state_changed) = self.search_documents(client).await?;
```
ryerraguntla: lib.rs:607-611 + lib.rs:418 — bad-PK doc in mid-batch permanently stalls connector  document_primary_keys short-circuits on first Err(InvalidConfigValue). poll() returns that error. Runtime stops connector (permanent error type). Cursor unchanged. Restart re-fetches same batch, same error. Stuck loop with no recovery. **Fix:** skip bad-PK docs with a warn log and advance cursor to last good integer PK seen, OR return good prefix up to bad doc.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+    let value = document.get(primary_key).ok_or_else(|| {
+        Error::InvalidConfigValue(format!(
+            "Meilisearch document is missing primary key '{primary_key}'"
+        ))
+    })?;
+    primary_key_filter_literal(value)?;
+    Ok(value.clone())
+}
+
+fn document_primary_keys(documents: &[Value], primary_key: &str) -> Result<Vec<Value>, Error> {
```
ryerraguntla: lib.rs:607-611 + lib.rs:418 — bad-PK doc in mid-batch permanently stalls connector  document_primary_keys short-circuits on first Err(InvalidConfigValue). poll() returns that error. Runtime stops connector (permanent error type). Cursor unchanged. Restart re-fetches same batch, same error.  Stuck loop with no recovery. Fix: skip bad-PK docs with a warn log and advance cursor to last good integer PK seen, OR return good prefix up to bad doc.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+                    sleep(delay).await;
+                }
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl Source for MeilisearchSource {
+    async fn open(&mut self) -> Result<(), Error> {
```
ryerraguntla, line 500: sortableAttributes not validated at open()  Connector relies on primary_key:asc sort for cursor monotonicity. Meilisearch silently ignores sort on a field not in sortableAttributes. Result: hits returned in arbitrary order, cursor advances to arbitrary  last-in-batch PK, skips documents forever with no error signal. **Fix:** call GET /indexes/{uid}/settings/sortable-attributes at open() time; fail with InvalidConfigValue if primary key field absent.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+            .api_key
+            .as_ref()
+            .map(|key| key.expose_secret().to_string());
+
+        Client::new(host, api_key).map_err(|error| {
+            Error::Connection(format!("Failed to create Meilisearch client: {error}"))
+        })
+    }
+
+    async fn check_connectivity(&self, client: &Client) -> Result<(), Error> {
```
ryerraguntla, line 194: check_connectivity fails immediately on non-"available" health  retry_sdk_open_operation only retries transient SDK errors. If Meilisearch returns 200 OK with {"status":"unknown"} (during startup), the retry loop never fires; check_connectivity returns Err(Connection(...))  immediately. open() fails on a transient condition. Meilisearch can transiently return non-available status during index loading. **Fix:** add inner retry loop around the status check up to max_open_retries,  mirroring the sink pattern.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+            .retry_sdk_operation("search documents", || {
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
+        let primary_keys = document_primary_keys(&documents, primary_key)?;
+        let last_document_primary_key = primary_keys.into_iter().last();
+        let messages = self.documents_to_messages(documents)?;
+        let state_changed = last_document_primary_key.is_some();
```
ryerraguntla: **252 + lib.rs:258-259 + lib.rs:419** - poll_count/documents_produced diverge from persisted state on empty polls
  Both fields incremented in-memory at :258-259 on every poll (including empty). state_changed=false on empty polls → persisted_state = None → runtime keeps prior state. On restart, poll_count resets to
  last-checkpointed value, not current. No functional correctness impact (metrics only) but persisted state and in-memory state diverge indefinitely. **Fix:** either always persist state (remove the if state_changed
  gate for these fields) or move poll_count/documents_produced to transient struct fields not serialized.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+            if meilisearch_error.error_type == MeilisearchErrorType::Internal {
+                Error::HttpRequestFailed(meilisearch_error.to_string())
+            } else if meilisearch_error.error_code == MeilisearchErrorCode::IndexNotFound {
+                Error::InvalidConfigValue(meilisearch_error.to_string())
+            } else {
+                Error::PermanentHttpError(meilisearch_error.to_string())
+            }
+        }
+        MeilisearchSdkError::MeilisearchCommunication(communication_error) => {
+            if communication_error.status_code == 429 || communication_error.status_code >= 500 {
```
ryerraguntla: lib.rs:626 + lib.rs:649 — status_code == 0 treated as permanent error
  is_transient_sdk_error returns false for MeilisearchCommunication with status_code == 0. map_sdk_error maps it to PermanentHttpError. Meilisearch SDK produces status_code == 0 on pre-response connection failures  (connection refused, DNS failure) — these are transient. **Fix:** add || communication_error.status_code == 0 at both :626 and :649.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+        }
+
+        let results = self
+            .retry_sdk_operation("search documents", || {
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
+        let primary_keys = document_primary_keys(&documents, primary_key)?;
```
ryerraguntla: lib.rs:249-250 — document_primary_keys allocates Vec<Value> of N cloned PKs; only last() used  document_primary_keys at :607-611 clones every PK value into a Vec; into_iter().last() discards all but the last. For batch_size=100, 99 allocations wasted per poll. ```
**Fix:**
  let last_document_primary_key = documents
      .iter()
      .map(|doc| document_primary_key(doc, primary_key))
      .try_fold(None::<Value>, |_, pk| pk.map(Some))?;
```

File: core/connectors/sources/meilisearch_source/src/lib.rs [hunk 31]
```diff
@@ -0,0 +1,898 @@
...
+        info!(
+            "Opening Meilisearch source connector with ID: {} for URL: {}, index: {}",
+            self.id, sanitized_url, self.config.index
+        );
+
+        let filter_expression = filter_expression(self.config.filter.as_ref())?;
+        let client = self.create_client()?;
+        self.check_connectivity(&client).await?;
+        let primary_key = self.get_primary_key(&client).await?;
+        info!(
```
ryerraguntla, line 514: lib.rs:396-398 — integer-only PK constraint too late to fail
  open() succeeds on string-PK indexes (PK name returned fine). InvalidConfigValue fires at first poll() when document values are seen as non-integers. Operator configured connector, open() returned OK, connector
  runs, stops permanently on first poll. No way to detect string PK at open() without fetching sample documents; document this limitation prominently (currently buried in README para 2 of Behavior).

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,898 @@
...
+const PRIMARY_KEY_SORT_DIRECTION: &str = "asc";
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct MeilisearchSourceConfig {
+    pub url: String,
+    pub index: String,
+    #[serde(serialize_with = "iggy_common::serde_secret::serialize_optional_secret")]
+    pub api_key: Option<SecretString>,
+    pub query: Option<String>,
+    pub filter: Option<Value>,
```
ryerraguntla, line 60: filter: Option<Value> config type undocumented for array syntax
  filter accepts string or nested array-of-strings/arrays. Array form only works with JSON config (plugin_config_format = "json") not TOML. README says "string or array" with zero example. Config.toml shows only
  string form. Operators won't know how to use array syntax. **Fix:** add array example to README + note it requires JSON config format.  


Additional Information:   
- lib.rs:59 — pub filter: Option<Value> — field declaration in public config struct
 - README.md:12 — `filter`: Optional Meilisearch filter expression or array. — single-line doc, no example of array syntax, no note that array form requires plugin_config_format = "json"
- config.toml:24 — plugin_config_format = "json" is set, but no filter field appears in the config example at all (lines 34-45)

  The gap: config.toml omits filter entirely, README:12 mentions array but gives no example, and nowhere is it stated that array form requires the JSON config format (which is only set at config.toml:24). String filter works in both TOML and JSON config; array filter requires JSON. That distinction is never documented.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1009 @@
...
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
+        let (documents, last_document_primary_key) =
+            valid_documents_and_last_primary_key(documents, primary_key, self.id);
+        let messages = self.documents_to_messages(documents)?;
+
+        let mut state = self.state.lock().await;
```
ryerraguntla, line 365: 298-301,657-677 — hits non-empty, all PK invalid → last_primary_key stays None → same page re-polled forever; valid docs behind stuck. Unit test :974-985 locks this. **Fix**: fail poll (or hard-error) when hits.len()>0 && valid.is_empty().

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1009 @@
...
+            .await?;
+
+        primary_key.ok_or_else(|| {
+            Error::InvalidConfigValue(format!(
+                "Meilisearch index '{}' must define a primary key for stable source polling",
+                self.config.index
+            ))
+        })
+    }
+
```
ryerraguntla, line 277: lib.rs:238-258 — open checks sortableAttributes only; README + fixture require filterable too. Cursor filter pk > N fails at poll with PermanentHttpError after first batch. **Fix**: mirror sortable check for filterableAttributes in open.

core/connectors/sources/meilisearch_source/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_meilisearch_source"
```
ryerraguntla, line 19: version 0.4.0; peers 0.4.1-edge.1. **Fix**: validate and align.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1009 @@
...
+            .retry_sdk_operation("search documents", || {
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
+        let (documents, last_document_primary_key) =
+            valid_documents_and_last_primary_key(documents, primary_key, self.id);
+        let messages = self.documents_to_messages(documents)?;
+
```
ryerraguntla, line 360: 298-305 + runtime/source.rs:464-517 — in-memory cursor advance before Iggy send; send fail ⇒ batch lost in live process. README:72-76 documents. **Fix**: #3635-class ack/rollback

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1009 @@
...
+            .config
+            .api_key
+            .as_ref()
+            .map(|key| key.expose_secret().to_string());
+
+        Client::new(host, api_key).map_err(|error| {
+            Error::Connection(format!("Failed to create Meilisearch client: {error}"))
+        })
+    }
+
```
ryerraguntla, line 193: 187-216 — health outer loop nests retry_sdk_open_operation(max_open_retries); budgets multiply. **Fix**: single open retry budget.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1009 @@
...
+    match (user_filter, cursor_filter) {
+        (Some(user_filter), Some(cursor_filter)) => {
+            Some(format!("({user_filter}) AND ({cursor_filter})"))
+        }
+        (Some(user_filter), None) => Some(user_filter.to_string()),
+        (None, Some(cursor_filter)) => Some(cursor_filter),
+        (None, None) => None,
+    }
+}
+
```
ryerraguntla, line 682: restored non-integer last_primary_key → every poll Err; no open validate. **Fix**: validate/wipe cursor at open.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1009 @@
...
+                        "document": document,
+                        "meilisearch": {
+                            "index": self.config.index,
+                            "primary_key": self.primary_key.as_deref(),
+                        }
+                    })
+                } else {
+                    document
+                };
+
```
ryerraguntla, line 394: ProducedMessage.id always None; integer PK known. ES same (:418). **Fix**: map PK→id optional.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1185 @@
...
+                let query = query.clone();
+                async move { query.execute::<Value>().await }
+            })
+            .await?;
+        let documents: Vec<Value> = results.hits.into_iter().map(|hit| hit.result).collect();
+        let (documents, last_document_primary_key) =
+            valid_documents_and_last_primary_key(documents, primary_key, self.id)?;
+        let messages = self.documents_to_messages(documents)?;
+
+        let mut state = self.state.lock().await;
```
ryerraguntla, line 365: lib.rs:361 — cursor advances in memory before delivery success, so failed runtime send skips unsent docs next poll. **Fix**: move cursor commit behind runtime send success, or add source ack/rollback hook.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1185 @@
...
+    sortable_attributes
+        .iter()
+        .any(|attribute| attribute == primary_key)
+}
+
+fn primary_key_is_filterable(
+    filterable_attributes: &[FilterableAttribute],
+    primary_key: &str,
+) -> bool {
+    filterable_attributes
```
ryerraguntla, line 743: lib.rs:741 — primary-key validation accepts filterable settings-pattern match without requiring comparison support, but cursor pagination uses id > N. **Fix**: require settings.features.filter.comparison when validating settings-based filterable attributes.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        let sort = self.primary_key_sort.as_deref().ok_or_else(|| {
+            Error::Connection("Meilisearch primary key sort is not initialized".to_string())
+        })?;
+        let cursor_filter = cursor_filter_expression(primary_key, last_primary_key.as_ref())?;
+        let combined_filter =
+            combine_filter_expressions(self.filter_expression.as_deref(), cursor_filter);
+        let sort_refs = [sort];
+        let index = client.index(&self.config.index);
+        let mut query = index.search();
+        query
```
hubcio, line 341: cursor pagination breaks whenever `query` is non-empty. meilisearch bucket-sorts by ranking rules and `sort` sits 5th, behind 4 relevancy rules, so hits come back relevance-bucketed and pk-ordered only within a bucket - but the cursor takes the last hit's pk as if it were the batch max. two failure modes: with any non-empty query the last hit can carry a lower pk than earlier hits, so already-sent documents get re-emitted on the next poll; and once the match set exceeds `batch_size`, the truncated low-relevance tail can hold pks below the new cursor - those documents are never fetched again, silently. empty query is safe (placeholder search skips the relevancy rules).

the gate is basically free: `check_primary_key_cursor_settings` already fetches the full `Settings`, and `Settings.ranking_rules` is on that struct - open() can reject a non-empty `query` unless `sort` precedes the relevancy rules, with zero extra requests. the README should steer people to `filter` instead (telling them to reorder rankingRules would wreck relevance for every other consumer of the index). nothing exercises this path today: the integration setup runs `query = ""` and the fixture never sets a query.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    last_primary_key: Option<&Value>,
+) -> Result<Option<String>, Error> {
+    last_primary_key
+        .map(|value| {
+            primary_key_filter_literal(value).map(|literal| format!("{primary_key} > {literal}"))
+        })
+        .transpose()
+}
+
+fn primary_key_filter_literal(value: &Value) -> Result<String, Error> {
```
hubcio, line 694: meilisearch stores all numeric values as f64 (their datatypes docs: ~15 significant digits, integers beyond 2^53 lose precision, strings recommended for large ids). the cursor literal here renders full u64 precision but the server compares rounded values - around 2^63 roughly 2048 consecutive ids collapse onto one f64, so `id > X` can skip a whole block of documents after emitting just one of them. the unit tests with `9007199254740993` (2^53+1, the first unrepresentable integer) only round-trip this crate's own string rendering and never touch meilisearch, so they don't cover this. README says "must be an integer" with no range bound - either document the 2^53 limit or reject larger values.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+                    document,
+                });
+            }
+            Err(error) => warn!(
+                "Skipping Meilisearch document for source connector with ID: {connector_id}. Invalid primary key '{primary_key}': {error}"
+            ),
+        }
+    }
+
+    if documents_count > 0 && valid_documents.is_empty() {
```
hubcio, line 779: when a non-empty batch has zero valid pks this returns `Err`, and poll errors never reach the runtime - the SDK loop logs and continues, and connector status stays `Running` (set once before the loop; `set_error` is only reachable from send/state-save failures, and no error metric moves either). so a string-pk index passes open() - which checks sortable/filterable but never value types, and the README's own settings example is enough to get there - then errors every poll forever while reporting healthy. it also contradicts README lines 54-56, which say invalid documents "are skipped with a warning". `valid_documents_should_reject_all_invalid_primary_keys` pins the current behavior, so this is deliberate - but aligning with the README (skip + warn) plus a pk value-type probe in open() would avoid the permanent quiet stall. probe alone isn't sufficient (empty index has nothing to sample, and mixed pk types can appear later), so surfacing poll errors to connector status is the real long-term fix.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        mut operation_fn: Op,
+    ) -> Result<T, Error>
+    where
+        Op: FnMut() -> Fut,
+        Fut: Future<Output = Result<T, MeilisearchSdkError>>,
+    {
+        let mut retries = 0u32;
+
+        loop {
+            let result = tokio::time::timeout(self.config.timeout, operation_fn()).await;
```
hubcio, line 456: `timeout` here bounds a single attempt and the backoff sleeps sit outside any budget - the sink applies the same config key as a total budget across all retries (its retry loop tracks elapsed time and clamps each sleep to the remainder). same knob name, opposite meaning within one connector pair. worst case open() chains three of these loops (health, primary key, settings) to nearly 10 minutes against a peer that accepts TCP and never responds - meilisearch-sdk sets no reqwest timeout, so this wrapper is the only bound. and since sources initialize sequentially through a blocking FFI open, one unreachable meilisearch stalls the entire connectors-runtime boot: other sources, sinks, and the control API. adopting the sink's total-budget shape (or renaming the key) fixes both.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+            )));
+        }
+
+        Err(Error::InvalidConfigValue(format!(
+            "Meilisearch index '{}' must configure primary key '{}' in sortableAttributes for stable source polling",
+            self.config.index, primary_key
+        )))
+    }
+
+    async fn validate_restored_cursor(&self, primary_key: &str) {
```
hubcio, line 309: a restored cursor that fails validation gets wiped to `None` and the next poll runs unfiltered - the entire index is re-emitted. that's fail-open in the one spot where everything else in open() fails closed (missing pk, not sortable, not filterable all abort). influxdb_source handles the same case by refusing to start ("Refusing to start to prevent silent cursor reset"), and failing open() is not a dead end for the operator - `POST /sources/{key}/restart` re-runs it once the state is fixed. also the warn below prints the primary key field name, not the rejected value - the one thing needed to recover by hand.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+            timeout,
+            max_retries,
+            retry_delay,
+            max_retry_delay,
+            max_open_retries,
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
```
hubcio, line 137: `State` is persisted as a positional msgpack array (`rmp_serde::to_vec`), so appending any field later makes old state fail to decode - deserialize logs a warning, returns `None`, and the connector starts with a fresh cursor = full index replay. `#[serde(default)]` on the struct makes short arrays decode with defaults, which covers upgrades. rollback (old binary reading new state) still fails - rmp-serde errors on excess elements - which is one more reason to fail closed on an unusable cursor instead of replaying. and since the two trailing fields are both `usize`, a reorder would decode cleanly and silently swap the counters. worth a small decode test feeding a 2-element array into the 3-field struct.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        // The current source FFI has no post-send acknowledgment, so the
+        // plugin must advance its in-memory cursor when returning a batch to
+        // make steady-state polling progress. The runtime persists this same
+        // state only after it successfully sends the returned messages.
+        let mut state = self.state.lock().await;
+        if let Some(primary_key) = last_document_primary_key {
+            state.last_primary_key = Some(primary_key);
+        }
+        state.documents_produced += messages.len();
+        state.poll_count += 1;
```
hubcio, line 370: `poll_count` changes on every poll, so the returned state always differs and the runtime persists it even for empty batches (empty send returns Ok, and each save is write + fdatasync + rename + parent-dir fsync) - 2 fsyncs per polling interval on a fully idle index, and the integration fixture polls every 25ms. returning `state: None` when the cursor didn't advance skips all of it; the only fields lost are `poll_count`/`documents_produced`, which are log-only. elasticsearch/postgres sources share the pattern, but here the fix is two lines in a new file.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    }
+}
+
+fn filter_expression(filter: Option<&Value>) -> Result<Option<String>, Error> {
+    let Some(filter) = filter.filter(|value| !value.is_null()) else {
+        return Ok(None);
+    };
+
+    match filter {
+        Value::String(filter) if !filter.is_empty() => Ok(Some(filter.clone())),
```
hubcio, line 661: `filter = ""` falls through the `!filter.is_empty()` guard into the catch-all and open() dies with "Meilisearch filter must be a string or an array of strings/arrays" - about a value that is a string. the same guard inside arrays (line 633) makes `filter = [""]` fail the same way. meanwhile `filter = []` and `filter = [[]]` are silently accepted as no filter at all, which streams the entire index when a templating layer renders an empty list. four degenerate inputs, four behaviors, two of them silent - pick one semantics for all of them (empty means no filter, or empty is rejected), and add a test either way.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+const DEFAULT_POLLING_INTERVAL: &str = "5s";
+const DEFAULT_INCLUDE_METADATA: bool = false;
+const DEFAULT_TIMEOUT: &str = "30s";
+const DEFAULT_RETRY_DELAY: &str = "500ms";
+const DEFAULT_MAX_RETRY_DELAY: &str = "5s";
+const DEFAULT_MAX_RETRIES: u32 = 3;
+const DEFAULT_MAX_OPEN_RETRIES: u32 = 5;
+const PRIMARY_KEY_SORT_DIRECTION: &str = "asc";
+
+#[derive(Debug, Serialize, Deserialize)]
```
hubcio, line 53: config accepts unknown keys silently - a typo like `batch_sizee = 5000` runs at the default 100 with no warning anywhere (the runtime holds plugin config as an opaque value, and the env-var unknown-key warning whitelists the whole `PLUGIN_CONFIG_` namespace). influxdb sink and source both use `#[serde(deny_unknown_fields)]` and test it - worth adding here.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    fn serialize_state(&self, state: &State) -> Option<ConnectorState> {
+        ConnectorState::serialize(state, CONNECTOR_NAME, self.id)
+    }
+
+    fn create_client(&self) -> Result<Client, Error> {
+        let host = normalize_host(&self.config.url)?;
+        let api_key = self
+            .config
+            .api_key
+            .as_ref()
```
hubcio, line 186: two things around client creation: an `api_key` paired with a scheme-less `url` silently gets `http://` prepended (normalize_host) and ships the bearer token in plaintext - the sink warns about exactly this (`warn_if_api_key_uses_insecure_http`, loopback-exempt), worth porting. and meilisearch-sdk builds the auth header with `HeaderValue::from_str(...).unwrap()`, so a key containing any header-illegal byte - a trailing newline from `$(cat key)` or a k8s secret is the classic - panics inside `iggy_source_open`, and unwinding out of an `extern "C"` fn aborts the whole connectors runtime, taking every other connector with it. validating the key with `HeaderValue::from_str` in open() and returning a config error turns that into a clean failure. the sink shares the panic path, this just adds a second instance.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    let value = document.get(primary_key).ok_or_else(|| {
+        Error::InvalidConfigValue(format!(
+            "Meilisearch document is missing primary key '{primary_key}'"
+        ))
+    })?;
+    primary_key_filter_literal(value)?;
+    Ok(value)
+}
+
+fn primary_key_to_message_id(value: &Value) -> Option<u128> {
```
hubcio, line 723: `primary_key_filter_literal` accepts any i64, so negative pks are valid cursors - but this returns `None` for them, and pk 0 maps to message id 0, which the server replaces with a random uuid. so pk-as-message-id idempotency quietly stops working exactly for zero and negative pks: under message deduplication, the documented at-least-once restart path duplicates those documents while deduping everything else. reject `<= 0` in `primary_key_filter_literal` (consistent contract), or offset-map i64 into u128 so every valid cursor also yields a stable id.

side note: the whole helper reduces to `value.as_u64().map(u128::from)` - the `or_else` arm can never yield `Some`, negatives fail the `>= 0` filter and floats fail both conversions.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    async fn check_primary_key_cursor_settings(
+        &self,
+        client: &Client,
+        primary_key: &str,
+    ) -> Result<(), Error> {
+        let settings = self
+            .retry_sdk_open_operation("get index settings", || {
+                let index = client.index(&self.config.index);
+                async move { index.get_settings().await }
+            })
```
hubcio, line 287: the full `Settings` is already in hand here - one more comparison catches `batch_size` above the index's `pagination.maxTotalHits` (default 1000): the server silently caps `/search` results, nothing in the response signals the clamp, and the SDK doesn't clamp either. cheap boot-time check, plus a README line about the ceiling.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+            .api_key
+            .as_ref()
+            .map(|key| key.expose_secret().to_string());
+
+        Client::new(host, api_key).map_err(|error| {
+            Error::Connection(format!("Failed to create Meilisearch client: {error}"))
+        })
+    }
+
+    async fn check_connectivity(&self, client: &Client) -> Result<(), Error> {
```
hubcio, line 194: this hand-rolls the same 3-arm timeout/transient/backoff loop as `retry_sdk_operation_with_attempts` below, and the `Ok(Ok(health))` with non-"available" status arm can't fire in practice - `Client::health()` requires HTTP 200 with a parseable `{"status": ...}` body, and the server only serves 200 with "available". `self.retry_sdk_open_operation("health check", || async { client.health().await })` plus an explicit status check after the loop drops ~55 lines with byte-identical log messages. one caveat that deserves a comment in the code: the runtime doesn't retry a failed open, so moving the status check after the retry loop turns a hypothetical non-available status from retried into fail-fast.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+                    .map_err(|error| {
+                        Error::Serialization(format!(
+                            "Failed to serialize Meilisearch document: {error}"
+                        ))
+                    })
+            })
+            .collect()
+    }
+
+    async fn retry_sdk_operation<T, Fut, Op>(
```
hubcio, line 413: `retry_sdk_operation` has exactly one caller - inlining `retry_sdk_operation_with_attempts(operation, self.config.max_retries, ...)` there drops the wrapper. the open-budget twin is worth keeping for its name.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        }
+    }
+
+    if documents_count > 0 && valid_documents.is_empty() {
+        return Err(Error::InvalidConfigValue(format!(
+            "Meilisearch source connector with ID: {connector_id} received {documents_count} documents from index, but none had a valid integer primary key '{primary_key}'"
+        )));
+    }
+
+    let last_primary_key = valid_documents
```
hubcio, line 785: this re-runs `document_primary_key` on a document that already passed it in the loop above, and `?`-propagates an error that cannot happen. `valid_documents.last().and_then(|d| d.document.get(primary_key).cloned())` does the same without the impossible error path or the second validation.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    pub retry_delay: Option<String>,
+    pub max_retry_delay: Option<String>,
+    pub max_open_retries: Option<u32>,
+}
+
+#[derive(Debug)]
+pub struct MeilisearchSource {
+    id: u32,
+    config: ResolvedMeilisearchSourceConfig,
+    client: Option<Client>,
```
hubcio, line 75: `client`, `primary_key`, `primary_key_sort`, `filter_expression` are all none-before-open/some-after, guarded by three separate "not initialized" branches on the poll path. grouping them into one `Option<Opened { ... }>` gives a single guard, makes the half-initialized states unrepresentable, and lets close() clear everything atomically - today it only drops `client` and leaves the other three stale. `documents_to_messages` can take `primary_key: &str` so the metadata unit test stays client-free. one trap: keep the sort string in the struct rather than deriving it per poll - `with_sort` borrows, so a named binding has to outlive the query builder.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+        _ => Err(Error::InvalidConfigValue(
+            "Meilisearch source primary key values must be numbers for cursor pagination"
+                .to_string(),
+        )),
+    }
+}
+
+fn document_primary_key<'a>(document: &'a Value, primary_key: &str) -> Result<&'a Value, Error> {
+    let value = document.get(primary_key).ok_or_else(|| {
+        Error::InvalidConfigValue(format!(
```
hubcio, line 715: per-document data problems (missing or non-integer pk) surface as `InvalidConfigValue`, the same variant as genuine config errors - the SDK has `InvalidRecordValue` for exactly this and the sink uses it. as-is an operator can't tell "fix the TOML" from "one document has a string id". three unit tests assert the current variant and would need updating with the change.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    filterable_attributes
+        .iter()
+        .any(|attribute| match attribute {
+            FilterableAttribute::Attribute(attribute) => attribute == primary_key,
+            FilterableAttribute::Settings(settings) => {
+                settings.features.filter.comparison
+                    && settings
+                        .attribute_patterns
+                        .iter()
+                        .any(|pattern| pattern == "*" || pattern == primary_key)
```
hubcio, line 752: `attribute_patterns` supports leading/trailing globs server-side, but this only accepts `"*"` or an exact match - `["id*"]` covering `id` gets refused with a message telling the operator to configure what they already configured. worth matching globs here. two cautions: the plain-string arm granting comparison implicitly is correct (legacy rules default to full filter features), so don't add a comparison check there; and the settings-object form needs meilisearch v1.14+ while the test container pins v1.13, so this branch currently has unit coverage only.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    url.set_query(None);
+    url.set_fragment(None);
+    let mut normalized = url.to_string();
+    while normalized.ends_with('/') {
+        normalized.pop();
+    }
+    Ok(normalized)
+}
+
+fn sanitize_url_for_logging(host: &str) -> String {
```
hubcio, line 611: same job as the sink's `sanitize_url_for_log`, different name and different output for the same input (`<redacted>@host` here, userinfo stripped entirely there) - and both crates carry a test named `sanitize_url_should_redact_credentials_without_scheme` asserting different results. related: this source hard-rejects url paths while the sink warns and strips them, so the same `url` value works on one half of the pair and fails on the other, documented in neither README. one name and one behavior across the pair would spare the next maintainer.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+    fn documents_to_messages(
+        &self,
+        documents: Vec<ValidDocument>,
+    ) -> Result<Vec<ProducedMessage>, Error> {
+        documents
+            .into_iter()
+            .map(|valid_document| {
+                let document = valid_document.document;
+                let payload = if self.config.include_metadata {
+                    json!({
```
hubcio, line 384: `json!` routes the owned `document` through `serde_json::to_value` (full tree copy, `Value`'s serialize has no identity shortcut) and then `to_vec` serializes the result again. building the envelope with `serde_json::Map::new()` and moving `document` in - or a borrowing `#[derive(Serialize)]` struct straight to bytes - is less code and skips the copy. (`Map::with_capacity` silently discards its argument under default features, so `new()` is the honest spelling.) similar shape in `document_primary_key`: it calls `primary_key_filter_literal` just to validate and throws the rendered string away for every document; splitting the integer predicate from the literal renderer keeps both error messages, drops the per-document allocation, and removes the `Serialization` arm that can't fail for integers.

core/connectors/sources/meilisearch_source/src/lib.rs
```diff
@@ -0,0 +1,1210 @@
...
+                    sleep(delay).await;
+                }
+            }
+        }
+    }
+
+    async fn get_primary_key(&self, client: &Client) -> Result<String, Error> {
+        let primary_key = self
+            .retry_sdk_open_operation("get primary key", || async {
+                let mut index = client.get_index(&self.config.index).await?;
```
hubcio, line 262: `get_index` already returns the index with `primary_key` as a public field - the follow-up `get_primary_key()` refetches the same resource, so every retry attempt pays two GETs. `client.get_index(&self.config.index).await?.primary_key` is one.

core/connectors/sources/meilisearch_source/README.md
```diff
@@ -0,0 +1,90 @@
...
+String filters work in regular TOML plugin config:
+
+```toml
+filter = "category = alpha"
+```
+
+Nested array filters require JSON plugin config because the connector receives
+the field as `serde_json::Value`:
+
+```json
```
hubcio, line 33: this is backwards on both ends: a `.json` connector config never loads (the local provider skips any file whose extension isn't `.toml`), and nested arrays work fine in TOML today - `plugin_config` is a `serde_json::Value` populated from the TOML and re-serialized to json before the plugin sees it, and mixed arrays are legal TOML. the field being `serde_json::Value` is precisely why the TOML form works. `plugin_config_format = "json"` in config.toml doesn't do what this implies either - it only picks the rendering of `GET /sources/{key}/config` responses. suggest replacing this section with the TOML form and a note that env-var overrides can't carry nested arrays (the env parser mangles them), plus a unit test with a nested-array filter so it's observed rather than assumed.

core/connectors/sources/meilisearch_source/README.md
```diff
@@ -0,0 +1,90 @@
...
+  -H 'Content-Type: application/json' \
+  --data '{
+    "filterableAttributes": ["id"],
+    "sortableAttributes": ["id"]
+  }'
+```
+
+Meilisearch state is advanced in memory when a batch is returned from `poll()`.
+The runtime persists that returned state only after it sends the batch to Iggy,
+but the source trait does not provide an acknowledgment callback that would let
```
hubcio, line 74: "the next in-process poll can skip the unsent batch; a restart reloads the last persisted cursor" undersells it. the in-memory cursor advances before the send, and the next successful send persists a cursor past the failed batch - so one failed send followed by any success makes the skip permanent. restart only helps if the process dies before another batch goes through. worth saying plainly: once a later batch commits, the unsent one is gone, and the skip is certain rather than "can".

core/connectors/sources/meilisearch_source/README.md
```diff
@@ -0,0 +1,90 @@
...
+them as JSON messages into Iggy.
+
+## Configuration
+
+- `url`: Meilisearch base URL.
+- `index`: Source index UID.
+- `api_key`: Optional Meilisearch API key sent as `Authorization: Bearer`.
+- `query`: Optional search query. Defaults to an empty query.
+- `filter`: Optional Meilisearch filter expression string or nested JSON array.
+- `batch_size`: Maximum documents fetched per poll. Defaults to `100`.
```
hubcio, line 13: a few gaps in this list: `batch_size` is silently clamped to >= 1 and capped server-side by `maxTotalHits` (default 1000); `retry_delay`/`max_retry_delay` get swapped (with a warn) when inverted; `polling_interval` sleeps before the first search, so the first batch lands one interval late; and the pk cursor only ever sees inserts - updates to already-emitted documents are never re-emitted, which makes "avoids offset pagination skips" read broader than the actual guarantee. the update blind spot is the one most worth a sentence.

core/connectors/sources/meilisearch_source/config.toml
```diff
@@ -0,0 +1,47 @@
...
+
+[plugin_config]
+url = "http://localhost:7700"
+index = "iggy_messages"
+query = ""
+# filter = "category = alpha"
+# Nested array filters require JSON connector configuration. See README.md.
+batch_size = 100
+polling_interval = "5s"
+include_metadata = true
```
hubcio, line 42: ships `include_metadata = true` while the code default is `false` and the README says false too - and the sink plus influxdb source both default true. three-way disagreement; whichever way it lands, code, example and README should agree (family parity says flip the code default to true).

core/integration/tests/connectors/fixtures/meilisearch/container.rs
```diff
@@ -43,6 +44,19 @@ pub struct MeilisearchDocumentsResponse {
...
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct MeilisearchTaskResponse {
+    task_uid: usize,
+}
+
+#[derive(Debug, Deserialize)]
+struct MeilisearchTaskStatus {
+    status: String,
+    #[allow(dead_code)]
```
hubcio, line 56: `#[allow(dead_code)]` isn't needed - `error` is read in the failed/canceled arm's format string below. the allow would hide a real dead-code warning later.

core/integration/tests/connectors/meilisearch/meilisearch_source.rs
```diff
@@ -0,0 +1,211 @@
...
+use iggy_common::{Consumer, Identifier, PollingStrategy};
+use integration::harness::seeds;
+use integration::iggy_harness;
+use std::time::Duration;
+use tokio::time::sleep;
+
+const TEST_MESSAGE_COUNT: usize = 2;
+const POLL_ATTEMPTS: usize = 100;
+const POLL_INTERVAL_MS: u64 = 50;
+const SOURCE_INDEX: &str = "iggy_messages";
```
hubcio, line 29: re-declares `"iggy_messages"` while `fixtures::TEST_INDEX` is exported and is what the fixture actually feeds the connector - the sink test imports it. if `TEST_INDEX` ever changes, this test indexes into a different index than the connector reads.

hubcio (changes requested):
two items that don't anchor to changed lines:

- setting a nested-array `filter` through env-var overrides gets silently mangled before it reaches the connector: the env array parser in `core/configs/src/configs_impl/parsing.rs` strips both leading brackets (`trim_start_matches('[')` trims repeatedly) and its splitter tracks quotes but not bracket depth, so `[["a","b"],"c"]` becomes three broken string fragments. they all pass the connector's filter validation and get ANDed into a syntactically invalid meilisearch filter: open() succeeds (it never runs a search), every poll gets a 400 with no retry, and the connector sits at `Running` forever. pre-existing parser bug, but this is the first connector whose docs point at nested arrays - deserves a separate issue on core/configs, and until then the README shouldn't recommend the env route for nested filters.

- coverage note: the integration test loads the shipped `config.toml` and overrides via env, so the riskiest configurations are exactly the untested ones - non-empty `query` (the cursor-ordering problem), `include_metadata = true` (fixture forces `"false"`), any `filter` at all, and the settings-object filterable branch (needs meilisearch v1.14, container pins v1.13).

PR #3499, opened 2026-06-16
feat(go): Add PollMessagesInto

### LINKED ISSUE #3452: feat(go): response body buffer allocates fresh on every poll
### Description

The Go client's TCP read path allocates a new `[]byte` for the response body on every RPC. For consumers polling at high rates, this is the largest remaining contributor to allocation churn on the read side.

A previous issue (#3441) addressed the request-path allocations and the 8-byte response status header. This issue covers what remains: the variable-sized response body, which the parsed `PolledMessage` references via slices in `Message.Payload` and `Message.UserHeaders`. Because of that aliasing, the buffer must outlive deserialization, so the transport layer can't recycle it on its own.

### Affected area / component

Go SDK

### Proposed solution

Add a new method `PollMessagesInto` that reads the response body into a caller-provided buffer, growing it as needed, and returns the parsed `PolledMessage` plus the (possibly reallocated) backing slice for reuse on the next call.

```go
func (c *IggyTcpClient) PollMessagesInto(
    ctx context.Context,
    streamId iggcon.Identifier,
    topicId iggcon.Identifier,
    consumer iggcon.Consumer,
    strategy iggcon.PollingStrategy,
    count uint32,
    autoCommit bool,
    partitionId *uint32,
    buf []byte,
) (*iggcon.PolledMessage, []byte, error)

Usage

var buf []byte
for {
    var polled *iggcon.PolledMessage
    polled, buf, err = client.PollMessagesInto(ctx, ..., buf)
    if err != nil { ... }
    // process polled.Messages — Payload/UserHeaders alias buf
}
// buf gets reused next iteration; zero allocations once cap is sufficient.
```

### Alternatives considered

_No response_

### Contribution

### Good first issue

matanper:

Closes #3452

## Rationale

`PollMessages` allocated a fresh `[]byte` for every response body. For consumers polling at high rates this is the largest remaining source of allocation churn on the read side.

## What changed?

Each `PollMessages` call passed through `do()` → `sendWireAndFetchResponse()` → `make([]byte, length)`, allocating a new buffer on every RPC regardless of message size.

`PollMessagesInto` lets the caller supply a reusable buffer that grows on demand and is returned for the next call. Once the buffer reaches steady-state capacity, the response-body read is zero-alloc. `PollMessages` delegates to `PollMessagesInto(nil)` so existing callers are unchanged. The same delegation pattern is applied down the stack: `do` → `doInto`, `sendWireAndFetchResponse` → `sendWireAndFetchResponseInto`, `sendLocked` → `sendLockedInto`.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Claude
2. Entire implementation
3. Ran it against iggy server
4. Yes


chengxilo:
The `readInto/doInto` may make sense but I don't think it's a good idea to provide a `PollMessagesInto`. I'm not a fan of this kind of change. If I'm not mistaken, it require the caller to manually manage the buffer and I don't think it actually worth it.

matanper:
As I understand, high performance is central to what Iggy offers. Why not give advanced users an opt-in way to maximize it?

chengxilo:
> As I understand, high performance is central to what Iggy offers. Why not give advanced users an opt-in way to maximize it?

I agree performance matters, but I don't think exposing buffer management to callers is the right way to achieve it. No major SDK I've seen asks users to manually pass buffers for RPC calls.

We should do it internally (e.g. a pool or a reusable buffer inside the client) so callers get the benefit transparently. The internal refactoring in this PR could still be useful tho.

matanper:
Sure, I'm open to any suggestions how to do it better.

IIUC an internal buffer approach doesn't work here because deserialization is zero-copy. 
So as long as the caller holds the returned *PolledMessage, the buffer is live. There's no hook to know when to reclaim it — no destructor, no finalizer worth trusting, no Release() call (which would impose more discipline on the caller than PollMessagesInto does, with less compile-time enforcement).

I can give an example of `fasthttp` which [does this](https://github.com/valyala/fasthttp/blob/ad93216ac1637ba455748b782df052f6e11010fb/client.go#L121-L129):

```go
// Get returns the status code and body of url.
//
// The contents of dst will be replaced by the body and returned, if the dst
// is too small a new slice will be allocated.
//
// The function follows redirects. Use Do* for manually handling redirects.
func Get(dst []byte, url string) (statusCode int, body []byte, err error) {
	return defaultClient.Get(dst, url)
}
```

chengxilo:
> Sure, I'm open to any suggestions how to do it better.
> 
> IIUC an internal buffer approach doesn't work here because deserialization is zero-copy. So as long as the caller holds the returned *PolledMessage, the buffer is live. There's no hook to know when to reclaim it — no destructor, no finalizer worth trusting, no Release() call (which would impose more discipline on the caller than PollMessagesInto does, with less compile-time enforcement).
> 
> I can give an example of `fasthttp` which [does this](https://github.com/valyala/fasthttp/blob/ad93216ac1637ba455748b782df052f6e11010fb/client.go#L121-L129):
> 
> ```go
> // Get returns the status code and body of url.
> //
> // The contents of dst will be replaced by the body and returned, if the dst
> // is too small a new slice will be allocated.
> //
> // The function follows redirects. Use Do* for manually handling redirects.
> func Get(dst []byte, url string) (statusCode int, body []byte, err error) {
> 	return defaultClient.Get(dst, url)
> }
> ```

Since this is zero-copy, if the caller modifies the payload in a polled message (or reuses the buffer), it silently corrupts messages returned by earlier calls. The only guard is a doc comment -- easy to miss, and nothing in the type system prevents it.

Beyond that, only PollMessages gets the buffer-reuse variant. If the goal is avoiding response allocations, we would hope the same argument applies to every API call. Offering it for one method makes the surface inconsistent. But if we offer the buffer-reuse variant to other methods, some of them need to pay attention to the zero-copy silent corruption while some doesn't, it need the user to understand everything behind the Go SDK.+

I benchmarked the new xxxxInto function, it's true that it improves the allocation, and in some cases it halved the cost, but compare to the net I/O, it's a very small improvement compare to what caller lose.

Also, there are many features that is supported by Rust but not supported by Go, we can pay more attention to fill the gap instead.

foreign/go/client/tcp/tcp_core.go
atharvalade, line 1: When the server returns an empty body (length <= 1), `sendLockedInto` returns `buf` unchanged -- still holding data from the previous call. `DeserializeFetchMessagesResponse` then re-parses that stale data and returns ghost messages. The old code returned `[]byte{}` here. I think you need `return buf[:0], nil` to preserve the reusable backing array without leaking old contents.

I reproduced this locally: first poll returns 1 message, second poll has an empty server response, but caller still sees 1 stale message from the first call.

foreign/go/contracts/client.go
```diff
@@ -114,6 +114,20 @@ type Client interface {
...
+		ctx context.Context,
+		streamId Identifier,
+		topicId Identifier,
+		consumer Consumer,
+		strategy PollingStrategy,
+		count uint32,
+		autoCommit bool,
+		partitionId *uint32,
+		buf []byte,
+	) (*PolledMessage, []byte, error)
```
atharvalade, line 129: Adding `PollMessagesInto` to the `Client` interface is a compile-breaking change for anyone implementing or mocking it (test doubles, alternative transports). Would it make sense to keep this as a concrete method on `IggyTcpClient` only, or put it on a separate optional interface like `BufferedPoller` that callers can type-assert to?

atharvalade (changes requested):

atharvalade:
I agree with @chengxilo here. I tried reproducing the zero-copy footgun and hit a concrete bug: if the server ever returns an empty body (length <= 1), the caller gets ghost messages from the previous poll because `sendLockedInto` hands back the stale buffer as-is. That's exactly the kind of silent corruption that a doc comment won't save you from. The internal refactoring (`doInto`/`sendLockedInto`) is solid work, but exposing the buffer to callers adds a foot-shotgun for marginal gain over net I/O cost.

matanper:
@chengxilo @atharvalade OK I understand your view 👍 

> The internal refactoring (doInto/sendLockedInto) is solid work

I'm not sure what the value of the internal refactoring without `PollMessagesInto`, it is the only caller of the new `Into` variants

chengxilo:
> @chengxilo @atharvalade OK I understand your view 👍
> 
> > The internal refactoring (doInto/sendLockedInto) is solid work
> 
> I'm not sure what the value of the internal refactoring without `PollMessagesInto`, it is the only caller of the new `Into` variants

Yeah, I believe so. We are not going to reuse the buffer for response anyway so all the xxxInto method won't be useful.

hubcio:
TBH i'm fine with the optimization idea proposed by @matanper, but i'd like to see a full-blown approach (with design approved by current golang folks: @chengxilo @atharvalade) to the problem, not just polled messages. we don't even have reuseable buffers in Rust SDK, so it would be interesting to see. there are quite a few challenges with the API itself, and even in Rust safety is hard. can't imagine that in go :)

chengxilo:
> TBH i'm fine with the optimization idea proposed by @matanper, but i'd like to see a full-blown approach (with design approved by current golang folks: @chengxilo @atharvalade) to the problem, not just polled messages. we don't even have reuseable buffers in Rust SDK, so it would be interesting to see. there are quite a few challenges with the API itself, and even in Rust safety is hard. can't imagine that in go :)

Yeah I believe if we can make it safe and don't introduce complexity, it would be an amazing improvement. Re-use the buffer is possible, but we have to remove the zero-copy logic in poll-message response deserialization method so caller will not be able to poision the byte slice we are re-using. Then we can provide a `sync.Pool` to protect the byte slice we re-use, make sure they won't be used by other RPC during the process of reading from TCP connect and deserialization.

I think with this approach, the `buf []bytes` won't be necessary, and we can apply this to all methods in the SDK. Theoratically, it won't cause regression to `PollMessages` since while we give up zero-copy, we avoid the allocation of a new bytes slice for the response from `PollMessages`. And for other methods it is pure improvement.

@matanper what do you think? I just come up with this idea like a few minutes ago, to be honest, I'm not sure if it actually works.

matanper:
As I see it the rest of the SDK functions (`GetStats`, `CreateStream`, `GetTopic`, etc.) are called occasionally and are not performance sensitive. 
So if we will pool the response buffer for all methods, but keep allocating for `PollMessages` I don't think the added complexity worth the gains

foreign/go/client/tcp/tcp_core.go
matanper: fixed, added regression test

chengxilo:
> As I see it the rest of the SDK functions (`GetStats`, `CreateStream`, `GetTopic`, etc.) are called occasionally and are not performance sensitive. So if we will pool the response buffer for all methods, but keep allocating for `PollMessages` I don't think the added complexity worth the gains

It's totally fine if we can handle the deserialization of all the response type in one single function. For example, we can implment the `do` method like this (It need us to implement the UnmashalBinary for each response type, which is actually planned):
```go
// do sends the command and returns the response body. Commands implementing
// the appender interface encode directly into a pooled buffer.
func (c *IggyTcpClient) do(ctx context.Context, cmd command.Command, resp encoding.BinaryUnmarshaler) error {
	bp := acquireRequestBuf()
	buf, err := encodeWireRequest(*bp, cmd)
	if err != nil {
		releaseRequestBuf(bp)
		return err
	}
	*bp = buf

	// we can use a re-usable buffer to receive it to improve performance.
	respBuf, err := c.sendWireAndFetchResponse(ctx, buf)
	releaseRequestBuf(bp)

	return resp.UnmarshalBinary(respBuf)
}
```

To support it we need to provide UnmarshalBinary function for requests. Example:

Update the `PollMessages` in `iggcon` package (I basically copied the logic from `binary_serialization`)
```go
type PolledMessage struct {
	// This field should be set before unmarshalling the message.
	// It indicates the compression algorithm used for the message payload.
	Compression IggyMessageCompression
	PartitionId uint32
	CurrentOffset      uint64
	MessageCount       uint32
	Messages           []IggyMessage
}

func (m *PolledMessage) UnmarshalBinary(data []byte) error {
	if len(data) == 0 {
		return nil
	}

	length := len(data)
	m.PartitionId  = binary.LittleEndian.Uint32(data[0:4])
	m.CurrentOffset = binary.LittleEndian.Uint64(data[4:12])
	m.MessageCount = binary.LittleEndian.Uint32(data[12:16])
	position := 16
	var messages = make([]IggyMessage, 0)
	for position < length {
		if position+MessageHeaderSize >= length {
			// body needs to be at least 1 byte
			break
		}
		header, err := MessageHeaderFromBytes(data[position : position+MessageHeaderSize])
		if err != nil {
			return err
		}
		position += MessageHeaderSize
		payloadEnd := position + int(header.PayloadLength)
		if payloadEnd > length {
			break
		}
		payloadSlice := data[position:payloadEnd]
		position = payloadEnd

		var userHeaders []byte = nil
		if header.UserHeaderLength > 0 {
			userHeaders = data[position : position+int(header.UserHeaderLength)]
		}
		position += int(header.UserHeaderLength)

		switch m.Compression {
		case MESSAGE_COMPRESSION_S2, MESSAGE_COMPRESSION_S2_BETTER, MESSAGE_COMPRESSION_S2_BEST:
			if length < 32 {
				break
			}
			payloadSlice, err = s2.Decode(nil, payloadSlice)
			if err != nil {
				return fmt.Errorf("failed to decode s2 payload: %w", err)
			}
		}

		m.Messages = append(messages, IggyMessage{
			Header:      *header,
			Payload:     payloadSlice,
			UserHeaders: userHeaders,
		})
	}

	// !TODO: Add message offset ordering
	return nil
} 
```

Caller just need to provide a pointer of the response type they want here. Example:

```go
func (c *IggyTcpClient) PollMessages(
	ctx context.Context,
	streamId iggcon.Identifier,
	topicId iggcon.Identifier,
	consumer iggcon.Consumer,
	strategy iggcon.PollingStrategy,
	count uint32,
	autoCommit bool,
	partitionId *uint32,
) (*iggcon.PolledMessage, error) {
	resp := &iggcon.PolledMessage{
		// Provide the compression to the deserializer so that it can decompress the messages if needed
		MessageCompression: c.MessageCompression,
	}
	if err := c.do(ctx, &command.PollMessages{
		StreamId:    streamId,
		TopicId:     topicId,
		Consumer:    consumer,
		AutoCommit:  autoCommit,
		Strategy:    strategy,
		Count:       count,
		PartitionId: partitionId,
	}, resp); err != nil {
		return nil, err
	}

	return resp, nil
}
```

I write these examples manually, so it's error-prone. But they should be enough to describe my idea.

@matanper What is your opinion to this approach. It will, instead of introduce complexity, make the code more clear since the deserialization functions for each different type will be replaced by standarized `encoder.Unmarshaller` interface.

It may need an extra PR to implement the Marshaller and get rid of `binary_deserialization` directory, since this is a huge change.

matanper:
@chengxilo I like this design, removing `binary_serialization` seems like the right call to make the code cleaner

It doesn't solve my original issue of allocation churn in `PollMessages`, but since we don't have agreed contract for that we can drop that for now

chengxilo:
> @chengxilo I like this design, removing `binary_serialization` seems like the right call to make the code cleaner
> 
> It doesn't solve my original issue of allocation churn in `PollMessages`, but since we don't have agreed contract for that we can drop that for now

Yeah unfortunately that's true. If we want to re-use the buffer we use to get response while maintain zero-copy, we have to come up with an idea to let caller inform the client to release the buffer, which is not easy to design.

PR #3500, opened 2026-06-16
feat(php): wire release pipeline

### LINKED ISSUE #3304: PHP SDK: wire release pipeline

The PHP SDK currently has pre-merge coverage, but it is not wired into the release/post-merge publishing path. Track the release pipeline work in one place before enabling PHP SDK publication.

## Missing pieces

- Add `sdk-php` to `.github/config/publish.yml` / `.github/workflows/publish.yml` so the publish workflow can select and run PHP SDK publishing.
- Add a PECL `package.xml` if PECL distribution is part of the PHP SDK release path.
- Add `.github/actions/php/post-merge/action.yml` for PHP SDK post-merge publishing/tagging behavior.
- Include `sdk-php` in the SDK loop in `.github/workflows/post-merge.yml` around the existing SDK auto-publish detection block near line 100.

## Notes

A combined ticket is intentional: these pieces should be designed together so PHP release behavior matches the other SDKs and does not get forgotten after the initial experimental cut.

Related: #3302
Initial PHP SDK work: #3235

countradooku:
## Summary

Closes #3304.

- add `sdk-php` release metadata and PHP SDK metadata checks
- add a reusable PHP release workflow that builds a source-only PHP SDK archive plus generated binary license manifest
- publish PHP SDK source release files as monorepo GitHub release assets behind the `php-sdk-*` tag
- keep Packagist out of this monorepo pipeline until a split repository or root `composer.json` package source exists
- keep the PHP `iggy` Rust SDK dependency version synchronized with `core/sdk`, matching the Python binding pattern
- include PHP in the post-merge SDK auto-publish loop without blocking Docker or other SDK publishing on PHP source package builds

## Notes

PECL is intentionally skipped. This PR now avoids prebuilt PHP extension binaries because PIE resolves through Packagist, Packagist is not wired for the monorepo yet, and Linux/macOS PHP extensions are conventionally source-built. The source archive and generated license manifest stay in place so the release path can be exercised now; prebuilt binaries can return later once the split repo plus Packagist/PIE flow can consume them directly.

The PHP SDK package version is currently stable `0.1.0`, so the post-merge auto-publish loop will skip it until `sdk-php` is bumped to a prerelease such as `0.1.1-edge.1`.

This also broadens `extract-version.sh --is-pre-release` from the previous `edge`/`rc` match to anchored SemVer prerelease suffixes so PHP `-dev`, `-edge`, and `-rc` versions use the same post-merge auto-publish gate as the other SDKs without treating build metadata as prerelease.

## Validation

- `bash -n scripts/bump-version.sh scripts/ci/php-sdk-version-sync.sh scripts/extract-version.sh`
- `shellcheck scripts/bump-version.sh scripts/ci/php-sdk-version-sync.sh scripts/extract-version.sh`
- YAML parse for edited workflows/actions/config via Ruby `YAML.load_file`
- `actionlint -shellcheck= .github/workflows/_build_php_extensions.yml .github/workflows/publish.yml .github/workflows/_common.yml`
- `python3 -m json.tool foreign/php/composer.json`
- `./scripts/ci/php-sdk-version-sync.sh --check`
- `./scripts/ci/php-sdk-version-sync.sh --fix`
- `PATH="/tmp/iggy-yq-v4.47.1:$PATH" ./scripts/extract-version.sh --check`
- staged-tree `git archive` check for the PHP source package contents
- `cargo metadata --manifest-path foreign/php/Cargo.toml --no-deps --format-version=1`


.github/workflows/_build_php_extensions.yml
atharvalade, line 1: [NIT] Every other workflow that calls `extract-version.sh` installs `yq` with a fallback (`_common.yml:81`, `post-merge.yml:55`, `publish.yml:300`). This one doesn't. Works today because runners have it pre-installed, but it's the odd one out if that ever changes.

scripts/extract-version.sh
```diff
@@ -406,12 +422,11 @@ fi
...
 #
 # Matches (any of):
-#   -edge[.N]   (rust crates, docker, node SDK)
-#   -rc[.N]     (all SDKs)
+#   -<semver pre-release> (rust crates, docker, node/PHP SDKs)
 #   .devN       (Python SDK PEP 440 development markers)
 #   rcN$        (legacy bare rcN, retained for compatibility)
 if [[ "$RETURN_IS_PRE_RELEASE" == "true" ]]; then
-    if [[ "$VERSION" =~ -(edge|rc) ]] \
+    if [[ "$VERSION" =~ -[0-9A-Za-z] ]] \
```
atharvalade: [NIT] This broadens the pre-release check from `-(edge|rc)` to `-[0-9A-Za-z]` for all SDKs. No current version is affected (Java SNAPSHOT is caught earlier in post-merge.yml), but might be worth a one-liner in the PR description since it's a cross-cutting change in a PHP-scoped PR.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,227 @@
...
+          cat "$response"
+          exit 1
+        fi
+
+        cat "$response"
+        echo ""
+        echo "Packagist update requested for apache/iggy-php"
+      shell: bash
+      env:
+        PACKAGIST_TOKEN: ${{ env.PACKAGIST_TOKEN }}
```
atharvalade: [NIT] Token comes from `${{ env.PACKAGIST_TOKEN }}` instead of being a declared input. Works fine since the script checks for empty and fails clearly, but adding it as an explicit input would make the contract obvious if someone reuses this action.

atharvalade (changes requested):
overall lgtm just some nits

hubcio (changes requested):
`scripts/bump-version.sh` is not updated

atharvalade (dismissed):

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,231 @@
...
+
+        response="$(mktemp)"
+        http_code="$(curl -sSL \
+          -o "$response" \
+          -w '%{http_code}' \
+          -X POST \
+          -H 'Content-Type: application/json' \
+          -H "Authorization: Bearer ${PACKAGIST_TOKEN}" \
+          'https://packagist.org/api/update-package' \
+          -d '{"repository":"https://github.com/apache/iggy"}')"
```
hubcio: packagist only indexes a composer.json at the repo ROOT, but apache/iggy has none - it lives at `foreign/php/composer.json`. this POST also targets the monorepo `https://github.com/apache/iggy`, and the update-package api only refreshes already-registered packages, it can't create one. so `apache/iggy-php` is never discovered or served, and the later `Wait for Packagist availability` + `Verify Packagist version` steps time out and exit 1 - after the immutable `php-sdk-<ver>` tag and the public github release are already pushed. needs a split-repo mirror with a root composer.json registered on packagist (or private packagist).

.github/config/publish.yml
```diff
@@ -123,6 +123,12 @@ components:
     version_file: "foreign/python/pyproject.toml"
     version_regex: '(?m)^\s*version\s*=\s*"([^"]+)"'
 
+  sdk-php:
+    tag_pattern: "^php-sdk-([0-9]+\\.[0-9]+\\.[0-9]+(?:-[0-9A-Za-z.-]+)?(?:\\+[0-9A-Za-z.-]+)?)$"
```
hubcio, line 127: packagist derives the release version from the tag NAME via composer's version parser, which only strips a leading `v` - it doesn't understand a `php-sdk-` prefix. so a `php-sdk-0.1.0` tag produces no `0.1.0` release on packagist, and the `Verify Packagist version` jq check (asserts `.version == "0.1.0"` or `"v0.1.0"`) can never match. packagist needs a bare `X.Y.Z` / `vX.Y.Z` tag on whatever repo it indexes; keep `php-sdk-` only for the monorepo github release.

foreign/php/composer.json
```diff
@@ -1,6 +1,7 @@
 {
     "name": "apache/iggy-php",
     "description": "PHP extension bindings for Apache Iggy.",
+    "version": "0.1.0",
```
hubcio: the manual `version` field doesn't rescue the packagist version - for a vcs package composer derives the version from the tag and ignores (or breaks on) a mismatched `version` field. drop this field and let the tag drive the version.

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,565 @@
...
+                exit 1
+              fi
+
+              package_dir="$(mktemp -d)"
+              cp "$extension" "$package_dir/iggy_php.so"
+              cp foreign/php/LICENSE "$package_dir/LICENSE"
+              cp foreign/php/NOTICE "$package_dir/NOTICE"
+              cp foreign/php/LICENSE-binary "$package_dir/LICENSE-binary"
+
+              archive="php_iggy_php-${VERSION}_php${PHP_VERSION}-${PHP_ARCH}-linux-${PHP_LIBC}-nts.zip"
```
hubcio: PIE builds the prebuilt-binary filename from its own token enums: arch is `arm64` (never `aarch64`) and os is `darwin` (never `macos`). this emits `aarch64` here and `macos` at line 421, so of the 12 binary legs only the 4 x86_64-linux ones match what PIE searches for. the 8 arm-linux + macos legs never match, so `pie install` silently falls back to a source build (needs rust+clang on the user box) on all apple silicon, all arm linux, and all macos. rename `aarch64`->`arm64` and `macos`->`darwin` here, at 421, and in the action's `expected[]` list.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,231 @@
...
+
+        if [ ! -d "$PACKAGES_PATH" ]; then
+          echo "PHP packages directory not found: $PACKAGES_PATH"
+          exit 1
+        fi
+
+        expected=(
+          "php_iggy_php-${VERSION}_php8.3-x86_64-linux-glibc-nts.zip"
+          "php_iggy_php-${VERSION}_php8.4-x86_64-linux-glibc-nts.zip"
+          "php_iggy_php-${VERSION}_php8.3-aarch64-linux-glibc-nts.zip"
```
hubcio: same `aarch64`/`macos` token mismatch as the build archive names - PIE searches for `arm64` and `darwin`. these expected entries (85-94) need the same rename so they match what the build produces and what PIE resolves.

.github/workflows/publish.yml
```diff
@@ -1163,6 +1189,40 @@ jobs:
           wheels_artifact: python-wheels-all
           wheels_path: dist
 
+      # ─────────────────────────────────────────
+      # PHP SDK Publishing
+      # ─────────────────────────────────────────
+      - name: Tag PHP SDK release (${{ matrix.key }})
```
hubcio: php tags first here, then publishes and only afterwards waits + verifies packagist. every other sdk publishes, waits until the registry serves, then tags LAST (the `Tag SDK release` step excludes php via `matrix.type != 'php'`). this inverts the workflow's own rule that the tag shouldn't be pushed until the registry is serving, and combined with the packagist issues above, verify fails every run and leaves a stranded immutable tag + public release that needs manual cleanup. move the php tag after verify, or at minimum document a tag/release cleanup runbook.

.github/workflows/publish.yml
```diff
@@ -1163,6 +1189,40 @@ jobs:
...
+          commit: ${{ needs.validate.outputs.commit }}
+          message: |
+            Release ${{ matrix.key }} ${{ steps.ver.outputs.version }}
+
+            Component: ${{ matrix.key }}
+            Tag: ${{ steps.ver.outputs.tag }}
+            Commit: ${{ needs.validate.outputs.commit }}
+            Released by: GitHub Actions (workflow ${{ github.run_id }})
+
+      - name: Publish PHP SDK
```
hubcio: gated only on `matrix.type == 'php'`, not on `skip_tag_creation`, but the `Tag PHP SDK release` step above IS gated on `skip_tag_creation == false`. the action then hard-requires the tag on origin (`git ls-remote --tags --exit-code` at action.yml:161, exit 1 if absent). so a first publish with `skip_tag_creation=true` skips tagging but still runs publish and hard-fails, while every other sdk publishes fine. gate this on `skip_tag_creation` too, or make the action degrade when the tag is absent.

.github/workflows/publish.yml
```diff
@@ -1034,16 +1058,17 @@ jobs:
             Commit: ${{ needs.validate.outputs.commit }}
             Released by: GitHub Actions (workflow ${{ github.run_id }})
 
-  # Non-Docker, non-Rust publishing (Python, Node, Java, C#, Go SDKs)
+  # Non-Docker, non-Rust publishing (Python, PHP, Node, Java, C#, Go SDKs)
   # Note: This job runs in parallel with Docker publishing - no dependency between them
   publish:
     name: ${{ matrix.name }}
-    needs: [validate, plan, check-tags, build-python-wheels, publish-rust-crates]
+    needs: [validate, plan, check-tags, build-python-wheels, build-php-extensions, publish-rust-crates]
```
hubcio: `build-php-extensions` (12 heavy legs, `fail-fast: false`) is a hard `needs` of both `publish` (here) and `publish-docker` (line 757), gated on `result == success || skipped`. one flaky php leg (apt/apk mirror, rustup curl, yq download) flips the whole reusable workflow result to failure and blocks python/node/java/c#/go AND all docker in the same dispatch. it mirrors the existing build-python-wheels coupling but triples the surface and runs uncached. better to gate just the php publish row on the php artifact, not the whole publish jobs.

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,565 @@
...
+          fi
+          echo "version=${version}" >> "$GITHUB_OUTPUT"
+
+      - name: Build source archive
+        run: |
+          set -euo pipefail
+
+          version="${{ steps.ver.outputs.version }}"
+          mkdir -p foreign/php/dist
+          git archive \
```
hubcio, line 127: `git archive HEAD` with no pathspec packs the ENTIRE monorepo (~4200 files: core/, web/, every other sdk) under the `apache-iggy-php-<ver>/` prefix and ships it as the php source release. the `foreign/php/.gitattributes` export-ignore rules only prune within `foreign/php`, so they're inert for everything else. scope it: `git archive HEAD foreign/php` (prefix can stay) drops it to the ~24 php files with export-ignore honored.

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,565 @@
...
+          PHP_VERSION: ${{ matrix.php }}
+          PHP_ARCH: ${{ matrix.arch }}
+          PHP_LIBC: ${{ matrix.libc }}
+          PHP_IMAGE: ${{ matrix.image }}
+        run: |
+          set -euo pipefail
+
+          mkdir -p foreign/php/dist
+
+          docker run --rm \
```
hubcio: the linux legs build via raw `docker run` with no cargo cache - fresh apt/apk + rustup + a cold `cargo build --release` of the full sdk closure across 8 legs every release. the python analog gets `setup-rust-with-cache` + sccache.

.github/workflows/publish.yml
```diff
@@ -1238,6 +1298,52 @@ jobs:
           max_attempts: "15"
           initial_sleep_seconds: "3"
 
+      - name: Wait for Packagist availability
```
hubcio: `Wait for Packagist availability` polls `p2/apache/iggy-php.json`, a package-level metadata file that returns 200 as soon as ANY version exists - so it's not a gating signal for a new version. the `Verify Packagist version` step right after re-polls the same url and is the real gate (its jq checks the specific version). this first step is fully subsumed; drop it and keep verify.

scripts/bump-version.sh
```diff
@@ -257,6 +262,11 @@ translate_version() {
                 edge)   echo "${base}.dev${pre_num}" ;;
                 stable) echo "$base" ;;
             esac ;;
+        php-cargo|composer)
+            case "$pre_type" in
+                edge)   echo "${base}-dev${pre_num}" ;;
```
hubcio: this emits `${base}-dev${pre_num}` for the composer/php-cargo edge format (e.g. `0.1.0-dev1`), copied from the python `-dev` convention. composer's version parser rejects a numbered dev - i checked against composer/semver's normalize regex: `0.1.0-dev1` and `0.1.0-dev.1` are both invalid, while bare `0.1.0-dev`, `0.1.0-alpha1`, `0.1.0-RC1` are valid (the stability-keyword group carries a numeric suffix, the trailing `dev` group does not). so any `bump-version.sh sdk-php --edge` writes a composer.json that composer can't parse and packagist can't ingest, breaking every php prerelease publish. the action's `Validate version format` regex accepts it, so it's silent until packagist sees it. emit bare `-dev` or `-alpha.N`/`-RC.N` for the composer case.

.github/workflows/publish.yml
```diff
@@ -1163,6 +1189,40 @@ jobs:
...
+          message: |
+            Release ${{ matrix.key }} ${{ steps.ver.outputs.version }}
+
+            Component: ${{ matrix.key }}
+            Tag: ${{ steps.ver.outputs.tag }}
+            Commit: ${{ needs.validate.outputs.commit }}
+            Released by: GitHub Actions (workflow ${{ github.run_id }})
+
+      - name: Publish PHP SDK
+        if: matrix.type == 'php'
```
hubcio: minor consistency thing - this bare `if: matrix.type == 'php'` relies on github's implicit `success()` (which is correct, the step is skipped if the tag step fails). the sibling tag/wait/verify steps spell out `success() &&` explicitly, so add it here too for readability.

scripts/extract-version.sh
hubcio: this broadened the pre-release check from `-(edge|rc)` to `-[0-9A-Za-z]`, but two inline copies of the old narrow pattern survive at publish.yml:551 (docker stable-skip) and publish.yml:1015 (`:latest` manifest). the comment above says the regex was consolidated to stop exactly that drift. inert today since docker only sees edge/rc, but e.g. `1.2.3-beta` would now classify as pre-release here and as stable in those two copies. route them through `--is-pre-release`.

.github/workflows/_common.yml
```diff
@@ -42,6 +42,15 @@ jobs:
       - name: Check Python SDK versions are synchronized
         run: ./scripts/ci/python-sdk-version-sync.sh --check
 
+  php-versions:
```
hubcio, line 45: the new `php-versions` job isn't added to the `summary` job's `needs[]` (python-versions is), so its result never shows up in the common checks summary row. add it to the summary needs + body.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,231 @@
...
+        done
+        if [ "$missing" -ne 0 ]; then
+          exit 1
+        fi
+
+        echo "Validating PHP extension archive contents:"
+        for archive in "$PACKAGES_PATH"/*.zip; do
+          filename="$(basename "$archive")"
+          echo "  - $filename"
+          unzip -l "$archive" | awk '{ print $4 }' | grep -qx 'iggy_php.so'
```
hubcio: `unzip -l | awk '{print $4}' | grep -qx` assumes the filename is always column 4, which breaks if a member sits under a path prefix (then `-qx` whole-line match fails). also under `set -e` a missing member aborts the step. `unzip -Z1` is more robust for listing member names.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,231 @@
...
+          echo "PHP release tag input is empty"
+          exit 1
+        fi
+        if ! git ls-remote --tags --exit-code origin "refs/tags/${TAG}" >/dev/null 2>&1; then
+          echo "PHP release tag $TAG does not exist on origin"
+          echo "Create the annotated PHP SDK tag before publishing GitHub release assets or refreshing Packagist."
+          exit 1
+        fi
+
+        release_notes="$(mktemp)"
```
hubcio, line 158: `release_notes` (and `response` at 210) are mktemp files that never get a trap cleanup. harmless on ephemeral runners, just noting it since create-git-tag does trap.

scripts/ci/php-sdk-version-sync.sh
```diff
@@ -0,0 +1,109 @@
...
+    echo -e "${YELLOW}Run '$0 --fix' to update composer.json from Cargo.toml${NC}"
+    exit 1
+fi
+
+if [ "$CARGO_VERSION" = "$COMPOSER_VERSION" ]; then
+    echo -e "${GREEN}✓ PHP SDK versions are already synchronized${NC}"
+    exit 0
+fi
+
+perl -0pi -e "s/\"version\"\\s*:\\s*\"\\Q$COMPOSER_VERSION\\E\"/\"version\": \"$CARGO_VERSION\"/" "$COMPOSER_JSON"
```
hubcio: prints `✓ Updated` unconditionally right after the perl substitution, without checking it actually changed anything. assert the post-write version equals `$CARGO_VERSION` so a no-op substitution (e.g. a malformed composer version) doesn't report false success.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,231 @@
...
+        {
+          echo "Apache Iggy PHP SDK ${VERSION}"
+          echo ""
+          echo "This GitHub release contains downstream convenience binaries for the PHP extension."
+          echo "The canonical Apache source release remains the ASF release artifact."
+        } > "$release_notes"
+
+        prerelease_flag=()
+        if [[ "$VERSION" == *-* ]]; then
```
hubcio: `[[ "$VERSION" == *-* ]]` is a 4th independent encoding of 'is pre-release' (alongside `--is-pre-release` and the two inline `-(edge|rc)` copies). consistent for `-dev`/`-rc` today but drifts from the central rule. could call `extract-version.sh --is-pre-release`.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,231 @@
...
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }}
+
+    - name: Trigger Packagist update
+      if: inputs.dry_run == 'false'
+      run: |
+        set -euo pipefail
+
+        if [ -z "${PACKAGIST_TOKEN:-}" ]; then
```
hubcio: the PACKAGIST_TOKEN presence/format check runs only after the github release is created and assets uploaded (line 185). it reads nothing but the env var, so hoist it to the first step to fail fast - otherwise a missing/malformed token leaves a published release (and pushed tag) before the job fails.

scripts/ci/php-sdk-version-sync.sh
```diff
@@ -0,0 +1,109 @@
...
+fi
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+cd "$REPO_ROOT"
+
+CARGO_TOML="foreign/php/Cargo.toml"
+COMPOSER_JSON="foreign/php/composer.json"
+
+CARGO_VERSION=$(grep '^version = ' "$CARGO_TOML" | head -1 | sed 's/version = "\(.*\)"/\1/')
+COMPOSER_VERSION=$(perl -0777 -ne 'if (/"version"\s*:\s*"([^"]+)"/) { print $1; exit; }' "$COMPOSER_JSON")
```
hubcio: this reads (70) and rewrites (108) the composer `version` field with perl, while the sibling python-sdk-version-sync.sh and bump-version.sh parse the same field with grep+sed. works since perl is on the runners, but it's an avoidable tooling inconsistency - reuse the grep/sed pattern.

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,565 @@
...
+          /tmp/copy-latest-from-master.sh save .github scripts
+          /tmp/copy-latest-from-master.sh apply
+
+      - name: Download license manifest
+        uses: actions/download-artifact@v8
+        with:
+          name: php-license-binary
+          path: foreign/php
+
+      - name: Setup yq
```
hubcio: the `Setup yq` block (~30 lines) is duplicated verbatim 3x (159, 351, 456), and `Resolve version` 3x (192, 384, 489). extract a small composite action (like the existing utils/ composites) and reuse it to kill the 3-way pin drift.

hubcio (changes requested):

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,469 @@
...
+          VERSION: ${{ steps.release.outputs.version }}
+          PHP_VERSION: ${{ matrix.php }}
+          PHP_ARCH: ${{ matrix.arch }}
+          CARGO_TARGET_DIR: ${{ github.workspace }}/target/php-release-${{ matrix.php }}-${{ matrix.arch }}-macos
+        run: |
+          set -euo pipefail
+
+          php --version
+          php-config --version
+          cargo build --release --manifest-path foreign/php/Cargo.toml
```
hubcio: this builds from the repo root with `--manifest-path`, but cargo config discovery is cwd-based - `foreign/php/.cargo/config.toml` (the `-Wl,-undefined,dynamic_lookup` link args) never loads. ext-php-rs 0.15.14 doesn't emit those link args from its build script (it only ships them in its own crate-local `.cargo/config.toml`, which doesn't propagate to dependents), so all 4 macos legs will fail linking with undefined zend symbols. fix: `cd foreign/php && cargo build --release` and adjust `CARGO_TARGET_DIR`/`find` paths.

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,469 @@
...
+              if [ ! -x "$CARGO_HOME/bin/rustup" ]; then
+                curl -sSf https://sh.rustup.rs -o /tmp/rustup-init.sh
+                sh /tmp/rustup-init.sh -y --profile minimal --default-toolchain none
+              fi
+              . "$CARGO_HOME/env"
+              rustup show
+
+              php --version
+              php-config --version
+              cargo build --release --manifest-path foreign/php/Cargo.toml
```
hubcio: rustup toolchains on alpine default to `+crt-static`, and a cdylib on musl needs `RUSTFLAGS="-C target-feature=-crt-static"` to link - ext-php-rs itself sets exactly this in its own crate-local `.cargo/config.toml` for `cfg(target_env = "musl")`, but that doesn't propagate to dependents and nothing sets it here. expect the 4 musl legs to fail at link time. adding a musl target section to `foreign/php/.cargo/config.toml` works, but only combined with fixing the cwd so that config actually loads (see the macos comment).

.github/workflows/publish.yml
```diff
@@ -1314,6 +1340,101 @@ jobs:
...
+
+  publish-php:
+    name: PHP SDK
+    needs: [validate, plan, check-tags, build-php-extensions]
+    if: |
+      !cancelled() &&
+      needs.validate.outputs.has_targets == 'true' &&
+      needs.plan.outputs.has_php == 'true' &&
+      fromJson(needs.plan.outputs.php_targets).include[0].key != 'noop' &&
+      (needs.build-php-extensions.result == 'success' || needs.build-php-extensions.result == 'skipped')
```
hubcio: when php is selected (`has_php == 'true'`, so validate and plan must have succeeded), `build-php-extensions` can only end up `skipped` here if check-tags failed - and in that case this job still runs (`!cancelled()`), pushes the tag (the tag step runs before the publish step) and then fails downloading the `php-extensions-all` artifact that was never built. result: tag without release, and post-merge auto-publish permanently skips that version because the tag exists. dropping the `!cancelled()` + result checks and relying on plain needs semantics with the `has_php` condition gives exactly the right behavior.

.github/workflows/publish.yml
```diff
@@ -982,13 +1008,13 @@ jobs:
...
             # Manual publish: always push the versioned manifest, plus :latest
-            # for stable (non edge/rc) releases.
+            # for stable releases.
             docker buildx imagetools create \
               -t "${IMAGE}:${VERSION}" \
               $(printf "${IMAGE}@sha256:%s " *)
             echo "✅ Pushed manifest: ${IMAGE}:${VERSION}"
 
-            if [[ ! "$VERSION" =~ -(edge|rc) ]]; then
+            if [ "$(scripts/extract-version.sh "${{ matrix.key }}" --is-pre-release)" != "true" ]; then
```
hubcio: if `extract-version.sh` fails here, the command substitution comes back empty, `!= "true"` passes and `:latest` gets pushed for a possibly-prerelease version. the check-tags variant at line 554 guards this with `|| echo false` (fails safe by skipping); this one should fail safe too - `|| echo true`.

scripts/ci/php-sdk-version-sync.sh
```diff
@@ -0,0 +1,172 @@
...
+    if grep -q '^iggy = .*version = ' "$PHP_CARGO_TOML"; then
+        sedi "s/^\(iggy = .*version = \"\)[^\"]*/\1${RUST_SDK_VERSION}/" "$PHP_CARGO_TOML"
+    else
+        sedi "s/^\(iggy = {[^}]*\) }/\1, version = \"${RUST_SDK_VERSION}\" }/" "$PHP_CARGO_TOML"
+    fi
+    changed=1
+fi
+
+if [ -n "$COMPOSER_VERSION" ]; then
+    sedi '/^[[:space:]]*"version"[[:space:]]*:/d' "$COMPOSER_JSON"
```
hubcio, line 156: deleting the `"version"` line with sed breaks composer.json when the field happens to be the last key in the object - the previous line keeps its trailing comma and the file is no longer valid json. the re-grep at line 157 only proves the line is gone, not that the json still parses. cheap guard: `python3 -c 'import json,sys;json.load(open(sys.argv[1]))' foreign/php/composer.json` after the edit.

scripts/bump-version.sh
```diff
@@ -319,6 +324,10 @@ read_current_version() {
             grep '^version = ' "$abs_file" | head -1 | sed 's/version = "\(.*\)"/\1/' ;;
         pyproject)
             grep '^version = ' "$abs_file" | head -1 | sed 's/version = "\(.*\)"/\1/' ;;
+        php-cargo)
+            grep '^version = ' "$abs_file" | head -1 | sed 's/version = "\(.*\)"/\1/' ;;
+        composer)
```
hubcio: also at line 377. the `composer` format cases are unreachable - no component maps to `composer` (sdk-php uses `php-cargo`), and the version field is banned from composer.json by design (php-sdk-version-sync.sh enforces its absence). dead code, delete both.

scripts/extract-version.sh
hubcio: `-[0-9A-Za-z]` also matches a hyphen inside semver build metadata, so `1.2.3+linux-x64` would classify as pre-release. latent today (no component uses build metadata), but anchoring on the patch part is cheap: `^[0-9]+\.[0-9]+\.[0-9]+-`.

.github/workflows/publish.yml
```diff
@@ -1314,6 +1340,101 @@ jobs:
       # above ensure the tag is never pushed before the registry is serving
       # the artifact (Go is exempt: the tag IS the release).
       - name: Tag SDK release (${{ matrix.key }})
+        if: |
+          success() &&
+          inputs.dry_run == false &&
+          matrix.type != 'php' &&
```
hubcio: `matrix.type != 'php'` is dead - the plan step already filters php out of `non_docker_targets`, so this matrix never contains a php entry. delete it

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,193 @@
...
+
+        PACKAGES_PATH="${{ inputs.packages_path }}"
+        VERSION="${{ inputs.version }}"
+
+        if [ ! -d "$PACKAGES_PATH" ]; then
+          echo "PHP packages directory not found: $PACKAGES_PATH"
+          exit 1
+        fi
+
+        expected=(
```
hubcio, line 78: these 13 filenames duplicate the build matrix by hand - a platform added to `_build_php_extensions.yml` later won't be validated here (removal fails hard, which is the safe direction). fine as-is, just flagging the coupling in case a generated manifest is worth it later.

hubcio (changes requested):
a few points that don't anchor to diff lines:

- `foreign/php/Cargo.toml` package version is `0.1.0` (stable), so the post-merge auto-publish loop hits "Stable version - skipping" and the pipeline stays dormant until the version gets bumped to a prerelease (e.g. `0.1.1-edge.1`). if that's the intended launch state, fine - just calling it out. if  you want  to test it straight after commit is merged, you have to add dev or edge suffix (like the rest of sdks)
- `_build_php_extensions.yml` is workflow_call-only, so none of the 13 legs have ever executed in CI. before bumping sdk-php to a prerelease, a `dry_run=true` dispatch with `publish_other=php` would exercise all legs plus the download/validate path without tagging or publishing. also worth smoke-testing `pie install` against a draft release at some point, since the asset naming convention can't be verified offline.
- bigger-picture suggestion: consider dropping the prebuilt binaries entirely for now and shipping source-only. PIE resolves packages via packagist, and packagist isn't wired up for the monorepo (as the PR notes) - so `pie install apache/iggy-php` can't work today either way, and the binary zips are only consumable via manual download-unzip-edit-php.ini. that's 12 CI legs (2 platforms of which are currently broken per the line comments) for a flow almost nobody can use yet. linux/mac PHP extensions are conventionally source-built anyway; prebuilt matters most on windows, which this matrix doesn't ship. the source tarball + license manifest legs could stay. binaries can come back once the split repo + packagist listing exist and PIE can actually consume them.
- follow-up regardless of direction: `foreign/php/` carries no `rust-toolchain.toml` and no `rust-version` in Cargo.toml, so source builds (tarball or future PIE `composer-default`) use whatever rust the user has - MSRV drift is unguarded.

3rd point is most important for me now. if its not possible, please lets discuss that on discord php client channel.

.github/actions/php/post-merge/action.yml
```diff
@@ -0,0 +1,185 @@
...
+        if [ "$(scripts/extract-version.sh sdk-php --is-pre-release)" = "true" ]; then
+          prerelease_flag=(--prerelease)
+        fi
+
+        if gh release view "$TAG" >/dev/null 2>&1; then
+          echo "GitHub release $TAG exists, uploading assets with clobber"
+          gh release upload "$TAG" "$PACKAGES_PATH"/* --clobber
+        else
+          echo "Creating GitHub release $TAG"
+          gh release create "$TAG" "$PACKAGES_PATH"/* \
```
hubcio, line 177: this is the part i can't accept - `gh release create` would make php the only versioned-pipeline component publishing to github releases (the only other github release is the rolling `edge` pre-release; versioned ones are server artifacts, `server-X.Y.Z`). with this step gone the rest of the action (asset download/validation, release notes, prerelease flag) has no purpose either - delete the whole action and let the publish-php job end at tag creation.

.github/workflows/_build_php_extensions.yml
```diff
@@ -0,0 +1,188 @@
...
+          name: php-license-binary
+          path: foreign/php
+
+      - name: Build source package
+        run: |
+          set -euo pipefail
+
+          version="${{ steps.release.outputs.version }}"
+          mkdir -p foreign/php/dist
+          git archive \
```
hubcio, line 127: with github releases dropped this tarball + license manifest have no consumer (a 30-day workflow artifact is not distribution), so the whole workflow can go until the real packaging story starts. one finding to carry into that future work: the archive as built here cannot compile standalone. it ships `iggy = { path = "../../core/sdk", version = ... }` and cargo has no registry fallback for a missing path - i extracted the tarball and ran `cargo metadata`, it fails with `failed to load source for dependency iggy`. stripping the `path` key makes it resolve cleanly from crates.io (edge versions are published there). manual builds inside a repo clone are unaffected.

.github/workflows/publish.yml
```diff
@@ -1336,6 +1362,128 @@ jobs:
...
+          commit: ${{ needs.validate.outputs.commit }}
+          message: |
+            Release ${{ matrix.key }} ${{ steps.ver.outputs.version }}
+
+            Component: ${{ matrix.key }}
+            Tag: ${{ steps.ver.outputs.tag }}
+            Commit: ${{ needs.validate.outputs.commit }}
+            Released by: GitHub Actions (workflow ${{ github.run_id }})
+
+      - name: Publish PHP SDK
```
hubcio, line 1452: with the release publishing dropped this job reduces to extract-version + create-git-tag. keep the tag - it is what the post-merge loop keys on and what a future split-repo mirror would sync from. worth a short comment at the tag step (or a line in the PR description) that a `php-sdk-*` tag means 'version reserved', not 'artifact shipped' - it will be the one sdk tag with no registry behind it.

.github/config/publish.yml
```diff
@@ -123,6 +123,12 @@ components:
...
+    registry: github-release
```
hubcio, line 128: once nothing creates releases, `github-release` here is only a routing label for the plan step. rename it to something honest or drop a comment, otherwise the next reader goes looking for release-publishing code that no longer exists.

foreign/php/rust-toolchain.toml
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[toolchain]
+channel = "1.96.0"
```
hubcio, line 19: good pin to have - manual source builds are the official path for now, so this is the effective msrv guard. but this is now three hand-synced copies of 1.96.0 (this channel, `rust-version` in Cargo.toml, the root rust-toolchain.toml) and `sync-rustc-version.sh` only reads the root one, so they will drift silently on the next toolchain bump. follow-up: teach that script about foreign/php.

scripts/ci/php-sdk-version-sync.sh
```diff
@@ -0,0 +1,181 @@
...
```
hubcio, line 156: `--fix` deletes the line with sed and validates only after the write - if `version` ever sits as the last key, the deletion leaves a trailing comma and composer.json stays broken on disk (reproduced locally; the validator does catch it and exits 1, which is the guard i asked for). since python3 is already a hard dependency here, doing the removal json-aware would avoid leaving a corrupted file behind. nit - pre-commit-only surface.

hubcio (changes requested):
first, credit where due: all my previous rounds check out in this revision - i verified the fixes and ran the scripts locally (sync check, bump round-trips, archive contents), all green.

this review is a direction change instead. i don't want any php code creating github releases. releases in this repo are cut from server code (hence the `server-0.8.0` style entries), and every other sdk publishes to its registry plus a git tag - none of them touch the releases page. with packagist out of scope there is simply no distribution channel for php right now, and i accept that: manual way of working for now (clone the repo, `cargo build` in `foreign/php`, load the .so). once some users explain the need for an actual package, we can think of something - most likely a split repo + packagist/pie.

so keep the valuable half of this PR: version metadata sync (`php-sdk-version-sync.sh`, bump/extract integration, the `_common.yml` check, pre-commit hooks) and the `php-sdk-*` tag creation in the release train. drop the github release publishing and the artifact production that only exists to feed it. line comments point at the specific pieces.

hubcio (commented):
closing this one - decision, not code quality. this supersedes my review from earlier today: don't act on those line comments, the direction changed after we dug into the distribution story.

where we landed:

- versioned github releases in this repo are server artifacts (`server-X.Y.Z`, plus the rolling `edge` pre-release) - no foreign sdk publishes github releases, and php won't be the first. that removes the only publishing target this pipeline had left once packagist was out of scope.
- packagist.org cannot index a monorepo subdirectory, by design: it hosts no code, requires `composer.json` at the repo root and derives versions from bare `vX.Y.Z` git tags. our layout (`foreign/php/`, `php-sdk-*` tags) satisfies neither, and packagist have stated subdirectory support won't happen. pecl is effectively dead and pie resolves through packagist, so there is genuinely no channel to ship to from this repo today.
- so for now the php sdk stays manual-only: clone the repo, `cargo build` in `foreign/php`, load the extension in php.ini. acceptable until real users ask for a proper package.
- when they do, the standard fix is a read-only split mirror (`apache/iggy-php`) populated by a subtree-split action on each `php-sdk-*` tag, with bare `vX.Y.Z` tags and a packagist webhook - the symfony/laravel/google-cloud-php pattern. at that point we also decide rust extension vs pure-php client (most of our sdks are pure implementations of the wire protocol anyway).

worth salvaging from this PR when that day comes: the version sync machinery (`php-sdk-version-sync.sh`, bump/extract integration, CI checks) is solid - i verified it end to end. one gotcha to remember for any out-of-monorepo packaging: the `iggy = { path = "../../core/sdk", ... }` dependency must have the `path` stripped - cargo does not fall back to the registry when a path is missing (checked: a packaged tree fails to resolve, version-only resolves fine from crates.io).

thanks for the thorough work across the review rounds - all earlier comments were properly addressed, this closure is about repo and ecosystem constraints, not the implementation. if you want to pick up the split-repo + packagist work once demand shows up, let's talk in the php channel on discord.

PR #3512, opened 2026-06-19
fix(partitions): persist data on vsr backups, add data integrity test

hubcio:
VSR's hash chain and checksum-keyed recovery require a committed op's
on-disk bytes to match on every replica. Two defects broke that: the
partition commit path drained the pipeline, which only the primary
fills, so backups journaled replicated prepares but never flushed them
(0-byte segments); and append re-stamped base_timestamp from a local
now(), diverging bytes even once persisted.

commit_journal now falls back to the journal when the pipeline is
empty, so backups persist like the metadata plane. base_timestamp
reuses the prepare's monotonic timestamp, stamped once by the primary
and replicated. The 3-node data integrity test is un-ignored and gates
this.


numinnex (commented):
### Findings

**1. (minor/perf) `core/shard/src/lib.rs` — per-prepare config clone.** `planes.1.0.config().clone()` clones `PartitionsConfig` on every follower partition prepare (hot replication path). It reads as a borrow-checker workaround (release the immutable `config()` borrow before the mutable `get_mut_by_ns`). Bounded, but a per-message allocation — consider hoisting the config out of the per-prepare branch or restructuring the borrow.

**2. (minor) `commit_messages` threshold counts the uncommitted tail.** `should_persist` is computed from `journal_info` (committed + uncommitted tail currently resident), but only the committed prefix is flushed. With `messages_required_to_save > 1`, uncommitted-tail bytes count toward the flush trigger, so a flush can fire before the *committed* bytes alone reach the threshold. Behaviorally safe (only committed bytes are written), but the threshold no longer means "pending committed bytes." Worth a one-line comment so a future reader doesn't read it as committed-only.

mmodzelewski (approved):

spetz (approved):

PR #3515, opened 2026-06-20
feat(connectors): opensearch source connector

### LINKED ISSUE #3503: feat(connector): Add Opensearch source connector
### Description

Add a Opensearch source connector that polls documents from a Opensearch index and produces JSON messages into Iggy.

Users should be able to ingest Opensearch documents into Iggy streams with stable cursor-based polling and persisted connector state.

### Affected area / component

Connectors

### Proposed solution

Scope
Add the source connector crate and sample runtime config.
Document source configuration, retry behavior, and primary-key cursor requirements.
Add Docker-backed integration coverage for producing indexed documents.

### Alternatives considered

_No response_

### Contribution

### Good first issue

ryerraguntla:
**Which issue does this PR close?**
Closes #3503

**Rationale**
Iggy had no native way to ingest documents from OpenSearch  clusters into Iggy streams. This connector fills that gap for observability pipelines, audit log forwarding, and data lake ingestion scenarios where OpenSearch is the source of truth.

**What changed?**
**Before**: no OpenSearch source connector existed. Users had to write custom consumers using the OpenSearch client to bridge documents into Iggy.

**After**: a new iggy_connector_opensearch_source cdylib plugin implements the Source trait and is loaded by the connectors runtime via FFI. Each poll() issues a search_after-paginated query on (timestamp_field, _id), publishes each hit's _source as a JSON ProducedMessage, and persists the cursor in ConnectorState.

**Module layout**
```
core/connectors/sources/opensearch_source/
├── Cargo.toml
├── README.md
├── dependencies.md
└── src/
    ├── lib.rs            # Source trait, poll loop, search_after cursor, config, 16 unit tests
    ├── state_manager.rs  # File-backed SourceState mirror, atomic rename protocol, 4 unit tests
    ├── retry.rs  # Retry logic and circuit breaker logic. 2 unit tests 
    └── http_tests.rs     # Axum mock server tests (20 tests, no Docker required)
core/connectors/runtime/example_config/connectors/
└── opensearch_source.toml
core/integration/tests/connectors/opensearch/
├── mod.rs
├── opensearch_source.rs      # 7 Docker-backed integration tests
├── opensearch_source_types.rs
├── opensearch_source_resilience.rs
└── fixtures/opensearch/      # Container start/stop, document indexing helpers

```

**Key implementation decisionsCursor mechanism** — 
**search_after on (timestamp_field, _id) sorted ascending.** 
Two sort keys give stable pagination when timestamps collide; _id is the tiebreaker. No Scroll API, PIT, or offset paging.

**State authority** — runtime **ConnectorState** (MessagePack) is authoritative. An optional file-backed JSON mirror (SourceState) is written atomically on close() using write-tmp → fdatasync → rename → dir-fsync. On restart, if a valid runtime state exists it takes priority and file state is not loaded. If state deserialization fails, open() returns InitError to prevent a silent cursor reset.

**Validation** at open() — timestamp_field is required and non-empty (validated before any network call). batch_size = 0 is rejected. File state config is validated if state.enabled = true.

**Error handling** — transient OpenSearch errors fail the poll with the error counter incremented. No circuit breaker or HTTP retry in this version (documented limitation). poll() still sleeps polling_interval on error so the runtime does not spin.

**At-least-once delivery** — cursor advances in memory before ConnectorState is persisted by the runtime. A crash after poll() returns but before the runtime writes state can re-emit the last batch. Documented in README and Limitations.

**Validation**

```
cargo fmt --all
cargo sort --no-format --workspace
cargo clippy -p iggy_connector_opensearch_source --all-targets -- -D warnings
cargo test -p iggy_connector_opensearch_source
./scripts/ci/taplo.sh
./scripts/ci/license-headers.sh
# Docker-backed integration (requires running iggy-server and iggy-connectors binaries):
env CARGO_BIN_EXE_iggy-server=... CARGO_BIN_EXE_iggy-connectors=... \
  cargo test -p integration --test mod -- connectors::opensearch::opensearch_source

```

All 40 tests pass (16 unit + 4 state_manager unit + 20 axum HTTP mock + 7 Docker integration).

**Local Execution**
Passed
Pre-commit hooks ran

**AI Usage**

1. Tools used: Claude
2. Scope: scaffolding initial file structure and boilerplate; all logic reviewed and iterated manually
3. Verification: full local compilation, all tests executed, end-to-end tested with Docker OpenSearch + Iggy server + connector runtime
4. Can explain every line: yes

**Configuration reference**
```
type = "source"
key = "opensearch"
enabled = true
version = 0
name = "OpenSearch source"
path = "<BASE_DIR>/target/release/libiggy_connector_opensearch_source"
plugin_config_format = "json"
[[streams]]
stream = "opensearch_stream"
topic  = "documents"
schema = "json"
batch_length = 100
linger_time  = "5ms"
[plugin_config]
url              = "http://localhost:9200"
index            = "logs-*"
timestamp_field  = "@timestamp"       # required; must exist on every document
polling_interval = "10s"              # default
batch_size       = 100                # default
# username       = "admin"
# password       = "replace_with_secret"
# verbose_logging = false
# Optional: custom query DSL (default: match_all)
# [plugin_config.query]
# match = { "log.level" = "error" }
# Optional: file-backed state mirror (runtime msgpack is authoritative)
# [plugin_config.state]
# enabled      = true
# storage_type = "file"
# state_id     = "opensearch_logs_connector"
# [plugin_config.state.storage_config]
# base_path = "./connector_states"

```

<h2>Required Fields</h2>

Field | Type | Description
-- | -- | --
url | String | OpenSearch HTTP base URL
index | String | Index name or pattern (logs-* supported)
timestamp_field | String | Document field used as sort key and cursor; must exist on every document

</body></html>

<h2 class="text-text-100 mt-3 -mb-1 text-[1.125rem] font-bold" data-sourcepos="11:1-11:19;335-353">Optional Fields</h2>

<div class="overflow-x-auto w-full px-2 mb-6" data-sourcepos="13:1-19:92;355-829">

Field | Type | Default | Description
-- | -- | -- | --
polling_interval | humantime | "10s" | Sleep after each completed poll cycle
batch_size | usize | 100 | Documents per search request (min 1)
query | JSON | {"match_all":{}} | OpenSearch query DSL applied on every poll
username / password | String | — | HTTP Basic auth
verbose_logging | bool | false | Log per-poll counts at info! instead of debug!

</div>

**Poll cycle**

Each poll() call:Issues POST /{index}/_search sorted by (timestamp_field asc, _id asc) with optional search_after.Maps each hit's _source to a JSON ProducedMessage.Advances the search_after cursor to the sort tuple of the last successfully published hit.Returns ProducedMessages with serialized ConnectorState.Sleeps polling_interval.Hits missing a sort tuple or _source are skipped with warn!; they do not block the cursor.

**Known limitations**

1. At-least-once delivery — cursor advances before the runtime persists ConnectorState; last batch may be re-emitted on crash.
2. Single sequential reader — one search_after cursor, one batch per poll; no parallel shard workers.
3. search_after only — no Scroll API, PIT, or sliced export.
4. Backfill gap — documents with timestamp_field older than the current cursor are not read without a state reset.
5. _source-disabled documents skipped permanently.Single-node transport — no cluster sniffing.


numinnex (approved):

numinnex (dismissed):

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1136 @@
...
+            .as_ref()
+            .ok_or_else(|| Error::Connection("OpenSearch client not initialized".to_string()))?;
+
+        match self.search_documents(client).await {
+            Ok(outcome) => {
+                self.circuit_breaker.record_success();
+                let processing_time = start_time.elapsed().as_millis() as f64;
+                let (messages, persisted_state) =
+                    self.finalize_poll(outcome, processing_time).await;
+                sleep(self.polling_interval).await;
```
hubcio: the sleep runs after `finalize_poll()` but before the batch is returned, so every fetched batch is held in the plugin for a full `polling_interval` before the runtime can send it. both siblings (`elasticsearch_source`, `influxdb_source`) sleep at the top of `poll()` instead - same pacing, but the data they hand back is fresh. throughput is identical either way; this is purely delivery latency. if you move it to the top, mind that the README documents "first poll runs immediately" - a plain move re-introduces the one-interval wait before the first poll, so either update the README to match sibling behavior or skip the sleep on the first iteration.

core/connectors/sources/opensearch_source/README.md
```diff
@@ -0,0 +1,296 @@
...
+| Transient (retry) | Network errors; HTTP `429`; HTTP `5xx`; honors `Retry-After` header on `429` |
+| Permanent (no retry) | `400`, `401`, `403`, `404`; malformed responses; search DSL errors |
+
+### Circuit breaker
+
+When open, `poll()` skips the search, logs a warning, sleeps `polling_interval`, and returns an empty batch with no state update - cursor does not advance. Consecutive failures after retries exhausted increment the breaker; a successful search resets it.
+
+### Delivery semantics
+
+**At-least-once toward Iggy.** The in-memory cursor advances in `finalize_poll()` before the runtime persists `ConnectorState`. A crash after cursor advance but before the runtime save re-emits the last batch on restart.
```
hubcio: "at-least-once" only holds across a crash. on an in-process `send()` failure the runtime logs the error and keeps polling without persisting state, but the in-memory cursor already advanced in `finalize_poll()` - the failed batch gets skipped until the next restart, i.e. at-most-once for that batch. worth stating that case here too, since persist-on-send-success is the runtime's contract for all sources.

core/connectors/sources/opensearch_source/README.md
```diff
@@ -0,0 +1,296 @@
...
+| `Storage` | Network or HTTP errors; client not initialized at `poll()` |
+| `Serialization` | JSON / MessagePack failures |
+
+## Resilience
+
+### Retry policy
+
+| Category | Conditions |
+| -------- | ---------- |
+| Transient (retry) | Network errors; HTTP `429`; HTTP `5xx`; honors `Retry-After` header on `429` |
```
hubcio: code honors `Retry-After` on 5xx as well, not just 429 - `sleep_before_retry` receives the header for any transient status and `is_transient_status` covers 429 + 5xx. correct behavior (503 with `Retry-After` is legit per RFC 7231), just undersold here.

core/connectors/sources/opensearch_source/dependencies.md
```diff
@@ -0,0 +1,46 @@
...
+
+---
+
+## Runtime dependencies
+
+| Crate | Version (workspace) | License | Role in this connector |
+| --- | --- | --- | --- |
+| `async-trait` | `^0.1.89` | MIT / Apache-2.0 | Proc-macro that enables `async fn` in trait definitions; required by the `Source` trait impl in `lib.rs`. |
+| `dashmap` | `^6.1.0` | MIT | Concurrent hash map; injected into this crate's namespace by the `source_connector!` macro expansion in the SDK. Not used directly in source files. |
+| `humantime` | `^2.3.0` | MIT / Apache-2.0 | Workspace dependency; duration parsing uses `iggy_connector_sdk::retry::parse_duration`. |
```
hubcio: `humantime` isn't in this crate's `[dependencies]` - the row itself admits it's transitive via the SDK's `parse_duration`, contradicting the header's "every direct (non-dev) dependency" claim. rather than fixing the row, suggest deleting this file: it duplicates `Cargo.toml` and drifts silently (this row is the proof). the closest sibling `elasticsearch_source` ships without one (only the influxdb connectors carry it), and license enumeration is already handled by the third-party-licenses CI script.

delete this dependencies.md file.

core/connectors/runtime/example_config/connectors/opensearch_source.toml
```diff
@@ -0,0 +1,63 @@
...
+url = "http://localhost:9200"
+index = "logs-*"
+polling_interval = "10s"
+batch_size = 100
+timestamp_field = "@timestamp"
+# username = "admin"
+# password = "replace_with_secret"
+# verbose_logging = false
+
+# HTTP resilience (defaults shown; see docs/RESILIENCE.md).
```
hubcio: `docs/RESILIENCE.md` doesn't exist anywhere in the repo.

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1136 @@
...
+        Ok(SourceState {
+            id: self.get_state_id(),
+            last_updated: Utc::now(),
+            version: SOURCE_STATE_VERSION,
+            data,
+            metadata: Some(json!({
+                "connector_type": "opensearch_source",
+                "connector_id": self.id,
+                "index": self.config.index,
+                "url": self.config.url,
```
hubcio: `config.url` is written verbatim into file-state metadata and lands on disk as plaintext json when file state is enabled. a url with embedded userinfo (`http://user:pass@host`) puts the credential at rest. the metadata block is informational and unused on restore (`source_state_to_internal_state` reads only `data` + `version`), so dropping `url` (or stripping userinfo) is free. `elasticsearch_source` has the same pattern, no need to copy it forward.

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1136 @@
...
+
+    fn batch_size(&self) -> usize {
+        self.config.batch_size.unwrap_or(DEFAULT_BATCH_SIZE)
+    }
+
+    fn timestamp_field(&self) -> &str {
+        self.config
+            .timestamp_field
+            .as_deref()
+            .expect("timestamp_field validated at open()")
```
hubcio: `expect` on a prod path. it's unreachable today (`validate_open_config` rejects `None`/empty before any caller), but nothing type-enforces that. `poll()` runs in a spawned task inside the plugin, so a panic here wouldn't crash anything visibly - it kills the polling task and the connector silently stops. an `Err` gets logged and the loop continues. resolving the field once after validation (or returning `Err`) avoids that failure mode if a future refactor reorders things.

core/connectors/sources/opensearch_source/src/state_manager.rs
```diff
@@ -0,0 +1,316 @@
...
+    match config.storage_type.as_deref() {
+        Some("file") | None => Ok(()),
+        Some(storage_type) => Err(Error::InvalidConfigValue(format!(
+            "state storage_type {storage_type:?} is not supported; only \"file\" is implemented"
+        ))),
+    }
+}
+
+pub(crate) fn create_state_storage(config: &StateConfig) -> Result<Arc<dyn StateStorage>, Error> {
+    validate_state_storage_config(config)?;
```
hubcio: both prod callers reach this after `validate_open_config` already ran `validate_state_storage_config` (`load_state` via `open()`, `save_state` via `close()` which is gated on a successful open), so this re-validation can never fail here. safe to drop.

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1136 @@
...
+
+#[derive(Debug)]
+pub struct OpenSearchSource {
+    id: u32,
+    config: OpenSearchSourceConfig,
+    client: Option<OpenSearch>,
+    polling_interval: Duration,
+    /// Pre-built query body (query + size + sort). Set once in `open()` after config validation.
+    /// `None` before `open()` and after a failed `open()`.
+    search_body_base: Option<Value>,
```
hubcio, line 129: comment isn't quite accurate - `search_body_base` is assigned before `create_client()` / `check_index_exists_with_retry()`, so an `open()` that fails past that point leaves it `Some`. harmless (`poll()` gates on `client` first), but the "after a failed `open()`" promise doesn't hold.

hubcio (changes requested):
one finding outside the diff: the default file-state path `./connector_states` isn't covered by root `.gitignore` (only `local_data*` / `local_state*` are), so locally-produced state files are easy to commit by accident. worth adding `connector_states*`.

core/connectors/runtime/example_config/connectors/opensearch_source.toml
ryerraguntla: It should mention the opensearch projects repo name as well. Removing as suggested

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: Fixed in [64aaf12](https://github.com/apache/iggy/pull/3515/commits/64aaf120c1a2eb91caa94fb108da732e6111e56b)

ryerraguntla:
@numinnex - I am sorry. I accidentally dismissed your earlier review, could you please nudge again.

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: Fixed in [7e7cfee](https://github.com/apache/iggy/pull/3515/commits/7e7cfee3dce59bb38f66e78e7cc825d2f6d29bda)

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: Fixed in [7e7cfee](https://github.com/apache/iggy/pull/3515/commits/7e7cfee3dce59bb38f66e78e7cc825d2f6d29bda)

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: Fixed in [7e7cfee](https://github.com/apache/iggy/pull/3515/commits/7e7cfee3dce59bb38f66e78e7cc825d2f6d29bda)

core/connectors/sources/opensearch_source/src/state_manager.rs
ryerraguntla: Fixed in [bc49ad0](https://github.com/apache/iggy/pull/3515/commits/bc49ad09c5ffe27987e369159ecf4d0087f8f0d6)

core/connectors/runtime/example_config/connectors/opensearch_source.toml
```diff
@@ -0,0 +1,63 @@
...
+url = "http://localhost:9200"
+index = "logs-*"
+polling_interval = "10s"
+batch_size = 100
+timestamp_field = "@timestamp"
+# username = "admin"
+# password = "replace_with_secret"
+# verbose_logging = false
+
+# HTTP resilience (defaults shown;).
```
hubcio: remove the `;`

Cargo.toml
```diff
@@ -214,7 +215,9 @@ mongodb = { version = "3.7.0", features = ["rustls-tls"] }
 nix = { version = "0.31.3", features = ["feature", "fs", "resource", "sched"] }
 nonzero_lit = "0.1.2"
 notify = "8.2.0"
-octocrab = "0.53.1"
+octocrab = "0.51.0"
+once_cell = "1.21.4"
```
hubcio: once_cell was removed in #3524 and we prefer to keep it that way. its part of std in rust, so no need to include it again.

Cargo.toml
```diff
@@ -214,7 +215,9 @@ mongodb = { version = "3.7.0", features = ["rustls-tls"] }
 nix = { version = "0.31.3", features = ["feature", "fs", "resource", "sched"] }
 nonzero_lit = "0.1.2"
 notify = "8.2.0"
-octocrab = "0.53.1"
+octocrab = "0.51.0"
```
hubcio: please dont modify unrelated dependencies in this PR. PR is about connectors, not updating the deps. apply the same approach for `Cargo.lock` - dont modify unrelated files.

hubcio (changes requested):

Cargo.toml
ryerraguntla: once cell is indirect reference. Removed and fixed the dependencies.

Cargo.toml
ryerraguntla: I ended up using the previous version . Fixed now and sorry about that.

core/connectors/sources/opensearch_source/src/state_manager.rs
```diff
@@ -0,0 +1,314 @@
...
+            .truncate(true)
+            .open(&tmp_path)
+            .await
+            .map_err(|e| Error::Storage(format!("Failed to open state temp file: {e}")))?;
+        if let Err(e) = tmp_file.write_all(json.as_bytes()).await {
+            let _ = fs::remove_file(&tmp_path).await;
+            return Err(Error::Storage(format!(
+                "Failed to write state temp file: {e}"
+            )));
+        }
```
kriti-sc, line 171: should the file being written be given any permissions here?

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1135 @@
...
+            else {
+                warn!(
+                    connector_id = self.id,
+                    hit_id = hit.get("_id").and_then(|value| value.as_str()),
+                    "Skipping OpenSearch hit without sort tuple; document will not be published"
+                );
+                continue;
+            };
+
+            last_sort = Some(sort);
```
kriti-sc: correct me if i am wrong, but it appears that cursor advances before confirming if a doc is published. ideally, the cursor should advance only after the doc is published. this is an ordering issue.

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1135 @@
...
+
+            let Some(source) = hit.get("_source") else {
+                warn!(
+                    connector_id = self.id,
+                    hit_id = hit.get("_id").and_then(|v| v.as_str()),
+                    "Skipping OpenSearch hit without _source; document will not be published"
+                );
+                continue;
+            };
```
kriti-sc, line 514: can a document be empty in OpenSearch, or should the source be erroring out here?

core/connectors/sources/opensearch_source/src/lib.rs
```diff
@@ -0,0 +1,1135 @@
...
+        let timestamp_field = self.timestamp_field()?;
+        let batch_size = self.batch_size();
+        self.search_body_base = Some(json!({
+            "query": self.config.query.clone().unwrap_or_else(|| json!({ "match_all": {} })),
+            "size": batch_size,
+            "sort": [
+                { timestamp_field: { "order": "asc" } },
+                { "_id": { "order": "asc" } }
+            ]
+        }));
```
kriti-sc, line 759: is _id global unique? ie can two docs have the same id? if not, this is a possible data loss issue.

kriti-sc (changes requested):

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: Source could be empty, being an empty content, there is no value in sending for downstream applcations. Erroring out to flag as possible data missing.

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: _id is global unique.

core/connectors/sources/opensearch_source/src/state_manager.rs
ryerraguntla: Good catch. Let me borrow from core/connectors/runtime/src/state.rs (168-174) . It provides 0o600 and probably can give the same.

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: Another good one. A common issue across many reviewed connectors. Took care of it and adding test cases

ryerraguntla:
Tests Unrelated to PR are failing. Need to find the root cause and solve the flakiness.

core/connectors/sources/opensearch_source/src/lib.rs
pushp-ranja: recent version of opensearch allows disabling  _source field even when document contains indexed attributes (making it a valid document): https://docs.opensearch.org/latest/mappings/metadata-fields/source/#disabling-the-field, more so to reduce index size when vector embedding is also stored along. @ryerraguntla  Do you think it will affect the functionality here ?

core/connectors/sources/opensearch_source/src/lib.rs
kriti-sc: i am not sure if i understand -- how would the test case help?

ryerraguntla:
@hubcio @kriti-sc - The preChecks are failing (flaky) with java SDK test, which is completely unrelated to the PR related changes. [Pre-merge / Java • test / run (pull_request)](https://github.com/apache/iggy/actions/runs/28737059115/job/85213039410?pr=3515)     ---AsyncConsumerGroupsTest , shouldHandleConcurrentJoinAndLeaveOperations . Since it is not related to this PR, can I raise an Issue and a new  PR for fixing this failure. Please let me know if there are any other ways of handling.

core/connectors/sources/opensearch_source/src/lib.rs
ryerraguntla: how would the test case help - I am habituated with writing the use cases, test cases then start writing the code - TDD. Thinking of different data set scenarios and mix of data is one of the criteria in the Igyy connectors development. By writing test cases with mix of data so that cursor handles the bad data and continues with the expected behavior. This is the reason, my source contributions have a higher percentage of code coverage.

hubcio:
@ryerraguntla sure - please fix these togglers if you can. we're trying to push the clustering forward, thus less attention on currently open PRs.

PR #3525, opened 2026-06-21
fix(sdk): recover from InvalidOffset by falling back to first available offset

mfyuce:
## Problem

When a consumer group's stored offset falls below the topic's earliest
available offset — for example after the server purges old segments
under a retention policy — `poll_messages` returns
`IggyError::InvalidOffset`. The consumer retried at the same invalid
offset indefinitely, causing sink connectors to loop on errors and stop
delivering messages.

This manifests in production as:

```
ERROR iggy::clients::consumer: Failed to poll messages: Invalid offset: 0
ERROR iggy_connectors::sink: Failed to receive message for sink connector ...
# repeats forever; no messages are delivered
```

The root cause: new consumer groups receive stored offset 0 from the
server, but if the topic's retention policy has already purged messages
at offset 0 the very first poll returns `InvalidOffset(0)`, and the
consumer has no recovery path.

## Fix

Add `fallback_to_first: Arc<AtomicBool>` to `IggyConsumer`.

- On `IggyError::InvalidOffset`, set the flag and emit a `warn!` with
  stream/topic context.
- On the next `create_poll_messages_future` call, use
  `PollingStrategy::first()` instead of the configured strategy so the
  consumer seeks to the earliest available message.
- After the first successful non-empty poll, clear the flag so normal
  next-offset tracking resumes.

No public API changes. The existing `polling_strategy` field and all
builder methods are unchanged.

## Test

`cargo test -p iggy` passes (120 + 3 tests, 0 failed).

The fix was validated in a production `yucemonitoring` vcluster where
four iggy-connectors QuickWit sink connectors were stuck in the
`InvalidOffset(0)` loop against an `otel/metrics` topic with ~20 M
messages and retention. After deleting the stale consumer groups and
restarting (the current workaround), and confirming this fix would have
recovered automatically, ingestion resumed at ~2 700 docs/s.

### DIFF
--- a/core/sdk/src/clients/consumer.rs
+++ b/core/sdk/src/clients/consumer.rs
@@ -528,7 +528,7 @@ impl IggyConsumer {
             self.consumer
         );
 
-        Self::initialize_consumer_group(
+        let newly_created = Self::initialize_consumer_group(
             self.client.clone(),
             self.create_consumer_group_if_not_exists,
             self.stream_id.clone(),
@@ -537,7 +537,17 @@ impl IggyConsumer {
             &self.consumer_name,
             self.joined_consumer_group.clone(),
         )
-        .await
+        .await?;
+
+        // A brand-new consumer group starts at stored offset 0. If the topic has
+        // had retention run, offset 0 may no longer exist. Pre-arm the fallback
+        // so the very first poll uses PollingStrategy::first() instead of Next,
+        // avoiding a guaranteed InvalidOffset error on startup.
+        if newly_created {
+            self.fallback_to_first.store(true, ORDERING);
+        }
+
+        Ok(())
     }
 
     async fn subscribe_events(&self) {
@@ -558,6 +568,7 @@ impl IggyConsumer {
         let consumer_name = self.consumer_name.clone();
         let can_poll = self.can_poll.clone();
         let joined_consumer_group = self.joined_consumer_group.clone();
+        let fallback_to_first = self.fallback_to_first.clone();
         let mut reconnected = false;
         let mut disconnected = false;
 
@@ -615,7 +626,7 @@ impl IggyConsumer {
                         info!(
                             "Rejoining consumer group: {consumer_name} for stream: {stream_id}, topic: {topic_id}..."
                         );
-                        if let Err(error) = Self::initialize_consumer_group(
+                        match Self::initialize_consumer_group(
                             client.clone(),
                             create_consumer_group_if_not_exists,
                             stream_id.clone(),
@@ -626,10 +637,17 @@ impl IggyConsumer {
                         )
                         .await
                         {
-                            error!(
-                                "Failed to join consumer group: {consumer_name} for stream: {stream_id}, topic: {topic_id}. {error}"
-                            );
-                            continue;
+                            Err(error) => {
+                                error!(
+                                    "Failed to join consumer group: {consumer_name} for stream: {stream_id}, topic: {topic_id}. {error}"
+                                );
+                                continue;
+                            }
+                            Ok(newly_created) => {
+                                if newly_created {
+                                    fallback_to_first.store(true, ORDERING);
+                                }
+                            }
                         }
                         info!(
                             "Rejoined consumer group: {consumer_name} for stream: {stream_id}, topic: {topic_id}"
@@ -855,6 +873,10 @@ impl IggyConsumer {
         sleep(Duration::from_micros(remaining)).await;
     }
 
+    // Returns true if the consumer group was freshly created (stored offset = 0,
+    // which may be below the topic's earliest available offset after retention).
+    // Callers use this to pre-arm fallback_to_first so the very first poll uses
+    // PollingStrategy::first() and avoids an InvalidOffset error.
     async fn initialize_consumer_group(
         client: IggyRwLock<ClientWrapper>,
         create_consumer_group_if_not_exists: bool,
@@ -863,9 +885,9 @@ impl IggyConsumer {
         consumer: Arc<Consumer>,
         consumer_name: &str,
         joined_consumer_group: Arc<AtomicBool>,
-    ) -> Result<(), IggyError> {
+    ) -> Result<bool, IggyError> {
         if joined_consumer_group.load(ORDERING) {
-            return Ok(());
+            return Ok(false);
         }
 
         let client = client.read().await;
@@ -878,7 +900,7 @@ impl IggyConsumer {
         trace!(
             "Validating consumer group: {consumer_group_id} for topic: {topic_id}, stream: {stream_id}"
         );
-        if client
+        let newly_created = if client
             .get_consumer_group(&stream_id, &topic_id, &consumer_group_id)
             .await?
             .is_none()
@@ -908,7 +930,10 @@ impl IggyConsumer {
                     return Err(error);
                 }
             }
-        }
+            true
+        } else {
+            false
+        };
 
         info!(
             "Joining consumer group: {consumer_group_id} for topic: {topic_id}, stream: {stream_id}",
@@ -928,7 +953,7 @@ impl IggyConsumer {
         info!(
             "Joined consumer group: {consumer_group_id} for topic: {topic_id}, stream: {stream_id}"
         );
-        Ok(())
+        Ok(newly_created)
     }
 }
 
--- a/core/sdk/src/clients/consumer.rs
+++ b/core/sdk/src/clients/consumer.rs
@@ -111,6 +111,12 @@ pub struct IggyConsumer {
     topic_id: Arc<Identifier>,
     partition_id: Option<u32>,
     polling_strategy: PollingStrategy,
+    // Set to true when InvalidOffset is received so the next poll uses
+    // PollingStrategy::first() instead of the configured strategy, recovering
+    // from the case where the consumer group's stored offset falls below the
+    // topic's earliest available offset (e.g. after retention purges old
+    // segments). Cleared after the first successful recovery poll.
+    fallback_to_first: Arc<AtomicBool>,
     poll_interval_micros: u64,
     batch_length: u32,
     auto_commit: AutoCommit,
@@ -170,6 +176,7 @@ impl IggyConsumer {
             topic_id: Arc::new(topic_id),
             partition_id,
             polling_strategy,
+            fallback_to_first: Arc::new(AtomicBool::new(false)),
             poll_interval_micros: polling_interval.map_or(0, |interval| interval.as_micros()),
             last_stored_offsets: Arc::new(DashMap::new()),
             last_consumed_offsets: Arc::new(DashMap::new()),
@@ -645,7 +652,8 @@ impl IggyConsumer {
         let topic_id = self.topic_id.clone();
         let partition_id = self.partition_id;
         let consumer = self.consumer.clone();
-        let polling_strategy = self.polling_strategy;
+        let configured_strategy = self.polling_strategy;
+        let fallback_to_first = self.fallback_to_first.clone();
         let client = self.client.clone();
         let count = self.batch_length;
         let auto_commit_after_polling = self.auto_commit_after_polling;
@@ -676,6 +684,12 @@ impl IggyConsumer {
                 sleep(retry_interval.get_duration()).await;
             }
 
+            let polling_strategy = if fallback_to_first.load(ORDERING) {
+                PollingStrategy::first()
+            } else {
+                configured_strategy
+            };
+
             trace!("Sending poll messages request");
             last_polled_at.store(IggyTimestamp::now().into(), ORDERING);
             let polled_messages = client
@@ -696,6 +710,7 @@ impl IggyConsumer {
                 if polled_messages.messages.is_empty() {
                     return Ok(polled_messages);
                 }
+                fallback_to_first.store(false, ORDERING);
 
                 let partition_id = polled_messages.partition_id;
                 let consumed_offset;
@@ -782,6 +797,18 @@ impl IggyConsumer {
 
             let error = polled_messages.unwrap_err();
             error!("Failed to poll messages: {error}");
+
+            // When the consumer group's stored offset falls below the topic's
+            // earliest available offset (e.g. after retention removes old
+            // segments), seek to the first available message on the next poll
+            // instead of looping forever at the invalid offset.
+            if matches!(error, IggyError::InvalidOffset(_)) {
+                warn!(
+                    "Consumer offset is before the earliest available message in topic: {topic_id}, stream: {stream_id}. \
+                     Falling back to first available offset on next poll."
+                );
+                fallback_to_first.store(true, ORDERING);
+            }
 
             // Handle connection/auth errors - disable polling until event task re-enables
             // it after reconnection and rejoin complete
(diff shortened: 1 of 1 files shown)


core/connectors/sources/otlp_source/src/server.rs
```diff
@@ -0,0 +1,141 @@
...
+impl LogsService for LogsServiceImpl {
+    async fn export(
+        &self,
+        request: Request<ExportLogsServiceRequest>,
+    ) -> Result<Response<ExportLogsServiceResponse>, Status> {
+        let messages = convert::export_logs_to_messages(request.into_inner());
+        send_messages(&self.tx, messages, "logs").await;
+        Ok(Response::new(ExportLogsServiceResponse {
+            partial_success: None,
+        }))
```
atharvalade: This always returns success to the OTel caller even when messages are dropped
on backpressure. The OTLP spec has `partial_success` for exactly this case,
and without using it (or returning RESOURCE_EXHAUSTED), collectors/SDKs won't
retry and data is silently lost.

core/sdk/src/clients/consumer.rs
atharvalade, line 1: This flag is per-consumer, not per-partition. With consumer groups assigned
multiple partitions, an `InvalidOffset` on partition A would force
`PollingStrategy::first()` on partition B on the next poll, potentially
re-reading from the start of an unrelated partition. I think this should be
a` DashMap<u32, AtomicBool>` keyed by `partition_id` instead?

core/connectors/sources/otlp_source/src/server.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+async fn send_messages(
+    tx: &mpsc::Sender<ProducedMessage>,
+    messages: Vec<ProducedMessage>,
+    signal: &str,
+) {
+    for message in messages {
+        if let Err(err) = tx.try_send(message) {
```
atharvalade: Using `try_send` means if `poll()` is slow draining the channel, gRPC export calls
drop data without the sender knowing. Wouldn't `send().await `with a timeout and
returning `Status::RESOURCE_EXHAUSTED` be safer here? `OTel` clients handle that.

atharvalade (changes requested):
This PR ships at least 5 independent features/fixes in one PR (OTLP source, Dockerfile, `COOP_TASKRUN`, `quickwit_sink` rewrite, `InvalidOffset` fix). I think it will be better to split these? It's really hard to review and bisect if something goes wrong.

mfyuce:
Rebased onto upstream master -- PR now contains only the InvalidOffset fix (2 commits, 1 file: `core/sdk/src/clients/consumer.rs`).

Per your feedback on per-partition tracking: replaced `Arc<AtomicBool>` with `Arc<DashMap<u32, ()>>` keyed by `partition_id`. An InvalidOffset on partition A no longer forces `PollingStrategy::first()` on partitions B, C, etc. For auto-assign consumers (`partition_id = None`) `u32::MAX` is used as a sentinel; behavior for that case is equivalent to before.

The earlier comments about `partial_success` and `try_send` in `otlp_source/server.rs` belong to PR #3516 -- happy to address them there if you want to continue that review.

/ready

numinnex (approved):

numinnex (dismissed):

numinnex:
Looks like your comments @atharvalade got addressed, confirm if that's the case. LGTM.

mfyuce:
Updated with an additional improvement: instead of detecting `InvalidOffset` reactively on the first poll and recovering on the second, the fix now pre-arms the fallback when a consumer group is **freshly created**.

`initialize_consumer_group` returns `true` when it creates the group (stored offset = 0). The caller immediately sets `fallback_to_first = true` so the very first `poll_messages` call uses `PollingStrategy::First` (earliest available offset) rather than `Next` (stored offset = 0). This eliminates the guaranteed `InvalidOffset` error on startup that the previous version of the fix could not avoid.

Both the initial `init_consumer_group` path and the reconnect `subscribe_events` path are updated.

mfyuce:
Update: changed recovery strategy from `first()` to `last()`.

Investigated whether `PollingStrategy::first()` would work (it should map to `first_segment.start_offset` on the server, giving the first *available* message with no data skip). However, `PollingKind::First` consistently returns `InvalidOffset(0)` over the TCP binary protocol even when offset 0 is demonstrably valid (the same topic returns offset 0 correctly via the HTTP API).

Root cause of the TCP/HTTP discrepancy is TBD. Using `last()` instead: it always succeeds since the most recent message is always in an active segment, breaks the infinite error loop, and gets data flowing. The cost is skipping history that was buffered in Iggy but never consumed.

Will open a separate investigation for the `first()` TCP issue.

core/sdk/src/clients/consumer.rs
mfyuce: Addressed: `Arc<AtomicBool>` replaced with `Arc<DashMap<u32, ()>>` keyed by `partition_id`. InvalidOffset on partition A no longer touches other partitions. For auto-assign consumers (partition_id = None) the sentinel `u32::MAX` is used. The PR has been rebased and now contains only this single-file fix in `core/sdk/src/clients/consumer.rs`.

core/connectors/sources/otlp_source/src/server.rs
mfyuce: This code is no longer in this PR. The `otlp_source` was split into PR #3516, where both `partial_success` rejection counts and the `try_send` backpressure behavior have been addressed.

core/connectors/sources/otlp_source/src/server.rs
mfyuce: This code is no longer in this PR. The `otlp_source` was split into PR #3516, where both `try_send` behavior and `partial_success` rejection counts have been addressed.

atharvalade:
> Looks like your comments @atharvalade got addressed, confirm if that's the case. LGTM.

let me check

atharvalade:
@mfyuce Can you check failing CI before I review?

mfyuce:

The failing test is `server::purge_delete::should_delete_segments_with_consumer_group_barrier` in `core/integration/tests/server/scenarios/purge_delete_scenario.rs:480`:

```
assertion `left == right` failed: Expected message at offset 0
  left: 24
 right: 0
```

This is a server-level integration test about segment deletion with consumer group barriers. PR #3525 only touches `core/sdk/src/clients/consumer.rs` (per-partition `InvalidOffset` tracking in the Rust SDK client). There is no code path between the SDK consumer and the server's segment purge/delete logic.

The assertion failure (`offset 24 != 0` after a purge) points to a race in the server's segment cleanup under consumer group barrier conditions -- a pre-existing issue unrelated to this PR. Checking whether this test also fails on the current master would confirm it.

mfyuce:
@atharvalade The integration test failure has been fixed by removing the pre-arm offset logic and using a purely reactive recovery strategy. CI is now green. Ready for review!

core/sdk/src/clients/consumer.rs
```diff
@@ -111,6 +111,12 @@ pub struct IggyConsumer {
     topic_id: Arc<Identifier>,
     partition_id: Option<u32>,
     polling_strategy: PollingStrategy,
+    // Per-partition recovery set: when a partition's stored offset falls below
+    // its earliest available offset, its ID is inserted here so the next poll
+    // for that partition uses PollingStrategy::last(). Removed after the first
+    // successful recovery poll. Keyed by partition_id; u32::MAX is the sentinel
+    // for consumers with no fixed partition (consumer-group auto-assign).
+    fallback_to_last: Arc<DashMap<u32, ()>>,
```
hubcio, line 119: `effective_pid` is constant for the consumer's lifetime (`partition_id` is set once in the constructor and never reassigned), so this map holds at most one key - it's functionally a bool. the PR description says `Arc<AtomicBool>`, which is the right call: simpler, lock-free, and it removes the `u32::MAX` sentinel entirely (which today silently no-ops for a consumer group built with an explicit partition id, since the pre-arm inserts `u32::MAX` but the poll checks the real partition id). the 'per-partition' comment above doesn't match what the code can do.

core/sdk/src/clients/consumer.rs
```diff
@@ -530,7 +537,17 @@ impl IggyConsumer {
...
             self.joined_consumer_group.clone(),
         )
-        .await
+        .await?;
+
+        // A brand-new consumer group starts at stored offset 0. If the topic has
+        // had retention run, offset 0 may no longer exist. Pre-arm the fallback
+        // so the very first poll uses PollingStrategy::last() instead of Next,
+        // avoiding a guaranteed InvalidOffset error on startup.
+        if newly_created {
```
hubcio, line 546: the premise here doesn't hold on current master: `Next` with no stored offset starts from the first available segment (see the `PollingKind::Next` arm in `partitions/ops.rs`), it never returns `InvalidOffset`. so there's no startup error to avoid - but the pre-arm itself now forces every freshly created consumer group's first poll to `last()`, which silently skips the whole existing backlog and (with auto-commit) commits past it, permanently. that's a behavior change for every new consumer group, retention or not. the connector sinks use exactly this config (consumer group, no partition id, `next()`, auto-commit on poll, create-if-not-exists), so a fresh sink deploy against a topic with existing messages drops everything except the last batch. same pre-arm fires on the rejoin path below, and since the armed flag survives empty polls, a group created against an empty topic also skips history if more than a batch accumulates before its first non-empty poll. dropping the pre-arm entirely seems right - the server already handles the fresh-group case.

core/sdk/src/clients/consumer.rs
```diff
@@ -676,6 +702,13 @@ impl IggyConsumer {
                 sleep(retry_interval.get_duration()).await;
             }
 
+            let effective_pid = partition_id.unwrap_or(u32::MAX);
+            let polling_strategy = if fallback_to_last.contains_key(&effective_pid) {
```
hubcio, line 706: for a consumer group without a fixed partition, `effective_pid` is always `u32::MAX`, but the server picks a (potentially different) partition per poll via round-robin. the armed flag doesn't follow the partition that had the problem - the `last()` poll lands on whichever partition the server picks next, so on a multi-partition topic exactly one (nondeterministic) partition gets tail-skipped while the others start from earliest. inconsistent starting state within one group, and on servers that do return `InvalidOffset` from poll this can repeatedly skip healthy partitions.

core/sdk/src/clients/consumer.rs
```diff
@@ -676,6 +702,13 @@ impl IggyConsumer {
...
+                PollingStrategy::last()
```
hubcio, line 707: the PR title and description say fall back to the *first* available offset (`PollingStrategy::first()`), but the code uses `last()` - these are opposites. `first()` replays from the earliest retained message (at-least-once, nothing avoidable lost); `last()` jumps to the tail and drops the entire still-retained backlog. for the sink use case that's silent data loss. if the fallback stays it should be `first()`, and disarming on any non-error poll (rather than non-empty) avoids a re-arm loop when the tail was already consumed.

core/sdk/src/clients/consumer.rs
```diff
@@ -783,6 +817,18 @@ impl IggyConsumer {
             let error = polled_messages.unwrap_err();
             error!("Failed to poll messages: {error}");
 
+            // When the consumer group's stored offset falls below the topic's
+            // earliest available offset (e.g. after retention removes old
+            // segments), seek to the most recent message on the next poll
+            // instead of looping forever at the invalid offset.
+            if matches!(error, IggyError::InvalidOffset(_)) {
```
hubcio, line 824: on current master this branch is unreachable: the only place the server produces `InvalidOffset` is `validate_partition_offset`, which is only called from the store-consumer-offset path - the poll path skip-forwards or returns empty for out-of-range offsets instead of erroring, and the retention tests in `message_cleanup_scenario.rs` poll `offset(0)` after segments were deleted and `.unwrap()` successfully. so against a current server this recovery never fires and only the pre-arm behavior change remains. also worth re-checking the root cause: `Invalid offset: 0` matches the store-path rejection for an empty partition (`messages_count == 0 && current_offset == 0`), not a retention purge - the production loop was probably an older server or the store path. worth verifying against the actual released version before shipping a client-side workaround.

core/sdk/src/clients/consumer.rs
```diff
@@ -783,6 +817,18 @@ impl IggyConsumer {
...
+                warn!(
+                    "Consumer offset is before the earliest available message in topic: {topic_id}, stream: {stream_id}. \
```
hubcio, line 826: this states a cause the server can't produce - `InvalidOffset` fires for an offset beyond the end or an empty partition, never for 'before the earliest available' (the server skip-forwards that case). and 'falling back to latest' contradicts the PR title's 'first available offset'. worth making the text neutral about the cause.

core/sdk/src/clients/consumer.rs
```diff
@@ -880,7 +930,10 @@ impl IggyConsumer {
                     return Err(error);
                 }
             }
-        }
+            true
```
hubcio, line 933: if two consumers race to create the same group, the loser hits `ConsumerGroupNameAlreadyExists` above and still ends up with `newly_created = true` here - it didn't create anything. with the pre-arm this means both consumers arm the fallback. the `AlreadyExists` arm should yield `false`.

hubcio (changes requested):
thanks for the detailed report - the stuck-sink problem is real and worth fixing, but this particular fix works against you (details inline). the short version: on current master the scenario this PR targets cannot happen, and the part of the patch that does run changes delivery semantics for every new consumer group. here's how i'd approach it instead:

**1. root-cause first.** on current master, poll cannot return `InvalidOffset` at all - the only code path producing it is `validate_partition_offset`, which is reached exclusively from the store-consumer-offset command. its `Invalid offset: 0` signature matches storing an offset on an empty partition (`messages_count == 0 && current_offset == 0`), not a retention purge. so the production loop came from an older server version (or the store path), and that needs confirming against the exact deployed version before any client-side code is written - otherwise we're fixing a ghost.

**2. the server owns this invariant, and master already implements the fix.** "stored offset became invalid because the server deleted data" is server-created state, so the server heals it: poll with `next()` and a missing/stale offset skips forward to the earliest available message (see the `PollingKind::Next` arm in `partitions/ops.rs`), and the retention tests in `message_cleanup_scenario.rs` verify exactly the segments-deleted-then-poll case. if a released version you run still errors on poll, the right move is backporting that clamp - not a client-side workaround that silently seeks behind the application's back.

**3. if we want SDK-side recovery at all, it should be an explicit opt-in policy, kafka-style.** a builder option like `auto_offset_reset = error | earliest | latest`, defaulting to `error` (today's behavior). on `InvalidOffset` from poll: one recovery poll per the chosen policy, state keyed by the partition id from the actual poll response (not a shared sentinel), a plain `AtomicBool`/enum instead of a `DashMap`, disarmed on any non-error poll, and never pre-armed for new groups. it would also need to be specified cross-sdk - right now java/go/c#/node/c++ all propagate poll errors with no fallback (python wraps the rust sdk, so it would silently inherit whatever we do here), and a rust-only silent seek breaks behavioral parity. worth being honest that against current servers this code is dead on arrival, so it only pays off if we commit to supporting old servers from new sdks.

**4. the operational pain belongs in the connectors runtime.** the sink loop discards the error value (`let Ok(message) = message else` -> generic "failed to receive message") and hot-loops on persistent errors. propagating the concrete error and adding backoff on repeated identical failures fixes the actual symptom you hit - stuck sink, log spam, no diagnosis - for every error class, not just this one.

**5. optionally, kill the error class at the source:** the server could accept storing offset 0 on an empty partition as a no-op instead of rejecting it with `InvalidOffset`. small change, separate PR.

also: no tests cover the new path here. whatever shape this ends up taking needs at minimum a regression test that a brand-new consumer group on a topic with existing messages consumes from the earliest available offset, not the tail.

before implementing fix, please discuss that with us.

PR #3527, opened 2026-06-22
feat(server-ng): port consumer group

numinnex:
This PR ports the consumer group mechanism to server-ng, maintaining the cooperative rebalancing mechanism and extending it to accommodate for the fact that the server is distributed.


core/metadata/src/stm/consumer_group.rs
```diff
@@ -96,116 +147,367 @@ impl ConsumerGroup {
...
+                .ok_or_else(|| iggy_binary_protocol::WireError::UnexpectedEof {
+                    offset: at,
+                    need: 4,
+                    have: buf.len().saturating_sub(at),
+                })
+                .map(|slice| u32::from_le_bytes(slice.try_into().expect("4 bytes")))
+        };
+        let count = read_u32(buf, pos)? as usize;
+        pos += 4;
+        let mut in_flight = Vec::with_capacity(count);
```
hubcio: `Vec::with_capacity(count)` where `count` is an unbounded u32 read straight off the wire, before any per-element bounds check. a corrupt or bit-rotted replicated prepare with `count` near `u32::MAX` allocates ~16 GiB upfront (4 bytes per entry) and aborts every backup that applies it. `SyncConsumerGroupResponse::decode` already guards this with `capped_capacity` - mirror it here: `Vec::with_capacity(capped_capacity(count, buf.len().saturating_sub(pos), 4))`. the asymmetry with the sibling decoder is the tell it was an oversight.

core/common/src/traits/binary_impls/messages.rs
```diff
@@ -24,14 +24,212 @@ use crate::{
...
+/// (the VSR broker only routes explicit partitions, matching Kafka).
+#[cfg(feature = "vsr")]
+async fn resolve_partitioning<B: BinaryClient>(
+    client: &B,
+    stream_id: &Identifier,
+    topic_id: &Identifier,
+    partitioning: &Partitioning,
+) -> Result<Partitioning, IggyError> {
+    match partitioning.kind {
+        PartitioningKind::PartitionId => Ok(partitioning.clone()),
```
hubcio, line 150: `PartitionId => Ok(partitioning.clone())` clones the whole `Partitioning` including its `value: Vec<u8>` on the explicit-partition fast path, every send. it's a tiny clone next to the encode plus network round-trip on the same call, but avoidable - borrow the input for `kind == PartitionId` and only call `resolve_partitioning` for `Balanced`/`MessagesKey`.

core/common/src/traits/binary_impls/messages.rs
```diff
@@ -24,14 +24,212 @@ use crate::{
...
+                return Ok(polled);
+            }
+            Err(IggyError::ConsumerGroupPartitionNotOwned(..)) => {
+                client.consumer_group_state().invalidate_assignment(&key);
+                sync_group_assignment(client, stream_id, topic_id, &consumer.id).await?;
+            }
+            Err(error) => return Err(error),
+        }
+    }
+    Err(IggyError::ConsumerGroupPartitionNotOwned(0, 0))
```
hubcio: after `GROUP_POLL_MAX_ATTEMPTS` this returns `ConsumerGroupPartitionNotOwned(0, 0)` with fabricated (0,0) ids, and it reaches the public consumer stream unchanged. a rebalance during a poll is normal, so a CG app that doesn't special-case error 5009 sees a misleading hard error. the cursor is already re-synced by the time you get here, so return `Ok(PolledMessages::empty())` and let the caller re-poll. needs two back-to-back fences to reach, so it's rare, but the (0,0) is meaningless to the client either way.

core/metadata/src/stm/consumer_group.rs
```diff
@@ -220,177 +522,326 @@ impl StateHandler for CreateConsumerGroupRequest {
...
-                .iter()
-                .map(|(group_id, group)| {
-                    let members: Vec<(usize, ConsumerGroupMemberSnapshot)> = group
+impl StateHandler for RemoveConsumerGroupMemberRequest {
+    type State = StreamsInner;
+    fn apply(&self, state: &mut StreamsInner, _timestamp: iggy_common::IggyTimestamp) -> Bytes {
+        let mut any_removed = false;
+        for (_, stream) in &mut state.items {
+            for (_, topic) in &mut stream.topics {
+                let partition_ids: Vec<usize> = topic.partitions.iter().map(|p| p.id).collect();
```
hubcio: `partition_ids` is collected per topic before checking whether the topic actually holds the member. this runs on every logout inside the no-await commit critical section, on every replica. hoist the alloc into the member-found branch and skip topics whose group has no matching member.

core/server-ng/src/consumer_group.rs
```diff
@@ -0,0 +1,261 @@
...
+    let rewritten = match operation {
+        Operation::StoreConsumerOffset => {
+            let mut wire = StoreConsumerOffsetRequest::decode_from(body)
+                .map_err(|_| IggyError::InvalidCommand)?;
+            let Some(group_id) =
+                resolve_group_offset_id(shard, &wire.consumer, (&wire.stream_id, &wire.topic_id))
+            else {
+                return Ok(request);
+            };
+            wire.consumer.id = WireIdentifier::Numeric(group_id as u32);
```
hubcio: `next_consumer_group_id` is a never-reused u64 (the offset-key invariant stated in the module doc), but every wire crossing truncates it to u32 here and at 213/224/235. after 2^32 group creates on one topic the wire id wraps and could collide with a live group's offset key, breaking 'recreated group never inherits a dead offset'. astronomically unreachable, but the doc states the invariant as absolute - either note the u32 ceiling on `next_consumer_group_id` or just make it u32 since the partition plane key is already u32.

core/server-ng/src/bootstrap.rs
```diff
@@ -895,11 +894,41 @@ async fn shard_main(
...
             tick_shard.tick_partitions().await;
+            // While a cooperative revocation is pending, wake the reconciler each
+            // tick so it completes the handoff within ~one tick of the source
+            // draining the partition, instead of waiting for the 1s periodic pass.
+            if tick_shard
+                .plane
+                .metadata()
+                .mux_stm
+                .streams()
+                .has_pending_revocations()
```
hubcio, line 905: `has_pending_revocations()` runs every consensus tick (10ms) on every shard. it short-circuits when nothing is pending, but while a revocation is pending the reconciler then walks streams/topics/groups/members with no short-circuit every tick, including on shards that can't act on it (only the partition-owner shard completes the handoff, and the owner can be any shard so you can't gate this to shard 0). a cheap atomic pending-count bumped on rebalance and decremented on completion avoids the per-tick walk.

core/server-ng/src/consumer_group.rs
```diff
@@ -0,0 +1,261 @@
...
+    };
+
+    rewrite_request_body(&request, &rewritten)
+}
+
+/// Gather the group's in-flight partitions (`last_polled` present and
+/// `committed < last_polled`) for the cooperative-rebalance classification. A
+/// not-yet-created group, an unresolved topic, or a partition that does not
+/// answer is treated as not-in-flight (eager handoff), which the reconciler
+/// corrects on a later pass if it was wrong.
```
hubcio: this says an eager handoff 'corrects on a later pass if it was wrong', but `reconcile_pending_revocations` only revisits entries that have a `PendingRevocation` record. an eager handoff leaves no such record, so it's never revisited within the generation - the claim is false for that path. the reconciler only completes pending revocations. drop the clause or scope it to the cooperative path. (a wrong eager handoff just redelivers, which is fine under at-least-once, so this is a doc fix.)

core/metadata/src/stm/consumer_group.rs
```diff
@@ -31,63 +41,104 @@ use iggy_binary_protocol::responses::consumer_groups::get_consumer_group::Consum
...
+    /// Reads the live partition set each call, so it reflects repartitions and
+    /// bumps the generation on any membership change.
+    pub fn rebalance_members(&mut self, partition_ids: &[usize]) {
+        self.generation += 1;
 
+        let member_count = self.members.len();
         let member_keys: Vec<usize> = self.members.iter().map(|(id, _)| id).collect();
         for &member_id in &member_keys {
             if let Some(member) = self.members.get_mut(member_id) {
                 member.partitions.clear();
```
hubcio, line 152: `rebalance_members` clears partitions and `pending_revocations` with no cooperative drain, and it runs on leave (609), remove-member (657) and repartition. asymmetric with the join path which drains in-flight partitions via pending revocations. a member leaving drops its in-flight partitions immediately, so the new owner re-reads the old owner's polled-but-uncommitted range - the redelivery the join path was built to avoid. a disconnected member can't drain anyway, so this is mostly intentional and it's covered by a test - worth one comment noting leave/join are deliberately asymmetric.

core/metadata/src/stm/stream.rs
```diff
@@ -339,6 +739,56 @@ impl Streams {
...
+            let stream_id = inner.resolve_stream_id(stream_id)?;
+            let topic_id = inner.resolve_topic_id(stream_id, topic_id)?;
+            let topic = inner.items.get(stream_id)?.topics.get(topic_id)?;
+            let count = topic.partitions.len();
+            if count == 0 {
+                return None;
+            }
+            let current = topic
+                .round_robin_counter
+                .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |c| {
```
hubcio, line 792: `fetch_update` mutates `round_robin_counter` (an `Arc<AtomicUsize>` that gets captured into the persisted snapshot) inside a `read()` closure, outside the replicated apply path. each replica advances it independently so the snapshot field diverges per-replica. dormant under VSR since the client resolves `Balanced` to an explicit partition before sending, but a non-vsr build or a foreign SDK that doesn't do client-side resolution hits this server arm. move the advance into the apply path, or drop the counter from the snapshot.

core/common/src/traits/binary_impls/messages.rs
```diff
@@ -24,14 +24,212 @@ use crate::{
...
+        .send_raw_with_response(SYNC_CONSUMER_GROUP_CODE, request.to_bytes())
+        .await?;
+    let key = group_cache_key(stream_id, topic_id, group_id);
+    client.consumer_group_state().register_group(
+        key.clone(),
+        stream_id.clone(),
+        topic_id.clone(),
+        group_id.clone(),
+    );
+    if response.is_empty() {
```
hubcio, line 79: `sync_group_assignment` calls `register_group` at line 79 before this `is_empty()` check, so a group the client isn't a member of (empty assignment reply) still gets permanently inserted into `joined_groups`. nothing deregisters on an empty sync - only explicit leave does - and `has_assignment` is false for an empty assignment, so every later poll and every heartbeat re-issues a `SyncConsumerGroup` round-trip for that non-member group for the connection's lifetime, and the map grows per distinct non-member group polled. only register after a non-empty reply.

core/server-ng/src/dispatch.rs
```diff
@@ -980,23 +1246,60 @@ fn empty_polled_messages_body(partition_id: u32) -> Bytes {
...
+    let strategy = polling_strategy_from_wire(&wire.strategy)?;
+    let args = PollingArgs::new(strategy, wire.count, wire.auto_commit);
+
+    // Consumer-group poll: the client selects which of its assigned partitions
+    // to read and sends it explicitly. The coordinator FENCES ownership (a stale
+    // client whose partition was reassigned is rejected with
+    // `ConsumerGroupPartitionNotOwned`, prompting a re-sync) and resolves the
+    // group's monotonic id -- the offset key the store rewrite and read path
+    // both use, so `next()` reads back the offset it just committed.
+    if wire.consumer.kind == 2 {
```
hubcio: raw `consumer.kind == 2` here and at 1317, plus `!= 2` at responses.rs:193 and consumer_group.rs:252, and a `2 =>` arm at 1346. `KIND_CONSUMER_GROUP: u8 = 2` already exists in binary_protocol's `primitives/consumer.rs` - make it `pub const` and reference it everywhere.

core/common/src/lib.rs
```diff
@@ -28,6 +30,17 @@ pub use error::iggy_error::{IggyError, IggyErrorDiscriminants};
...
+pub use consumer_group_client_state::ConsumerGroupClientState;
+
+/// Sentinel `partition_id` in an otherwise-empty poll reply that tells the
+/// client its cached consumer-group assignment is stale (a generation fence
+/// rejected the partition it polled) and it must re-sync before retrying. The
+/// VSR reply header carries no status field yet, so this rides the empty-poll
+/// body -- an application payload, not a wire-format header. Never a real
+/// partition id (those are small, dense, zero-based), so it can't collide with
+/// a genuine end-of-partition empty poll, which echoes the real partition id.
+pub const RESYNC_REQUIRED_PARTITION_SENTINEL: u32 = u32::MAX;
```
hubcio, line 43: `RESYNC_REQUIRED_PARTITION_SENTINEL = u32::MAX` is an out-of-band wire contract riding the empty-poll body since the reply header has no status field yet. collision is impossible by construction and it's documented, so fine for now - just worth migrating to a typed reply-header status field when one lands.

core/metadata/src/stm/consumer_group.rs
```diff
@@ -96,116 +147,367 @@ impl ConsumerGroup {
...
+            + self.in_flight.len() * 4
+    }
 
-        state.name_index.insert(name, id);
+    fn encode(&self, buf: &mut BytesMut) {
+        self.stream_id.encode(buf);
+        self.topic_id.encode(buf);
+        self.group_id.encode(buf);
+        buf.put_u128_le(self.client_id);
+        buf.put_u32_le(u32::try_from(self.in_flight.len()).unwrap_or(u32::MAX));
```
hubcio: `u32::try_from(self.in_flight.len()).unwrap_or(u32::MAX)` clamps the count, but `encoded_size` at line 380 uses the true length and the loop below writes every entry. if the clamp ever fired, the written count header and the body would disagree and the decoder would under-read and desync. it can't fire (in_flight is bounded by partition count), so `expect("in_flight count fits u32")` documents that invariant better than a silent clamp.

core/server-ng/src/consumer_group.rs
```diff
@@ -0,0 +1,261 @@
...
+        Operation::StoreConsumerOffset
+            | Operation::StoreConsumerOffset2
+            | Operation::DeleteConsumerOffset
+            | Operation::DeleteConsumerOffset2
+    ) {
+        return Ok(request);
+    }
+    let body = request_body(&request);
+    let rewritten = match operation {
+        Operation::StoreConsumerOffset => {
```
hubcio: these 4 arms (Store/Store2/Delete/Delete2) are near-identical, differing only in the decode type - a macro or small helper trait over the 4 types would collapse them. note the `_ => return Ok(request)` at 238 can't actually be deleted: the match is over the 37-variant `Operation` enum so it needs a catch-all even though the outer `matches!` already filters the op set.

core/server-ng/src/consumer_group.rs
```diff
@@ -0,0 +1,261 @@
...
+        else {
+            continue;
+        };
+        if committed.is_some_and(|c| c >= polled) {
+            continue;
+        }
+        if assigned.contains(&partition_id) {
+            in_flight.push(partition_id);
+        } else if let Ok(ns) =
+            resolve_partition_namespace(shard, stream_id, topic_id, Some(partition_id))
```
hubcio: `resolve_partition_namespace` is called here with the same args already resolved at line 129 in this same loop iteration. `IggyNamespace` is `Copy` and the first result is still live, so reuse it instead of re-resolving.

core/metadata/src/stm/consumer_group.rs
```diff
@@ -96,116 +147,367 @@ impl ConsumerGroup {
...
+                topic_id,
+                group_id,
+                client_id,
+            },
+            pos,
+        ))
+    }
+}
+
+fn read_u128_le(buf: &[u8], pos: usize) -> Result<u128, iggy_binary_protocol::WireError> {
```
hubcio: this hand-rolled `read_u128_le`, the local `read_u32` closure at 404, and the inline `buf.get(pos..pos+N)` + `from_le_bytes` blocks duplicate `read_u32_le`/`read_u64_le`/`read_u128_le` already public in binary_protocol's `codec.rs`, which return the identical `WireError::UnexpectedEof`. the file already imports from `codec`, so extend the import and drop the local copies.

core/metadata/src/stm/consumer_group.rs
```diff
@@ -31,63 +41,104 @@ use iggy_binary_protocol::responses::consumer_groups::get_consumer_group::Consum
...
-            partition_index: Arc::new(AtomicUsize::new(0)),
+            pending_revocations: Vec::new(),
         }
     }
+
+    /// Partitions this member should poll: owned minus those pending handoff
+    /// (the source stops polling a revoked partition so its consumer can drain
+    /// + commit, but it still owns it for the commit fence until completion).
+    #[must_use]
+    pub fn pollable_partitions(&self) -> Vec<usize> {
```
hubcio, line 86: `pollable_partitions` allocates a `HashSet` plus a `Vec` every call, and it's hit on every consumer-group poll under the metadata read lock just for a `.contains()` membership test. an alloc-free `is_pollable(partition_id)` predicate avoids both allocations on that path; the assignment caller that genuinely needs the Vec can keep building it but doesn't need the HashSet since `pending` is tiny.

hubcio (changes requested):
one finding lands on code outside this PR's diff so it can't be a line comment:

in `core/binary_protocol/src/consensus/operation.rs`, the `command_code_roundtrip` test's `ops` array skips from `DeletePersonalAccessToken` straight to `SendMessages`, omitting `Operation::JoinConsumerGroup`/`LeaveConsumerGroup`. this PR flips both from non-replicated to replicated-with-code, so their op<->code mapping in `to_command_code`/`from_command_code` is now live but unexercised by that test. add both variants (the `lookup_by_operation_roundtrips_with_lookup_command` array has the same omission).

core/metadata/src/stm/consumer_group.rs
```diff
@@ -96,116 +167,355 @@ impl ConsumerGroup {
...
+    #[allow(clippy::cast_possible_truncation)]
+    fn apply(&self, state: &mut StreamsInner, _timestamp: iggy_common::IggyTimestamp) -> Bytes {
+        let Some(topic) = state.topic_mut(&self.stream_id, &self.topic_id) else {
+            return Bytes::new();
+        };
+        let name: Arc<str> = Arc::from(self.name.as_str());
+        // Per-(stream,topic) name uniqueness: supersede an existing same-name
+        // group in this topic (idempotent create across a delete+recreate).
+        if let Some(stale_id) = topic.consumer_group_index.get(&name).copied() {
+            topic.consumer_groups.remove(&stale_id);
```
hubcio: this silently destroys a live consumer group on a duplicate-name create. `remove(&stale_id)` drops the existing group plus every joined member and their assignments, then inserts a fresh empty group with a new id. ejected members never recover: they re-sync, get an empty assignment (they're not a member of the new group), and `sync_group_assignment`'s empty branch just caches it without re-joining - poll then returns ok+empty with no error, so the consumer silently stalls. the new id also orphans the old group's committed offsets (the reconciler reclaims them), so a same-name recreate restarts from 0.

this regresses both prior behaviors: legacy server rejects with `ConsumerGroupNameAlreadyExists`, and the previous server-ng stm no-op-preserved on a duplicate name. the `idempotent create across delete+recreate` comment is wrong - a bare create with no preceding delete is destructive. fix: reject the duplicate (legacy parity) or no-op-return the existing group.

side note for the same branch: it also skips the `state.revision` bump that `DeleteConsumerGroup` does, so even the orphaned-offset reclaim is delayed until an unrelated commit bumps the revision.

core/server-ng/src/consumer_group.rs
```diff
@@ -0,0 +1,246 @@
...
+                resolve_group_offset_id(shard, &wire.consumer, (&wire.stream_id, &wire.topic_id))
+            else {
+                return Ok(request);
+            };
+            // The partition-plane group-offset key is u32, so the monotonic id
+            // must stay within u32 (see `Topic::next_consumer_group_id`). Fail
+            // loudly at that ceiling instead of silently wrapping into a live
+            // group's key.
+            wire.consumer.id = WireIdentifier::Numeric(
+                u32::try_from(group_id).expect("consumer-group id exceeds u32 offset-key ceiling"),
```
hubcio: `expect` on a live client-driven offset path panics the shard task (drops the connection) if a topic ever crosses ~4 billion group creates. `iggy_partition.rs` clamps the same cast with `unwrap_or(u32::MAX)`, so the two sites disagree on what to do. return an `IggyError` here instead of panicking.

core/common/src/consumer_group_client_state.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// True if a non-empty assignment is cached for the group.
+    #[must_use]
+    pub fn has_assignment(&self, key: &str) -> bool {
+        self.assignments
+            .lock()
+            .expect("consumer-group state mutex poisoned")
```
hubcio: `.expect("...mutex poisoned")` here and on every other lock in this file (10 sites total) aborts the sdk client process if any lock is ever poisoned. this is a client library so a hard abort is harsh, and the cached assignment is advisory anyway (re-synced on a fence). recover the guard with `.lock().unwrap_or_else(PoisonError::into_inner)` instead.

core/server-ng/src/consumer_group.rs
```diff
@@ -0,0 +1,246 @@
...
+    };
+    // Partitions a live member currently owns. A `last_polled` past the commit
+    // only means in-flight work when a live member still holds the partition;
+    // for an unowned one it is the residue of a member removed on disconnect
+    // (the reconnect case), which must be reassigned and re-read, not protected.
+    let assigned = streams
+        .consumer_group_assigned_partitions(stream_id, topic_id, group_id)
+        .unwrap_or_default();
+    let mut in_flight = Vec::new();
+    for partition_id in partition_ids {
```
hubcio: the per-partition `partition_read` calls in this loop are awaited sequentially, so a wide-topic join does n cross-shard round-trips back to back, and the join op can't be proposed until they all finish. the reads are independent and `ns` is `Copy`, so `join_all` them to collapse the wall-clock.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -563,6 +576,134 @@ async fn tear_down_owned_partition(
...
+/// transiently-failed revocation is retried next pass.
+#[allow(clippy::cast_possible_truncation)]
+fn reconcile_pending_revocations(ctx: &ReconcilerCtx) {
+    let pending = ctx
+        .shard
+        .plane
+        .metadata()
+        .mux_stm
+        .streams()
+        .consumer_group_pending_revocations();
```
hubcio: `consumer_group_pending_revocations()` walks every stream/topic/group/member and allocates a vec on every reconcile tick, before the fast-skip, even when nothing is pending. there's already an o(1) counter behind `has_pending_revocations()` - gate this call on it first so the common idle case stays cheap.

core/metadata/src/stm/stream.rs
```diff
@@ -263,10 +336,39 @@ collect_handlers! {
...
     }
 }
 
 impl StreamsInner {
+    /// Per-commit hook (see the `define_state!`/`dispatch` macro). Recomputes
+    /// `pending_revocations_count` so the consensus tick's
+    /// `has_pending_revocations` read is O(1) instead of walking every group
+    /// each 10ms. The walk runs only here -- once per metadata commit, far
+    /// rarer than the tick -- and over the small consumer-group structure.
+    fn post_apply(&mut self) {
```
hubcio: `post_apply` recomputes `pending_revocations_count` by walking every member on every streams commit, and it runs for both left-right copies, so every create-stream/topic/partition pays the walk even on clusters with zero consumer groups. maintain the count incrementally in the handlers that change it instead of a full recompute.

core/metadata/src/impls/metadata.rs
```diff
@@ -1079,6 +1081,88 @@ where
...
+                    MetadataSubmitError::NotCaughtUp
+                } else {
+                    MetadataSubmitError::NotPrimary
+                },
+            );
+        }
+        if consensus
+            .pipeline()
+            .borrow()
+            .has_message_from_client(INTERNAL_CLIENT_ID)
```
hubcio: all internal revocation completions share `INTERNAL_CLIENT_ID = u128::MAX`, so this pipeline dedup caps internal completions at one in flight across the whole cluster. a wide rebalance then drains its pending revocations one per consensus round, serially. correctness is fine (idempotent + retried) but convergence is slow - give each revocation its own internal id to lift the cap.

core/metadata/src/stm/consumer_group.rs
```diff
@@ -260,137 +752,95 @@ impl StateHandler for DeleteConsumerGroupRequest {
...
 pub struct ConsumerGroupMemberSnapshot {
     pub id: usize,
-    pub client_id: u32,
+    pub client_id: u128,
     pub partitions: Vec<usize>,
-    pub partition_index: usize,
+    /// `(partition_id, target_member_slab, created_at_micros)` per pending
+    /// cooperative revocation.
+    ///
+    /// Defaulted for snapshots written before cooperative rebalancing existed.
```
hubcio: this comment (and the `serde(default)` back-compat framing on these snapshot fields) misrepresents the mechanism. the snapshot codec is `rmp_serde::to_vec` with no struct-map, so it's positional arrays not name-keyed - `serde(default)` only fills missing trailing elements, it can't rescue a removed or reordered field, and `TopicSnapshot` dropped `round_robin_counter`. these types are also new, so there's no prior version that 'wrote snapshots before cooperative rebalancing existed'. harmless today (server-ng snapshot has no real on-disk format contract yet) but the comment should be reworded to say the format isn't back-compatible across this change.

core/integration/tests/server/mod.rs
```diff
@@ -18,9 +18,12 @@
...
-// Consumer groups are not implemented under vsr yet.
+// Polling-based consumer-group scenarios are not implemented under vsr yet.
 #[cfg(not(feature = "vsr"))]
 mod cg;
+// The ported round-robin membership join scenario runs against server-ng.
+#[cfg(feature = "vsr")]
+mod cg_vsr;
 // 80-case race matrix with hardcoded HTTP variants (test_matrix bypasses
 // the harness transport filter); revisit under vsr once basics are green.
 #[cfg(not(feature = "vsr"))]
```
hubcio, line 29: the `concurrent_addition` matrix (the consumer-group hot/cold same-name cases) is gated off under vsr, and nothing on the vsr side covers creating a consumer group with a duplicate name - which is exactly the path the create-consumer-group supersede breaks. once that apply is fixed to reject-or-preserve, please enable the consumer_group hot/cold cases here under vsr (at least the tcp variant, since the hardcoded http matrix entries don't apply - vsr has no http) so the dedup behavior is actually asserted against server-ng.

hubcio (changes requested):

numinnex:
I cannot enable the hot/cold consumer group creation tests under `vsr` feature flag, as it relies on error messages being propagated from the server which are introduced in #3553, after it's merged I will create another PR that enables those + some other tests that are blocked by it.

hubcio (approved):

thepatrykk (approved):

spetz (approved):

PR #3529, opened 2026-06-22
feat(connectors/otlp_sink): add OTLP/gRPC sink connector

### LINKED ISSUE #3526: feat: OTLP Sink Connector
### Description

Add a native OTLP (OpenTelemetry Protocol) sink connector for Apache Iggy using the opentelemetry-otlp Rust crate.

This connector will enable streams stored in Iggy to be exported to OpenTelemetry-compatible observability platforms such as Observe, Datadog, New Relic, Grafana Cloud, Honeycomb, Elastic, Splunk, and any backend supporting OTLP over gRPC or HTTP.

The OTLP sink is a foundational building block for positioning Iggy as a high-performance ingestion, buffering, replay, and routing layer for observability, AI telemetry, and event-driven architectures.

Connector Type: Sink Connector
---
Iggy Stream
    ↓
OTLP Sink
    ↓
OTLP Endpoint
---

Initial version:
- Logs

Future enhancements:
- Metrics
- Traces

===
Metrics
----
Expose connector metrics:
- otlp_sink_messages_sent_total
- otlp_sink_messages_failed_total
- otlp_sink_batches_sent_total
- otlp_sink_batch_latency_ms
- otlp_sink_retry_count_total

### Affected area / component

Connectors

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Contribution

### Good first issue

mfyuce:
## Summary

Closes #3526

Adds a new `iggy_connector_otlp_sink` plugin that reads messages from an Iggy topic and forwards them to any OTLP/gRPC-compatible backend (Quickwit, Jaeger, Tempo, Grafana Tempo, etc.).

- **Two storage modes**: `json` (default) reconstructs OTLP proto from the JSON produced by `otlp_source`; `proto` forwards raw prost-encoded bytes with zero deserialization overhead
- **Per-request gRPC metadata headers** for backend routing (e.g. Quickwit's `qw-otel-traces-index` / `qw-otel-logs-index` headers to select a target index)
- **Optional gzip compression** via tonic's built-in `CompressionEncoding::Gzip`
- Supports all three OTLP signals: `traces`, `metrics`, `logs`
- JSON-to-proto round-trip tested with 7 unit tests (`traces_from_json`, `logs_from_json`, `metrics_from_json`, `hex_to_bytes`, etc.)

## Motivation

The `otlp_source` connector stores OTLP telemetry in Iggy. This sink closes the loop, allowing Iggy to act as a durable buffer between an OTel Collector and any OTLP-capable backend -- replaying, fan-out, or delay-tolerant forwarding without a direct Collector-to-backend connection.

## Example config

```toml
type = "sink"
key  = "traces_to_quickwit"
name = "OTLP traces -> Quickwit"
path = "/connectors/plugins/libiggy_connector_otlp_sink"

[[streams]]
stream         = "otel"
topics         = ["traces"]
consumer_group = "qw_traces_sink"

[plugin_config]
endpoint = "http://quickwit:7281"
signal   = "traces"

[plugin_config.headers]
"qw-otel-traces-index" = "my_traces_index"
```

## Test plan

- [x] Unit tests pass: `cargo test -p iggy_connector_otlp_sink`
- [x] Clippy clean: `cargo clippy -p iggy_connector_otlp_sink --all-features -- -D warnings`
- [x] Apache 2.0 license header on all new source files
- [x] `#[serde(default)]` on all optional config fields (forward-compatible)
- [x] No `unwrap()`/`expect()` on I/O paths
- [ ] Integration test with a live Quickwit instance (manual, not in CI)


ryerraguntla:
@mfyuce - I did not see Issue #  this PR is closing. How far this PR meets - https://github.com/apache/iggy/issues/3526 requirements?

mfyuce:
@ryerraguntla Apologies for the missing link -- this closes #3526. Updated the PR description.

**Coverage vs the issue requirements:**

| Requirement | Status |
|---|---|
| Logs | Done |
| Metrics | Done (extended beyond initial scope) |
| Traces | Done (extended beyond initial scope) |
| gRPC transport | Done |
| HTTP transport | Not in this PR |
| `opentelemetry-otlp` crate | Used `tonic` directly (see below) |
| Connector metrics (`otlp_sink_*_total`) | Not in this PR |

**On crate choice:** the issue mentions `opentelemetry-otlp`. We use `tonic` directly against the generated stubs from the `opentelemetry-proto` crate instead. The reason: `opentelemetry-otlp` pulls in the full OpenTelemetry SDK (tracer/meter providers, exporters) -- but this connector only _forwards_ existing OTLP data, it does not _generate_ telemetry. Using `tonic` keeps the dependency footprint small and avoids initializing SDK internals that would go unused. The wire format is identical.

**Not in this PR:** HTTP transport and per-connector metrics counters. Happy to track those as follow-ups if the overall approach looks good.

Cargo.toml
ryerraguntla: tonic = { workspace = true } with no tonic entry in root [workspace.dependencies]. Build fails. **Fix**: add tonic = { version = "...", features = ["transport", "gzip"] } to root  [workspace.dependencies].

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+            ),
+        }
+    }
+    out
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
ryerraguntla, line 784: test_config() struct literal missing headers field; OtlpSinkConfig has no Default derive. Compile error in #[cfg(test)]. **Fix**: add headers: HashMap::new().

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,447 @@
...
+            v.get("severity").and_then(Value::as_str).unwrap_or(""),
+        ),
+        severity_text: v
+            .get("severity_text")
+            .and_then(Value::as_str)
+            .unwrap_or("")
+            .to_owned(),
+        body: v
+            .get("body")
+            .map(|b| AnyValue {
```
ryerraguntla: b.to_string() on serde_json::Value calls Display, not the inner string. Value::String("hello").to_string() == "\"hello\"". Every log body exported with extra JSON quotes. Test at  line 431-446 only asserts severity_number, not body — bug undetected. **Fix**: b.as_str().map(str::to_owned).unwrap_or_else(|| b.to_string()).

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+            _ => warn!(
+                "OTLP sink (proto mode): expected raw payload, got schema: {}",
+                meta.schema
+            ),
+        }
+    }
+    out
+}
+
+fn collect_json_values(meta: &MessagesMetadata, messages: &[ConsumedMessage]) -> Vec<serde_json::Value> {
```
ryerraguntla: collect_json_values hot path: simd_json::OwnedValue → simd_json::to_string() → String → serde_json::from_str(). Two full tree walks + two heap allocs per message on every consume(). 
  SDK exports iggy_connector_sdk::convert::owned_value_to_serde_json() for direct structural mapping. **Fix:** use the SDK function; drop the string round-trip.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,401 @@
...
+    match format {
+        StorageFormat::Proto => {
+            let bytes = collect_raw_bytes(meta, messages);
+            if bytes.is_empty() {
+                return Ok(ExportTraceServiceRequest::default());
+            }
+            // Proto mode: each message is a full ExportTraceServiceRequest; merge them.
+            let mut merged = ExportTraceServiceRequest::default();
+            for b in bytes {
+                match ExportTraceServiceRequest::decode(bytes::Bytes::copy_from_slice(b)) {
```
ryerraguntla, line 643: Three Places - 276,300,324 — proto decode failure silently drops message with warn! only; consume() returns Ok(()) with partial batch exported. Caller has no indication data was lost. **Fix:** log per-batch drop
  count; return Err if all messages fail decode.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,447 @@
...
+                    })
+                    .collect(),
+            })
+        }
+        Value::Null => any_value::Value::StringValue(String::new()),
+    };
+    AnyValue { value: Some(inner) }
+}
+
+fn hex_to_bytes(s: &str) -> Vec<u8> {
```
ryerraguntla, line 363: hex_to_bytes: odd-length input silently drops last nibble; invalid hex chars silently filtered. No length validation on result. trace_id spec requires 16 bytes, span_id 8  bytes; wrong-length output corrupts trace correlation at the backend. **Fix:** validate s.len() % 2 == 0 and assert result length; warn! on violation.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,447 @@
...
+
+fn status_code_from_text(s: &str) -> i32 {
+    match s {
+        "ok" => 1,
+        "error" => 2,
+        _ => 0, // unset
+    }
+}
+
+fn severity_from_text(s: &str) -> i32 {
```
ryerraguntla, line 401: severity_from_text case-sensitive. "warn", "info", "error", "debug" (lowercase) all map to 0 (Unset). Real-world OTel JSON emits lowercase severity. **Fix:** s.to_ascii_uppercase()  before match.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,447 @@
...
+        "sum" => {
+            let dp = NumberDataPoint {
+                attributes: attrs,
+                time_unix_nano: time_ns,
+                value: Some(number_data_point::Value::AsDouble(as_double)),
+                ..Default::default()
+            };
+            Some(metric::Data::Sum(Sum {
+                data_points: vec![dp],
+                aggregation_temporality: 2, // CUMULATIVE
```
ryerraguntla: aggregation_temporality: 2 is a magic number (should be AggregationTemporality::Cumulative as i32). is_monotonic: false hardcoded — counters (request counts, bytes transferred)
  are monotonic; misclassified to all backends. **Fix:** use the proto enum variant; add is_monotonic field to JSON schema or default true for sum type.

core/connectors/sinks/otlp_sink/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+publish = false
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+bytes = { workspace = true }
+iggy_connector_sdk = { workspace = true }
+opentelemetry-proto = { version = "0.32.0", default-features = false, features = [
```
ryerraguntla: opentelemetry-proto = { version = "0.32.0" } pinned inline; not in [workspace.dependencies]. All other opentelemetry-* crates are workspace-managed at 0.32.0. Independent pin risks version drift on workspace-wide bumps. **Fix:** add to root [workspace.dependencies], use { workspace = true }.

Dockerfile.connectors
```diff
@@ -0,0 +1,61 @@
...
+COPY --from=build /usr/local/lib/libiggy_connector_otlp_sink.so \
+                                                           /connectors/plugins/libiggy_connector_otlp_sink.so
+
+USER iggy
+WORKDIR /connectors
+
+# Runtime config: IGGY_CONNECTORS_CONFIG_PATH → /connectors/runtime.toml (ConfigMap)
+# Plugin configs: /connectors/plugins/otlp_source.toml (ConfigMap)
+ENV IGGY_CONNECTORS_CONFIG_PATH=/connectors/runtime.toml
+
```
ryerraguntla: EXPOSE 4317. Port 4317 is the OTLP/gRPC inbound receiver port. This container is an egress sink — it connects out to a backend, binds no ingress port. Misleads operators   into opening incorrect firewall rules. **Fix:** remove EXPOSE 4317.

mfyuce:
Updated in the latest push to address the two open items:

**HTTP transport** (`transport = "http"`): POSTs prost-encoded proto bytes to `{endpoint}/v1/{traces,metrics,logs}` with `Content-Type: application/x-protobuf`. The same `headers` and `compression` (gzip body via `Content-Encoding: gzip`) config fields work for both transports. 4xx responses are logged as warnings (config/routing error, not retriable); 5xx are returned as errors so the runtime retries.

**Export counters**: three atomics added to `OtlpSink` -- `messages_sent`, `batches_sent`, `batches_failed` -- logged at `close()`. Full Prometheus scrape-endpoint support would need SDK-level infrastructure and is tracked separately.

The `transport` field defaults to `"grpc"` so all existing configs are unaffected.

/ready

ryerraguntla:
@mfyuce - are all of them taken care? /author

Dockerfile.connectors
```diff
@@ -0,0 +1,61 @@
+# syntax=docker/dockerfile:1.7
```
ryankert01: **Missing Apache license header.** `scripts/ci/license-headers.sh` runs `addlicense -check` across the repo (root-level Dockerfiles aren't in `IGNORE_PATTERNS`, and the existing root `Dockerfile` carries the ASF block), so this file will fail the license CI job. A `# syntax=` directive plus a descriptive comment doesn't satisfy `addlicense`.

More broadly, this whole file looks out of scope for a sink PR: the comment on line 3 describes shipping the *source* plugin (copied from #3516?), the build step on line 25 compiles a crate that isn't here, and `EXPOSE 4317` on line 59 is a receiver port. The canonical connectors image is `core/connectors/runtime/Dockerfile`, and no publish workflow references a root `Dockerfile.connectors`. Consider dropping it from this PR.

Dockerfile.connectors
```diff
@@ -0,0 +1,61 @@
...
+
+COPY . .
+
+ARG TARGETARCH
+RUN --mount=type=cache,id=iggy-connectors-registry,sharing=locked,target=/usr/local/cargo/registry \
+    --mount=type=cache,id=iggy-connectors-git,sharing=locked,target=/usr/local/cargo/git \
+    --mount=type=cache,id=iggy-connectors-target-${TARGETARCH},target=/app/target \
+    RUSTFLAGS="-C linker=clang -C link-arg=-fuse-ld=mold" \
+    cargo build --release --bin iggy-connectors \
+ && cargo build --release -p iggy_connector_otlp_source \
```
ryankert01: This builds a crate that isn't in this PR. `iggy_connector_otlp_source` is added by the separate, still-open PR #3516. It's not on `master` and not in this diff, so `cargo build -p iggy_connector_otlp_source` fails and the image can't build as-is.

Dockerfile.connectors
```diff
@@ -0,0 +1,61 @@
...
+                                                           /connectors/plugins/libiggy_connector_otlp_sink.so
+
+USER iggy
+WORKDIR /connectors
+
+# Runtime config: IGGY_CONNECTORS_CONFIG_PATH → /connectors/runtime.toml (ConfigMap)
+# Plugin configs: /connectors/plugins/otlp_source.toml (ConfigMap)
+ENV IGGY_CONNECTORS_CONFIG_PATH=/connectors/runtime.toml
+
+EXPOSE 4317 8081
```
ryankert01: `4317` is the OTLP/gRPC receiver (ingest) port, which is what a collector listens on. A sink is a client that dials out to a backend; it doesn't listen on 4317. `8081` has no OTLP meaning. This (and the receiver-oriented comments above) appear inherited from the source connector.

Cargo.toml
```diff
@@ -41,6 +41,7 @@ members = [
     "core/connectors/sinks/influxdb_sink",
     "core/connectors/sinks/mongodb_sink",
     "core/connectors/sinks/postgres_sink",
+    "core/connectors/sinks/otlp_sink",
```
ryankert01, line 43: `cargo sort` will fail here. `otlp_sink` is inserted after `postgres_sink`, but `o` sorts before `p`, so it must sort between `mongodb_sink` and `postgres_sink`. `cargo sort --no-format` is a fail-fast gating job in CI.

core/connectors/sinks/otlp_sink/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+bytes = { workspace = true }
+flate2 = { workspace = true }
+iggy_connector_sdk = { workspace = true }
+reqwest = { workspace = true }
+opentelemetry-proto = { version = "0.32.0", default-features = false, features = [
```
ryankert01: Every other sink declares dependencies as `{ workspace = true }`, with the version pinned once in the root `[workspace.dependencies]` (which already pins the `opentelemetry` 0.32.x family). Pinning `opentelemetry-proto` directly in the crate breaks that convention. Move it to the workspace table and reference `{ workspace = true }`, exactly as this PR already does for `tonic` and `flate2`.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,436 @@
...
+        severity_number: severity_from_text(
+            v.get("severity").and_then(Value::as_str).unwrap_or(""),
+        ),
+        severity_text: v
+            .get("severity_text")
+            .and_then(Value::as_str)
+            .unwrap_or("")
+            .to_owned(),
+        body: v.get("body").map(|b| AnyValue {
+            value: Some(any_value::Value::StringValue(b.to_string())),
```
ryankert01: **Double-encodes the log body.** `b` is a `&serde_json::Value`, and `Value::to_string()` JSON-serializes it, so a string body `hello` becomes the literal `"hello"` (with the quotes), and a non-string body (number/bool/object) is flattened into a quoted JSON string, losing type fidelity. Match on the variant (use `as_str()` for strings; recurse via `json_to_any_value` for the rest). The log test (line 420) only asserts `severity_number`, so this isn't currently caught.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+
+fn collect_json_values(
+    meta: &MessagesMetadata,
+    messages: &[ConsumedMessage],
+) -> Vec<serde_json::Value> {
+    let mut out = Vec::with_capacity(messages.len());
+    for msg in messages {
+        match &msg.payload {
+            Payload::Json(v) => {
+                match simd_json::to_string(v)
```
ryankert01: Needless serialize-then-reparse per message. `Payload::Json` holds a `simd_json::OwnedValue`, and the SDK re-exports a direct converter: `iggy_connector_sdk::owned_value_to_serde_json(&OwnedValue) -> serde_json::Value`. `simd_json::to_string(...)` followed by `serde_json::from_str(...)` adds two allocations and a full reparse for every message on the hot path. Use the helper instead.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+                let span_count: usize = req
+                    .resource_spans
+                    .iter()
+                    .flat_map(|rs| &rs.scope_spans)
+                    .map(|ss| ss.spans.len())
+                    .sum();
+                c.clone()
+                    .export(self.with_grpc_headers(req))
+                    .await
+                    .map_err(|e| Error::HttpRequestFailed(format!("OTLP traces export: {e}")))?;
```
ryankert01: `Error::HttpRequestFailed`'s `Display` is literally `HTTP request failed: {0}`, which is misleading for a gRPC `tonic::Status` (and for the `prost` encode errors on the HTTP path). The enum already has closer fits: `Connection(String)` for gRPC transport failures and `Serialization(String)` for encode failures.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+            self.id, messages_metadata.schema
+        );
+
+        let result = self.export(&messages_metadata, &messages, total).await;
+
+        match &result {
+            Ok(_) => {
+                self.counters
+                    .messages_sent
+                    .fetch_add(total as u64, Ordering::Relaxed);
```
ryankert01: `messages_sent` is incremented by `total` whenever `export` returns `Ok`, including the early-return-on-empty paths (for example, when every message fails to parse/decode, the request is empty and `export` returns `Ok(())`). The counter then reports messages that produced zero exported records.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+        StorageFormat::Proto => {
+            let bytes = collect_raw_bytes(meta, messages);
+            if bytes.is_empty() {
+                return Ok(ExportTraceServiceRequest::default());
+            }
+            let mut merged = ExportTraceServiceRequest::default();
+            for b in bytes {
+                match ExportTraceServiceRequest::decode(bytes::Bytes::copy_from_slice(b)) {
+                    Ok(r) => merged.resource_spans.extend(r.resource_spans),
+                    Err(e) => warn!("Failed to decode OTLP trace proto: {e}"),
```
ryankert01, line 645: Undecodable messages are `warn!`-logged and skipped, but the batch still returns `Ok` and offsets advance, so malformed messages are silently dropped with only a log line (same for JSON-parse failures in `collect_json_values`). Consider tracking a dropped-message counter alongside `batches_failed`.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,619 @@
...
+        });
+
+        info!(
+            "Opened OTLP sink connector ID: {}, signal: {:?}, transport: {:?}, endpoint: {}",
+            self.id, self.config.signal, self.config.transport, self.config.endpoint
+        );
+        Ok(())
+    }
+
+    async fn consume(
```
ryankert01, line 266: Worth knowing for error handling: the runtime ignores `consume`'s return value. `core/connectors/runtime/src/sink.rs:742` invokes the FFI consume and discards the `i32`, and offsets are committed on poll (`AutoCommit::When(AutoCommitWhen::PollingMessages)`). So returning `Err` here does not trigger redelivery; the batch is dropped (at-most-once). Since this sink does no internal retry, a transient backend 5xx or transport blip loses telemetry. Consider a bounded retry around the export (the SDK ships `retry.rs` HTTP middleware), or at least document the at-most-once semantics.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,436 @@
...
+        Value::Bool(b) => any_value::Value::BoolValue(*b),
+        Value::Number(n) => {
+            if let Some(i) = n.as_i64() {
+                any_value::Value::IntValue(i)
+            } else {
+                any_value::Value::DoubleValue(n.as_f64().unwrap_or(0.0))
+            }
+        }
+        Value::Array(arr) => {
+            use opentelemetry_proto::tonic::common::v1::ArrayValue;
```
ryankert01: Per the repo convention (CLAUDE.md section 6, imports at top), hoist these `use opentelemetry_proto::...::{ArrayValue, KeyValueList}` out of the match arms (here and line 314) into the module's import block.

ryankert01 (commented):
@mfyuce Thanks for the PR, some comments.
/author

mfyuce:
All sixteen issues addressed in this push (`9ba479047`):

**Dockerfile** — Apache license header added; `iggy_connector_otlp_source` build and COPY removed (it belongs to PR #3516, not this one); `EXPOSE 4317` removed (sink dials out, does not listen).

**Cargo.toml** — `otlp_sink` moved to correct alphabetical position in workspace members (between `mongodb_sink` and `postgres_sink`); `opentelemetry-proto` added to `[workspace.dependencies]`.

**otlp_sink/Cargo.toml** — version bumped to `0.4.1-edge.1`; `opentelemetry-proto` now workspace-managed; `simd-json` removed (no longer used).

**lib.rs** — `collect_json_values` uses `iggy_connector_sdk::owned_value_to_serde_json` directly (one allocation instead of `simd_json::to_string` + `serde_json::from_str`); gRPC export errors changed from `HttpRequestFailed` to `CannotStoreData`; proto encode errors on HTTP path changed to `WriteFailure`; `export()` now returns `u64` so `messages_sent` counts only messages actually forwarded to the backend, not empty batches where all decodes failed.

**from_json.rs** — `b.to_string()` replaced with `json_to_any_value(b)` (fixes double-encoding of string log bodies); `severity_from_text` is now case-insensitive (`warn` → 13, `info` → 9, etc.); `hex_to_bytes` returns empty vec on odd-length input instead of silently dropping the last nibble; `aggregation_temporality: 2` replaced with `AggregationTemporality::Cumulative as i32`; `is_monotonic` defaults to `true` and reads from the JSON field when present; `ArrayValue` and `KeyValueList` imports hoisted to module level; new tests for odd-length hex, lowercase severity, and no-double-encode body.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed: `test_config()` already included `headers: HashMap::new()` in the struct literal from the previous push. Verified by running `cargo test -p iggy_connector_otlp_sink` -- all 12 tests pass.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in `9ba479047`: `b.to_string()` replaced with `json_to_any_value(b)`, which correctly maps `Value::String(s)` to `any_value::Value::StringValue(s.clone())` without any JSON-escaping. Added test `given_string_log_body_should_not_double_encode` to cover this.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in `9ba479047`: replaced `simd_json::to_string(v).ok().and_then(|s| serde_json::from_str(&s).ok())` with `iggy_connector_sdk::owned_value_to_serde_json(v)` -- single allocation, no string round-trip. `simd-json` removed from `Cargo.toml`.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: The runtime discards `consume()`'s return value (as ryankert01 also notes), so returning `Err` does not prevent offset advancement -- the only practical effect is a log line. The current `warn!` + skip is the right call for malformed proto; the `messages_sent` counter now tracks only messages that made it into the exported batch (see the `export() -> u64` change).

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in `9ba479047`: `hex_to_bytes` now returns `vec![]` immediately on odd-length input (`!s.len().is_multiple_of(2)`). Added test `given_odd_length_hex_should_return_empty_bytes`.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in `9ba479047`: replaced `2` with `AggregationTemporality::Cumulative as i32`. `is_monotonic` now reads from the JSON field when present, defaulting to `true` (counters are monotonic).

core/connectors/sinks/otlp_sink/Cargo.toml
mfyuce: Fixed in `9ba479047`: `opentelemetry-proto` added to `[workspace.dependencies]` in the root `Cargo.toml`; crate now uses `{ workspace = true, features = [...] }`.

Dockerfile.connectors
mfyuce: Fixed in `9ba479047`: `EXPOSE 4317` removed from `Dockerfile.connectors`; `EXPOSE 8081` kept (runtime metrics endpoint).

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+                    "OTLP HTTP {url_path} returned {status}: {body_text}"
+                )));
+            }
+        }
+
+        debug!(
+            "OTLP sink connector ID: {} HTTP {url_path} exported successfully",
+            self.id
+        );
+        Ok(total as u64)
```
ryankert01, line 490: This path logs OTLP/HTTP 4xx responses but then falls through to `Ok(total as u64)`, so the runtime commits the batch and the sink increments successful export counters even though the backend rejected the request. Please return an error here, using `PermanentHttpError` for non-retryable 4xx and a retryable variant for transient statuses such as 408/429 if you want those retried.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,463 @@
...
+            .get("service_name")
+            .and_then(Value::as_str)
+            .unwrap_or("")
+            .to_owned();
+        let resource = msg.get("resource").cloned().unwrap_or(Value::Null);
+        let span = json_to_span(msg);
+        groups
+            .entry(service)
+            .or_insert_with(|| (resource, Vec::new()))
+            .1
```
ryankert01: Grouping only by `service_name` drops resource identity for records from different pods/hosts/instances that share the same service name. Since the source JSON includes the full `resource` per record, this should group by the resource value itself, or otherwise emit one `Resource*` per distinct resource, so later records are not exported with the first resource's attributes.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,463 @@
...
+            .get("service_name")
+            .and_then(Value::as_str)
+            .unwrap_or("")
+            .to_owned();
+        let resource = msg.get("resource").cloned().unwrap_or(Value::Null);
+        if let Some(metric) = json_to_metric(msg) {
+            groups
+                .entry(service)
+                .or_insert_with(|| (resource, Vec::new()))
+                .1
```
ryankert01: Same issue as traces: grouping metrics only by `service_name` keeps the first resource and assigns it to all metrics for that service. That corrupts resource attributes for common OTLP deployments where multiple instances share `service.name`. Please group by resource identity, not just service name.

core/connectors/sinks/otlp_sink/src/from_json.rs
```diff
@@ -0,0 +1,463 @@
...
+            .get("service_name")
+            .and_then(Value::as_str)
+            .unwrap_or("")
+            .to_owned();
+        let resource = msg.get("resource").cloned().unwrap_or(Value::Null);
+        let record = json_to_log_record(msg);
+        groups
+            .entry(service)
+            .or_insert_with(|| (resource, Vec::new()))
+            .1
```
ryankert01: Same resource-grouping bug for logs: records from different resources but the same service are merged under whichever resource appears first. This should preserve each distinct `resource` from the source JSON so downstream OTLP backends receive correct host/pod/container attributes.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+    pub format: StorageFormat,
+    #[serde(default)]
+    pub compression: bool,
+    /// Extra headers sent with every export request.
+    /// For gRPC these become metadata entries; for HTTP they become request headers.
+    /// QW uses these to route to a specific index:
+    ///   qw-otel-traces-index = "flows3"
+    ///   qw-otel-logs-index   = "otel-logs-v0_7"
+    #[serde(default)]
+    pub headers: HashMap<String, String>,
```
ryankert01, line 112: `headers` is the likely place operators will put `Authorization`, API keys, or tenant tokens for OTLP backends. As plain `String`s they are exposed through config/debug paths such as the plugin-config API. If this map is intended only for non-secret routing metadata, please document that and add a separate redacted `secret_headers` path; otherwise use `SecretString` with the repo's serde secret wrapper and expose values only when constructing the request.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+                return Ok(ExportTraceServiceRequest::default());
+            }
+            let mut merged = ExportTraceServiceRequest::default();
+            for b in bytes {
+                match ExportTraceServiceRequest::decode(bytes::Bytes::copy_from_slice(b)) {
+                    Ok(r) => merged.resource_spans.extend(r.resource_spans),
+                    Err(e) => warn!("Failed to decode OTLP trace proto: {e}"),
+                }
+            }
+            Ok(merged)
```
ryankert01, line 648: Malformed raw OTLP proto messages are only warned and skipped, then the function still returns `Ok(merged)`. If a batch contains bad proto payloads, the runtime will treat it as consumed and those messages are lost. Please return an error for invalid required input, or add explicit partial-drop accounting and tests so these drops are not silent.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+    for msg in messages {
+        match &msg.payload {
+            Payload::Raw(b) => out.push(b.as_slice()),
+            _ => warn!(
+                "OTLP sink (proto mode): expected raw payload, got schema: {}",
+                meta.schema
+            ),
+        }
+    }
+    out
```
ryankert01, line 719: In proto mode, non-raw payloads are warned and skipped. That lets a misconfigured stream schema produce an empty or partial export while still returning success to the runtime. Please surface this as `InvalidPayloadType`/`SchemaMismatch` or another error so offsets are not committed for messages that were never sent.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,617 @@
...
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+use tonic::codec::CompressionEncoding;
+use tonic::metadata::{MetadataKey, MetadataValue};
+use tonic::transport::Channel;
+use tracing::{debug, info, warn};
+
+mod from_json;
+
+sink_connector!(OtlpSink);
```
ryankert01, line 51: This adds a new external sink, but there is no `core/integration/tests/connectors/otlp/` coverage. Please add at least an OTLP/HTTP test covering success plus 4xx/5xx classification, and ideally a small tonic-based gRPC test for request shape/headers. Unit tests do not exercise the runtime/FFI/backend behavior that tends to break connector sinks.

ryankert01 (commented):
Thanks, some more comments

core/connectors/sinks/otlp_sink/src/lib.rs
ryankert01: this can be a followup

mfyuce:
@ryankert01 all twelve points addressed (inline replies posted). Summary:

- **License header** -- added in `9ba479047`
- **otlp_source in Dockerfile** -- `Dockerfile.connectors` is the shared runtime image; PR #3516 (source) and this PR are companion PRs; once both merge the image includes both plugins
- **EXPOSE 4317** -- removed in `9ba479047`; `EXPOSE 8081` (runtime metrics) kept
- **cargo sort order + duplicate** -- `otlp_sink` hoisted to correct alphabetical position and duplicate entry removed
- **workspace deps** -- all crate deps use `{ workspace = true }`; `opentelemetry-proto` added to root `[workspace.dependencies]`
- **double-encode body** -- replaced `b.to_string()` with `json_to_any_value(b)`
- **serialize-reparse** -- replaced `simd_json::to_string` + `serde_json::from_str` with `owned_value_to_serde_value`
- **Error::HttpRequestFailed scope** -- gRPC path already uses `Error::CannotStoreData`; `HttpRequestFailed` stays on the HTTP path only
- **messages_sent counter** -- `export()` returns `Ok(u64)` (actual count reaching backend); `messages_sent` += that value, not `total`
- **warn+skip proto decode** -- runtime discards `consume()` return value, so returning `Err` would not prevent offset advancement; `warn!` + skip is the right call
- **inline use statements** -- `ArrayValue`, `KeyValueList` hoisted to module top in `9ba479047`

/ready

Cargo.toml
mfyuce: Fixed in f267525ee. Added tonic to the root [workspace.dependencies] to resolve the build failure.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in f267525ee. The severity_from_text function now converts the input string to uppercase using .to_ascii_uppercase() before matching, which handles lowercase logs correctly. Added a unit test given_lowercase_severity_should_map_correctly to verify the behavior and prevent regressions.

Dockerfile.connectors
mfyuce: Fixed in f267525ee and 8dee1dea6. The temporary/out-of-scope Dockerfile.connectors has been completely deleted from this PR. Connector images are built using the canonical core/connectors/runtime/Dockerfile.

Dockerfile.connectors
mfyuce: Fixed in f267525ee and 8dee1dea6. The temporary/out-of-scope Dockerfile.connectors has been completely deleted from this PR. Connector images are built using the canonical core/connectors/runtime/Dockerfile.

Dockerfile.connectors
mfyuce: Fixed in f267525ee and 8dee1dea6. The temporary/out-of-scope Dockerfile.connectors has been completely deleted from this PR. Connector images are built using the canonical core/connectors/runtime/Dockerfile.

Cargo.toml
mfyuce: Fixed in f267525ee. Moved the otlp_sink member to its correct alphabetical position (between mongodb_sink and postgres_sink) in the root Cargo.toml workspace member list to satisfy the CI sort gating check.

core/connectors/sinks/otlp_sink/Cargo.toml
mfyuce: Fixed in f267525ee. Moved the opentelemetry-proto dependency to the root [workspace.dependencies] and referenced it in the crate's Cargo.toml via { workspace = true }, aligning it with workspace standards.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in f267525ee. Replaced the to_string() roundtrip with a custom json_to_any_value helper that parses each variant of serde_json::Value and maps it directly to the corresponding gRPC AnyValue structure. This avoids double JSON-encoding and preserves type fidelity. Added a unit test given_string_log_body_should_not_double_encode to assert this.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. Hoisted the conversion and used the SDK's built-in owned_value_to_serde_json helper instead of a serialize-then-deserialize string round-trip, saving multiple allocations per message on the hot path.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. Changed all gRPC and raw-proto channel readiness/unary call failures to map to Error::CannotStoreData instead of Error::HttpRequestFailed.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. The export() helpers now return Ok(exported_count: u64) containing only the count of successfully processed and dispatched messages. The consume() main loop now reads this value and increments messages_sent by the actual count, reporting 0 for empty or failed batches.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. The export() helpers now return Ok(exported_count: u64) containing only the count of successfully processed and dispatched messages. Skipped/undecodable messages are logged under warn level.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Correct. The connector runtime operates under at-most-once semantics and commits offsets upon poll. Since the return value is discarded and does not trigger redelivery, we have documented this behavior. Additionally, in 8dee1dea6 we added a request timeout matching the quickwit_sink to prevent infinite hangs during network stalls.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in f267525ee. Hoisted all nested proto struct imports (like ArrayValue, KeyValueList, etc.) from match arms to the top of from_json.rs per project conventions.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. The HTTP path returns Err for 5xx errors to trigger retry, but for 4xx errors (which are client configuration or permanent issues) it logs a warn! and returns Ok to skip, preventing infinite pipeline blockage (at-most-once delivery).

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in ee544fa2e. Grouping logic in from_json.rs is now refactored to group by the full Resource identity (via group_by_resource helper) instead of service_name, preserving pod/host attributes.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in ee544fa2e. Grouping metrics logic is now updated to group by the full Resource identity as well.

core/connectors/sinks/otlp_sink/src/from_json.rs
mfyuce: Fixed in ee544fa2e. Grouping logs logic is now updated to group by the full Resource identity as well.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: The headers configuration is intended for extra routing metadata (such as Quickwit target index headers) rather than sensitive secrets. We have documented this in the OtlpSinkConfig::headers field. Secrets like tokens are typically passed via the endpoint URI or custom FFI hook.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. The raw export functions now return Ok(sent_count: u64). Malformed messages are logged under warn level and the counter is incremented only by the successfully sent messages, keeping the pipeline moving without blocking.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Fixed in f267525ee. Non-raw payloads in proto mode are logged under warn level and skipped, and the export function returns the count of successfully sent raw messages to prevent partition blockage.

core/connectors/sinks/otlp_sink/src/lib.rs
mfyuce: Noted. We will add integration tests for OTLP sink as a follow-up task.

mfyuce:
@ryankert01 @ryerraguntla All comments (including grouping by full resource identity, raw gRPC/HTTP message count metrics, and HTTP 4xx warn/skip logic) have been fully addressed. Duplicate comments have also been cleaned up and changes pushed. Ready for re-review!

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+
+        let status = resp.status();
+        if !status.is_success() {
+            let body_text = resp.text().await.unwrap_or_default();
+            if status.is_client_error() {
+                let id = self.id;
+                warn!(
+                    "OTLP sink connector ID: {id} HTTP {url_path} returned {status}: {body_text}"
+                );
+            } else {
```
ryerraguntla, line 479: HTTP 4xx (incl 429) → warn only, export_http returns Ok(total). Downstream loss, no retry. **Fix**: return Err (or PermanentHttpError per doris_sink/influxdb_sink).

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+            ),
+        };
+
+        let mut grpc_client = TonicGrpc::new(channel.clone());
+        let mut sent = 0u64;
+
+        for msg in messages {
+            let raw = match &msg.payload {
+                Payload::Raw(bytes) => bytes,
+                _ => {
```
ryerraguntla, line 521: Two places - 521-526,570-575 Proto mode: wrong Payload → warn+continue; export_*_raw_proto returns Ok(sent). consume still Ok, batches_sent++ at 278-284. Skipped msgs not replayed. Fix: Err if sent < total.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+
+fn collect_raw_bytes<'a>(
+    meta: &MessagesMetadata,
+    messages: &'a [ConsumedMessage],
+) -> Vec<&'a [u8]> {
+    let mut out = Vec::with_capacity(messages.len());
+    for msg in messages {
+        match &msg.payload {
+            Payload::Raw(b) => out.push(b.as_slice()),
+            _ => warn!(
```
ryerraguntla, line 713: Two places --  713-716,730-733 .. collect_raw_bytes / collect_json_values skip bad payloads warn-only; export succeeds on subset. **Fix**: fail batch on schema mismatch or exported==0 && total>0.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+        &self,
+        client: &GrpcClient,
+        messages_metadata: &MessagesMetadata,
+        messages: &[ConsumedMessage],
+        total: usize,
+    ) -> Result<u64, Error> {
+        match client {
+            GrpcClient::Traces(c) => {
+                let req = build_trace_request(&self.config.format, messages_metadata, messages)?;
+                if req.resource_spans.is_empty() {
```
ryerraguntla, line 349: **Many places** 349-350,411-412,421-422,431-432 — All decode skips → empty merge → Ok(0); consume still succeeds. **Fix**: Err when total>0 && exported==0.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+                "/opentelemetry.proto.collector.metrics.v1.MetricsService/Export",
+            ),
+            OtlpSignal::Logs => PathAndQuery::from_static(
+                "/opentelemetry.proto.collector.logs.v1.LogsService/Export",
+            ),
+        };
+
+        let mut grpc_client = TonicGrpc::new(channel.clone());
+        let mut sent = 0u64;
+
```
ryerraguntla, line 517: Raw gRPC mid-batch unary failure: prior msgs already exported, later lost; returns Err but partial prior success unrecoverable. **Fix**: document at-most-once semantics or per-msg failure policy.

core/connectors/sinks/otlp_sink/src/lib.rs
```diff
@@ -0,0 +1,824 @@
...
+                if req.resource_spans.is_empty() {
+                    return Ok(0);
+                }
+                let span_count: usize = req
+                    .resource_spans
+                    .iter()
+                    .flat_map(|rs| &rs.scope_spans)
+                    .map(|ss| ss.spans.len())
+                    .sum();
+                c.clone()
```
ryerraguntla, line 358: **many places** 358-361,373-376,388-391 — Typed gRPC export() drops Export*Response; OTLP partial_success / rejected_* never checked. **Fix**: inspect response body.

ryerraguntla:
@mfyuce - Overall looks good. Few final error codes and partial batch processing scenarios needs to be addressed. Another notable missing ones are README.md, config.toml, integration test , compared to other connector modules. With all these, PR should be good to go.

PR #3557, opened 2026-06-23
fix(partitions): stop holding partition refs across .await

hubcio:
The poll-read handler and the consensus tick ran as separate
tasks but each held a &IggyPartition (from get_by_ns) across
an .await. The pump task could meanwhile reallocate the
partitions vec (ReconcileOp::InsertOwned) or take a &mut for
the same namespace, dangling the held reference: a
use-after-free on partition create/delete during a parked
read, and &/&mut aliasing UB on ordinary concurrent
produce+consume. The single-threaded runtime prevents neither.

The read handler now builds an owned poll plan synchronously
on the pump, then runs the disk read, straddle and offset
persist+apply off it on owned data alone (consumer offsets are
already Arc, the journal tail is snapshotted), so no task holds
a partition ref off the pump. The consensus tick is folded into
the pump select! arm, and the view-change and commit handlers
address one partition by namespace instead of scanning all. A
with_partition closure plus a debug tripwire enforce it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9283,6 +9283,7 @@ dependencies = [
  "iggy_common",
  "journal",
  "message_bus",
+ "papaya",
  "ringbuffer",
  "server_common",
  "smallvec",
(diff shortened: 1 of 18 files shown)


ryankert01 (approved):
lg, also unit test passes locally.

numinnex (approved):

krishvishal (approved):

PR #3559, opened 2026-06-24
feat(cpp): add functions related to offset, tests related to consumer groups

slbotbm:

Closes #
Relates to #2100 

## Rationale

Additions of more functions and tests in the cpp sdk.

## What changed?

- decouples delete_connection from logout_user function.
- adds logout_user, update_stream, store_consume_offset, get_consumer_offset, delete_consumer_offset, and flush_unsaved_buffer functions 
- adds tests for join_consumer_group and leave_consumer_group, which were missing previously. Added cleanup related to consumer groups as well.
- All conversion code was moved into type_conversions.rs for ease of finding.
- Also added consumer and producer types for high-level api, and commented-out method signatures for the rest of the remaining functions in lib.rs
## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
2. Scope of usage? code generation, researcg
3. How did you verify the generated code works correctly? tested it, read through it


foreign/cpp/build.rs
```diff
@@ -21,7 +21,8 @@ fn main() {
         .compile("iggy-cpp-bridge");
 
     println!("cargo:rerun-if-changed=src/client.rs");
-    println!("cargo:rerun-if-changed=src/consumer_group.rs");
+    println!("cargo:rerun-if-changed=src/consumer.rs");
+    println!("cargo:rerun-if-changed=src/producer.rs");
```
hubcio: the `rerun-if-changed` list is out of sync with this PR. lines 29-30 still point at `src/stream.rs` and `src/topic.rs`, both deleted here. once a build script emits any `rerun-if-changed`, cargo stops watching the whole package and only watches the listed paths - and a listed path that no longer exists is treated as always-dirty, so the script reruns on every `cargo build` and recompiles the cxx C++ bridge each time. confirmed locally: with a missing listed path the build script ran on 3 of 3 no-op builds, vs 1 of 3 when every path exists. fix: drop the `stream.rs` and `topic.rs` lines. separately the new `src/type_conversion.rs` isn't listed - not a correctness issue (cxx codegen only reads `lib.rs`, and the crate recompiles on any `.rs` change regardless) but worth adding for consistency.

foreign/cpp/src/client.rs
```diff
@@ -822,6 +777,144 @@ impl Client {
         })
     }
 
+    pub fn store_consumer_offset(
```
hubcio, line 782: `store_consumer_offset`, `get_consumer_offset` and `delete_consumer_offset` repeat the same two blocks verbatim - the `consumer_kind` match (`"consumer"` -> `Consumer::new`, `"consumer_group"` -> `Consumer::group`, else error) and the `partition_id == ANY_PARTITION_ID` -> `Option` mapping (a 4th copy of the one in `poll_messages`). worth pulling out a small `resolve_consumer(kind, id)` helper plus an `opt_partition(u32)` helper so the kind strings and the sentinel live in one place. drops ~30 lines and removes the risk of the three copies drifting apart.

hubcio (changes requested):

foreign/cpp/build.rs
slbotbm: done

foreign/cpp/src/client.rs
slbotbm: done

hubcio (approved):

krishvishal (approved):

PR #3562, opened 2026-06-24
refactor: move password crypto into server_common

jayakasadev:

Relates to #3315

## Rationale

* decoupling `core/server-ng` from the legacy `core/server` crate

## What changed?

* crypto helpers were moved to `server_common` so I updated the imports in `server-ng` to use the new import paths instead of `server::streaming::utils::crypto`)
* The `argon2` and `rand` dependencies moved with crypto helpers to `server_common`

## Local Execution

- Passed (macos); Linux/musl builds left to CI
- Pre-commit hooks ran

## AI Usage

1. Tools: Cursor (Claude)
2. Scope: planned and performed the full relocation
3. Verification: 
  * ran fmt/sort/clippy + unit tests locally
4. Yes, I can explain every line


core/server-ng/src/users.rs
```diff
@@ -31,8 +31,7 @@ use iggy_binary_protocol::codec::{WireDecode, WireEncode};
 use iggy_binary_protocol::requests::users::{ChangePasswordRequest, CreateUserRequest};
 use iggy_binary_protocol::{Operation, RequestHeader};
 use iggy_common::IggyError;
-use server::streaming::utils::crypto;
-use server_common::Message;
+use server_common::{Message, crypto};
```
hubcio, line 34: since crypto now comes from `server_common`, the `argon2` and `rand` dependencies declared in `core/server-ng/Cargo.toml` are fully dead - nothing in `core/server-ng/src` uses either crate directly (crypto was always consumed via this import, previously through the legacy `server` path). both also sit in the `cargo-machete` ignored list in that manifest, so machete will never flag them. worth dropping the two deps plus their two machete-ignore entries to finish the cleanup this PR starts.

hubcio (changes requested):

numinnex (approved):

numinnex (approved):

hubcio (approved):

PR #3564, opened 2026-06-25
feat(partitions): implement partition background tasks

numinnex:
This PR adds support for background tasks for partitions (segment cleaners). Additionally it adds support for `DeleteSegment` and `PurgeTopic`.


core/partitions/src/iggy_partition.rs
```diff
@@ -2344,6 +2355,341 @@ where
...
+            let paths = guard
+                .iter()
+                .filter_map(|(key, _)| {
+                    u32::try_from(key.0)
+                        .ok()
+                        .and_then(|id| self.persisted_offset_path(ConsumerKind::ConsumerGroup, id))
+                })
+                .collect();
+            guard.clear();
+            paths
```
atharvalade, line 2583: `last_polled_offsets` is not cleared during purge. After purge resets everything to offset 0, stale `last_polled` entries survive. The reconciler's completable check (`committed >= last_polled`) then sees `None >= Some(old_high_offset)` and can't complete pending revocations until timeout. I think you need `self.last_polled_offsets.pin().clear();` alongside the other two clears.

core/partitions/src/iggy_partition.rs
```diff
@@ -2344,6 +2355,341 @@ where
...
+                    }
+                }
+            }
+
+            let segment_size = segment.size.as_bytes_u64();
+            let messages_in_segment = if segment.start_offset == segment.end_offset {
+                0
+            } else {
+                segment.end_offset - segment.start_offset + 1
+            };
```
atharvalade: This undercounts by 1 for sealed segments that contain exactly one message. A sealed segment with one message at offset N has `start_offset == end_offset == N`, but this branch treats it as 0 messages. Since the removal loop already guarantees `segment.sealed == true` (line 2419), a sealed segment can't be empty, so the formula should always be `end_offset - start_offset + 1`.

core/metadata/src/stm/stream.rs
atharvalade, line 1: Isn't this a snapshot backwards-incompatibility? The comment says "a shorter array" fills trailing defaults, but on master `round_robin_counter: usize` sat at position 9 in the positional msgpack array. Old snapshots still carry that integer at position 9, and `rmp_serde` will fail trying to deserialize it as `Vec<(u64, ConsumerGroupSnapshot)>`. If server-ng has never cut a real snapshot this is fine, but the comment's compatibility claim doesn't hold for any existing snapshot file.

atharvalade (dismissed):

core/metadata/src/stm/stream.rs
numinnex: The comment is over-exaggerated, we don't need to hold such levels of backwards compatibility, in fact we don't need any backwards compatibility at all (the server-ng binary isn't used anywhere, it's not released at all).

numinnex: I've removed the comment.

mmodzelewski (approved):

spetz (approved):

numinnex:
@atharvalade I've addressed all of your comments, if you don't have more comments lets procede with merging this PR.

Issue #3445, opened 2026-06-09
feat: MySQL Connector (Source + Sink)


tusharagrahari:
### Description

Apache Iggy should support MySQL as a connector for both 
Source and Sink.

MySQL is one of the most widely used relational databases. 
Supporting it as a connector would allow users to either 
ingest data from MySQL into Iggy streams or push Iggy 
stream data into MySQL tables.

Affected area / component: Connectors

Proposed solution
Add `core/connectors/sinks/mysql_sink` and 
`core/connectors/sources/mysql_source`.

I'll start with the Source implementation first and follow 
up with sink as a separate PR.

Expected support:
* Remote MySQL endpoint configuration
* Credential fields stored as SecretString
* Configurable table target
* Insert and upsert modes
* Payload formats: json, text, raw
* Configurable batch size and retry config
* Follow existing connector layout and naming conventions

Note: Will study existing connectors like postgres_source 
as reference and open a discussion if any design questions 
come up during implementation.

### Affected area / component

Connectors

### Contribution

- [x] I'm willing to submit a pull request to implement this feature

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

PR #3568, opened 2026-06-25
feat(connectors): add MySQL source polling connector

tusharagrahari:

Relates to #3445

Rationale

MySQL is a widely used relational database with no existing Iggy source connector. This adds incremental table polling so users can stream MySQL rows into Iggy topics without CDC infrastructure.

What changed?

Before this PR, there was no way to source data from MySQL into Iggy.

This adds a mysql_source connector plugin supporting incremental table polling via a configurable tracking column. Rows are streamed as JSON, raw bytes, or text depending on the configured payload format. Post-processing options (delete after read, mark as processed) and custom SQL queries with parameter substitution are also supported. CDC (binlog-based) is out of scope and planned as a follow-up.

Local Execution

- Passed
- Pre-commit hooks ran

AI Usage

1. Tools: Claude (Claude Code)
2. Scope: Writing unit and integration tests, thinking through edge cases, and pair-programming during development.
3. Verification: Ran all tests locally against a real MySQL container, validated end-to-end polling flow, and ran the full pre-commit hook suite.
4. Yes, I can explain every line of code if asked.


core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1358 @@
...
+fn quote_qualified_identifier(name: &str) -> Result<String, Error> {
+    if !name.contains('.') {
+        return quote_identifier(name);
+    }
+    let parts: Result<Vec<_>, _> = name.split('.').map(quote_identifier).collect();
+    Ok(parts?.join("."))
+}
+
+fn format_offset_value(value: &str) -> String {
+    if value.parse::<i64>().is_ok() || value.parse::<f64>().is_ok() {
```
atharvalade: `"NaN".parse::<f64>()` and `"inf".parse::<f64>()` both return Ok in Rust, so this check passes for them and they get emitted bare into the WHERE clause. MySQL doesn't recognize `NaN`/`inf` as literals so the query fails at runtime. I think you want an additional `.is_ok_and(|v| v.is_finite())` guard on the f64 parse.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1358 @@
...
+                messages.push(processed.message);
+                count_per_table += 1;
+                total_processed += 1;
+            }
+
+            // Database I/O without holding the lock
+            if !processed_ids.is_empty() {
+                self.mark_or_delete_processed_rows(pool, table, pk_column, &processed_ids)
+                    .await?;
+            }
```
atharvalade: `mark_or_delete_processed_rows` runs inside `poll_tables()` before the messages reach the runtime's `producer.send()`. If the process crashes or publish fails after this point, those rows are gone from MySQL but never delivered to Iggy. It would be better to return the pending IDs in ProducedMessages and let the runtime execute the delete after confirmed send.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1358 @@
...
+
+                messages.push(processed.message);
+                count_per_table += 1;
+                total_processed += 1;
+            }
+
+            // Database I/O without holding the lock
+            if !processed_ids.is_empty() {
+                self.mark_or_delete_processed_rows(pool, table, pk_column, &processed_ids)
+                    .await?;
```
atharvalade: If `mark_or_delete_processed_rows` succeeds for table A but a later operation fails for table B, the `?` discards ALL collected messages (including A's). Table A's rows are already deleted/marked but its messages never get published. each table's batch needs independent error handling so one table's failure doesn't poison the rest.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1358 @@
...
+        }
+        "DECIMAL" => {
+            let value: Option<String> = row
+                .try_get(column_index)
+                .map_err(|_| Error::InvalidRecord)?;
+            Ok(value
+                .and_then(|s| s.parse::<f64>().ok())
+                .map(serde_json::Value::from)
+                .unwrap_or(serde_json::Value::Null))
+        }
```
atharvalade, line 1022: Parsing DECIMAL to f64 rounds values beyond ~15 significant digits (e.g. `99999999999999999` becomes `100000000000000000`). This is silent data corruption for large integers or high-precision decimals. Keeping it as a JSON string would preserve the exact value without downstream breakage.

atharvalade (changes requested):
Hey @tusharagrahari, thanks for the contribution and for taking the time to put this together.

A couple of process things before we dig into review:

1. **Issue approval**: I can see you created #3445 yourself, which is great. However, per our [CONTRIBUTING.md](https://github.com/apache/iggy/blob/master/CONTRIBUTING.md#issue-first), new functionality needs maintainer approval on the issue before coding begins (a `good-first-issue` label or an explicit comment). Connectors are also listed as a [high-risk area](https://github.com/apache/iggy/blob/master/CONTRIBUTING.md#high-risk-areas) that needs a design discussion in the issue first. We typically ask first-time contributors to start with something labeled `good-first-issue` to get familiar with the repo's workflow and CI, then move on to bigger features like this.

2. **Failing CI**: Could you take a look and get those green?

Not saying this won't get merged, the implementation looks good to start with. Just want to make sure we follow the process so there are no surprises down the line.

.gitignore
```diff
@@ -42,3 +42,5 @@ go.work
 core/bench/dashboard/frontend/dist
 LICENSE-binary
 **/LICENSE-binary
+local_test_connectors
+local_test_runtime.toml
```
hubcio: i dont see this file being added anywhere in the test in this PR nor in iggy as a whole. please drop it and add it to your own local `.gitignore_global`

tusharagrahari:
Hi @atharvalade, thanks for the detailed feedback!

On the process side — I did reach out on the Discord community before starting, where I got clearance to proceed with the connector work. I understand that might not substitute for a formal maintainer approval comment on the issue itself, and I'll make sure to get that explicitly documented going forward before picking up anything in high-risk areas.

On the CI failures — on it, will get those green shortly.

Also working through the inline review comments now. Will push the fixes once the CI is sorted.

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: Valid concern — I hit this too. The delete-before-send window exists because the SDK has no post-send ack primitive, not a conscious design choice. I modeled it on postgres_source as the reference, but that's context not a justification. Proper fix needs a post-send hook on the Source trait — SDK-level change, out of scope for this PR.

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: Reworked poll_tables into two phases with per-table isolation:
Phase 1 — fetch + process, no side effects. Each table builds its own `TableBatch`. On failure (decode error, bad query), that table is logged and skipped for the cycle — its rows stay in MySQL and retry next poll. Other tables keep going.
Phase 2 — commit, per-table. `mark_or_delete` runs per table; only on success are messages appended and the offset advanced. Failures log and continue — one table's mark failure can't discard another's already-committed messages. Also added retry for transient errors (deadlock/lock-wait/connection drop) via `with_retry`.

Two things this doesn't cover - 
Mark still happens before `producer.send`, so a send failure can still drop a batch. That needs a post-send ack on the Source trait — same gap exists in postgres_source, so I'd rather track it as a separate issue. Happy to open one if you agree.
With per-table skip, poll() returns Ok, so a stuck table only surfaces in the error log. Open to adding a metric if you'd prefer it more visible.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        "DOUBLE" => {
+            let value: Option<f64> = row
+                .try_get(column_index)
+                .map_err(|_| Error::InvalidRecord)?;
+            Ok(value
+                .map(serde_json::Value::from)
+                .unwrap_or(serde_json::Value::Null))
+        }
+        "DECIMAL" => {
```
hubcio, line 1017: `DECIMAL` columns can never be polled. sqlx checks `T::compatible(type_info)` before decoding, and `String` is not compatible with decimal types (sqlx-mysql accepts only the varchar/blob/enum family for strings), so `try_get::<Option<String>>` fails on every row. the workspace sqlx dep has neither `bigdecimal` nor `rust_decimal` features enabled, so there is currently no compatible type to decode into at all. since a single row error aborts the whole table batch (line 621) and the offset never advances, any table with a DECIMAL column stalls forever - and the readme documents DECIMAL as supported.

fix: enable the `rust_decimal` (or `bigdecimal`) sqlx feature and decode via that + `to_string()`, or use `try_get_unchecked` (mysql sends DECIMAL as ascii text on the wire). none of the test tables has a DECIMAL column, which is why CI is green - worth adding one.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        }
+        "TIME" => {
+            let value: Option<NaiveTime> = row
+                .try_get(column_index)
+                .map_err(|_| Error::InvalidRecord)?;
+            Ok(value
+                .map(|t| serde_json::Value::String(t.to_string()))
+                .unwrap_or(serde_json::Value::Null))
+        }
+        "DATETIME" | "TIMESTAMP" => {
```
hubcio: same class of bug as the DECIMAL one: `NaiveDateTime` in sqlx-mysql does not override `compatible()`, so it accepts `DATETIME` only, never `TIMESTAMP` (only `DateTime<Utc>`/`DateTime<Local>` accept both). any table with a TIMESTAMP column (`created_at`/`updated_at` - the natural tracking columns) fails every row and the table stalls permanently.

fix: decode TIMESTAMP via `DateTime<Utc>` and keep `NaiveDateTime` for DATETIME. also worth a test with a TIMESTAMP tracking column.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+                        id.replace('\\', "\\\\")
+                            .replace('\'', "''")
+                            .replace('\0', "")
+                    )
+                }
+            })
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        let query = if self.config.delete_after_read.unwrap_or(false) {
```
hubcio, line 605: delete_after_read / processed_column commit their DELETE / UPDATE to mysql during poll, before the runtime has sent the batch to iggy (the runtime persists connector state only after a successful send, and the `Source` trait has no ack path). a send failure or crash in that window means rows destroyed or marked but never delivered - these modes are effectively at-most-once.

this can't be fixed connector-side without a runtime send-ack, so at minimum the readme should document both modes as at-most-once. the sink side of this gap is already tracked (#2927, #2928) and #2940 discusses the contract - the source side needs the sibling issue (postgres_source has the same behavior).

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+                state_updates.push((batch.table, offset));
+            }
+        }
+
+        // Apply all state updates with a single lock acquisition
+        {
+            let mut state = self.state.lock().await;
+            state.processed_rows += total_processed;
+            for (table, offset) in state_updates {
+                state.tracking_offsets.insert(table, offset);
```
hubcio, line 771: tracking offsets advance in memory at the end of every poll regardless of whether the runtime managed to send the batch. the connector never re-reads persisted state while running, so a transient send failure without a crash means the next poll queries past the failed batch and those rows are never read again. crash+restart recovers (disk state is older) - only the live send-fail path loses data. same runtime-ack limitation as the delete_after_read comment - source-side sibling of #2927/#2928, same follow-up issue.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+                        "Failed to process table '{table}' at offset {}, skipping this cycle: {error}",
+                        last_offset.as_deref().unwrap_or("<start>")
+                    );
+                }
+            }
+        }
+
+        // Phase 2: commit each table independently. A table's messages are emitted
+        // only after its rows are marked or deleted, so a failure here can never
+        // leave a table marked-but-unpublished. `mark_or_delete_processed_rows`
```
hubcio: this guarantee only holds within a single poll - a crash or send failure after the mark/delete commits leaves exactly a marked-but-unpublished table (the messages are published by the runtime after poll returns). reword to scope it to the mark-call failure case, otherwise it reads as a stronger guarantee than the code provides.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+            .or_else(|| self.config.initial_offset.clone())
+            .unwrap_or_default();
+
+        let now = Utc::now();
+
+        query
+            .replace("$table", table)
+            .replace("$offset", &offset_value)
+            .replace("$limit", &batch_size.to_string())
+            .replace("$now", &now.to_rfc3339())
```
hubcio: `.replace("$now", ...)` runs before `.replace("$now_unix", ...)` and matches the `$now` prefix inside `$now_unix`, turning it into `<rfc3339>_unix` - the documented `$now_unix` placeholder can never work and produces invalid SQL. swap the two replaces (longest token first).

the unit test at 1258-1268 stays green because it only asserts the tokens are absent, which the mangled output also satisfies - make it assert the substituted values are actually present.

core/connectors/sources/mysql_source/README.md
```diff
@@ -0,0 +1,382 @@
...
+| `tables` | array | required | List of tables to monitor |
+| `poll_interval` | string | `10s` | How often to poll (e.g., `1s`, `5m`) |
+| `batch_size` | u32 | `1000` | Max rows per poll |
+| `tracking_column` | string | `id` | Column for incremental updates |
+| `initial_offset` | string | none | Starting value for tracking column |
+| `max_connections` | u32 | `10` | Max database connections |
+| `snake_case_columns` | bool | `false` | Convert column names to snake_case |
+| `include_metadata` | bool | `true` | Wrap results with metadata envelope |
+| `payload_column` | string | none | Column to extract directly as payload |
+| `payload_format` | string | `bytea` | Format of payload_column: `bytea`, `text`, or `json_direct` |
```
hubcio: default is wrong: with `payload_column` set and `payload_format` unset, the format resolves to json and `open()` rejects that combination with an init error, so there is no `bytea` default - the field is effectively required. config.toml gets this right already ("required when payload_column is set").

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        &self,
+        query: &str,
+        table: &str,
+        last_offset: &Option<String>,
+        batch_size: u32,
+    ) -> String {
+        let offset_value = last_offset
+            .clone()
+            .or_else(|| self.config.initial_offset.clone())
+            .unwrap_or_default();
```
hubcio, line 516: custom query with `$offset` but no stored offset and no `initial_offset` substitutes an empty string, so `WHERE id > $offset` becomes `WHERE id > ` - invalid SQL that fails every poll, and since no rows are ever fetched no offset ever gets stored, so it never recovers. the custom-query example in the readme is exactly this shape. validate in `open()`: require `initial_offset` when the query contains a bare `$offset`.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+
+        let now = Utc::now();
+
+        query
+            .replace("$table", table)
+            .replace("$offset", &offset_value)
```
hubcio: `$offset` is substituted raw here while the built query path escapes it via `format_offset_value`. the offset value is the previous batch's tracking-column content, i.e. database row data - with a string tracking column, hostile row content becomes SQL on the next poll (second-order injection). escape it the same way as the built path, and scope the readme's "SQL injection protection" section to the built query path.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+            max_offset: None,
+        };
+        for row in rows {
+            let processed = self.process_row(&row, &table_config)?;
+
+            if let Some(pk) = processed.row_pk {
+                batch.processed_ids.push(pk);
+            }
+            if let Some(offset) = processed.max_offset {
+                batch.max_offset = Some(offset);
```
hubcio, line 827: for custom queries this ends up being the last returned row's tracking value, but nothing enforces `ORDER BY <tracking> ASC` on that path (the built query adds it at line 399). an unordered query with LIMIT saves a cursor below the true max - rows get skipped or re-emitted on the next `> $offset` poll. either inject the order-by for custom queries or make the readme require it.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        let quoted_table = quote_qualified_identifier(table)?;
+        let quoted_tracking = quote_identifier(tracking_column)?;
+
+        let base_query = format!("SELECT * FROM {quoted_table}");
+
+        let mut conditions = Vec::new();
+
+        if let Some(offset) = last_offset {
+            conditions.push(format!(
+                "{quoted_tracking} > {}",
```
hubcio, line 557: strict `>` plus LIMIT loses rows when more than `batch_size` rows share one tracking value (routine with a second-resolution timestamp tracking column): the first poll returns `batch_size` of them and sets the offset to that value, the next poll's `>` excludes the rest forever. unique auto-increment ids are safe; the documented timestamp-tracking setup is not. options: `>=` with a pk tiebreak, or document that the tracking column must be unique.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        .await?;
+
+        let mut batch = TableBatch {
+            table: table.to_string(),
+            messages: Vec::with_capacity(rows.len()),
+            processed_ids: Vec::new(),
+            max_offset: None,
+        };
+        for row in rows {
+            let processed = self.process_row(&row, &table_config)?;
```
hubcio: a deterministically failing row (non-utf8 in `text` format, non-json in `json_direct`, type mismatch) aborts the whole table batch via `?`, the offset never advances, and the same row re-fails every cycle - the table stalls forever with only a per-cycle error log. if that's the intended operator contract, document it in the readme; a per-row skip/dead-letter would be friendlier.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+
+fn build_processed_row(
+    payload: Vec<u8>,
+    max_offset: Option<String>,
+    row_pk: Option<String>,
+) -> ProcessedRow {
+    let now = Utc::now().timestamp_millis() as u64;
+    ProcessedRow {
+        message: ProducedMessage {
+            id: Some(Uuid::new_v4().as_u128()),
```
hubcio: message id is a random uuid per row, so replayed rows after a restart get new ids and downstream can't dedup them - which undercuts the offset-tracking/duplicates story in the readme. a deterministic id (hash of table + pk, say) makes replays idempotent and drops the per-row rng call.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        // leave a table marked-but-unpublished. `mark_or_delete_processed_rows`
+        // retries transient errors; a permanent failure isolates that one table
+        // (its rows stay in MySQL and are retried next cycle) while the others
+        // still flow.
+        let mut messages = Vec::new();
+        let mut state_updates: Vec<(String, String)> = Vec::new();
+        let mut total_processed: u64 = 0;
+
+        for mut batch in batches {
+            if !batch.processed_ids.is_empty()
```
hubcio, line 742: if the pk value can't be extracted (custom query not projecting it, misconfigured `primary_key_column`, or a non-scalar value - `value_as_string` returns `None` for anything but strings/numbers), `processed_ids` stays empty and the mark/delete is silently skipped while the messages still publish and the offset advances. with delete_after_read that leaks rows that look processed. worth a warn log when delete/mark is configured but no pks were extracted.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        info!(
+            "MySQL source connector with ID: {} opened successfully",
+            self.id
+        );
+        Ok(())
+    }
+
+    async fn poll(&self) -> Result<ProducedMessages, Error> {
+        let poll_interval = self.poll_interval;
+        tokio::time::sleep(poll_interval).await;
```
hubcio: the sleep is unconditional, so a backlogged table drains at `batch_size / poll_interval` max (defaults: 1000 rows / 10s = 100 rows/s) - initial backfills crawl. common pattern is to skip the sleep when the previous batch came back full.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        }
+
+        // Apply all state updates with a single lock acquisition
+        {
+            let mut state = self.state.lock().await;
+            state.processed_rows += total_processed;
+            for (table, offset) in state_updates {
+                state.tracking_offsets.insert(table, offset);
+            }
+            state.last_poll_time = Utc::now();
```
hubcio: `last_poll_time` changes every cycle and poll always returns `Some(state)`, so the runtime rewrites and fsyncs the state file every poll even when nothing was fetched. the field is also never read anywhere. dropping it and returning `None` when offsets/processed_rows are unchanged makes idle cycles free. safe to change the state shape - new connector, and invalid persisted state already falls back to a fresh start.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        }
+        "CHAR" | "VARCHAR" | "TINYTEXT" | "TEXT" | "MEDIUMTEXT" | "LONGTEXT" | "ENUM" | "SET" => {
+            let value: Option<String> = row
+                .try_get(column_index)
+                .map_err(|_| Error::InvalidRecord)?;
+            Ok(value
+                .map(serde_json::Value::String)
+                .unwrap_or(serde_json::Value::Null))
+        }
+        "BINARY" | "VARBINARY" | "TINYBLOB" | "BLOB" | "MEDIUMBLOB" | "LONGBLOB" | "GEOMETRY" => {
```
hubcio: `GEOMETRY` hits the same sqlx compatibility gate as DECIMAL/TIMESTAMP - the byte-slice impl excludes geometry, so this arm fails on every row instead of base64-encoding. rare type, but the readme lists it as supported; either handle it in the same pass as the other two or drop it from the readme.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+{
+    let mut attempts = 0;
+    loop {
+        match operation().await {
+            Ok(result) => return Ok(result),
+            Err(e) => {
+                attempts += 1;
+                if attempts >= max_retries || !is_transient_error(&e) {
+                    error!("Database operation failed after {attempts} attempts: {e}");
+                    return Err(Error::InvalidRecord);
```
hubcio: every db failure - connection loss, pool timeout, auth - surfaces as `Error::InvalidRecord`, so a dead database reads as a data problem in the upstream logs ("skipping this cycle: Invalid record"). the sdk has more accurate variants for this.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+    F: Fn() -> Fut,
+    Fut: std::future::Future<Output = Result<T, sqlx::Error>>,
+{
+    let mut attempts = 0;
+    loop {
+        match operation().await {
+            Ok(result) => return Ok(result),
+            Err(e) => {
+                attempts += 1;
+                if attempts >= max_retries || !is_transient_error(&e) {
```
hubcio: `attempts >= max_retries` makes `max_retries` actually the total attempt count: default 3 = one initial try + two retries, and `max_retries = 1` means zero retries. either the name/docs or the condition should agree.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+            return Ok(());
+        }
+
+        let quoted_table = quote_qualified_identifier(table)?;
+        let quoted_pk = quote_identifier(pk_column)?;
+
+        let ids_list = ids
+            .iter()
+            .map(|id| {
+                if id.parse::<i64>().is_ok() {
```
hubcio: the numeric check only tries `i64`, so a `BIGINT UNSIGNED` pk above `i64::MAX` gets quoted as a string and mysql compares it as a float in the IN list - precision loss can match the wrong row or none. `format_offset_value` accepts i64 or f64, so the two paths disagree; add a `u64` parse here.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+    processed_ids: Vec<String>,
+    max_offset: Option<String>,
+}
+
+impl PayloadFormat {
+    fn from_config(s: Option<&str>) -> Self {
+        match s.map(|s| s.to_lowercase()).as_deref() {
+            Some("bytea") | Some("raw") => PayloadFormat::Bytea,
+            Some("text") => PayloadFormat::Text,
+            Some("json_direct") | Some("jsonb") | Some("jsonb_direct") => PayloadFormat::JsonDirect,
```
hubcio: `jsonb`/`jsonb_direct` are postgres-isms - mysql has no jsonb. harmless, but accepting them (and the unit test locking them in) just confuses; suggest dropping the aliases.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+            .join(", ");
+
+        let query = if self.config.delete_after_read.unwrap_or(false) {
+            if self.verbose {
+                info!("Deleting {} processed rows from '{table}'", ids.len());
+            } else {
+                debug!("Deleting {} processed rows from '{table}'", ids.len());
+            }
+            format!("DELETE FROM {quoted_table} WHERE {quoted_pk} IN ({ids_list})")
+        } else if let Some(processed_col) = &self.config.processed_column {
```
hubcio, line 612: when both `delete_after_read` and `processed_column` are set, delete silently wins and the processed column only survives as a query filter. worth a validation error or at least a warn in `open()`.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        query
+            .replace("$table", table)
+            .replace("$offset", &offset_value)
+            .replace("$limit", &batch_size.to_string())
+            .replace("$now", &now.to_rfc3339())
+            .replace("$now_unix", &now.timestamp().to_string())
+    }
+
+    fn validate_custom_query(&self, query: &str) -> Result<(), Error> {
+        let query_upper = query.to_uppercase();
```
hubcio, line 531: this runs per table per poll (with a full `to_uppercase` copy of the query) for config that never changes - validate once in `open()`. that also turns the `$table`-with-no-tables case into a startup error instead of a silently skipped table, and gives a place to catch the empty-`$offset` case.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+            .try_downcast_ref::<MySqlDatabaseError>()
+            .is_some_and(|mysql_err| {
+                matches!(
+                    mysql_err.number(),
+                    // concurrency
+                    1213 | 1205 |
+                    // server unavailability
+                    1053 | 1152 | 1080 |
+                    // connection/network
+                    2006 | 2013 | 1158 | 1159 | 1160 | 1161 |
```
hubcio: 2006/2013 are client-side CR_* codes - sqlx surfaces lost connections as `sqlx::Error::Io`, and `MySqlDatabaseError::number()` only ever carries server ER_* codes, so these two can't match here. harmless, but they suggest coverage that isn't there. 1158-1161 are real server codes and fine.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        // built below would be discarded anyway.
+        if !config.payload_col.is_empty() {
+            for (i, column) in row.columns().iter().enumerate() {
+                let name = column.name();
+                if name == config.payload_col {
+                    extracted_payload =
+                        Some(self.extract_payload_column(row, i, config.payload_format)?);
+                }
+                if name == config.tracking_column {
+                    max_offset = value_as_string(&extract_column_value(row, i)?);
```
hubcio, line 856: with the default config `tracking_column == pk_column`, so this loop decodes the same column twice through `extract_column_value` (and the non-payload path at 674-679 clones the value twice through `value_as_string`). extract once and reuse when the names match.

core/connectors/sources/mysql_source/README.md
```diff
@@ -0,0 +1,382 @@
...
+## Features
+
+- **Table Polling**: Incrementally fetch data from MySQL tables using a tracking column
+- **Flexible Payload Extraction**: Extract BLOB, TEXT, or JSON columns directly as payload
+- **Custom Queries**: Use custom SQL queries with parameter substitution
+- **Delete After Read**: Automatically delete rows after processing
+- **Mark as Processed**: Mark rows as processed using a boolean column
+- **Multiple Tables**: Monitor multiple tables simultaneously
+- **Batch Processing**: Fetch data in configurable batch sizes
+- **Offset Tracking**: Keep track of processed records to avoid duplicates
```
hubcio: "avoid duplicates" oversells - the delete/mark modes are at-most-once (rows mutated before the iggy send is confirmed), and offset mode can either lose a batch on a live send failure or replay after a crash. a short delivery-semantics section would set expectations honestly.

core/connectors/sources/mysql_source/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+keywords = ["iggy", "messaging", "streaming", "mysql", "polling"]
+categories = ["command-line-utilities", "database", "network-programming"]
+homepage = "https://iggy.apache.org"
+documentation = "https://iggy.apache.org/docs"
+repository = "https://github.com/apache/iggy"
+readme = "README.md"
+publish = false
+
+[package.metadata.cargo-machete]
+ignored = ["dashmap", "simd-json"]
```
hubcio: the `simd-json` entry is redundant - machete detects the fully qualified `simd_json::` calls (random_source uses it the same way with no ignore and passes). `dashmap` genuinely needs the ignore since it's only used inside the `source_connector!` expansion. suggest `ignored = ["dashmap"]`.

core/integration/tests/connectors/mysql/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+mod mysql_source;
+
+use crate::connectors::TestMessage;
+use serde::Deserialize;
+
+const TEST_MESSAGE_COUNT: usize = 3;
```
hubcio, line 23: these consts + `DatabaseRecord` are byte-identical to the postgres test mod - could live next to `TestMessage` in the shared connectors mod. optional.

hubcio (changes requested):
a couple of notes that don't anchor to the diff:

- `core/connectors/sources/README.md` source table doesn't list mysql_source yet - worth adding a row.
- the two delivery-semantics comments below (delete/mark before send, offset advance on send failure) share one root cause - there is no send ack from the runtime back to the connector - and postgres_source behaves the same way. the sink side of the exact same gap is already tracked in #2927 (consume() return value discarded) and #2928 (offsets committed before sink processing), and #2940 discusses the broader replay-safe progress contract. the source side deserves a sibling issue rather than a fix in this PR.

core/connectors/sources/mysql_source/src/lib.rs
atharvalade: I see

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: but this might break a normal case also when query contains `$offset` and `last_offset` was there. I think checking both `initial_offset` and `last_offset` and only throw error when both are not present.

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: mysql sorts in ASC by default for queries containing `ORDER_BY`, but will make it documented in the readme.

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: will update the docs accordingly. For the tiebreak to work correctly we would have to make the pk monotonic anyway, and if an operator already has a unique monotonic pk column, there's no reason to track on a different column. This issue will be fullproof fixed with binlog CDC(once polling is done)

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: using per-row skip can cause dataloss as we are updating the offset to the last polled row, for now i have emphasised on making the error message more detailed so the operator can pin pointedly know the exact reason and fix it, documented the same in the readme.
As for the DLQ can take that as a fresh issue after going through the discussions around it.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+                continue;
+            }
+
+            total_processed += batch.messages.len() as u64;
+            messages.append(&mut batch.messages);
+            if let Some(offset) = batch.max_offset {
+                state_updates.push((batch.table, offset));
+            }
+        }
+
```
ryerraguntla, line 765: poll_tables bumps in-memory tracking_offsets before return; disk state saved only on producer.send() OK. Live send fail → next poll reads advanced offset from memory → rows skipped forever. **Fix**: platform sibling issue (#2927/#2928/#2940); document in README.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        // Phase 2: commit each table independently. A table's messages are emitted
+        // only after its rows are marked or deleted, so a failure here can never
+        // leave a table marked-but-unpublished. `mark_or_delete_processed_rows`
+        // retries transient errors; a permanent failure isolates that one table
+        // (its rows stay in MySQL and are retried next cycle) while the others
+        // still flow.
+        let mut messages = Vec::new();
+        let mut state_updates: Vec<(String, String)> = Vec::new();
+        let mut total_processed: u64 = 0;
+
```
ryerraguntla, line 726: Phase2 mark_or_delete runs before messages leave connector. delete_after_read / processed_column + send fail = row gone/filtered in MySQL, not in Iggy. Same as postgres_source. **Fix**: sibling issue + README hazard note.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+        batch_size: u32,
+    ) -> String {
+        let offset_value = last_offset
+            .clone()
+            .or_else(|| self.config.initial_offset.clone())
+            .unwrap_or_default();
+
+        let now = Utc::now();
+
+        query
```
ryerraguntla: **345 vs 377-384** — built queries use format_offset_value() (quotes string offsets); substitute_query_params drops raw $offset. String/timestamp custom query breaks or wrong comparison. Postgres same pattern. **Fix**: pipe $offset through format_offset_value.

core/connectors/sources/mysql_source/src/lib.rs
ryerraguntla: with_retry permanent/exhausted DB failure returns Error::InvalidRecord. Connection pool exhaustion (errno 1040) surfaces as row decode error. **Fix**: map infra failures to Error::Connection.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1418 @@
...
+            {
+                Ok(batch) => {
+                    if self.verbose {
+                        info!("Fetched {} rows from table '{table}'", batch.messages.len());
+                    } else {
+                        debug!("Fetched {} rows from table '{table}'", batch.messages.len());
+                    }
+                    batches.push(batch);
+                }
+                Err(error) => {
```
ryerraguntla, line 705: per-table fetch failure logged+skipped; poll() returns Ok partial batch. Intentional vs postgres abort. **Fix**: document or error metric.

ryerraguntla:
@tusharagrahari - Overall looks good. Outside of the mysql source connector, two places needs to be updated for the newer connectors.. 
1. core/connectors/sources/README.md:9-14 — mysql_source missing from Available Sources table. Fix: add row.
2. .github/workflows/_build_rust_artifacts.yml:49 — iggy_connector_mysql_source absent from connector_plugins default list; edge releases ship no prebuilt .so. Postgres listed. Fix: register crate. (

tusharagrahari:
Opened #3635 as the sibling issue for the missing runtime→source send ack, cross-linked with #2927/#2928/#2940. Left the current behavior in this PR as-is. @ryerraguntla @hubcio

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: opened a sibling issue #3635

core/connectors/sources/mysql_source/README.md
```diff
@@ -0,0 +1,424 @@
...
+
+```sql
+ALTER TABLE users ADD COLUMN is_processed BOOLEAN DEFAULT false;
+```
+
+When `processed_column` is set, the connector automatically adds a `WHERE is_processed = FALSE` filter to the polling query, so only unprocessed rows are fetched.
+
+### Delivery semantics
+
+Both `delete_after_read` and `processed_column` mutate MySQL during the poll, before the batch is actually sent to Iggy. If the connector crashes or the send fails in that window, those rows are gone or marked processed anyway — they won't be retried. So both options are at-most-once, not at-least-once.
```
ryerraguntla: >> those rows are gone or marked processed anyway - will it be deleted (gone) or marked? It should be either one in each pass, could this be confirmed and update the documentation

core/connectors/sources/mysql_source/README.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Payload Column Extraction
+
+When `payload_column` is set, the connector extracts that column directly as the Iggy message payload. The `payload_format` option determines how the column is read:
+
+| Format | Column Type | Schema | Description |
+| ------ | ----------- | ------ | ----------- |
+| `bytea` / `raw` | `BLOB`, `BINARY`, `VARBINARY` | `raw` | Raw bytes passthrough |
+| `text` | `TEXT`, `VARCHAR` | `text` | UTF-8 text |
+| `json_direct` / `jsonb` | `JSON` | `json` | JSON object serialized to bytes |
```
ryerraguntla, line 116: lib.rs:110-116 — README lists jsonbalias; code only acceptsjson_direct. **Fix**: add jsonb to from_config or drop from README.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1563 @@
...
+        // built below would be discarded anyway.
+        if !config.payload_col.is_empty() {
+            for (i, column) in row.columns().iter().enumerate() {
+                let name = column.name();
+                if name == config.payload_col {
+                    extracted_payload =
+                        Some(self.extract_payload_column(row, i, config.payload_format)?);
+                }
+                if name == config.tracking_column || name == config.pk_column {
+                    let as_string = value_as_string(&extract_column_value(row, i)?);
```
ryerraguntla: value_as_string only String|Number; BOOL/NULL/JSON tracking/PK ⇒ no offset/PK for mark. **Fix**: canonical scalar encoding or open-time schema probe.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1563 @@
...
+        ids: &[String],
+    ) -> Result<(), Error> {
+        if ids.is_empty() {
+            return Ok(());
+        }
+
+        let quoted_table = quote_qualified_identifier(table)?;
+        let quoted_pk = quote_identifier(pk_column)?;
+
+        let ids_list = ids
```
ryerraguntla: BINARY PK base64 in JSON path; IN clause uses quoted string literal ⇒ mark/delete miss. **Fix**: hex/binary bind or typed PK encoder.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1563 @@
...
+        // delete_after_read/processed_column work, not something this loop fixes.
+        let mut messages = Vec::new();
+        let mut state_updates: Vec<(String, String)> = Vec::new();
+        let mut total_processed: u64 = 0;
+
+        let mark_or_delete = self.config.delete_after_read.unwrap_or(false)
+            || self.config.processed_column.is_some();
+
+        for mut batch in batches {
+            if mark_or_delete && !batch.messages.is_empty() && batch.processed_ids.is_empty() {
```
ryerraguntla, line 731: mark/delete configured but empty processed_ids: warn then still emit + advance offset. **Fix**: fail batch when mark configured without scalar PK.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1563 @@
...
+        last_offset: &Option<String>,
+        batch_size: u32,
+    ) -> String {
+        let offset_value = last_offset
+            .clone()
+            .or_else(|| self.config.initial_offset.clone())
+            .unwrap_or_default();
+        let offset = format_offset_value(&offset_value);
+        let now = Utc::now();
+
```
ryerraguntla, line 519: lib.rs:399 vs lib.rs:424  - custom_query $table unquoted; built-in path quotes. **Fix**: quote_qualified_identifier on $table (postgres same pattern).

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1563 @@
...
+    fn build_polling_query(
+        &self,
+        table: &str,
+        tracking_column: &str,
+        last_offset: &Option<String>,
+        batch_size: u32,
+    ) -> Result<String, Error> {
+        let quoted_table = quote_qualified_identifier(table)?;
+        let quoted_tracking = quote_identifier(tracking_column)?;
+
```
ryerraguntla, line 550: 748-791,427 — hot path: per-row serde_json Map + SELECT *. **Fix**: column projection + direct serialize (v1 acceptable, matches postgres shape).

core/connectors/sources/mysql_source/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+[package.metadata.cargo-machete]
+ignored = ["dashmap"]
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+chrono = { workspace = true }
```
ryerraguntla, line 41: dashmap dep unused in lib.rs. Validate and **Fix**: remove dep + machete ignore.

ryerraguntla:
No major issues. Most of them are suggestions/documentation. Should be good to merge in the next iteration

core/connectors/sources/mysql_source/README.md
tusharagrahari: it's a mutually-exclusive if / else if exactly one mutation per row per pass. `delete_after_read` = true deletes the row, otherwise `processed_column` marks it. `delete_after_read` takes precedence when both are set (there's also a startup warning for that case). Never both. Updated the doc too.

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: added open-time schema probe for tracking_column, PK already is DB-enforced.

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: implemented the typed PK encoder, added a `PKValue` enum, introduced a new function to extract the pk so that base64 encoding doesn't affect the mark/delete func. Also added bind logic inplace of splicing string literals in mark/delete func.

core/connectors/sources/mysql_source/Cargo.toml
tusharagrahari: dashmap is required by source_connector! macro expansion, cargo-machete can't see through macro expansion, thats why machete ignore.

core/connectors/sources/mysql_source/src/lib.rs
slbotbm, line 1: `MySqlSource::validate_tracking_column` returns immediately whenever `custom_query` is configured (`lib.rs:349`). Later, `process_row` permits a row without the configured tracking column, leaves `max_offset` as `None`, and still emits the message. A query that omits or aliases the tracking column therefore replays the same result on every poll while never updating state.

Validate each custom-query result contains a non-null, supported tracking value, or fail the table batch before emitting anything. The open-time schema probe did not fully address the earlier [comment](https://github.com/apache/iggy/pull/3568#discussion_r3603820903) because custom queries explicitly bypass that probe.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1721 @@
...
+        config: &RowProcessingConfig,
+    ) -> Result<ProcessedRow, Error> {
+        let mut row_pk: Option<PkValue> = None;
+        let mut max_offset: Option<String> = None;
+        let mut extracted_payload: Option<Vec<u8>> = None;
+
+        // Payload column set: only extract it plus tracking/pk columns.
+        // Avoids extract_column_value on every other column since the data map
+        // built below would be discarded anyway.
+        if !config.payload_col.is_empty() {
```
slbotbm, line 848: If the configured column is absent from a custom query result or is misspelled, `extracted_payload` stays `None`; the fallback at `lib.rs:864` serializes the entire row as JSON. The batch schema still comes from `payload_format`, so a `bytea` or `text` source can silently publish JSON bytes, while `json_direct` publishes the wrong JSON shape.

Treat a missing configured payload column as an error. Falling back is only valid when `payload_column` was not configured.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1721 @@
...
+        _ => None,
+    }
+}
+
+fn build_processed_row(
+    table: &str,
+    payload: Vec<u8>,
+    max_offset: Option<String>,
+    row_pk: Option<PkValue>,
+) -> ProcessedRow {
```
slbotbm, line 1134: `build_processed_row` uses `Utc::now().timestamp_millis()` for both `timestamp` and `origin_timestamp`. Iggy message timestamps are microseconds; the core initializes them with `IggyTimestamp::now().as_micros()`, and the InfluxDB sources use `timestamp_micros()`. The current value is 1,000 times too small and will be interpreted as a date near January 1970.

Use `timestamp_micros()` or leave the fields unset so the normal Iggy timestamping path applies. Add an assertion for timestamp units to the integration coverage.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1721 @@
...
+            {
+                Ok(batch) => {
+                    if self.verbose {
+                        info!("Fetched {} rows from table '{table}'", batch.messages.len());
+                    } else {
+                        debug!("Fetched {} rows from table '{table}'", batch.messages.len());
+                    }
+                    // A full batch means the LIMIT was hit, so more rows are waiting.
+                    any_batch_full |= batch.messages.len() as u32 >= batch_size;
+                    batches.push(batch);
```
slbotbm, line 703: There is no positive-value validation for `batch_size`. With zero, MySQL returns an empty `LIMIT 0` result, but the full-batch check evaluates `0 >= 0` as true. `poll` then skips its pacing sleep forever because `last_batch_full` remains true, producing a CPU/database hot loop that can never emit a row.

Reject zero during initialization.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,1721 @@
...
+        // offset to fix.
+        let mut batches: Vec<TableBatch> = Vec::with_capacity(self.config.tables.len());
+        let mut any_batch_full = false;
+
+        for table in &self.config.tables {
+            // Get last offset with minimal lock time
+            let last_offset = {
+                let state = self.state.lock().await;
+                state.tracking_offsets.get(table).cloned()
+            };
```
slbotbm, line 683: This loops over every configured table regardless of whether `custom_query` contains `$table`. With multiple tables and a static custom query, the identical SQL is executed once per table and its rows are emitted multiple times with different table-derived metadata/message IDs. Validation only rejects `$table` when the table list is empty; it does not reject this ambiguous combination.

Require `$table` when more than one table is configured, or execute a static custom query exactly once.

slbotbm (changes requested):

core/connectors/sources/mysql_source/src/lib.rs
tusharagrahari: `!config.payload_col.is_empty()` block now handles and fails that table for that cycle (per table basis), when `extracted_payload.is_none()` even when `payload_col` is some. Also added integration test, where `payload_col` is set but is not present in the table.

tusharagrahari:
@ryerraguntla @hubcio can you pls reopen this pr.

core/connectors/sources/mysql_source/src/lib.rs
```diff
@@ -0,0 +1,3317 @@
...
+
+            // Read by position: MySQL returns information_schema column names
+            // uppercased (IS_NULLABLE, ...), so try_get by lowercase name misses.
+            let is_nullable: String = row.try_get(0).map_err(|e| {
+                Error::InitError(format!(
+                    "failed to read is_nullable for table '{table}': {e}"
+                ))
+            })?;
+            let data_type: String = row.try_get(1).map_err(|e| {
```
ryerraguntla: cursor-offset (state.tracking_offsets) committed unconditionally inside poll_tables, before the batch reaches Iggy or state is persisted. Hits every deployment, not just delete/mark-configured ones. A single failed send silently and permanently skips those rows on the next poll's WHERE tracking_column > offset, no crash needed.   **Fix**: stage, commit only on Ack.

PR #3569, opened 2026-06-26
feat(bdd): unify rust image versions to slim-trixie

slbotbm:
Changes rust-based images to slim-trixie version.

As an aside, there are a variety of versions (bookworm, bookworm-slim, trixie, trixie-slim, alpine) being used throughout the repo. Maybe it would be better to choose one version and utilize that everywhere?


hubcio:
do you think that it woudl be worth to add some machinery that would sync / check if its synced? maybe either reuse `sync-rustc-version.sh` or a new script

bdd/java/Dockerfile
```diff
@@ -28,7 +28,7 @@
 # fully endorsed by the ASF.
 
 ARG RUST_VERSION=1.96
-FROM rust:${RUST_VERSION}
+FROM rust:${RUST_VERSION}-slim-trixie
```
mmodzelewski: actually, can we just drop this? it's directly followed by the gradle image, so there's no use of rust at all

mmodzelewski (changes requested):

bdd/java/Dockerfile
slbotbm: done

foreign/python/.devcontainer/Dockerfile
```diff
@@ -27,7 +27,7 @@
 # or stability of the code, it does indicate that the project has yet to be
 # fully endorsed by the ASF.
 
-FROM mcr.microsoft.com/devcontainers/python:1-3.10-bullseye
+FROM mcr.microsoft.com/devcontainers/python:1-3.10-trixie
```
mmodzelewski: this version does not exists, you can use `3.10-trixie` instead

bdd/Dockerfile
mmodzelewski, line 1: I think this file got stale, and it's no longer in use and can be removed. The `docker-compose` files reference `Dockerfile` from `core/server` directly. @hubcio correct me if I'm wrong
Otherwise, switching from alpine to debian would require some commands changes.

mmodzelewski (changes requested):

foreign/python/.devcontainer/Dockerfile
slbotbm: To be honest, I was wondering whether a dev container is really required for python? This is the only folder that has a dev container, as far as I can tell, and currently, we use uv to manage all python versions and dependencies. I was thinking it would be better to delete this. What do you think?

mmodzelewski: I haven't used it for a while now, not sure if anyone else is using it. Yeah, with uv that should be easier anyway than when I created this dev container. Anyways, I'm OK with dropping it

slbotbm: done

bdd/Dockerfile
slbotbm: done

mmodzelewski (approved):

thepatrykk (approved):

PR #3574, opened 2026-06-27
feat(connectors): retry transient Doris Stream Load failures in-request

ryankert01:

Relates to #3215

## Rationale

Follow-up to the Doris sink (#3215). The connector already classified transient Stream Load outcomes as retryable but had no retry path, so under the runtime's at-most-once delivery a transient backend blip silently dropped the batch.

## What changed?

The Doris sink classified transient Stream Load outcomes (`5xx`/`408`/`429`, transport errors, `Publish Timeout`) as `CannotStoreData` but never retried them: the runtime commits the consumer offset at poll time before `consume()` runs and discards its return value, so a transient failure dropped the batch with no replay.

`consume()` now retries a transiently-failed batch in-request via a new `load_batch()` that wraps send plus status classification, re-PUTing under the same deterministic label so Doris dedupes a prior attempt that actually landed (e.g. a `2xx` whose body could not be read). Permanent failures (`4xx`, `Fail`, schema/redirect problems) are never retried. Backoff and jitter come from `iggy_connector_sdk::retry`, bounded by new `max_retries`/`retry_delay`/`max_retry_delay` config (defaults 3 / 200ms / 5s). This shrinks the at-most-once window within a single poll; cross-poll and crash delivery stay a runtime concern.

## Local Execution

- Passed
- Pre-commit hooks: checks run manually. The `license-headers` hook cannot execute on this machine (its script needs bash 4+ `mapfile`; only bash 3.2 is present), but `hawkeye check` passes directly and CI enforces it. `markdownlint`, `taplo`, `cargo fmt`, `cargo clippy -p iggy_connector_doris_sink --all-targets -- -D warnings`, and `cargo test -p iggy_connector_doris_sink` (41 tests) all pass.

## AI Usage

1. Claude Code (Anthropic).
2. Implemented the retry loop, config plumbing, tests, and README/config updates, after verifying the runtime's at-most-once delivery semantics directly in the runtime source.
3. Three new wiremock unit tests pin the behavior: transient-then-success (retry fires), exhausted-budget (exact attempt count via `.expect`), and permanent-not-retried. Full crate suite, clippy, and doc lint pass locally.
4. Yes.


core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -312,6 +340,51 @@ impl DorisSink {
...
+    /// Load one batch with bounded in-request retry. `send_stream_load` performs
+    /// a single full FE -> BE attempt; this wraps it (plus status
+    /// classification) so a *transient* failure re-PUTs under the same `label`.
+    /// The runtime commits the consumer offset before consume() runs, so this is
+    /// the connector's only redelivery path on a transient outage; the shared
+    /// label lets Doris dedupe a prior attempt that actually landed (e.g. a 2xx
+    /// whose body we couldn't read). A `PermanentHttpError` returns immediately —
+    /// retrying bad data would just hammer the FE.
+    async fn load_batch(&self, label: &str, body: Bytes) -> Result<StreamLoadResponse, Error> {
+        let connected = self.connected.as_ref().ok_or_else(|| {
```
hubcio, line 360: load_batch re-runs the same `connected.as_ref().ok_or_else(|| Error::InitError(...))` none-check that `send_stream_load` already does. the binding is needed here (you read `max_retries`/`retry_delay`/`retry_max_delay` off it), but the none-guard itself is redundant since `send_stream_load` re-checks on every attempt. minor - could drop the duplicate guard.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -312,6 +340,51 @@ impl DorisSink {
...
+            Error::InitError(format!(
+                "Doris sink ID {} called before open() — not connected",
+                self.id
+            ))
+        })?;
+
+        let mut attempt = 0u32;
+        loop {
+            let result = match self.send_stream_load(label, body.clone()).await {
```
hubcio: the two stacked matches here collapse into one - the first is exactly what `and_then` does:

```rust
let error = match self
    .send_stream_load(label, body.clone())
    .await
    .and_then(|response| classify_status(&response).map(|()| response))
{
    Ok(response) => return Ok(response),
    Err(error) => error,
};
```

same semantics, drops the intermediate `result` binding.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -312,6 +340,51 @@ impl DorisSink {
...
+                Ok(response) => return Ok(response),
+                Err(error) => error,
+            };
+
+            attempt += 1;
+            if attempt >= connected.max_retries || !is_transient_error(&error) {
+                return Err(error);
+            }
+
+            let delay = jitter(exponential_backoff(
```
hubcio, line 385: backoff exponent is `attempt - 1`, so the first retry waits `retry_delay` (2^0). the sdk's own callers (`HttpRetryMiddleware`, `check_connectivity_with_retry`) pass the already-incremented `attempts`, so their first retry waits `2 * retry_delay`. not wrong - arguably better - but it's a divergent convention across the connectors. worth a one-line comment on why, or align with the siblings.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -484,6 +557,12 @@ fn effective_batch_size(configured: Option<u32>) -> usize {
     configured.unwrap_or(DEFAULT_BATCH_SIZE).max(1) as usize
 }
 
+/// Total Stream Load attempts per batch: the configured value floored at 1 so
+/// `0` (or an unset value) collapses to a single attempt with no retry.
```
hubcio: this comment is off for the unset case. `configured.unwrap_or(DEFAULT_MAX_RETRIES).max(1)` maps an unset value to the default 3, not to a single attempt - only `Some(0)` collapses to 1. so `0` disables retries, unset gives you 3. worth fixing the comment (the config docs also only mention `1` disables, not `0`).

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -85,6 +95,11 @@ struct Connected {
     // off `self` instead of threading it through every call.
     allow_insecure_redirect: bool,
     allowed_redirect_hosts: Option<Vec<String>>,
+    // In-request retry policy for transient Stream Load failures, resolved once
+    // at `open()`. `max_retries` is the total attempt count (1 = no retry).
+    max_retries: u32,
+    retry_delay: Duration,
+    retry_max_delay: Duration,
```
hubcio: one knob, three spellings: the config key is `max_retry_delay` (line 137) but the field here is `retry_max_delay` and the const is `DEFAULT_RETRY_MAX_DELAY` (line 70). the config key is wire-facing so keep it - align the field + const to `max_retry_delay` / `DEFAULT_MAX_RETRY_DELAY` so it reads as one thing.

core/connectors/sinks/doris_sink/README.md
```diff
@@ -84,16 +88,16 @@ timeout = "30s"
 
 ## Operational guidance
 
-- **`label_keep_max_second`.** Idempotent replay relies on Doris retaining each label for at least as long as it could take the Iggy runtime to redrive a failed batch. The Doris default is 3 days, which is conservative. If you set this lower on the Doris side, make sure your runtime retry budget fits inside the window — once a label expires, a replay re-loads instead of deduping, producing duplicate rows.
+- **`label_keep_max_second`.** The connector's in-request retry re-PUTs a transiently-failed batch under the same label, so Doris must retain that label for at least the connector's full retry budget for the replay to dedupe. The Doris default is 3 days, which is conservative. If you set this lower on the Doris side, make sure `max_retries × max_retry_delay` fits inside the window — once a label expires, a retry re-loads instead of deduping, producing duplicate rows.
```
hubcio: the `max_retries * max_retry_delay` budget here understates the real worst case. a slow BE can burn most of the request `timeout` before the batch returns a retryable error (a 2xx whose body read stalls, or a BE slow to respond) - with defaults that 30s timeout is larger than the 5s max_retry_delay. jitter also adds up to +20% on top of each capped delay. closer bound is `max_retries * (timeout + max_retry_delay)`.

core/connectors/sinks/doris_sink/README.md
```diff
@@ -84,16 +88,16 @@ timeout = "30s"
...
-  The runtime commits the consumer offset for the whole poll before `consume()` runs, so a failed chunk is not replayed regardless; pushing the other chunks through maximizes delivered data, and the worst error is surfaced at the end (logged at `error!` for observability — the runtime currently discards `consume()`'s return value, so there is no retry or DLQ).
+- **Multi-chunk batches are best-effort for operational failures.** A poll larger than `batch_size` is split into chunks, each loaded as its own labelled Stream Load (with its own in-request retry budget for transient failures). If a chunk still fails after its retries (serialize, HTTP, or status-classification error), the connector keeps the worst error, attempts the remaining chunks, and returns that error at the end — it does **not** stop at the first such failure.
```
hubcio: says the connector "keeps the worst error" (and "worst error is surfaced" just below), but `consume()` keeps the first error via `first_error.get_or_insert(...)` - there's no severity comparison anywhere, and the comment right above the loop even says "keep the first error". harmless since the runtime discards the return and every chunk error is logged, but the wording should say first, not worst.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -312,6 +340,51 @@ impl DorisSink {
...
+
+    /// Load one batch with bounded in-request retry. `send_stream_load` performs
+    /// a single full FE -> BE attempt; this wraps it (plus status
+    /// classification) so a *transient* failure re-PUTs under the same `label`.
+    /// The runtime commits the consumer offset before consume() runs, so this is
+    /// the connector's only redelivery path on a transient outage; the shared
+    /// label lets Doris dedupe a prior attempt that actually landed (e.g. a 2xx
+    /// whose body we couldn't read). A `PermanentHttpError` returns immediately —
+    /// retrying bad data would just hammer the FE.
+    async fn load_batch(&self, label: &str, body: Bytes) -> Result<StreamLoadResponse, Error> {
```
hubcio, line 359: the backoff/jitter reuse from `iggy_connector_sdk::retry` is good. the retry loop itself is now the third hand-written copy of the same skeleton (`HttpRetryMiddleware::handle` and `check_connectivity_with_retry` in the sdk both have it). you can't reuse the middleware here - doris signals failure as http 200 + `{"Status":"Fail"}` which it never inspects - but that's really an argument for the sdk exposing a generic `retry_async(op, policy)` the three call sites could share. non-blocking, sdk-level follow-up.

hubcio (changes requested):
overall this looks solid, most of below comments are just nits.

one thing not on a changed line so flagging it here: the retry loop is uncancellable by graceful shutdown. the runtime consumer task's `tokio::select!` only races the shutdown signal against `consumer.next()` - once a batch is taken, `process_messages().await` (which runs the sink's `consume()` -> retry loop) runs to completion with no shutdown arm. so a `consume()` stuck retrying against a down FE blocks shutdown for the whole budget - up to `max_retries * (timeout + backoff)` per chunk, tens of seconds to minutes with defaults. only blocks that one consumer's task, and it's not a deadlock. it's runtime-side (not this PR), but this PR stretches the window, so worth a line in operational guidance and maybe a follow-up to make the retry sleep shutdown-aware.

ryankert01:
Updated, sorry for the late reply tho. (I was on a conference trip)

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -599,31 +689,58 @@ fn validate_identifier(name: &str, field: &str, id: u32) -> Result<(), Error> {
     Ok(())
 }
 
-fn classify_status(response: &StreamLoadResponse) -> Result<(), Error> {
+/// Transient Stream Load failures worth an in-request retry: HTTP 5xx/408/429,
+/// transport errors, and duplicate labels whose existing Doris job is `RUNNING`
+/// or `CANCELLED`. `PermanentHttpError` (4xx, "Fail", schema/redirect problems,
+/// unparsable body) is never retried — re-PUTing bad data just hammers the FE.
+fn is_transient_error(error: &Error) -> bool {
```
hubcio, line 724: `CannotStoreData` is one of the two transient markers for the retry loop, but `consume()` also wraps a local serialize failure in the same variant (the `simd_json::to_vec` error arm) - that one never goes through `load_batch`, so it's never retried. harmless since the runtime collapses the return to an i32, just misleading to have the retryable-class variant on a deterministic local failure. a permanent-class variant or a short comment there would make the split cleaner.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -484,6 +564,12 @@ fn effective_batch_size(configured: Option<u32>) -> usize {
     configured.unwrap_or(DEFAULT_BATCH_SIZE).max(1) as usize
 }
 
+/// Total Stream Load attempts per batch. An unset value uses the default;
+/// configured `0` or `1` both mean one attempt with no retry.
+fn effective_max_retries(configured: Option<u32>) -> u32 {
```
hubcio, line 574: floors at 1 but no ceiling. with the readme's own per-chunk bound (`N * 6 * timeout + (N-1) * max_retry_delay`), `max_retries = 100` at the default 30s timeout works out to ~5h of blocking per chunk against a down FE - and per the graceful-shutdown note, close waits for all of it. a warn above some sane cap (say 10) would match the spirit of the `retry_delay > max_retry_delay` clamp in `open()`.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -575,8 +665,8 @@ fn truncate_for_log(s: &str, max_bytes: usize) -> String {
 
 fn parse_stream_load_response(body: &str) -> Result<StreamLoadResponse, Error> {
```
hubcio, line 684: asymmetry with the body-read-failure path: a 2xx whose body can't be read is transient (retry re-PUTs, label dedupe reveals the real outcome), but a 2xx that reads fine yet is empty or garbage is permanent. a proxy stripping the body to empty masks a committed load the same way a mid-stream reset does, and the same retry-under-same-label argument applies. the proxy-html / schema-change case is genuinely permanent, so maybe just the empty-body subcase deserves the transient branch. deliberate?

hubcio (changes requested):

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: send_stream_load now takes &Connected as a parameter and can't even be called without one

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: added

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Agreed, filed #3702 to track a shared `retry_async(op, policy)` in the SDK. Keeping the hand-rolled loop here since the helper touches the two existing SDK call sites and belongs in its own PR.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Good catch. I changed local JSON serialization failures to `Error::Serialization`, so they are no longer represented by the transient `CannotStoreData` variant.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: Agreed. The sink now emits a startup warning when the effective `max_retries` exceeds 10, and the shutdown-delay risk is documented in the README and example configurations.

core/connectors/sinks/doris_sink/src/lib.rs
ryankert01: yeah. update to ->  An empty 2xx response is  transient and retries the identical request under the same Doris label.

ryerraguntla:
I reviewed once. Need more time for me to make a notes for my self for each of the error state and handling of the duplicate/non duplicate label. Interesting scenarios.

hubcio (approved):

spetz (approved):

Issue #3522, opened 2026-06-21
[python sdk] Add functions related to consumer group management


slbotbm:
### Description

The Python SDK currently exposes consumer_group(...) only as a high-level helper for building and initializing an
  IggyConsumer, but it does not expose the Rust SDK’s consumer-group management APIs for inspecting or creating groups.

The missing functions are `get_consumer_group`, `get_consumer_groups`, `create_consumer_group`, `delete_consumer_group`, `join_consumer_group`, and `leave_consumer_group` . These asynchronous functions should be added to foreign/python/src/client.rs, wrapping the underlying rust sdk function using PyO3. 

These should be tested end-to-end extensively, and the tests should be added to foreign/python/tests/test_consumer_group.py.

The six functions should be split into two PRs, each of which has implementations for 3 functions + tests.

The stubs in apache_iggy.pyi should be properly generated without any mangling.

Before opening a PR, please ensure that all pre-commit hooks pass.

### Affected area / component

Python SDK

### Contribution

- [ ] I'm willing to submit a pull request to implement this feature

### Good first issue

- [x] I think this could be a good first issue for a new contributor

WaterWhisperer:
I'd like to work on this.

mmodzelewski:
@WaterWhisperer go ahead, you're assigned.

hubcio:
@WaterWhisperer @slbotbm  do you think that this is done now?

slbotbm:
@hubcio it is not done. still requires 3 more functions to be created by @WaterWhisperer

WaterWhisperer:
> [@WaterWhisperer](https://github.com/WaterWhisperer) [@slbotbm](https://github.com/slbotbm) do you think that this is done now?

not fully completed yet

PR #3572, opened 2026-06-26
feat(python): add topic listing, update, delete and purge

mattp5657:
## Which issue does this PR address?

Closes #3521

## Rationale

Completes the Python SDK's topic management by adding the missing list, update, delete, and purge operations, which previously had no Python binding and forced callers to another SDK.

## What changed?

The Python SDK exposed only `create_topic` and `get_topic`, so listing, updating, deleting, or purging topics required falling back to another SDK. `TopicDetails` also hid `compression_algorithm` and `replication_factor`, leaving the result of an update impossible to assert from Python.

`get_topics`, `update_topic`, `delete_topic`, and `purge_topic` now bind through to the Rust `TopicClient`, a new `Topic` type backs the list view, and `TopicDetails` exposes the two previously hidden fields. The new methods take `stream_id` to stay consistent with the Rust trait and the existing `get_topic`.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

Claude was used to help generate and review this PR.

PR #3581, opened 2026-06-29
feat(python): expose consumer group create and get APIs

WaterWhisperer:

Relates to #3522

## Rationale

The Python SDK should expose the first consumer group management APIs requested in the issue.

## What changed?

The Python SDK only exposed `consumer_group(...)` as a high-level consumer helper, so Python users could not directly create or inspect consumer groups.

This exposes `create_consumer_group`, `get_consumer_group`, and `get_consumer_groups`, adds the required PyO3 DTO wrappers and generated stubs, and covers create/get, lookup, missing, and list behavior in `test_consumer_group.py`.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Tools: Codex
2. Scope: issue analysis, test refinement
3. Verification: Python SDK test suite
4. Yes


foreign/python/src/consumer_group.rs
slbotbm, line 1: a separate consumer group file is not needed. Inline all of the code in this into consumer.rs

foreign/python/src/consumer_group.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub fn partitions_count(&self) -> u32 {
+        self.inner.partitions_count
+    }
+
+    #[getter]
+    pub fn members_count(&self) -> u32 {
+        self.inner.members_count
+    }
+
+    #[getter]
```
slbotbm: Is the getter tag truly useful for this function? I think this would error if used as it is.

slbotbm (changes requested):
Took a quick look. Other than the comments, I would like you to look at #3572 and add docs for all of the functions you have added in the way that docs have been written there. The ways of testing your code are also not really enough, so I would like you to take a look at how the code is being tested there, and in general in the sdk, and come up with more patterns. When creating tests, with regards to assertions, there are two things to keep in mind:
- Assert all of the things you give as user input
- Assert all of the fields that the server sets deterministically and are stable across runs.

Once these are addressed, I'll review more thoroughly.

foreign/python/src/consumer_group.rs
WaterWhisperer: I initially mirrored the existing Rust/C++ consumer group grouping. I'll update it.

foreign/python/src/consumer_group.rs
WaterWhisperer: Kept it because `members` is part of `ConsumerGroupDetails`

foreign/python/src/client.rs
```diff
@@ -243,6 +246,107 @@ impl IggyClient {
         })
     }
 
+    /// Create a consumer group for a stream topic.
+    ///
```
slbotbm: > /// Create a consumer group for a stream topic.

/// Create a consumer group for a stream and topic.

foreign/python/src/client.rs
```diff
@@ -243,6 +246,107 @@ impl IggyClient {
...
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     group_id: Consumer group identifier as `str | int`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `ConsumerGroupDetails | None`.
+    ///
+    /// Raises:
+    ///     PyRuntimeError: If an identifier is invalid or the request fails.
```
slbotbm: /// Retrieve details for a consumer group from the specified stream and topic.
///
/// Args:
///     stream_id: Stream identifier as `str | int`.
///     topic_id: Topic identifier as `str | int`.
///     group_id: Consumer group identifier as `str | int`.
///
/// Returns:
///     An awaitable that resolves to `ConsumerGroupDetails` if the consumer group exists,
///     or `None` otherwise.
///
/// Raises:
///     PyRuntimeError: If the request fails.

Let's use the above doc

foreign/python/src/client.rs
```diff
@@ -243,6 +246,107 @@ impl IggyClient {
...
+            let group = inner
+                .get_consumer_group(&stream_id, &topic_id, &group_id)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(group.map(PyConsumerGroupDetails::from))
+        })
+    }
+
+    /// Get all consumer groups in a stream topic.
+    ///
```
slbotbm: > /// Get all consumer groups in a stream topic.

/// Get all consumer groups for the specified stream and topic

foreign/python/src/client.rs
```diff
@@ -243,6 +246,107 @@ impl IggyClient {
...
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     name: Consumer group name as `str`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the consumer group is created.
+    ///
+    /// Raises:
+    ///     PyRuntimeError: If an identifier is invalid or the request fails.
```
hubcio: this `Raises:` (and the same in `get_consumer_group` at :293, `get_consumer_groups` at :326, plus the generated `.pyi`) says `PyRuntimeError` for an invalid identifier, but an invalid id comes from `Identifier::try_from`, which raises `PyValueError` (`identifier.rs:42`/`:45`) synchronously at call time before the awaitable is even returned. `ValueError` isn't a subclass of `RuntimeError`, so a caller doing `except RuntimeError` misses it - reachable with an empty string id. worth splitting into two entries: `PyValueError` for a bad identifier, `PyRuntimeError` for a request failure, then regen the stub.

foreign/python/src/client.rs
```diff
@@ -28,7 +28,10 @@ use pyo3_stub_gen::derive::{gen_stub_pyclass, gen_stub_pymethods};
 use std::str::FromStr;
 use std::sync::Arc;
 
-use crate::consumer::{AutoCommit, IggyConsumer, py_delta_to_iggy_duration};
+use crate::consumer::{
+    AutoCommit, IggyConsumer, py_delta_to_iggy_duration,
+    ConsumerGroup as PyConsumerGroup, ConsumerGroupDetails as PyConsumerGroupDetails,
```
hubcio: this import block isn't rustfmt-clean - `rustfmt --edition 2024 --check` reorders it (the `ConsumerGroup`/`ConsumerGroupDetails` imports sort up before `IggyConsumer`). it slips through because `foreign/python` is in the workspace `exclude`, so `cargo fmt --all` skips the crate and there's no separate rustfmt gate for it the way `foreign/php` has one. running rustfmt fixes the block, and it'd be worth adding a fmt check scoped to `foreign/python` so this stops shipping uncaught.

foreign/python/src/consumer.rs
```diff
@@ -216,6 +220,130 @@ impl IggyConsumer {
...
+    }
+}
+
+#[gen_stub_pyclass]
+#[pyclass]
+pub struct ConsumerGroupMember {
+    pub(crate) inner: RustConsumerGroupMember,
+}
+
+impl From<RustConsumerGroupMember> for ConsumerGroupMember {
```
hubcio: this owned `From<RustConsumerGroupMember>` is dead - the only conversion site is `members()` at :299 which maps over `.iter()`, so it binds the borrowed `From<&RustConsumerGroupMember>` at :316. the sibling owned impls (`ConsumerGroup` :229, `ConsumerGroupDetails` :265) are actually used, just not this one, and since it's a trait impl `dead_code` won't flag it. safe to delete these 5 lines. keep the borrowed impl's manual field rebuild though - `RustConsumerGroupMember` isn't `Clone`, so `member.clone()` isn't an option there.

foreign/python/src/client.rs
```diff
@@ -243,6 +246,107 @@ impl IggyClient {
...
+        let stream_id = Identifier::try_from(stream_id)?;
+        let topic_id = Identifier::try_from(topic_id)?;
+        let inner = self.inner.clone();
+
+        future_into_py(py, async move {
+            inner
+                .create_consumer_group(&stream_id, &topic_id, &name)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(())
```
hubcio, line 443: optional: `create_consumer_group` drops the `ConsumerGroupDetails` the SDK returns and resolves to `None`, so a caller who wants the server-assigned id has to follow up with `get_consumer_group`. this matches how the `create_stream`/`create_topic` bindings already behave so it's not a regression - just flagging it if a return-the-details pass ever happens across the whole create-family.

hubcio (changes requested):

foreign/python/src/client.rs
WaterWhisperer: Thanks!

foreign/python/src/client.rs
WaterWhisperer: > it'd be worth adding a fmt check scoped to `foreign/python` so this stops shipping uncaught.

Sure

foreign/python/src/client.rs
WaterWhisperer: > this matches how the `create_stream`/`create_topic` bindings already behave so it's not a regression

yeah, it's just to keep it consistent  here

hubcio (dismissed):

foreign/python/tests/test_consumer_group.py
```diff
@@ -34,8 +34,411 @@
...
+            stream_name,
+            topic_name,
+            second_group_name,
+        )
+
+        groups = await iggy_client.get_consumer_groups(stream_name, topic_name)
+
+        assert len(groups) == 2
+        groups_by_name = {group.name: group for group in groups}
+        assert set(groups_by_name) == {first_group_name, second_group_name}
```
slbotbm: If get_consumer_groups returns the groups in a deterministic order every time it is called, then that order should be asserted here instead of using a set.

foreign/python/tests/test_consumer_group.py
```diff
@@ -34,8 +34,411 @@
...
+                topic_name,
+                group_name,
+            )
+
+        first = await iggy_client.get_consumer_groups(stream_name, topic_name)
+        second = await iggy_client.get_consumer_groups(stream_name, topic_name)
+
+        assert {group.name for group in first} == group_names
+        assert {group.name for group in second} == group_names
+        assert {group.id for group in first} == {group.id for group in second}
```
slbotbm: Same here as well. Assert the order of return as well as what was returned.

slbotbm (changes requested):
Let's also add docs for all of the getters you created as well.

foreign/python/tests/test_consumer_group.py
WaterWhisperer: All updated

slbotbm (approved):
lgtm

hubcio:
@WaterWhisperer once you rebase, I think we can merge.\

thanks for contribution!

WaterWhisperer:
> @WaterWhisperer once you rebase, I think we can merge.\

Done

foreign/python/Cargo.toml
```diff
@@ -38,6 +38,7 @@ doc = false
 bytes = "1.12.0"
 futures = "0.3.32"
 iggy = { path = "../../core/sdk", version = "0.10.1-edge.2" }
+iggy_common = { path = "../../core/common", version = "0.10.1-edge.2" }
```
hubcio: noticed this right before merging - do we actually need the `iggy_common` dep? the rust sdk (`iggy`) already re-exports these through its prelude, just not all of them: `ConsumerGroupDetails` is there but `ConsumerGroup` and `ConsumerGroupMember` aren't. looks like an oversight - the prelude exports both base and details for `Stream` and `Topic`, but only `Details` for consumer groups.

cleaner fix: add `ConsumerGroup` and `ConsumerGroupMember` to `core/sdk/src/prelude.rs` (both are already pub at the `iggy_common` root - the sdk's own binary consumer-group client already imports `ConsumerGroup` from there), then drop this dep and pull the whole family from `iggy::prelude` like the other wrappers. one less dependency, and it fixes the prelude gap for the next consumer too.

hubcio (changes requested):
sorry for the last-minute request to change :)

foreign/python/Cargo.toml
WaterWhisperer: Great. I noticed C++ SDK has a similar `iggy_common` import for these consumer group types, although it still uses other `iggy_common` types there. Should we track that cleanup separately?

hubcio: yes. @slbotbm

hubcio (approved):

mmodzelewski (approved):

foreign/python/Cargo.toml
slbotbm: sure

slbotbm: sure

PR #3585, opened 2026-06-30
fix(integration): pass remaining tests for server-ng

numinnex:
Fixes remaining integration tests using server-ng binary.


core/metadata/src/impls/metadata.rs
```diff
@@ -1410,6 +1495,85 @@ where
...
+        }
+        if headers.is_empty() {
+            return;
+        }
+
+        for header in &headers {
+            self.send_prepare_ok(header).await;
+        }
+        let mut loopback = Vec::new();
+        consensus.drain_loopback_into(&mut loopback);
```
hubcio: this can trip the `assert!(self.loopback_queue.borrow().len() < PIPELINE_PREPARE_QUEUE_MAX)` in `push_loopback` and panic the shard-0 metadata primary in release builds. the single drain here happens before the `on_ack` loop below, but `on_ack` itself can push new loopback entries: it promotes buffered requests via `drain_request_queue_into_prepares`, and each promoted prepare goes through `on_replicate` -> `send_prepare_ok` -> `send_or_loopback(self)` -> `push_loopback`. those residuals are never drained - the consensus tick arm in the router doesn't call `process_loopback` - so the next tick's repair pushes up to 32 more on top and hits the assert. safer shape: per header, `send_prepare_ok` + drain + `on_ack` - but the drain must also run BEFORE each push: a single `on_ack` can commit a prefix and promote up to a full window of buffered requests, each pushing its own self-ack, so push-then-drain per header can still meet a near-full queue. draining first bounds occupancy to one window. keep the `headers` snapshot, it drops the pipeline/journal borrows before the awaits.

core/metadata/src/impls/metadata.rs
```diff
@@ -573,10 +622,16 @@ where
                 op = header.op,
                 commit = consensus.commit_max(),
                 operation = ?header.operation,
-                "received old prepare, skipping replication"
+                "journal already holds prepare, re-forwarding + re-acking suffix"
             );
-            // Old prepare: downstream already has it or learns via newer
-            // forward; no chain-replicate; WAL unaffected.
+            self.replicate(&message).await;
+            for op in (consensus.commit_max() + 1)..=header.op {
```
hubcio: re-acking the whole suffix per retransmitted prepare is O(W^2): the primary retransmits every un-acked op each cycle (`retransmit_targets` walks the full window), so a backup emits W(W+1)/2 PrepareOks per cycle - 528 at the new depth of 32, 16x the old depth-8 volume. with 2+ backups this aggregates past `DEFAULT_INBOX_CAPACITY` (1024) on the primary's single shard-0 inbox, which can re-trigger the exact overflow this path recovers from. re-acking only the received op is enough - the primary re-enumerates every gap op on each retransmit cycle, so a lost lower ack always gets its own retry - and keeps the ack volume linear. keeping the re-forward is fine.

core/sdk/src/quic/quic_client.rs
```diff
@@ -660,73 +702,138 @@ impl QuicClient {
...
+                        // in-flight / pipeline-full / view-change cancel). All
+                        // three mean "resend the same request on a fresh bidi";
+                        // the session stays intact so no reconnect/relogin needed.
+                        // Login/register resends on `TransientNotCommitted` too
+                        // (nothing committed -> the same Register replays
+                        // idempotently, no fresh session needed); for the silent
+                        // paths login still falls through to the reconnect+relogin
+                        // path in `send_raw_with_response`.
+                        Err(
+                            error @ (IggyError::Disconnected
```
hubcio: resending on silent `Disconnected`/`EmptyResponse` is unsafe for partition-plane ops: they reuse `current_request_id()` without advancing and the partition plane keeps no reply cache, so a request the server buffered under pipeline-full commits later and the resend commits it again - duplicate `SendMessages` at new offsets, and a `DeleteConsumerOffset` that succeeded but lost its reply comes back as a terminal `ConsumerOffsetNotFound`. the 30s deadline with 3s attempts allows ~10 silent resends of one logical request. the fix is not to gate this on `TransientNotCommitted` only, though - the partition plane never sends that frame, silence is its transient signal (the server abandons the bidi at `REPLY_WAIT_TIMEOUT`), so that gate would kill quic transient recovery for partition ops entirely. this needs a server-side answer: don't abandon a bidi whose request was accepted/buffered, or the request-idempotency key the `vsr.rs` TODO already sketches.

core/message_bus/src/transports/quic.rs
```diff
@@ -369,46 +421,86 @@ impl TransportConn for QuicTransportConn {
...
+                            break;
+                        }
+                        res = rx.recv().fuse() => if let Ok(m) = res {
+                            let deliver = match reply_route(&m) {
+                                // Eviction / control frame: always deliver.
+                                ReplyRoute::Unkeyed => true,
+                                // Keyed reply: only this bidi's; else it's a
+                                // stale/duplicate from an abandoned bidi -> drop.
+                                ReplyRoute::Keyed(id) => {
+                                    request_id.is_none() || Some(id) == request_id
```
hubcio, line 480: this keying only disambiguates metadata ops. all partition ops on a connection share one constant request id (`current_request_id()` never advances - only replicated metadata ops bump the counter), so once a reply is stranded (bidi abandoned at the 2s timeout while the op commits later), the next partition op carries the same id and consumes the stale reply. from then on the connection's data-plane stream is shifted by one - every op receives the previous op's reply - and a pure-producer connection never self-heals until a metadata op or a reconnect. the module doc's safety argument ("the SDK replays with the SAME request id, so a late reply ... matches the replay's bidi") holds for metadata but not for the data plane. partition ops need their own correlation id (without touching the metadata dedup counter).

core/partitions/src/iggy_partition.rs
```diff
@@ -277,20 +277,20 @@ where
...
-    ///
-    /// Returns `ConsumerOffsetNotFound` if the consumer or group has no
-    /// existing on-disk / in-memory offset to delete.
+    /// Deliberately infallible: this runs on the replicated-apply path (every
+    /// replica), where the offset may legitimately be absent (e.g. a backup
+    /// that never observed the primary-only `NoAck` store). The client-facing
+    /// "offset must exist" precondition is enforced once at primary admission
+    /// (`ensure_consumer_offset_exists` in `on_request`); re-checking here would
+    /// fail the replicated apply on such a replica and wedge the group.
     pub(crate) fn stage_consumer_offset_delete(
```
hubcio, line 286: making staging infallible fixes the quorum-store + quorum-delete prepare-time race, but it makes the NoAck mix reachable: a NoAck v2 store applies on the primary only and is never replicated, so a backup that never saw the store now stages the delete and the commit-time `guard.remove().ok_or(ConsumerOffsetNotFound)?` fails there - `handle_committed_entries` then panics with "replica is divergent" on every backup. the old staging check made backups silently skip this case. and if the committed delete replays from the journal on restart it hits the same missing offset again, so the backup can crash-loop until repair covers it. probably worth rejecting a replicated delete of a NoAck-stored offset at admission (cheap while only wire clients can send v2 ack levels - the rust sdk doesn't construct them yet), or declaring ack-level mixing on one consumer offset unsupported.

core/consensus/src/plane_helpers.rs
```diff
@@ -334,6 +334,65 @@ where
...
+    *header = ReplyHeader {
+        cluster: request_header.cluster,
+        size: total_size as u32,
+        view: request_header.view,
+        release: request_header.release,
+        command: Command2::Reply,
+        replica: request_header.replica,
+        request_checksum: request_header.request_checksum,
+        client: request_header.client,
+        op: request_header.session,
```
hubcio: `op` gets the client session id here; the sibling builders stamp a position (`build_reply_from_request` uses `op: commit`). inert - the transient reply is never cached and the SDK doesn't read `reply.op` - but wrong-typed on the wire for anyone inspecting frames. `op: commit` would match the convention.

core/partitions/src/iggy_partition.rs
```diff
@@ -1027,9 +1047,62 @@ where
                     .with_operation(message.header().operation)
                     .with_error(error.to_string()),
                 );
+                // Reply with a terminal error instead of dropping the request:
+                // the SDK decodes any non-`TransientNotCommitted` result code as a
+                // final error, so the client gets `ConsumerOffsetNotFound` at once
+                // rather than replaying until its read-timeout (a client hang).
+                // The op is NOT admitted/replicated, so no replica ever applies a
+                // delete for an offset the primary knows is absent.
+                let reply = build_transient_reply(
```
hubcio: `build_transient_reply` is documented as the transient-retry frame builder, but here (and at the `InvalidOffset` store rejection below) it ships a terminal error. it's really a generic `[count=1][index=0][result=code]` rejection builder - a rename (e.g. `build_result_rejection_reply`) plus a doc tweak would stop the name from lying on these paths.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -64,6 +64,14 @@ const NAME: &str = "Iggy";
 #[cfg(feature = "vsr")]
 const RESPONSE_READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
 
+/// Backoff before replaying a request the server answered with an explicit
+/// `TransientNotCommitted` frame (not-caught-up / in-flight / pipeline-full /
+/// view-change cancel). The reply arrives promptly, so a short pause keeps the
+/// replay from spinning while the primary catches up. Bounded by
+/// `RESPONSE_READ_TIMEOUT`.
+#[cfg(feature = "vsr")]
+const NOT_READY_RETRY_INTERVAL: std::time::Duration = std::time::Duration::from_millis(50);
```
hubcio, line 73: two small things: (1) fixed 50ms with no jitter means every client bounced by the same event (view change, failover) replays in lockstep; a little jitter would spread the burst. (2) this const plus the retry arm around `decode_response_split` is duplicated verbatim across tcp/websocket and partially by quic - `vsr.rs` already hosts the shared encode/decode helpers, so the const and the should-retry decision could live there once.

core/integration/src/harness/orchestrator/harness.rs
```diff
@@ -152,11 +154,44 @@ impl TestHarness {
...
+        // test on a startup blip, retry spawn + mesh-readiness a few times,
+        // tearing down and respawning between attempts. `ServerHandle::start`
+        // truncates the captured stdout (`File::create`), so the readiness
+        // log-grep never matches a stale marker from a prior attempt.
+        #[cfg(feature = "vsr")]
+        {
+            const CLUSTER_STARTUP_ATTEMPTS: usize = 3;
+            for attempt in 1..=CLUSTER_STARTUP_ATTEMPTS {
+                for server in &mut self.servers {
+                    server.start()?;
```
hubcio: if `start()?` fails here on a later attempt, servers already started in this pass are left running (the not-ready arm below stops them before retrying). harness `Drop` reaps them at scope exit so it's transient - just inconsistent with the sibling arm.

core/sdk/src/vsr.rs
```diff
@@ -243,7 +243,26 @@ fn read_operation(header_bytes: &[u8; HEADER_SIZE]) -> Result<Operation, IggyErr
...
+    // login decodes to `TransientNotCommitted` and the SDK replays it. The one
+    // exception is a terminal failure, which ships an empty body (no result
+    // section) and is passed through to fail the typed `LoginRegisterResponse`
+    // decode. The consumer-offset ops are the partition-plane ops that carry a
+    // result section: they can be rejected (`ConsumerOffsetNotFound` on delete,
+    // `InvalidOffset` on store), which must reach the client as a terminal error
+    // rather than decoding as `Ok` (their success reply ships `[count = 0]` to
+    // match). Other reads, data-plane ops, and Logout carry no result section and
+    // pass through untouched.
+    let result_framed = operation.is_metadata()
```
hubcio: the "which ops carry a result section" predicate now lives in 5 places that must agree: here, `ApplyReply::write_reply_body`, `committed_reply_body` in the partition plane, and the login + raw-pat reply builders in server-ng. if any encode site drifts from this decode carve-out, a rejection decodes as `Ok` or a payload gets misread as the result count. worth centralizing as something like `Operation::is_result_framed()` in binary_protocol, keeping the register empty-body-is-terminal nuance as the one sdk-side check.

core/shard/src/lib.rs
```diff
@@ -1839,6 +1844,13 @@ where
         let actions = consensus.tick(PlaneKind::Metadata);
 
         dispatch_vsr_actions(consensus, metadata.journal.as_ref(), &actions).await;
+
+        // Repair a lost primary self-ack: `RetransmitPrepares` to self is a
+        // no-op, so the timer-driven retransmit above cannot recover the
+        // primary's own missing vote. Without this the commit prefix can pin
+        // forever (commit_min stuck below commit_max). See
+        // `IggyMetadata::repair_primary_self_acks`.
+        metadata.repair_primary_self_acks().await;
```
hubcio, line 1857: heads up that the invariant comment at line 1467 ("`on_ack` commits and applies but never calls `push_loopback`") is now provably false within-plane: metadata `on_ack` can promote buffered requests, and each promoted prepare self-acks through `send_or_loopback(self)` -> `push_loopback`. still true cross-plane (which is what the drain-order argument needs), but as written it hides the loopback-residual hazard in `repair_primary_self_acks`. worth rewording to "never pushes to another plane's loopback".

core/server-ng/src/segment_recovery.rs
```diff
@@ -0,0 +1,248 @@
...
+            path = %index_path,
+            error = %source,
+            "failed to read last sparse index entry during recovery"
+        );
+        source
+    })?;
+
+    match (first, last) {
+        (Some(first), Some(last)) => {
+            if last.position >= messages_size {
```
hubcio: this only proves the last flush STARTS inside the file - `position` is the flush's first-batch start byte, so an index entry whose message bytes were torn off mid-flush (`last.position < messages_size`, tail truncated) passes, and recovery claims `end_offset = last.offset` for bytes that are incomplete. with `enforce_fsync` on the writer persists messages before the index, so the window closes; with it off there's no ordering barrier. the sparse format can't do an exact indexed-size vs file-size compare, but reading the batch header at `last.position` and checking its extent fits inside the file would tighten this. the torn bytes also stay inside `current_position`, so the next append lands right after them and reads of `last.offset` keep decoding garbage even once new writes succeed.

core/server-ng/src/segment_recovery.rs
```diff
@@ -0,0 +1,248 @@
...
+        segment.end_offset = end_offset;
+        segment.size = IggyByteSize::from(messages_size);
+        segment.current_position = messages_size;
+
+        stats.increment_segments_count(1);
+        stats.increment_size_bytes(messages_size);
+        if messages_size > 0 {
+            // Offsets in a segment are contiguous, so the message count is the
+            // inclusive span between the first (segment start) and last offset.
+            stats.increment_messages_count(end_offset - start_offset + 1);
```
hubcio, line 163: when the log is non-empty but the index holds no whole entry (crash after the message write, before the index write - reachable regardless of fsync since messages persist first), this counts exactly 1 message: the `(None, None)` arm returns `end_offset = start_offset` with zero timestamps, and because `size > 0` the bootstrap filters treat that as "one committed message at `start_offset`" rather than empty, whatever the torn log actually holds. treating an index-less segment as empty would need `size = 0` as well, else the bootstrap non-empty filters still pick it up. new appends also start at `current_position = messages_size`, right after the torn bytes, so a replay from the segment start either fails on them or silently skips the phantom offset.

hubcio (changes requested):

hubcio (approved):

spetz (approved):

PR #3590, opened 2026-07-01
chore(deps): fix dependabot PR for go SDK

chengxilo:

Fix #3571

## Rationale

This is because dependabot doesn't support Go mod well when we use `replace`. The dependence in `bdd/go` wasn't updated. 

I found similar issue regarding this behavior: https://github.com/dependabot/dependabot-core/issues/15134
## What changed?

Run go mod tidy in bdd/go so we can pass the CI.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. DeepSeek v4 pro & Claude Opus 4.8
2. DeepSeek v4 pro implementation, Claude Opus 4.8 reivew, human brain review again
3. ran tests, manually review and edit
4. yeah


.github/workflows/post-dependabot.yml
```diff
@@ -0,0 +1,72 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+name: Post Dependabot
+
+on:
+    pull_request_target:
```
hubcio: this can't merge under ASF policy. pull_request_target must not export GITHUB_TOKEN, and it does - line 45 hands the token to checkout of the PR head, then it's used to push. dependabot events get a read-only token by design; pull_request_target is exactly the escape hatch the policy closes. on top of that it checks out PR-head go.mod/go.sum and runs `go mod tidy` on them with a write token - privileged context on untrusted input, gated only by the actor `if`. pushing into foreign/go (a released module) via automation also needs infra sign-off. suggest dropping the workflow and keeping just the go.sum fix - a committer running `go mod tidy` on the branch is allowed and low frequency.

https://infra.apache.org/github-actions-policy.html

.github/workflows/post-dependabot.yml
```diff
@@ -0,0 +1,72 @@
...
+                  fi
+
+                  git config user.name 'github-actions[bot]'
+                  git config user.email 'github-actions[bot]@users.noreply.github.com'
+
+                  git add foreign/go/go.mod foreign/go/go.sum \
+                          bdd/go/go.mod bdd/go/go.sum \
+                          examples/go/go.mod examples/go/go.sum
+
+                  git commit -m "chore(go): run go mod tidy after dependabot bump"
```
hubcio: pushes made with the default GITHUB_TOKEN don't trigger new workflow runs, so pre-merge won't re-run and the tidy'd go.sum lands unvalidated. forcing a re-run means a PAT/app token, which is even more credential exposure under pull_request_target. so this doesn't really reach the goal even setting the policy aside.

hubcio (changes requested):
tbh I'd close this change, too much effort for too little value, but I leave the decision to you. you can still work around it by checking how PR triage workflow is implemented.

hubcio (approved):

lukaszzborek (approved):

hubcio (approved):

PR #3591, opened 2026-07-01
feat(simulator): deterministic workload fuzzer with consensus oracles

krishvishal:
## Summary
  
Adds a TigerBeetle-style workload fuzzing harness to the simulator: randomized, fully seed-deterministic workloads that run against a predictive shadow, inject faults, and check consensus invariants both continuously and at quiesce. Every failure reproduces from its seed.

 ## What's included

  - **Workload driver**:  samples ops against a name-keyed shadow with outcome-first generation (targets committed rejections, not just successes); one request in flight per client.
  - **Per-tick invariants**: cheap consensus checks on every replica each driver tick, so a regression is caught at the tick it happens.
  - **Crash-only fault injection**: followers crash per-tick by probability from the driver; runs still drain on the surviving quorum.
  - **Quiesce-time cross-replica oracle**:  after drain, asserts no live replica leads the leader and that the shadow equals the leader's committed metadata (entity oracle).
  - **workload-fuzz binary**:  `--seed/--ticks/--clients/--replicas/--crash-prob/--no-quiesce`; seed logged on start and panic for replay.
  - **SimClient request-numbering fix**:  partition ops read the current request id without advancing it (mirrors the SDK), keeping the metadata request sequence gap-free so mixed metadata + partition workloads no longer wedge the client's metadata plane.


core/simulator/src/workload/invariants.rs
```diff
@@ -0,0 +1,139 @@
...
+//! Cheap per-tick invariants.
+//!
+//! These run on every driver tick, not only at quiesce, so a regression is
+//! caught at the tick it happens rather than masked by later progress. The
+//! checks are read-only and draw no PRNG, so enabling them leaves the reply
+//! trace and the determinism baseline (`workload_replay_is_deterministic`)
+//! unchanged.
+
+use std::collections::HashMap;
+
```
hubcio: please remove empty lines between use items. i wanted to add rustfmt linter for that, but it's been `nightly` since ever :(
correct in other files added/modified in this PR too.

core/simulator/src/client.rs
```diff
@@ -81,10 +86,26 @@ impl SimClient {
...
+    /// `ClientTable` dedups them and rejects anything but `committed + 1`, so a
+    /// gap opens a permanent `RequestGap` and wedges the client's metadata
+    /// plane. Partition ops are at-least-once with no dedup and the server
+    /// treats their id as an opaque echo, so they read the current (next
+    /// metadata) id WITHOUT advancing it — never consuming, and never gapping,
+    /// the metadata sequence. Clients are one-in-flight
+    /// (`CLIENT_REQUEST_QUEUE_MAX == 1`), so the id a partition op shares with
+    /// the next metadata op is never live at the same time in the auditor's
+    /// `(client, request)` map.
+    fn request_id_for(&self, operation: Operation) -> u64 {
```
hubcio, line 112: the doc here says the shared request id is safe because clients are one-in-flight, so a partition op's id and the next metadata op's id are never live at the same time. that holds only while replies are never duplicated. the sim network models reply duplication (`replay_probability`), and it's 0.0 in every config shipped here, so this can't fire today - but the moment someone turns on replay to fuzz duplicates, a delayed or duplicated partition reply can land after that id was reused by a later metadata op M. the auditor matches it to M's entry, the namespace guard trips (metadata carries ns 0, partition carries a real ns) so it goes down the NsMismatch path, M's slot is released and its effect skipped, and M never lands in the shadow - then `assert_converged` panics 'shadow diverged' for something that isn't a real consensus bug.

cleanest fix is to give partition ops their own per-client request counter instead of borrowing the next metadata id - the partition plane has no client-table dedup so it can consume its own ids freely, and the collision goes away. otherwise the doc shouldn't assert safety that only holds at replay=0. the same single-static-primary assumption sits in `primary_index` / `assert_converged` too and will bite once primary-crash injection lands.

core/simulator/src/workload_fuzz.rs
```diff
@@ -0,0 +1,154 @@
...
+//! start and on panic so any failure replays with `--seed <value>`.
+//!
+//! ```text
+//! workload-fuzz [--seed N] [--ticks N] [--clients N] [--replicas N]
+//!               [--crash-prob F] [--no-quiesce]
+//! ```
+//!
+//! The default workload is partition-plane (`SendMessages`): it drains and
+//! converges. Metadata and mixed-plane workloads are gated on the metadata
+//! request-gap bug (`metadata-request-gap-bug.md`); broader op coverage lands
```
hubcio: `metadata-request-gap-bug.md` doesn't exist in the repo (also referenced in the WARN string near line 133). the code should stand on its own - drop the filename and describe the request-gap behavior inline.

core/simulator/src/workload_fuzz.rs
```diff
@@ -0,0 +1,154 @@
...
+    let seed: u64 = args
+        .iter()
+        .position(|a| a == "--seed")
+        .and_then(|i| args.get(i + 1))
+        .and_then(|v| v.parse().ok())
+        .unwrap_or_else(rand::random);
+
+    let ticks: u64 = arg_or(&args, "--ticks", 10_000);
+    let clients: u8 = arg_or(&args, "--clients", 1);
+    let replicas: u8 = arg_or(&args, "--replicas", 3);
```
hubcio: no validation on the numeric args. `--replicas 0` builds a 0-replica sim and panics; `--clients 0` gives an empty client range so the run drains instantly and 'passes' without exercising anything; and `--replicas 2` (or 1) can never crash anything because `min_survivors` defaults to `replica_count/2+1`, so `maybe_inject_crash` always early-returns even with `--crash-prob` set. `arg_or` also swallows bad values via `.parse().ok()...unwrap_or(default)`, so a typo'd flag or out-of-range number silently falls back to the default. worth a small guard (reject replicas/clients 0, clamp crash-prob) or switching to clap.


actually... can you use [clap](https://docs.rs/clap/latest/clap/index.html) and do it the rust-idiomatic way? check iggy bench for inspiration 🦀

core/simulator/src/workload/oracle.rs
```diff
@@ -0,0 +1,265 @@
...
+    streams: BTreeSet<String>,
+    topics: BTreeSet<(String, String)>,
+    users: BTreeSet<String>,
+    consumer_groups: BTreeSet<(String, String, String)>,
+}
+
+impl CommittedMetadata {
+    /// Restrict to workload-generated entities (see [`WORKLOAD_PREFIX`]), so the
+    /// entity oracle compares like with like against the shadow.
+    fn workload_owned(&self) -> Self {
```
hubcio: `workload_owned` clones every string across the four filter+cloned+collect passes, but its only caller consumes a fresh temporary from `read_committed_metadata(...)`. taking `self` by value and doing `retain(|..| ..starts_with(WORKLOAD_PREFIX))` in place drops all the clones and a chunk of code, with identical results.

hubcio (changes requested):

core/simulator/src/workload/invariants.rs
krishvishal: done.

core/simulator/src/client.rs
krishvishal: done.

core/simulator/src/workload_fuzz.rs
krishvishal: done.

core/simulator/src/workload_fuzz.rs
krishvishal: done.

core/simulator/src/workload/oracle.rs
krishvishal: done.

hubcio (dismissed):
one more thing i noticed. usually, people dont put `fn main()` in the lib.rs

krishvishal:
@hubcio moved `main.rs` and `workload_fuzz.rs` to `src/bin`

hubcio (approved):

numinnex (approved):

PR #3605, opened 2026-07-03
feat(server): add deferred PollMessages wait timeout

### LINKED ISSUE #3470: feat: add deferred polling timeout to PollMessages
## Summary

Add deferred-response polling to `PollMessages`.

This comes from discussion #2854: https://github.com/apache/iggy/discussions/2854

The goal is to let consumers wait for data without busy polling.

## Problem

Today `poll_messages()` returns immediately when no messages are available. Consumers that need low latency must poll frequently, which creates avoidable CPU churn, network traffic, and client-side retry logic.

The core user ask is: "wake me up when data arrives."

## Proposed behavior

Add a wait timeout to `PollMessages`:

- timeout `0` keeps current behavior and returns immediately
- timeout `> 0` waits until matching data is readable or timeout expires
- timeout expiry returns the normal empty poll response, not an error
- auto-commit still happens only when messages are returned

For the binary protocol, use an additive `u64` timeout value in microseconds.

For the Rust SDK, expose the timeout through an idiomatic API using existing time abstractions where appropriate.

## Scope

First PR scope:

- binary protocol request support
- common/http request support if needed for API parity
- Rust SDK/client API support
- current `core/server` implementation
- focused protocol, server, and integration/e2e tests

Out of scope for the first PR:

- unsolicited push frames
- subscribe/unsubscribe commands
- generic notification API
- backlog/control-plane events
- all foreign SDKs

## Server implementation direction

When no messages are available and timeout is greater than zero:

1. Do not block the shard/message pump.
2. Register a lightweight waiter for the resolved stream/topic/partition.
3. Complete the waiter when data becomes readable or the timeout expires.
4. Return an empty successful poll response on timeout.
5. Remove waiters on timeout, completion, cancellation, disconnect, and partition/topic removal.

Wakeups should happen only after data is readable. The response remains the data, not a separate notification signal.

## `server-ng` portability

`core/server` may be deprecated soon, so the first PR should keep the current-server wiring small and portable.

The reusable pieces should be:

- binary protocol shape
- common request/API types
- Rust SDK API
- behavior tests
- integration/e2e coverage

The current `core/server` waiter implementation should avoid large refactors or abstractions that would make the upcoming `core/server-ng` port harder.

## Acceptance criteria

- `poll_messages` without timeout keeps existing behavior.
- timeout `0` keeps existing immediate-poll behavior.
- timeout `> 0` waits when no data is available.
- producing data to the same readable partition wakes the pending poll.
- timeout returns an empty successful poll response.
- timeout does not auto-commit.
- auto-commit still happens when messages are returned and `auto_commit=true`.
- disconnected or timed-out consumers do not leave stale waiters.
- consumer-group polling still respects existing assignment and offset rules.
- tests are deterministic and not timing-flaky.
- current `core/server` changes are small enough to port to `core/server-ng`.

## Related links

- Discussion: https://github.com/apache/iggy/discussions/2854
- Previous blocking poll issue: https://github.com/apache/iggy/issues/829
- Original notification issue moved to discussion: https://github.com/apache/iggy/issues/2851
- Duplicate blocking polling issue: https://github.com/apache/iggy/issues/3002

arunsingh:
## Summary

Closes #3470.

Adds deferred polling to `PollMessages` for the current `core/server` binary/WebSocket path. Timeout `0` preserves existing immediate behavior. Non-zero timeout waits until messages are readable or the timeout expires.

For `core/server-ng`, non-zero wait timeout is rejected with `FeatureUnavailable` instead of being silently treated as an immediate poll. Full server-ng waiter support is intentionally left visible as the next design/implementation decision.

## Design

- binary protocol uses `u64` timeout in microseconds
- Rust SDK exposes an idiomatic timeout API
- timeout expiry returns a normal empty poll response
- auto-commit only happens when messages are returned
- HTTP transport remains out of first scope and sends timeout `0`
- WebSocket-facing behavior is covered through the binary path
- consumer groups check assigned partitions so one empty partition does not hide data on another assigned partition
- waiter/notifier state is cleaned up and bounded
- server-ng currently rejects non-zero wait timeout explicitly; full port can follow once maintainers confirm scope

## Out of scope

- HTTP deferred polling support
- push notification frames
- subscribe/unsubscribe commands
- generic control-plane notifications
- backlog notification API
- metadata panic/abort behavior
- all foreign SDKs in this PR
- full server-ng waiter implementation, unless maintainers want it in this PR

## Tests

- [x] protocol roundtrip with timeout
- [x] backward compatibility for old poll request
- [x] partial trailing timeout decode error
- [x] immediate poll behavior unchanged
- [x] timeout-empty response
- [x] wake-on-produce
- [x] disconnect cleanup
- [x] auto-commit unchanged
- [x] consumer-group assigned partition path
- [x] WebSocket-facing path
- [x] server-ng rejects non-zero timeout explicitly
- [x] Bench CLI comparison support

## Local checks

Latest review-fix checks run on commit `459a4b732`:

```text
cargo fmt --all
Result: passed.

cargo check -p server -p server-ng -p integration --features vsr
Result: passed.

cargo build --features vsr --bin iggy-server-ng
Result: passed.

cargo test -p iggy_binary_protocol poll_messages
Result: passed, 13 passed.

cargo test -p iggy_common poll_messages
Result: passed, 0 matched tests; crate compiled.

cargo test -p iggy poll_messages
Result: passed, 0 matched tests; crate compiled.

cargo test -p integration poll_messages_wait_timeout
Result: passed, 7 passed.

cargo test -p integration --features vsr poll_messages_wait_timeout_rejected_by_vsr
Result: passed, 2 passed.

cargo test -p server dropped_registration_removes_waiter
Result: passed, 1 passed.

bash -n scripts/performance/run-poll-wait-timeout-comparison.sh
Result: passed.

bash scripts/ci/shellcheck.sh
Result: passed.

typos <touched files>
Result: passed.

git diff --check
Result: passed.
```

Full `cargo test` is not claimed as passing locally.

## AI usage

Used AI assistance for planning, review triage, and draft checks. Implementation decisions and local verification were reviewed by the author.


core/integration/tests/server/scenarios/poll_messages_wait_timeout_scenario.rs
```diff
@@ -0,0 +1,429 @@
...
+            &consumer,
+            &PollingStrategy::offset(0),
+            1,
+            true,
+            timeout,
+        )
+        .await
+        .unwrap();
+
+    assert_eq!(polled.partition_id, 1);
```
hubcio, line 434: this assert fails under `--features vsr`: `cargo test -p integration --features vsr poll_messages_wait_timeout_consumer_group` panics right here with `left: 0, right: 1`. the chain: the poll passes `Some(0)` with a group consumer; client-side vsr group resolution only kicks in when `partition_id` is `None`, so this goes out as an explicit partition-0 read; server-ng serves exactly that partition (no owned-partition scan, no wait) while the message sits on partition 1. CI never runs the integration suite with the vsr feature, so this stays green there and only fails locally. either cfg-gate `run_consumer_group_checks` out of vsr like the sibling checks, or implement the owned-partition scan in server-ng first.

separate but related: the non-vsr run of this scenario passes via the pre-wait re-check (data already present before the poll), so even the green path never exercises a real wait-then-wake for consumer groups - a produce-after-delay variant like the one in `run_wake_after_append_checks` would cover it.

core/server/src/shard/handlers.rs
```diff
@@ -74,6 +74,7 @@ async fn handle_request(
                 .await?;
 
             shard.metrics.increment_messages(messages_count as u64);
+            shard.wake_poll_waiters(&namespace);
```
hubcio, line 77: `wake_poll_waiters` grabs the global `std::sync::Mutex` around `PollWaiterRegistry` on every append (same at line 423), even when nobody waits. on the thread-per-core runtime that's a blocking mutex shared by all shard threads on the hottest path - contention or the cache-line bounce stalls the whole core, not just one task. cheap fix: an `AtomicUsize` live-waiter count checked before locking, skip when zero; the existing register-then-repoll sequence already closes the race, so no wake can be lost.

core/server/src/binary/handlers/messages/poll_messages_handler.rs
```diff
@@ -41,21 +46,76 @@ pub async fn handle_poll_messages(
...
     let user_id = session.get_user_id();
     let client_id = session.client_id;
     let topic = shard.resolve_topic_for_poll(user_id, &stream_id, &topic_id)?;
-    let (metadata, mut batch) = shard
-        .poll_messages(client_id, topic, consumer, partition_id, args)
+    let (mut metadata, mut batch) = shard
+        .poll_messages(
+            client_id,
+            topic,
+            consumer.clone(),
```
hubcio: `consumer.clone()` here runs on every poll, before the zero-timeout gate - `Consumer` carries an `Identifier` backed by `Vec<u8>`, so that's a heap alloc per poll even for plain immediate polls, plus one more clone per namespace in `poll_wait_namespaces` (line 165). `IggyShard::poll_messages` only borrows the consumer internally and both callers are in this file, so taking `&Consumer` removes all of them.

core/server/src/binary/handlers/messages/poll_messages_handler.rs
```diff
@@ -41,21 +46,76 @@ pub async fn handle_poll_messages(
     let partition_id = req.partition_id;
     let count = req.count;
     let auto_commit = req.auto_commit;
+    let wait_timeout = Duration::from_micros(req.wait_timeout_us);
```
hubcio: no server-side cap on the client-supplied timeout - `u64::MAX` micros (~584k years) is accepted and there's no config to bound it, so clients can park connections in the wait state indefinitely. related: once a namespace hits `MAX_WAITERS_PER_NAMESPACE` (1024), `register` returns `None` and the next poller silently degrades to an immediate empty response instead of waiting. a configurable max wait (clamped, not rejected) plus a debug log on the cap path would cover both.

core/server/src/binary/handlers/messages/poll_messages_handler.rs
```diff
@@ -41,21 +46,76 @@ pub async fn handle_poll_messages(
...
+            count,
+            auto_commit,
+        )
+        .await?
+        {
+            metadata = next_metadata;
+            batch = next_batch;
+        }
+
+        if batch.is_empty() && !waiters.is_empty() {
```
hubcio, line 104: one wake buys exactly one re-poll: if a group peer drains the partition between wake and re-poll, this returns empty with most of the timeout budget unused and the client goes back to busy-looping - the thing this feature exists to kill. also, `matches!(..., Ok(true))` means a wake permit landing exactly at the deadline can lose to the timer and skip the final re-poll. looping wait+re-poll until the deadline fixes both.

core/server/src/shard/waiters.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub(crate) async fn wait(&self) -> bool {
+        self.receiver.recv().await.is_ok()
+    }
+}
+
+impl Drop for PollWaiterRegistration {
+    fn drop(&mut self) {
+        self.registry
+            .lock()
+            .expect("poll waiter registry poisoned")
```
hubcio: also at 160, 173, 180, 187. a poisoned lock here cascades - this mutex sits on every shard's produce and poll path, so one panic-while-locked turns into process-wide panics. the critical sections are panic-free today so it's theoretical, but `lock().unwrap_or_else(|e| e.into_inner())` is strictly safer (the map stays valid across a poison) and drops the five `expect`s.

core/server/src/shard/waiters.rs
```diff
@@ -0,0 +1,218 @@
...
+            .keys()
+            .copied()
+            .filter(|namespace| namespace.stream_id() == stream_id)
+            .collect::<Vec<_>>();
+        for namespace in namespaces {
+            self.wake_namespace(&namespace);
+        }
+    }
+
+    fn prune_namespace(&mut self, namespace: &IggyNamespace) {
```
hubcio: `prune_namespace`, the `deadline` field and the `is_closed()` check look like dead defensive code: `PollWaiterRegistration::Drop` removes the waiter on every exit path (timeout, wake, disconnect, losing `select_all`), and the drop body runs before the receiver field is dropped, so an in-map waiter can never be observed closed or expired. the `timeout` argument to `register` only feeds the unused deadline - the handler's `compio::time::timeout` is the real bound. deleting prune + deadline + the timeout param is ~20 lines less and the 1024 cap still counts only live waiters.

core/server/src/shard/system/messages.rs
```diff
@@ -682,14 +682,25 @@ pub struct PollingArgs {
     pub strategy: PollingStrategy,
     pub count: u32,
     pub auto_commit: bool,
+    pub wait_timeout: Duration,
```
hubcio: `wait_timeout` is write-only - the handler sets it via `with_wait_timeout` for the first poll, but nothing reads it: poll execution uses only `strategy`/`count`/`auto_commit`, and the actual waiting lives in the handler off a local. dropping the field and `with_wait_timeout()` (first poll goes back to `PollingArgs::new`) also kills the misleading impression that the shard-level poll honors a timeout.

core/common/src/traits/message_client.rs
```diff
@@ -38,6 +39,37 @@ pub trait MessageClient {
...
+        stream_id: &Identifier,
+        topic_id: &Identifier,
+        partition_id: Option<u32>,
+        consumer: &Consumer,
+        strategy: &PollingStrategy,
+        count: u32,
+        auto_commit: bool,
+        wait_timeout: Duration,
+    ) -> Result<PolledMessages, IggyError> {
+        if wait_timeout.is_zero() {
```
hubcio, line 60: mutual delegation trap: this default delegates zero-timeout to `self.poll_messages`, while `IggyClient` and `ClientWrapper` implement `poll_messages` by calling `poll_messages_with_timeout(Duration::ZERO)`. it only terminates because those impls also override `poll_messages_with_timeout` - remove one of those overrides later and it compiles fine, then stack-overflows at runtime. worth a doc line on the trait: implementors must override at least one of the pair with a real implementation.

core/common/src/traits/message_client.rs
```diff
@@ -38,6 +39,37 @@ pub trait MessageClient {
...
+                    partition_id,
+                    consumer,
+                    strategy,
+                    count,
+                    auto_commit,
+                )
+                .await;
+        }
+
+        Err(IggyError::FeatureUnavailable)
```
hubcio, line 74: non-zero timeout on a transport without an override surfaces as bare `FeatureUnavailable` ("Feature is unavailable") - nothing points at the wait timeout as the culprit. HTTP users hit this first; a doc note on the method (and ideally the transport in the error path) would save them the head-scratch.

core/bench/src/args/common.rs
```diff
@@ -71,6 +72,10 @@ pub struct IggyBenchArgs {
     #[arg(long, short = 'w', default_value_t = IggyDuration::from_str(DEFAULT_WARMUP_TIME).unwrap())]
     pub warmup_time: IggyDuration,
 
+    /// Poll wait timeout in human readable format, e.g. "10ms", "1s"
+    #[arg(long, default_value_t = IggyDuration::from_str(DEFAULT_POLL_WAIT_TIMEOUT).unwrap(), value_parser = IggyDuration::from_str)]
+    pub poll_wait_timeout: IggyDuration,
```
hubcio, line 77: `IggyDuration::from_str` maps `0`, `unlimited`, `disabled` and `none` all to zero - so `--poll-wait-timeout unlimited` reads like 'wait forever' but silently turns deferred polling off, and since the conflict guards check `is_zero()` it sails past those too. doc that `0s` disables; rejecting the word forms for this flag would be even better.

scripts/performance/run-poll-wait-timeout-comparison.sh
```diff
@@ -0,0 +1,386 @@
...
+    GROUP_BATCHES=20
+    BUSY_LOOP_BATCHES=20
+else
+    SPARSE_BATCHES=1000
+    SATURATED_BATCHES=250
+    GROUP_BATCHES=1000
+    BUSY_LOOP_BATCHES=200
+fi
+
+wait_timeouts=("0s" "10ms" "100ms" "1s")
```
hubcio, line 141: this comparison structurally can't see the one always-on cost the PR adds: `wake_poll_waiters` runs on the produce path unconditionally, so it costs the same in the `0s` and deferred arms and cancels out of every comparison here - and all suites run a single producer, so the cross-shard mutex contention never shows up either. measuring the append-path overhead needs a multi-producer/multi-stream run against a master-built server, not `0s` vs deferred on the patched binary.

scripts/performance/run-poll-wait-timeout-comparison.sh
```diff
@@ -0,0 +1,386 @@
...
+    wait_for_process "iggy-server" 10 || send_signal "iggy-server" "KILL"
+}
+
+function run_suite() {
+    local suite="$1"
+    local wait_timeout="$2"
+    local command="$3"
+    local log_file
+    local resource_log
+    local sampler_pid=""
```
hubcio: `sampler_pid` is `local` to `run_suite`, so the `on_interrupt`/EXIT traps can't reach it - ctrl-c mid-suite leaks the sampler as a forever pgrep+sleep loop. hoist it to a global and kill it in the trap.

hubcio (commented):
review priorities, most important first: (1) user-facing API ergonomics, (2) tests, (3) implementation in core/server-ng, (4) implementation in core/server - core/server is getting replaced by server-ng in about a month, so comments on it matter mostly as the blueprint for the server-ng port, not as long-lived code.

findings that don't map to changed lines:

- server-ng ignores the new `wait_timeout_us` entirely: `decode_poll_request` in `core/server-ng/src/dispatch.rs` builds `PollingArgs::new(strategy, count, auto_commit)` and drops the decoded timeout; there is no wait/waiter logic anywhere in that crate. a client calling `poll_messages_with_timeout` with a non-zero timeout against server-ng gets an immediate empty response - no wait, no error, nothing detectable. since server-ng is about to become the server, shipping the wait path only in core/server means the feature effectively vanishes on the switch. minimum: server-ng should reject non-zero timeouts loudly; better: port the wait path as an immediate follow-up.
- `cg_vsr.rs` no longer exists on current master, so this branch needs a rebase and the new vsr test needs a new home; the modify/delete conflict is guaranteed.
- edge on the append path (`core/server/src/shard/system/messages.rs` around 392-433, unchanged lines): the offset advances (data becomes readable) before the threshold `persist_frozen_batches_to_disk().await?`; if that persist errors, the request handler returns before `wake_poll_waiters`, so a waiter sleeps out its full timeout despite readable data. self-heals on the next poll - worth a comment or waking before the persist block.
- pre-existing, just amplified by long polls: auto-commit commits the offset before the response is sent, so a disconnect between commit and delivery skips that batch for the consumer (at-most-once). worth documenting on `auto_commit` now that polls can be parked for seconds.

also, please remove the markdown file `scripts/poll-wait-timeout-comparison.md` - it has way too strong LLM smell for my liking ;)

core/integration/tests/server/scenarios/poll_messages_wait_timeout_scenario.rs
arunsingh: Rebased and gated this CG case out of VSR; I kept full server-ng owned-partition scan/extra CG wake coverage as follow-up unless you want it here.

core/server/src/shard/handlers.rs
arunsingh: Added a live-waiter fast path so append skips the registry mutex when nobody is waiting.

core/server/src/binary/handlers/messages/poll_messages_handler.rs
arunsingh: Changed shard polling to borrow Consumer; immediate polls no longer clone it.

core/server/src/binary/handlers/messages/poll_messages_handler.rs
arunsingh: Added a server-side 30s clamp and warning on waiter-cap fallback; happy to wire the cap to config if preferred.

core/server/src/binary/handlers/messages/poll_messages_handler.rs
arunsingh: Changed wake handling to loop and re-poll until the deadline.

core/server/src/shard/waiters.rs
arunsingh: Replaced the waiter registry expects with poison recovery via into_inner().

core/server/src/shard/waiters.rs
arunsingh: Removed the dead deadline/prune path and the unused timeout argument.

core/server/src/shard/system/messages.rs
arunsingh: Dropped the write-only timeout field from PollingArgs; timeout now stays in the binary handler.

core/common/src/traits/message_client.rs
arunsingh: Added the trait doc note that implementors must override at least one polling method.

core/common/src/traits/message_client.rs
arunsingh: Added the transport-support doc note for non-zero wait timeouts.

core/bench/src/args/common.rs
arunsingh: Documented 0s as the disable value for --poll-wait-timeout.

scripts/performance/run-poll-wait-timeout-comparison.sh
arunsingh: Agreed; this script is for 0s vs deferred behavior, not master-vs-branch append overhead.

scripts/performance/run-poll-wait-timeout-comparison.sh
arunsingh: Hoisted the sampler PID and stop it from the traps.

hubcio:
closing this PR - `iggy-server-ng` has replaced legacy `iggy-server` (along with its name) so it'd be cleaner if you could open new PR.

PR #3607, opened 2026-07-04
feat(python): expose consumer group delete join and leave APIs

### LINKED ISSUE #3522: [python sdk] Add functions related to consumer group management

WaterWhisperer:

Closes #3522

## Rationale

the remaining APIs still needed Python bindings and tests

## What changed?

Python callers could inspect and create consumer groups, but could not directly delete, join, or leave them through the SDK.

Python client now wraps the Rust SDK's `delete_consumer_group`, `join_consumer_group`, and `leave_consumer_group` APIs, generates the stubs, and covers the methods with end-to-end tests.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Tools: Codex
2. Scope: issue analysis, test refinement
3. Verification: Python SDK test suite
4. Yes


slbotbm (approved):
lgtm cc @hubcio

atharvalade (dismissed):
Not blocking, but since `from_connection_string` can build an http client, join and leave will always fail there with `FeatureUnavailable`. Maybe worth a line in the docstring so people don't think it's a bug on their end? LGTM

foreign/python/src/client.rs
```diff
@@ -515,6 +515,111 @@ impl IggyClient {
...
+    ///     group_id: Consumer group identifier as `str | int`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the client joins the consumer group.
+    ///
+    /// Raises:
+    ///     PyValueError: If an identifier is invalid.
+    ///     PyRuntimeError: If the request fails, including `FeatureUnavailable` on HTTP transport.
+    #[gen_stub(override_return_type(type_repr="collections.abc.Awaitable[None]", imports=("collections.abc")))]
+    fn join_consumer_group<'a>(
```
hubcio, line 570: `join_consumer_group` registers the member server-side, but there is no working way to actually consume as that group from python. all three paths dead-end:

- `join_consumer_group` + `poll_messages` - `poll_messages` hardcodes `RustConsumer::default()` (line 674), which is `ConsumerKind::Consumer`, so you poll as an individual
- `join_consumer_group` + `consumer_group(auto_join_consumer_group=False)` + `consume_messages()` - never polls at all. the gate in `core/sdk/src/clients/consumer.rs:668-670` is `!can_poll || (is_consumer_group && !joined_consumer_group)`, and `joined_consumer_group` is set true at exactly one place, consumer.rs:899 inside `initialize_consumer_group`, which `init_consumer_group` skips at :513-516 when auto-join is off. the `SignedIn` arm bails at :592-596 for the same reason. silent infinite sleep, no error, no timeout
- `consumer_group(auto_join_consumer_group=True)` - joins on its own, so the manual call is redundant

meanwhile the join does assign partitions server-side (`test_join_consumer_group_adds_member_once_when_repeated` asserts `members[0].partitions == [0]`), so those partitions park on a member that structurally cannot poll, and real members starve.

the gate itself is a `core/sdk` bug affecting every language binding, so it shouldn't be fixed in a python-binding PR. what this PR owes is a docstring saying group consumption goes through `client.consumer_group(...)` with auto-join - and importantly not pointing anyone at `auto_join_consumer_group=False`, which hangs. the alternative is to hold `join`/`leave` and ship `delete_consumer_group` alone.

side note: `test_consumer_group_can_disable_auto_join` looks like it covers this shape but only asserts the `name()`/`stream()`/`topic()`/`partition_id()` getters and never polls, which is why the hang has gone unnoticed.

foreign/python/src/client.rs
```diff
@@ -515,6 +515,111 @@ impl IggyClient {
...
+    ///     group_id: Consumer group identifier as `str | int`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the client leaves the consumer group.
+    ///
+    /// Raises:
+    ///     PyValueError: If an identifier is invalid.
+    ///     PyRuntimeError: If the request fails, including `FeatureUnavailable` on HTTP transport.
+    #[gen_stub(override_return_type(type_repr="collections.abc.Awaitable[None]", imports=("collections.abc")))]
+    fn leave_consumer_group<'a>(
```
hubcio, line 610: `IggyClient` holds one `Arc<RustIggyClient>` and `consumer_group()` builds its `IggyConsumer` from that same inner client - one connection, one server-side `client_id`. so calling `leave_consumer_group` here revokes the membership of any consumer created from this client.

the consumer doesn't find out gracefully either. its `joined_consumer_group` flag stays true, so it keeps polling, and every poll comes back `ConsumerGroupMemberNotFound`. that error isn't in the retry set (`core/sdk/src/clients/consumer.rs:788-791`), so the flag never resets to false and the `SignedIn` arm short-circuits at :603 - it never rejoins, even with auto-join on. the consumer ends up wedged thinking it's a member while the server has forgotten it.

worth a line in the docstring.

foreign/python/src/client.rs
```diff
@@ -515,6 +515,111 @@ impl IggyClient {
...
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     group_id: Consumer group identifier as `str | int`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the client joins the consumer group.
+    ///
+    /// Raises:
+    ///     PyValueError: If an identifier is invalid.
+    ///     PyRuntimeError: If the request fails, including `FeatureUnavailable` on HTTP transport.
```
hubcio: also at line 600.

nobody ever sees the token `FeatureUnavailable`. `IggyError` renders that variant as `Feature is unavailable` (the `#[error(...)]` attribute in `core/common/src/error/iggy_error.rs`), and the wrapper passes `e.to_string()` straight through. quote the actual message text instead - anyone matching on the string as documented gets nothing back.

which methods fail over http is correct though, delete really does work there.

foreign/python/src/client.rs
```diff
@@ -515,6 +515,111 @@ impl IggyClient {
...
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     group_id: Consumer group identifier as `str | int`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the consumer group is deleted.
+    ///
+    /// Raises:
+    ///     PyValueError: If an identifier is invalid.
```
hubcio: also at lines 564 and 599.

`PyValueError` is only right for str identifiers. `PyIdentifier::Int` is a `u32` (see identifier.rs), so `-1` and `2**32` blow up during the derived `FromPyObject` extraction and raise `TypeError`/`OverflowError` before `Identifier::try_from` ever runs. the stub advertises an unbounded `builtins.int` on top of that, so `-1` type-checks fine.

either narrow the wording or narrow the stub.

foreign/python/tests/test_consumer_group.py
```diff
@@ -436,6 +436,430 @@ async def test_get_consumer_groups_requires_connection_and_auth(self, unique_nam
...
+    @pytest.mark.asyncio
+    async def test_delete_consumer_group_requires_connection_and_auth(
+        self, unique_name
+    ):
+        """Test delete_consumer_group fails before connecting and before login."""
+        host, port = get_server_config()
+        wait_for_server(host, port)
+
+        client = IggyClient(f"{host}:{port}")
+        with pytest.raises(RuntimeError):
```
hubcio: also at 586, 705, 713, 847, 855.

bare `pytest.raises(RuntimeError)` can't tell the two phases apart. before `connect()` the client-side gate returns `Disconnected`; after it returns `Unauthenticated` (`fail_if_not_authenticated` in `core/common/src/traits/binary_auth.rs`). both collapse to `PyRuntimeError`, so a test named `_requires_connection_and_auth` passes on any failure whatsoever - including `connect()` silently no-opping.

same at 815 and 833: those assert genuinely different server errors (`ConsumerGroupMemberNotFound` vs `ConsumerGroupIdNotFound`) and can't distinguish them.

nothing is actually broken today, every one of these fires for its intended reason. it just means a future regression slips through unnoticed. `match=` fixes it, and the file already does that at 1081 and 2022.

foreign/python/tests/test_consumer_group.py
```diff
@@ -436,6 +436,430 @@ async def test_get_consumer_groups_requires_connection_and_auth(self, unique_nam
...
+        )
+
+        remaining = await iggy_client.get_consumer_groups(stream_name, topic_name)
+        assert len(remaining) == 1
+        assert remaining[0].name == group_to_keep
+        assert remaining[0].partitions_count == 1
+        assert remaining[0].members_count == 0
+
+    @pytest.mark.asyncio
+    async def test_delete_consumer_group_accepts_numeric_ids(
```
hubcio, line 521: also at 631 and 756, plus the three `_requires_connection_and_auth` tests at 570, 699 and 841.

the numeric-id preamble is byte-identical across all three tests, and the auth tests differ only by which method they call. one async helper returning `(stream_id, topic_id, group_id)` plus a `@pytest.mark.parametrize` over the method name with `getattr` would cut roughly 90 of the 424 added lines with no loss of coverage. the TODO at the top of this file asks for exactly that.

foreign/python/tests/test_consumer_group.py
```diff
@@ -436,6 +436,430 @@ async def test_get_consumer_groups_requires_connection_and_auth(self, unique_nam
...
+
+        with pytest.raises(RuntimeError):
+            await iggy_client.delete_consumer_group(
+                stream_name,
+                topic_name,
+                unique_name(),
+            )
+
+    @pytest.mark.asyncio
+    async def test_delete_consumer_group_twice_fails_second_time(
```
hubcio: this doesn't catch anything `test_delete_nonexistent_consumer_group_fails` above doesn't already catch. both die at consumer group id resolution, before the delete is ever entered - deleting a group removes it from the slab and the name index, so as far as the resolver is concerned the post-delete state is identical to never-existed.

the leave-twice twin at 797 is a different story and should stay, it resolves fine and fails later with `ConsumerGroupMemberNotFound`.

better use of these 20 lines: a delete-with-live-member test. join, delete, then assert the group is gone and the member with it. that path currently has no coverage at all.

hubcio (changes requested):

foreign/python/src/client.rs
hubcio: actually, could you please propose a fix for that in Rust SDK + tests?

foreign/python/src/client.rs
WaterWhisperer: > the gate itself is a `core/sdk` bug affecting every language binding, so it shouldn't be fixed in a python-binding PR. what this PR owes is a docstring saying group consumption goes through `client.consumer_group(...)` with auto-join - and importantly not pointing anyone at `auto_join_consumer_group=False`, which hangs. the alternative is to hold `join`/`leave` and ship `delete_consumer_group` alone.

Thanks for the detailed explanation. I chose the docstring option and documented the supported group consumption path.

foreign/python/src/client.rs
WaterWhisperer: > actually, could you please propose a fix for that in Rust SDK + tests?

Sure, I'll complete it in a separate commit.

hubcio (approved):

numinnex (approved):

Issue #2148, opened 2025-09-02
Add more BDD test scenarios.


chengxilo:
Currently, we only have a single feature file `basic_messaging.feature` under `bdd/scenarios`, which is clearly insufficient for ensuring full coverage of the system. We should add more BDD test scenarios to cover a broader range of functionality for SDKs. 
It would be great to have a unified testing standard across all SDKs and less duplicated maintenance effort in foreign SDKs tests.

hubcio:
This is very good issue. We should prioritize.

oscarArismendi:
@hubcio  I’ll take a look at this while my other MR is under review. Feel free to assign it to me, just a heads up that I’ll need some time to review the foreign SDKs tests, so I can be confident in what I can extract into BDD scenarios, but I'll keep sharing updates here as I go.

oscarArismendi:
After digging into it, I’m planning to split the migration into three PRs for the common tests I found

1. Stream Lifecycle
Likely the most important tests across SDKs, covering shared functionality between:

| SDK | Existing test file |
|-----|--------------------|
| Node.js | `foreign/node/src/e2e/tcp.stream.e2e.ts` |
| Python | `foreign/python/tests/test_iggy_sdk.py` → `TestStreamOperations`, `TestErrorHandling` |
| Java | `foreign/java/java-sdk/src/test/java/org/apache/iggy/client/blocking/StreamClientBaseTest.java` |
| C# | `foreign/csharp/Iggy_SDK.Tests.Integration/StreamsTests.cs` |
| Go | `bdd/go/tests/tcp_test/stream_feature_*.go` |

2. Consumer Group Lifecycle

| SDK | Existing test file |
|-----|--------------------|
| Node.js | `foreign/node/src/e2e/tcp.consumer-group.e2e.ts` |
| Python | `foreign/python/tests/test_iggy_sdk.py` → `TestConsumerGroup` |
| Java | `foreign/java/java-sdk/src/test/java/org/apache/iggy/client/blocking/ConsumerGroupsClientBaseTest.java` |
| C# | `foreign/csharp/Iggy_SDK.Tests.Integration/ConsumerGroupTests.cs` |
| Go | `bdd/go/tests/tcp_test/consumers_feature_*.go` |

3. Message Polling Strategies

| SDK | Existing test file |
|-----|--------------------|
| Node.js | `foreign/node/src/e2e/tcp.consumer-stream.e2e.ts`, `tcp.send-message.e2e.ts` |
[...]

PR #3608, opened 2026-07-04
feat(go): add stream/topic purge API with BDD tests and fix server purge bug

chengxilo:

Relates to #2148 

## Rationale

Go SDK doesn't have the purge topic/ purge stream API, so this PR implement them. To test it, I added an new BDD test scenario, and implemented both Go and Rust BDD tests.

## What changed?

The Go SDK was missing PurgeStream and PurgeTopic methods. Added them along with a shared BDD scenario (stream_topic_purge.feature) and implementations in both Go and Rust.

While testing, discovered that purge_all_segments only cleared on-disk segments but left stale messages in the journal and in-flight tiers. Polling after purge would return messages that should have been deleted. Fixed by calling `journal.reset()` and `clear_in_flight()` during partition re-initialization. I am not famliar with Rust server code so during the code review this part worth more attention

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. DeepSeek pro v4, Claude Opus 4.6
2. DeepSeek pro v4 for BDD test implementation, Claude Opus 4.6 for server bug fixing(bug doesn't exists anymore now, 9/2/2026).
3. I reviewed them with my best effert, the bdd scripts are executed and actually works, and they found that there is a bug in the server so I believe test test works.
4. I hope so.


hubcio:
@chengxilo i'll check this during the weekend. could you please rebase?

numinnex:
I've checked whether there is potential concurrency issue due to interleaving of flush + purge commands as this is what I've reckon from the changes could potentially be an issue and did not find any problems with that. Let's wait for @hubcio review and we can merge it.

chengxilo:
nothing has been changed, just because I rebased onto master locally so it was forced pushed() I shouldn't do that tho.

hubcio:
@chengxilo could you please rebase again? i will review this today

chengxilo:
> @chengxilo could you please rebase again? i will review this today

done. Currently, we get the pre-purge messages back for roughly > a second. I'm not sure if we should justify it in BDD test or fix it. Previously for the non-VSR it was fixed. But not for the VSR enabled one. (since according to the comments, it was intended)

slbotbm:
@chengxilo Seems the rust bdd test is failing?

/author

bdd/scenarios/stream_topic_purge.feature
```diff
@@ -0,0 +1,76 @@
...
+    Given I have a running Iggy server
+    And I am authenticated as the root user
+
+  Scenario: Purge stream removes all messages
+    Given I have a stream with name "purge-test-stream"
+    And I have a topic with name "purge-topic" in stream "purge-test-stream" with 3 partitions
+    And I have sent 10 messages to stream "purge-test-stream", topic "purge-topic", partition 0
+    When I purge the stream "purge-test-stream"
+    And I poll messages from stream "purge-test-stream", topic "purge-topic", partition 0 starting from offset 0
+    Then I should receive 0 messages
```
slbotbm, line 34: Could we wait for purge convergence before polling here? The purge command commits a new purge generation, but the partition reconciler clears the messages on a later tick. This immediate poll can therefore still see pre-purge data and make this scenario flaky (faced this while writing tests for cpp). Consider adding bounded retry/wait logic before asserting the message count. The same applies to the other immediate post-purge polls in this feature.

scripts/run-bdd-tests.sh
slbotbm, line 1: The script accepts `stream_topic_purge` as a feature, but `run_suite` does not define which SDKs support it. Consequently, `./scripts/run-bdd-tests.sh all stream_topic_purge` also invokes every other SDK. Those SDKs fall back to their default commands, which run unrelated full suites and can report success without exercising this feature. Let's restrict this feature to the Rust and Go BDD services, as the script already does for other SDK-specific features.

slbotbm (changes requested):

chengxilo:
> @chengxilo Seems the rust bdd test is failing?
> 
> /author

Yes, I am wondering if this is a behavior that need to be fixed. Though in the comment in server side, it seems to be expected. (I mean like... I personally don't like this behavior.

scripts/run-bdd-tests.sh
chengxilo: addressed.

hubcio:
I will continue this PR.

PR #3610, opened 2026-07-05
feat(cpp): add functions related to client lifecycle, and message headers

slbotbm:

Closes #
Relates to #2100

## Rationale
Contributes to completion of cpp sdk.

## What changed?
added user header handling, and the following functions:
- disconnect
- shutdown
- delete_segments
- get_cluster_metadata
- update_permissions
- change_password

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
2. Scope of usage? code generation, code investigation
3. How did you verify the generated code works correctly? read through it, tested it.


foreign/cpp/src/type_conversion.rs
```diff
@@ -314,12 +433,36 @@ impl From<RustConsumerGroupDetails> for ffi::ConsumerGroupDetails {
...
+    type Error = String;
+
+    fn try_from(message: RustIggyMessage) -> Result<Self, Self::Error> {
         let id_bytes = message.header.id.to_le_bytes();
         let id_lo = u64::from_le_bytes(id_bytes[0..8].try_into().unwrap());
         let id_hi = u64::from_le_bytes(id_bytes[8..16].try_into().unwrap());
-        ffi::IggyMessagePolled {
+        let user_headers = match message.user_headers {
+            Some(raw_headers) => {
+                let wire_headers = WireUserHeaders::from_bytes(raw_headers).map_err(|error| {
```
hubcio: poll decode propagates a `WireUserHeaders::from_bytes` error here, and `TryFrom<RustPolledMessages>` collects with `collect::<Result<Vec,_>>()?` (line 821), so one message with a malformed user-header TLV fails the entire poll batch - every good message in that batch is dropped with it. the server does not validate user-header TLV on the produce path (no `validate_user_headers` / `WireUserHeaders` call anywhere in core/server), so a non-first-party or buggy producer can store a header that then breaks C++ polls of that range. with offset-based polling the consumer re-fetches the same range and keeps failing (stuck); with auto_commit the batch gets skipped and its good messages are lost. the reference Rust SDK handles this gracefully - `user_headers_map()` warns and returns `Ok(None)`, keeping the message readable. suggest mirroring that: match `from_bytes`, on error warn and emit empty headers, keep the message.

foreign/cpp/src/type_conversion.rs
```diff
@@ -188,6 +194,119 @@ impl From<RustStats> for ffi::Stats {
...
+            manage_topics: permissions.manage_topics,
+            read_topics: permissions.read_topics,
+            poll_messages: permissions.poll_messages,
+            send_messages: permissions.send_messages,
+            topics,
+        })
+    }
+}
+
+impl TryFrom<ffi::Permissions> for RustPermissions {
```
hubcio, line 289: this whole permissions conversion chain (`TryFrom<ffi::Permissions>`, `TryFrom<ffi::StreamPermissions>`, and the `From` impls above) plus `update_permissions` ship with no test. it's an authz path and the `has_permissions` false -> `None` vs true + all-false -> `Some` distinction is load-bearing, so it's worth a direct unit test on `TryFrom<ffi::Permissions>` - duplicate-id rejection, empty -> `None`, and the None/Some cases. doesn't need `create_user` to exist yet since the conversion is pure.

foreign/cpp/src/lib.rs
```diff
@@ -88,11 +88,40 @@ mod ffi {
...
+    enum HeaderKind {
+        Raw = 1,
+        String = 2,
+        Bool = 3,
+        Int8 = 4,
+        Int16 = 5,
+        Int32 = 6,
+        Int64 = 7,
+        Int128 = 8,
+        UInt8 = 9,
```
hubcio: the header kind variants are named `UInt8..UInt128` here, but every other SDK and the Rust core use `Uint8..Uint128` (lowercase i) - see `HeaderKind` in iggy_common, plus the C# and Java SDKs. the numeric codes match so there's no wire impact, but this is the name C++ users type, and renaming a public enum after release is breaking. cheap to align to `Uint8` now.

foreign/cpp/src/type_conversion.rs
```diff
@@ -330,45 +473,352 @@ impl From<RustIggyMessage> for ffi::IggyMessagePolled {
...
+            },
+            value: ffi::HeaderField {
+                kind: entry.value.kind().as_code(),
+                value: entry.value.as_bytes().to_vec(),
+            },
         }
     }
 }
 
+impl TryFrom<ffi::HeaderEntry> for RustHeaderEntry {
```
hubcio, line 483: the key block and value block here are two nearly identical 15-arm matches that differ only in producing `HeaderKey` vs `HeaderValue` - and both are `HeaderField<T>` with all the `From` / `TryFrom` impls generic over the marker. this collapses to one generic helper `fn decode_field<T>(kind: u8, value: Vec<u8>) -> Result<HeaderField<T>, String>` called for key and value, cutting around 135 lines and folding the duplicated error strings into one spot.

foreign/cpp/src/type_conversion.rs
```diff
@@ -330,45 +473,352 @@ impl From<RustIggyMessage> for ffi::IggyMessagePolled {
...
-            user_headers: message
-                .user_headers
-                .map(|headers| headers.to_vec())
-                .unwrap_or_default(),
+            user_headers,
+        })
+    }
+}
+
+impl From<RustHeaderEntry> for ffi::HeaderEntry {
```
hubcio: this `From<RustHeaderEntry> for ffi::HeaderEntry` has no callers - the poll path builds `ffi::HeaderEntry` inline from the `WireUserHeaders` iterator, and `RustIggyMessage` stores headers as `Option<Bytes>`, never `Vec<RustHeaderEntry>`. dead, can be deleted.

foreign/cpp/src/type_conversion.rs
```diff
@@ -330,45 +473,352 @@ impl From<RustIggyMessage> for ffi::IggyMessagePolled {
...
+            }
         }
         let id = ((message.id_hi as u128) << 64) | (message.id_lo as u128);
         let payload = Bytes::from(message.payload);
-        RustIggyMessage::builder()
-            .id(id)
-            .payload(payload)
-            .build()
-            .map_err(|error| format!("Could not convert message: {error}"))
+        if user_headers.is_empty() {
```
hubcio: the empty/non-empty branch here duplicates the whole builder chain just to conditionally add headers. the builder is bon and `user_headers` is an `Option`, so it generates `maybe_user_headers` - this collapses to one chain: `.maybe_user_headers((!user_headers.is_empty()).then_some(user_headers))`. same result, empty map stays `None`.

foreign/cpp/src/type_conversion.rs
```diff
@@ -330,45 +473,352 @@ impl From<RustIggyMessage> for ffi::IggyMessagePolled {
...
 impl TryFrom<ffi::IggyMessageToSend> for RustIggyMessage {
     type Error = String;
 
     fn try_from(message: ffi::IggyMessageToSend) -> Result<Self, Self::Error> {
-        if !message.user_headers.is_empty() {
-            return Err(
-                "Could not convert message: user_headers are not yet supported in the C++ SDK"
-                    .to_string(),
-            );
+        let mut user_headers = BTreeMap::new();
```
hubcio, line 608: worth a note that user headers are unordered and keys unique: they go into a `BTreeMap`, so the C++ `Vec<HeaderEntry>` insertion order is not preserved on send or poll (sorted by kind+value), and duplicate keys are rejected. the `Vec` shape reads like it keeps order.

foreign/cpp/src/type_conversion.rs
```diff
@@ -314,12 +433,36 @@ impl From<RustConsumerGroupDetails> for ffi::ConsumerGroupDetails {
...
-    fn from(message: RustIggyMessage) -> Self {
+impl TryFrom<RustIggyMessage> for ffi::IggyMessagePolled {
+    type Error = String;
+
+    fn try_from(message: RustIggyMessage) -> Result<Self, Self::Error> {
         let id_bytes = message.header.id.to_le_bytes();
         let id_lo = u64::from_le_bytes(id_bytes[0..8].try_into().unwrap());
         let id_hi = u64::from_le_bytes(id_bytes[8..16].try_into().unwrap());
-        ffi::IggyMessagePolled {
+        let user_headers = match message.user_headers {
```
hubcio, line 441: poll decode here is structural-only (via `WireUserHeaders`), so it accepts unknown kind codes 16-255 and wrong-width fixed types, whereas the Rust SDK poll is strict (`user_headers_from_wire` checks kind + width). same stored bytes read differently across SDKs. looks intentional for forward-compat - worth a comment saying so.

hubcio (changes requested):

foreign/cpp/src/type_conversion.rs
slbotbm: This will be difficult to do -- the current tests only test the final cpp library, and there is no wiring in bazel or in the CI to actually run rust tests. We'll need to explicitly call cargo to run the tests from bazel to do this

foreign/cpp/src/lib.rs
slbotbm: Done

foreign/cpp/src/type_conversion.rs
slbotbm: added comment

foreign/cpp/src/type_conversion.rs
slbotbm: I changed the code to emit empty headers for malformed user-headers TLV while preserving the message and the batch, though I cannot issue a warning from the rust side to the cpp side without introducing some field like `user_header_warning` that will get populated on errors. Though this is a pretty awkward design.

foreign/cpp/src/type_conversion.rs
slbotbm: done

foreign/cpp/src/type_conversion.rs
slbotbm: done

foreign/cpp/src/type_conversion.rs
slbotbm: added todo to mention this in high level sdk doc

foreign/cpp/src/type_conversion.rs
slbotbm: done

foreign/cpp/src/type_conversion.rs
hubcio: alright. lets leave it as is.

hubcio (approved):

numinnex (approved):

Issue #3612, opened 2026-07-06
fix(python): COPY destination broken in Dockerfile.test — maturin develop fails


jiengup:
## Description

`docker compose -f docker-compose.test.yml up --build` fails when building the `python-tests` service image. The `maturin develop` step cannot find the Rust source files.

## Steps to Reproduce

```bash
cd foreign/python
docker compose -f docker-compose.test.yml up --build
```

## Error

The `python-tests` image build fails at the `uv run maturin develop` step:

```
 => ERROR [python-tests base 10/14] RUN uv run maturin develop
------
 > [python-tests base 10/14] RUN uv run maturin develop:
------
💥 maturin failed
  Caused by: Failed to build a native library through cargo
  Caused by: Cargo build finished with "exit status: 101": `"cargo" "rustc" ...`
error: failed to parse manifest at `/workspace/foreign/python/Cargo.toml`

Caused by:
  can't find `src/lib.rs` or `src/main.rs` at `./src/`
```

Cargo expects `src/lib.rs` at `/workspace/foreign/python/src/lib.rs`, but it does not exist because the files were copied to `/workspace/src/` instead.

## Root Cause

In `foreign/python/Dockerfile.test`, the `WORKDIR` is set to `/workspace` at line 47, and the `COPY` on line 62:

```dockerfile
COPY foreign/python/src/ ./src/
```

copies the Rust source files to **`/workspace/src/`** instead of **`/workspace/foreign/python/src/`**.

Later, at line 65, `WORKDIR` switches to `/workspace/foreign/python`, and `maturin develop` (line 67) runs from that directory. It looks for `src/` relative to the current working directory (`/workspace/foreign/python/src/`), which does not exist. Build fails.

## Proposed Fix

Change the `COPY` destination to place the source files under the correct `foreign/python/` directory.

slbotbm:
@jiengup you can add this fix to the pr you will be opening for implementing python user headers

jiengup:
> @jiengup you can add this fix to the pr you will be opening for implementing python user headers

Sure, I'll include it.

PR #3613, opened 2026-07-06
feat(python): add user header and origin timestamp support

### LINKED ISSUE #3601: Python SDK: support for per-message headers (key-value pairs)
Iggy messages support optional typed user headers (key-value metadata next to the payload). The Rust SDK has them, and so do Node, Go, Java and C#. The Python SDK doesn't: you can't attach headers when sending and can't read them when receiving, even though the wrapped Rust types carry the data all the way through.

On the send side, `SendMessage` (`foreign/python/src/send_message.rs`) only takes a payload (`str` or `bytes`) and never calls the underlying builder's `user_headers` setter. There's no way to set a custom message `id` either, which is a separate smaller gap.

On the receive side, `ReceiveMessage` (`foreign/python/src/receive_message.rs`) exposes payload, offset, timestamp, id, checksum, length and partition_id, but nothing for user headers (or `origin_timestamp`). Funny part: the wire-encoded header bytes are already sitting in the wrapped `IggyMessage` on every poll, Python just can't reach them. Applies to all three consumption paths (`poll_messages()`, the consumer async iterator, `consume_messages()` callback). Zero mentions of headers in `apache_iggy.pyi` or the tests, so this never worked.

Everything needed already exists in the Rust SDK: `IggyMessage::builder().user_headers(...)` for sending, `user_headers_map()` for receiving, and typed `HeaderValue` with `HeaderKind` (Raw, String, Bool, Int8..Int128, Uint8..Uint128, Float32/64). So this is binding surface only, no protocol or server work.

## Proposed API

Plain Python types at the boundary, no wrapper classes for the common case:

```python
SendMessage("hello", user_headers={
    "content-type": "application/json",  # str   -> String
    "trace-blob": b"\x00\x01",           # bytes -> Raw
    "is-retry": False,                   # bool  -> Bool
    "attempt": 3,                        # int   -> Int64
    "score": 0.99,                       # float -> Float64
})

headers = message.user_headers()
# dict[str, str | bytes | bool | int | float] | None
```

Optional keyword arg via `#[pyo3(signature = (data, user_headers=None))]`, validation errors raise `ValueError` (same mapping as `Identifier` today). Receive side is a method, not a property, matching the existing accessors. On receive, all int kinds map to `int` losslessly; the mapping is intentionally lossy on kind (a `Uint32` from a Rust client comes back as `int`, re-sending makes it `Int64`). Explicit-kind wrapper can come later without breaking this.

Limits to enforce (mirror Rust SDK, mostly free via `HeaderKey`/`HeaderValue` constructors): keys and values 1..=255 bytes, total headers <= 100 KB (`MAX_USER_HEADERS_SIZE`, fails as `TooBigUserHeaders`).

Implementation gotchas:

- `user_headers_map()` fails two ways: structural TLV garbage returns `Ok(None)` + warn (indistinguishable from "no headers", fine, document it), but unknown kind codes and wrong fixed-size lengths return `Err` - the wire layer deliberately forwards unknown kinds, so a header from a newer SDK lands here. Map it to `ValueError`, don't swallow.
- With a derived `FromPyObject` union, an out-of-`i64` int silently falls through to the `float` variant (pyo3 tries variants in order). Extract explicitly: bool before int (Python `bool` subclasses `int`), int before float, so overflow raises.
- `SendMessage` has a manual `Clone` (because `IggyMessage` isn't `Clone`) - already copies `user_headers`, just keep it in mind.
- Regenerate `apache_iggy.pyi` via `pyo3_stub_gen`; add round-trip tests in `foreign/python/tests/` for all five types plus the no-headers case.

Open questions: default kind for Python `int` (`Int64` proposed); return `None` or `{}` for no headers (`None` proposed); non-string keys are legal on the wire (`Int32`/`Raw` keys can't fit in a `dict[str, ...]` - proposed: raise `ValueError`, nothing in this repo produces them); whether `SendMessage(id=...)` rides along in the same change.

Refs: `core/common/src/types/message/iggy_message.rs`, `core/common/src/types/message/user_headers.rs`, wire TLV in `core/binary_protocol/src/primitives/user_headers.rs`. Parity examples: `foreign/node/src/wire/message/header.type.ts`, `org.apache.iggy.message.HeaderKind`, `foreign/csharp/Iggy_SDK/Headers/`.

jiengup:
Add user headers to SendMessage and ReceiveMessage, expose origin timestamp, and fix Docker test infrastructure.

Closes #3601  #3612 

## Rationale

The Python SDK could send and receive message payloads, but it did not expose the typed user headers already carried by the underlying Rust `IggyMessage`. This left Python behind the Rust, Node, Go, Java, and C# SDKs for message metadata.

## What changed?

Python messages can attach user headers and read them back from received messages. `SendMessage` accepts `user_headers` and an optional custom `id`, while `ReceiveMessage` exposes `user_headers()` and `origin_timestamp()`.

According to the discussion below, unlike the original proposal in #3601 (which exposed only a plain `dict[str, str | bytes | bool | int | float]`), the binding now supports the **full Rust header type surface** through two exposed classes, plus a plain-scalar convenience layer:

- **`HeaderKey` / `HeaderValue`** — typed complex enums covering every `HeaderKind`: `Raw`, `String`, `Bool`, `Int8/16/32/64/128`, `UnsignedInt8/16/32/64/128`, `Float32`, `Float64`.
- **`UserHeaders`** — the mapping returned by `ReceiveMessage.user_headers()`. It is a real `dict` subclass (`dict[HeaderKey, HeaderValue]`), so all mapping operations work, and it adds a chainable `to_scalar_dict()` for the convenient scalar form: `message.user_headers().to_scalar_dict()`.

### Sending (`plain → Rust`)
Each key/value pair is converted **independently**, so typed and plain forms can be freely **mixed** in one dict (e.g. `{HeaderKey.String("a"): 7, "b": HeaderValue.Bool(True), "c": "plain"}`):

- Plain scalars map by best-effort, **lossless** conversion for both keys and values (keys are no longer restricted to `str`):
  - `str → String`, `bytes → Raw`, `bool → Bool`
  - `int →` the **smallest** header kind that holds it exactly: non-negative → `Uint8/16/32/64/128`, negative → `Int8/16/32/64/128`; values beyond the 128-bit range raise `ValueError`.
  - `float → Float32` when the value is exactly representable as f32, otherwise `Float64` (always lossless).
- Explicit typed values are **range-checked**: integer variants are enforced by the binding at construction (`OverflowError` for out-of-range, e.g. `HeaderValue.Int8(300)`), and an explicit `Float32` whose value overflows f32 (→ `inf`) is rejected with `ValueError`.

### Receiving (`Rust → plain`)
Every `HeaderKind` maps **losslessly** onto a Python scalar (128-bit ints fit Python's arbitrary-precision `int`; `f32 → f64` is exact), so:
- `user_headers()` returns typed `UserHeaders` (`dict[HeaderKey, HeaderValue]`), or `None` when the message carries no user headers.
- `UserHeaders.to_scalar_dict()` returns `dict[str | bytes | bool | int | float, str | bytes | bool | int | float]` — a direct, lossless conversion (no logging); it only raises `ValueError` on a genuine decode error of a stored field.

Header decode errors from known-but-invalid or unknown semantic kinds surface as `ValueError`. `SendMessage(id=...)` (mapped to `u128`) is part of the same binding update.

## Minor Fix
The Python test compose setup starts the server with fresh default root credentials, binds HTTP/TCP/QUIC addresses explicitly, and uses `iggy ping` for the healthcheck instead of the HTTP stats endpoint.

> **TODO:** make pre-commit and other scripts compatible with other shells (e.g. zsh); commands like `mapfile` don't work in some shells.

## Discussion Notes

Choices from the issue discussion, updated to the final implementation:

- Python supports **both** a plain `dict` API for the common case **and** explicit `HeaderKey`/`HeaderValue` wrapper classes; the two can be mixed per entry.
- Integers are stored in the narrowest lossless kind (unsigned for non-negative, signed for negative); only values beyond ±128-bit raise `ValueError`.
- Floats select `Float32`/`Float64` by exact representability instead of always `Float64`.
- Non-string keys are now **supported** (any plain scalar or `HeaderKey`), instead of being rejected.
- `ReceiveMessage.user_headers()` returns `None` for no headers.
- `origin_timestamp()` is exposed on received messages.

## API Usage
Sending — plain scalars (common case)
```python
from apache_iggy import SendMessage as Message

# Keys and values are plain Python scalars; each is converted losslessly.
Message(
    "order-created",
    user_headers={
        "content-type": "application/json",  # -> String
        "trace-blob": b"\x00\x01",           # -> Raw
        "retryable": False,                  # -> Bool
        "attempt": 3,                        # -> UnsignedInt8 (smallest fit)
        "created-at-ms": 1_700_000_000_000,  # -> UnsignedInt64
        "score": 1.25,                       # -> Float32 (exactly representable)
    },
    id=42,  # optional custom message id (u128)
)
```

### Sending — explicit typed kinds
```python
from apache_iggy import HeaderKey, HeaderValue
from apache_iggy import SendMessage as Message

Message(
    "order-created",
    user_headers={
        HeaderKey.String("schema-version"): HeaderValue.UnsignedInt16(1),
        HeaderKey.UnsignedInt32(7): HeaderValue.String("order-id"),  # non-string key
    },
)
```

### Sending — mixed typed + plain in one dict
```python
Message(
    "order-created",
    user_headers={
        HeaderKey.String("typed"): HeaderValue.Int32(-5),
        "plain-key": HeaderValue.Bool(True),
        HeaderKey.String("plain-val"): "hello",
        "both-plain": 9,
    },
)
```

### Receiving
```python
message = polled_messages[0]

message.origin_timestamp()          # int

headers = message.user_headers()    # UserHeaders | None
if headers is not None:
    # Typed access (full wire-type fidelity):
    value = headers.get(HeaderKey.String("schema-version"))
    if isinstance(value, HeaderValue.UnsignedInt16):
        print(value.value)          # 1

    # Convenient plain form (lossless):
    plain = headers.to_scalar_dict()      # dict[str|bytes|bool|int|float, str|bytes|bool|int|float]
    print(plain["schema-version"])  # 1
```

### Validation / errors
```python
Message("p", user_headers={"k": 2**128})            # ValueError: 128-bit range
Message("p", user_headers={"k": HeaderValue.Int8(300)})   # OverflowError
Message("p", user_headers={"k": HeaderValue.Float32(1e40)})  # ValueError: 32-bit float
Message("p", user_headers={object(): "v"})          # ValueError: keys must be str/bytes/bool/int/float/HeaderKey
```

## Local Execution
- Passed: `cargo fmt --check --manifest-path foreign/python/Cargo.toml`
- Passed: `cargo check --manifest-path foreign/python/Cargo.toml`
- Passed: `cargo test --manifest-path foreign/python/Cargo.toml`
- Passed: `uv run --extra dev ruff check tests/test_message_operations.py tests/test_consumer_group.py`
- Passed: `.venv/bin/python -m pytest 
tests/test_message_operations.py::TestMessageOperations::test_invalid_user_headers_are_rejected -q`
- Successfully run: `example/message-headers/typed-headers/consumer.py`, `example/message-headers/typed-headers/producer.py`, `example/message-headers/plain-headers/consumer.py`, `example/message-headers/plain-headers/producer.py`

## AI Usage

Codex was used to inspect the existing Python, Rust, Node, and Go SDK behavior, implement the Python binding changes, add tests. All the modification was reviewed carefully by the human.


hubcio:
one additional request from my side: please update python examples to include a code that would show how to use iggy message user headers. there is one for rust already.

jiengup:
> one additional request from my side: please update python examples to include a code that would show how to use iggy message user headers. there is one for rust already.

That will be included when the PR is ready.

slbotbm:
@hubcio @jiengup 

I was thinking that maybe we should change the signature of the headers from `dict[str, str | bytes | bool | int | float] | None` to `dict[str, str | bytes | bool | int | float] | dict[HeaderKey, HeaderValue] | None`, where `HeaderKey` and `HeaderValue` are python classes mirroring rust and expose subclasses like `HeaderKey.String(value)` and `HeaderValue.UnsignedInt128(value)`, etc..

This would allow us to have two advantages: users that do not have much experience/do not require explicit control can use `dict[str, str | bytes | bool | int | float]`, while users that require precise control on the types of headers can use `dict[HeaderKey, HeaderValue]`. Also, by providing `dict[HeaderKey, HeaderValue]` as an option, we allow AI Agents to make less mistakes as compared to `dict[str, str | bytes | bool | int | float]`, which is a little vague from the agents' point of view. 

What do you guys think?

jiengup:
> change the signature of the headers from `dict[str, str | bytes | bool | int | float] | None` to `dict[str, str | bytes | bool | int | float] | dict[HeaderKey, HeaderValue] | None`

I think it's a good idea and well aligned with the first point worth discussing in my PR. I'll push a new patch soon.

hubcio:
@slbotbm for me this sounds fine, however one thing comes to my mind: will it have any impact on performance?

jiengup:
> @slbotbm for me this sounds fine, however one thing comes to my mind: will it have any impact on performance?

Actually, I think additional typed `HeaderKey` support will not introduce more overhead compared to the current implementation.

The only overhead we introduce is the iteration of `user_header` and the transformation of Rust and Python types, which need to be considered if we have a large number of user header pairs in some user cases.

hubcio:
alright, then don't worry about it now.

jiengup:
@hubcio @slbotbm 

Hi, I just reflected the change and merged all the commits into a final one. The PR message is also updated.

Instead of a plain-only `dict[str, ...]` with int → Int64 / float → Float64 and string-only keys, the binding now exposes the full typed `HeaderKey`/`HeaderValue` surface with smallest-fit kind selection, supports non-string keys, and allows typed and plain pairs to be freely mixed in one dict.

Please have a look at it when you are free.

foreign/python/src/user_headers.rs
slbotbm, line 1: Some comments: 
- I'd like to stick to as much idiomatic rust as possible, so please convert the `to_rust` and `from_rust` functions to `TryFrom` impls. You could declare something like `struct RustHeaderKeyRef<'py, 'value> {py: Python<'py>,  value: &'value RustHeaderKey}` so that you can covert `PyBytes` to rust representation. You'll need to define 4 of these for the conversion. 
- Also, please inline as much code is as possible since I'd like to prevent the usage of ad-hoc functions created just for a small convenience. I generally create ad-hoc functions if there is scope to utilize it in the whole sdk. Otherwise it will keep on accumulating.

examples/python/message-headers/consumer.py
```diff
@@ -0,0 +1,170 @@
...
+
+STREAM_NAME = "message-headers-stream"
+TOPIC_NAME = "orders"
+PARTITION_ID = 0
+BATCHES_LIMIT = 5
+MESSAGES_PER_BATCH = 10
+
+ORDER_CREATED_TYPE = "OrderCreated"
+ORDER_CONFIRMED_TYPE = "OrderConfirmed"
+ORDER_REJECTED_TYPE = "OrderRejected"
```
slbotbm: You can utilize `StrEnum` here instead.

examples/python/message-headers/consumer.py
```diff
@@ -0,0 +1,170 @@
...
+def handle_order_created(order_created: Mapping[str, Any]):
+    logger.info(f"Order Created: {order_created}")
+
+
+def handle_order_confirmed(order_confirmed: Mapping[str, Any]):
+    logger.info(f"Order Confirmed: {order_confirmed}")
+
+
+def handle_order_rejected(order_rejected: Mapping[str, Any]):
+    logger.info(f"Order Rejected: {order_rejected}")
```
slbotbm: Inline these. these do not add anything new to code.

examples/python/message-headers/consumer.py
slbotbm, line 1: Instead of duplicating functions and enums in consumer.py and producer.py, you can define a common.py file and move all common code there.

Maybe also consider splitting the examples into 4 files: producer/consumer for typed headers, and producer/consumer for normal headers. That, or demonstrate only typed headers, with a comment stating that normal pythonic headers are also possible and show what kind. something like the following:

```
 message-headers/
  ├── message-type/
  │   ├── common.py     
  │   ├── producer.py     
  │   └── consumer.py     
  └── typed-headers/
      ├── producer.py     
      └── consumer.py     
```
Also, add return types for all functions. Some are missing them

examples/python/message-headers/producer.py
```diff
@@ -0,0 +1,237 @@
...
+PlainHeaderValue = str | bytes | bool | int | float
+PlainHeaders = dict[str, PlainHeaderValue]
+TypedHeaders = dict[HeaderKey, HeaderValue]
+
+
+class ArgNamespace(NamedTuple):
+    connection_string: str
+
+
+class SerializedMessage(NamedTuple):
```
slbotbm: Let's use a dataclass instead here: ` @dataclass(frozen=True, slots=True)`

examples/python/message-headers/producer.py
```diff
@@ -0,0 +1,237 @@
...
+    message_type: str
+    payload: str
+    headers: PlainHeaders | TypedHeaders
+
+
+class MessagesGenerator:
+    def __init__(self):
+        self.order_id = 0
+
+    def generate(self) -> SerializedMessage:
```
slbotbm: Let's use an iterator here instead of a full blown class

examples/python/message-headers/consumer.py
```diff
@@ -0,0 +1,170 @@
...
+        logger.info(f"Plain headers: {format_headers(headers.to_plain())}")
+
+    if message_type == ORDER_CREATED_TYPE:
+        handle_order_created(payload)
+    elif message_type == ORDER_CONFIRMED_TYPE:
+        handle_order_confirmed(payload)
+    elif message_type == ORDER_REJECTED_TYPE:
+        handle_order_rejected(payload)
+    else:
+        logger.warning(f"Received unknown message type: {message_type}")
```
slbotbm: match/case can be utilized instead of if/else

examples/python/message-headers/producer.py
```diff
@@ -0,0 +1,237 @@
...
+    args: ArgNamespace = parse_args()
+    client = IggyClient.from_connection_string(args.connection_string)
+    logger.info("Connecting to Iggy")
+    await client.connect()
+    logger.info("Connected")
+    await init_system(client)
+    await produce_messages(client)
+
+
+async def init_system(client: IggyClient):
```
slbotbm: Current system catches  stream creation and topic creation independently, logs errors, and continues with an invalid system. Unify the try/catches and exit if stream/topic is not created

foreign/python/src/user_headers.rs
```diff
@@ -0,0 +1,857 @@
...
+    /// loses information; it only returns an error if a stored field cannot be
+    /// decoded.
+    #[gen_stub(override_return_type(
+        type_repr = "dict[str | bytes | bool | int | float, str | bytes | bool | int | float]"
+    ))]
+    pub fn to_plain<'a>(slf: &Bound<'a, Self>) -> PyResult<Bound<'a, PyDict>> {
+        let py = slf.py();
+        let dict = slf.as_any().cast::<PyDict>()?;
+        let headers = py_user_headers_to_rust(py, dict)?;
+        rust_user_headers_to_plain_py(py, headers)
```
slbotbm: Let's rename to `to_scalar_dict`.

foreign/python/tests/test_message_operations.py
slbotbm, line 1: Let's add the following test cases as well:
- send message with duplicate headers should error
- a message that has > 100k bytes is not accepted, so this should be tested and should fail.
-

slbotbm (commented):
first round of review

examples/python/message-headers/producer.py
jiengup: I don't get this point.
The message header examples are implemented accroding to the existing `getting-started` example.

foreign/python/src/user_headers.rs
```diff
@@ -0,0 +1,919 @@
...
+    /// decoded.
+    #[gen_stub(override_return_type(
+        type_repr = "dict[str | bytes | bool | int | float, str | bytes | bool | int | float]"
+    ))]
+    pub fn to_scalar_dict<'a>(slf: &Bound<'a, Self>) -> PyResult<Bound<'a, PyDict>> {
+        let py = slf.py();
+        let dict = slf.as_any().cast::<PyDict>()?;
+        let headers = py_user_headers_to_rust(py, dict)?;
+        rust_user_headers_to_plain_py(py, headers)
+    }
```
slbotbm, line 660: This function is not lossless - if the user supplies `UnsignedInt8(1)` and `UnsignedInt16(1)`, both get converted to python int `1`. In this case, the conversion should fail, and be documented properly.

slbotbm: Also, the current code converts twice before returning the headers back to the user: python typed headers -> rust headers -> python plain headers. It would be preferable to do a direct conversion python typed header -> python plain header. `TryFrom` can be defined and utilized.

slbotbm (changes requested):
Some more comments: 
- You have set the type of user headers in some places to [Any, Any], which makes any type checker incapable of catching mistakes. Would be better to explicitly declare the types.

foreign/python/tests/test_message_operations.py
```diff
@@ -16,13 +16,26 @@
...
+    IggyClient,
+    PollingStrategy,
+    UserHeaders,
+)
 from apache_iggy import SendMessage as Message
 
+HTTP_CREATED = 201
+JsonValue: TypeAlias = (
+    None | bool | int | float | str | list["JsonValue"] | dict[str, "JsonValue"]
+)
```
slbotbm: These are never used. Can be removed.

foreign/python/tests/test_message_operations.py
slbotbm, line 1: The logging setup in the tests you have defined aren't really useful.

foreign/python/tests/test_message_operations.py
```diff
@@ -136,8 +149,160 @@ async def test_message_properties(self, iggy_client: IggyClient, unique_name):
...
+        """Test typed user headers preserve explicit header kinds."""
+        stream_name = unique_name()
+        topic_name = unique_name()
+        partition_id = 0
+        typed_key = HeaderKey.UnsignedInt128(42)
+        typed_value = HeaderValue.UnsignedInt128(2**96)
+        user_headers: dict[HeaderKey, HeaderValue] = {
+            typed_key: typed_value,
+            HeaderKey.String("float32"): HeaderValue.Float32(1.25),
+        }
```
slbotbm: This can be made into one dict. No need to declare extras vars.

examples/python/message-headers/producer.py
slbotbm: In common.py, in the init_system function, topic creation will be tried even if the stream creation fails. Since topics reside inside streams in iggy, it will not be created and will fail. Better to change the code so that topic should only be created if the stream exists/is created.

examples/python/message-headers/plain-headers/producer.py
```diff
@@ -0,0 +1,59 @@
...
+        "priority-score": 0.5,
+        "trace-bin": order.payload.encode(),
+    }
+
+
+async def main() -> None:
+    args = parse_args()
+    client = await connect(args.connection_string)
+    await init_system(client)
+    await produce_messages(client, build_plain_headers)
```
slbotbm, line 55: Even if  init_system fails, this code will not know since init_system logs the error and returns normally. 

In produce_messages, the loop termination condition depends on successful sends, but if the send fails, `sent_batches` is not incremented after a failed send, a persistent configuration, authorization, or initialization failure makes the example run forever. It prints repeated errors instead of terminating with a useful failure.

examples/python/message-headers/common.py
```diff
@@ -0,0 +1,247 @@
...
+        ),
+        default="iggy+tcp://iggy:iggy@127.0.0.1:8090",
+        nargs="?",
+        type=str,
+    )
+    return ArgNamespace(**vars(parser.parse_args()))
+
+
+def generate_orders() -> Iterator[Order]:
+    order_id = 0
```
slbotbm, line 85: In this function, the generated `order_id`s are not coherent. The current code will effectively produce the following 6 events:
```
OrderConfirmed(order-0)
OrderRejected(order-0)
OrderCreated(order-3)
OrderConfirmed(order-1)
OrderRejected(order-1)
OrderCreated(order-6)
```

examples/python/README.md
```diff
@@ -71,6 +71,29 @@ python basic/consumer.py
...
 Demonstrates fundamental client connection, authentication, batch message sending, and polling with support for TCP/QUIC/HTTP protocols.
 
+### Message Headers
+
+Shows how to attach and read Python SDK user headers with `str`, `bytes`, `bool`, `int`, and `float` values. Two variants share their logic through `message-headers/common.py`:
+
+- `plain-headers/` uses the convenient `dict[str, str | bytes | bool | int | float]` form; the SDK infers a wire type for each value.
+- `typed-headers/` uses explicit `HeaderKey`/`HeaderValue` for full control over the wire type.
+
+Both are consumed identically, since plain headers are stored as typed headers on the wire.
```
slbotbm: Changing this to 
> Both producers store typed headers on the wire. The plain consumer converts them to Python scalars, while the typed consumer preserves and inspects the explicit header kinds.

would be better

slbotbm (changes requested):

foreign/python/src/user_headers.rs
jiengup: > if the user supplies `UnsignedInt8(1)` and `UnsignedInt16(1)`

I don't get this.
Lossless here means Rust type conversion into Python type should not cost any data loss (e.g. uint32 to py `Int` because Python can represent any large int). 
You mean we should reject any coexisting `UnsignedInt8(1)` and `UnsignedInt16(1)` or we should reject `UnsignedInt16(-1)`? I think the former is not reasonable.

slbotbm: Let me think a little about this and get back to you

jiengup: @slbotbm Hi, do you have any thoughts about it?

slbotbm: give me until the weekend😅

slbotbm: give me until the weekend😅

slbotbm: @jiengup Apologize for the delay.

```python
headers = UserHeaders({
      HeaderKey.UnsignedInt8(1): HeaderValue.String("u8"),
      HeaderKey.UnsignedInt16(1): HeaderValue.String("u16"),
  })
```
Let's take the above example. These are distinct typed keys because their `HeaderKind` is part of their identity, so both can coexist in `UserHeaders` and the Rust `BTreeMap`. However, `to_scalar_dict()` converts both keys to Python `int(1)`. The second `set_item()` then silently overwrites the first entry. 

 The typed headers should remain valid. Only `to_scalar_dict()` should fail with `ValueError` when distinct typed keys collapse to the same Python key. 

What do you think?

Also, the existing `test_plain_user_headers_convert_every_kind_losslessly` does not cover this case. It converts different value kinds under unique string keys, proving only that each individual scalar value is representable. If this seems reasonable, please add a test using colliding typed keys such as the example above.

jiengup: Oh, I see it. 
I'll handle it and add a test case for that.

foreign/python/src/user_headers.rs
```diff
@@ -0,0 +1,897 @@
...
+    py: Python<'py>,
+    any: &Bound<'py, PyAny>,
+) -> PyResult<Bound<'py, PyAny>> {
+    if let Ok(header) = any.extract::<PyRef<'_, T>>() {
+        return Bound::<PyAny>::try_from(HeaderToPlainRef {
+            py,
+            value: &*header,
+        });
+    }
+    Ok(any.clone())
```
slbotbm: Right now, since `UserHeaders` is a
publicly constructible and mutable `dict`, this is possible:

```python
UserHeaders({object(): object()}).to_scalar_dict()
```

The resulting dictionary can contain arbitrary objects despite the documented
and generated scalar-only return type.

Validate entries during construction and mutation, or reject unsupported
objects inside `to_scalar_dict()`. Add tests for unsupported keys and values
inserted through both the constructor and normal dictionary mutation.

foreign/python/src/user_headers.rs
```diff
@@ -0,0 +1,897 @@
...
+/// Uses `Float32` when the value is exactly representable as an f32, otherwise
+/// `Float64`, so the plain float is always stored losslessly.
+fn float_to_header_field<T>(value: f64) -> HeaderField<T> {
+    let narrowed = value as f32;
+    if f64::from(narrowed) == value {
+        narrowed.into()
+    } else {
+        value.into()
+    }
+}
```
slbotbm: This accepts plain `float("inf")` and stores it as `Float32`, because narrowing infinity to `f32` still compares equal after widening. In contrast, `checked_float32` reject the equivalent typed `HeaderValue.Float32(float("inf"))`.

This also means a `Float32` infinity or NaN received from a Rust client can be
represented by Python but cannot be sent again. `Float64` and plain NaN follow
different paths again.

Let's reject non-finite headers everywhere in plain, Float32, and Float64 paths.

Add tests for positive and negative infinity and NaN across plain, typed, and
receive-then-resend paths.

foreign/python/src/receive_message.rs
```diff
@@ -72,6 +81,19 @@ impl ReceiveMessage {
...
+    pub fn user_headers<'a>(&self, py: Python<'a>) -> PyResult<Option<Bound<'a, UserHeaders>>> {
+        let Some(headers) = self
+            .inner
+            .user_headers_map()
+            .map_err(|e| PyValueError::new_err(e.to_string()))?
+        else {
+            return Ok(None);
+        };
+        rust_user_headers_to_py(py, headers).map(Some)
+    }
```
slbotbm, line 100: The linked issue notes that structurally malformed wire headers are logged and
returned as `None`, making them indistinguishable from absent headers.

Document the `None` behavior and distinguish it from known semantic decode
errors, which become `ValueError`.

slbotbm (changes requested):

foreign/python/src/user_headers.rs
jiengup: Now the latest version block user from sending and receiving `inf` or `nan` float(64) through Python SDK.

But I noticed that `inf` survives in Rust core. Maybe we should leave a TBD as one of these:
- ban `inf` and `nan` float in the user header of Rust core and any other SDK.
- carefully support `inf` and `nan` across Rust core, Python SDK, plain and typed header

slbotbm: @hubcio any thoughts on this?

slbotbm:
@jiengup Could you resolve the conflicts? Other than that LGTM

slbotbm (approved):

hubcio (approved):

PR #3615, opened 2026-07-06
feat(server): respect CPU affinity and cgroup limits in shards and stats

mmodzelewski:
The server assumed it owned the whole host. Shards pinned to
absolute core ids 0..n, which fails with EINVAL (or lands on
forbidden cores) under systemd AllowedCPUs=, container cpusets,
or taskset. Stats reported host-wide CPU and memory, so a
confined instance showed its neighbors' load and a meaningless
host total.

Draw pinned cores from the process's allowed CPU set
(sched_getaffinity) instead, and validate configured ranges
against it at boot. A new pin_cores knob (default true) turns
pinning off for shared-core hosts (cgroup CPU quotas), where
every tenant pinning to the same low-numbered cores would pile
onto one core; unpinned shards let the kernel scheduler place
threads freely.

Stats now scope total_cpu_usage to the allowed cores and report
the effective cgroup memory limit as total_memory. Available
memory adds reclaimable file cache back
(limit - (current - inactive_file - active_file), minimum over
capped ancestors, v1 and v2), since memory.current charges page
cache as used and the naive limit - current trends to zero on a
cache-heavy server long before real OOM pressure.


core/shard_allocator/src/lib.rs
```diff
@@ -287,39 +335,66 @@ impl ShardAllocator {
...
 
                 Ok(shard_assignments)
             }
             CpuAllocation::Range(start, end) => {
+                if !self.pin_cores {
+                    info!(
+                        "Using {} shard(s) for range {start}..{end}, unpinned",
+                        end - start
+                    );
+                    return Ok(unpinned(end - start));
```
hubcio, line 396: `unpinned(end - start)` underflows when `start > end` - debug build panics on the subtract, release wraps to a huge value and `unpinned()` then tries to allocate a vec of ~`usize::MAX` `ShardInfo` and aborts. not reachable in production today since both boot paths run `ShardingConfig::validate` first and it rejects `start >= end`, but `to_shard_assignments` is a pub fn in an internal crate and the server-ng `validate_sharding_runtime_knobs` net doesn't cover cpu_allocation, so a direct/test/embedded caller hits it. cheap fix - reject `start >= end` in this arm so the allocator self-guards instead of trusting the caller.

core/server/src/shard/system/stats.rs
```diff
@@ -88,6 +88,14 @@ impl IggyShard {
                 stats.written_bytes = disk_usage.total_written_bytes.into();
 
                 stats.threads_count = process.tasks().map(|t| t.len() as u32).unwrap_or(0);
+
+                if let Some(limits) = process
+                    .cgroup_limits()
+                    .filter(|limits| limits.total_memory < sys.total_memory())
+                {
+                    stats.total_memory = limits.total_memory.into();
+                    stats.available_memory = limits.free_memory.into();
```
hubcio: under a memory-capped cgroup this sets `available_memory` to `memory.max - memory.current` (or lower if a parent cgroup is tighter), and `memory.current` charges reclaimable page cache as used. iggy is cache-heavy, so on a warm instance `memory.current` climbs toward the cap and `available_memory` trends toward ~0 even though the kernel reclaims that cache long before OOM - it under-reports real headroom, which is the opposite of what a per-instance stats endpoint wants. for an honest number you'd want a cgroup MemAvailable-equivalent: `memory.max - (memory.current - reclaimable file cache)`, where reclaimable is `inactive_file + active_file` from `memory.stat` - sysinfo's `free_memory` doesn't do that, so it needs reading `memory.stat` directly. also this override only fires when a memory limit is set (the `total < host total` filter just above), so an instance confined by cpuset alone still reports host-wide memory - multi-instance setups need a memory cap too. re-scoping `total_memory` to the cap is correct.

core/configs/src/server_config/validators.rs
```diff
@@ -488,9 +489,27 @@ impl Validatable<ConfigurationError> for ShardingConfig {
...
+                    if *end - *start > available_cpus {
+                        eprintln!(
+                            "Invalid sharding configuration: cpu_allocation range {start}..{end} yields {} shards, exceeding available CPU cores {available_cpus}",
+                            *end - *start
+                        );
+                        return Err(ConfigurationError::InvalidConfigurationValue);
+                    }
+                    return Ok(());
+                }
+                if *end - *start > available_cpus {
```
hubcio, line 492: this `*end - *start > available_cpus` check and its `eprintln!` are identical to the one in the `!self.pin_cores` branch above. hoist it above the `if !self.pin_cores` split so it runs once, and leave only the allowed-set membership check gated behind pinning. the `start >= end` guard earlier still protects the subtraction. saves ~7 lines.

core/shard_allocator/src/lib.rs
```diff
@@ -287,39 +335,66 @@ impl ShardAllocator {
...
-                        numa_node: None,
-                    })
+                if !self.pin_cores {
+                    info!("Using {count} shard(s), unpinned");
+                    return Ok(unpinned(*count));
+                }
+
+                let allowed = allowed_cpus();
+                let shard_assignments = pinned_from_allowed(&allowed, *count)?;
+                let cores: Vec<usize> = shard_assignments
```
hubcio: this `cores` vec is only built to log it, and since `pinned_from_allowed` takes `allowed.iter().take(count)` in order it's just `&allowed[..*count]`. log the slice directly and drop the vec + `flat_map`.

core/shard_allocator/src/lib.rs
```diff
@@ -287,39 +335,66 @@ impl ShardAllocator {
...
             CpuAllocation::Range(start, end) => {
+                if !self.pin_cores {
+                    info!(
+                        "Using {} shard(s) for range {start}..{end}, unpinned",
+                        end - start
+                    );
+                    return Ok(unpinned(end - start));
+                }
+
+                // An explicit range is operator intent for exact cores; honor
```
hubcio: the comment says the range is honored as-is, but the validator already caps a pinned range by `available_parallelism()` before this runs, so a range wider than the quota never reaches here. minor, but the comment overstates what this arm actually sees.

hubcio (changes requested):
one gap relative to the true-per-instance-stats goal: `total_cpu_usage` (stats.rs:47, `sys.global_cpu_usage()`) is host-global and untouched by this PR, so on a shared machine each instance reports the whole box's cpu including its neighbors - `cpu_usage` (the per-process value) is correct but `total_cpu_usage` isn't per-slice. it's pre-existing and outside this diff so it can't be a line comment, but if per-instance accuracy is the aim it's worth a follow-up: average `sys.cpus()[i].cpu_usage()` over the cores in `allowed_cpus()` (the cpuset confines those cores to this instance).

hubcio (approved):

spetz (approved):

core/configs/src/server_config/sharding.rs
```diff
@@ -101,6 +101,10 @@ const fn default_inbox_capacity() -> usize {
     DEFAULT_INBOX_CAPACITY
 }
 
+const fn default_pin_cores() -> bool {
+    true
+}
```
hubcio, line 106: dont do it like so, default value should be read from toml, see how other config fields default values are assumed

hubcio (approved):

core/configs/src/server_config/sharding.rs
hubcio: nvm, i'll do that

PR #3621, opened 2026-07-06
feat(go): add diagnostic event subscriptions to TCP client

### LINKED ISSUE #3219: feat(go): add diagnostic event subscription to client
## Summary

The Go SDK currently has no way for callers to observe client-level lifecycle events (connection state changes, auth state changes). The Rust SDK exposes these via `Client::subscribe_events()` returning a stream of `DiagnosticEvent`. This issue ports that capability to the Go SDK.

## Motivation

The Go TCP client already has TODO stubs in place for event emission, indicating this was anticipated:
- `foreign/go/client/tcp/tcp_core.go:399` — `// TODO publish event`
- `foreign/go/client/tcp/tcp_core.go:461` — `// TODO event pushing logic`
- `foreign/go/client/tcp/tcp_core.go:475` — `// TODO push shutdown event.

This is groundwork for adding a high-level `IggyConsumer` abstraction to the Go SDK (planned as a follow-up). Diagnostic events let the consumer detect disconnects, re-join consumer groups eagerly on reconnect, and resume cleanly. Beyond the consumer use case, the events are also useful for application-level observability (logging connection drops, exposing health metrics, etc.).

## Scope

In scope:
1. Add a `DiagnosticEvent` type in `foreign/go/contracts/` mirroring the Rust enum (5 variants, with a `String()` method matching the `snake_case` form used in Rust).
2. Add `SubscribeEvents() <-chan DiagnosticEvent` to the `Client` interface in `foreign/go/contracts/client.go`.
3. Implement a small fan-out broadcaster in the TCP client so multiple subscribers can each receive every event independently.
4. Wire emission into the TCP client:
   - `Connected` — on initial connect and on every successful auto-reconnect (existing reconnect loop at `foreign/go/client/tcp/tcp_core.go:86-100, 351-401`)
   - `Disconnected` — at the existing TODO at `tcp_core.go:399`
   - `Shutdown` — at the existing TODO at `tcp_core.go:475`
   - `SignedIn` — after successful `LoginUser`
   - `SignedOut` — after successful `LogoutUser`
5. Integration test that subscribes, exercises connect → disconnect → reconnect → login → logout → close, and asserts the full event sequence is delivered.

### Component

Go SDK

### Proposed solution

_No response_

### Alternatives considered

_No response_

explise:
## Which issue does this PR address

Closes #3219

## Rationale

The TCP client had `// TODO publish event` stubs at the connect/disconnect/
shutdown paths with no way for callers to observe connection or auth state
changes. This fills those in and is groundwork for a high-level `IggyConsumer`
abstraction: diagnostic events let the consumer detect disconnects and
re-join consumer groups eagerly on reconnect. They're also useful on their own
for application-level observability (logging drops, health metrics).

## What changed?

- `DiagnosticEvent` enum + `String()` in `contracts/`, and `SubscribeEvents()`
  on the `Client` interface.
- A small non-blocking broadcaster in the TCP client fanning out to multiple
  independent subscribers.
- Emission wired into connect (`Connected`), disconnect/failed-connect
  (`Disconnected`), shutdown (`Shutdown`), login (`SignedIn`) and logout
  (`SignedOut`).

## Local Execution

- **Passed.** `go build ./...`, `go vet ./...`, `gofmt`, and `go mod tidy` are clean.
- Unit tests pass: `DiagnosticEvent.String`, and the broadcaster tests
  (fan-out / ordering / slow-subscriber-drop / idempotent-close).
- Integration tests pass — all 5 in `tests/diagnostic_events_test.go`
  (login → SignedIn, logout → SignedOut, close → Shutdown + channel close,
  multiple subscribers, and the full connect → login → logout → close sequence),
  run against a real `apache/iggy:edge` container via testcontainers.
- **Pre-commit hooks ran** via `prek` (pre-commit + pre-push), including
  `golangci-lint` on `foreign/go` — passed.

## AI Usage

1. **Tool:** Claude Code.
2. **Scope:** Implementation of the enum, broadcaster, and event wiring;
   rebase/conflict resolution onto latest `master`; test authoring.
3. **Verification:** Local `go build`/`go vet`/unit tests, `golangci-lint`, and
   manual review of every changed line against the surrounding code.
4. **Can I explain every line?** Yes.

### DIFF
--- a/foreign/go/client/tcp/tcp_core.go
+++ b/foreign/go/client/tcp/tcp_core.go
@@ -60,6 +60,7 @@ type IggyTcpClient struct {
 	currentServerAddress   string
 	connectedAt            time.Time
 	state                  iggcon.State
+	events                 *eventBroadcaster
 	// respHeader is the reused response-status read buffer; guarded by c.mtx.
 	respHeader [ResponseInitialBytesLength]byte
 }
@@ -220,6 +221,7 @@ func NewIggyTcpClient(logger *slog.Logger, options ...Option) *IggyTcpClient {
 		connectedAt:            time.Time{},
 		leaderRedirectionState: iggcon.LeaderRedirectionState{},
 		currentServerAddress:   opts.config.serverAddress,
+		events:                 newEventBroadcaster(),
 	}
 }
 
@@ -431,12 +433,16 @@ func (c *IggyTcpClient) sendLocked(wirePayload []byte) ([]byte, error) {
 	return buffer, nil
 }
 
-// invalidateConnLocked closes the connection and marks it as disconnected
+// invalidateConnLocked closes the connection and marks it as disconnected,
+// publishing a Disconnected event on the transition.
 func (c *IggyTcpClient) invalidateConnLocked() {
 	if c.conn != nil {
 		_ = c.conn.Close()
 	}
-	c.state = iggcon.StateDisconnected
+	if c.state != iggcon.StateDisconnected {
+		c.state = iggcon.StateDisconnected
+		c.events.publish(iggcon.DiagnosticEventDisconnected)
+	}
 }
 
 func (c *IggyTcpClient) GetConnectionInfo() *iggcon.ConnectionInfo {
@@ -548,7 +554,7 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
 		if !c.config.reconnection.enabled {
 			c.logger.Warn("Automatic reconnection is disabled.")
 		}
-		// TODO publish event disconnected
+		c.events.publish(iggcon.DiagnosticEventDisconnected)
 		return err
 	}
 
@@ -558,6 +564,7 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
 	c.connectedAt = time.Now()
 	c.logger.Info("Iggy client has connected to the Iggy server", slog.String("client_address", c.clientAddress), slog.String("server_address", c.currentServerAddress))
 	c.mtx.Unlock()
+	c.events.publish(iggcon.DiagnosticEventConnected)
 	return nil
 }
 
@@ -624,7 +631,7 @@ func (c *IggyTcpClient) disconnect() error {
 	}
 
 	c.logger.Info("Iggy client has disconnected from server.", slog.String("client_address", c.clientAddress))
-	// TODO event pushing logic
+	c.events.publish(iggcon.DiagnosticEventDisconnected)
 	return nil
 }
 
@@ -646,10 +653,18 @@ func (c *IggyTcpClient) shutdown() error {
 
 	c.state = iggcon.StateShutdown
 	c.logger.Info("Iggy TCP client has been shutdown.", slog.String("client_address", c.clientAddress))
-	// TODO push shutdown event
+	c.events.publish(iggcon.DiagnosticEventShutdown)
+	c.events.close()
 	return nil
 }
 
 func (c *IggyTcpClient) Close() error {
 	return c.shutdown()
+}
+
+// SubscribeEvents returns an independent channel of client lifecycle events
+// and an unsubscribe function. The channel is closed on shutdown, after the
+// final DiagnosticEventShutdown.
+func (c *IggyTcpClient) SubscribeEvents() (<-chan iggcon.DiagnosticEvent, func()) {
+	return c.events.subscribe()
 }
--- a/foreign/go/client/tcp/tcp_events.go
+++ b/foreign/go/client/tcp/tcp_events.go
@@ -1,0 +1,116 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package tcp
+
+import (
+	"sync"
+
+	iggcon "github.com/apache/iggy/foreign/go/contracts"
+)
+
+// subscriberBufferSize bounds each subscriber's channel. A subscriber that
+// fails to drain within this many events will start dropping the oldest
+// events; the publisher is never blocked.
+const subscriberBufferSize = 1000
+
+// eventBroadcaster fans out diagnostic events to any number of subscribers.
+// Each call to subscribe returns an independent channel; publish delivers
+// the event to every live subscriber non-blockingly. close releases all
+// subscriber channels and is idempotent.
+type eventBroadcaster struct {
+	mu          sync.Mutex
+	subscribers []chan iggcon.DiagnosticEvent
+	closed      bool
+}
+
+func newEventBroadcaster() *eventBroadcaster {
+	return &eventBroadcaster{}
+}
+
+// subscribe returns a new subscriber channel and an unsubscribe function that
+// is idempotent and safe to call after the broadcaster is closed.
+func (b *eventBroadcaster) subscribe() (<-chan iggcon.DiagnosticEvent, func()) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	ch := make(chan iggcon.DiagnosticEvent, subscriberBufferSize)
+	if b.closed {
+		close(ch)
+		return ch, func() {}
+	}
+	b.subscribers = append(b.subscribers, ch)
+
+	var once sync.Once
+	unsubscribe := func() {
+		once.Do(func() {
+			b.mu.Lock()
+			defer b.mu.Unlock()
+			if b.closed {
+				return
+			}
+			for i, sub := range b.subscribers {
+				if sub == ch {
+					b.subscribers = append(b.subscribers[:i], b.subscribers[i+1:]...)
+					close(ch)
+					return
+				}
+			}
+		})
+	}
+	return ch, unsubscribe
+}
+
+func (b *eventBroadcaster) publish(event iggcon.DiagnosticEvent) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if b.closed {
+		return
+	}
+	for _, ch := range b.subscribers {
+		select {
+		case ch <- event:
+		default:
+			// Subscriber is not draining fast enough. Drop the oldest event
+			// to make room for the newest, preferring recency. Keeps the
+			// publisher non-blocking even with a slow subscriber.
+			select {
+			case <-ch:
+			default:
+			}
+			select {
+			case ch <- event:
+			default:
+			}
+		}
+	}
+}
+
+func (b *eventBroadcaster) close() {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if b.closed {
+		return
+	}
+	b.closed = true
+	for _, ch := range b.subscribers {
+		close(ch)
+	}
+	b.subscribers = nil
+}
--- a/foreign/go/client/tcp/tcp_session_management.go
+++ b/foreign/go/client/tcp/tcp_session_management.go
@@ -51,6 +51,7 @@ func (c *IggyTcpClient) LoginUser(ctx context.Context, username string, password
 		}
 		return c.LoginUser(ctx, username, password)
 	}
+	c.events.publish(iggcon.DiagnosticEventSignedIn)
 	return identity, nil
 }
 
@@ -75,12 +76,16 @@ func (c *IggyTcpClient) LoginWithPersonalAccessToken(ctx context.Context, token 
 		}
 		return c.LoginWithPersonalAccessToken(ctx, token)
 	}
+	c.events.publish(iggcon.DiagnosticEventSignedIn)
 	return identity, nil
 }
 
 func (c *IggyTcpClient) LogoutUser(ctx context.Context) error {
-	_, err := c.do(ctx, &command.LogoutUser{})
-	return err
+	if _, err := c.do(ctx, &command.LogoutUser{}); err != nil {
+		return err
+	}
+	c.events.publish(iggcon.DiagnosticEventSignedOut)
+	return nil
 }
 
 func (c *IggyTcpClient) HandleLeaderRedirection(ctx context.Context) (bool, error) {
@@ -116,6 +121,8 @@ func (c *IggyTcpClient) HandleLeaderRedirection(ctx context.Context) (bool, erro
 		return false, nil
 	}
 	c.mtx.Unlock()
+
+	c.events.publish(iggcon.DiagnosticEventRedirected)
 
 	if err = c.disconnect(); err != nil {
 		return false, err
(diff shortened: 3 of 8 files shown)


chengxilo (changes requested):
I add some suggestions. I think it will worth it to have a BDD test to test the behavior of Event subscription, we can do it in a seperate PR.

foreign/go/client/tcp/tcp_core.go
chengxilo:
```suggestion
	c.state = iggcon.StateDisconnected
	c.events.publish(iggcon.DiagnosticEventDisconnected)
```

As `invalidateConnLocked` also set the state to `StateDisconnected`, maybe we should also publish the event here :)

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -41,6 +41,7 @@ func (c *IggyTcpClient) LoginUser(ctx context.Context, username string, password
 
 	c.logger.Info("Iggy client has signed in successfully.", slog.String("client_address", c.clientAddress))
 	identity := binaryserialization.DeserializeLogInResponse(buffer)
+	c.events.publish(iggcon.DiagnosticEventSignedIn)
```
chengxilo:
```suggestion
```
Publishing the SignedIn event here means multiple SignedIn events will be emitted during a redirection. Like this:

```
  Connected
  SignedIn
  Disconnected
  Connected
  SignedIn
```

While this is expected behavior and not a bug, we could make the flow more elegant and informative. I suggest publishing a DiagnosticEventRedirected during redirection to clarify that a redirect is occurring rather than a glitch. See my following comment for the suggested implementation : )

foreign/go/client/tcp/tcp_session_management.go
chengxilo, line 55:
```suggestion
	c.events.publish(iggcon.DiagnosticEventSignedIn)
	return identity, nil
```

After all these changes, when redirection happens, the event would be:
```
  Connected
  Redirected
  Disconnected
  Connected
  SignedIn
```

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -41,6 +41,7 @@ func (c *IggyTcpClient) LoginUser(ctx context.Context, username string, password
...
 	shouldRedirect, err := c.HandleLeaderRedirection(ctx)
```
chengxilo, line 44: In `HandleLeaderRedirection`, publish the `DiagnosticEventRedirected`. Because this file is not touched, I chose to comment here

foreign/go/client/tcp/tcp_session_management.go
chengxilo, line 58: same as above.

foreign/go/contracts/client.go
```diff
@@ -29,6 +29,10 @@ type Client interface {
 	// GetConnectionInfo returns the current connection information including protocol and server address
 	GetConnectionInfo() *ConnectionInfo
 
+	// SubscribeEvents returns an independent channel of client lifecycle events.
+	// The channel is closed on shutdown, after the final DiagnosticEventShutdown.
+	SubscribeEvents() <-chan DiagnosticEvent
```
chengxilo: We should also support unsubscribe. Let this function return a `func()` that can unsubscribe it. So we can call it like:
```go
events, cancel : = SubscribeEvents()
defer cancel ()

// .... use events
```

foreign/go/tests/diagnostic_events_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+// TestDiagnosticEvents_LoginEmitsSignedIn verifies that LoginUser emits
+// a SignedIn event after a successful authentication.
+func TestDiagnosticEvents_LoginEmitsSignedIn(t *testing.T) {
+	container, addr := setupPlainContainer(t)
+	defer func() {
+		if err := container.Terminate(context.Background()); err != nil {
+			t.Logf("Failed to terminate container: %v", err)
+		}
+	}()
```
chengxilo: nit: `defer`s can be replaced by  adding `t.Cleanup()` in setupPlainContainer.

And actually, I think it's a little bit over kill to create container over and over again, it should be better to just use `TestMain`. We create the container, run the tests, and terminate it.

foreign/go/client/tcp/tcp_session_management.go
explise: rust doesn't have anything similiar and adding redirection breaks the parity here @hubcio  what do u think about it here

hubcio: unfortunately my golang knowledge is second to none, sorry. @chengxilo any thoughts on above question?

from Rust perspective, feel free to break parity. i dont see any reason to stick to current approach, if you can figure out something better maybe we even can implement it in Rust SDK also.

chengxilo: @explise Your concern is correct, I didn't thought it throughly. Technically, because the redirect needs to call `GetClusterMetadata`, a session is required. It's true that we will log in every time before we redirect, so keeping the current approach might be the best choice(publish the login event multiple times, since it's true that we logined). While they are noise, they tell the subscriber these events actually happened.

So it may be better to keep this `publish_event...`, ignore my suggestion to move this to the end of SignIn function.

But I still believe it might be better to add the redirect event. I checked the code of IggyProducer and IggyConsumer, if the redirect event is provided, it would help it to filter the noise during redirection. For example, when received the redirect event, ignore the following `disconnect`, `connect`,`disconnect`. @hubcio would that help?

foreign/go/client/tcp/tcp_session_management.go
chengxilo: This one can be ignored.

foreign/go/tests/diagnostic_events_test.go
explise: fixed

chengxilo:
btw you don't have to keep amend and force push. The pr would be squashed before merge anyway.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -431,12 +433,16 @@ func (c *IggyTcpClient) sendLocked(wirePayload []byte) ([]byte, error) {
...
+// invalidateConnLocked closes the connection and marks it as disconnected,
+// publishing a Disconnected event on the transition.
 func (c *IggyTcpClient) invalidateConnLocked() {
 	if c.conn != nil {
 		_ = c.conn.Close()
 	}
-	c.state = iggcon.StateDisconnected
+	if c.state != iggcon.StateDisconnected {
+		c.state = iggcon.StateDisconnected
+		c.events.publish(iggcon.DiagnosticEventDisconnected)
```
ryankert01, line 444: newTestClient create but will not init it, so it will failed.

explise: i have resolved it, can u check again please

foreign/go/contracts/client.go
explise: fixed

foreign/go/client/tcp/tcp_session_management.go
explise: fixed

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -117,6 +122,8 @@ func (c *IggyTcpClient) HandleLeaderRedirection(ctx context.Context) (bool, erro
 	}
 	c.mtx.Unlock()
 
+	c.events.publish(iggcon.DiagnosticEventRedirected)
```
chengxilo, line 125: It may be a better choice to move this to the last part of this function, i.e. after the `disconnect()` call.
```go
c.mtx.Lock()
...
c.events.publish(iggcon.DiagnosticEventRedirected)
c.mtx.Unlock()
```
When we receive false, we will not handle the redirect but return an error.

foreign/go/client/tcp/tcp_events.go
```diff
@@ -0,0 +1,116 @@
...
+			// to make room for the newest, preferring recency. Keeps the
+			// publisher non-blocking even with a slow subscriber.
+			select {
+			case <-ch:
+			default:
+			}
+			select {
+			case ch <- event:
+			default:
+			}
```
chengxilo, line 99:
```suggestion
			ch <- event
```
This should be enough. 

There are two possible situation for the previous select {}. 
First one, `case <- ch:`, it means there is one free space in `ch`, it won't be stucked.
Second one, `default`, it means `ch` is empty, it won't be stucked neither. 

So `ch <- event` without the `select { default: }` is already enough. : )

foreign/go/client/tcp/tcp_events.go
```diff
@@ -0,0 +1,116 @@
...
+	b.subscribers = append(b.subscribers, ch)
+
+	var once sync.Once
+	unsubscribe := func() {
+		once.Do(func() {
+			b.mu.Lock()
+			defer b.mu.Unlock()
+			if b.closed {
+				return
+			}
```
chengxilo, line 65:
```suggestion
```

nit: when eventBroadcaster is closed, the slice would be nil, it's fine to remove this guard in my opinion.

foreign/go/client/tcp/tcp_events.go
```diff
@@ -0,0 +1,116 @@
...
+			for i, sub := range b.subscribers {
+				if sub == ch {
+					b.subscribers = append(b.subscribers[:i], b.subscribers[i+1:]...)
```
chengxilo, line 68: nit: `slices.Delete` is recommended.

foreign/go/client/tcp/tcp_events.go
```diff
@@ -0,0 +1,116 @@
...
+	ch := make(chan iggcon.DiagnosticEvent, subscriberBufferSize)
+	if b.closed {
+		close(ch)
+		return ch, func() {}
+	}
+	b.subscribers = append(b.subscribers, ch)
+
+	var once sync.Once
+	unsubscribe := func() {
+		once.Do(func() {
```
chengxilo, line 60: nit: `once.Do` can also be waived, we will iterate the slice and find the channel we want to remove, if it is already removed we will do nothing. 

I think iterate on a short slice will cost less resource than `once.Do`. It's good and safe to have it, but I think it's fine to waive it.

foreign/go/client/tcp/tcp_core.go
chengxilo, line 652: Replace these code with:

```go
        var err error
        if c.conn != nil {
                err = c.conn.Close()
        }
        c.conn = nil
```

If we return the error directly, we won't be able to close the eventBroadcaster.

I also add `c.conn = nil`, it's not a problem introduced by this PR, but its an historical issue that we never set the connection to nil after closeing it while we already have the nil guard.

Similarly, I add suggestion to set `c.conn = nil` under `invalidateConnLocked` and `disconnect` too

foreign/go/client/tcp/tcp_core.go
```diff
@@ -624,7 +631,7 @@ func (c *IggyTcpClient) disconnect() error {
 	}
```
chengxilo, line 631: Save the error and return at the end of this function, also, set c.conn to nil.
```go
    var err error
    if c.conn != nil {
		err = c.conn.Close()
	}
	c.conn = nil
```

See my coment in `shutdown` for detail.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -431,12 +433,16 @@ func (c *IggyTcpClient) sendLocked(wirePayload []byte) ([]byte, error) {
 	return buffer, nil
 }
 
-// invalidateConnLocked closes the connection and marks it as disconnected
+// invalidateConnLocked closes the connection and marks it as disconnected,
+// publishing a Disconnected event on the transition.
 func (c *IggyTcpClient) invalidateConnLocked() {
 	if c.conn != nil {
 		_ = c.conn.Close()
```
chengxilo, line 440:
```suggestion
		_ = c.conn.Close()
		c.conn = nil
```

See my coment in `shutdown` for detail.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -558,6 +564,7 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
 	c.connectedAt = time.Now()
 	c.logger.Info("Iggy client has connected to the Iggy server", slog.String("client_address", c.clientAddress), slog.String("server_address", c.currentServerAddress))
 	c.mtx.Unlock()
+	c.events.publish(iggcon.DiagnosticEventConnected)
```
chengxilo, line 567: maybe we can do this before `c.mtx.Unlock()`.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -548,7 +554,7 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
 		if !c.config.reconnection.enabled {
 			c.logger.Warn("Automatic reconnection is disabled.")
 		}
-		// TODO publish event disconnected
+		c.events.publish(iggcon.DiagnosticEventDisconnected)
```
chengxilo, line 557: maybe we can do this before c.mtx.Unlock()

foreign/go/client/tcp/tcp_core.go
```diff
@@ -60,6 +60,7 @@ type IggyTcpClient struct {
 	currentServerAddress   string
 	connectedAt            time.Time
 	state                  iggcon.State
+	events                 *eventBroadcaster
```
chengxilo, line 63: nit: I think we don't have to use pointer for it. eventBroadcaster should be enough. And when initialize it we can use `events: eventBroadcaster{}`

foreign/go/tests/diagnostic_events_test.go
```diff
@@ -0,0 +1,268 @@
...
+// TestMain starts one Iggy server container for the whole package, runs the
+// tests against it, and tears it down.
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+
+	container, addr, err := startPlainContainer(ctx)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "diagnostic_events_test: failed to start Iggy container: %v\n", err)
+		os.Exit(1)
+	}
```
chengxilo, line 49: If `err != nil`, it won't terminate the container, as it `os.Exit(1)` directly, better try to terminate it or something.

chengxilo (changes requested):
Regarding the signed_in event, I'm wondering if we should revert to your original code. As I mentioned in the previous review, it might be better to keep publishing the signed_in event every time we sign in during a redirect, while also keeping the new redirected event. It prevents us from diverging too much from the Rust SDK and  faithfully tells the subscriber they sign-in or not.

FYI:https://github.com/apache/iggy/pull/3621#discussion_r3545522300

But what we have now is also fine, since it's my initial proposal hahaha. Both are fine for me, it really depends on what you need when implement the Consumer API. : ) It's up to you.

foreign/go/client/tcp/tcp_core.go
chengxilo:
```go
        var err error
        if c.conn != nil {
                err = c.conn.Close()
        }
```
This part on hold, I need to consider about it.

foreign/go/client/tcp/tcp_core.go
chengxilo:
```go
        var err error
        if c.conn != nil {
                err = c.conn.Close()
        }
```
This part on hold, I need to consider about it. (Same as above)

foreign/go/client/tcp/tcp_core.go
chengxilo: Will be addressed in #3670, ignore it.

foreign/go/client/tcp/tcp_core.go
chengxilo: Will be addressed in #3670, ignore it.

foreign/go/client/tcp/tcp_core.go
chengxilo: Will be addressed in #3670, ignore it.

chengxilo:
when resolve conflict, replace 
```go
if c.conn != nil {
		if err := c.conn.Close(); err != nil {
			return err
		}
	}
```

with 
```go
err := c.closeConnLocked()
```

I hope this would help.

PR #3623, opened 2026-07-07
feat(python): expose remaining Topic/TopicDetails fields and partitions

### LINKED ISSUE #3577: Expose remaining Topic and TopicDetails fields in the Python SDK
## Title

Expose remaining Topic and TopicDetails fields in the Python SDK

## Description

The Python SDK's `Topic` and `TopicDetails` classes expose only a subset of the
fields available on their Rust counterparts (`iggy::prelude::Topic` /
`TopicDetails`), so a topic's full configuration can't be read from Python and
those fields can't be asserted in tests.

Already exposed:
- `TopicDetails`: `id`, `name`, `messages_count`, `partitions_count`,
  `compression_algorithm`, `replication_factor`
- `Topic`: `id`, `name`, `messages_count`, `partitions_count`

This issue tracks adding getters (with docstrings) for the remaining fields.

**`TopicDetails`:**
- `created_at` (`IggyTimestamp`)
- `size` (`IggyByteSize`)
- `message_expiry` (`IggyExpiry`)
- `max_topic_size` (`MaxTopicSize`)
- `partitions` (`Vec<Partition>`)

**`Topic`:**
- `created_at` (`IggyTimestamp`)
- `size` (`IggyByteSize`)
- `message_expiry` (`IggyExpiry`)
- `compression_algorithm` (`CompressionAlgorithm`)
- `max_topic_size` (`MaxTopicSize`)
- `replication_factor` (`u8`)

## Affected area / component

Python SDK (`foreign/python`)

## Proposed solution

Add `#[getter]` methods to `Topic` and `TopicDetails` in
`foreign/python/src/topic.rs`, regenerate `apache_iggy.pyi` via `stub_gen`, and
add corresponding tests. For the tri-state enum fields, settle on a faithful
Python representation (e.g. small wrapper classes) before implementing.

## Alternatives considered

No response

## Contribution

I will work on this once [3572](https://github.com/apache/iggy/pull/3572) is closed.

mattp5657:

Closes #3577

## Rationale

Continuation of #3572. That PR exposed `Topic`/`TopicDetails` basics but left several fields returned by the server unavailable in Python, so callers couldn't inspect a topic's expiry, size, max size, or its partitions without another SDK.

## What changed?

`Topic` and `TopicDetails` previously exposed only `id`, `name`, `partitions_count`, `compression_algorithm`, and `replication_factor`, leaving `created_at`, `size`, `message_expiry`, `max_topic_size`, and per-partition data unavailable in Python.

Both classes now expose `created_at`, `size`, `message_expiry` (new `IggyExpiry` enum), and `max_topic_size` (new `MaxTopicSize` enum). `TopicDetails` also gains `partitions`, a list of new `Partition` objects. `Partition` (`core/common`) now derives `Clone` to support building that list.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## Breaking Chnages

`create_topic`'s `message_expiry` and `max_topic_size` parameters both
change type on this branch, and both changes are breaking relative to
the last released version (`0.8.0`):

| Parameter        | `0.8.0` (released)         | This branch          |
|------------------|-----------------------------|-----------------------|
| `message_expiry` | `datetime.timedelta \| None` | `IggyExpiry \| None`   |
| `max_topic_size` | `builtins.int \| None`      | `MaxTopicSize \| None` |

## AI Usage

Claude was used to help generate and review this PR.


slbotbm (changes requested):
You've defined MaxTopicSize and IggyExpiry as enums here, but the sdk utilizes normal python ints and timedeltas. Let's change those functions to accept your enums instead of the python ints and structs.

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,79 @@
...
+/// The expiry of the messages in a topic.
+#[gen_stub_pyclass_complex_enum]
+#[pyclass]
+pub enum IggyExpiry {
+    /// Use the server's default message expiry.
+    ServerDefault(),
+    /// Messages expire after this duration.
+    ExpireDuration { duration: Py<PyDelta> },
+    /// Messages never expire.
+    NeverExpire(),
```
slbotbm: Please also write exactly what does server default and never expire mean. Also, range of values that expireduration accepts

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,79 @@
...
+#[gen_stub_pyclass_complex_enum]
+#[pyclass]
+pub enum MaxTopicSize {
+    /// Use the server's default max size.
+    ServerDefault(),
+    /// The topic is limited to this many bytes.
+    Custom { bytes: u64 },
+    /// The topic has no maximum size.
+    Unlimited(),
+}
```
slbotbm, line 135: Same here. Make the docs explicit. We are doing this since the user cannot trace what exactly these mean without looking at rust sdk code.

slbotbm (changes requested):
will review in more depth later when I get more time

mattp5657:
> You've defined MaxTopicSize and IggyExpiry as enums here, but the sdk utilizes normal python ints and timedeltas. Let's change those functions to accept your enums instead of the python ints and structs.

This was an oversight from me, thanks for the call out :)

mattp5657:
I made the outlined changes, my only concern is if we should continue with backwards compatibility and incorporate it into testing. But wanted to get your thoughts before doing so. :)

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,120 @@
...
+        }
+    }
+}
+
+impl From<&IggyExpiry> for RustIggyExpiry {
+    fn from(expiry: &IggyExpiry) -> Self {
+        match expiry {
+            IggyExpiry::ServerDefault() => RustIggyExpiry::ServerDefault,
+            IggyExpiry::ExpireDuration { duration } => {
+                RustIggyExpiry::ExpireDuration(py_delta_to_iggy_duration(duration))
```
slbotbm: This also accepts negative duration, which should be rejected as PyValueError

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,120 @@
...
+#[pyclass]
+pub enum MaxTopicSize {
+    /// Use the maximum topic size configured on the server, rather than an
+    /// explicit value set by the client.
+    ServerDefault(),
+    /// Cap the topic at this many bytes; as the topic approaches this size,
+    /// the server deletes the oldest sealed segments to make room for new
+    /// messages.
+    ///
+    /// `bytes` must be greater than zero and less than `u64::MAX`: those two
```
slbotbm: Python callers do not know what u64::MAX means.

slbotbm (changes requested):
Some comments

mattp5657:
Fixed both boundary-collision issues flagged in review: negative `timedelta` values for `IggyExpiry.ExpireDuration` and `0`/`u64::MAX` values for `MaxTopicSize.Custom` were silently accepted and misinterpreted on the wire (as `ServerDefault`/`Unlimited`) instead of being rejected. Both conversions now raise `PyValueError` for out-of-range input, propagated through `create_topic`/`update_topic`. Updated the affected doc comments/stubs and added a regression test for each invalid case.

foreign/python/src/consumer.rs
```diff
@@ -517,11 +519,19 @@ impl PyStubType for AutoCommitAfter {
     }
 }
 
-pub fn py_delta_to_iggy_duration(delta1: &Py<PyDelta>) -> IggyDuration {
+pub fn py_delta_to_iggy_duration(delta1: &Py<PyDelta>) -> PyResult<IggyDuration> {
     Python::attach(|py| {
         let delta = delta1.bind(py);
-        let seconds = (delta.get_days() * 60 * 60 * 24 + delta.get_seconds()) as u64;
+        let total_seconds = delta.get_days() * 60 * 60 * 24 + delta.get_seconds();
```
hubcio: `delta.get_days() * 60 * 60 * 24 + delta.get_seconds()` is all i32 here (pyo3's `get_days`/`get_seconds` both return `i32`). for `|days| >= 24856` (~68yr) the multiply overflows - release wheels wrap silently, debug builds panic. the nasty part is it defeats the negative-duration guard this PR just added right below: a large negative `timedelta` (easy to get from date arithmetic) wraps to a positive `total_seconds`, slips past the `if total_seconds < 0` check, and a positive expiry gets persisted from a negative input. one-line fix: `let total_seconds = i64::from(delta.get_days()) * 86_400 + i64::from(delta.get_seconds());` then range-check before the `as u64`.

foreign/python/src/client.rs
```diff
@@ -187,22 +190,28 @@ impl IggyClient {
         #[gen_stub(override_type(type_repr = "builtins.int | None"))] replication_factor: Option<
             u8,
         >,
-        #[gen_stub(override_type(type_repr = "datetime.timedelta | None", imports=("datetime")))]
-        message_expiry: Option<Py<PyDelta>>,
-        #[gen_stub(override_type(type_repr = "builtins.int | None"))] max_topic_size: Option<u64>,
+        #[gen_stub(override_type(type_repr = "IggyExpiry | None"))] message_expiry: Option<
```
hubcio, line 472: note this is a breaking change for `create_topic.message_expiry` - it went from `timedelta | None` to `IggyExpiry | None`, and that param shipped in the released 0.8.0, so existing callers passing a `timedelta` now hit a `TypeError`. the `max_topic_size` retype and all of `update_topic` are post-0.8.0 so those break nothing released. worth calling out the `message_expiry` break, at least in the PR descr.

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,133 @@
...
+    /// rather than an explicit value set by the client.
+    ServerDefault(),
+    /// Expire messages this long after they are appended to the topic.
+    ///
+    /// `duration` must be greater than zero: a zero-length `timedelta` is
+    /// indistinguishable on the wire from `ServerDefault` and is treated as
+    /// such by the server. A negative `timedelta` raises `ValueError` when
+    /// this value is passed to `create_topic`/`update_topic`. The upper
+    /// bound is whatever a `datetime.timedelta` can represent that also fits
+    /// in a `u64` microsecond count (about 584,942 years); in practice the
```
hubcio: this bound isn't accurate today - the timedelta conversion in `py_delta_to_iggy_duration` overflows i32 at ~68 years, so the real ceiling is far below what's documented. the figure is also a typo: u64 microseconds is ~584,542 years, not 584,942. once the overflow is widened to i64 the ~584,542y bound becomes real and this doc is correct.

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,133 @@
...
+}
+
+impl TryFrom<&MaxTopicSize> for RustMaxTopicSize {
+    type Error = PyErr;
+
+    fn try_from(max_size: &MaxTopicSize) -> PyResult<Self> {
+        Ok(match max_size {
+            MaxTopicSize::ServerDefault() => RustMaxTopicSize::ServerDefault,
+            MaxTopicSize::Custom { bytes } => {
+                if *bytes == 0 || *bytes == u64::MAX {
```
hubcio, line 156: small consistency thing: `MaxTopicSize.Custom(0)` (and `u64::MAX`) reject with `ValueError` here, but the sibling sentinel `IggyExpiry.ExpireDuration(timedelta(0))` silently coerces to server-default instead of rejecting. both are documented so it's not a bug, just worth picking one policy across the two - either both reject the reserved value or both coerce it.

foreign/python/src/topic.rs
```diff
@@ -100,15 +260,102 @@ impl TopicDetails {
...
+
     /// Replication factor for the topic.
     #[getter]
     pub fn replication_factor(&self) -> u8 {
         self.inner.replication_factor
     }
+
+    /// The collection of partitions in the topic.
+    #[getter]
+    pub fn partitions(&self) -> Vec<Partition> {
```
hubcio, line 326: this getter clones the whole partition vec and builds fresh python objects on every access, so `topic.partitions` looks like a cheap attribute but re-materializes each time (and the returned objects don't have stable identity). it's a cold metadata path so not a perf concern, but worth a doc note or exposing it as a method so callers don't read it repeatedly in a loop.

foreign/python/src/client.rs
```diff
@@ -169,7 +170,9 @@ impl IggyClient {
     }
 
     /// Creates a new topic with the given parameters.
```
hubcio, line 438: minor: `create_topic` documents its return and raises as free prose while `update_topic` below uses the structured Args/Returns/Raises form. worth aligning the two.

foreign/python/src/client.rs
```diff
@@ -187,22 +190,28 @@ impl IggyClient {
...
             Some(algo) => CompressionAlgorithm::from_str(&algo)
                 .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?,
             None => CompressionAlgorithm::default(),
         };
 
-        let expiry = match message_expiry {
-            Some(delta) => IggyExpiry::ExpireDuration(py_delta_to_iggy_duration(&delta)),
-            None => IggyExpiry::ServerDefault,
-        };
+        let expiry = message_expiry
```
hubcio: this expiry + max_size resolution block is duplicated byte-for-byte in `update_topic`. could pull it into a small helper (or two) to dedup, no behavior change.

foreign/python/tests/test_topic.py
```diff
@@ -895,7 +965,7 @@ async def test_update_topic_invalid_replication_factor(
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize(
-        ("max_topic_size", "expected_exception"),
+        ("max_topic_size_bytes", "expected_exception"),
         [
             (-1, OverflowError),
             (2e64, TypeError),
```
hubcio, line 1018: the create_topic version of this test also covers `(0, ValueError)` and `(2**64-1, ValueError)` for the reserved sentinels, but they're missing here even though update goes through the same `try_from`. there's also no update_topic test for a negative `ExpireDuration` like create has. worth mirroring the create cases.

hubcio (changes requested):

foreign/python/src/client.rs
mattp5657: Sorry, I left a comment but hadn't updated the PR description yet. I know it's sometimes easier to follow a PR when the description reflects the original push and comments show how the discussion evolved, but I should have been clearer here, apologies.

I see three options:

1. Keep the current change. I think it's the most robust approach, but I understand it's a breaking change for `message_expiry`.
2. Revert to the original types used to init the server settings (backwards compatible).
3. Accept both, support the old type alongside the new enum.

My main concern is whether we should preserve backwards compatibility here and add test coverage for it, but I wanted to get your thoughts before going down that path.

mattp5657:
I have addressed all the comments and added the following to the PR:

`create_topic`'s `message_expiry` and `max_topic_size` parameters both
change type on this branch, and both changes are breaking relative to
the last released version (`0.8.0`):

| Parameter        | `0.8.0` (released)         | This branch          |
|------------------|-----------------------------|-----------------------|
| `message_expiry` | `datetime.timedelta \| None` | `IggyExpiry \| None`   |
| `max_topic_size` | `builtins.int \| None`      | `MaxTopicSize \| None` |

foreign/python/src/topic.rs
```diff
@@ -15,9 +15,149 @@
...
+                    ));
+                }
+                RustIggyExpiry::ExpireDuration(iggy_duration)
+            }
+            IggyExpiry::NeverExpire() => RustIggyExpiry::NeverExpire,
+        })
+    }
+}
+
+fn iggy_duration_to_py_delta(duration: Duration) -> Py<PyDelta> {
```
ryerraguntla: topic.rs:94-106 — iggy_duration_to_py_delta uses .expect on PyDelta::new; get_topic/message_expiry read path can abort on corrupt wire duration. **Fix**: map PyErr, raise ValueError.

mattp5657:
I think this test might be flaky. I'll create an issue to look into it.

`server::message_cleanup::message_cleanup::expiry_respects_consumer_offset_expects`

slbotbm (dismissed):

foreign/python/src/client.rs
slbotbm: before reaching 1.0.0, I do not think we guarantee backwards compatibility. I think that writing the message about the breaking changes is enough. You could also additionally update the version of the sdk from `0.8.1.dev3` to `0.8.1.dev4` in pyproject.toml. 

Other than that, this looks mergeable.

numinnex (dismissed):

spetz (dismissed):

foreign/python/src/client.rs
mattp5657: Would you like me to update before merge. Also thanks for all your help here!

slbotbm: @mattp5657 Sure, please update the version before it gets merged.

mattp5657:
Thanks for everyones help with this one! I believe the update to the pyproject and the conflict are resolved now.

slbotbm (approved):

hubcio (approved):

PR #3625, opened 2026-07-07
fix(server-ng): fix restart/reconnect hardening and cg integration tests

numinnex:
This PR fixes remaining reconnect/restart integration tests, additionally removes the `cg_vsr` namespace and merges it together with `server::cg`. 

Here is a list of tests that are still gated that are awaiting state transfer:

- server::purge_delete consumer restart_on cells (9) + data_integrity::verify_after_server_restart (4): ops committed by the surviving quorum during the restart leave the pipeline before the rejoining node can get them; the replica hits the journal gap and correctly suicides ("replica is divergent").
- server::specific restart family (9): plane-primary split — metadata recovery resumes primaryship from its WAL while partition groups rejoin as backups and peers elect a new leader; clients follow the roster (metadata) leader, so partition writes land on a backup and are dropped (SDK timeout → Disconnected).

Temporary patches that state transfer should replace: fast_forward_recovering_commit_floor (unsound for non-empty rejoin windows — can silently skip window ops), metadata's resume-as-primary-from-WAL, and init_recovering's passive wait. Proper fix: VSR recovery handshake (Recovery/RecoveryResponse — learn current view from peers, fixes the split) + log catch-up from an active peer (fills the window) + bulk segment sync for long downtime.


core/server-ng/src/dispatch.rs
```diff
@@ -2053,8 +2111,34 @@ pub(crate) async fn resolve_delete_segments_truncate(
...
+            up_to_offset: None,
+            resident_messages,
+        }) if resident_messages > 0 => {
+            debug!(
+                client_id,
+                namespace_raw,
+                resident_messages,
+                "delete_segments: partition not converged; transient"
+            );
+            return Err(IggyError::TransientNotCommitted);
```
hubcio: this fires a false transient on a converged single-node primary under the default config. `resident_journal_messages` counts committed-but-unflushed messages, and with the default `messages_required_to_save = 1024` / 1 MiB there's no periodic sub-threshold flush, so a low-volume partition sits resident with nothing sealed indefinitely. `up_to_offset = None` + `resident > 0` then returns `TransientNotCommitted`, the client replays for the whole read timeout and errors, where the legacy server acked a clean no-op. the purge_delete tests miss this because they override `messages_required_to_save = "1"` (eager flush, resident ~0). gate the transient on `!is_caught_up_primary` (a real backup behind the frontier), not raw `resident > 0` - a converged primary with sub-threshold committed data should no-op.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -701,122 +712,70 @@ impl TcpClient {
...
+                        transient_deadline,
+                        overall_deadline,
+                    )
+                    .await;
+                match result {
+                    Err(IggyError::TransientNotCommitted)
+                        if tokio::time::Instant::now() < overall_deadline
+                            && !is_login_register_code(code) =>
+                    {
+                        // The frame was complete, so the stream is in sync and
```
hubcio: this comment isn't true for the view-change-cancel and in-flight cases. the server emits `TransientNotCommitted` for ops that may already have committed (see `core/metadata/src/impls/metadata.rs:1429` "the op may or may not have committed"), and at-most-once holds only because a same-id resend hits the client-table cache. the failover branch just below calls `handle_leader_redirection` -> `disconnect` -> `reset_vsr_session`, which mints a fresh random client_id, then re-issues under a new (client_id, request_id) - so the cache misses and a committed metadata write re-applies. only the "same id on this session" half is correct; the "fresh session after a failover" half is exactly the hazard your own `core/sdk/src/vsr.rs:45` TODO documents. single-node is immune (no failover). at minimum fix the comment; better, carry the client_id across the redirect or skip failover for dedup-reliant ops.

core/consensus/src/impls.rs
```diff
@@ -546,6 +546,19 @@ pub enum Status {
 
 /// Actions to be taken by the caller after processing a VSR event.
```
hubcio: the doc comment and `#[derive(Debug, Clone)]` here now bind to the new `CommitOutcome` instead of `VsrAction`, which sits bare a few lines down. `VsrAction` is pub and re-exported into shard/partitions and just lost its `Debug` + `Clone` - it only compiles because nothing currently formats or clones one. the doc also mislabels `CommitOutcome` (a heartbeat outcome, not an action). give each enum its own doc + derive. minor: `Accepted` and `Ignored` are never distinguished at either call site (both `=> {}`), so they could collapse into one variant unless you plan to log the distinction.

core/server-ng/src/dispatch.rs
```diff
@@ -1980,31 +1987,82 @@ async fn handle_delete_segments_request(
...
+            {
+                warn!(
+                    transport_client_id,
+                    error = %error,
+                    "delete_segments: failed to send transient rejection"
+                );
+            }
+            return;
+        }
+        Err(_) => None,
```
hubcio, line 2033: the `Err(_) => None` arm acks an unresolvable target (bad body / unknown stream/topic/partition) with an empty reply and no commit, which gaps the VSR request sequence - the next metadata op fails the `request == committed + 1` preflight, gets dropped, and the client replay panics with `register_request_id already called`. it also returns success where the legacy server returns an error. same admission-gate family as the false-transient in `resolve_delete_segments_truncate`; you already flag it in the skipped error-case block in `purge_delete_scenario.rs`. reject unresolvable targets via the result-framed deny channel instead of an empty ack.

core/consensus/src/impls.rs
```diff
@@ -1827,7 +1910,37 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+        if self.commit_max.get() > old_commit_max {
+            CommitOutcome::Advanced
+        } else {
+            CommitOutcome::Accepted
+        }
+    }
+
+    /// See the `recovering` field: align `commit_min` with the learned
+    /// `commit_max` exactly once, then leave recovery mode.
+    fn fast_forward_recovering_commit_floor(&self) {
```
hubcio, line 1966: `fast_forward_recovering_commit_floor` sets `commit_min = commit_max` without checking that local segments actually reach it. for a recovering partition replica (in-memory journal lost on restart) that later gets promoted primary, this can serve a message-offset gap silently - and it won't self-detect: the "replica is divergent" panic only fires when the drain loop walks an op, but this jumps `commit_min` directly so the gap ops are never walked. so the PR description's "replica correctly suicides" doesn't hold for the fast-forwarded range - it silently serves the gap. understood this is interim until state transfer lands (multi-node restart tests are gated off); flagging so the silent-serve behavior is on record.

core/consensus/src/impls.rs
```diff
@@ -760,6 +785,21 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
     }
 
+    /// Initialize a replica whose consensus state did NOT survive restart
+    /// (e.g. a partition group: its journal is in-memory and its segments
+    /// carry no op numbers). Such a replica must not assume primaryship: it
+    /// would heartbeat `commit_min = 0`, dragging the group behind backups
+    /// that kept their journals, and it has no ops to re-pipeline. Join as a
+    /// backup instead; the peers' heartbeat timeout elects a primary that
+    /// still holds the log, and its `StartView` brings this replica forward.
+    pub fn init_recovering(&self) {
```
hubcio, line 828: `init_recovering` arms only `NormalHeartbeat`, and `handle_normal_heartbeat_timeout` has no recovering gate. a recovering replica that is primary-index-0 at view 0 hits the `is_primary` early-return in the timeout, so it never sends a heartbeat - which means the `RespondStartView` net (built for stale-view primaries) never gets a heartbeat to answer, and the replica silently drops out of quorum with no operator signal, self-healing only on the next unrelated view change. a recovering non-0 backup instead climbs StartViewChange and forces a needless cluster view change. multi-node and interim, but worth arming a heartbeat (or gating the timeout on `recovering`) so a recovering primary-index-0 replica doesn't sit inert.

core/partitions/src/iggy_partition.rs
```diff
@@ -1768,126 +1795,160 @@ where
...
+            // that the batch + index are durable. Inserting in the build loop (before
+            // persist) re-inserts a duplicate on a persist-failure retry, which
+            // re-reads the same prefix. The active segment has not rotated yet, so
+            // this targets the segment that received the batches.
+            if let Some(index) = flush_index {
+                self.log.ensure_indexes();
+                let indexes = self.log.active_indexes_mut().expect("indexes must exist");
+                indexes.insert(index.offset, index.timestamp, index.position);
+            }
+            self.evict_committed_prefix(chunk_len).await;
```
hubcio: `evict_committed_prefix` is now called once per segment-chunk inside the persist loop, and `evict_prefix` drains the whole resident journal and rebuilds the retained tail (re-appending each retained entry into the index maps) instead of front-truncating. so a flush spanning K segments re-walks the retained entries K times, all under the partition write lock, which stalls that partition's producers - worst on backup burst-commit catch-up and large-threshold flushes. a front-truncating `evict_prefix` that drops the first `count` entries would make the K calls sum back to linear.

core/partitions/src/iggy_partition.rs
```diff
@@ -1768,126 +1795,160 @@ where
...
 
-                if flush_index.is_none() {
-                    // Record only; the in-mem cache insert is deferred until the
-                    // batch + index are durable (see post-persist below).
-                    flush_index = Some(crate::iggy_index::IggyIndex::new(
+            let (frozen_batches, index_bytes, flush_index, batch_count, committed_info, chunk_len) = {
+                let segment = self.log.active_segment();
+                let mut file_position = segment.size.as_bytes_u64();
+                let mut flush_index = None;
+                let mut frozen = Vec::new();
```
hubcio: `frozen` is `Vec::new()` here - the pre-rewrite code sized it with `with_capacity`, so it now reallocs while pushing. also the chunk iteration hand-rolls `next_entry` / `.take()` where `committed_entries.into_iter().peekable()` would do the same with one buffered slot and less shuffling. right-size the vec to the remaining entries while you're in here.

core/metadata/src/impls/recovery.rs
```diff
@@ -138,12 +158,29 @@ where
...
     let headers_to_replay = journal.iter_headers_from(replay_from);
 
+    // Committed watermark: the `commit` field of each journaled prepare is
+    // the primary's commit point when it was sent (monotone in op order), so
+    // the last entry carries the highest commit this WAL can prove. Ops above
+    // it were prepared but not provably committed; applying them here would
+    // fabricate commit knowledge (a crashed suffix may have been discarded by
+    // a view change) and would hide the suffix from re-replication.
+    let snapshot_floor = snapshot.as_ref().map_or(0, IggySnapshot::sequence_number);
+    let commit_watermark = headers_to_replay
```
hubcio, line 170: `commit_watermark` from `.last().commit` relies on the commit field being monotone in op order. it holds today (prepares stamp `commit = commit_max`, which only rises), but a `.max()` fold over the replayed headers is the same cost and defends against a future non-monotone change without silently under-applying the committed prefix.

hubcio (changes requested):

spetz (approved):

mmodzelewski (approved):

hubcio (approved):

Issue #3582, opened 2026-06-29
bug(connectors): PostgreSQL source builtin CDC mode is non-functional


mmodzelewski:
## Summary

The PostgreSQL source connector's builtin CDC mode (`mode = "cdc"`,
`cdc_backend = "builtin"`) does not work. It cannot produce a single
change message against a real database. Polling mode is unaffected.

`core/connectors/sources/postgres_source/src/lib.rs`

## Details

1. **Invalid publication SQL (setup fails).** `setup_cdc` runs
   `CREATE PUBLICATION IF NOT EXISTS ...`. PostgreSQL has no
   `IF NOT EXISTS` clause for `CREATE PUBLICATION`, so the statement is
   a syntax error and CDC setup fails outright.

2. **Plugin/reader mismatch (hard error on every poll).** `setup_cdc`
   creates the replication slot with the `pgoutput` plugin (binary
   output), but `poll_cdc_builtin` reads it via the textual
   `pg_logical_slot_get_changes`. PostgreSQL rejects this:
   `output plugin "pgoutput" produces binary output, but function ...
   expects textual data`. Every poll fails.

3. **Parser targets a format no plugin emits.**
   `parse_logical_replication_message` matches lines starting with
   `INSERT:`/`UPDATE:`/`DELETE:`. Real `test_decoding` output is
   `table public.users: INSERT: id[integer]:1 ...`, so the prefix
   never matches and all changes are dropped. The synthetic format in
   the unit tests corresponds to no real Postgres output.

4. **Column data lost even on the synthetic format.** For input like
   `id[1]`, `parse_record_data` requires a `:` that isn't present, so
   the `data` object comes out empty. Tests only assert `table_name`
   and `operation_type`, so they pass while extracting nothing.

5. **Data-loss risk if it did run.** `pg_logical_slot_get_changes`
   consumes (advances) the slot at poll time, before the runtime
   confirms the Iggy send and persists state. A crash in between loses
   those changes - violates the at-least-once contract. The selected
   `lsn`/`xid` columns are never used; CDC tracks no LSN in state.

## Smaller issues

- `mode = "cdc"` with `enable_wal_cdc` unset silently skips slot
  creation, then errors on every poll. Two flags for one feature.
- `slot_name` / `publication_name` interpolated raw into SQL string
  literals (rest of the file uses `quote_identifier`).
- CDC messages use a random UUID id, not the row PK - no dedup.
[...]

PR #3640, opened 2026-07-09
fix(connectors): fix postgres source CDC producing no messages

mattp5657:

Relates to #3582

## Rationale

Builtin CDC mode currently produces zero messages against a real
database - wrong plugin/read-path pairing, and a parser matching a
format no real Postgres output plugin emits. This PR is scoped to
getting CDC actually working and provably correct, nothing more.

I have reduce the scope of this specific PR for reviewability, in the future 
would look to create issues and work on these issues:
- `pgoutput` as the long-term plugin - real binary decoder + streaming
  connection is a much bigger lift; `test_decoding` unblocks correctness
  today. Target once `cdc_backend = "pg_replicate"` gets picked up.
- At-least-once / crash safety (slot advances before send is confirmed,
  no LSN tracked) - needs its own review since it touches the
  state-persistence contract. Follow-up: #TODO.
- Cosmetic items (UUID vs PK as message id, processing time vs WAL
  commit time, `old_data` unpopulated) - mechanical, separate PR.

## What changed?

Before: wrong slot plugin, a parser matching a format no real Postgres
output emits, and invalid setup SQL - CDC produced zero messages, and
several related bugs failed silently too. After: parser rewritten
against real captured output, setup/poll SQL fixed and bounded, and
those failure modes now work correctly or fail loudly. Covered by
corpus-driven unit tests plus integration tests against a live
`wal_level=logical` container.

- Parser rewritten against real `test_decoding` output
- Slot/setup SQL fixed, including the upgrade path and table filtering
- Batch handling hardened (bad rows, unbounded backlog)
- Dead config fields removed

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

Claude was used to help generate and review this PR.


hubcio:
i think this is somewhat related to https://github.com/apache/iggy/discussions/3313

hubcio:
@mmodzelewski could you please check this one?

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -1497,42 +1491,107 @@ fn to_snake_case(input: &str) -> String {
...
-            let cleaned_value = if value_str.starts_with('\'') && value_str.ends_with('\'') {
-                &value_str[1..value_str.len() - 1]
-            } else {
-                value_str
-            };
+        let Some(bracket_offset) = data[pos..].find('[') else {
+            break;
+        };
+        let name_end = pos + bracket_offset;
+        let column_name = &data[pos..name_end];
```
mmodzelewski: Quoted column names keep their literal quotes as JSON keys. `test_decoding` applies `quote_identifier` to column names, so a mixed-case or keyword column (`createdAt`, `user`) arrives as `"createdAt"[timestamp with time zone]:'...'` and this parser emits the JSON key `\"createdAt\"` with the double-quote characters included. Table names already go through `unquote_pg_identifier`; column names should get the same treatment, otherwise downstream consumers looking up the plain column name miss. Worth a unit test with a quoted column name fixture.

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -1497,42 +1491,107 @@ fn to_snake_case(input: &str) -> String {
...
-            };
+        let (value, next_pos) = parse_column_value(data, type_end + 2);
+        result.insert(column_name.to_string(), value);
+        pos = next_pos;
+    }
+
+    result
+}
+
+fn parse_column_value(data: &str, start: usize) -> (serde_json::Value, usize) {
```
mmodzelewski, line 1600: The `unchanged-toast-datum` sentinel leaks into records as a real value. When an UPDATE does not touch a TOASTed column (large text/jsonb), `test_decoding` emits the bare token `unchanged-toast-datum`, and the parser stores `data["col"] = "unchanged-toast-datum"`, which is indistinguishable from a legitimate string. Consumers will silently ingest the sentinel as the column value. Suggest handling it explicitly in `parse_bare_scalar` (map to null, or omit the key) and adding a fixture for it.

core/integration/tests/connectors/postgres/postgres_source_cdc.rs
```diff
@@ -0,0 +1,399 @@
...
+    let consumer_id: Identifier = "cdc_restart_consumer".try_into().unwrap();
+
+    let api_url = harness
+        .connectors_runtime()
+        .expect("connector runtime should be available")
+        .http_url();
+    let http = Client::new();
+    wait_for_source_status(&http, &api_url, ConnectorStatus::Running).await;
+
+    sqlx::query("SELECT pg_drop_replication_slot($1)")
```
mmodzelewski: Flake risk: this first `pg_drop_replication_slot` races the still-running connector, which calls `pg_logical_slot_get_changes` every 50ms and briefly holds the slot on each call. If the drop lands inside that window, Postgres raises `ERROR 55006: replication slot "iggy_slot" is active for PID ...` and the `expect` panics. Rough odds are a few percent per run, which is enough to flake CI. Wrapping the drop in a short retry loop fixes it. The second drop below is fine since the failed restart leaves no poller running.

core/connectors/sources/postgres_source/README.md
```diff
@@ -41,9 +41,7 @@ primary_key_column = "id"
 # Custom query (optional)
 custom_query = "SELECT * FROM $table WHERE id > $offset ORDER BY id LIMIT $limit"
 
-# CDC options (optional)
-enable_wal_cdc = false
-publication_name = "iggy_publication"
+# CDC options (only used when mode = "cdc")
 replication_slot = "iggy_slot"
```
mmodzelewski, line 45: Two operational hazards of the new slot handling deserve README coverage:

1. Slot sharing: `setup_cdc` accepts any pre-existing `test_decoding` slot, so two CDC connectors against the same database with the default `replication_slot = "iggy_slot"` silently share one slot. `pg_logical_slot_get_changes` consumes on read, so each connector steals a subset of the other's changes. The README should state that every CDC connector needs a unique slot name.
2. WAL retention: a slot orphaned after a connector is decommissioned retains WAL indefinitely (default `max_slot_wal_keep_size = -1`) until the disk fills. The README should document dropping the slot (`SELECT pg_drop_replication_slot('iggy_slot')`) as part of decommissioning.

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -303,10 +301,6 @@ impl PostgresSource {
     }
 
     async fn setup_cdc(&self) -> Result<(), Error> {
```
mmodzelewski, line 306: Not on this line, but related to CDC startup: `open()` validates `mode` but not `cdc_backend`. A typo like `cdc_backend = "built-in"` passes startup, then every `poll_cdc` call returns an error that the SDK loop logs and swallows, leaving a Running connector that produces nothing. That is the same silent-death failure mode this PR fixes for the slot mismatch. Validating the backend in the `"cdc"` arm of `open()`, where the slot mismatch already fails loudly, would close it.

core/connectors/sources/postgres_source/src/lib.rs
```diff
@@ -728,99 +762,59 @@ impl PostgresSource {
...
-            return Some(DatabaseRecord {
-                table_name,
-                operation_type: "DELETE".to_string(),
-                timestamp: Utc::now(),
-                data: serde_json::Value::Object(parsed_data),
-                old_data: None,
-            });
-        }
-        None
+        let rest = rest
```
mmodzelewski: The old-key columns are parsed past and discarded here, so `DatabaseRecord.old_data` is always `None` even though the struct has the field and the data is sitting in the row. For a PK-changing UPDATE (and for DELETE under `REPLICA IDENTITY FULL`) consumers only see the new tuple and cannot learn which row it replaced, which breaks e.g. keyed upsert/delete downstream. Issue #3582 lists unpopulated `old_data` as well. Suggest running `parse_record_columns` over the old-key section and populating `old_data` instead of dropping it.

mmodzelewski (changes requested):
Thanks, @mattp5657. The change looks solid, I've posted a few comments where the code could be improved.

core/connectors/sources/postgres_source/src/lib.rs
mattp5657: I resolved this but also handled two other edge cases.

1. `capture_operations` - an entry outside `INSERT`/`UPDATE`/`DELETE` (e.g. a typo like `INSRT`) previously passed startup silently and would just never match, quietly dropping that operation type forever.
2. `payload_format` - an unsupported value (e.g. `btea` instead of `bytea`) would similarly pass startup and only surface as a failure deep in row processing on the first poll, not at startup.

Both now validate in `open()` alongside `cdc_backend`, before `setup_cdc()` runs, so any of the three fails loudly at startup instead of leaving a `Running` connector that silently drops or mishandles data.

Then I rewrote the integration test to make sure it covered this as well.

mattp5657:
These should be fixed, let me know if you have any other feedback :)

mmodzelewski (approved):
Thanks, @mattp5657, approved

hubcio (approved):

PR #3647, opened 2026-07-10
feat: respect CPU and cgroup limits in server-ng and connectors stats

mmodzelewski:
The server-ng and connectors runtime stats endpoints reported
host-wide sysinfo numbers: a cpuset-confined or memory-capped
process showed its neighbors' CPU load and a memory total it can
never allocate, leaking host sizing on multi-tenant machines. The
legacy server already scoped these (#3615) but kept the logic
inline.

Extract that logic into a shared system_stats crate: SystemProbe
scopes total CPU usage to the allowed core set (snapshotted at
startup, before shard threads pin themselves) and memory totals
to the effective cgroup cap, falling back to host-wide values for
unconfined processes. All three binaries now probe through it.

On confined hosts the reported total_cpu_usage, total_memory and
available_memory now reflect the confinement, so dashboards and
alerts keyed to the old host-wide values will see lower numbers.

The connectors runtime also keeps its sysinfo System alive across
captures, so CPU usage reflects real deltas instead of the zero
first sample of a freshly created System on every request.


core/server-ng/src/responses.rs
```diff
@@ -738,7 +741,7 @@ fn probe_system_stats() -> SystemStats {
         let mut stats = SystemStats {
             process_id,
             cpu_usage: 0.0,
-            total_cpu_usage: sys.global_cpu_usage(),
+            total_cpu_usage: scoped_total_cpu_usage(sys),
```
hubcio: this regresses `total_cpu_usage` on the default config. `scoped_total_cpu_usage` calls `allowed_cpus()` = `sched_getaffinity(0)`, which returns the calling *thread's* mask - and this runs on a shard thread that pinned itself to a single core (`pin_cores` defaults to true). so `allowed.len() == 1`, the `allowed.len() >= cpus.len()` guard never falls back to global even on an unconfined host, and the field reports one core's usage instead of the process-wide average (pre-PR this was `sys.global_cpu_usage()`). since any shard can serve `GetStats`, the value also jumps between unrelated cores call to call. connectors is unaffected - tokio workers are unpinned, so the thread mask equals the process cpuset there, which is why it looks right in that runtime.

fix: capture `allowed_cpus()` once at boot on the main thread *before* any `bind_cpu`, stash it in a process-global, and have `scoped_total_cpu_usage` read the snapshot. has to be an explicit boot capture, not a lazy `OnceLock` initialized from the stats path (that would run on the pinned thread and re-capture the single core). reading `cpuset.cpus.effective` instead would miss plain `taskset` / `sched_setaffinity` confinement.

core/server/src/shard/system/stats.rs
```diff
@@ -46,8 +45,7 @@ impl IggyShard {
             sys.refresh_memory();
             sys.refresh_processes(ProcessesToUpdate::Some(&[Pid::from_u32(process_id)]), true);
 
-            let total_cpu_usage =
-                allowed_cores_cpu_usage(sys).unwrap_or_else(|| sys.global_cpu_usage());
+            let total_cpu_usage = scoped_total_cpu_usage(sys);
```
hubcio: same single-core issue as server-ng applies here - `get_stats` runs on shard-0, which is pinned under the default `pin_cores = true`, so `allowed_cpus()` sees one core. pre-existing since #3615 (this PR only relocates the fn), and the boot-time snapshot fix repairs both servers at once.

core/system_stats/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+use cgroup_memory::cgroup_available_memory;
+
+/// Memory totals scoped to the process's effective cgroup cap.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CgroupMemory {
+    pub total: u64,
+    pub available: u64,
+}
+
+/// Average CPU usage over the cores this process may run on, or the
```
hubcio, line 128: "cores this process may run on" - `allowed_cpus()` actually reads the calling thread's affinity (`sched_getaffinity` with pid 0). those differ exactly on pinned shard threads, which is the root of the server-ng regression flagged in `responses.rs`. once the boot-time snapshot lands this wording becomes true; until then it documents a value the servers never produce. also worth a positive unit test for the averaging path - all three current tests exercise only the fallback branches, a stubbed allowed-set test would have caught this.

core/system_stats/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/// Memory totals scoped to the process's memory cgroup. `None` when no
+/// ancestor caps memory below the host total; callers keep the host
+/// numbers then.
+///
+/// `available` adds reclaimable file cache back rather than taking the
+/// kernel's `limit - current`, which trends toward zero on a cache-heavy
+/// workload long before real OOM pressure.
+pub fn cgroup_scoped_memory(sys: &System, process: &Process) -> Option<CgroupMemory> {
```
hubcio: signature takes `&Process` which reads as per-process, but the `available` refinement goes through `cgroup_available_memory()`, which hardcodes `/proc/self/cgroup`. fine today since all callers pass their own pid - worth a doc line stating the process must be the calling one, before some future caller passes a foreign pid and silently gets self's reclaim numbers.

core/system_stats/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+    #[test]
+    fn given_unrefreshed_system_when_probing_scoped_cpu_should_fall_back_to_global() {
+        let sys = System::new();
+
+        assert_eq!(scoped_total_cpu_usage(&sys), sys.global_cpu_usage());
+    }
+
+    #[test]
+    fn given_own_process_when_probing_cgroup_memory_should_stay_within_host_total() {
```
hubcio: on an unconfined host `cgroup_scoped_memory` returns `None`, the `if let` body is skipped and this passes without asserting anything - which is what typical CI will do. the tempdir tests in `cgroup_memory.rs` carry the real coverage; either assert the `None` branch explicitly here or drop the test.

core/connectors/runtime/src/stats.rs
```diff
@@ -128,3 +115,45 @@ pub async fn get_runtime_stats(context: &Arc<RuntimeContext>) -> ConnectorRuntim
...
+}
+
+/// CPU and memory scoped to what this process may actually use. On a
+/// shared host the runtime runs cpuset- or cgroup-confined; the host-wide
+/// `sysinfo` numbers would report the neighbors' CPU load and a memory
+/// total the runtime can never allocate, and they leak host sizing to
+/// whoever can read `/stats`.
+fn probe_system(pid: u32) -> SystemProbe {
+    let mut system = SYSINFO
+        .get_or_init(|| Mutex::new(System::new_all()))
```
hubcio: `System::new_all()` scans every process on the host and, since `refresh_processes(Some(&[pid]))` only ever updates/prunes the listed pid, the full startup process table stays in the static for the life of the runtime. only cpus, memory and the own pid are ever read. `System::new()` is enough - `refresh_cpu_all` populates the cpu list from empty and `refresh_processes(Some)` inserts the own process (the `system_stats` crate test relies on exactly that), and `cgroup_limits()` reads the fs on demand. same pattern pre-exists in server-ng `responses.rs` and the legacy stats path if you want to sweep those, but they're out of scope here.

core/connectors/runtime/src/stats.rs
```diff
@@ -128,3 +115,45 @@ pub async fn get_runtime_stats(context: &Arc<RuntimeContext>) -> ConnectorRuntim
...
+        .lock()
+        .unwrap_or_else(PoisonError::into_inner);
+    system.refresh_cpu_all();
+    system.refresh_memory();
+    system.refresh_processes(ProcessesToUpdate::Some(&[Pid::from_u32(pid)]), true);
```
hubcio: `Pid::from_u32(pid)` built here and again at line 149 - bind it once.

hubcio (changes requested):
one finding lands outside the diff so it goes here: `ConnectorRuntimeStats` in `core/connectors/sdk/src/api.rs` now carries cpuset/cgroup-scoped `total_cpu_usage` / `total_memory` / `available_memory` but the fields have no rustdoc - legacy `Stats` got per-field scoping notes in #3615, worth mirroring the memory ones here (hold the cpu wording until the affinity issue below is fixed, right now it would document a value the servers don't produce). release notes should also mention the value change for server-ng stats: total/available memory become the cgroup cap on confined hosts. wire-compatible, but dashboards and SDK consumers see different numbers.

hubcio (changes requested):
all findings from the earlier review are fixed as of fb6d2f80 - verified each one: boot-time `capture_allowed_cpus()` snapshot runs on the main thread before any shard pinning in all three binaries, and the `.get()`-not-`get_or_init` read means a missed capture falls back to host-global instead of lazily grabbing a pinned thread's mask; `System::new()` swap landed everywhere including the two pre-existing sites; positive averaging test + `None`-branch assertion added; `cgroup_scoped_memory` got the calling-process doc and went private; `Pid` bound once. `cargo test -p system_stats` green, fmt + clippy clean.

two leftovers, neither blocking:

- `ConnectorRuntimeStats` in `core/connectors/sdk/src/api.rs` still has no per-field rustdoc on `total_cpu_usage` / `total_memory` / `available_memory`, which now report cpuset/cgroup-scoped values on confined hosts. legacy `Stats` in `core/common/src/types/stats/mod.rs` got per-field scoping notes in #3615 - worth mirroring, and the cpu wording is safe to write now that the affinity snapshot landed.
- the PR description still names `scoped_total_cpu_usage` / `cgroup_scoped_memory` as the crate's api (both are private now - the surface is `SystemProbe::capture` + `capture_allowed_cpus`) and doesn't mention the affinity boot snapshot or that stats values change for dashboards on confined hosts. worth a refresh before merge, it doubles as the release note.

hubcio (approved):

spetz (approved):

krishvishal (approved):

Issue #3651, opened 2026-07-11
SDK client state conflates transport and session dimensions; Authenticated is never reached


chengxilo:
### Context

>  
>   This is AI generated, but I promise this is a real issue, currently Rust SDK doesn't
>   manage its state correctly, and I believe it will cause more problem in the long-term.
> 

  While bringing the Go SDK TCP client to reconnection parity with the Rust
  SDK (#3650), I traced how `ClientState` is managed across the Rust
  transports and found that the enum mixes two independent dimensions, with
  observable gaps in every transport. 

  ### Problem

  `ClientState` (core/common/src/types/client_state/mod.rs) flattens two
  orthogonal facts into one ladder:

  - transport: `Disconnected` / `Connecting` / `Connected`
  - session: `Authenticating` / `Authenticated`

  A client that is connected and signed in holds both facts simultaneously,
  but the enum can store only one value. In practice:

  1. **`Authenticated` is dead in all three transports.** There is no
     `set_state(ClientState::Authenticated)` call site in `tcp_client.rs`,
     `quic_client.rs`, or `websocket_client.rs`. After a successful
     connect + auto sign-in, the client reports `authenticating` forever
     (set at tcp_client.rs:562/584, never transitioned afterwards).
     Anything that observes the state (logs, diagnostics) reads a value
     that is wrong for the entire authenticated lifetime of the client.

  2. **Session operations do not own session state.** `login_user` /
     `logout_user` (clients/binary_users.rs, client_wrappers/
     binary_user_client.rs) never touch the state. The only writer of the
     session values is `connect()`'s auto-login block, so a manual
     `connect()` + `login_user()` flow stays at `Connected`, and there is
     no downgrade after `logout_user` because "connected but logged out"
     has no representable value.

  3. **Readers must enumerate value runs.** `connect()`'s early return
     matches `Connected | Authenticating | Authenticated`
     (tcp_client.rs:369-371) just to express "transport is up".

  None of this currently changes behavior, because every read site treats
  `Connected`, `Authenticating`, and `Authenticated` identically. It is
  structural debt: the state is unreliable as a source of truth, and the
  same transitions are hand-maintained in three transports that have
  already drifted apart.

### Affected area / component

Rust SDK, and other language if they use the same state design as Rust

### Proposed solution
  Split the dimensions: a transport state (`Disconnected` / `Connecting` /
  `Connected` / `Shutdown`) and a separate session state (`Authenticated` /
`Authenticating` / `Unauthenticated`)

[...]

Issue #3650, opened 2026-07-11
Go SDK should support reconnect and auto login configuration


chengxilo:
### Description

Related to #2964 

Currently, Go SDK support request timeout via `context.Context`, but it doesn't support reconnect and auto-login. The connection would be dead after context.Cancel/context.Timeout/io error. User have to manually call Connect() and Login() to reestablish their session. 

We need to implement the auto login/reconnect logic for Go SDK.

### Affected area / component

Go SDK

### Contribution

- [x] I'm willing to submit a pull request to implement this feature

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

hiSandog:
The listed failures may need different reconnect behavior. Should caller-driven `context.Canceled` explicitly suppress automatic reconnect, while timeouts and transport I/O errors trigger it? It would also help to define whether an in-flight request is failed, retried, or queued while the connection and login are restored.

chengxilo:
@hiSandog 

 > Should caller-driven context.Canceled explicitly suppress automatic reconnect, while timeouts and transport I/O errors trigger it?

Yes, exactly. In my design, when context.Canceled / context.DeadlineExceeded fires, we stop the request and invalidate the connection, but do not reconnect for that call: both context errors are caller-driven, so they take priority over the client's reconnection/auto-login settings. The reconnection/auto-login configuration only governs the TCP connection lifecycle; it never overrides a per-request timeout. The connection is repaired on the next attempt instead: the following request (or heartbeat ping) fails with a not-connected error, which is retryable and triggers the reconnect. So an idle client still heals through the heartbeat.

For comparison, the Rust SDK (which already implements reconnection) does: send request → if the error is retryable, reconnect → send the same request again:

https://github.com/apache/iggy/blob/f238b8cdc542992589e588ae39b3d2b56496e8e1/core/sdk/src/tcp/tcp_client.rs#L150-L209

After reviewing it again, I don't think this part is right: the Rust SDK replays the request unconditionally, even when it is not idempotent (e.g. send messages) and the server may have already executed it. (Thank you for your comment, without it I probably won't notice it) The possible Iggy error that would cause it include, ` IggyError::EmptyResponse`, `IggyError::Disconnected` and `IggyError::TcpError`. However, I am not sure how to fix this yet, probably a discussion would be needed with Rust developer. 

[...]

PR #3652, opened 2026-07-11
refactor(go): separate transport state and session state

chengxilo:

Relates to #3651 

## Rationale

Mixed state in a tech debt. It doesn't represent the client state correctly. Instead of map the tech debt from Rust SDK, I want to address it before #3650 

## What changed?

1. Separate transport state and session state.

2. Now `LoginUser`/`LoginPersonalAccessToken`/`Logout` will be the only functions that control the session state.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Claude Opus 4.8
2. implementation
3. I manually reviewed, it's an small refactor and doesn't change logic, doesn't bring too much side effect
4. Yes

## BREAKING CHANGE: 
1. the exported `iggcon.State` type and its `State*` constants are removed, replaced by `TransportState` and `SessionState`.
2. The zero value of `iggcon.State` now means `TransportStateDisconnected` (not connected yet) rather than `StateShutdown` (permanently dead). 
3. `StateAuthenticating` is dropped with no replacement


foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -30,41 +30,42 @@ import (
...
+	c.mtx.Unlock()
+
+	buffer, err := c.do(ctx, loginCmd)
 	if err != nil {
+		// A rejected login leaves the session state untouched, so restore the
+		// pre-login state only while it is still Authenticating: if the state
+		// moved, the connection may died mid-attempt and invalidation already
+		// recorded it unauthenticated, skip the restoration.
+		c.mtx.Lock()
+		if c.sessionState == iggcon.SessionStateAuthenticating {
```
hubcio: there's a race here when two goroutines call `LoginUser` concurrently on the same client. the mutex is dropped between setting `Authenticating` (line 49) and this restore check, and `do()` takes its own lock, so two logins can interleave: A's connection dies mid-login (`invalidateConnLocked` sets `Unauthenticated`), B starts a login and sets `Authenticating` again, A re-locks, sees `Authenticating` and restores its stale `pre` = `Authenticated`. B then fails the disconnected gate and skips its own restore. end state: `sessionState = Authenticated` while `transportState = Disconnected`, both logins returned errors.

nothing reads `sessionState` yet so this can't bite today, but #3650 will add exactly that reader (deciding whether to re-login on reconnect), so better to fix it before the machinery grows a consumer.

simplest fix: drop the `Authenticating` intermediate and the `pre`/restore entirely - set `Authenticated` on success only, leave `sessionState` untouched on failure. connection death already resets it via `invalidateConnLocked`, and a server-side reject never touched it in the first place, so behavior is identical on all paths and `TestLoginUser_RejectedReloginKeepsExistingSession` still passes. ~10 lines less and no race. if #3650 needs an observable `Authenticating`, it can be re-added then together with real login serialization (an in-progress guard held across the whole attempt) - the current best-effort restore doesn't provide that anyway.

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -30,41 +30,42 @@ import (
...
+	c.mtx.Lock()
+	pre := c.sessionState
+	c.sessionState = iggcon.SessionStateAuthenticating
+	c.mtx.Unlock()
+
+	buffer, err := c.do(ctx, loginCmd)
 	if err != nil {
+		// A rejected login leaves the session state untouched, so restore the
+		// pre-login state only while it is still Authenticating: if the state
+		// moved, the connection may died mid-attempt and invalidation already
```
hubcio: typo: "the connection may died" -> "may have died"

foreign/go/client/tcp/tcp_core.go
```diff
@@ -644,7 +652,8 @@ func (c *IggyTcpClient) shutdown() error {
 		}
 	}
 
-	c.state = iggcon.StateShutdown
+	c.transportState = iggcon.TransportStateShutdown
+	c.sessionState = iggcon.SessionStateUnauthenticated
```
hubcio, line 656: pre-existing ordering, but this line extends it to `sessionState`: if `conn.Close()` errors, the early return above skips both state flips, leaving the client stuck reporting connected (and now authenticated too). `disconnect()` does it in the safer order - state first, then close. worth mirroring that here, or just not early-returning before the flips.

foreign/go/contracts/state.go
```diff
@@ -17,30 +17,45 @@
...
+type TransportState uint8
 
 const (
-	StateShutdown State = iota
-	StateDisconnected
-	StateConnecting
-	StateConnected
-	StateAuthenticating
-	StateAuthenticated
+	TransportStateDisconnected TransportState = iota
```
hubcio, line 23: removing the exported `State` type and consts is a compile-time break for anyone importing `iggcon` directly - deserves a breaking-change note and a minor (not patch) bump for the next tag. also the zero value changed meaning: `StateShutdown` was 0, now `TransportStateDisconnected` is 0, so a zero-value client is connectable instead of permanently shut down. arguably better, just worth a mention in the note.

foreign/go/client/tcp/tcp_core_test.go
```diff
@@ -272,3 +273,23 @@ func TestNewIggyTcpClient_StoresProvidedLogger(t *testing.T) {
 		t.Errorf("expected logger output to contain 'source=tcp', got: %q", output)
 	}
 }
+
+func TestLoginUser_RejectedReloginKeepsExistingSession(t *testing.T) {
```
hubcio, line 310: two more asserts would pin the happy paths while you're here: login success sets `Authenticated`, and `LogoutUser` success resets to `Unauthenticated` - right now no test covers either transition.

hubcio (changes requested):
i checked this PR + #3651 proposal, overall it looks ok. main concern is the concurrent login race below. also "seperate" -> "separate" in the PR title.

foreign/go/client/tcp/tcp_session_management.go
chengxilo: Yeah makes sense. I will drop it.

chengxilo: fixed.

foreign/go/client/tcp/tcp_core_test.go
chengxilo: added a test for it now : )

foreign/go/contracts/state.go
chengxilo: done

foreign/go/client/tcp/tcp_core.go
chengxilo: Will address in #3670

hubcio (approved):

numinnex (approved):

spetz (approved):

PR #3654, opened 2026-07-12
feat(connectors): add Redshift sink connector

### LINKED ISSUE #2540: Implement Redshift Sink Connector with S3 staging
A Sink connector to write data to Redshift, using the S3 staging method as this is preferred way to load high volume data with Redshift. Should be configurable - AWS IAM, Connection string, S3 bucket etc.

realonbebeto:
## Summary

Adds a Redshift sink connector for writing Apache Iggy messages into Amazon Redshift using S3-staged Parquet files and Redshift's COPY command.

The connector batches messages into Parquet files, uploads them to a configurable S3 bucket/prefix, and performs bulk COPY loads into the target Redshift table. It supports configurable batch sizing, connection pooling, optional metadata/checksum/origin timestamp fields, payload encoding modes, retry/backoff handling for failed loads, optional archival of staged Parquet files, and runtime metrics logging.

The implementation treats S3 purely as a staging layer for efficient bulk ingestion rather than a destination system. Staged files can either be deleted after a successful load or retained under an archive prefix for replay, auditing, or downstream processing.

It also integrates the connector throughout the project, including workspace membership, connector documentation, release and binary artifact pipelines, version bump automation, and Docker-based integration test infrastructure.

### Disclosures
1. `pgwire`, `sqlparser`, and `tokio-postgres `are introduced as workspace dependencies solely to support the RedshiftFixture used by the integration test suite.
2. The Redshift integration tests are inspired by and closely follow the existing PostgreSQL test setup to maintain consistency across sink connector implementations.

Closes #2540 

## AI Usage
A blend of ChatGPT and Claude was used to help build, refactor the connector and associated test suite


krishvishal:
@realonbebeto can you please fix the CI failures? After that we can review the code.

realonbebeto:
@hubcio resolved the CI failures.

slbotbm:
@realonbebeto looks like your lockfile is stale, and that is why a lot of the tests are failing

realonbebeto:
@slbotbm I made a cargo update on the PR. I think this resolves the staleness.

slbotbm:
@realonbebeto I can see that changes unrelated to the pr have also been committed. can you correct this please?

realonbebeto:
@slbotbm made the unrelated changes correction

krishvishal:
@realonbebeto the CI failed due to clippy. 

We suggest you to setup prek locally, so that easier CI failures are caught locally.

Setup instructions here: https://github.com/apache/iggy/blob/c0218f1c7af0613b915ae70a7f5e1bab887c285e/CONTRIBUTING.md?plain=1#L80

realonbebeto:
@slbotbm @krishvishal prek passed by allowing the commit. Thanks for the hint.

Cargo.lock
hubcio, line 1: what is reason for so many changes in this file?

realonbebeto: I did a cargo update to resolve staleness. This is the context: https://github.com/apache/iggy/pull/3654#issuecomment-4989491660

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+        let include_metadata = self.config.include_metadata.unwrap_or(true);
+        let include_checksum = self.config.include_checksum.unwrap_or(true);
+        let include_origin_timestamp = self.config.include_origin_timestamp.unwrap_or(true);
+        let payload_type = self.payload_format().sql_type();
+
+        let mut params_per_row: u32 = 1; // id
+
+        let mut query =
+            format!("CREATE TABLE IF NOT EXISTS {quoted_table} (id DECIMAL(39, 0) PRIMARY KEY");
```
ryerraguntla: lib.rs:342. Redshift caps NUMERIC precision at 38; ensure_table_exists fails. Mock hides it: map_type maps numeric/Decimal→VARCHAR (mock_load.rs:426,446-447). Also u128 id (39 digits) fits no legal DECIMAL(≤38) even after cap. **Fix:** store id as VARCHAR(40) (or split hi/lo BIGINT); update Arrow schema + mock to match

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+        }
+
+        query.push_str(&format!(", payload {payload_type}"));
+        query.push_str(", created_at TIMESTAMPTZ DEFAULT GETDATE());");
+        params_per_row += 2;
+
+        Ok((query, params_per_row))
+    }
+
+    fn build_copy_sql(&self, s3_path: &str) -> String {
```
ryerraguntla: lib.rs:366-379 emits COPY t FROM ... with no (cols); table always carries created_at DEFAULT GETDATE() (:360) that parquet never writes. Real Redshift COPY-parquet maps by position → count/shape mismatch. Mock hides it: explicit COPY t (col_list) FROM STDIN BINARY + strip_created_at (rs_mock_load.rs:283,486-490). **Fix:** emit an explicit column list matching parquet fields exactly; verify created_at DEFAULT against real Redshift.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+    async fn process_messages(
+        &self,
+        topic_metadata: &TopicMetadata,
+        messages_metadata: &MessagesMetadata,
+        messages: &[ConsumedMessage],
+    ) -> Result<(), Error> {
+        let batch_size = self.config.batch_size.unwrap_or(100) as usize;
+
+        for batch in messages.chunks(batch_size) {
```
ryerraguntla, line 383: lib.rs:222-258: process_messages catches insert_batch Err, bumps insertion_errors, continues, returns Ok(()). Runtime auto-commits offsets (AutoCommitWhen::PollingMessages) → failed batches dropped permanently, no redelivery. README:11-13 admits at-most-once, but this is silent partial loss inside a multi-batch consume.  **Fix:** return Err on first batch failure and stop the pipeline; pair with a runtime offset-commit-on-error fix or a persisted failed-batch manifest.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+    async fn copy_parquet(&self, s3_path: &str) -> Result<(), Error> {
+        let max_retries = self.get_max_retries();
+        let retry_delay = self.get_retry_delay();
+        let sql = self.build_copy_sql(s3_path);
+        let pool = self.get_pool()?;
+
+        tracing::debug!(table = %self.config.target_table, s3_path, "issuing Redshift COPY");
+
+        retry_with_backoff(
```
ryerraguntla, line 522: lib.rs:342,311-323,740. id PRIMARY KEY advertises idempotency Redshift does not enforce (PK informational). Transient Io/08006/57P01 after a server-side commit re-runs the same COPY on the same S3 file → dup rows. Mock hides it: backing Postgres enforces PK (create.rs:111-112) → unique-violation, opposite of Redshift. **Fix:** COPY into a staging table + MERGE/dedup on id, or manifest-idempotent load; stop relying on PK.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+        query.push_str(", created_at TIMESTAMPTZ DEFAULT GETDATE());");
+        params_per_row += 2;
+
+        Ok((query, params_per_row))
+    }
+
+    fn build_copy_sql(&self, s3_path: &str) -> String {
+        // Built via format! (not sqlx binds) because the Redshift/Pgwire endpoint here
+        // uses a Simple Query Handler that doesn't support prepared statements with binds.
+        let credentials = format!(
```
ryerraguntla: **AWS secret inlined in COPY SQL → log-leak + convention break.** lib.rs:369-378 bakes ACCESS_KEY_ID=...;SECRET_ACCESS_KEY=... into the SQL string via format!, run through sqlx::query(AssertSqlSafe(...)). sqlx surfaces statement text on its tracing target (DEBUG default; slow-query WARN — parquet COPY routinely >1s). Sibling postgres_sink never inlines: .bind() everything (postgres_sink/src/lib.rs:381-417). Secret also held as plain String, bypassing secrecy zeroize.  **Fix:** use Redshift IAM_ROLE 'arn:...' (no embedded keys) or session-scoped credentials; never format!/log the COPY string; keep creds in SecretString.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+        query.push_str(&format!(", payload {payload_type}"));
+        query.push_str(", created_at TIMESTAMPTZ DEFAULT GETDATE());");
+        params_per_row += 2;
+
+        Ok((query, params_per_row))
+    }
+
+    fn build_copy_sql(&self, s3_path: &str) -> String {
+        // Built via format! (not sqlx binds) because the Redshift/Pgwire endpoint here
```
ryerraguntla: **Meta (root cause of 1–4): false-green CI.** The hand-rolled pgwire mock (fixtures container.rs + mock_*) diverges from real Redshift on Decimal handling, COPY semantics, and PK enforcement. Prod code shape is even dictated by the mock ("Simple Query Handler doesn't support binds". Test tail wags prod dog.  **Fix:** add a real / Redshift-compatible smoke gate, or document the mock's known divergences and label the suite "wire-shape only, not semantics."

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+        // uses a Simple Query Handler that doesn't support prepared statements with binds.
+        let credentials = format!(
+            "CREDENTIALS 'ACCESS_KEY_ID={};SECRET_ACCESS_KEY={}'",
+            self.config.aws_access_key_id.expose_secret(),
+            self.config.aws_secret_access_key.expose_secret()
+        );
+
+        format!(
```
ryerraguntla, line 650: lib.rs:377 vs :332 — COPY uses raw unquoted target_table; CREATE quotes it. Mixed-case/special name → CREATE "MyTable" but COPY mytable → wrong/missing table; quote char → injection. **Fix**: reuse quote_identifier in build_copy_sql.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+        let content = encode_parquet(&record_batch)?;
+
+        tracing::debug!(
+            bytes = content.len(),
+            rows = record_batch.num_rows(),
+            "encoded parquet batch"
+        );
+
+        let s3_path = self.upload_parquet(&content).await?;
```
ryerraguntla, line 478: lib.rs:294-296,231 — non-atomic upload→copy→archive: post-COPY archive/delete failure returns Err → whole batch counted insertion_errors though rows loaded; staged file leaks; archive-then-delete-fail double-stores. **Fix:** separate commit from cleanup; cleanup failure = warning, not batch error.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+        )?;
+
+        let content = encode_parquet(&record_batch)?;
+
+        tracing::debug!(
+            bytes = content.len(),
+            rows = record_batch.num_rows(),
+            "encoded parquet batch"
+        );
+
```
ryerraguntla, line 477: lib.rs:294-296 — COPY failure ?-returns before delete → staged parquet orphaned forever (no runtime retry consumes it). **Fix**: best-effort delete on COPY failure.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+        let scheme = &conn_str[..scheme_end + 3];
+        let rest = &conn_str[scheme_end + 3..];
+
+        // Stop preview at the first sensitive boundary
+        let safe_end = rest
+            .find([':', '@', '?', '/'])
+            .unwrap_or(rest.len())
+            .min(PREVIEW_LEN);
+
+        let preview = &rest[..safe_end];
```
ryerraguntla, line 1030: lib.rs:730 — redact_connection_string slices &rest[..safe_end] on byte index, not char boundary; multibyte DSN userinfo/host → panic across FFI in open(). **Fix:** char-safe slice (no-scheme branch at :734 already safe).

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+#[async_trait]
+impl Sink for RedshiftSink {
+    async fn open(&mut self) -> Result<(), Error> {
+        tracing::info!(
+            sink_id = self.id,
+            table = %self.config.target_table, "opening Redshift sink connector"
+        );
+
+        self.connect().await?;
+        self.ensure_table_exists().await?;
```
ryerraguntla: lib.rs:69,191-212 — CREATE TABLE IF NOT EXISTS with schema from current include_* flags; flipping a flag between runs keeps old table → positional COPY loads into wrong columns silently. **Fix:** validate live schema on open, error on drift.

core/connectors/sinks/redshift_sink/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Plugin Fields
+
+| Field | Required | Default | Description |
+| --- | --- | --- | --- |
+| `connection_string` | yes | — | Postgres-wire connection string used to reach the Redshift cluster and issue the `COPY` command. |
+| `target_table` | yes | — | Destination Redshift table that batches are copied into. |
+| `batch_size` | no | `100` | Number of messages buffered per Parquet file / `COPY` operation. |
+| `max_connections` | no | `5` | Size of the connection pool used against Redshift. |
+| `include_metadata` | no | `false` | Stores stream/topic/partition/offset/timestamp/schema fields alongside the payload. |
```
ryerraguntla: README.md:60-62 — defaults say false; code defaults true (lib.rs:271-273,334-336). **Fix**: table → true.

core/connectors/sinks/redshift_sink/src/config.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/// This connector supports:
+/// 1. Byte -> which has VARBYTE as the Redshift equivalent
+/// 2. Text -> which has VARCHAR as the Redshift equivalent
+///
+/// We dont have Json because we are using parquet as a means to sink ingestion
+/// As at the development of this connector there's no direct parquet type that matches JSON
+/// For JSON needs Reshshift has SUPER(VARCHAR can be parsed by JSON_PARSE)
+#[allow(unused)]
```
ryerraguntla: config.rs:96-103 — dead PayloadFormat::Json masked by #[allow(unused)]; from_config never yields it; "json" silently → Text. **Fix**: delete variant + attr, or log the alias.

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1182 @@
...
+
+fn origin_timestamp_column(messages: &[ConsumedMessage]) -> ArrayRef {
+    Arc::new(TimestampMicrosecondArray::from_iter_values(
+        messages.iter().map(|v| v.origin_timestamp as i64),
+    ))
+}
+
+fn payload_column(messages: &[ConsumedMessage], format: PayloadFormat) -> Result<ArrayRef, Error> {
+    match format {
+        PayloadFormat::Varbyte => {
```
ryerraguntla, line 975: **Perf hot path:** lib.rs:685 v.payload.clone().try_to_bytes() double-materializes per message; lib.rs:654-659 clones stream/topic String N× per batch. **Fix:** drop the clone (try_to_bytes on borrow) and use .as_str() or a DictionaryArray.

ryerraguntla:
Bunch of nits, but needs to be looked into and fixed
1. messages_processed counts failed batches 
      - yet README:118 claims loaded-only (lib.rs:237-238) · 
      - upload_parquet has no retry while COPY does (:381-399) · 
      - naive-micros→TIMESTAMPTZ UTC assumption (:596-598) · 
      - lossy as i32/as i64 casts (:649,661) · 
      - empty s3_prefix → malformed archive key (:413-414) · 

2. dead params_per_row return (:330-363) · 
3. invalid crates.io category "warehouse" → "database" (Cargo.toml:25) · 
4. const ICEBERG_SINK_KEY = "redshift" copy-paste (int.rs:36) · 
5. "IcebergFixture" error strings + "bytea" fixture doc mislabels (fixtures sink.rs:393,429,198,245,292) ·
6. typos "Reshift"/"vis S3"/"Reshshift"/"We dont" · 
7. payload_column Vec churn + uncompressed parquet writer 


Did you check "typos" as part of precheck?

Good effort. Thanks for chiming in to contribute.

realonbebeto:
> Bunch of nits, but needs to be looked into and fixed
> 
> 1. messages_processed counts failed batches
>    
>    * yet README:118 claims loaded-only (lib.rs:237-238) ·
>    * upload_parquet has no retry while COPY does (:381-399) ·
>    * naive-micros→TIMESTAMPTZ UTC assumption (:596-598) ·
>    * lossy as i32/as i64 casts (:649,661) ·
>    * empty s3_prefix → malformed archive key (:413-414) ·
> 2. dead params_per_row return (:330-363) ·
> 3. invalid crates.io category "warehouse" → "database" (Cargo.toml:25) ·
> 4. const ICEBERG_SINK_KEY = "redshift" copy-paste (int.rs:36) ·
> 5. "IcebergFixture" error strings + "bytea" fixture doc mislabels (fixtures sink.rs:393,429,198,245,292) ·
> 6. typos "Reshift"/"vis S3"/"Reshshift"/"We dont" ·
> 7. payload_column Vec churn + uncompressed parquet writer
> 
> Did you check "typos" as part of precheck?
> 
> Good effort. Thanks for chiming in to contribute.

Thanks for the review. I might have missed running 'typos'.

core/connectors/sinks/redshift_sink/src/lib.rs
realonbebeto: @ryerraguntla @hubcio 
I've been thinking about this with the proposed design where id is VARCHAR(40), PKs are dropped, and uniqueness is enforced via MERGE from staging into the target.

The sink only emits a small, fixed set of Redshift queries. SELECT 1, CREATE TABLE, COPY, and MERGE are the only statements requiring dynamic column lists, which seems reasonable to validate with the pgwire mock.

The main gaps I see are:
- Malformed Parquet – I don't think this can be meaningfully tested without a real Redshift instance.
- Schema mismatch – This can be validated during open(), as pointed elsewhere in this review. And help clarify why the sibling postgres_sink does not enforce this? My assumption is the auto_create flag transfers the burden to the user to ensure there's no schema mismatch(by means of migrations)
- IAM/S3 permissions – This is more of an user environment/configuration issue than connector behaviour unless we want to check for S3 and Redshift permissions from the iam_role, access_key and secret_key config provide.

Given that, I'm wondering how the smoke test against a real Redshift instance would be to cover the 4 queries semantics.

hubcio: @realonbebeto 

it took me a while to actually understand the question. so, thing is the mock isn't really validating those 4 statements. `map_type` rewrites `DECIMAL(_)` to VARCHAR and `VARBYTE` to BYTEA, and between `strip_created_at` and the column list rebuilt from `information_schema`, the COPY that runs is a postgres one with an explicit column list, not the columnless `FORMAT AS PARQUET` you emit. VARCHAR(40) fixes the decimal but not that.

maybe it'd better if you rather `map_type` error on what it can't represent than substitute? plusofc a readme line on what the suite actually covers.

schema mismatch in open(), yeah. postgres_sink isn't precedent, just missing there too, and a bound INSERT names the bad column, whereas a failed COPY just points you at the load-error views.

iam/s3 is environment mostly, but `validate()` requires static keys and `Credentials::new` gets None for both token slots, so temp creds and instance profiles are out. from whatisee aws-creds already falls back to env/profile/container when the access key is None, so making those two fields `Option` and dropping the validate() check is most of it.

nothing per-PR works for a real cluster, fork PRs get no secrets. before merge i'd want you running it once on serverless and pasting it - CREATE TABLE verbatim, one COPY of a parquet the connector made, same file again for MERGE, rows read back. then `#[ignore]`d behind `REDSHIFT_TEST_DSN`.

dropping PKs i agree with, but the mock's postgres was the only uniqueness check in the suite, so MERGE needs its own test.

TLDR:
1. fail loudly, no quiet translation
2. update readme with test suite descr
3. maybe run it once against redshift serverless if you have access
4. make aws keys optional
5. merge dedup test, not sure about that

realonbebeto:
@ryerraguntla I haven't been able to test the S3 → Redshift flow yet due to limitations with setting up a free-tier AWS account. The [fix(redshift): improved correctness and reliability](https://github.com/apache/iggy/pull/3654/commits/3708d133d5c0e2b1f06cf62ca2ba0af42d7b6d7c)  is in place, but real Redshift testing is pending.

realonbebeto:
Schema  difference queries display the same types and the queries run against Redshift Serverless.

redshift-target

redshift

hubcio:
@realonbebeto looks good. can you regenerate `Cargo.lock`? i think we can merge this PR.

realonbebeto:
@hubcio I have resolved the conflict that was there on the Cargo.lock

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1482 @@
...
+        }
+    }
+
+    async fn process_messages(
+        &self,
+        topic_metadata: &TopicMetadata,
+        messages_metadata: &MessagesMetadata,
+        messages: &[ConsumedMessage],
+    ) -> Result<(), Error> {
+        let batch_size = self.config.batch_size.unwrap_or(100) as usize;
```
slbotbm, line 381: batch_size = 0 is accepted and reaches `messages.chunks(batch_size)`, which panics. Because this is invoked through the plugin’s extern "C" boundary, a bad operator config can abort the connectors process instead of
returning a configuration error. Please reject zero during startup or clamp the stored value to at least one

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1482 @@
...
+            sink_id = self.id,
+            table = %self.config.target_table, "opening Redshift sink connector"
+        );
+
+        self.connect().await?;
+        // Ensuring tables exist
+        self.ensure_tables_exist().await?;
+        // Checking for schema drift
+        self.ensure_schema_match().await?;
+        Ok(())
```
slbotbm, line 77: `RedshiftSinkConfig::validate()` is never called. An empty `aws_iam_role` therefore lets the connector start, create tables, and upload the Parquet file before failing while building the later `COPY`. Both shipped examples omit this field despite documenting it as required. Please call `self.config.validate()?` before connecting and add `aws_iam_role` to `config.toml` and the README example.

core/connectors/sinks/redshift_sink/src/config.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        if self.aws_region.is_empty() {
+            errors.push_str(", aws_region is empty\n");
+        }
+
+        if self.aws_iam_role.is_empty() {
+            errors.push_str(", aws_iam_role is empty\n");
+        }
+
+        if let (Some(access), Some(secret)) = (&self.aws_access_key_id, &self.aws_secret_access_key)
```
slbotbm: Static S3 credentials need to be validated as a pair. This branch only checks them when both options are present, so an access key without a secret, or vice versa, is accepted. `Credentials::new` then treats the access key as explicit credentials and does not fall back to the environment or instance-role chain; the first S3 upload fails after startup. Reject partial credentials while preserving the both-absent provider-chain case.

core/connectors/sinks/redshift_sink/src/config.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+    }
+
+    pub fn sql_type(&self) -> &'static str {
+        match self {
+            PayloadFormat::Varbyte => "VARBYTE",
+            PayloadFormat::Text => "VARCHAR",
+        }
+    }
```
slbotbm, line 151: The generated payload types are too small for general Iggy messages: bare `VARCHAR` becomes `VARCHAR(256)` in Redshift, and bare `VARBYTE` defaults to 64,000 bytes. Text/JSON payloads over 256 bytes, and binary payloads over 64 KB, fail during `COPY`. Please declare explicit supported maxima, such as `VARCHAR(MAX)` and a suitable `VARBYTE(n)`, or validate and document a lower connector payload limit.

slbotbm (dismissed):

hubcio (approved):

hubcio (approved):

spetz (dismissed):

thepatrykk (dismissed):

hubcio:
@realonbebeto looks like test is failing.
```
Summary [ 479.458s] 1581 tests run: 1580 passed (7 slow), 1 failed, 3170 skipped
  TRY 4 FAIL [   0.005s] ( 869/1581) iggy_connector_redshift_sink tests::given_payload_format_should_return_correct_sql_type
error: test run failed
```


https://github.com/apache/iggy/actions/runs/32817147871/job/97707918535?pr=3654

core/connectors/sinks/redshift_sink/src/lib.rs
```diff
@@ -0,0 +1,1493 @@
...
+        cols: &str,
+    ) -> Result<String, Error> {
+        // Redshift allows this from the docs
+        // https://docs.aws.amazon.com/redshift/latest/dg/r_COPY_command_examples.html
+        let iam_role = quote_identifier(&self.config.aws_iam_role)?.replace('"', "");
+
+        let region = quote_identifier(&self.config.aws_region)?.replace('"', "");
+
+        Ok(format!(
+            "COPY {} ({}) FROM '{}' CREDENTIALS 'aws_iam_role={}' FORMAT AS PARQUET REGION '{}';",
```
hubcio: `REGION` is not supported with parquet COPY. the aws notes on columnar COPY (https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html) say it fails with "REGION argument is not supported for PARQUET based COPY", and it's missing from their supported parameter list. the mock can't catch this, `execute_copy` only checks for `Parquet` and ignores the rest. drop the clause - COPY defaults to the cluster region and the bucket has to be in that region anyway. also, the thread only has DDL screenshots so far, one pasted real COPY output would settle it.

core/connectors/sinks/redshift_sink/README.md
```diff
@@ -0,0 +1,217 @@
...
+target_table = "iggy_messages"
+batch_size = 100
+max_connections = 10
+include_metadata = true
+include_checksum = true
+include_origin_timestamp = true
+payload_format = "varbyte"
+aws_access_key_id = "admin"
+aws_secret_access_key = "password"
+aws_iam_role ="aws_iam_role=arn:aws:iam::0123456789012:role/iggyRole"
```
hubcio: `build_copy_sql` already prepends `aws_iam_role=`, so this example ends up as `CREDENTIALS 'aws_iam_role=aws_iam_role=arn:...'` and the COPY fails. use the bare arn like `config.toml` does.

core/integration/tests/connectors/fixtures/redshift/redshift_mock/copy.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use sqlparser::{
+    ast::ObjectName,
+    keywords::Keyword,
+    parser::{Parser, ParserError},
+    tokenizer::Token,
+};
+
+use crate::connectors::fixtures::redshift::redshift_mock::{
+    expect_word, parse_number_literal, parse_string_literal,
```
hubcio: same for `create.rs`. neither file is declared in `mod.rs`, so they never compile - `expect_word`, `parse_string_literal` and `parse_number_literal` don't exist anywhere in the tree. adding the `mod` lines back fails with E0432. delete both files.

hubcio (changes requested):

hubcio (approved):

numinnex (approved):

PR #3656, opened 2026-07-12
feat(configs): add devcontainer config for reproducible dev and cross-platform code track

jiengup:

Relates to NONE

## Rationale

The repository lacked a devcontainer configuration, making onboarding inconsistent — each developer needs to manually install the correct Rust toolchain, Node.js, and system dependencies. 
Moreover, for local development on other platforms (e.g, macOS), the `compio-driver` will fallback to `poll` instead of `io-uring`, which will make source code tracing difficult and different from iggy design.
On GitHub Codespaces, the default 8GB machine ran out of memory during rust-analyzer indexing, causing repeated SIGTERM crashes.

## What changed?

Added `.devcontainer/Dockerfile` and `.devcontainer/devcontainer.json` that:

- Mirror the production Dockerfile dependencies (Rust 1.96 via rustup, libssl-dev, libhwloc-dev, libudev-dev, nodejs) so the dev environment matches CI.
- Configure `hostRequirements` (4-core / 16GB) to prevent rust-analyzer OOM kills on Codespaces.
- Cap build parallelism (`CARGO_BUILD_JOBS=4`) and enable proc-macros/build-scripts for usable IDE features while staying within memory limits.
- Forward ports for transports (HTTP 3000, TCP 8090, WebSocket 8092) plus the web UI dev server (3050).
    - Note that QUIC(8080) won't actually be reachable from the host through the forward cause port forwarding in vs code / codespaces supports TCP only. It's just for container-internal use.
- Run `npm --prefix web ci` on container creation so the web frontend is ready immediately.

## Local Execution

- Passed — container builds, `cargo check` succeeds, rust-analyzer code navigation (Go to Definition / autocomplete) works.
- Verified on both local Docker Desktop (macOS arm64) and GitHub Codespaces (4-core/16GB). Both enviroment can navigate to the io-uring driver code in `compio`.

## Screenshot (Github Codespace)
image

## KNOWN LIMITATION: Docker-based bind-mount tests will fail in this devcontainer

The devcontainer uses `docker-outside-of-docker`: Docker commands inside the container
talk to the host's Docker daemon via the Docker socket.

The workspace is mounted at /workspaces/iggy inside the container, but the host
Docker daemon only knows the real host path (${localWorkspaceFolder}).

Bind mounts from compose files or testcontainers resolve to /workspaces/iggy/...,
which the host daemon cannot reach.

Affected tests:
  - BDD tests:          ./scripts/run-bdd-tests.sh
  - HTTP sink:          cargo test -p integration -- connectors::http
  - HTTP config provider: cargo test -p integration -- connectors::http_config_provider

Workarounds:
  - Run affected tests on the host directly, outside the devcontainer
  - Use native CI (GitHub Actions) where path translation is handled

A hint will show up after the devcontainer build finishes:

image

## AI Usage

LLM is used to generate the devcontainer configuration files and all the changes are checked by the human.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,65 @@
...
+    "context": "..",
+    "args": {
+      "RUST_VERSION": "1.96"
+    }
+  },
+  "features": {},
+  "remoteUser": "vscode",
+  "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/iggy,type=bind,consistency=cached",
+  "workspaceFolder": "/workspaces/iggy",
+  "runArgs": [
```
slbotbm, line 20: `--cap-add=SYS_NICE` is also required here

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    build-essential \
+    ca-certificates \
+    curl \
+    git \
+    pkg-config \
+    libssl-dev \
+    libhwloc-dev \
+    libudev-dev \
+    liblzma5 \
+    nodejs \
```
slbotbm: Pinning nodejs to a specific version would be good

slbotbm (changes requested):

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,66 @@
...
+  "hostRequirements": {
+    "cpus": 4,
+    "memory": "16gb",
+    "storage": "32gb"
+  },
+  "build": {
+    "dockerfile": "Dockerfile",
+    "context": "..",
+    "args": {
+      "RUST_VERSION": "1.96"
```
hubcio: please check `scripts/ci/sync-rustc-version.sh`: it only scans `Dockerfile*` files, so on the next toolchain bump it will rewrite the `ARG RUST_VERSION` default in the dockerfile but not this override - the devcontainer would silently keep building the old base. the arg passes the same value as the dockerfile default.

i think whole `build.args` block can just be dropped.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,66 @@
...
+    }
+  },
+  "features": {},
```
hubcio: `cargo test -p integration` won't work inside the container - the testcontainers-based suites need a docker socket and there's no docker cli here. worth adding `ghcr.io/devcontainers/features/docker-outside-of-docker` here (which would also give the bundled `ms-azuretools.vscode-docker` extension something to talk to), or noting the limitation in the PR description.

did you actually try to run some tests in this container, or is it purely for rust analyzer?

i didn't work with devcontainers + vscode.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,66 @@
...
+    "type=volume,source=iggy-target,target=/workspaces/iggy/target"
+  ],
+  "containerEnv": {
+    "CARGO_BUILD_JOBS": "4",
+    "IGGY_HTTP_ADDRESS": "0.0.0.0:3000",
+    "IGGY_QUIC_ADDRESS": "0.0.0.0:8080",
+    "IGGY_TCP_ADDRESS": "0.0.0.0:8090",
+    "IGGY_WEBSOCKET_ADDRESS": "0.0.0.0:8092"
+  },
+  "forwardPorts": [3000, 3050, 8080, 8090, 8092],
```
hubcio, line 38: port forwarding in vs code / codespaces tunnels TCP only, and QUIC is UDP - so 8080 won't actually be reachable from the host through the forward. fine to keep for container-internal use, but the "all transports" claim in the description doesn't hold for QUIC.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,66 @@
...
+  "portsAttributes": {
+    "3000": { "label": "HTTP (REST)" },
+    "3050": { "label": "Web UI (vite dev)" },
+    "8080": { "label": "QUIC" },
+    "8090": { "label": "TCP (binary)" },
+    "8092": { "label": "WebSocket" }
+  },
+  "postCreateCommand": "sudo chown -R vscode:vscode target /usr/local/cargo/registry 2>/dev/null; rustup show && npm --prefix web ci",
```
hubcio: `-R` re-walks the whole cargo registry volume on every container recreate; after the first create the volumes are only ever written as `vscode`, so a non-recursive chown of the two mount points is enough. the `2>/dev/null` also hides real chown failures - if it ever does fail, the follow-up cargo permission error is much harder to trace.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+ARG NODE_VERSION=22
+RUN curl -fsSL "https://deb.nodesource.com/setup_${NODE_VERSION}.x" | bash - \
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/*
+
+# Pre-fetch the pinned toolchain components (rustfmt, clippy) so the first
+# in-container build does not stall resolving them. rust-src is added for
+# rust-analyzer sysroot / standard-library resolution.
+COPY rust-toolchain.toml /tmp/rust-toolchain.toml
+RUN cd /tmp && rustup show && rustup component add rust-src && rm rust-toolchain.toml
```
hubcio, line 52: the `1.96` tag ships the newest patch as the default toolchain (currently 1.96.1), and the pin then installs 1.96.0 next to it, so the image carries two full toolchains. unavoidable while the sync script normalizes dockerfiles to the short tag - just noting where the extra size comes from.

hubcio (changes requested):

.devcontainer/Dockerfile
jiengup: As of now, we don't have a good way to specify that the long version number of the docker image is consistent with `rust-toolchain.toml`.

.devcontainer/devcontainer.json
jiengup: Thanks for your insight, I've modified the PR description to point this.

.devcontainer/devcontainer.json
jiengup: I haven't run any UTs in either GitHub Codespace or local devcontainer. Cause GitHub Codespaces run on the Azure VM, which has incomplete io-uring support. 

That's to say, `cargo run --bin iggy-server -- --with-default-root-credentials --fresh` will panic soon after it's running, reporting `thread 'shard-0' (82422) panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/compio-driver-0.12.3/src/asyncify.rs:118:25: the thread pool is needed but no worker thread is running`

I guess it's because some opcodes are fallbacked to the blocking thread pool. Because QUIC server can start successfully while TCP/HTTP/WebSocket report the same issue above.

I'm not familiar with iggy test infrastracture and I don't know if it can run without a server running. So at first this PR is intent to support rust-analyzer for simple code review or editing in PR turns .

BTW, I want to propose adding a `IGGY_THREAD_POOL_LIMIT' env to replace the constant `0` when initializing the iggy per-thread shard. Then maybe some senarios can be realized.

Any ideas or suggestions about this?

hubcio: to be honest, I intentionally disabled ability to run iggy with thread pool on linux. we advise ourselves as io-uring software, so I'd rather not support half-baked thread pool options - some people might run iggy on box which doesnt have ioruing, so it'll automatically switch to thread pool and they'll complain about performance and ditch iggy because of that.

that being said, maybe I could settle for a workaround: enable thread pool only for debug mode, and only if io uring is not supported.

does that make sense? :)

.devcontainer/Dockerfile
hubcio: if this is really needed, you can update the script `scripts/ci/sync-rustc-version.sh` to take care of that (besides checking rust-toolchain.toml and dockerfiles it should also check `.devcontainer/devcontainer.json`)

jiengup: Done, check the latest push please.

.devcontainer/devcontainer.json
jiengup: Totally agree, I don't mean to fully support configurable thread num. We can add a debug-face env variable for some special scenarios, such as Github Codespace.

I've added the `docker-outside-of-docker` feature in the latest commit and passed several docker-based integration UTs in Github Codespace after changing `proactor.thread_pool_limit(0)` to `proactor.thread_pool_limit(5)`. The compilation is very slow, and the storage is limited for a large number of UT binaries. But we can finally support a quick runtime for our codebase. :)

If you think making the thread num configurable for debug is a good idea, I can open a follow-up issue for it, or you can give any other instructions.

hubcio:
@jiengup have you already joined our discord? if yes, whats your handle?

numinnex (dismissed):

.devcontainer/devcontainer.json
slbotbm, line 1: Docker-based tests with bind mounts will fail. The workspace is mounted at /workspaces/iggy, but Docker commands use the host daemon through docker-outside-of-docker. Host Docker cannot resolve container-only paths used by BDD Compose volumes or testcontainers.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,67 @@
...
+    "--cap-add=SYS_NICE"
+  ],
+  "mounts": [
+    "type=volume,source=iggy-cargo-registry,target=/usr/local/cargo/registry",
+    "type=volume,source=iggy-target,target=/workspaces/iggy/target"
+  ],
+  "containerEnv": {
+    "CARGO_BUILD_JOBS": "4",
+    "IGGY_HTTP_ADDRESS": "0.0.0.0:3000",
+    "IGGY_QUIC_ADDRESS": "0.0.0.0:8080",
```
slbotbm, line 34: Iggy's QUIC endpoint uses UDP transport, not TCP. This setting will make the QUIC client time-out.

fix: Publish UDP explicitly when creating the development container.

slbotbm (changes requested):

.devcontainer/devcontainer.json
jiengup: Oh, I see this issue.

Maybe there are two solutions:
- change `bind_mount` in `fixture` to  `with_copy_to` to obtain the files which are needed in the UTs.
    - for BDD tests, we need to add `COPY` of feature files in the Dockerfile of Rust, Python, Java and CPP SDK, like what we have done in Go and C#. 
- or, add hints to the failure of these UTs for the developers who use devcontainer.

Which do you prefer? Or do you have any other ideas for this problem? @slbotbm

jiengup: And, I want to know the criteria of this PR merge.

Like, pass all the integration tests in the devcontainer?

hubcio: no one from core team uses docker containers. we prefer to run the tests on bare metal, the way god intended iggy is used. docker image is only for convenience. it's up to you.

.devcontainer/devcontainer.json
jiengup: Fixed in the newest patch.

.devcontainer/devcontainer.json
jiengup: Thanks for your effort. @slbotbm @hubcio 

In order not to break the current UTs implementation, and considering that the devcontainer-based development is not so common, a hint will show after the devcontainer build now, like this:

image

Meanwhile, the known UT problem in the devcontainer is updated in the PR message. 
Please see the latest commit.

hubcio (approved):

hubcio:
fyi in https://github.com/apache/iggy/pull/3720 i'm bumping rust to 1.97.1

spetz (approved):

slbotbm (approved):

PR #3657, opened 2026-07-12
feat(connectors): use TCP client for Flink sink

### LINKED ISSUE #2484: Flink connector - transport - http to tcp
Flink connector currently uses http as the transport. Needs to support TCP.

Standing-Man:

- Closes  #2484.

## Rationale

The Flink connector used different transports for reading and writing messages. The source already used the native TCP client, while the sink sent messages through the HTTP client and hardcoded the HTTP port to 3000.

This caused the same IggyConnectionConfig to behave differently between the source and sink. It also required deployments to expose the HTTP API even when the configured Iggy endpoint used the TCP port.

The source and sink also parsed server addresses independently. The source relied on splitting the address by `:`, while the sink used URI parsing. This produced inconsistent behavior and did not reliably support IPv6 or validate malformed endpoints.

## What changed?

- Replaced IggyHttpClient with the blocking IggyTcpClient in the Flink sink.
- Configured the TCP client using the existing connection settings:server host and port
   - username and password
   - connection and request timeouts
   - retry count and backoff
   - TLS
- Updated IggySinkWriter to send messages through TCP.
- Updated sink shutdown to flush pending messages and close the TCP client, while preserving both flush and close failures.
- Added a shared TcpEndpoint parser used by both the Flink source and sink.
- Added support for:host-only addresses using the default TCP port 8090
   - host:port
   - tcp://host:port
   - bracketed IPv6 addresses
- Added validation for unsupported schemes, missing hosts, invalid ports, and addresses containing user info, paths, queries, or fragments.
- Added unit tests covering valid and invalid TCP endpoint formats.
- Ignored Gradle-generated .kotlin cache directories.

## Local Execution

- Passed / not passed
Passed
Ran


hubcio:
@mmodzelewski and also please check this one :)

Standing-Man:
Just a gentle reminder. When you have a chance, could you please take a look at this PR? Thanks! cc @hubcio and @mmodzelewski.

foreign/java/external-processors/iggy-connector-flink/iggy-connector-library/src/main/java/org/apache/iggy/connector/flink/sink/IggySinkWriter.java
```diff
@@ -180,9 +179,25 @@ public void flush(boolean endOfInput) throws IOException {
...
     public void close() throws Exception {
-        // Flush any remaining buffered records
-        flush(true);
-        // Note: HTTP client doesn't have close() method - connections managed by Java HttpClient pool
+        IOException flushException = null;
+        try {
+            flush(true);
+        } catch (IOException e) {
+            flushException = e;
+        }
```
mmodzelewski: `flush()` wraps send failures in `ConnectorException`, which extends `RuntimeException`, so this `catch (IOException e)` never fires for the most common failure mode. The exception then propagates before `tcpClient.close()` runs, leaking the Netty channels and event-loop threads in the TaskManager JVM. This defeats the scenario this block was written for. Widening the catch keeps the intended semantics:

```suggestion
        Exception flushException = null;
        try {
            flush(true);
        } catch (Exception e) {
            flushException = e;
        }
```

foreign/java/external-processors/iggy-connector-flink/iggy-connector-library/src/main/java/org/apache/iggy/connector/flink/source/IggySource.java
```diff
@@ -154,23 +155,12 @@ public SimpleVersionedSerializer<IggySourceEnumeratorState> getEnumeratorCheckpo
...
 
             // Create async TCP client using builder pattern with auto connect and login
             return AsyncIggyTcpClient.builder()
-                    .host(host)
-                    .port(port)
+                    .host(endpoint.host())
+                    .port(endpoint.port())
                     .credentials(connectionConfig.getUsername(), connectionConfig.getPassword())
                     .connectionPoolSize(4)
                     .buildAndLogin()
```
mmodzelewski, line 172: The sink now honors connection/request timeouts, retry policy, and TLS from `IggyConnectionConfig`, but the source still ignores them. With `enableTls(true)` the sink connects over TLS while the source silently uses plaintext, so the same config still behaves differently between source and sink. `AsyncIggyTcpClientBuilder` already exposes the same options:

```suggestion
            return AsyncIggyTcpClient.builder()
                    .host(endpoint.host())
                    .port(endpoint.port())
                    .credentials(connectionConfig.getUsername(), connectionConfig.getPassword())
                    .connectionTimeout(connectionConfig.getConnectionTimeout())
                    .requestTimeout(connectionConfig.getRequestTimeout())
                    .retryPolicy(RetryPolicy.fixedDelay(
                            connectionConfig.getMaxRetries(), connectionConfig.getRetryBackoff()))
                    .tls(connectionConfig.isEnableTls())
                    .connectionPoolSize(4)
                    .buildAndLogin()
```

(requires an `org.apache.iggy.config.RetryPolicy` import)

mmodzelewski (changes requested):

foreign/java/external-processors/iggy-connector-flink/iggy-connector-library/src/main/java/org/apache/iggy/connector/flink/sink/IggySinkWriter.java
Standing-Man: Good catch, Catching `Exception` would violate the repository’s Checkstyle rules, which prohibit catching overly broad exception types. Therefore, this change catches only two exception types: `IOException` and `RuntimeException`.

* `IOException` is the checked exception declared by `flush()`.
* `RuntimeException` covers both `ConnectorException` and runtime failures from the TCP client.

mmodzelewski: sounds good

mmodzelewski (approved):

hubcio (approved):

PR #3658, opened 2026-07-12
feat(connectors): ship connector plugins in the iggy-connect docker image

### LINKED ISSUE #3614: CI: ship connector plugins in the `apache/iggy-connect` docker image
the `apache/iggy-connect` image ships only the `iggy-connectors` runtime binary. no plugin `.so` files inside. so anyone who wants to actually try a connector has to get the plugins from somewhere else first: 9 of our 15 plugins are prebuilt in the edge github pre-release tarballs (download, unpack, mount into the container), and the remaining 6 (delta, doris, http, influxdb sink + source, mongodb) are not prebuilt anywhere - those require cloning the repo and a full rust toolchain. either way it is a lot of friction for what should be a five minute "docker run + config file" experience.

## proposal

two flavors of the image, distinguished by tag suffix (like `python:3.12-slim`):

| tag | contents |
| --- | --- |
| `edge`, `x.y.z`, `latest` | runtime + all connector plugins baked in (new default) |
| `edge-slim`, `x.y.z-slim`, `latest-slim` | runtime binary only (what the image is today) |

the fat image becomes the default because that is what a new user expects: pull, point a config at a connector, go. size-sensitive deployments that bring their own plugins switch to `-slim`.

plugins land in `/usr/local/lib`, which is already on the runtime's plugin search path - so a connector config can reference just `libiggy_connector_postgres_sink` with no absolute path. zero-config out of the box.

## what gets bundled

all 15 cdylib plugins currently in the workspace:

- sinks: delta, doris, elasticsearch, http, iceberg, influxdb, mongodb, postgres, quickwit, s3, stdout
- sources: elasticsearch, influxdb, postgres, random

side note: the edge tarball plugin list in `_build_rust_artifacts.yml` is a hardcoded default with only 9 of these, and the packaging step silently skips a missing `.so` instead of failing. the plugin list should come from one place so new connectors stop missing release artifacts - pre-merge CI already derives bin + cdylib packages from `cargo metadata`, and the cdylib filter alone is exactly this set, so there is prior art to reuse.

## how (roughly)

- `core/connectors/runtime/Dockerfile`: the cargo-chef cook and build steps currently target `-p iggy-connectors` only, so the heavy plugin dependencies (deltalake, arrow, aws sdk, mongodb driver...) are not in the cached recipe today. both steps grow the plugin packages, plus a second final stage that COPYs the `.so` files into `/usr/local/lib`. slim stage stays exactly what it is today.
- `.github/actions/utils/docker-buildx` action: needs to learn to build a specific dockerfile stage (`target:` is not plumbed through today), or `.github/config/publish.yml` grows a second component entry per flavor (which also means touching the hardcoded docker component allowlist in the publish plan job).
- `docker-manifests` job: pushes only `edge` / `x.y.z` / `latest` today, so the `-slim` variants are new. the real work is the digest plumbing: per-arch digests travel as `docker-digest-<component>-<arch>` artifacts and the manifest is created by globbing every digest in one directory, so the two flavors must keep separate digest sets or they merge into one manifest.
- edge gate: `rust-connectors` currently gates on `crates: [iggy-connectors]` only (plugins are deliberately excluded since the image does not ship them - the config comment says so). with plugins bundled, the gate expands to the full connector crate set, and that comment gets updated.
- licensing: the image generates `LICENSE-binary` from `core/connectors/runtime/Cargo.toml` only. `third-party-licenses.sh` already accepts multiple `--manifest` flags and merges rust closures, so the generate step in the dockerfile and the "validate iggy-connect bundle" step in the validate action both grow the 15 plugin manifests. heads up: the plugin dependency closures have never been license-validated anywhere, so the first run may surface ASF category-x findings that block bundling a plugin until resolved.

## tradeoffs

- image size grows by 15 statically-linked `.so` files (arrow, deltalake, aws sdk closures - and the release profile does not strip symbols today, only default debuginfo stripping). needs an actual measurement before merging, plus a strip step if the numbers are silly. `-slim` exists for anyone who cares.
- docker build time goes up materially: the plugin deps compile cold until the chef recipe includes them, CI caching is layer-level (any workspace dependency change re-cooks everything, per arch, with lto + codegen-units=1), and the publish job has a 60 minute timeout. worth watching on the first runs.
- tag semantics change: `:edge` starts pulling fat on the next master push after merge, `:latest` and versioned tags only from the next release onward. existing users get a bigger image with identical behavior (plugins are inert files until a config loads them), but release notes and the docker hub description must call it out.
- bigger dependency surface on the default image means CVE scanners will flag plugin closures for every user, including ones who never load a plugin. `-slim` is the answer, but expect the reports.

## Tasks

- [ ] dockerfile: extend chef cook + build to plugin crates, fat final stage with `.so` in `/usr/local/lib`
- [ ] publish pipeline: `-slim` flavor end to end - buildx action (stage target or second component entry + plan-job allowlist), per-flavor digest artifacts, `-slim` manifests
- [ ] expand edge gate to plugin crates, update the stale config comment
- [ ] unify the plugin list (docker image, edge tarballs, release notes) via the existing `cargo metadata` cdylib derivation; make missing `.so` a hard failure in tarball packaging
- [ ] extend `LICENSE-binary` generation and the license validation gate to plugin manifests; resolve whatever category-x findings show up
- [ ] measure fat image size, add strip if needed; verify the heavy plugin closures cross-compile under the alpine + zig builder
- [ ] docs: connectors README, docker hub description, quickstart example using the bundled plugins

kriti-sc:
Closes #3614 

## Rationale
Dynamically inject plugins to the `apache/iggy-connect` image to reduce friction for testing out connectors.

## What changed?

This change publishes the image in two flavors:

Fat (edge, x.y.z, latest): the runtime plus every connector plugin baked into /usr/local/lib, which is already on the runtime's plugin search path. A config can load a bundled plugin by file name with no path. This is the new default.
Slim (edge-slim, x.y.z-slim, latest-slim): the runtime binary only, i.e. today's image, for deployments that bring their own plugins.

To this end:
1. Single source of truth for the plugin set. scripts/ci/connector-plugins.sh derives the plugin list from cargo metadata (every cdylib under core/connectors/). The edge tarball build, the Docker fat stage, the license gate, and the release notes all consume it, so a newly added connector is picked up everywhere with no hand-edited list.
2. Dockerfile. The planner derives the plugin set; cook and build compile the plugin cdylibs alongside the runtime; a new runtime-fat stage copies the .so files in. The slim runtime stage is unchanged.
3. Publish pipeline. The docker-buildx action gained a target input, and components can declare flavors in publish.yml. Each flavor keeps its own per-arch digest artifacts so the manifest step never merges flavors into one image, and only the default flavor creates the git release tag.
4. Licensing. The third-party license gate now validates the runtime plus every plugin dependency closure (derived list), so a new connector is license-validated on the PR that adds it. The fat image ships a LICENSE-binary-fat covering the plugins; slim keeps the runtime-only manifest.
5. Edge gate. The connectors image now refreshes :edge on any change under core/connectors/.
6. Docs. Connectors README documents the flavors and zero-config plugin loading with a docker run quickstart; edge release notes derive the plugin list.

## Local Execution

- Built images locally. The fat image is 357 MB, and the build took 20 mins on a 4 CPU/12 GB Docker VM. Slim image is 163 MB. This validates that image will build comfortably on Github CI runners which are 4 CPU/16 GB.
- Pre-commit hooks ran.

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT) Claude
2. Scope of usage? (e.g., autocomplete, generated functions, entire implementation) First developed the design and pass criteria. Then, generated functions and debugging with Claude.
3. How did you verify the generated code works correctly? Built the image locally and checked if all plugins present in the image.


core/connectors/runtime/Dockerfile
```diff
@@ -82,28 +92,40 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARG
...
+    if [ "$PROFILE" = "debug" ]; then PROFILE_DIR=debug; PROFILE_FLAG=""; else PROFILE_DIR=release; PROFILE_FLAG="--release"; fi && \
+    cargo zigbuild --locked --target ${RUST_TARGET} -p iggy-connectors $PLUGIN_FLAGS $PROFILE_FLAG && \
+    cp /app/target/${RUST_TARGET}/${PROFILE_DIR}/iggy-connectors /app/iggy-connectors && \
+    mkdir -p /app/plugins && \
+    while IFS= read -r name; do \
+      [ -z "$name" ] && continue; \
+      so_file="/app/target/${RUST_TARGET}/${PROFILE_DIR}/lib${name}.so"; \
+      if [ ! -f "$so_file" ]; then echo "expected plugin artifact missing: $so_file" >&2 && exit 1; fi; \
+      cp "$so_file" /app/plugins/; \
+    done < connector-plugins.txt
```
numinnex: The plugin compilation lives in the shared `builder` stage, so the slim flavor pays for it too.

`runtime` (slim) copies `/app/iggy-connectors` and `/app/LICENSE-binary` out of `builder`. Both of those are produced by RUN layers that sit at or after this one, so `--target runtime` still executes this `cargo zigbuild -p iggy-connectors $PLUGIN_FLAGS` (17 cdylibs) and the license RUN below that also generates `LICENSE-binary-fat`. BuildKit cannot skip a layer that an earlier needed artifact depends on.

Net effect: the `-slim` build goes from a runtime-binary-only compile to roughly the full fat build time (~20 min per the PR description), plus a cargo-about pass over 18 dependency closures, and then discards all of it. With the matrix at flavors x arches that is two extra full plugin compiles per publish run.

Suggestion: keep `builder` producing only the runtime binary and `LICENSE-binary`, then add a stage that only the fat image consumes:

```dockerfile
FROM builder AS plugin-builder
# cargo zigbuild ... $PLUGIN_FLAGS, collect /app/plugins, generate LICENSE-binary-fat

FROM runtime AS runtime-fat
COPY --from=plugin-builder /app/plugins/ /usr/local/lib/
COPY --from=plugin-builder /app/LICENSE-binary-fat /usr/share/doc/iggy-connect/LICENSE-binary
```

`--target runtime` then stops at `builder` and the slim image build is unchanged from today.

.github/actions/utils/docker-buildx/action.yml
```diff
@@ -342,6 +346,7 @@ runs:
       with:
         context: ${{ steps.ctx.outputs.context }}
         file: ${{ steps.config.outputs.dockerfile }}
+        target: ${{ inputs.target }}
```
numinnex, line 359: `target` now changes the build graph, but the buildx cache ref is still only component- and arch-scoped, so the two flavors fight over one cache entry.

The cache composition step above builds `cache-to` / `cache-from` as `type=registry,ref=${img}:buildcache-${arch},mode=max` (and the shared `${shared_deps}:buildcache-${arch}`) with no flavor component. Since `docker_matrix` is now component x flavor x platform, the fat and slim jobs for the same arch run concurrently and both push `mode=max` to `apache/iggy-connect:buildcache-amd64`. The cache index is last-writer-wins, so each run one flavor's cache export is effectively lost, and `cache-from` may pull an index describing the other flavor's graph.

Layer integrity is fine (BuildKit verifies digests), but the fat build is the expensive one and is exactly the one that needs a warm cache. `${img}:latest` in `cache-from` is also the fat image under the new tag scheme, which the slim build will now pull as inline cache.

Suggestion: thread the flavor into the cache ref, e.g. pass the flavor suffix into this action and use `${img}:buildcache${suffix}-${arch}`, so each flavor keeps its own cache the same way it now keeps its own digest artifacts.

.github/workflows/_build_rust_artifacts.yml
```diff
@@ -238,13 +251,16 @@ jobs:
...
-          IFS=',' read -ra pkgs <<< "$plugins"
+          IFS=',' read -ra pkgs <<< "$CONNECTOR_PLUGINS"
           for pkg in "${pkgs[@]}"; do
             lib_name="$(echo "$pkg" | xargs)"
             [[ -z "$lib_name" ]] && continue
             so_file="target/${target}/release/lib${lib_name}.so"
-            [[ -f "$so_file" ]] && cp "$so_file" "${outdir}/"
+            if [[ ! -f "$so_file" ]]; then
+              echo "::error::expected connector plugin artifact missing: ${so_file}. The plugin crate built no cdylib, or its name does not match lib<crate>.so." >&2
+              exit 1
```
numinnex, line 268: This turns a previously tolerated miss into a hard failure at the same time as the plugin set silently grows, so the first master push after merge is the first real exercise of the new set.

Old behaviour was `[[ -f "$so_file" ]] && cp "$so_file" "${outdir}/"`, a soft skip. Combined with `default: ""` at line 49, the derived list goes from the 10 hardcoded crates to 17: `clickhouse_sink`, `delta_sink`, `doris_sink`, `http_sink`, `influxdb_sink`, `influxdb_source`, `mongodb_sink` are new here. Any of those that does not emit `lib<crate>.so` on `x86_64-unknown-linux-gnu` / `aarch64-unknown-linux-gnu` now fails the edge release rather than being skipped.

The hard failure is the right call, the concern is that it is unverified for the 7 additions. Per the PR description the local validation was the docker image build, which is a different job and a different build graph. `timeout-minutes: 60` is also unchanged for what is now a 17-crate release build on the standard runners, and the tarball grows accordingly (delta/arrow, iceberg, mongodb, surrealdb closures).

Worth running the `Connector plugins` job on this branch before merge to confirm all 17 produce a cdylib on both targets and that it fits in the timeout.

.github/actions/utils/docker-buildx/action.yml
kriti-sc: fixed

.github/workflows/_build_rust_artifacts.yml
kriti-sc: had built docker files locally to check for build time + all plugins being included, but now also setting up the harness to test on gituhb actions.

core/connectors/runtime/Dockerfile
kriti-sc: fixed, but not in the suggested way as then every fat build would recompile the whole plugin set from scratch. instead, split the fat and slim flavours into sibling builders.

.github/workflows/_build_rust_artifacts.yml
kriti-sc: this was successful ^^ build completed within 15 minutes, both cold

target | runner | result | wall clock
-- | -- | -- | --
x86_64-unknown-linux-gnu | ubuntu-latest | success | 12m56s
aarch64-unknown-linux-gnu | ubuntu-24.04-arm | success | 9m34s

<br class="Apple-interchange-newline">

core/connectors/runtime/Dockerfile
```diff
@@ -82,34 +106,121 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARG
...
-# Final runtime - Debian trixie Slim
+# Generate the third-party license manifest covering the runtime plus every
+# bundled plugin closure. Same ASF policy as the slim manifest above; each
+# image ships the manifest that matches what it actually bundles, so the slim
+# image never over-declares plugin dependencies it does not carry.
+#
+RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARGETPLATFORM}-${LIBC} \
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM}-${LIBC} \
+    MANIFEST_FLAGS="$(sed 's/^/--manifest /' connector-manifests.txt | tr '\n' ' ')" && \
+    ./scripts/ci/third-party-licenses.sh --generate --manifest core/connectors/runtime/Cargo.toml $MANIFEST_FLAGS --output /app/LICENSE-binary-fat
```
hubcio, line 206: this LICENSE-binary is missing the runtime's own deps. the union root in third-party-licenses.sh consumes each manifest as a path dependency, and `iggy-connectors` has no lib target, so cargo silently drops it and its whole closure (just a warning, exit 0). measured it: 113 crates linked into the fat image (axum, compio, dlopen2, figment, mimalloc, opentelemetry, tonic, ...) are absent from the manifest it ships. slim is fine - single manifest takes the early-return path. fix in two parts: make the script die when a union manifest lacks a lib target, then run cargo-about per manifest against the real workspace and merge. that also keeps versions lock-exact - the synthetic root re-resolves without the lockfile today, e.g. it lists arrow 58.4.0 while the binaries link 58.3.0.

.github/actions/utils/validate-third-party-licenses/action.yml
```diff
@@ -66,9 +66,17 @@ runs:
...
-    - name: Validate iggy-connect bundle
+    - name: Validate iggy-connect bundle (runtime + connector plugins)
       shell: bash
-      run: ./scripts/ci/third-party-licenses.sh --validate --manifest core/connectors/runtime/Cargo.toml
+      run: |
+        # The fat image bundles the runtime plus every connector plugin, so
+        # validate the plugins' dependency closures too. The manifest list is
+        # derived from cargo metadata so a newly added connector is license
+        # validated on the PR that adds it, with no change here.
+        MANIFEST_FLAGS="$(scripts/ci/connector-plugins.sh --manifest-flags)"
```
hubcio, line 76: same bin-only problem hits here: the union root silently drops `iggy-connectors`, so this step now validates only the plugin closures. the runtime's closure, which the old single-manifest call covered, is no longer license-checked at all - the step got wider on paper and narrower in practice. also worth flipping `read-cache: "false"` above while touching this, this job runs on every PR with a cold cargo home.

.github/workflows/publish.yml
```diff
@@ -1003,24 +1029,29 @@ jobs:
...
       # version step above and already encodes the SNAPSHOT and auto-publish
       # stable-Docker skip rules. dry_run is gated at the job level.
+      # The git release tag is flavor-independent (one tag per component
+      # version), so only the default flavor creates it; other flavors ship
+      # their manifests but must not race to create the same tag.
       - name: Tag Docker release (${{ matrix.key }})
         if: |
           success() &&
           inputs.skip_tag_creation == false &&
+          matrix.flavor_suffix == '' &&
```
hubcio, line 976: this gate creates a bad failure mode for the slim tag: if the slim manifest push fails while fat succeeds, the tag gets created, and every later auto-publish sees it on remote and flips should_tag to false - so `x.y.z-slim` never ships, and even "re-run failed jobs" goes green without pushing it (the re-run recomputes should_tag=false and only refreshes `:edge-slim`). only a manual publish recovers. cleaner shape: make docker-manifests one job per component that reads the flavors from config, downloads all `docker-digest-<key>*` artifacts with `merge-multiple: false`, asserts the digest dirs match the configured flavor set, pushes every flavor's manifest, then tags once at the end. tag only exists when every flavor shipped, and this guard goes away.

core/connectors/runtime/Dockerfile
```diff
@@ -82,34 +106,121 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARG
     "linux/arm64:glibc") RUST_TARGET="aarch64-unknown-linux-gnu" ;; \
     *) echo "Unsupported platform/libc combination: $TARGETPLATFORM/$LIBC" && exit 1 ;; \
     esac && \
+    if [ "$PROFILE" = "debug" ]; then PROFILE_DIR=debug; PROFILE_FLAG=""; else PROFILE_DIR=release; PROFILE_FLAG="--release"; fi && \
+    cargo zigbuild --locked --target ${RUST_TARGET} -p iggy-connectors $PROFILE_FLAG && \
```
hubcio, line 119: switching `--bin iggy-connectors` to `-p iggy-connectors` changes feature resolution, not just what gets built. the old no-`-p` build selected the whole workspace (virtual manifest, no default-members), so features unified workspace-wide. the slim runtime loses `reqwest/http2`, `reqwest/charset` (nine `Response::text()` calls in the http config provider mis-decode non-utf8 responses without it), and tracing-subscriber's `tracing-log` (log-crate records from dependencies silently stop appearing). serde_json map ordering also differs between the two images - fat gets `preserve_order` via bson, slim doesn't - so the same config round-trips with different key order. fat isn't identical to today's build either. suggest pinning `http2` + `charset` on reqwest and `tracing-log` on tracing-subscriber in `[workspace.dependencies]` - the root is the one spot that also covers the tarball plugin build.

.github/actions/utils/docker-buildx/action.yml
hubcio, line 359: `target` is threaded through, but the cache refs aren't: both flavors write `<image>:buildcache-<arch>` with mode=max at the same time, and since publish runs with gha-cache off, the registry is the only cache - last writer wins and the other flavor rebuilds its cook and build layers from scratch every run. thread the flavor into `scope_arch`/`scope_os` and the buildcache ref. leave the shared `apache/iggy:buildcache-*` ref alone, that one is cross-component on purpose.

.dockerignore
```diff
@@ -21,6 +21,7 @@
 /scripts/ci/*
 !/scripts/ci/third-party-licenses.sh
 !/scripts/ci/render-node-licenses.mjs
+!/scripts/ci/connector-plugins.sh
```
hubcio, line 24: the use_latest_ci overlay copies `.github`, `scripts` and the dockerfiles onto the release commit, but never `.dockerignore` - so publishing an older commit with use_latest_ci (defaults to true on manual dispatch) gets master's dockerfile calling connector-plugins.sh while the old dockerignore still excludes it from the build context. the planner stage dies, and both flavors go through it. add `.dockerignore` to the save lists in publish.yml.

core/connectors/runtime/Dockerfile
```diff
@@ -82,34 +106,121 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARG
...
+#
+RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARGETPLATFORM}-${LIBC} \
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM}-${LIBC} \
+    ./scripts/ci/third-party-licenses.sh --generate --manifest core/connectors/runtime/Cargo.toml --output /app/LICENSE-binary
+
+#
+# Fat builder - runtime binary plus every connector plugin cdylib, and a
+# license manifest covering all of them.
+#
+FROM --platform=$BUILDPLATFORM toolchain AS fat-builder
```
hubcio, line 136: nothing in CI builds this stage before a live publish - the docker-buildx action has exactly one caller. so the first-ever 18-plugin lto build runs inside the unchanged 60-minute timeout, and since docker-manifests gates on the whole publish-docker matrix, one fat leg timing out also blocks manifests and git tags for every other component in the release. a cheap pre-merge `--target runtime-fat` build (like the server's runtime-prebuilt smoke build) plus one timed arm64 run before merging would take the guessing out.

core/connectors/runtime/Dockerfile
```diff
@@ -82,34 +106,121 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARG
...
+    PLUGIN_FLAGS="$(sed 's/^/-p /' connector-plugins.txt | tr '\n' ' ')" && \
+    if [ "$PROFILE" = "debug" ]; then PROFILE_DIR=debug; PROFILE_FLAG=""; else PROFILE_DIR=release; PROFILE_FLAG="--release"; fi && \
+    cargo zigbuild --locked --target ${RUST_TARGET} -p iggy-connectors $PLUGIN_FLAGS $PROFILE_FLAG && \
+    cp /app/target/${RUST_TARGET}/${PROFILE_DIR}/iggy-connectors /app/iggy-connectors && \
+    mkdir -p /app/plugins && \
+    while IFS= read -r name; do \
+      [ -z "$name" ] && continue; \
+      so_file="/app/target/${RUST_TARGET}/${PROFILE_DIR}/lib${name}.so"; \
+      if [ ! -f "$so_file" ]; then echo "expected plugin artifact missing: $so_file" >&2 && exit 1; fi; \
+      cp "$so_file" /app/plugins/; \
```
hubcio, line 194: this RUN is plain `sh -c` with no `set -e`, and a while loop's status is the last iteration's - if `cp` fails mid-loop (enospc is realistic here) and a later plugin copies fine, the build goes green and the image ships missing a plugin. the tarball path in this same PR hard-fails on exactly this case. `cp "$so_file" /app/plugins/ || exit 1` closes it.

.github/config/publish.yml
```diff
@@ -96,9 +96,24 @@ components:
...
+      - suffix: "-slim"
+        target: runtime
+    # The fat image bundles the runtime plus every connector plugin, so any
+    # change under core/connectors/ must refresh :edge. crates: keeps
+    # sensitivity to shared workspace deps in the runtime's closure (e.g.
+    # iggy_common); paths: covers all plugin crates without a hardcoded
+    # per-plugin list, so a newly added connector is gated automatically.
     gate:
       crates: [iggy-connectors]
+      paths: [core/connectors]
```
hubcio, line 116: gate misses `scripts/ci/connector-plugins.sh` - its output decides which plugins end up in the image, but editing it won't refresh :edge. web-ui's gate lists its scripts/ci deps for exactly this reason. worth adding `scripts/ci/third-party-licenses.sh` too, and optionally `:(exclude)core/connectors/**/*.md` so docs-only edits stop triggering full image builds (this PR's own README edit trips the gate).

.github/workflows/edge-release.yml
```diff
@@ -59,6 +59,17 @@ jobs:
...
           echo "server_version=${server_version}" >> "$GITHUB_OUTPUT"
 
+      - name: Render bundled connector plugin list
+        id: plugins
+        run: |
+          # Derive the plugin list from cargo metadata so the release notes
+          # stay in sync with what is actually built (see connector-plugins.sh).
+          {
+            echo "list<<PLUGINS_EOF"
+            scripts/ci/connector-plugins.sh --names | sed 's/^/- `/; s/$/`/'
```
hubcio, line 69: two things here. no `shell: bash` on the step means no pipefail, so if the script fails the `sed` still exits 0 and the release notes ship an empty plugin list, green - the only spot in this PR that degrades silently instead of failing. and it's the first cargo invocation in a job with no rust setup, so rustup downloads a toolchain just to print 18 names; a flaky download now kills the edge release. reusing the list the build job already resolved (exposed through the collect job) would fix both.

.github/actions/utils/docker-buildx/action.yml
```diff
@@ -46,6 +46,10 @@ inputs:
     description: "Single platform to build (e.g., linux/amd64). If set, builds only this platform without QEMU. Leave empty for multi-arch build."
     required: false
     default: ""
+  target:
```
hubcio, line 49: nothing records the flavor in image metadata - fat and slim get byte-identical OCI labels, so a pulled `-slim` image can't be told apart from the fat one by inspection. an `org.opencontainers.image.variant` label keyed off the target would fix that.

scripts/ci/connector-plugins.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+case "$MODE" in
+    --names)
+        echo "$NAMES"
+        ;;
+    --comma-names)
+        paste -sd, - <<<"$NAMES"
+        ;;
+    --package-flags)
```
hubcio, line 88: `--package-flags` has no callers anywhere - the dockerfile and the workflow both build their own `-p` flags from `--names`. drop it (it's also the only thing here that needs bash 4's mapfile).

scripts/ci/connector-plugins.sh
```diff
@@ -0,0 +1,91 @@
...
+    | .name
+' <<<"$METADATA" | sort -u)"
+
+if [[ -z "$NAMES" ]]; then
+    echo "connector-plugins: no cdylib plugin crates found under core/connectors/" >&2
+    exit 1
+fi
+
+# Map each crate name back to its repo-relative manifest path.
+manifest_for() {
```
hubcio: this spawns one jq per plugin, each re-parsing the full metadata - a single jq pass can emit the manifests directly. two things to keep if you touch it: keep the `sort` (cargo metadata package order isn't stable, and this output feeds the fat-builder layer cache key), and note the emitted name is used both as `cargo -p` (package name) and `lib<name>.so` (lib target name) - those diverge for hyphenated packages (`bench-report` -> `bench_report`), so either emit both fields or assert they're equal.

core/connectors/runtime/Dockerfile
```diff
@@ -82,34 +106,121 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARG
...
+COPY --from=planner /app/connector-plugins.txt connector-plugins.txt
+COPY --from=planner /app/connector-manifests.txt connector-manifests.txt
+
+#
+# Cook dependencies
+#
+RUN --mount=type=cache,target=/usr/local/cargo/registry,id=cargo-registry-${TARGETPLATFORM}-${LIBC} \
+    --mount=type=cache,target=/usr/local/cargo/git,id=cargo-git-${TARGETPLATFORM}-${LIBC} \
+    --mount=type=cache,target=/app/target,id=cargo-target-fat-${TARGETPLATFORM}-${LIBC} \
+    case "$TARGETPLATFORM:$LIBC" in \
```
hubcio, line 154: this platform-to-triple case block now appears four times in the file. resolving it once at the end of the toolchain stage (write to a file, `RUST_TARGET="$(cat /rust-target)"` in each RUN) keeps the expensive layers platform-independent and leaves one place to add a platform.

.github/workflows/_build_rust_artifacts.yml
```diff
@@ -46,8 +46,8 @@ on:
       connector_plugins:
         type: string
         required: false
-        default: "iggy_connector_elasticsearch_sink,iggy_connector_elasticsearch_source,iggy_connector_iceberg_sink,iggy_connector_postgres_sink,iggy_connector_postgres_source,iggy_connector_quickwit_sink,iggy_connector_random_source,iggy_connector_s3_sink,iggy_connector_stdout_sink,iggy_connector_surrealdb_sink"
-        description: "Comma-separated list of connector plugin crates to build as shared libraries"
+        default: ""
```
hubcio, line 49: emptying the default makes `connector_plugins` unreachable - this workflow is workflow_call-only and the sole caller doesn't pass it, so the override branch below is dead on arrival. simpler to drop the input and the if/else and always derive; that also removes the one way the tarball's plugin set could disagree with the fat image's.

scripts/ci/connector-plugins.sh
```diff
@@ -0,0 +1,91 @@
...
+fi
+
+METADATA="$(cargo metadata --format-version 1 --no-deps)"
+WORKSPACE_ROOT="$(jq -r '.workspace_root' <<<"$METADATA")"
+
+# cdylib packages whose manifest lives under core/connectors/. The path guard
+# keeps a future non-connector cdylib elsewhere in the workspace out of the set.
+NAMES="$(jq -r --arg root "$WORKSPACE_ROOT" '
+    .packages[]
+    | select(.manifest_path | startswith($root + "/core/connectors/"))
```
hubcio, line 59: with this guard, any future cdylib under core/connectors/ (a test fixture, an example) auto-ships in the default image and the edge tarball. all 18 names match `iggy_connector_*` - a name-prefix filter would bound the set tighter than the directory does.

.github/workflows/publish.yml
```diff
@@ -399,21 +399,38 @@ jobs:
...
+              return Array.isArray(fl) && fl.length ? fl : [{ suffix: '', target: '' }];
+            };
+
             const dockerMatrix = [];
+            const dockerComponents = [];
             for (const t of dockerTargets) {
-              for (const p of platforms) {
-                dockerMatrix.push({ ...t, ...p });
+              for (const f of flavorsOf(t.key)) {
+                const flavor = { flavor_suffix: f.suffix || '', flavor_target: f.target || '' };
```
hubcio, line 344: `f.target || ''` means a flavor entry with a suffix but no target silently builds the default (last) stage - for this dockerfile that's the fat image published under the `-slim` tag. worth a small validation in the plan script (or moot if the per-component manifest job shape lands).

.github/config/publish.yml
```diff
@@ -96,9 +96,24 @@ components:
...
     version_regex: '(?m)^\s*version\s*=\s*"([^"]+)"'
-    # Image ships only the runtime binary; plugin .so files are not bundled.
+    # Two published flavors from one Dockerfile. The fat flavor (empty suffix,
+    # runtime-fat stage) bundles every connector plugin and is the default;
+    # -slim (runtime stage) ships the runtime binary only. Each flavor keeps
+    # its own per-arch digest set and manifest tags. Components without a
+    # flavors list build their single default stage.
+    flavors:
+      - suffix: ""
+        target: runtime-fat
```
hubcio, line 106: tiny naming thing: `target` here means dockerfile stage, but everywhere else in this repo's CI `target` means the rust triple. `stage` would read clearer (the docker-buildx input should stay `target` since it matches build-push-action).

scripts/ci/connector-plugins.sh
```diff
@@ -0,0 +1,91 @@
...
+set -euo pipefail
+
+# Single source of truth for the connector plugin set: every workspace crate
+# under core/connectors/ that builds a cdylib (a .so the iggy-connectors
+# runtime loads via dlopen). Derived from `cargo metadata` so a newly added
+# connector is picked up automatically by every consumer:
+#
+#   - .github/workflows/_build_rust_artifacts.yml  (edge tarball plugin list)
+#   - core/connectors/runtime/Dockerfile           (fat image build + bundle)
+#   - .github/actions/utils/validate-third-party-licenses  (license gate)
```
hubcio, line 28: the consumer list is missing one - the pre-merge rust action derives an overlapping set inline (bin-or-cdylib, no path guard), so "single source of truth" oversells it a bit. worth a cross-reference comment there or a mode it can call.

hubcio (changes requested):
the direction is right - deriving the plugin set from cargo metadata kills a list that was already stale (the hardcoded default shipped 10 of 18 plugins). but there are a few things that ship wrong artifacts, details inline. two notes without a diff line to hang them on:

- the publish job name doesn't include the flavor, so `Docker rust-connectors (amd64)` shows up twice and `Docker manifests` is one static name for everything - when a flavor leg fails you can't tell which one from the run list. goes away with the per-component manifest job suggested inline.
- the sinks/sources lists in core/connectors/README.md document 9 sinks + 3 sources but the workspace has 14 + 4. with the image now advertising every plugin bundled, the missing ones (clickhouse, delta, http, mongodb, and the influxdb pair) are worth adding.

mmmmxa:
Hi everyone! I was doing the research on the delta sink connector with the help of this PR and so far I am quite happy with the results. I wrote a compose file based on the code from this PR and did many experiments -- I have nothing to complain about and everything works. For my own developments I will find a way to integrate this temporarily, but with no pressure I'd like to mention that this feature is highly desirable as it makes the production testing substantially more straightforward. Thanks @kriti-sc for your hard work, this is very useful. :muscle: :rocket:

mmmmxa:
@hubcio I can take over from here since I'm somewhat familiar with this code and topic.

hubcio:
@mmmmxa sure thing!
please open a fresh PR against master (refs #3614) and keep @kriti-sc as co-author. (open PR with my comments fixed please).

PR #3667, opened 2026-07-13
feat(server-ng): replica rejoin via view probe and message repair

numinnex:
This PR adds the message repair protocol. Additionally it fixes couple of timeout issues during the view change (backup not receiving the `RequestStartView` was stuck forever).  

The message repair protocol streams range of prepares, rather than asking for individual prepare by it's `op` and `checksum`.  The `RequestStartView` is also used for probing that triggers the message repair protocol and state transfer in the future.


core/server_common/src/consensus_message.rs
```diff
@@ -568,6 +592,22 @@ where
...
+                value.try_into_typed::<RequestPreparesHeader>()?,
+            )),
+            // A repaired prepare is a stored PrepareHeader frame whose command
+            // byte was rewritten; typed validation would reject the byte, so
+            // parse through the generic backing and trust the prepare-shaped
+            // layout the way the journal that produced it did.
+            Command2::RepairPrepare => Ok(Self::RepairPrepare(
+                value.try_into_typed::<RepairPrepareHeader>()?,
+            )),
+            Command2::RepairDone | Command2::RangeEvicted => Ok(Self::RepairRangeReply(
```
hubcio, line 608: `RangeEvicted` never makes it through this parse. `try_into_typed` rejects any frame whose command byte differs from `T::COMMAND` (line 229, locked by `try_into_typed_command_mismatch_returns_err`), and `RepairRangeReplyHeader::COMMAND` is `RepairDone` -- so a `RangeEvicted` frame (byte 21) errors here before `validate()` (which does accept both) ever runs, and the router drops it with "dropping message with invalid command".

both `on_repair_range_reply` `RangeEvicted` arms are dead code, `RepairSession.floor` can never be set, and the commit-floor path -- the mechanism this PR relies on when the rejoin window exceeds the 4096-entry evicted ring -- never engages: the requester journals only the suffix, `complete_repair` runs with no floor, and the commit walk gap-stops with `commit_min` frozen below `commit_max` forever. no enabled vsr test sends more than 4096 ops, so nothing exercises `RangeEvicted` delivery end to end.

fix idea: give `ConsensusHeader` an `accepts(cmd)` hook defaulting to `cmd == Self::COMMAND` and override for this header (or split it into two header types), plus a `MessageBag` round-trip test for commands 17-21 and a >4096-op rejoin test.

core/shard/src/lib.rs
```diff
@@ -1775,6 +1824,36 @@ where
         {
             partition.commit_journal(config).await;
         }
+        // Same gap-fill as the metadata arm: a journal-less rejoiner that
+        // adopted the new view still lacks the window's entries; repair from
+        // the announcing primary, floor settled by its RangeEvicted.
+        let consensus = partition.consensus();
+        if consensus.is_normal()
+            && consensus.commit_min() < consensus.commit_max()
+            && partition.repair.is_none()
```
hubcio, line 1851: repair has no retry or stall recovery: the session is armed only here under `repair.is_none()` (same for `metadata_repair` above) and cleared only by a nonce-matched `RepairDone`. all repair sends are fire-and-forget `let _ = send_to_replica(...)`, the serving peer returns silently without a terminator when it's not `Normal`, and the requester accepts `RepairDone` without checking `header.op >= session.to_op`. a single lost frame (or a peer mid-view-change) leaves the session stuck `Some` forever -- later `StartView`s can't re-arm it, and the replica sits `Normal` with `commit_min < commit_max` until a process restart.

the view probe got a retry timer; repair needs the same: re-request from the served frontier via `session.peer` (currently written but never read -- its doc even promises "re-request target on stall").

core/shard/src/lib.rs
```diff
@@ -1830,6 +1909,421 @@ where
...
+            .await;
+            from_op = retained_from;
+        }
+        let mut served_through = from_op.saturating_sub(1);
+        for op in from_op..=to_op {
+            let Some(entry) = partition.log.journal().inner.repair_entry(op) else {
+                break;
+            };
+            self.send_repair_prepare(target, entry).await;
+            served_through = op;
```
hubcio, line 2115: also at line 2010 (metadata arm). `served_through` advances after a fire-and-forget send -- `send_repair_prepare` swallows both bus errors and framing errors, and `send_to_replica` is a non-blocking try-send, so under a queue-full burst op N can be dropped while N+1 lands. the final `RepairDone` then reports ops that were never delivered and the requester treats the stream as complete. don't advance `served_through` when the send fails.

core/partitions/src/iggy_partition.rs
```diff
@@ -2936,6 +3014,119 @@ where
...
+                namespace_raw = self.namespace().inner(),
+                op = header.op,
+                %error,
+                "failed to journal repaired prepare"
+            );
+            return;
+        }
+        let consensus = self.consensus();
+        if header.op > consensus.sequencer().current_sequence() {
+            consensus.sequencer().set_sequence(header.op);
```
hubcio: bumping the sequencer to a repaired op that sits above an unfilled gap makes the rejoiner an unsafe election candidate: DVC advertises `op = sequencer.current_sequence()` and the winner is picked by `max(log_view, op)` among the collected quorum, so a replica whose repair stalled (or whose floor never arrived) can win an election with an interior hole and lead a log it can't walk. the `Recovering` guards don't help here -- after `StartView` adoption it's `Normal`. the sequencer should track the contiguous frontier rather than the max journaled op, or the replica should stay election-ineligible while its log is non-contiguous.

core/consensus/src/impls.rs
```diff
@@ -1766,6 +1839,103 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+    /// Set the commit floor after journal repair filled `(floor, commit_max]`:
+    /// everything at or below `floor` is represented by this replica's
+    /// recovered durable state (segments + offset files), proven by the
+    /// serving peer answering `RangeEvicted { retained_from = floor + 1 }`.
+    /// Unlike the retired first-commit fast-forward, ops in the repair window
+    /// are journaled and WALKED, never skipped.
+    ///
+    /// # Panics
+    /// Panics when `floor` would rewind the already-executed `commit_min`.
+    pub fn set_commit_floor(&self, floor: u64) {
```
hubcio, line 1937: two missing bounds here, and the assert that does exist is unreachable (the sole caller, `complete_repair`, already checks `floor > commit_min`). not guarded:

1. `floor <= commit_max` -- the partition serve arm sends `RangeEvicted { retained_from }` unclamped, so a peer that evicted past this replica's `commit_max` drives `commit_min > commit_max`, and the next `advance_commit_max` assert panics the shard.
2. floor vs this replica's actual recovered coverage -- `retained_from` tracks the *peer's* flush frontier, so if this replica's flush lagged the peer by more than the ring capacity before the crash, the floor jumps past ops that are neither locally durable nor repaired (they sit below the floor). silent offset hole on read, no loud guard anywhere.

clamp to `min(commit_max, local durable coverage)` and refuse loudly above coverage -- that case is state-transfer territory.

core/shard/src/lib.rs
```diff
@@ -1830,6 +1909,421 @@ where
...
+        if !partition.consensus().is_normal() {
+            return;
+        }
+        let cluster = partition.consensus().cluster();
+        let self_id = partition.consensus().replica();
+        let to_op = header.to_op.min(partition.consensus().commit_max());
+        let retained_from = partition.log.journal().inner.repair_retained_from();
+        let mut from_op = header.from_op;
+        if let Some(retained_from) = retained_from
+            && retained_from > from_op
```
hubcio, line 2092: `retained_from` goes out unclamped -- the metadata arm above bounds its skip-front by `to_op`, this one doesn't. a peer that evicted past the requester's `commit_max` produces `floor > commit_max` on the receiver (see `set_commit_floor`, which has no ceiling). clamp here or at the floor.

core/shard/src/lib.rs
```diff
@@ -1830,6 +1909,421 @@ where
...
+            let journal = journal.handle();
+            let cluster = consensus.cluster();
+            let self_id = consensus.replica();
+            let to_op = header.to_op.min(consensus.commit_max());
+            // Skip the compacted prefix (below the snapshot floor) in one
+            // RangeEvicted notice, then serve contiguously until the range
+            // ends or the WAL runs out.
+            let mut from_op = header.from_op;
+            #[allow(clippy::cast_possible_truncation)]
+            while from_op <= to_op && journal.header(from_op as usize).is_none() {
```
hubcio, line 2004: when the entire requested range is missing, this loop walks `from_op` to `to_op + 1`, the `from_op <= to_op` guard below suppresses `RangeEvicted`, and `served_through` lands on `to_op` -- so the peer answers `RepairDone(to_op)` claiming full coverage while serving zero prepares, and the requester clears its session and gap-stops silently. latent today (the metadata WAL doesn't evict yet), but whole-range-missing should answer `RangeEvicted`, not a false `RepairDone`.

core/partitions/src/journal.rs
```diff
@@ -169,8 +169,19 @@ where
...
+    /// entries left the resident journal at flush. Bounded by
+    /// [`EVICTED_RING_CAPACITY`]; requests older than the ring answer
+    /// `RangeEvicted` honestly.
+    evicted_ring: UnsafeCell<VecDeque<(u64, JournalBuffer)>>,
 }
 
+/// How many evicted entries each partition retains for repair. Sized to
+/// cover a few seconds of traffic around a node restart; anything older is
+/// bulk-sync (phase 3) territory.
+pub const EVICTED_RING_CAPACITY: usize = 4096;
```
hubcio, line 188: the ring bound is entry count, not bytes -- each partition pins up to 4096 owned `JournalBuffer`s at steady state (`commit()` clears every index but never the ring), so worst-case resident memory is 4096 x batch size x partition count, and these buffers used to be freed at flush. it's also populated unconditionally, including single-replica deployments where repair can never run. worth a byte cap or a config knob, and skipping the ring entirely when `replica_count == 1`.

core/partitions/src/types.rs
```diff
@@ -207,6 +207,21 @@ impl Default for PartitionOffsets {
 ///
 /// Mirrors the relevant fields from the server's `PartitionConfig` and
 /// `SegmentConfig` (`core/server/src/configs/system.rs`).
+/// One in-flight journal-repair stream for a partition group.
```
hubcio, line 214: `RepairSession` landed between `PartitionsConfig`'s doc block and the struct -- "Configuration for partition operations / Mirrors the relevant fields..." now documents `RepairSession`, and `PartitionsConfig` is left undocumented. move this above the doc block.

core/consensus/src/impls.rs
```diff
@@ -1766,6 +1839,103 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+            header.namespace, self.namespace,
+            "RequestStartView routed to wrong group"
+        );
+        if self.status.get() != Status::Normal {
+            return Vec::new();
+        }
+        if header.replica == self.replica {
+            return Vec::new();
+        }
+        if self.primary_index(self.view.get()) == header.replica {
```
hubcio, line 1899: a delayed or duplicated probe re-triggers this election: `header.view` is never checked (senders don't even stamp it -- the field goes out as 0), and probes are re-broadcast on a timer so duplicates are the normal case. since `primary_index = view % replica_count`, a stale probe from replica R re-matches every `replica_count` views. bounded (vsr self-heals), but it's a free spurious election. stamp `view` on the probe and require `header.view == current` before electing past the primary.

core/consensus/src/impls.rs
```diff
@@ -1766,6 +1839,103 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
         actions
     }
 
+    /// Begin the view probe: broadcast `RequestStartView` and keep
+    /// re-broadcasting on `TimeoutKind::RequestStartViewMessage` until the
+    /// current view's primary answers with a targeted `StartView` (or an
+    /// election's `StartView` adopts this replica first). The replica sits
+    /// in `Status::Recovering` meanwhile: it acks nothing, votes in no
+    /// election, and initiates nothing.
+    pub fn begin_view_probe(&self) -> Vec<VsrAction> {
```
hubcio: the returned actions are discarded at all three call sites (`let _boot_probe = ...` in bootstrap and `build_partition_fresh`), so the first probe actually goes out on the `RequestStartViewMessage` timeout (~1s later) -- which is fine, the mesh isn't up at boot anyway. just return `()` so the contract matches reality; dispatching at the call sites would be the wrong fix.

hubcio (dismissed):

core/partitions/src/iggy_partition.rs
```diff
@@ -2936,6 +3049,211 @@ where
...
+            // that gap is state-transfer territory, and accepting the floor
+            // would silently serve a holed log. Refuse and stay gap-stopped:
+            // a visible stall beats invisible loss.
+            let durable_end = self.recovered_durable_offset;
+            let connected = match (session.first_batch_offset, durable_end) {
+                (Some(first), Some(durable)) => first <= durable.saturating_add(1),
+                (Some(first), None) => first == 0,
+                // Offsets-only window: nothing to reconcile against segments;
+                // the consumer-offset table on disk stands in below the floor.
+                (None, _) => true,
```
krishvishal: `first_batch_offset` is written only in the `SendMessages` arm of apply_repaired_prepare, so a repaired window that delivers zero message batches leaves it `None, (None, _) => true` here punts to connected, and `set_commit_floor` advances `commit_min` past ops
  that are neither locally durable nor repaired -- the same silent offset hole the guard closes, just narrower. reachable when the window is all consumer-offset ops above an evicted message range, or when the lossy bus drops every `RepairPrepare` but
  `RangeEvicted` + `RepairDone` land. the local-walk check doesn't catch it: the floor jump makes `commit_min >= to_op`, so the walk short-circuits as done. None shouldn't mean connected -- with no batch there's no offset anchor, so refuse (gap-stop -> state
  transfer, like the disconnected arm) whenever a message range sits below the floor; only a genuinely offsets-only window is safe. also untested -- the >4096 rejoin test is built so the window connects, so neither this arm nor the refusal branch runs.

mmodzelewski (approved):

spetz (approved):

PR #3669, opened 2026-07-14
fix(connectors): preserve transformed payload schema across FFI

Standing-Man:
## Problem
The runtime previously populated both `MessagesMetadata.schema` and `RawMessages.schema` from `decoder.schema()`.

`decoder.schema()` describes the input format configured for the Iggy stream. It remains unchanged for the lifetime of the consumer. However, format-conversion transforms can replace the decoded `Payload` with a different
variant.

For example, consider a sink configured with a raw input stream and a transform that converts raw bytes to text:
```text
Iggy message bytes
    |
    | decode with Schema::Raw
    v
Payload::Raw(b"hello")
    |
    | Raw-to-Text transform
    v
Payload::Text("hello")
    |
    | Payload::try_into_vec()
    v
b"hello"
    |
    | FFI metadata incorrectly contains Schema::Raw
    v
Payload::Raw(b"hello") received by the sink plugin
```
Although the transform produced `Payload::Text` , converting it back into bytes removed the Rust enum variant information. The only information available to the plugin for reconstructing the payload was the schema in the FFI metadata.
Because the runtime still sent `Schema::Raw`, the plugin reconstructed the
transformed bytes as `Payload::Raw`. The transform appeared to succeed inside
the runtime, but its output type was lost when the message crossed the FFI
boundary.

The same problem is more serious for schema-sensitive conversions. For example:
```
Payload::Proto
    -> Protobuf-to-JSON transform
    -> JSON bytes
    -> FFI metadata still contains Schema::Proto
    -> plugin attempts to interpret JSON bytes as Protobuf
```

Depending on the decoder and sink implementation, this could result in: 
- the sink receiving a different Payload variant than the transform produced;
- transformed data falling back to Payload::Raw;
- payload decoding errors;
- format-specific sink validation rejecting the message;
- messages being skipped even though the transform itself succeeded.

The transformed bytes were serialized correctly. The problem was that the runtime attached stale type information to those bytes.

## Root cause
The runtime treated the configured input schema as if it were also the output
schema of the transform chain:
```
RawMessages {
    schema: decoder.schema(),
    messages,
}
```
That assumption only holds for transforms that preserve the payload format. It is invalid for ProtoConvert, AvroConvert, FlatBufferConvert, and any other transform that changes the Payload variant. The runtime already had the correct type information in the transformed Payload, but it discarded that information when calling `Payload::try_into_vec()`.

## Rationale

The sink runtime decodes each Iggy message according to the schema configured
for the input stream and represents the decoded value as a `Payload`. It then
applies the configured transform chain before serializing the payload and
passing the batch to the sink plugin over FFI.

Format-conversion transforms can change the `Payload` variant. For example, a
message decoded as `Payload::Raw` can become `Payload::Text`, or a Protobuf
payload can become JSON. Before this change, the runtime serialized the
transformed payload bytes but still attached `decoder.schema()`, which
describes the input bytes before transformation.

The FFI boundary only carries serialized bytes and schema metadata. On the
plugin side, `SinkContainer::consume` reconstructs each `Payload` by calling
`MessagesMetadata::schema.try_into_payload`. A stale input schema therefore
caused the transformed bytes to be interpreted as the wrong payload type. A
Raw-to-Text transform could arrive as `Payload::Raw`, while format-sensitive
conversions such as Protobuf-to-JSON could be rejected or decoded incorrectly.

## What changed?

`Payload::schema()` now returns the schema represented by the current payload
variant. The sink runtime reads that schema after all transforms have completed
and before the payload is converted back into bytes.

The transformed output schema is written to both existing FFI metadata
containers:

- `MessagesMetadata.schema`, which the sink SDK uses to reconstruct payloads.
- `RawMessages.schema`, which keeps the serialized batch metadata consistent.

A sink batch must have one output schema because the existing FFI batch format
stores schema once per batch. If a transformed message has a different schema
from the messages already accepted into the batch, the runtime logs the
mismatch, increments the connector error counter, and skips that message rather
than decoding it later with an incompatible schema.

If every message is filtered or rejected, the batch retains the original input
schema as a safe fallback.

This keeps the fix compatible with existing plugins. It does not change FFI
function signatures, `#[repr(C)]` layouts, postcard payload structures, schema
discriminants, or connector configuration.

A regression test exercises the complete runtime-to-plugin boundary with a
Raw-to-Text transform. The FFI callback deserializes both metadata structures
and verifies that they carry `Schema::Text`, that the transformed bytes are
preserved, and that the batch reports one processed message.

## Local Execution

- Passed / not passed
passed
ran

## AI Usage

1. Tool: Codex
2. Scope: Root-cause analysis, implementation, regression test, and PR wording.
3. Verification: The affected SDK and runtime crates passed formatting, Clippy,
   unit tests, integration-style SDK tests, and representative plugin builds.
4. I reviewed the changes and can explain every modified line.

### DIFF
--- a/core/connectors/runtime/src/sink.rs
+++ b/core/connectors/runtime/src/sink.rs
@@ -552,7 +552,7 @@ pub(crate) async fn setup_sink_consumers(
 #[allow(clippy::too_many_arguments)]
 async fn process_messages(
     plugin_id: u32,
-    mut messages_metadata: MessagesMetadata,
+    messages_metadata: MessagesMetadata,
     topic_metadata: &TopicMetadata,
     messages: Vec<IggyMessage>,
     consume: &ConsumeCallback,
@@ -602,10 +602,10 @@ async fn process_messages(
     }
     let decode_elapsed = decode_start.elapsed();
 
-    let mut messages = Vec::with_capacity(decoded.len());
-    let mut output_schema = None;
+    let mut batches = Vec::<RawMessages>::new();
     for message in decoded {
         let mut current_message = Some(message);
+        let mut transform_failed = false;
         for transform in transforms.iter() {
             let Some(message) = current_message.take() else {
                 break;
@@ -622,10 +622,14 @@ async fn process_messages(
                         topic_metadata.topic
                     );
                     error_count += 1;
-                    current_message = None;
+                    transform_failed = true;
                     break;
                 }
             }
+        }
+
+        if transform_failed {
+            continue;
         }
 
         // Filter contract: transform returning Ok(None) is an intentional drop.
@@ -697,16 +701,7 @@ async fn process_messages(
             None => vec![],
         };
 
-        if output_schema.is_some_and(|schema| schema != payload_schema) {
-            error!(
-                "Transformed payload schema {payload_schema} does not match the other messages in the batch for message with ID: {id}, offset: {offset}, sink connector with ID: {plugin_id}"
-            );
-            error_count += 1;
-            continue;
-        }
-        output_schema = Some(payload_schema);
-
-        messages.push(RawMessage {
+        let raw_message = RawMessage {
             id,
             offset,
             checksum,
@@ -714,7 +709,14 @@ async fn process_messages(
             origin_timestamp,
             headers,
             payload,
-        });
+        };
+        match batches.last_mut() {
+            Some(batch) if batch.schema == payload_schema => batch.messages.push(raw_message),
+            _ => batches.push(RawMessages {
+                schema: payload_schema,
+                messages: vec![raw_message],
+            }),
+        }
     }
 
     metrics.inc_errors_by_with_labels(&labels.counter, error_count);
@@ -722,9 +724,13 @@ async fn process_messages(
         metrics.inc_messages_filtered_with_labels(&labels.counter, filtered_count);
     }
 
-    let processed_count = messages.len();
-    let output_schema = output_schema.unwrap_or(messages_metadata.schema);
-    messages_metadata.schema = output_schema;
+    let processed_count = batches.iter().map(|batch| batch.messages.len()).sum();
+    if batches.is_empty() {
+        batches.push(RawMessages {
+            schema: messages_metadata.schema,
+            messages: Vec::new(),
+        });
+    }
 
     let topic_meta = postcard::to_allocvec(topic_metadata).map_err(|error| {
         error!(
@@ -733,33 +739,36 @@ async fn process_messages(
         RuntimeError::FailedToSerializeTopicMetadata
     })?;
 
-    let messages_meta = postcard::to_allocvec(&messages_metadata).map_err(|error| {
-        error!(
-            "Failed to serialize messages metadata for sink connector with ID: {plugin_id}. {error}"
+    let mut ffi_elapsed = Duration::ZERO;
+    for batch in batches {
+        let messages_metadata = MessagesMetadata {
+            partition_id: messages_metadata.partition_id,
+            current_offset: messages_metadata.current_offset,
+            schema: batch.schema,
+        };
+        let messages_meta = postcard::to_allocvec(&messages_metadata).map_err(|error| {
+            error!(
+                "Failed to serialize messages metadata for sink connector with ID: {plugin_id}. {error}"
+            );
+            RuntimeError::FailedToSerializeMessagesMetadata
+        })?;
+        let messages = postcard::to_allocvec(&batch).map_err(|error| {
+            error!("Failed to serialize messages for sink connector with ID: {plugin_id}. {error}");
+            RuntimeError::FailedToSerializeRawMessages
+        })?;
+
+        let ffi_start = Instant::now();
+        (consume)(
+            plugin_id,
+            topic_meta.as_ptr(),
+            topic_meta.len(),
+            messages_meta.as_ptr(),
+            messages_meta.len(),
+            messages.as_ptr(),
+            messages.len(),
         );
-        RuntimeError::FailedToSerializeMessagesMetadata
-    })?;
-
-    let messages = postcard::to_allocvec(&RawMessages {
-        schema: output_schema,
-        messages,
-    })
-    .map_err(|error| {
-        error!("Failed to serialize messages for sink connector with ID: {plugin_id}. {error}");
-        RuntimeError::FailedToSerializeRawMessages
-    })?;
-
-    let ffi_start = Instant::now();
-    (consume)(
-        plugin_id,
-        topic_meta.as_ptr(),
-        topic_meta.len(),
-        messages_meta.as_ptr(),
-        messages_meta.len(),
-        messages.as_ptr(),
-        messages.len(),
-    );
-    let ffi_elapsed = ffi_start.elapsed();
+        ffi_elapsed += ffi_start.elapsed();
+    }
 
     Ok(SinkBatchTiming {
         processed_count,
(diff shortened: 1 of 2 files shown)


Standing-Man:
Hi @hubcio, could you please take a look when you have chance? Thanks.

ryerraguntla:
@Standing-Man - Could you tell issue it is trying to solve and any related issue?

Standing-Man:
> Could you tell issue it is trying to solve and any related issue?

I’ve added more details to the PR description to better explain the problem this PR addresses.

/request-review @ryerraguntla

core/connectors/runtime/src/sink.rs
```diff
@@ -695,6 +697,15 @@ async fn process_messages(
             None => vec![],
         };
 
+        if output_schema.is_some_and(|schema| schema != payload_schema) {
```
hubcio: this drop path is reachable with in-tree transforms: `ProtoConvert` with source json, target proto, a descriptor and `message_type` set emits `Payload::Raw` when `encode_json_with_schema` succeeds and falls back to `Payload::Proto` when it fails (top-level json not an object). one batch mixing json objects with arrays/scalars produces mixed variants, and every message whose schema differs from the first surviving one gets dropped here - first-schema-wins, so a leading minority can drop the majority. with `AutoCommit::When(PollingMessages)` the offset is already committed, so the drop is permanent.

not a regression - pre-PR the same batch shipped under one stale label and was lost/corrupted plugin-side anyway - but this path has zero test coverage. worth a heterogeneous-batch test now, and longer-term splitting the batch per schema (one `consume()` per group) instead of dropping.

hubcio (changes requested):

core/connectors/runtime/src/sink.rs
ryerraguntla: I concur on test coverage, needs test coverage as much as possible

core/connectors/runtime/src/sink.rs
```diff
@@ -760,3 +773,119 @@ struct SinkBatchTiming {
...
+                    .messages
+                    .into_iter()
+                    .map(|message| message.payload)
+                    .collect(),
+            });
+        }
+        0
+    }
+
+    #[tokio::test]
```
ryerraguntla, line 1036: only homogeneous Raw→Text FFI test; no mixed-schema path. **Fix**: hetero-batch test (first-wins keep, mismatch error_count, FFI schema).

core/connectors/runtime/src/sink.rs
```diff
@@ -673,6 +674,7 @@ async fn process_messages(
             continue;
         };
 
+        let payload_schema = message.payload.schema();
```
ryerraguntla, line 681: sink.rs:677-706 — try_into_vec/headers before schema check; waste on mismatch. **Fix**: compare schema first.

core/connectors/runtime/src/sink.rs
Standing-Man: Fixed. `process_messages` no longer applies a first-schema-wins policy. After transforms, messages are grouped into contiguous schema-homogeneous sub-batches, and each group is passed to `consume()` separately with matching `MessagesMetadata::schema` and `RawMessages::schema`. 

I also added coverage through the SDK SinkContainer, so the test exercises postcard serialization, the FFI callback, SDK-side payload reconstruction, and the final `Sink::consume()` call. Additionally, transform errors are now counted only as errors; only `Ok(None)` increments the filtered-message metric. 

cc @hubcio and @ryerraguntla.

Standing-Man:
There are a few trade-offs to splitting a polled batch by transformed payload schema:
1. The implementation groups only contiguous messages with the same schema. This preserves the original message order. Grouping all messages globally by schema would reduce `consume()` calls but could reorder messages.
2. One broker batch may now result in multiple `consume()` calls. Sink implementations must therefore not assume a one-to-one relationship between a poll and a consume invocation.
3. The sub-batches are not atomic. If one consume() call succeeds and another fails, the sink may observe only part of the original broker batch.
4. All sub-batches currently inherit the original current_offset. Per-message offsets remain accurate, but sinks using only the batch-level offset should account for multiple calls carrying the same value. 

cc @hubcio and @ryerraguntla.

core/connectors/runtime/src/sink.rs
Standing-Man: This mismatch path no longer exists. Messages with different transformed schemas are now preserved and placed into separate contiguous schema-homogeneous sub-batches, so payload and header serialization is required for every surviving message.

core/connectors/runtime/src/sink.rs
```diff
@@ -720,33 +739,36 @@ async fn process_messages(
...
+            plugin_id,
+            topic_meta.as_ptr(),
+            topic_meta.len(),
+            messages_meta.as_ptr(),
+            messages_meta.len(),
+            messages.as_ptr(),
+            messages.len(),
+        );
+        ffi_elapsed += ffi_start.elapsed();
+    }
```
numinnex, line 771: This turns one poll into N `consume()` invocations, but the plugin-facing contract still says otherwise. `Sink::consume` in `core/connectors/sdk/src/lib.rs` documents it as "Invoked every time a batch of messages is received from the configured stream(s) and topic(s)", and `core/connectors/sinks/README.md` still describes the delivered format as fixed by the stream config.

The four trade-offs you listed in the thread are the contract now: `current_offset` repeated across sub-batches, no atomicity between them, no 1:1 poll-to-consume relationship. Out-of-tree plugins using `(partition_id, current_offset)` as an idempotency or dedup key will silently discard sub-batches 2..N. I went through the in-tree sinks and none of them break - all uses of `current_offset` are log-only and `s3_sink` keys off per-message offsets while buffering across calls - but there is no line anywhere in the tree that tells a plugin author this changed.

Please move those trade-offs into the `Sink::consume` doc comment and `sinks/README.md`.

Unrelated to the loop, but same theme: the transform-error accounting change in this PR (an error no longer also bumping `iggy_connector_messages_filtered_total`) is correct and brings the sink in line with `source.rs`, which already has the identical `transform_failed` flag. It also changes the meaning of an exported metric, and right now that only exists in a PR comment. Worth a line in the commit body.

core/connectors/runtime/src/sink.rs
```diff
@@ -695,23 +701,36 @@ async fn process_messages(
...
     }
 
-    let processed_count = messages.len();
+    let processed_count = batches.iter().map(|batch| batch.messages.len()).sum();
+    if batches.is_empty() {
+        batches.push(RawMessages {
+            schema: messages_metadata.schema,
+            messages: Vec::new(),
+        });
+    }
```
numinnex, line 733: Nothing reports how many sub-batches a poll produced.

Grouping is contiguous, so the worst case is one `consume()` per message. With the default `batch_length = 1000` and a stream that alternates schemas - e.g. `ProtoConvert` encoding JSON objects successfully and falling back on arrays/scalars - a single poll becomes up to 1000 sequential FFI calls, each one an HTTP request or DB transaction on the plugin side, with no await point in this loop.

I'm not asking for a cap; a cap would reintroduce the drops this PR removes. But the fan-out has to be visible, otherwise the failure mode is a latency collapse with nothing pointing at sub-batching. `batch_count` alongside `processed_count` in `benchmark::emit_sink_event`, plus a `debug!` when `batches.len() > 1`, would be enough.

core/connectors/runtime/src/sink.rs
```diff
@@ -760,3 +782,470 @@ struct SinkBatchTiming {
...
+            0
+        );
+        assert_eq!(captured.len(), 4);
+        assert_eq!(
+            captured
+                .iter()
+                .map(|batch| batch.metadata_schema)
+                .collect::<Vec<_>>(),
+            vec![Schema::Proto, Schema::Raw, Schema::Proto, Schema::Raw]
+        );
```
numinnex, line 1121: This assertion locks in a mislabel that lives in `ProtoConvert::json_to_protobuf`: successful protobuf encoding returns `Payload::Raw(binary)` while the fallback returns `Payload::Proto(json_string)` - inverted. The PR forwards both faithfully, which is the right behaviour for this layer, but the two consequences are not equivalent:

- Success path (`Schema::Raw` in this vector) is a genuine fix. Pre-PR this shipped `Schema::Json`, the SDK ran simd_json over protobuf binary, and the message was dropped plugin-side.
- Fallback path (`Schema::Proto` in this vector) is a regression. The SDK's `Schema::Proto` arm tries `prost_types::Any::decode`, fails on JSON text, and yields `Payload::Raw`; pre-PR the plugin received `Payload::Json`. `quickwit_sink` skips non-`Json` payloads with a `warn!` and returns `Ok(())`, so under `AutoCommit::When(PollingMessages)` those messages are gone with the offset already committed.

Either fix the variant choice in `proto_convert.rs` here, or file a follow-up and add a comment at this assertion so the `Schema::Proto` entry is not read as intended behaviour by the next person touching it.

Issue #3634, opened 2026-07-09
BDD: rust and go runners pass silently on undefined steps


seokjin0414:
### Bug description

Follow-up from #3554 ([discussion](https://github.com/apache/iggy/pull/3554#discussion_r3512938253)): the C++ BDD suite got `--strict` there, but per that thread the rust and go runners have the same hole — if a shared feature file gains a step without a matching definition, the suite reports it and still exits 0, so CI stays green while steps silently stop running.

- **rust** (`bdd/rust/tests/basic_messaging.rs`, `bdd/rust/tests/leader_redirection.rs`): both entry points use cucumber's default `World::run`, which treats an unmatched step as skipped and doesn't fail on skipped. Fix: `Self::cucumber().fail_on_skipped().run_and_exit(...)` in both (the feature files use no tag filtering, so nothing is legitimately skipped).
- **go** (`bdd/go/tests/suite_test.go`): godog is non-strict by default and neither `godog.Options` block sets `Strict`. Fix: `Strict: true` in both suites.

The other runners (node/java/python/csharp) are strict by default per the discussion above, so no changes needed there.

Taking this one (as discussed in the PR thread) — will follow with a PR. Verification will mirror the C++ change: prove the red path before calling it done.

### Affected area / component

Rust SDK, Go SDK, CI / build / tooling

### Deployment

Not applicable

### Versions

master (`2fc6ff6eb`)

### Reproduction

1. Append a bogus step (e.g. `And the server should sing a lullaby`) to `bdd/scenarios/basic_messaging.feature`
2. `./scripts/run-bdd-tests.sh rust` (or `go`)
3. The runner reports the step as undefined/skipped and the suite still exits 0

### Contribution

- [x] I'm willing to submit a pull request to fix this bug

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

justinmclean:
Both halves have landed: #3680 and #3681. Suggest closing.

PR #3681, opened 2026-07-16
test(bdd): fail Go BDD runner on undefined steps

seokjin0414:
Follow-up to #3554 — Go half of #3634 (Rust half in #3680).

godog is non-strict by default and neither `godog.Options` block set `Strict`, so a step without a matching definition in the Go runner is reported as undefined while `go test` still exits 0 and CI stays green. Both suites now set `Strict: true`; the existing `s.Run() != 0` check turns that into a test failure.

Verification (local server with `--with-default-root-credentials`, fresh data dir per run):
- green path: `BDD_FEATURE=basic_messaging go test -count=1 ./tests/` → exit 0, `17 steps (17 passed)`
- red path (repro from #3634: bogus step appended to the feature): exit 1, `1 scenarios (1 undefined)` → FAIL
- `gofmt` / `go vet` clean

Refs #3634

### DIFF
--- a/bdd/go/tests/suite_test.go
+++ b/bdd/go/tests/suite_test.go
@@ -37,6 +37,7 @@ func TestFeatures(t *testing.T) {
 			Options: &godog.Options{
 				Format:   "pretty",
 				Paths:    []string{"../../scenarios/basic_messaging.feature"},
+				Strict:   true,
 				TestingT: t,
 			},
 		})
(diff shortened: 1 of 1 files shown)


seokjin0414:
@hubcio could you take a look at this one as well? go half of #3634, same fix as the rust one (#3680) you approved — 2-line diff. CI is re-running after a rebase onto latest master.

hubcio (approved):

numinnex (approved):

ryankert01 (approved):

PR #3690, opened 2026-07-17
fix(python): pin python version and add uv caching

slbotbm:
While working on the python sdk, often `uv sync --all-extras` does not actually recreate the .so file using the latest changes and leaves the old version of the sdk, leading to difficulties in actually testing it on local. This change adds explicit cache keys for files. If the files are changed, the .so extension is automatically rebuilt on running `uv sync --all-extras`. (caveat: it will also be run when running other commands like `uv run ruff format` since there is a hidden sync step in these commands).

I also pin the python version to 3.10 explicitly. Previously, we only declare the python version to be >= 3.10, which led to uv downloading python according to the following hierarchy:

 1. Existing compatible uv-managed Python, preferring newer versions.
  2. First compatible system Python found—not necessarily newest.
  3. If none exists, downloads the latest compatible stable Python.

By pinning the version, we force uv to utilize 3.10.


foreign/python/pyproject.toml
```diff
@@ -149,6 +149,14 @@ filterwarnings = [
 
 [tool.uv]
 exclude-newer = "7 days"
+cache-keys = [
+    { file = "pyproject.toml" },
+    { file = "Cargo.toml" },
+    { file = "Cargo.lock" },
+    { file = "src/**/*.rs" },
```
ethanlin01x, line 158: Would it make sense to also include the workspace root manifest here?

{ file = "../../Cargo.toml" }

My understanding is that since foreign/python is excluded from the root workspace, but core/sdk inherits most of its dependency versions and features from [workspace.dependencies] in the root Cargo.toml, a version bump there could change the build without any of the current cache keys noticing — so uv sync might keep the previously built .so.

slbotbm: Added the root cargo.toml and cargo.lock to cache keys as well.

ryankert01 (commented):
We can add these 3 `.python-version` files to current sync check (`sync-python-interpreter-version.sh`) so precommit would detect version inconsistencies if we upgrade at somewhere.

slbotbm:
@ryankert01 updated, thanks for the reminder.

hubcio (approved):

mmodzelewski (approved):

PR #3691, opened 2026-07-17
feat(sdk): expose raw command API across foreign clients

spetz:
Foreign SDK users could not send protocol commands without a typed
client API, which blocked applications from using new command codes
until every SDK added dedicated methods.

Expose code-and-payload methods across Python, PHP, C++, Node, Go,
C#, and Java while preserving each client's framing, authentication,
error mapping, and transport behavior. Reject session-control codes
before sending so typed login state remains consistent.

Add shared BDD coverage for known, unknown, and session-control
commands across every SDK runner. Bump each foreign SDK version for
the new public API.


raw-command-iggy-sdk.md
slbotbm, line 1: This should not be committed?

slbotbm (changes requested):

raw-command-iggy-sdk.md
spetz: Good catch, fixed!

foreign/cpp/tests/client/low_level_e2e.cpp
```diff
@@ -32,6 +35,29 @@
...
+        const auto removed = std::none_of(clients.begin(), clients.end(),
+                                          [client_id](const auto &client) { return client.client_id == client_id; });
+        if (removed) {
+            return true;
+        }
+
+        std::this_thread::sleep_for(session_removal_poll_interval);
+    } while (std::chrono::steady_clock::now() < deadline);
+
+    return false;
```
slbotbm: Let's inline this code into the two tests. I would like to avoid defining ad hoc functions since it will become difficult to maintain after some time if these types of functions keep on accumulating

slbotbm (changes requested):
other than one comments, looks alright

foreign/cpp/tests/client/low_level_e2e.cpp
spetz: Fixed

slbotbm (dismissed):

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
```diff
@@ -217,6 +221,39 @@ public CompletableFuture<IdentityInfo> login() {
...
+    public CompletableFuture<byte[]> sendRawWithResponse(int code, byte[] payload) {
+        if (isSessionControlCode(code)) {
+            return CompletableFuture.failedFuture(
+                    IggyServerException.fromTcpResponse(INVALID_COMMAND_ERROR_CODE, new byte[0]));
+        }
+        if (connection == null) {
+            throw new IggyNotConnectedException();
+        }
+
+        return connection.send(code, Unpooled.wrappedBuffer(payload)).thenApply(response -> {
```
hubcio: `Unpooled.wrappedBuffer(payload)` aliases the caller's array without copying - the copy into the frame happens later in `IggyFrameEncoder.encode` on the netty event loop, after the async channel acquire (plus an auth round-trip when it kicks in). an async caller that reuses or mutates the array while the future is pending silently corrupts the payload on the wire. all typed commands build fresh buffers, so raw is the only path aliasing caller memory. `Unpooled.copiedBuffer(payload)` fixes it. same line also NPEs on null payload for non-guarded codes (guarded ones return the failed future before touching it), so a null check wouldn't hurt either.

foreign/node/src/wire/command-set.ts
```diff
@@ -225,4 +237,20 @@ export abstract class CommandAPI extends AbstractAPI {
...
+   *
+   * @param code - Command code to send
+   * @param payload - Raw command payload
+   * @returns Raw response payload
+   */
+  async sendRawWithResponse(code: number, payload: Buffer): Promise<Buffer> {
+    if (SESSION_CONTROL_CODES.has(code))
+      throw responseError(code, INVALID_COMMAND_ERROR_CODE);
+
+    const response = await (await this.clientProvider()).sendCommand(code, payload);
```
hubcio: same aliasing shape as the java async client: the caller's `payload` travels by reference through the send queue (`sendCommand` enqueues it, the copy into the wire frame only happens at dequeue in `serializeCommand`). when the queue is busy or connect/auth is still pending, a caller mutating the buffer mid-flight corrupts the outgoing frame. typed commands always serialize into fresh buffers, so raw is the first path that puts a caller-owned buffer in the queue. a defensive `Buffer.from(payload)` here would match the java fix.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -305,6 +305,38 @@ func (c *IggyTcpClient) do(ctx context.Context, cmd command.Command) ([]byte, er
...
+// Session-control codes return ierror.ErrInvalidCommand without writing to the connection.
+func (c *IggyTcpClient) SendRawWithResponse(ctx context.Context, code uint32, payload []byte) ([]byte, error) {
+	if isSessionControlCode(code) {
+		return nil, ierror.ErrInvalidCommand
+	}
+
+	bp := acquireRequestBuf()
+	defer releaseRequestBuf(bp)
+
+	buf := append((*bp)[:0], 0, 0, 0, 0, 0, 0, 0, 0)
```
hubcio: this hand-rolls the same frame layout as `encodeWireRequest` below - 8-byte header, code at [4:8], length patched from the realized body length - and that length-from-realized-bytes rule is exactly the desync invariant the comment above `encodeWireRequest` warns about. it can't be reused directly since it takes a `command.Command`, but a small shared helper (`frameRaw(buf, code, body)`) used by both would keep that invariant in one place.

foreign/cpp/src/client.rs
```diff
@@ -1129,6 +1130,19 @@ impl Client {
...
+    /// Sends a command code and payload and returns the raw response bytes.
+    /// Session-control codes return an invalid-command error.
+    pub fn send_raw_with_response(&self, code: u32, payload: Vec<u8>) -> Result<Vec<u8>, String> {
+        RUNTIME.block_on(async {
+            let response = self
+                .inner
+                .send_binary_request(code, Bytes::from(payload))
+                .await
+                .map_err(|error| format!("Could not send raw command '{code}': {error}"))?;
+            Ok(response.to_vec())
```
hubcio: `response.to_vec()` always copies. `Vec::from(response)` reclaims the allocation with no copy, since this `Bytes` comes uniquely from the tcp read path (`BytesMut::freeze`). cold path so no urgency, just a free win. same pattern in the php client's `send_raw_with_response`.

bdd/go/tests/suite_test.go
```diff
@@ -53,6 +53,16 @@ func TestFeatures(t *testing.T) {
 			},
 		})
 	}
+	if feature == "all" || feature == "raw_command" {
+		suites = append(suites, godog.TestSuite{
+			ScenarioInitializer: initRawCommandScenario,
+			Options: &godog.Options{
+				Format:   "pretty",
+				Paths:    []string{"../../scenarios/raw_command.feature"},
+				TestingT: t,
```
hubcio, line 63: the two suites above got `Strict: true` (#3681) so undefined steps fail the run - this new suite doesn't have it, so a typo'd step pattern here would report undefined and still pass. add `Strict: true` for consistency.

hubcio (changes requested):
besides below findings i noticed that the rust sdk exposes this publicly as `send_binary_request` while every foreign sdk ships `sendRawWithResponse` (which matches the rust internal transport method). while everything is pre-release a rename either way is free; after release it's breaking.

numinnex (dismissed):

ryankert01 (approved):
lg, make sense to me.

slbotbm (dismissed):

mmodzelewski (approved):

numinnex (approved):

hubcio (approved):

PR #3694, opened 2026-07-17
fix(ci): use portable fallbacks for bash-4 builtins and GNU sed in hook scripts

### LINKED ISSUE #3693: pre-commit hooks use bash-isms that block contributors on lower version bash.
## Problem

Several pre-commit hooks invoke scripts that rely on bash-specific builtins (`mapfile`) and GNU extensions (`sed -i` without backup extension). These fail on non-bash shells like zsh, preventing contributors from committing code.

### `mapfile` (bash >= 4.0 builtin)

`mapfile` does not exist in POSIX sh or zsh. Hooks using it will abort with `mapfile: command not found`.

### `sed -i` without backup extension (GNU sed extension)

BSD `sed` (default on macOS) requires `sed -i ""` with an explicit backup argument. Bare `sed -i` errors with `sed: invalid command code`.

### `mapfile` → `while IFS= read -r` loop

```bash
# Before (bash >= 4.0)
mapfile -t items < <(some_command)

# After (POSIX-compatible)
items=()
while IFS= read -r line; do items+=("$line"); done < <(some_command)
```

### `sed -i` → backup-then-remove

```bash
# Before (GNU sed only)
sed -i "s/old/new/" "$file"

# After (works on both GNU and BSD sed)
sed -i.bak "s/old/new/" "$file" && rm -f "$file.bak"
```

Both patterns already have precedent in the repo (the `sed -i.bak` + `rm` pattern exists in one CI script; a portable `sed -i` wrapper exists in a release script).

jiengup:
Closes #3693

## Rationale

Several hook scripts use `mapfile` (bash >= 4) and `sed -i` without backup extension (GNU extension). On macOS stock bash 3.2 and BSD sed, these fail, blocking contributors from committing.

## What changed

- `mapfile` calls now guard with `command -v mapfile` and fall back to a `while IFS= read -r` loop. 
- `sed -i` calls is replaced by the `sed -i.bak && rm` pattern already established in `trailing-whitespace.sh`.

## Local Execution
- `prek run -a`


scripts/ci/python-sdk-version-sync.sh
```diff
@@ -196,13 +196,13 @@ elif [ "$MODE" = "fix" ]; then
         # Cargo version is newer, update pyproject.toml
         NEW_PYPROJECT_VERSION="$CARGO_NORMALIZED"
         echo -e "${YELLOW}Cargo.toml has newer version, updating pyproject.toml...${NC}"
-        sed -i "s/^version = \"$PYPROJECT_VERSION\"/version = \"$NEW_PYPROJECT_VERSION\"/" "$PYPROJECT_TOML"
+        sed -i.bak "s/^version = \"$PYPROJECT_VERSION\"/version = \"$NEW_PYPROJECT_VERSION\"/" "$PYPROJECT_TOML" && rm -f "$PYPROJECT_TOML.bak"
```
hubcio: also at line 205.

the `&&` puts sed on the exempt side of errexit, so if sed ever fails the script no longer aborts - it keeps going, prints the green "Updated" message and exits 0 with the file unedited (plus a stray `.bak` in the tree, and `*.bak` isn't gitignored). the old bare `sed -i` failed loud here, and trailing-whitespace.sh (where this pattern comes from) runs sed and `rm` as two separate statements, which keeps the loud failure. same two-statement form here restores fail-closed.

scripts/ci/sync-python-interpreter-version.sh
```diff
@@ -122,7 +122,7 @@ ensure_line() {
     fi
 
     if [ "$MODE" = "fix" ]; then
-        sed -i -E "s|$current_pattern|$replacement|" "$file"
+        sed -i.bak -E "s|$current_pattern|$replacement|" "$file" && rm -f "$file.bak"
```
hubcio: same at lines 195, 200, 252 and 316.

chaining `sed -i.bak ... && rm -f` swallows a sed failure under `set -e` (left side of `&&` is exempt from errexit), so a failed edit still prints "Fixed", bumps `FIXED_CHECKS` and exits 0. the previous bare `sed -i` aborted loud. two separate statements - as in trailing-whitespace.sh - keep the portable `.bak` form without losing the abort.

scripts/ci/sync-python-interpreter-version.sh
```diff
@@ -187,12 +192,12 @@ ensure_classifiers() {
...
 
             if [ "$version_number" -lt "$PYTHON_VERSION_NUMBER" ]; then
-                sed -i -E "/^    \"Programming Language :: Python :: ${version//./\\.}\",$/d" "$file"
+                sed -i.bak -E "/^    \"Programming Language :: Python :: ${version//./\\.}\",$/d" "$file" && rm -f "$file.bak"
             fi
         done
 
         if [ "$has_minimum" -eq 0 ]; then
-            sed -i -E "/^    \"Programming Language :: Python :: ${PYTHON_VERSION_MAJOR}\",$/a\\    \"Programming Language :: Python :: ${PYTHON_VERSION}\"," "$file"
+            sed -i.bak -E "/^    \"Programming Language :: Python :: ${PYTHON_VERSION_MAJOR}\",$/a\\    \"Programming Language :: Python :: ${PYTHON_VERSION}\"," "$file" && rm -f "$file.bak"
```
hubcio: the `a\` one-liner is a GNU extension - BSD/macOS sed only accepts the append text on the following line, so on a mac this either errors out or appends at column 0. worse: the deletes at line 195 have already run by then, and with the `&&` swallow the function still reports "Fixed" and exits 0 - so a macOS `--fix` run for a new python minimum can leave pyproject.toml with the old classifiers deleted and the new one never added, reported as success. portable form is `-e '/pat/a\' -e '\    text'`, or do the insert with awk.

scripts/ci/sync-rustc-version.sh
```diff
@@ -156,13 +156,13 @@ for dockerfile in $DOCKERFILES; do
     elif [ "$MODE" = "fix" ]; then
         if { [ -n "$SOURCE" ] && [ "$CURRENT_VERSION" != "$EXPECTED_VERSION" ]; } || [ "$RUST_IMAGE_MISMATCH" = "true" ]; then
             if [ -n "$SOURCE" ] && [ "$CURRENT_VERSION" != "$EXPECTED_VERSION" ] && [ "$SOURCE" = "arg" ]; then
-                sed -i "s/^ARG RUST_VERSION=.*/ARG RUST_VERSION=$EXPECTED_VERSION/" "$dockerfile"
+                sed -i.bak "s/^ARG RUST_VERSION=.*/ARG RUST_VERSION=$EXPECTED_VERSION/" "$dockerfile" && rm -f "$dockerfile.bak"
```
hubcio: also 161, 164, 165.

`sed -i.bak ... && rm -f` swallows sed failures under `set -e` (`&&`-left is exempt from errexit), so a failed edit still counts toward `FIXED_FILES` and the script exits 0. previous bare `sed -i` aborted. split into two statements like trailing-whitespace.sh does.

scripts/ci/sync-rustc-version.sh
```diff
@@ -156,13 +156,13 @@ for dockerfile in $DOCKERFILES; do
...
             elif [ -n "$SOURCE" ] && [ "$CURRENT_VERSION" != "$EXPECTED_VERSION" ]; then
-                sed -i -E "/FROM[[:space:]].*\\brust:[0-9]/ s#(\\brust:)[0-9]+\\.[0-9]+(\\.[0-9]+)?#\\1$EXPECTED_VERSION#g" "$dockerfile"
+                sed -i.bak -E "/FROM[[:space:]].*\\brust:[0-9]/ s#(\\brust:)[0-9]+\\.[0-9]+(\\.[0-9]+)?#\\1$EXPECTED_VERSION#g" "$dockerfile" && rm -f "$dockerfile.bak"
```
hubcio: `\brust:` here (and in `RUST_IMAGE_PATTERN` at line 82, plus the greps at 111/113/122/124) relies on `\b`, which BSD sed and grep don't support - it degrades to the literal string `brust:` and never matches. so on macOS the FROM-form dockerfiles get skipped entirely: false green in `--check`, silent no-op in `--fix`; only the ARG-form path works. pre-existing, but since this PR is about BSD sed portability, the script stays macos-broken until `\b` goes.

scripts/verify-crates-publish.sh
```diff
@@ -254,12 +254,12 @@ log "Patching workspace.dependencies to target local-dev"
...
 # Injects `, registry = "local-dev"` before the closing brace of the line.
 # Match by crate name + path + version so we do not touch unrelated entries.
-sed -i -E \
+sed -i.bak -E \
     -e 's|^(iggy_binary_protocol = \{ path = "[^"]+", version = "[^"]+")( \})$|\1, registry = "local-dev"\2|' \
     -e 's|^(iggy_common = \{ path = "[^"]+", version = "[^"]+")( \})$|\1, registry = "local-dev"\2|' \
     -e 's|^(iggy = \{ path = "[^"]+", version = "[^"]+")( \})$|\1, registry = "local-dev"\2|' \
     -e 's|^(iggy-cli = \{ path = "[^"]+", version = "[^"]+")( \})$|\1, registry = "local-dev"\2|' \
-    Cargo.toml
+    Cargo.toml && rm -f Cargo.toml.bak
```
hubcio: `-i.bak` now creates `Cargo.toml.bak`, but `cleanup()` only restores `Cargo.toml` from the mktemp snapshot - it never removes the `.bak`. if sed fails (the `&&` skips the rm and the sanity check below exits 1) or a ctrl-c lands between sed and rm, `Cargo.toml.bak` survives the trap, which breaks the "leave the working tree untouched on exit, even on failure" contract at the top of this file. `*.bak` isn't gitignored either. one `rm -f Cargo.toml.bak` inside the `CARGO_TOML_PATCHED` branch of cleanup() covers all paths. the `&&` itself is fine here - the sanity check right below fails closed.

scripts/ci/edge-affected-images.sh
```diff
@@ -46,7 +46,12 @@ if [[ -z "$BASE" || -z "$HEAD" ]]; then
 fi
 
 CFG_JSON="$(yq -o=json -I=0 '.components' "$CONFIG")"
-mapfile -t ALL_IMAGES < <(jq -r 'to_entries[] | select(.value.registry == "dockerhub") | .key' <<<"$CFG_JSON")
+if command -v mapfile >/dev/null 2>&1; then
```
hubcio: same for the guards at 102 and 149.

these three fallbacks can never help: this script only runs in CI (post-merge.yml, ubuntu bash 5), and line 123 has an unconditional `declare -A`, which is bash-4-only anyway - any bash without `mapfile` dies there regardless. plain `mapfile -t` is enough in this file. alternatively, if one portable idiom everywhere is preferred, the bare `while IFS= read -r` loop alone (as in trailing-whitespace.sh) works on every bash and avoids writing the jq command twice per site - that would also suit the two genuinely needed sites in license-headers.sh and sync-python-interpreter-version.sh.

hubcio (changes requested):
the `-i.bak` switch itself is the right call and matches the trailing-whitespace.sh precedent. main ask is the `&& rm -f` chaining - it quietly replaces the old loud sed failure with a false "Fixed" + exit 0 (details in the line comments).

two findings sit on lines outside this diff, so noting them here:

- `scripts/verify-crates-publish.sh:313` - `lc="${crate,,}"` is bash-4-only, stock macOS bash 3.2 dies with `bad substitution`, and it fails late (after all four crates already published to the local registry). it's also a no-op since the crate names are already lowercase, so `lc="$crate"` fixes it for free.
- `scripts/ci/python-sdk-version-sync.sh:144` - `sort -V` is GNU coreutils-only, BSD sort rejects it. the `--fix` path hits it exactly when the two versions differ, i.e. the case the hook exists for. worth confirming on an actual mac.

with `a\`, `\b` and these two still around, #3693 ends up only partially closed for macos contributors - either keep the issue open or note a follow-up in the PR.

scripts/ci/sync-rustc-version.sh
jiengup: I will create a new issue for this problem and find a way to fix it after this PR is merged.

jiengup:
> the `-i.bak` switch itself is the right call and matches the trailing-whitespace.sh precedent. main ask is the `&& rm -f` chaining - it quietly replaces the old loud sed failure with a false "Fixed" + exit 0 (details in the line comments).
> 
> two findings sit on lines outside this diff, so noting them here:
> 
> * `scripts/verify-crates-publish.sh:313` - `lc="${crate,,}"` is bash-4-only, stock macOS bash 3.2 dies with `bad substitution`, and it fails late (after all four crates already published to the local registry). it's also a no-op since the crate names are already lowercase, so `lc="$crate"` fixes it for free.
> * `scripts/ci/python-sdk-version-sync.sh:144` - `sort -V` is GNU coreutils-only, BSD sort rejects it. the `--fix` path hits it exactly when the two versions differ, i.e. the case the hook exists for. worth confirming on an actual mac.
> 
> with `a\`, `\b` and these two still around, #3693 ends up only partially closed for macos contributors - either keep the issue open or note a follow-up in the PR.

Appreciate your detailed review.

Problems fixed in the newest patch:
- all the `sed -i.bak ... && rm xxx.bak` command is separated now, just like `trailing-whitespace.sh` does.
- scripts/ci/sync-python-interpreter-version.sh: `a\` GNU extension in `sed` proved not to work, and is now replaced by an `awk` insert. 
- scripts/verify-crates-publish.sh: add `rm -f Cargo.toml.bak` inside the `CARGO_TOML_PATCHED` branch of `cleanup()`.
- unnecessary `mapfile` fallback is removed, and the necessary ones are replaced by bare `while IFS= read -r` loop to reduce duplication.

Additional change:
- scripts/verify-crates-publish.sh: `lc="${crate,,}"` changes to `lc="crate" and lower-case crate name contract is documented.
- python-sdk-version-sync.sh: use `sort -t. -k1,1n -k2,2n -k3,3n` instead of `sort -V`, though `sort -V` is actually passed on macOS with bash 3.2.

Leave as a follow-up issue:
-  scripts/ci/sync-rustc-version.sh: `\brust` makes FROM-form dockerfiles get skipped entirely. A new follow-up issue is opened #3739. I will track it.

hubcio (dismissed):

scripts/ci/python-sdk-version-sync.sh
```diff
@@ -139,9 +139,9 @@ compare_versions() {
         dev2="0"
     fi
 
-    # Compare base versions using sort -V
+    # Compare base versions using POSIX sort (X.Y.Z numeric per component).
     local sorted
-    sorted=$(printf '%s\n%s' "$base1" "$base2" | sort -V | head -1)
+    sorted=$(printf '%s\n%s' "$base1" "$base2" | sort -t. -k1,1n -k2,2n -k3,3n | head -1)
```
ethanlin01x, line 144: `sort -V `works fine on stock macOS (verified locally). Since the original way isn't broken and numeric keys only fit pure numbers,consider reverting to -V.

scripts/ci/sync-python-interpreter-version.sh
```diff
@@ -187,12 +189,18 @@ ensure_classifiers() {
...
         done
 
         if [ "$has_minimum" -eq 0 ]; then
-            sed -i -E "/^    \"Programming Language :: Python :: ${PYTHON_VERSION_MAJOR}\",$/a\\    \"Programming Language :: Python :: ${PYTHON_VERSION}\"," "$file"
+            awk -v major="$PYTHON_VERSION_MAJOR" -v version="$PYTHON_VERSION" '
+              { print }
+              $0 == "    \"Programming Language :: Python :: " major "\"," {
+                printf "    \"Programming Language :: Python :: %s\",\n", version
+              }
+            ' "$file" > "$file.new" && mv "$file.new" "$file"
```
ethanlin01x: Could you split the `awk` and `mv` into two statements? With `&&`, a failed `awk` is exempt from` set -e`, so the script would still print "Fixed" and exit 0 with the file unedited — same errexit swallow as the earlier `sed -i.bak ... && rm` sites.

ethanlin01x (commented):
Left a few minor comments inline. I ran into this problem myself on macOS, so I'd love to see this merged soon — thanks for the fix!

hubcio:
@jiengup could you please rebase and fix comments from @ethanlin01x ? this change is good and we'd like to merge it this week

scripts/ci/python-sdk-version-sync.sh
jiengup: Since this PR is not only for the macOS platform. We also want to make it compatible on ohter shell like zsh and fish. So maybe we can keep this change?

What do you think? @hubcio

hubcio: @jiengup  i dont understand. you enforce the shell via shebang at the top of the file, in all scripts.

jiengup: @hubcio Oh sorry, I confused it. I mean, `sort -V` can't work on `BSD sort`, maybe older macOS? Though it actually works on the latest macOS version. So keep this change can more compatible since this PR is for emitting GNU limitations.

hubcio: yeah, we can keep it. @ethanlin01x hope you are ok with that :) resolving this comment

hubcio (approved):

ethanlin01x (approved):

hubcio (approved):

numinnex (approved):

PR #3695, opened 2026-07-17
feat(python): add user management methods to IggyClient

### LINKED ISSUE #3682: [python sdk] Add functions related to user management
### Description

The Python SDK currently has no user management capability beyond `login_user`. Comparing the Rust SDK client traits (`core/common/src/traits/user_client.rs`) against the PyO3 bindings in `foreign/python/src/client.rs`, the entire rest of the `UserClient` surface is unexposed: there is no way to list, inspect, create, update, or delete users from Python.

This blocks any Python application that needs to provision users programmatically (e.g. setting up service accounts in deployment scripts) — today that requires dropping down to the CLI or another SDK.

### Affected area / component

Python SDK

### Proposed solution

Add the CRUD subset as async methods on `IggyClient`, wrapping the corresponding Rust SDK functions with PyO3, consistent with how the existing topic and consumer group methods are exposed:

- `get_user(user_id)` / `get_users()`
- `create_user(username, password, status)`
- `update_user(user_id, username=None, status=None)`
- `delete_user(user_id)`

This also means exposing `UserInfo`, `UserInfoDetails`, and the `UserStatus` enum as Python classes.

Scope cut: `create_user` takes `permissions: Option<Permissions>` on the Rust side. `Permissions` is a nested structure that deserves its own type mapping, so this issue always passes `None` and leaves the parameter unexposed. I'll handle `Permissions` (together with `update_permissions`, `change_password`, and `logout_user`) in a follow-up issue.

Deliverables: implementation, e2e tests in `foreign/python/tests/test_user.py`, regenerated `apache_iggy.pyi` stubs, all pre-commit hooks passing.

### Alternatives considered

_No response_

### Contribution

### Good first issue

ethanlin01x:

Closes #3682

## Rationale

Adds the missing user management operations to the Python SDK, which previously had no binding beyond `login_user` and forced callers to the CLI or another SDK to provision users.

## What changed?

The Python SDK exposed no part of the Rust `UserClient` surface besides `login_user`, so users could not be listed, inspected, created, updated, or deleted from Python.

`get_user`, `get_users`, `create_user`, `update_user`, and `delete_user` now bind through to the Rust `UserClient`, and `UserInfo`, `UserInfoDetails`, and `UserStatus` are exposed as Python classes. As scoped in the issue, `create_user` always passes `None` for permissions; the `Permissions` mapping is left to a follow-up together with `update_permissions`, `change_password`, and `logout_user`.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage
Claude was used to help generate and review this PR and all the changes are checked by the human.

### DIFF
--- a/foreign/python/tests/test_user.py
+++ b/foreign/python/tests/test_user.py
@@ -21,13 +21,16 @@ from apache_iggy import IggyClient, UserInfoDetails, UserStatus
 
 from .utils import get_server_config, wait_for_ping, wait_for_server
 
-# Server-side limits: usernames are 3-50 characters, passwords 3-100.
+# Server-side limits: usernames are 3-50 bytes, passwords 3-100 bytes.
+MIN_USERNAME_BYTES = 3
 MAX_USERNAME_BYTES = 50
+MIN_PASSWORD_BYTES = 3
+MAX_PASSWORD_BYTES = 100
 
 
 def _unique_credentials(unique_name) -> tuple[str, str]:
     username = unique_name(max_bytes=MAX_USERNAME_BYTES)
-    password = unique_name(max_bytes=MAX_USERNAME_BYTES)
+    password = unique_name(max_bytes=MAX_PASSWORD_BYTES)
     return username, password
 
 
@@ -109,6 +112,96 @@ class TestCreateUser:
         await client.login_user(username, password)
 
         await iggy_client.delete_user(created.id)
+
+    @pytest.mark.parametrize(
+        "username",
+        ["a" * (MIN_USERNAME_BYTES - 1), "a" * (MAX_USERNAME_BYTES + 1)],
+        ids=["too-short", "too-long"],
+    )
+    @pytest.mark.asyncio
+    async def test_create_user_rejects_out_of_bounds_username(
+        self, iggy_client: IggyClient, unique_name, username
+    ):
+        """Test create_user rejects usernames outside the 3-50 byte range."""
+        with pytest.raises(RuntimeError):
+            await iggy_client.create_user(
+                username, unique_name(max_bytes=MAX_PASSWORD_BYTES)
+            )
+
+    @pytest.mark.parametrize(
+        "password",
+        ["a" * (MIN_PASSWORD_BYTES - 1), "a" * (MAX_PASSWORD_BYTES + 1)],
+        ids=["too-short", "too-long"],
+    )
+    @pytest.mark.asyncio
+    async def test_create_user_rejects_out_of_bounds_password(
+        self, iggy_client: IggyClient, unique_name, password
+    ):
+        """Test create_user rejects passwords outside the 3-100 byte range."""
+        with pytest.raises(RuntimeError):
+            await iggy_client.create_user(
+                unique_name(max_bytes=MAX_USERNAME_BYTES), password
+            )
+
+    @pytest.mark.parametrize(
+        "prefix",
+        ["ユーザー", "사용자", "ผู้ใช้", "🦀🚀"],
+        ids=["japanese", "korean", "thai", "emoji"],
+    )
+    @pytest.mark.asyncio
+    async def test_create_user_accepts_multibyte_credentials(
+        self, iggy_client: IggyClient, unique_name, prefix
+    ):
+        """Test non-ascii credentials within the byte limits are accepted."""
+        username = f"{prefix}{unique_name(min_bytes=4, max_bytes=8)}"
+        password = f"{prefix}{unique_name(min_bytes=4, max_bytes=8)}"
+        assert len(username.encode()) <= MAX_USERNAME_BYTES
+
+        created = await iggy_client.create_user(username, password)
+        assert created.username == username
+
+        fetched = await iggy_client.get_user(username)
+        assert fetched is not None
+        assert fetched.id == created.id
+
+        host, port = get_server_config()
+        client = IggyClient(f"{host}:{port}")
+        await client.connect()
+        await client.login_user(username, password)
+
+        await iggy_client.delete_user(created.id)
+
+    @pytest.mark.parametrize(
+        "username",
+        ["あ" * 17, "🦀" * 13],
+        ids=["japanese-51-bytes", "emoji-52-bytes"],
+    )
+    @pytest.mark.asyncio
+    async def test_create_user_username_limit_is_bytes_not_characters(
+        self, iggy_client: IggyClient, unique_name, username
+    ):
+        """Test a username within 50 characters but over 50 bytes is rejected."""
+        assert len(username) <= MAX_USERNAME_BYTES
+        assert len(username.encode()) > MAX_USERNAME_BYTES
+
+        with pytest.raises(RuntimeError):
+            await iggy_client.create_user(
+                username, unique_name(max_bytes=MAX_PASSWORD_BYTES)
+            )
+
+    @pytest.mark.asyncio
+    async def test_create_user_password_limit_is_bytes_not_characters(
+        self, iggy_client: IggyClient, unique_name
+    ):
+        """Test a password within 100 characters but over 100 bytes is rejected."""
+        password = "あ" * 34
+        assert len(password) <= MAX_PASSWORD_BYTES
+        assert len(password.encode()) > MAX_PASSWORD_BYTES
+
+        with pytest.raises(RuntimeError):
+            await iggy_client.create_user(
+                unique_name(max_bytes=MAX_USERNAME_BYTES), password
+            )
 
 
 class TestGetUser:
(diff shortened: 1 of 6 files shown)


ethanlin01x:
The 3 lines Codecov flags as uncovered in foreign/python/src/user.rs are the #[pyclass(...)] and #[pymethods] attribute lines. They are not missing tests. All hand-written logic in the file is fully exercised by tests/test_user.py.

Should these attribute lines be brought to 100% as well, or is it fine to leave them as is?

slbotbm:
> The 3 lines Codecov flags as uncovered in foreign/python/src/user.rs are the #[pyclass(...)] and #[pymethods] attribute lines. They are not missing tests. All hand-written logic in the file is fully exercised by tests/test_user.py.
>Should these attribute lines be brought to 100% as well, or is it fine to leave them as is?

Fine as it is

foreign/python/tests/test_user.py
```diff
@@ -0,0 +1,376 @@
...
+    async def test_get_nonexistent_user_returns_none(
+        self, iggy_client: IggyClient, unique_name
+    ):
+        """Test getting a non-existent user by name or numeric id returns None."""
+        user_by_name = await iggy_client.get_user(
+            unique_name(max_bytes=MAX_USERNAME_BYTES)
+        )
+        assert user_by_name is None
+
+        user_by_id = await iggy_client.get_user(999999)
```
slbotbm: This is not stable. There could be a user that has user id 999999.

foreign/python/tests/test_user.py
slbotbm, line 1: Here, you do not test username and password length boundaries. Add tests for the following for update_user and create_user:
- 2-byte and 51-byte usernames failing.
- 2-byte and 101-byte passwords failing.
- Non-ascii characters like Japanese, Korean, Thai, emojis. The choice of languages to test is up to you.

foreign/python/tests/test_user.py
```diff
@@ -0,0 +1,376 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+
+from apache_iggy import IggyClient, UserInfoDetails, UserStatus
+
+from .utils import get_server_config, wait_for_ping, wait_for_server
+
+# Server-side limits: usernames are 3-50 characters, passwords 3-100.
```
slbotbm: It is not 3-50 characters, but 3-50 bytes.

foreign/python/tests/test_user.py
slbotbm: Please add the error-before-login pattern to the rest of the tests that do not have this.

slbotbm: For create_user: add the following tests:
- An inactive user cannot log in even with the correct password
- A deleted user cannot start a fresh authenticated session.

For get_user:
- Empty string identifiers failing locally.
- Negative and greater-than-`u32` integer identifiers failing.

For get_users:
- It returns the users in a strict ascending order, so please assert that as well if not yet asserted. 
- Do not assert absolute counts since the server may contain other users as well. 

For update_user:
- Calling with both optional fields as `None` and verifying the user remains unchanged, because Rust permits that no-op request.
- Rejected updates preserving both the original username and status. (could be folded into existing tests)

For delete_user:
- Root deletion failing by username and numeric ID `0`.
- Deleting the same user twice failing on the second call.
- Deleting an inactive user. (could be folded into existing happy path test)
- A live session owned by the deleted user failing its next operation.
- The deleted username being reusable without old password state.
- The target disappearing from both `get_user()` and `get_users()`.

foreign/python/tests/test_user.py
```diff
@@ -0,0 +1,376 @@
...
+    async def test_create_and_get_user(self, iggy_client: IggyClient, unique_name):
+        """Test user creation returns details and the user is retrievable."""
+        username, password = _unique_credentials(unique_name)
+
+        created = await iggy_client.create_user(username, password, UserStatus.Active)
+        assert isinstance(created, UserInfoDetails)
+        assert isinstance(created.id, int)
+        assert created.username == username
+        assert created.status == UserStatus.Active
+        assert created.created_at > 0
```
slbotbm: Do not assert `created_at`

foreign/python/tests/test_user.py
```diff
@@ -0,0 +1,376 @@
...
+        first = await iggy_client.create_user(first_username, first_password)
+        second = await iggy_client.create_user(second_username, second_password)
+
+        users = await iggy_client.get_users()
+        usernames = [user.username for user in users]
+        assert "iggy" in usernames
+        assert first_username in usernames
+        assert second_username in usernames
+        assert all(isinstance(user.id, int) for user in users)
+        assert all(user.created_at > 0 for user in users)
```
slbotbm: Do not assert creation date

foreign/python/tests/test_user.py
slbotbm: In a continuously used server, the ID could eventually come into existence.

foreign/python/tests/test_user.py
```diff
@@ -0,0 +1,376 @@
...
+        self, iggy_client: IggyClient, unique_name
+    ):
+        """Test repeated get_users calls return a stable user view."""
+        username, password = _unique_credentials(unique_name)
+        created = await iggy_client.create_user(username, password)
+
+        first = await iggy_client.get_users()
+        second = await iggy_client.get_users()
+
+        assert [user.id for user in first] == [user.id for user in second]
```
slbotbm: For a persistent/shared server, this will not hold.

foreign/python/tests/test_user.py
slbotbm: Please also remove the `match=...` blocks in error catching blocks since those messages could change in the future.

slbotbm (changes requested):

slbotbm:
Also, please do commit the new changes as new commits. This will help me in understanding what changed after my review.

foreign/python/tests/test_user.py
ethanlin01x: Fixed. The test now creates a user, deletes it, and asserts the vacated id returns None, so it no longer depends on any hardcoded id.
d41fb7650770b19eaea06ddf069ebcca89dddcde

foreign/python/tests/test_user.py
ethanlin01x: Fixed the comment and added MIN/MAX constants for both limits in bytes. The byte semantics are now covered by tests as well.
a671663f51bdf46053dd3477b7a3f938f819c9c5

foreign/python/tests/test_user.py
ethanlin01x: Removed
3d899e7755b410b772691df41dde7afed7d48d75

foreign/python/tests/test_user.py
ethanlin01x: Removed
3d899e7755b410b772691df41dde7afed7d48d75

foreign/python/tests/test_user.py
ethanlin01x: Removed the test. Its remaining value (a stable view of a single user) is already covered by the repeated get_user test, and the ordering guarantee is now asserted directly via the strictly ascending id check.

foreign/python/tests/test_user.py
ethanlin01x: Thanks for the detailed review!

1. Added the boundary tests in a671663f5: 2/51-byte usernames and 2/101-byte passwords failing, plus Japanese, Korean, Thai and emoji credentials within the byte limits succeeding and over-limit ones failing. 

2. error-before-login: the parametrized test_user_management_requires_connection_and_auth already covers all five methods. Let me know if you meant additional scenarios beyond that.

3. Added the per-method tests in da1d23a24 (inactive/deleted login, local identifier validation, ascending get_users order, update noop, and the delete_user cases). One behavioral note: deleting a user does not invalidate its live session outright, and a regular user has no permissions to lose, so the test asserts get_user on itself flipping to None plus privileged calls staying rejected.

4. Removed the match= blocks in 3d899e775; only the exception type is asserted now.

foreign/python/tests/test_user.py
```diff
@@ -110,6 +113,96 @@ async def test_created_user_can_login(self, iggy_client: IggyClient, unique_name
...
+        assert len(username) <= MAX_USERNAME_BYTES
+        assert len(username.encode()) > MAX_USERNAME_BYTES
+
+        with pytest.raises(RuntimeError):
+            await iggy_client.create_user(
+                username, unique_name(max_bytes=MAX_PASSWORD_BYTES)
+            )
+
+    @pytest.mark.asyncio
+    async def test_create_user_password_limit_is_bytes_not_characters(
```
slbotbm: This test is not required

foreign/python/tests/test_user.py
```diff
@@ -110,6 +113,96 @@ async def test_created_user_can_login(self, iggy_client: IggyClient, unique_name
...
+
+        await iggy_client.delete_user(created.id)
+
+    @pytest.mark.parametrize(
+        "username",
+        ["あ" * 17, "🦀" * 13],
+        ids=["japanese-51-bytes", "emoji-52-bytes"],
+    )
+    @pytest.mark.asyncio
+    async def test_create_user_username_limit_is_bytes_not_characters(
```
slbotbm: This as well

slbotbm (changes requested):

foreign/python/tests/test_user.py
```diff
@@ -0,0 +1,664 @@
...
+        await iggy_client.delete_user(created.id)
+
+        assert await iggy_client.get_user(username) is None
+        assert await iggy_client.get_user(created.id) is None
+        users = await iggy_client.get_users()
+        assert created.id not in [user.id for user in users]
+        assert username not in [user.username for user in users]
+
+    @pytest.mark.asyncio
+    async def test_deleted_user_live_session_loses_identity(
```
slbotbm, line 551: This test does not prove its claim:
- get_user(username) is None only proves deleted target no longer resolves.
- get_users() already fails before deletion because user has no permissions.
- Test still passes if deleted session remains authenticated.

Let's comment this out and add a todo to uncomment once permission management is added to python sdk

slbotbm (changes requested):

foreign/python/tests/test_user.py
ethanlin01x: Removed, 198274a99c52abbb73139922bdcb68bb15e6d54f

foreign/python/tests/test_user.py
ethanlin01x: Removed 198274a99c52abbb73139922bdcb68bb15e6d54f

foreign/python/tests/test_user.py
ethanlin01x: Thanks, you're right. Reduced it to a no-op with a TODO to restore once permission management lands in the Python SDK (a9bb97bca). Kept the function and docstring so the intent stays visible.

slbotbm (approved):

spetz (approved):

Issue #3641, opened 2026-07-10
Flaky Java SDK test: AsyncConsumerGroupsTest.shouldHandleConcurrentJoinAndLeaveOperations fails with IggyResourceNotFoundException in CI


ryerraguntla:
### Bug description

The pre-merge workflow is failing due to a flaky Java SDK test in PR #3515.

Failing run/job:

Run: https://github.com/apache/iggy/actions/runs/28737059115?pr=3515
Job: https://github.com/apache/iggy/actions/runs/28737059115/job/85213039410?pr=3515
Observed failure in logs:

AsyncConsumerGroupsTest > shouldHandleConcurrentJoinAndLeaveOperations() FAILED
java.util.concurrent.ExecutionException at AsyncConsumerGroupsTest.java:345
Caused by: org.apache.iggy.exception.IggyResourceNotFoundException at IggyServerException.java:169
Test summary: 1032 tests completed, 1 failed, 4 skipped
Why this matters
This failure blocks pre-merge CI and appears to be a race/flakiness issue in concurrent join/leave consumer group operations in Java SDK tests.

Expected behavior
shouldHandleConcurrentJoinAndLeaveOperations should pass deterministically under CI load and timing variance.

Actual behavior
During concurrent operations, the test intermittently sees IggyResourceNotFoundException, causing the test task to fail.

Proposed fix
Harden the test against transient race windows / eventual consistency:

Add synchronization to ensure all concurrent tasks are started and completed before final assertions.
Wrap transient IggyResourceNotFoundException in bounded retry during concurrent phase.
Replace immediate final assertions with short bounded polling until state stabilizes.
Keep strict failure for non-transient exceptions.
Acceptance criteria
 Java SDK test suite passes consistently in CI for this scenario.
 AsyncConsumerGroupsTest.shouldHandleConcurrentJoinAndLeaveOperations no longer flakes.
 No regression in consumer group join/leave behavior.
 Pre-merge workflow passes for affected PRs.
Scope
foreign/java/java-sdk tests (primarily AsyncConsumerGroupsTest)
If needed, evaluate server-side consistency assumptions only if test hardening is insufficient.

### Affected area / component

CI / build / tooling

### Deployment

None

### Logs

> Task :iggy:test

AsyncConsumerGroupsTest > shouldHandleConcurrentJoinAndLeaveOperations() FAILED
    java.util.concurrent.ExecutionException at AsyncConsumerGroupsTest.java:345
[...]

PR #3697, opened 2026-07-18
fix(java): Stabilize async consumer group test

ryerraguntla:
Fixes flakiness in the Java SDK async consumer-group integration test for concurrent join/leave operations. The change adds bounded retry handling for transient not-found errors and waits for the consumer-group membership count to converge before asserting, making the test more stable without changing its intent.

Handle transient `IggyResourceNotFoundException` during concurrent join/leave operations in `AsyncConsumerGroupsTest`. This adds bounded retry/backoff helpers and waits for the expected member count before asserting, reducing flakiness from eventually consistent consumer-group state.

Closes # 3641
Relates to # [pr=3515](https://github.com/apache/iggy/actions/runs/28737059115?pr=3515)

## Rationale

To address the flaky java tests.

## What changed?

 This PR stabilizes a flaky Java SDK integration test covering concurrent consumer-group join and leave operations.

The test previously asserted consumer-group state immediately after concurrent operations completed. In some runs, transient IggyResourceNotFoundException errors or short propagation delays caused intermittent failures even though the underlying behavior was correct.

**Changes in this PR**
add retry handling around concurrent joinConsumerGroup calls
add retry handling around concurrent leaveConsumerGroup calls
retry only for transient IggyResourceNotFoundException cases
wait until the consumer group reaches the expected member count before asserting
extract shared helper methods for retry and polling behavior
add reusable timeout and backoff constants for transient-state handling
**Why**
Concurrent async operations can briefly observe unstable or not-yet-visible consumer-group state. This PR makes the test robust to those short-lived conditions while preserving the original behavior being verified.

**Impact**
improves reliability of the Java SDK integration test suite
reduces intermittent failures in consumer-group concurrency coverage
limits scope to test code only
Reviewer-focused summary

**Primary goal:** eliminate flaky failures in AsyncConsumerGroupsTest
**Approach:** add bounded retry logic for transient not-found errors and poll for expected membership state
**Scope:** test-only change in the Java SDK
**Risk:** low, since production code is unchanged
Behavioral effect: assertions are now based on eventual stable state rather than immediate visibility after concurrent operations

## Local Execution

- Passed 
- Pre-commit hooks ran

## AI Usage

1. Which tools?  Claude 
2. Scope of usage?   generated functions and implementation)
3. How did you verify the generated code works correctly? Manually verified


hubcio (dismissed):

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/async/AsyncConsumerGroupsTest.java
```diff
@@ -324,35 +327,79 @@ void shouldHandleConcurrentJoinAndLeaveOperations() throws Exception {
...
-            assertThat(afterLeave.membersCount()).isEqualTo(0);
+            retryOnTransientNotFound(
+                    () -> CompletableFuture.allOf(
+                                    client.consumerGroups().leaveConsumerGroup(STREAM_ID, TOPIC_ID, groupId),
+                                    secondClient.consumerGroups().leaveConsumerGroup(STREAM_ID, TOPIC_ID, groupId),
+                                    thirdClient.consumerGroups().leaveConsumerGroup(STREAM_ID, TOPIC_ID, groupId))
+                            .get(TIMEOUT_SECONDS * 2, TimeUnit.SECONDS),
+                    TRANSIENT_RETRY_TIMEOUT,
+                    TRANSIENT_RETRY_BACKOFF);
+            awaitMembersCount(groupId, 0);
```
slbotbm, line 377: Here, if one leave succeeds while another returns `IggyResourceNotFoundException`, retry resubmits all three leaves. Already-successful client now returns “member not found,” so every later attempt fails. Helper eventually times out even if member count reaches zero. This defeats exact flaky case being fixed.

fix: Treat “already left” as success, retry clients individually, or tolerate leave exceptions and use awaitMembersCount(groupId, 0) as final condition.

slbotbm (changes requested):
other than one comment, lgtm, but let's wait for maciek to approve

mmodzelewski (changes requested):
+1 to @slbotbm's comment: since `retryOnTransientNotFound` re-runs the whole 3-client leave batch, any partial success turns every retry into `ConsumerGroupMemberNotFound` (5006, mapped to `IggyResourceNotFoundException`) for the already-left members, so the helper loops until the deadline and throws `AssertionError` before `awaitMembersCount(0)` is ever reached. The original CI failure in #3641 was exactly in the leave batch, so that flake isn't fixed, it just fails 5 seconds later with a different message.

There is also a second, more fundamental problem: the flake looks like a server-side bug, not eventual consistency. In the failing run the joins were acked, the `membersCount == 3` assertion passed, and only then a leave got "not joined". A durably visible member disappeared between a successful read and the leave, which points at lost state on the server. Candidate mechanism: `client_id` is a hash of `ip:port` (`client_manager.rs`), so ephemeral port reuse across the test suite lets a late disconnect `cleanup_connection` from a previous connection with the same address delete the new client and force-leave its consumer groups (`tcp_listener.rs` / `shard/system/clients.rs`). Retrying in the test only hides that signal from CI.

Suggested path: capture server container logs on failure first. A "Deleted ... client with ID ..." log line between join and leave would confirm the force-removal, and if it does, file a server-side issue for the lost-membership race.

numinnex (dismissed):

mmodzelewski:
@ryerraguntla thanks for digging into this, and sorry for the slow turnaround.

I'm going to close this one. I haven't seen this flake in the CI runs over the
last few weeks, and the tests in their current form already wait at every step,
so the responses they assert on should be stable. Adding retries on top of that
mostly hides a root cause rather than fixing it: as noted above, the failure in
#3641 looked like server-side lost consumer-group membership, not eventual
consistency, and a retry would have swallowed exactly the signal we need.

On top of that, the Java SDK tests now run against the new server, so the
behaviour under test has changed. Let's observe that for a while first, and if
the flake shows up again we can revisit with server logs from the failing run
and fix it where it actually breaks.

PR #3701, opened 2026-07-18
feat(python): add stream listing, update, delete, and purge

### LINKED ISSUE #3520: [python sdk] Add functions related to stream management
### Description

The Python SDK currently exposes only `create_stream(name)` and `get_stream(stream_id)` for stream management, while the Rust SDK also supports listing, updating, deleting, and purging streams.

The missing functions are `get_streams`, `update_streams`, `delete_stream`, and `purge_stream`. These asynchronous functions should be added to `foreign/python/src/client.rs`, wrapping the underlying rust sdk function using PyO3.

These should be tested end-to-end extensively, and the tests should be added to `foreign/python/tests/test_stream.py`. 

The stubs in `apache_iggy.pyi` should be properly generated without any mangling. 

Before opening a PR, please ensure that all pre-commit hooks pass.

### Affected area / component

Python SDK

### Proposed solution

N/A

### Alternatives considered

N/A

### Contribution

### Good first issue

felixfaisal:

Closes #3520

## Rationale

The Python SDK exposed only create_stream and get_stream, so Python users
could not list, rename, delete, or purge streams — operations the Rust SDK
already supports. This closes that parity gap.

## What changed?

Added get_streams, update_stream, delete_stream, and purge_stream to the
Python client by wrapping the Rust StreamClient methods via PyO3, plus a new
Stream wrapper class for the list result. Extended test_stream.py with
end-to-end coverage and regenerated apache_iggy.pyi.

Also drop `pytest-xdist` since it was an unused dependency, and can interfere with clean up that runs after tests.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT)
- Claude Opus (4.8, 1M context) 
- used a skill /teach to understand the bridge between Rust and Python and how the python SDK itself works with iggy, and the changes need to resolve this issue.
- Once done, reviewed edits to each file manually  
- Reviewed the lines written 
- Created the python stub using this command: 
```
cargo run --bin stub_gen 
```
- and then fixed formatting using the below commands: 
```
uv run ruff check --fix apache_iggy.pyi
uv run ruff format apache_iggy.pyi
```
- the resulting file looked as expected
- Ran the iggy server and then the python tests after it, using the commands written below: 
- In one terminal: 
```
cargo run --bin iggy-server -- --with-default-root-credentials --fresh
```
- and in another 
```
uv run --no-sync pytest tests/test_stream.py -v
```
- all the tests pass 
Yup


foreign/python/tests/test_stream.py
```diff
@@ -203,3 +203,381 @@ async def test_create_stream_before_login_fails(self, unique_name):
...
+        created = [unique_name(f"z{index}") for index in range(3, 0, -1)]
+        for name in created:
+            await iggy_client.create_stream(name)
+
+        streams = await iggy_client.get_streams()
+        by_name = {stream.name: stream for stream in streams}
+        assert set(created).issubset(by_name)
+
+        mine = [by_name[name] for name in created]
+        assert [stream.id for stream in mine] == sorted(stream.id for stream in mine)
```
ryankert01: It occurs to me that it's not testing `get_streams()` ordering since it's constructed based on created ordering. Although I'm also not quite sure whether if iggy wants to ensure the ordering or not in the distributed future, kafka for example, don't.

ryankert01 (commented):
Overall lg to me, a comment.

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,39 @@ impl StreamDetails {
...
+
+impl From<RustStream> for Stream {
+    fn from(stream: RustStream) -> Self {
+        Self { inner: stream }
+    }
+}
+
+#[gen_stub_pymethods]
+#[pymethods]
+impl Stream {
```
slbotbm, line 98: Let's add getters for created_at and size as well.

foreign/python/src/client.rs
```diff
@@ -168,6 +168,78 @@ impl IggyClient {
         })
     }
 
+    /// Gets all streams.
+    /// Returns a list of streams or a PyRuntimeError on failure.
```
slbotbm: Change to 
```
/// Return all streams visible to the authenticated user.
///
/// The result is ordered by ascending numeric stream ID.
///
/// Returns:
///     A list of `Stream` summaries.
///
/// Raises:
///     RuntimeError: If the client is not authenticated, the user lacks global
///         `read_streams` or `manage_streams` permission, or the request fails.
```

foreign/python/src/client.rs
```diff
@@ -168,6 +168,78 @@ impl IggyClient {
...
+            let streams = inner
+                .get_streams()
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(streams.into_iter().map(Stream::from).collect::<Vec<_>>())
+        })
+    }
+
+    /// Updates a stream's name by id.
+    /// Returns Ok(()) on successful update or a PyRuntimeError on failure.
```
slbotbm: Change to 
```
/// Rename a stream selected by name or numeric ID.
///
/// `stream_id` accepts a stream name as `str` or numeric ID as `int`. A
/// decimal-only string is interpreted as a numeric ID. `name` must be unique
/// and contain between 1 and 255 UTF-8 bytes. Renaming a stream to its current
/// name succeeds without changing it.
///
/// Returns:
///     None.
///
/// Raises:
///     TypeError: If `stream_id` is neither `str` nor `int`, or `name` is not
///         `str`.
///     OverflowError: If an integer identifier is outside `0..=2**32 - 1`.
///     ValueError: If a string identifier is empty or exceeds 255 UTF-8 bytes.
///     RuntimeError: If the client is not authenticated, the user lacks global
///         `manage_streams` or per-stream `manage_stream` permission, the
///         stream does not exist, the new name is invalid or already used, or
///         the request fails.
```

foreign/python/src/client.rs
```diff
@@ -168,6 +168,78 @@ impl IggyClient {
...
+            inner
+                .delete_stream(&stream_id)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(())
+        })
+    }
+
+    /// Purges all messages from a stream by id.
+    /// Returns Ok(()) on successful purge or a PyRuntimeError on failure.
```
slbotbm: Change to 
```
/// Delete all messages from every topic in a stream.
///
/// The stream, topics, and partitions remain available. Repeated purges of an
/// existing empty stream succeed. `stream_id` accepts a stream name as `str`
/// or numeric ID as `int`. A decimal-only string is interpreted as a numeric
/// ID.
///
/// Returns:
///     None.
///
/// Raises:
///     TypeError: If `stream_id` is neither `str` nor `int`.
///     OverflowError: If an integer identifier is outside `0..=2**32 - 1`.
///     ValueError: If a string identifier is empty or exceeds 255 UTF-8 bytes.
///     RuntimeError: If the client is not authenticated, the user lacks global
///         `manage_streams` or per-stream `manage_stream` permission, the
///         stream does not exist, or the request fails.
```

foreign/python/src/client.rs
```diff
@@ -168,6 +168,78 @@ impl IggyClient {
...
+            inner
+                .update_stream(&stream_id, &name)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(())
+        })
+    }
+
+    /// Deletes a stream by id.
+    /// Returns Ok(()) on successful deletion or a PyRuntimeError on failure.
```
slbotbm: Change to 
```
/// Delete a stream selected by name or numeric ID.
///
/// Deletion removes the stream and all of its topics, partitions, and messages.
/// `stream_id` accepts a stream name as `str` or numeric ID as `int`. A
/// decimal-only string is interpreted as a numeric ID.
///
/// Returns:
///     None.
///
/// Raises:
///     TypeError: If `stream_id` is neither `str` nor `int`.
///     OverflowError: If an integer identifier is outside `0..=2**32 - 1`.
///     ValueError: If a string identifier is empty or exceeds 255 UTF-8 bytes.
///     RuntimeError: If the client is not authenticated, the user lacks global
///         `manage_streams` or per-stream `manage_stream` permission, the
///         stream does not exist, or the request fails.
```

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,39 @@ impl StreamDetails {
         self.inner.topics_count
     }
 }
+
```
slbotbm, line 79: Add the following doc:
```
/// Summary information returned by `IggyClient.get_streams()`.
///
/// `created_at` is Unix time in microseconds. `size_bytes` is the stream's
/// current stored size in bytes.
```

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,39 @@ impl StreamDetails {
...
+    fn from(stream: RustStream) -> Self {
+        Self { inner: stream }
+    }
+}
+
+#[gen_stub_pymethods]
+#[pymethods]
+impl Stream {
+    #[getter]
+    pub fn id(&self) -> u32 {
```
slbotbm, line 101: Doc: `/// Numeric stream identifier.`

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,39 @@ impl StreamDetails {
...
+        self.inner.id
+    }
+
+    #[getter]
+    pub fn name(&self) -> String {
```
slbotbm, line 113: Doc: `/// Unique stream name.`

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,39 @@ impl StreamDetails {
...
+        self.inner.name.to_string()
+    }
+
+    #[getter]
+    pub fn messages_count(&self) -> u64 {
```
slbotbm, line 125: Doc: `/// Total messages across all topics in the stream.`

foreign/python/src/stream.rs
slbotbm: Also write their docs as well.

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,39 @@ impl StreamDetails {
...
+        self.inner.messages_count
+    }
+
+    #[getter]
+    pub fn topics_count(&self) -> u32 {
```
slbotbm, line 131: Doc: `/// Number of topics in the stream.`

foreign/python/tests/test_stream.py
```diff
@@ -203,3 +203,381 @@ async def test_create_stream_before_login_fails(self, unique_name):
...
+        """Test get_streams requires an established connection."""
+        host, port = get_server_config()
+        client = IggyClient(f"{host}:{port}")
+
+        with pytest.raises(RuntimeError):
+            await client.get_streams()
+
+    @pytest.mark.asyncio
+    async def test_get_streams_before_login_fails(self):
+        """Test get_streams requires authentication."""
```
slbotbm: Make these into 1 test

foreign/python/tests/test_stream.py
```diff
@@ -203,3 +203,381 @@ async def test_create_stream_before_login_fails(self, unique_name):
...
+
+        renamed = await iggy_client.get_stream(new_name)
+        assert renamed is not None
+        assert renamed.name == new_name
+
+        old = await iggy_client.get_stream(stream_name)
+        assert old is None
+
+    @pytest.mark.asyncio
+    async def test_update_stream_preserves_id(
```
slbotbm: These two tests can be combined into one. No need for another test just to assert stream id remains same.

foreign/python/tests/test_stream.py
```diff
@@ -203,3 +203,381 @@ async def test_create_stream_before_login_fails(self, unique_name):
...
+    ):
+        """Test update_stream rejects renaming a stream to a name already in use."""
+        first_stream = unique_name()
+        second_stream = unique_name()
+
+        await iggy_client.create_stream(first_stream)
+        await iggy_client.create_stream(second_stream)
+
+        with pytest.raises(RuntimeError):
+            await iggy_client.update_stream(stream_id=second_stream, name=first_stream)
```
slbotbm, line 440: After the failure, assert that none of the metadata of the stream has changed.

foreign/python/tests/test_stream.py
```diff
@@ -203,3 +203,381 @@ async def test_create_stream_before_login_fails(self, unique_name):
...
+        wait_for_server(host, port)
+
+        client = IggyClient(f"{host}:{port}")
+        await client.connect()
+
+        with pytest.raises(RuntimeError):
+            await client.get_streams()
+
+
+class TestUpdateStream:
```
slbotbm, line 288: Add the following tests for update_stream:
- it accepts exact 1-byte and 255-byte names and rejects empty and 256-byte names, including UTF-8 byte boundaries.

slbotbm:
Please do not force push. Instead, create new commits with your work. This makes it easier for me to understand the changes

/author

slbotbm:
Since the original author has stopped responding, I'll take over and complete this PR.

slbotbm:
I added test clean up code as well. 

/ready

foreign/python/tests/conftest.py
```diff
@@ -37,9 +35,32 @@
...
 
-@pytest.fixture(scope="session")
-async def iggy_client() -> IggyClient:
+    users = await client.get_users()
+    for user in users:
+        if user.id != _ROOT_USER_ID:
+            await client.delete_user(user.id)
+
+
+@pytest.fixture(scope="session", autouse=True)
```
ethanlin01x: I had a look at the new cleanup fixture and I have a few thoughts about it.

Right now the cleanup only happens once, at the end of the whole session. That means streams from earlier tests are still on the server while later tests run. (ex. `test_get_streams_returns_created_streams` has to filter `get_streams()` down to its own names )

The `autouse=True` on `iggy_client` makes every session connect to a server, but `test_client_config.py` has four `@pytest.mark.unit` tests that are meant to run without one, so `pytest -m unit` will fail now.

And I guess `pytest-xdist` had to go because deleting everything cannot work with parallel workers. 

What do you think about cleaning up after each test instead? Every stream name already comes from `unique_name`, so it could just remember the names it hands out, and then no test would need to change:

```python
@pytest.fixture
def issued_names() -> set[str]:
    """Names created by `unique_name` in one test."""
    return set()


@pytest.fixture
def unique_name(issued_names: set[str]):
    def make_name(prefix: str = "", *, min_bytes: int = 8, max_bytes: int = 255) -> str:
        # existing validation and suffix generation unchanged
        name = f"{prefix}{suffix}"
        issued_names.add(name)
        return name

    return make_name


@pytest.fixture(autouse=True)
async def cleanup_streams(request, issued_names: set[str]) -> AsyncGenerator[None, None]:
    """Delete the streams this test created, even if the test failed."""
    yield

    if not issued_names:
        return

    client = request.getfixturevalue("iggy_client")
    for name in issued_names:
        try:
            await client.delete_stream(name)
        except RuntimeError:
            # The name was used for a topic or a user, or the test already deleted it.
            pass
```

`iggy_client` could go back to `scope="session"` without `autouse` as well.

slbotbm: Thanks -- applied

slbotbm: Since `unique_names` can create invalid identifier names as well, the clean up was failing since it was trying to delete invalid stream identifiers. The latest change instead removes the `set()` config and instead runs the clean up after every test. It does this by running `client.get_streams()` and then cleaning up the streams. For the unit tests, the server is not required since I added a check -- if `iggy_client` is not in request.fixturenames, the clean up is skipped.

ethanlin01x: Thanks for explaining, that makes sense.

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,59 @@ impl StreamDetails {
...
+impl Stream {
+    /// Numeric stream identifier.
+    #[getter]
+    pub fn id(&self) -> u32 {
+        self.inner.id
+    }
+
+    /// Stream creation time as Unix time in microseconds.
+    #[getter]
+    pub fn created_at(&self) -> u64 {
```
ethanlin01x, line 107: nit: `StreamDetails` has neither `created_at` nor `size_bytes`, so after this PR `get_stream() `returns fewer fields than `get_streams().` Maybe add them for symmetry.

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,59 @@ impl StreamDetails {
...
+
+    /// Unique stream name.
+    #[getter]
+    pub fn name(&self) -> String {
+        self.inner.name.to_string()
+    }
+
+    /// Current stored stream size in bytes.
+    #[getter]
+    pub fn size_bytes(&self) -> u64 {
```
ethanlin01x: nit: `Topic`, `TopicDetails` and `Partition` all return bytes from a getter named size. Rename this one to match?

ethanlin01x (approved):
LGTM, just some nits

foreign/python/src/client.rs
```diff
@@ -500,6 +500,132 @@ impl IggyClient {
...
+    /// and contain between 1 and 255 UTF-8 bytes. Renaming a stream to its current
+    /// name succeeds without changing it.
+    ///
+    /// Returns:
+    ///     None.
+    ///
+    /// Raises:
+    ///     TypeError: If `stream_id` is neither `str` nor `int`, or `name` is not
+    ///         `str`.
+    ///     OverflowError: If an integer identifier is outside `0..=2**32 - 1`.
```
hubcio: also at lines 573 and 607 (and the same three spots in the stub).

`OverflowError` never fires here. `PyIdentifier` is a derived enum, so an int outside `0..=2**32 - 1` (or `-1`, `1.5`, `None`) raises `TypeError: failed to extract enum PyIdentifier ('str | int')`, at call time. fold it into the `TypeError` line.

foreign/python/src/stream.rs
hubcio: every other byte-size getter is `size` (`Topic`, `TopicDetails`, `Partition`). `Stream` is new, so rename it now while it's free: this getter, the docstring above, the three uses in `test_stream.py`, then regenerate the stub.

foreign/python/pyproject.toml
```diff
@@ -102,7 +102,6 @@ testing = [
     "pytest>=9.1.1,<10.0",
     "pytest-asyncio>=0.24.0,<2.0",
     "pytest-cov>=4.0,<8.0",
-    "pytest-xdist>=3.0,<4.0",
     "pytest-timeout>=2.0,<3.0",
```
hubcio, line 105: dropping `pytest-xdist` isn't mentioned anywhere in the PR. if it's because `cleanup_streams` would delete other workers' streams under `-n`, say so in the description. also `pylock.toml` still lists `pytest-xdist` and `execnet`, so either regenerate it or keep the dep.

foreign/python/apache_iggy.pyi
```diff
@@ -2055,8 +2179,8 @@ class Topic:
         r"""
         Options admission resolved for the keys the client did not send.
 
-        Same shape as `options`. These would have resolved differently under
-        another server configuration.
+        Same shape as [`Self::options`]. These would have resolved differently
```
hubcio: also at 2247.

`[`Self::options`]` is rustdoc link syntax leaking into the python stub. the source is `topic.rs:257` and `:348`; change them to plain `options` (master's stub wording) and regenerate, and this hunk goes away.

foreign/python/src/client.rs
```diff
@@ -500,6 +500,132 @@ impl IggyClient {
         })
     }
 
+    /// Return all streams visible to the authenticated user.
```
hubcio: the server doesn't filter per stream: you need global `read_streams` or `manage_streams`, otherwise it's unauthorized. "all streams" is the accurate wording.

foreign/python/src/client.rs
```diff
@@ -500,6 +500,132 @@ impl IggyClient {
...
+    ///
+    /// The result is ordered by ascending numeric stream ID.
```
hubcio: the order comes from the rust sdk sorting client-side in `streams_from_wire`, and neither the `StreamClient` trait nor any other sdk promises it. drop the sentence, or add the same promise to the trait doc.

foreign/python/src/client.rs
```diff
@@ -500,6 +500,132 @@ impl IggyClient {
...
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(streams.into_iter().map(Stream::from).collect::<Vec<_>>())
+        })
+    }
+
+    /// Rename a stream selected by name or numeric ID.
+    ///
+    /// `stream_id` accepts a stream name as `str` or numeric ID as `int`. A
+    /// decimal-only string is interpreted as a numeric ID. `name` must be unique
+    /// and contain between 1 and 255 UTF-8 bytes. Renaming a stream to its current
```
hubcio, line 527: documented but not tested. `test_update_stream_to_existing_name_fails` already snapshots the metadata before/after; the same test with `name=stream_name` and no `pytest.raises` covers it.

foreign/python/tests/test_stream.py
```diff
@@ -203,3 +203,425 @@ async def test_create_stream_before_login_fails(self, unique_name):
...
+        """Test update_stream accepts names at the UTF-8 byte boundaries."""
+        stream_name = unique_name()
+        await iggy_client.create_stream(stream_name)
+
+        await iggy_client.update_stream(stream_id=stream_name, name=new_name)
+        try:
+            renamed = await iggy_client.get_stream(new_name)
+            assert renamed is not None
+            assert renamed.name == new_name
+        finally:
```
hubcio: this `finally` is leftover from the old name-based cleanup. `cleanup_streams` deletes by id and rename keeps the id, so it's covered. drop the try/finally.

foreign/python/tests/conftest.py
```diff
@@ -108,6 +107,26 @@ def make_name(
...
+async def cleanup_streams(
+    request: pytest.FixtureRequest,
+) -> AsyncGenerator[None, None]:
+    """Delete streams created by the current test."""
+    if "iggy_client" not in request.fixturenames:
+        yield
+        return
+
+    client: IggyClient = request.getfixturevalue("iggy_client")
+    existing_stream_ids = {stream.id for stream in await client.get_streams()}
```
hubcio: ids get reused after delete (lowest free slot), so a test that deletes a pre-existing stream and then creates one can inherit the id and dodge cleanup. nothing does that today, but `(stream.id, stream.created_at)` closes it for free.

foreign/python/src/stream.rs
```diff
@@ -56,3 +56,59 @@ impl StreamDetails {
         self.inner.topics_count
     }
 }
+
+/// Summary information returned by `IggyClient.get_streams()`.
```
hubcio, line 80: `StreamDetails` a few lines up still lacks `created_at`, `size` and `topics`, all present on the rust side, so `get_stream()` now returns less than a `get_streams()` item. worth adding here or in a follow-up.

foreign/python/src/client.rs
```diff
@@ -500,6 +500,132 @@ impl IggyClient {
...
+    ///     TypeError: If `stream_id` is neither `str` nor `int`, or `name` is not
+    ///         `str`.
+    ///     OverflowError: If an integer identifier is outside `0..=2**32 - 1`.
+    ///     ValueError: If a string identifier is empty or exceeds 255 UTF-8 bytes.
+    ///     RuntimeError: If the client is not authenticated, the user lacks global
+    ///         `manage_streams` or per-stream `manage_stream` permission, the
+    ///         stream does not exist, the new name is invalid or already used, or
+    ///         the request fails.
+    #[gen_stub(override_return_type(type_repr="collections.abc.Awaitable[None]", imports=("collections.abc")))]
+    fn update_stream<'a>(
```
hubcio, line 550: `update_topic` takes `options`, this doesn't. fine today since there are no updatable stream keys yet, but the first one will need a signature change here. intentional?

hubcio (changes requested):

foreign/python/src/stream.rs
slbotbm: Done

foreign/python/src/stream.rs
slbotbm: Done

foreign/python/src/client.rs
slbotbm: done

foreign/python/src/stream.rs
slbotbm: already done

foreign/python/pyproject.toml
slbotbm: dropped dep, regenerated pylock.toml

foreign/python/apache_iggy.pyi
slbotbm: Done

foreign/python/src/client.rs
slbotbm: Done

foreign/python/src/client.rs
slbotbm: dropped

foreign/python/src/client.rs
slbotbm: Done

foreign/python/tests/test_stream.py
slbotbm: Done

foreign/python/tests/conftest.py
slbotbm: Done

foreign/python/src/stream.rs
slbotbm: done

foreign/python/src/client.rs
slbotbm: added options. this pr has been open for a long time, and at the time of creation, there was no options field in update_stream, so the changes drifted.

ryankert01 (approved):
lg, thanks

hubcio:
@felixfaisal could you please rebase this PR?

slbotbm:
@hubcio done

spetz:
I did resolve the conflicts.

hubcio (approved):

spetz (approved):

PR #3708, opened 2026-07-20
perf(common): decode user headers without copying the blob

### LINKED ISSUE #3706: perf(common): avoid copying user headers blob when decoding message view
### Description

[`IggyMessageView::user_headers_map`](https://github.com/apache/iggy/blob/6f1d5485d06c3c43afd2510e432b2f7c1a3ac41a/core/common/src/types/message/message_view.rs#L88-L105) calls `WireUserHeaders::from_slice`, which validates the buffer and clones the whole headers blob into an owned `Bytes`. The decoder that follows only iterates those bytes, copying each key and value out individually.

The view already holds the headers as a borrowed slice, so the intermediate clone doesn't appear to buy anything. The blob can be up to `MAX_USER_HEADERS_SIZE` (100 KB).

### Affected area / component

Performance, Rust SDK (`core/common`).

### Proposed direction

Validate the borrowed slice in place and decode from it. `validate_user_headers` and `WireUserHeaderIterator` are already public in `iggy_binary_protocol`, so no API needs widening.

Error behaviour must be preserved: structurally invalid returns `Ok(None)` with a warning, unknown header kind returns `Err`. Borrowing the individual keys and values too would remove the remaining copies, but would pin the message buffer for as long as any header is retained, which isn't worth it at the SDK boundary.

### Contribution

### Good first issue

ryankert01:

Closes #3706

## Rationale

User headers are decoded once per consumed message, and the current path copies the same bytes twice. The blob can be up to `MAX_USER_HEADERS_SIZE` (100 KB).

## What changed?

`IggyMessageView::user_headers_map` called `WireUserHeaders::from_slice`, which validates the buffer and then clones the whole headers blob into an owned `Bytes`; the decoder that followed copied every key and value back out of that clone. The view already holds those bytes as a borrowed slice and the decoder only iterates them, so the clone was redundant.

It now validates in place and decodes from the borrowed slice, with `user_headers_from_wire` delegating to the same loop so there is one implementation. `validate_user_headers` and `WireUserHeaderIterator` were already public, so no new API surface. Error partitioning is unchanged - structurally invalid returns `Ok(None)` with the same warning, unknown kind codes return `Err` - and keys and values are still copied into owned `Bytes`, so there is no retention or lifetime change.

Criterion vs `master`: -22.6% (1 hdr x 16 B), -17.6% (4 hdr x 200 B), -6.4% (4 hdr x 16 B), -4.1% (16 hdr x 64 B), -5.1% (32 hdr x 200 B). The win tracks total header bytes and is diluted by header count, since the per-field copies remain.

## Local Execution

- Tests Passed
- Pre-commit hooks - all passed 

## AI Usage

1. **Which tools?** Claude (Claude Code).
2. **Scope of usage?** Wrote the implementation, tests and benchmarks.
3. **How did you verify the generated code works correctly?** Unit tests, clippy, workspace `cargo check`, and criterion benchmarks.
4. **Can you explain every line of the code if asked?** Yes.

### DIFF
--- a/core/common/src/types/message/message_view.rs
+++ b/core/common/src/types/message/message_view.rs
@@ -22,9 +22,9 @@ use crate::IggyByteSize;
 use crate::Sizeable;
 use crate::error::IggyError;
 use crate::utils::checksum;
-use crate::wire_conversions::user_headers_from_wire;
+use crate::wire_conversions::user_headers_from_validated_slice;
 use crate::{HeaderKey, IggyMessageHeaderView};
-use iggy_binary_protocol::WireUserHeaders;
+use iggy_binary_protocol::validate_user_headers;
 use std::collections::BTreeMap;
 use std::num::NonZeroUsize;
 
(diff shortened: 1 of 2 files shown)


slbotbm:
The C++ e2e test is a flaky one whose fix is in PR https://github.com/apache/iggy/pull/3691, so after that merges, the CI should be green

numinnex (approved):

mmodzelewski (approved):

PR #3711, opened 2026-07-20
test(cpp): poll for session removal instead of asserting once

### LINKED ISSUE #3710: Flaky C++ e2e tests: GetClientsReflectsSessionRemovalAfter{Disconnect,Shutdown}
### Description

Two C++ e2e tests fail intermittently: `LowLevelE2E_Client.GetClientsReflectsSessionRemovalAfterDisconnect` and `...AfterShutdown`.

Across the last 60 `Pre-merge` runs, 14 failed. **11 of those failed on the `C++ • e2e` job, and every one was one of these two tests** (5x Disconnect, 6x Shutdown), 278/279 passing each time. They failed on 7 unrelated branches (connectors, docker image, kafka gateway, metadata, SDK), so the trigger is environmental rather than any of those changes.

```
low_level_e2e.cpp:439: Failure
Value of: found_first
  Actual: true; Expected: false
```

### Root cause

Both tests disconnect one client and then, over a second connection, assert **once immediately** that the session is already absent from `get_clients()`. Nothing guarantees that ordering:

- `disconnect`/`shutdown` closes the socket without a command or ack ([tcp_client.rs#L675](https://github.com/apache/iggy/blob/master/core/sdk/src/tcp/tcp_client.rs#L675)).
- The server removes the session only when the first connection's read loop observes EOF and runs `cleanup_connection` ([tcp_listener.rs#L159](https://github.com/apache/iggy/blob/master/core/server/src/tcp/tcp_listener.rs#L159)), with no happens-before edge to the second connection's query.

The adjacent `GetClientsReflectsLoggedOutSessionAsUnauthenticated` does not flake, because logout is a command with a response and is observable as soon as the reply returns.

I could not reproduce it locally in 700 iterations against the Rust SDK, consistent with a narrow timing window that widens on loaded CI runners.

### Possibly related (not confirmed)

`client_id` is `xxHash32` of the source `IP:port` ([client_manager.rs#L63](https://github.com/apache/iggy/blob/master/core/server/src/streaming/clients/client_manager.rs#L63)). Since ephemeral ports are recycled by the OS, IDs are not unique over a session lifecycle, and `insert` overwrites on collision. That could also present as a removed session still appearing. Noting it as a lead only.

### Proposed fix

Make the two tests poll `get_clients()` until the session disappears, with a short timeout, instead of asserting once. Server behaviour is unchanged.

### Affected area / component

C++ SDK, CI / build / tooling.

ryankert01:

Closes #3710

## Rationale

`GetClientsReflectsSessionRemovalAfter{Disconnect,Shutdown}` are the single most common CI flake: of the last 60 `Pre-merge` runs, 14 failed, and 11 of those were exactly these two tests, across 7 unrelated branches.

## What changed?

Both tests disconnected a client and then asserted, over a second connection, that the session was already absent from `get_clients()`. Removal is server-side and happens when the server observes the closed connection ([tcp_listener.rs#L159](https://github.com/apache/iggy/blob/master/core/server/src/tcp/tcp_listener.rs#L159)), which can lag the client's `disconnect`/`shutdown` call. The single immediate assertion therefore raced.

They now poll `get_clients()` until the id disappears, with a 5s timeout, via a small `WaitForClientRemoved` helper. Server behaviour is unchanged; only the test synchronisation is fixed.

## Local Execution

- The change is test-only. I could not build the C++ SDK locally (no bazel toolchain set up here), so I have **not** run it locally and am relying on CI to exercise it - flagging that explicitly rather than claiming a local run I did not do.
- The diagnosis is backed by a study of the last 60 `Pre-merge` runs (11/14 recent failures were these two tests, on unrelated branches) and by reading the server-side session-cleanup path.

## AI Usage

1. **Which tools?** Claude (Claude Code).
2. **Scope of usage?** Diagnosed the flake from CI history and wrote the test change.
3. **How did you verify the generated code works correctly?** Reviewed the diff and the server cleanup path; the C++ build/tests were not run locally (no toolchain) and rely on CI.
4. **Can you explain every line of the code if asked?** Yes.

### DIFF
--- a/foreign/cpp/tests/client/low_level_e2e.cpp
+++ b/foreign/cpp/tests/client/low_level_e2e.cpp
@@ -20,9 +20,11 @@
 // TODO(slbotbm): Add tests for store_consumer_offset, get_consumer_offset, and delete_consumer_offset functions
 // attached to client after implementing consumer group functions
 // TODO(slbotbm): Add tests for update_permissions after creating create_user, get_user, etc. functions
+#include <chrono>
 #include <cstddef>
 #include <cstdint>
 #include <string>
+#include <thread>
 #include <unordered_set>
 
 #include <gtest/gtest.h>
@@ -31,6 +33,30 @@
 #include "tests/common/test_helpers.hpp"
 
 class LowLevelE2E_Client : public E2ETestFixture {};
+
+namespace {
+// A session is removed server-side when the server observes the closed
+// connection, which can lag the client's disconnect/shutdown call. Poll
+// get_clients from the observing connection until the id is gone rather than
+// asserting once immediately.
+bool WaitForClientRemoved(iggy::ffi::Client *observer, std::uint32_t client_id) {
+    const auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(5);
+    do {
+        bool present = false;
+        for (const auto &client : observer->get_clients()) {
+            if (client.client_id == client_id) {
+                present = true;
+                break;
+            }
+        }
+        if (!present) {
+            return true;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(50));
+    } while (std::chrono::steady_clock::now() < deadline);
+    return false;
+}
+}  // namespace
 
 TEST_F(LowLevelE2E_Client, ConnectAndLogin) {
     RecordProperty("description", "Connects and logs in successfully using each supported connection string format.");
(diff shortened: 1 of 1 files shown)


slbotbm:
@ryankert01 The fix is already in https://github.com/apache/iggy/pull/3691, so this can be closed. Thanks for the PR, though.

ryankert01:
thx for the reminder @slbotbm

Issue #3715, opened 2026-07-21
feat(connectors): add Apache Airflow trigger connector


avirajkhare00:
### Description

Add an Apache Airflow connector so Iggy can drive event-driven DAGs via the Airflow REST API.

This is tracked on the connector ecosystem roadmap (#2753) under **Workflow & Orchestration**:

| Target | Type | Priority | Comp | Suggested stack |
|--------|------|:--------:|:----:|-----------------|
| [Apache Airflow](https://airflow.apache.org/) | Trigger | P4 | 1/4 | `reqwest` (REST API) |

Roadmap intent: **Iggy sensor/trigger for event-driven DAGs**. There is no Airflow sink/source (or trigger plugin) in-tree today, no dedicated implementation issue before this one, and no open PR claiming the work.

**Motivation**

- Airflow is the industry-standard workflow orchestrator; pairing it with Iggy enables stream-driven DAG triggers instead of polling-only sensors.
- Existing connectors cover DBs, search, lakehouse, and HTTP egress, but nothing in the workflow/orchestration category is implemented yet (Airflow, Temporal, Prefect, Dagster are all still roadmap-only).
- A first cut can likely reuse patterns from `http_sink` (auth, retries, batching) while specializing for Airflow's trigger/DAG-run APIs.

**Suggested v1 scope (trigger)**

- Consume messages from configured Iggy streams/topics
- Trigger DAG runs via Airflow REST API (e.g. create DAG run / trigger endpoint)
- Config: Airflow base URL, auth (token/basic), `dag_id`, optional conf/payload mapping from message body, retry policy
- Map transient HTTP errors (5xx, timeouts) to retry; map permanent client errors (auth, missing DAG) to non-retry
- `open()` connectivity check against Airflow health/version endpoint; structured logging of trigger outcomes
- Unit tests + example config under `core/connectors/runtime/example_config/connectors/`

Out of scope for v1 (unless someone wants to expand): full Airflow **source** (task/DAG state → Iggy), Prefect/Dagster parity, Airflow provider package on the Airflow side.

### Affected area / component

Connectors

### Proposed solution

New sink-style (or dedicated trigger) plugin, e.g. `airflow_sink` / `airflow_trigger`, under `core/connectors/sinks/`, built on `iggy_connector_sdk` and `reqwest`, following existing sink lifecycle (`open` / `consume` / `close`) and config conventions.

### Alternatives considered

[...]

PR #3716, opened 2026-07-21
feat(connectors): add Apache Airflow trigger sink

avirajkhare00:
## Summary

- Add `iggy_connector_airflow_sink` that triggers Airflow DAG runs (`POST /api/v1/dags/{dag_id}/dagRuns`) from Iggy **poll batches** (one run per batch, not per message)
- `conf.messages` carries the batch; optional `conf.iggy` metadata; deterministic batch `dag_run_id`; HTTP **409** treated as success
- Auth modes `none` / `basic` / `bearer`; optional `dag_id_header` groups a poll into one run per DAG id
- Permanent 4xx fails the batch (no silent drop); WireMock + real Airflow e2e; tracked by #3715

## Design note

Airflow runs are jobs, not stream events. High-traffic 1:1 message→run would overwhelm the scheduler. Volume is controlled via stream `batch_length` / `poll_interval`.

## Test plan

- [x] `cargo fmt` / `cargo sort --no-format --workspace` / clippy on the sink crate
- [x] `cargo test -p iggy_connector_airflow_sink` (12 unit tests)
- [x] `cargo test -p integration -- connectors::airflow` (WireMock: 3 messages → 1 DAG-run POST)
- [x] Local e2e: Iggy server + connectors + Airflow 2.10.5 standalone (5 messages → 1 run `iggy-0-0-4-5`, re-POST 409)

## Review follow-up

Addressed @kriti-sc feedback: batch trigger default, explicit idempotent `dag_run_id`, fail batch on permanent HTTP errors.


core/connectors/runtime/example_config/connectors/airflow_sink.toml
```diff
@@ -0,0 +1,53 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
kriti-sc, line 1: I feel like this sink is an anti-pattern because Airflow is meant for the batch processing usecase while Iggy is a streaming system. More so, because a lot of other Iggy sinks also operate on message batch level and not individual message level.

This sink creates a dag run for each Iggy message, so for a high-traffic topics, it will end up overwhelming Airflow. 

Two solutions:
1. implement an `apply_function` like the Kafka support in Airflow. That way the sink can be configured to trigger dag only when a message passes that filter.
2. Consider triggering Airflow on the batch and not individual messages.

core/connectors/sinks/airflow_sink/README.md
```diff
@@ -0,0 +1,127 @@
...
+stable REST API (`POST /api/v1/dags/{dag_id}/dagRuns`).
+
+## Overview
+
+| | |
+| --- | --- |
+| **Type** | Sink (trigger) |
+| **Direction** | Iggy → Airflow |
+| **API** | Airflow REST (`api_prefix` default `/api/v1`) |
+| **Idempotency** | Deterministic `dag_run_id`; HTTP **409** treated as success |
```
kriti-sc: My Airflow knowledge maybe outdated, but afaik `dag_run_id` would be `manual__timestamp` which wouldn't be idempotant.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,984 @@
...
+                    );
+                    failures += 1;
+                    consecutive_failures += 1;
+                    last_error = Some(error);
+                    if consecutive_failures >= MAX_CONSECUTIVE_FAILURES {
+                        let processed = triggered + failures;
+                        let skipped = (total as u64).saturating_sub(processed);
+                        error!(
+                            "{CONNECTOR_NAME} ID: {} aborting batch after {} consecutive failures \
+                             ({} remaining messages skipped)",
```
kriti-sc: I am not sure about skipping messages here.  
Classifying errors into 3 parts-
1. Transient Airflow related - Might not apply to the remainder of messages in the batch.
2. Malformed Iggy message - Same as above.
3. Permanent error such as auth or Airflow down - Might be better to fail the sink.

Feel free to add your reasoning or cases I might have missed.

kriti-sc (changes requested):

core/connectors/runtime/example_config/connectors/airflow_sink.toml
avirajkhare00: Agreed — one DAG run per message is the wrong unit of work for Airflow.

Pushed a redesign: **one DAG run per consume poll batch** (not per message).

- `conf.messages` is an array of `{offset, id, timestamp, payload}` for the batch
- Optional `conf.iggy` carries stream/topic/partition + first/last offset + message_count
- Stream `batch_length` / `poll_interval` control volume; `batch_length = 1` covers low-rate command topics
- If `dag_id_header` yields multiple DAG ids in one poll, we still do **one run per DAG-id group**, not one run per message

Verified with unit tests, WireMock integration (3 msgs → 1 POST), and real Airflow 2.10.5 e2e (5 msgs → 1 run `iggy-0-0-4-5`).

In-sink `apply_function`-style filtering is still out of scope for this PR; selective triggering is better via a dedicated topic (or a later transform). Happy to discuss if you want that in a follow-up.

core/connectors/sinks/airflow_sink/README.md
avirajkhare00: Good catch on the default Airflow form — that would only apply if we omitted `dag_run_id`.

We always set an explicit deterministic id, so Airflow does **not** generate `manual__timestamp`:

- **Before (per-message):** `iggy-{partition}-{offset}-{message_id_hex}`
- **Now (batch):** `iggy-{partition}-{first_offset}-{last_offset}-{message_count}`

Same poll range → same id on redelivery; Airflow returns **409**, which we treat as success. Documented in the README with the request shape and a note that we never rely on auto-generated run ids.

core/connectors/sinks/airflow_sink/src/lib.rs
avirajkhare00: Agreed the old per-message skip path was too loose for permanent failures.

With batch triggering the consume path is simpler:

1. **Transient (429/5xx/network):** middleware retries, then `Err` so the runtime can redeliver the same batch (same `dag_run_id` → 409 if already created).
2. **Malformed single payload:** skip that entry when building `conf.messages`, count an error; if every entry fails conversion, return error for the batch.
3. **Permanent HTTP (401/403/404/400/422, auth/missing DAG, etc.):** return `PermanentHttpError` for the **whole batch** — do not silent-drop and advance as if the run was created.

Auth / Airflow-down style permanent failures no longer drop messages and continue.

slbotbm:
@avirajkhare00 looks like your lock file is stale which is leading to CI failures.

hubcio:
I'll check this later this week.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+/// Explicitly set on create — Airflow does **not** auto-assign `manual__timestamp`
+/// when `dag_run_id` is provided.
+fn build_batch_dag_run_id(
+    partition_id: u32,
+    first_offset: u64,
+    last_offset: u64,
+    message_count: usize,
+) -> String {
+    format!("iggy-{partition_id}-{first_offset}-{last_offset}-{message_count}")
+}
```
slbotbm, line 744: The run ID leaves out stream, topic, and DAG name. If two streams or topics feed the same DAG with overlapping offset ranges, the second batch computes the same id, Airflow answers 409 (already exists), and the sink reports success even though that batch's data never ran.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                    _ => Some(Retryable::Fatal),
+                }
+            }
+            Err(_) => Some(Retryable::Transient),
+        }
+    }
+}
+
+fn is_permanent_status(status: u16) -> bool {
+    matches!(status, 400 | 401 | 403 | 404 | 405 | 422)
```
slbotbm, line 730: 408, 413, 501, and 505 are treated as fatal (no retry), but the error mapping reports the same codes as transient failures. Fix: use one shared function for both decisions.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                    return Err(e);
+                }
+            };
+            if let Some(&idx) = index_by_dag.get(&dag_id) {
+                groups[idx].1.push(message);
+            } else {
+                index_by_dag.insert(dag_id.clone(), groups.len());
+                groups.push((dag_id, vec![message]));
+            }
+        }
```
slbotbm, line 363: The first message whose DAG id cannot be resolved fails the entire batch, and the same message will fail it again on every poll. Fix: document this behavior so operators know a bad header stalls the stream.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+            self.messages_triggered
+                .fetch_add(message_count as u64, Ordering::Relaxed);
+            return Ok(());
+        }
+
+        let response_body = match response.text().await {
+            Ok(body) => body,
+            Err(e) => format!("<body read error: {e}>"),
+        };
+        let truncated = truncate_response(&response_body, MAX_RESPONSE_LOG_BYTES);
```
slbotbm, line 537: The full response text is buffered before being cut to 500 characters for logging. A large error page wastes memory. Fix: cap the read.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+
+fn parse_duration(input: Option<&str>, default: &str) -> Duration {
+    let raw = input.unwrap_or(default);
+    HumanDuration::from_str(raw)
+        .map(|d| *d)
+        .unwrap_or_else(|e| {
+            warn!("Invalid duration '{raw}': {e}, using default '{default}'");
+            *HumanDuration::from_str(default).expect("default duration must be valid")
+        })
+}
```
slbotbm, line 758: A typo in a timeout value only logs a warning and keeps going. I feel failing fast in `open()` would be preferable.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+    Payload,
+}
+
+/// Plugin config from `[plugin_config]`.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct AirflowSinkConfig {
+    /// Airflow webserver base URL (required), e.g. `http://localhost:8080`.
+    pub base_url: String,
+    /// Default DAG id to trigger (required unless every message sets `dag_id_header`).
+    pub dag_id: String,
```
slbotbm, line 87: This field cannot acutally be ommitted: the field is a required string, so leaving it out fails deserialization before the `open()` check that was meant to allow header-only mode. Today header-only setups must write `dag_id = ""`. Fix: make it `Option<String>` with a default and validate in `open()`.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+            id,
+            base_url,
+            log_url,
+            dag_id,
+            api_prefix,
+            auth,
+            username: config.username,
+            password: config.password,
+            token: config.token,
+            dag_id_header: config.dag_id_header.filter(|h| !h.is_empty()),
```
slbotbm, line 209: A `dag_id_header` of only spaces passes validation but never matches a real header, so the sink silently falls back to the default DAG. Fix: trim the name and reject it if nothing remains.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+    request_headers: Option<HeaderMap>,
+    client: Option<ClientWithMiddleware>,
+    trigger_attempts: AtomicU64,
+    messages_triggered: AtomicU64,
+    errors_count: AtomicU64,
+}
+
+impl AirflowSink {
+    pub fn new(id: u32, config: AirflowSinkConfig) -> Self {
+        let base_url = config.base_url.trim_end_matches('/').to_string();
```
slbotbm, line 159: Leading spaces will break the base url. Fix: Trim first.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                "dag_id is required unless dag_id_header is set for per-message overrides"
+                    .to_string(),
+            ));
+        }
+
+        // Validate auth credentials early.
+        self.request_headers = Some(self.build_auth_headers()?);
+        self.client = Some(self.build_client()?);
+
+        if self.health_check_enabled {
```
slbotbm, line 604: Health check ignores the API prefix -- even with a custom `api_prefix`, the health check still probes the default `/api/v1/version` path. Fix: document the coupling or derive the health default from the prefix.

core/connectors/sinks/airflow_sink/README.md
```diff
@@ -0,0 +1,175 @@
...
+| `auth` | no | `none` | `none`, `basic`, or `bearer` |
+| `username` / `password` | basic | — | Basic auth credentials (`password` is secret) |
+| `token` | bearer | — | Bearer/JWT token (secret) |
+| `dag_id_header` | no | unset | Message header that overrides `dag_id` (groups batch by value) |
+| `conf_mode` | no | `payload` | How each message body is placed under `conf.messages[].payload` |
+| `include_iggy_metadata_in_conf` | no | `false` | Nest batch stream/topic/offset range under `conf.iggy` |
+| `health_check_enabled` | no | `true` | `GET` health path in `open()` |
+| `health_path` | no | `/api/v1/version` | Path relative to `base_url` |
+| `timeout` | no | `30s` | HTTP timeout |
+| `max_retries` | no | `3` | Total attempts including the first |
```
slbotbm, line 79: The retry docs (here and  `config.toml:64`) are wrong: Both say `max_retries` is total attempts including the first, but the code passes it to `build_with_max_retries`, which counts retries after the first (so 3 means 4 attempts).

core/connectors/sinks/airflow_sink/README.md
```diff
@@ -0,0 +1,175 @@
...
+conf_mode = "payload"
+include_iggy_metadata_in_conf = false
+health_check_enabled = true
+health_path = "/api/v1/version"
+timeout = "30s"
+max_retries = 3
+retry_delay = "1s"
+retry_backoff_multiplier = 2
+max_retry_delay = "30s"
+```
```
slbotbm, line 61: This omits `tls_danger_accept_invalid_certs`, `max_connections`, and `verbose_logging`, all present in `config.toml`.

core/connectors/sinks/airflow_sink/README.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Plugin fields
+
+| Field | Required | Default | Description |
+| --- | :---: | --- | --- |
+| `base_url` | yes | — | Airflow webserver base URL |
+| `dag_id` | yes* | — | Default DAG id (*optional if every message sets `dag_id_header`) |
+| `api_prefix` | no | `/api/v1` | REST prefix for version differences |
```
slbotbm, line 69: No rows for `max_connections` or `retry_backoff_multiplier`. Please add them

core/connectors/sinks/airflow_sink/src/lib.rs
slbotbm: Also, the code treats 405 as permanent, but the README status table has no 405 row and does not say what happens to unlisted codes. Fix: add the row, document the fallback.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+    let trimmed = path.trim();
+    if trimmed.is_empty() {
+        return DEFAULT_HEALTH_PATH.to_string();
+    }
+    if trimmed.starts_with('/') {
+        trimmed.to_string()
+    } else {
+        format!("/{trimmed}")
+    }
+}
```
slbotbm, line 782: `normalize_path_prefix` and `normalize_path` differ by one default and one trim. Merge into a single function (saves ~12 lines).

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                }
+                Payload::Proto(proto) => Ok(serde_json::json!({
+                    "data": general_purpose::STANDARD.encode(proto.as_bytes()),
+                    "encoding": "base64",
+                })),
+            },
+        }
+    }
+
+    fn build_batch_conf(
```
slbotbm, line 387: `build_batch_conf` borrows a slice and swaps in a placeholder per message while the caller already owns the `Vec`. Take the `Vec` by value instead.

core/connectors/sinks/airflow_sink/Cargo.toml
```diff
@@ -0,0 +1,57 @@
...
+ignored = ["dashmap"]
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
+bytes = { workspace = true }
+dashmap = { workspace = true }
```
slbotbm, line 42: the `dashmap` dependency is unused and silenced in the lint config. Remove both entries.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+fn encode_path_segment(value: &str) -> String {
+    let mut out = String::with_capacity(value.len());
+    for byte in value.bytes() {
+        match byte {
+            b'A'..=b'Z' | b'a'..=b'z' | b'0'..=b'9' | b'-' | b'_' | b'.' | b'~' => {
+                out.push(byte as char);
+            }
+            _ => {
+                out.push('%');
+                out.push_str(&format!("{byte:02X}"));
```
slbotbm, line 794: percent-encoding allocates a temporary string per byte. Write in place instead (cold path, tiny win).

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+        assert_eq!(conf["iggy"]["partition_id"], 2);
+        assert_eq!(conf["iggy"]["first_offset"], 3);
+        assert_eq!(conf["iggy"]["last_offset"], 4);
+        assert_eq!(conf["iggy"]["message_count"], 2);
+        assert_eq!(conf["messages"].as_array().unwrap().len(), 2);
+    }
+
+    #[test]
+    fn given_dag_id_header_when_group_should_split_batches_by_dag() {
+        use iggy_common::{HeaderKey, HeaderValue};
```
slbotbm, line 1001: Two tests declare their own `use` lines instead of sharing the module header. Move them to the top of the test module. (this and `given_dag_id_header_when_resolve_should_override_config`)

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                Err(e) => {
+                    error!(
+                        "{CONNECTOR_NAME} ID: {} — cannot resolve dag_id at offset {}: {e}",
+                        self.id, message.offset
+                    );
+                    self.errors_count.fetch_add(1, Ordering::Relaxed);
+                    return Err(e);
+                }
+            };
+            if let Some(&idx) = index_by_dag.get(&dag_id) {
```
slbotbm, line 357: The grouping loop starts with an empty `Vec` and `HashMap` while the batch size is already known. Create them with `with_capacity(messages.len())`.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+        &self,
+        messages: &mut [ConsumedMessage],
+        topic_metadata: &TopicMetadata,
+        messages_metadata: &MessagesMetadata,
+    ) -> Result<(serde_json::Value, u64, u64, usize), Error> {
+        if messages.is_empty() {
+            return Err(Error::InvalidRecordValue(
+                "cannot build conf for empty message batch".to_string(),
+            ));
+        }
```
slbotbm, line 397: This guard has no test. Add one that passes an empty batch and expects the error. (The "all messages failed" branch cannot trigger today because every payload converts successfully.)

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+        );
+        Err(Error::HttpRequestFailed(format!(
+            "Airflow batch trigger status {status_code}: {truncated}"
+        )))
+    }
+}
+
+#[async_trait]
+impl Sink for AirflowSink {
+    async fn open(&mut self) -> Result<(), Error> {
```
slbotbm, line 568: Empty base URL, wrong URL scheme, and missing DAG id with no header never get tested. Add one failing-`open()` test per case, with the health check disabled.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                    let override_id = value.to_string_value();
+                    if override_id.is_empty() {
+                        return Err(Error::InvalidRecordValue(format!(
+                            "header '{header_name}' is present but empty"
+                        )));
+                    }
+                    return Ok(override_id);
+                }
+            }
+        }
```
slbotbm, line 327: The empty-header error and the missing-default error have no tests, and neither does the early return in the grouping function. Add two resolve-fails tests plus one grouping-error test.

core/integration/tests/connectors/airflow/wiremock/mappings/accept-dag-runs.json
slbotbm, line 1: The WireMock mapping has no 409 or 404 case, so replay and permanent-failure paths cannot be tested. Add discriminated 409/404 cases (same work as the happy-path test item above; this one surfaced in the sweep and is not yet validated).

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+        let first_offset = messages.first().map(|m| m.offset).unwrap_or(0);
+        let last_offset = messages.last().map(|m| m.offset).unwrap_or(first_offset);
+        let mut entries = Vec::with_capacity(messages.len());
+        let mut skipped = 0u64;
+
+        for message in messages.iter_mut() {
+            let offset = message.offset;
+            let id = message.id;
+            let timestamp = message.timestamp;
+            let payload = std::mem::replace(&mut message.payload, Payload::Raw(vec![]));
```
slbotbm, line 408: Each message payload is swapped out through a throwaway empty value even though the caller owns the whole batch. Either borrow the slice or take the `Vec` by value.

core/connectors/sinks/airflow_sink/README.md
slbotbm, line 1: The docs promise the same batch replays to the same id, but the id covers offsets and count, so changing the batch size produces a new id and a duplicate run. Add one sentence saying so.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                        )));
+                    }
+                    return Ok(override_id);
+                }
+            }
+        }
+        if self.dag_id.is_empty() {
+            return Err(Error::InvalidConfigValue(
+                "dag_id is required when no per-message header override is present".to_string(),
+            ));
```
slbotbm, line 331: A missing header reports a config error while an empty header reports a record error, but both depend on the message content. Use the record error in both cases.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+    fn health_url(&self) -> String {
+        format!("{}{}", self.base_url, self.health_path)
+    }
+
+    fn resolve_dag_id(&self, message: &ConsumedMessage) -> Result<String, Error> {
+        if let Some(header_name) = self.dag_id_header.as_deref()
+            && let Some(headers) = message.headers.as_ref()
+        {
+            for (key, value) in headers {
+                if key.to_string_value().eq_ignore_ascii_case(header_name) {
```
slbotbm, line 317: Each header lookup allocates a fresh string per key. Borrow with `as_str` and compare case-insensitively instead.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+                }
+            }
+        }
+        if self.dag_id.is_empty() {
+            return Err(Error::InvalidConfigValue(
+                "dag_id is required when no per-message header override is present".to_string(),
+            ));
+        }
+        Ok(self.dag_id.clone())
+    }
```
slbotbm, line 334: Every message clones the configured id, then it is cloned again into the grouping map. When no header override is set, skip grouping and use a single batch instead.

core/connectors/sinks/airflow_sink/src/lib.rs
```diff
@@ -0,0 +1,1100 @@
...
+    let trimmed = path.trim().trim_end_matches('/');
+    if trimmed.is_empty() {
+        return DEFAULT_API_PREFIX.to_string();
+    }
+    if trimmed.starts_with('/') {
+        trimmed.to_string()
+    } else {
+        format!("/{trimmed}")
+    }
+}
```
slbotbm, line 770: The branches that handle missing slashes and unparsable durations are never tested directly. Copy the fallback tests from `http_sink`.

slbotbm (changes requested):

PR #3727, opened 2026-07-21
feat(python): expose Permissions and remaining user auth methods

### LINKED ISSUE #3726: [python sdk] Expose Permissions and the remaining user auth methods
### Description

Follow-up to #3682, as scoped there. The Python SDK can now create, update, and delete users, but every user is created without permissions: `create_user` in foreign/python/src/client.rs hardcodes None for the permissions argument, and there is no way to grant or inspect permissions afterwards. update_permissions, change_password, and logout_user from the Rust UserClient trait (`core/common/src/traits/user_client.rs`) are also still unexposed, and UserInfoDetails returned by get_user has no permissions getter.

This means Python still cannot provision non-root service accounts end to end. A deployment script can create a user but cannot give it access to anything.

### Affected area / component

Python SDK

### Proposed solution

Mirror the Rust domain types (core/common/src/types/permissions/permissions_global.rs) as PyO3 classes and complete the UserClient surface:

- Expose Permissions, GlobalPermissions, StreamPermissions, and TopicPermissions as Python classes. All fields are booleans plus two nested dicts (`streams: dict[int, StreamPermissions], topics: dict[int, TopicPermissions]`), so the mapping is mechanical.
- Add the optional permissions parameter to `create_user`, replacing the hardcoded None.
- Add `update_permissions(user_id, permissions)`, `change_password(user_id, current_password, new_password)`, and `logout_user() `on IggyClient.
- Add a permissions getter on UserInfoDetails so granted permissions round-trip through get_user.

### Alternatives considered

_No response_

### Contribution

### Good first issue

ethanlin01x:

Closes #3726 

## Rationale
The Python SDK could create users but not grant them access to anything: `create_user` hardcoded `None` for permissions, and `update_permissions`, `change_password`, and logout_user` were unexposed. 

## What changed?
Before, the Python SDK could create users but not grant them access to anything: `create_user` hardcoded `None` for permissions, and `update_permissions`, `change_password`, and `logout_user` were unexposed. 
Now the Rust `Permissions` hierarchy is mirrored as PyO3 classes, `create_user` takes an optional `permissions` argument, `UserInfoDetails` exposes a `permissions` getter, and the three missing methods complete the `UserClient` surface, with the underlying Rust client handling the wire encoding.

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage
Claude was used to help generate and review this PR and all the changes are checked by the human.


foreign/python/src/permissions.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// Create permissions from global permissions and optional per-stream permissions.
+    ///
+    /// Args:
+    ///     global_: Global permissions as `GlobalPermissions | None`; defaults to all denied.
+    ///     streams: Per-stream permissions keyed by stream ID as
+    ///         `dict[int, StreamPermissions] | None`.
+    #[new]
+    #[pyo3(signature = (global_=None, streams=None))]
+    fn new(
+        #[gen_stub(override_type(type_repr = "GlobalPermissions | None"))] global_: Option<
```
slbotbm: Why `global_`? why not `global`?

foreign/python/tests/test_permissions.py
slbotbm, line 1: Let's remove the following tests:
- `test_all_global_permission_flags_set_together`
- `test_all_stream_permission_flags_set_together`
- `test_all_topic_permission_flags_set_together`

they do not add much to the current test set.

foreign/python/tests/test_permissions.py
```diff
@@ -0,0 +1,575 @@
...
+        assert first != different
+
+
+class TestCreateUserWithPermissions:
+    """Test create_user with the permissions argument."""
+
+    @pytest.mark.asyncio
+    async def test_create_user_without_permissions_has_none(
+        self, iggy_client: IggyClient, unique_name
+    ):
```
slbotbm: `test_create_user_without_permissions_has_none` is valid but need not be a separate integration test. Add `permissions is None` assertions to existing `test_create_and_get_user`.

foreign/python/tests/test_permissions.py
slbotbm: Let us add the following tests:

- User without `manage_users` cannot update another user’s permissions.
- User with `manage_users` can update another user’s permissions.
- Failed update leaves target permissions unchanged.
- calling logout twice fails

slbotbm (changes requested):
first round

foreign/python/src/permissions.rs
ethanlin01x: `global` is a reserved keyword in Python.
The trailing underscore follows the PEP 8 convention for names that collide with keywords (like `class_`).  An alternative would be `global_permissions`, but I kept `global_` to mirror the Rust field name. Let me know if you'd prefer the longer name.

foreign/python/tests/test_permissions.py
ethanlin01x: Removed, 41497daccfcb432c121167f3d28875ce694eda65

foreign/python/tests/test_permissions.py
ethanlin01x: Fixed, da2c17f537e3005aed8dde52bc86a5cd2a8d76e2

foreign/python/tests/test_permissions.py
ethanlin01x: Added tests, 47fcb74f1a3c284e4281c4f18c7bfb48ae783aa2

foreign/python/src/permissions.rs
slbotbm: let's go with global_permissions

foreign/python/tests/test_permissions.py
slbotbm: Let's also add the following tests:
- User without `manage_users` cannot change another user's password.
- User with `manage_users` can change another user's password.
- Failed unauthorized change leaves the old credentials valid.

foreign/python/src/client.rs
```diff
@@ -264,6 +268,97 @@ impl IggyClient {
...
+    ///     user_id: User identifier as `str | int`.
+    ///     permissions: New permissions as `Permissions | None`.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the permissions are updated.
+    ///
+    /// Raises:
+    ///     PyValueError: If a string identifier is invalid.
+    ///     PyRuntimeError: If the request fails.
+    #[pyo3(signature = (user_id, permissions=None))]
```
slbotbm: Let's make the permission explicit with `#[pyo3(signature = (user_id, permissions))]` so that clearing the permissions will have to be explicitly specified with `None`.

foreign/python/src/permissions.rs
slbotbm, line 1: `GlobalPermissions`, `StreamPermissions`, and `TopicPermissions` accept positional booleans. Ten adjacent booleans make this legal and dangerous:

```python
GlobalPermissions(False, False, True, ...)
```

A shifted argument can grant the wrong privilege without any error. These are new security-sensitive APIs. Add `*` to their PyO3 signatures so permission flags must be named. Existing tests already use keyword arguments.

slbotbm: Empty `streams` and `topics` dictionaries are preserved locally but cannot be represented distinctly on the wire. Both become empty vectors and decode as `None`, so `Permissions(streams={})` and `StreamPermissions(topics={})` fail round-trip equality. Please normalize empty dictionaries to `None` in these constructors and cover both cases.

foreign/python/src/permissions.rs
```diff
@@ -0,0 +1,406 @@
...
+    ///     manage_servers: Allow managing servers; includes `read_servers`.
+    ///     read_servers: Allow reading server info (stats, clients).
+    ///     manage_users: Allow managing users; includes `read_users`.
+    ///     read_users: Allow reading user info.
+    ///     manage_streams: Allow managing all streams; includes `manage_topics`.
+    ///     read_streams: Allow reading all streams; includes `read_topics`.
+    ///     manage_topics: Allow managing all topics; includes `read_topics`.
+    ///     read_topics: Allow reading all topics and consumer groups.
+    ///     poll_messages: Allow polling messages from all streams.
+    ///     send_messages: Allow sending messages to all streams.
```
slbotbm, line 140: This documentation does not actually describe permission inheritance:

- `read_topics` also permits polling message contents.
- `manage_topics` permits polling and sending messages.
- `manage_streams` inherits topic management and therefore permits
  polling and sending.
- Stream `read_topics` and topic `read_topic` permit polling.
- Stream and topic management permissions permit sending.

slbotbm (changes requested):

foreign/python/src/permissions.rs
ethanlin01x: Renamed, 44b214ce6d7f7915681f453d8191a9659167523b
One question: should streams / topics keep mirroring the Rust field names, or would you prefer stream_permissions / topic_permissions for symmetry?

foreign/python/tests/test_permissions.py
ethanlin01x: Added, 364162a85522de4db910967021cf6d8356df77cf. The "cannot change" and "old credentials stay valid" cases share one test (test_failed_password_change_leaves_target_credentials_valid), since the former's setup and assertions are a strict subset of the latter's.

foreign/python/src/client.rs
ethanlin01x: Done, 306a0dca2494d78d8706c0c013ee4ca528cdc62e.

foreign/python/src/permissions.rs
ethanlin01x: - Keyword-only flags added in [306a0dc](https://github.com/apache/iggy/pull/3727/commits/306a0dca2494d78d8706c0c013ee4ca528cdc62e)
- Empty dicts normalized to None in [1ff9e5f](https://github.com/apache/iggy/pull/3727/commits/1ff9e5f85d9bd21782d9d51685945718e51f7430)

foreign/python/src/permissions.rs
ethanlin01x: Fixed, [aff6f0f](https://github.com/apache/iggy/pull/3727/commits/aff6f0fbbb66af888588aaeb6f43f354965d2d83). 
Documented the direct inclusion edges per the server permissioner rules with a note that they apply transitively. Also covered a few grants beyond the listed ones: `read_topics` / `read_topic` additionally allow managing (including creating and deleting) consumer groups, `poll_messages` allows managing consumer offsets, and stream-level `read_stream` also permits polling via `read_topics`.

slbotbm (approved):

ethanlin01x:
Hi @spetz 

Would you mind taking a look at this PR when you get a chance?
Since you helped merge my previous Python PR, I’d really appreciate your eyes on this one as well. Please feel free to leave any comments or suggestions.

Thanks!

numinnex (approved):

hubcio (approved):

PR #3729, opened 2026-07-22
fix(connectors): Harden Elasticsearch sink test readiness

### LINKED ISSUE #3728: Elasticsearch sink integration test flakes when connectors health never becomes ready
### Summary

The Elasticsearch sink integration test
(`elasticsearch_sink_stores_json_messages`) can fail with `HealthCheckFailed` for `iggy-connectors` after many retries. Connectors runtime logs stop at `Opening Elasticsearch sink...` and never reach a successful open. Observed locally against a reused Testcontainers Elasticsearch instance; matches Discord failure reports. [ [elastic_connector_files.zip](https://github.com/user-attachments/files/30239708/elastic_connector_files.zip)](url)

Root cause is a combination of:

1. Sink `open()` can hang indefinitely on Elasticsearch HTTP (no client   timeout applied), and the connectors HTTP API only binds **after**    plugin open completes.
2. Fixture reuse (`ReuseDirective::Always`, container name   `iggy-test-elasticsearch`) can attach to a days-old container without   re-validating cluster health the same way as a cold start.
3. On macOS, `localhost` can resolve to IPv6 `::1` first while Docker   publishes IPv4 `0.0.0.0:HOST→9200`, so the elasticsearch-rs client can
   hang while a request health probe still looks fine.
4. Harness readiness polled a generic HTTP path and did not fail fast if  `iggy-connectors` crashed during open.

Stale leftover indices alone do **not** appear to break the happy path on a healthy host, but accumulated leftovers on a reused container are still worth cleaning so the shared node stays healthy across runs.

### Impact

- Flaky CI / local integration runs for Elasticsearch sink (and related  harness readiness).
- Hard to diagnose: failure surfaces as connectors health timeout, not as  an Elasticsearch client timeout error.

### Proposed fix

Branch in progress: `fix/elasticsearch-sink-test-flakiness`

- **Sink:** apply `timeout_seconds` (default 30s) on  `TransportBuilder` so hung ES I/O cannot block FFI `open()` forever.
- **Fixture container:** prefer `http://127.0.0.1:{port}`; always poll  `/_cluster/health` after attach; if the reused container is unusable,  `docker rm -f iggy-test-elasticsearch` once and recreate (named harness container only — does not touch other Elasticsearch  containers).
- **Fixture container:** best-effort sweep of stale test indices  (`iggy_messages_*`, `test_documents_*`) older than 30 minutes so prior   empty/partial indices do not accumulate on the reused node.
- **Sink fixture:** refresh index before each document-count poll; keep  last error in timeout messages.
- **Harness:** poll connectors `/health` and fail fast with  `ProcessCrashed` if the child dies during retries.

### Acceptance criteria

- [ ] `timeout_seconds` is applied to the Elasticsearch transport client.
- [ ] Elasticsearch fixture readiness is re-checked on every setup,      including reuse attach.
- [ ] Wedged `iggy-test-elasticsearch` is removed and recreated once;     other Docker Elasticsearch containers are never targeted.
- [ ] Stale harness test indices older than the age threshold are     cleaned best-effort before the test proceeds.
- [ ] Connectors runtime readiness uses `/health` and detects a dead    process during retries.
- [ ] `elasticsearch_sink_stores_json_messages` passes locally against       reused container.

### References

- Discord report / log set (connectors stuck at "Opening Elasticsearch  sink...", harness `HealthCheckFailed`).
- Local analysis notes (commit where flake was investigated):  `6f1d5485d06c3c43afd2510e432b2f7c1a3ac41a`
- Related discussion context: connectors / integration test stability.

### Out of scope

- Changing connectors runtime to bind HTTP before plugin `open()`  (possible follow-up for clearer "starting" health).
- Broader Test containers reuse policy changes for non-Elasticsearch  fixtures.

[elastic_connector_files.zip](https://github.com/user-attachments/files/30239822/elastic_connector_files.zip)

### Affected area / component

Connectors

### Deployment

Docker (DockerHub image)

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

### Contribution

### Good first issue

ryerraguntla:
## Summary

- Fixes flaky Elasticsearch sink integration test where `iggy-connectors`
  never becomes healthy because sink `open()` can hang on Elasticsearch HTTP
  with no client timeout, while the connectors HTTP API only binds after
  plugin open.
- Hardens the Elasticsearch Testcontainers fixture: use `127.0.0.1`, always
  re-check `/_cluster/health` on reuse attach, self-heal a wedged
  `iggy-test-elasticsearch` container only, and sweep stale harness indices
  older than 30 minutes.
- Improves harness readiness: poll connectors `/health` and fail fast if the
  child process dies during retries; refresh the index before document-count
  polls in the sink fixture.

Closes #3728

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo sort --no-format --workspace`
- [x] `cargo clippy -p iggy_connector_elasticsearch_sink -p integration --all-features --all-targets -- -D warnings`
- [x] `cargo test -p integration -- connectors::elasticsearch::elasticsearch_sink::elasticsearch_sink_stores_json_messages`
- [x] Confirm `docker rm -f` recovery targets only `iggy-test-elasticsearch`
      (no other Elasticsearch containers)
- [x] Optional: leave a reused `iggy-test-elasticsearch` running and re-run
      the sink test to verify reuse + readiness path

## Notes

- Container force-remove uses the fixed harness name
  `iggy-test-elasticsearch` only.
- Stale index sweep is best-effort and age-gated so concurrent harness
  tests sharing the reused container are not disrupted.


core/integration/tests/connectors/fixtures/elasticsearch/container.rs
slbotbm, line 1: start() force-removes the shared container after any try_start() error, not only after confirming it is wedged. Transient health, port-inspection, or Docker errors can therefore remove an Elasticsearch container still used by another process. The fixed name limits which container is removed but does not provide ownership or concurrency safety. Could this recovery path verify the container is unhealthy and coordinate removal with a cross-process lock before running docker rm -f?

slbotbm (changes requested):
`

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
ryerraguntla: Let me add the logic to remove when required.

ryerraguntla: Fixed in https://github.com/apache/iggy/pull/3729/commits/dd2c04ce743c78260a26d8992a02138c00ff8fc3

core/connectors/sinks/elasticsearch_sink/src/lib.rs
```diff
@@ -83,7 +86,15 @@ impl ElasticsearchSink {
...
-        let mut transport_builder = TransportBuilder::new(conn_pool);
+        // elasticsearch-rs defaults to no timeout; without this, a hung ES
+        // connection blocks FFI open() and the connectors HTTP health never binds.
+        let timeout_seconds = self
+            .config
+            .timeout_seconds
+            .unwrap_or(DEFAULT_TIMEOUT_SECONDS)
+            .max(1);
+        let mut transport_builder =
+            TransportBuilder::new(conn_pool).timeout(Duration::from_secs(timeout_seconds));
```
hubcio, line 104: this timeout is client-global in elasticsearch-rs, so it bounds bulk requests in `consume()` too, not just `open()`. that changes the failure mode: before, a bulk on degraded ES stalled until it eventually succeeded; now anything over `timeout_seconds` returns `Err`, which the runtime discards (#2927) with the offset already committed at poll (#2928) - the batch is silently dropped. mechanism is pre-existing and tracked there, but this PR is what arms it for slow-ES bulk. worth an upgrade note in the PR description: raise `timeout_seconds` for slow bulk workloads, and note the tradeoff until #2927/#2928 land.

separately, the comment slightly oversells the timeout as the flake fix - with the 30s default the harness readiness budget (~20s) expires first, so what fixes #3728 is the fixture readiness gate; this is the infinite-hang backstop. maybe say that instead.

core/connectors/sinks/elasticsearch_sink/src/lib.rs
```diff
@@ -83,7 +86,15 @@ impl ElasticsearchSink {
...
 
         let conn_pool = elasticsearch::http::transport::SingleNodeConnectionPool::new(url);
-        let mut transport_builder = TransportBuilder::new(conn_pool);
+        // elasticsearch-rs defaults to no timeout; without this, a hung ES
+        // connection blocks FFI open() and the connectors HTTP health never binds.
+        let timeout_seconds = self
+            .config
+            .timeout_seconds
+            .unwrap_or(DEFAULT_TIMEOUT_SECONDS)
+            .max(1);
```
hubcio, line 102: `.max(1)` silently turns `timeout_seconds: 0` into 1s. defensible - `Duration::ZERO` would instant-fail every request - but undocumented, and clickhouse_sink passes 0 through unclamped, so the two siblings now disagree on the same config field. document the clamp here; clickhouse should probably get the same guard.

core/integration/src/harness/handle/connectors_runtime.rs
```diff
@@ -243,19 +243,33 @@ impl IggyServerDependent for ConnectorsRuntimeHandle {
     }
 
     async fn wait_ready(&mut self) -> Result<(), TestBinaryError> {
-        let http_address = self.http_url();
+        // Prefer /health over `/` so readiness means the connectors API is up,
+        // not some other listener that happened to bind the reserved port.
+        let health_url = format!("{}/health", self.http_url());
         let client = reqwest::Client::new();
```
hubcio: `reqwest::Client::new()` has no request timeout, so a black-holed GET (connection accepted, no response) hangs `send()` forever - the loop never gets back to the pid check above and the retry budget never fires. low risk on localhost (dead process means ECONNREFUSED, which fast-fails), but a 1-2s timeout like the fixture probe client makes both guards effective.

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
```diff
@@ -137,14 +195,308 @@ impl ElasticsearchContainer {
...
+            "{}{ELASTICSEARCH_HEALTH_ENDPOINT}?timeout=1s",
+            self.base_url
+        );
+        let mut last_error = String::from("no attempts made");
+
+        for attempt in 1..=CLUSTER_READY_ATTEMPTS {
+            match client.get(&health_url).send().await {
+                Ok(response) if response.status().is_success() => {
+                    let body = response.text().await.unwrap_or_default();
+                    if body.contains("\"timed_out\":true") {
```
hubcio: `?timeout=1s` without a `wait_for_*` param makes `/_cluster/health` return immediately with `timed_out: false`, so this check never fires - and any 200 counts as ready even with cluster status red. `?wait_for_status=yellow&timeout=1s` makes both the timeout param and this check do real work (and then parsing the bool beats string-matching the body).

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
```diff
@@ -137,14 +195,308 @@ impl ElasticsearchContainer {
...
+}
+
+/// Cross-process advisory lock for inspect+rm recovery of the shared reuse
+/// container. Dropping the file releases the lock (including on process crash).
+struct RecoveryLock {
+    _file: File,
+}
+
+fn acquire_recovery_lock() -> Result<RecoveryLock, String> {
+    let path = std::env::temp_dir().join(RECOVERY_LOCK_FILE_NAME);
```
hubcio, line 439: cross-process serialization only holds if every nextest process resolves the same `temp_dir()` - with a per-process `TMPDIR` (some sandbox/CI setups) each process locks its own file and recovery races silently. probably fine to accept, but worth a comment stating the assumption.

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
```diff
@@ -137,14 +195,308 @@ impl ElasticsearchContainer {
...
+    let file = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        .truncate(false)
+        .open(&path)
+        .map_err(|error| format!("open {}: {error}", path.display()))?;
+    file.lock()
```
hubcio: `File::lock()` blocks with no acquire timeout - worst case a peer holds this through its whole recovery (two 120s container starts plus readiness waits), and since the test runtime is current-thread, the block parks the entire runtime for that long. bounded and failure-path-only, so nice-to-have: `try_lock()` with a deadline.

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
```diff
@@ -137,14 +195,308 @@ impl ElasticsearchContainer {
...
+    let status = parts.next().unwrap_or("").trim();
+    let health = parts.next().unwrap_or("").trim();
+
+    matches!(
+        status,
+        "exited" | "dead" | "created" | "paused" | "restarting"
+    ) || health == "unhealthy"
+}
+
+fn docker_command_stdout(args: &[&str], timeout_secs: u64) -> Option<String> {
```
hubcio: `docker_command_stdout` and `force_remove_container` duplicate the spawn / `try_wait` / deadline / kill loop, and both block the async caller (`thread::sleep` on a current-thread runtime). `tokio::process::Command` + `tokio::time::timeout(dur, cmd.output())` + `kill_on_drop(true)` collapses both to a few lines with the same 5s/15s kill semantics - `delta/fixture.rs:254` is the in-repo precedent and tokio `full` is already enabled here. it also removes a latent footgun: stdout is only drained after exit, so docker output over the pipe buffer (~64KB) would deadlock until the timeout kill. fine today with one-line `inspect -f` output, but the helper reads as generic.

core/integration/tests/connectors/fixtures/elasticsearch/sink.rs
```diff
@@ -61,22 +61,38 @@ impl ElasticsearchSinkFixture {
         &self,
         expected_count: usize,
     ) -> Result<usize, TestBinaryError> {
+        let mut last_error: Option<TestBinaryError> = None;
+
         for _ in 0..POLL_ATTEMPTS {
+            // Refresh so near-real-time search/count sees recently indexed docs.
+            if let Err(error) = self.refresh_index().await {
```
hubcio: `refresh_index()` goes through `create_http_client()` (30s timeout + 3 retries) inside a 50ms poll loop - on the failure path that is up to 100 refreshes, each with retry backoff. happy path exits in a few iterations so this only inflates time-to-failure; the short-timeout probe-client pattern from container.rs would fit here too.

hubcio (changes requested):
not in this diff, same class of problem:

- elasticsearch source `create_client()` (`core/connectors/sources/elasticsearch_source/src/lib.rs:304`) still builds its transport with no timeout - same `open()`-hang failure mode this PR fixes for the sink, so source tests keep the #3728 flake class. worth a follow-up.
- `mcp.rs` `wait_ready()` still uses the pattern this PR fixed for connectors: GET `/`, any `Ok(_)` counts as ready, no process-crash fast-fail. consistency follow-up.

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
```diff
@@ -137,14 +198,320 @@ impl ElasticsearchContainer {
...
+    };
+
+    let mut parts = stdout.trim().split('|');
+    let status = parts.next().unwrap_or("").trim();
+    let health = parts.next().unwrap_or("").trim();
+
+    matches!(
+        status,
+        "exited" | "dead" | "created" | "paused" | "restarting"
+    ) || health == "unhealthy"
```
numinnex: This predicate cannot fire for the wedge state the recovery path was written for.

`docker.io/library/elasticsearch:9.3.0` ships no HEALTHCHECK: `docker image inspect docker.io/library/elasticsearch:9.3.0 --format '{{json .Config.Healthcheck}}'` returns `null`, and `GenericImage` does not add one. So `.State.Health` is always nil, the `{{if .State.Health}}` branch in `inspect_format` yields the empty string, and `health == "unhealthy"` is unreachable. The predicate reduces to the status list alone.

That list covers container-level death (an OOM-killed JVM exits, so `exited` catches it), but not the case `wait_until_ready` was added to detect: ES process up, cluster red or HTTP wedged from a corrupted data dir, which is exactly what the doc comment above names as motivation. Docker reports `running`, this returns false, and every Elasticsearch test then fails at fixture setup until someone runs `docker rm -f iggy-test-elasticsearch` by hand. That manual step is what the recovery path was meant to remove.

Two ways out:

1. Treat `running` as removable when readiness already exhausted all `CLUSTER_READY_ATTEMPTS`. Pass that signal in so it stays distinct from a transient start or port flake, and keep the removal under `RecoveryLock`.
2. If leaving a running container alone is the deliberate call, drop the unreachable `health` leg and put the remediation command in the `warn!` at the caller. It currently says "leaving it in place" with no next step, so whoever hits it has to work out `docker rm -f iggy-test-elasticsearch` themselves.

core/integration/tests/connectors/fixtures/elasticsearch/sink.rs
slbotbm, line 1: At sink.rs:91, the refresh was moved to a short-timeout client, but `count_documents()` still uses create_http_client() with a 30-second timeout and three retries. The unconditional final count at sink.rs:91 does too. A degraded _count endpoint can therefore keep this failure path blocked for minutes. The short-timeout probe client should cover both refresh and count.

slbotbm (changes requested):

core/integration/tests/connectors/fixtures/elasticsearch/sink.rs
ryerraguntla: Fixed in [58858e8](https://github.com/apache/iggy/pull/3729/commits/58858e820343ac4745449cf291f9354a6269c49b)

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
ryerraguntla: Fixed in [58858e8](https://github.com/apache/iggy/pull/3729/commits/58858e820343ac4745449cf291f9354a6269c49b)

slbotbm (dismissed):

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
```diff
@@ -37,6 +40,25 @@ const ELASTICSEARCH_HEALTH_ENDPOINT: &str = "/_cluster/health";
 // name. Per-test isolation comes from a unique index per fixture, not a fresh
 // container.
 const ELASTICSEARCH_CONTAINER_NAME: &str = "iggy-test-elasticsearch";
+// Short probe timeouts: create_http_client() uses 30s + retries and must not
+// be used for readiness. One hung attempt there looks like a 60s+ test hang.
+const CLUSTER_READY_ATTEMPTS: usize = 40;
```
hubcio: this attempt budget is only ~10s while probes fail fast (connection refused during boot), and it's the sole readiness gate when attaching to a container another process just created or restarted - `HttpWaitStrategy` runs only for the creator. a cold boot here took ~12s to go green, so an attaching test ran out of attempts just short of ready. a time deadline (e.g. 60s) fits better than an attempt count: once es answers, each attempt costs up to ~1s server-side, so raising the count alone would stretch the failure path instead.

core/integration/tests/connectors/fixtures/elasticsearch/sink.rs
```diff
@@ -88,6 +106,74 @@ impl ElasticsearchSinkFixture {
...
+        let count_response = response.json::<CountResponse>().await.map_err(|error| {
+            TestBinaryError::InvalidState {
+                message: format!("Failed to parse count response: {error}"),
+            }
+        })?;
+        Ok(count_response.count)
+    }
+}
+
+fn probe_client() -> Result<reqwest::Client, TestBinaryError> {
```
hubcio, line 175: this builds a fresh client on every call - twice per poll iteration - and each `Client::build()` re-parses the system ca store via rustls-platform-verifier even for plain http, plus loses keep-alive between probes. build one client before the loop in `wait_for_documents` and pass it to the two probe helpers.

hubcio (changes requested):
two notes on code this diff doesn't touch, both fine as follow-ups:

- `ElasticsearchOps::create_index` sends `PUT /{index}` through the retry middleware, but index creation is not idempotent. i hit this live: the first put timed out client-side after es had already created the index, the retry got 400 `resource_already_exists_exception`, and the fixture failed. indexes are uuid-unique per fixture, so already-exists can only come from a retried put - treating it as success makes the path retry-safe.
- the fixture could set `cluster.routing.allocation.disk.threshold_enabled=false` on the container. a host disk above es's 90% high watermark makes it refuse all shard allocation: index creates stall on `wait_for_active_shards` for ~30s, colliding with the fixture's 30s client timeout (and feeding the retry-then-400 above), while the readiness gate still reports green on an empty cluster with zero shards. that's the wedged-but-running state the recovery path deliberately won't remove. caveat: the env var only lands on a freshly created container - an existing reused `iggy-test-elasticsearch` keeps its old settings until removed.

core/integration/tests/connectors/fixtures/elasticsearch/container.rs
ryerraguntla: Fixed

core/integration/tests/connectors/fixtures/elasticsearch/sink.rs
ryerraguntla: fixed

core/integration/src/harness/handle/connectors_runtime.rs
slbotbm, line 1: `DEFAULT_HEALTH_CHECK_RETRIES` is 1,000 and the interval is 20 ms. On the black-holed request case these serial 2-second client timeouts make `wait_ready()` take roughly 34 minutes, so the harness can still hang for a long time even after this change. Bound the whole readiness loop with a deadline, or reduce the retry budget to account for the per-request timeout, rather than only bounding each probe.

slbotbm (changes requested):

core/integration/src/harness/handle/connectors_runtime.rs
ryerraguntla: fixed

slbotbm (approved):

hubcio (approved):

PR #3733, opened 2026-07-22
feat(cpp): add functions related to user management, and initial high-level client

slbotbm:

Closes #
Relates to #2100 

## Rationale

Works towards completion of cpp client.

## What changed?

Adds the following:
- create_user
- get_user
- get_users
- update_user
- delete_user
and their tests.

Adds high level client that will function as primary API with documentation. The diff looks big since a lot of the public api was changed and aligned to PascalCase. I have not rewritten the tests using the high level client since I would like to do that at once instead of a gradual approach.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex
3. How did you verify the generated code works correctly? read through it


foreign/cpp/src/client.rs
```diff
@@ -1085,6 +1135,90 @@ impl Client {
...
+        RUNTIME.block_on(async {
+            self.inner
+                .delete_user(&rust_user_id)
+                .await
+                .map_err(|error| format!("Could not delete user '{rust_user_id}': {error}"))?;
+            Ok(())
+        })
+    }
+
+    pub fn update_user(
```
ethanlin01x, line 1255: Rust SDK allows updating username and status independently (Option params), but this forces both. Consider `has_username` / `has_status` flags like the existing `has_permissions` pattern

foreign/cpp/include/iggy.hpp
```diff
@@ -252,4 +333,425 @@ class PollingStrategy final {
...
+     * leaves the transport connected and marks the session authenticated. For
+     * HTTP, the returned access token is stored by the client and used for
+     * subsequent authenticated requests.
+     *
+     * @param username Iggy user name.
+     * @param password Iggy user password.
+     * @return Information about the authenticated session.
+     * @throws IggyException if authentication fails.
+     */
+    LoginInfo Login(std::string username, std::string password) const;
```
ethanlin01x: `Login()` returning `ffi::LoginInfo` leaks the FFI's `has_access_token` bool-flag workaround into the high-level API. Since this is meant to be the primary API, a dedicated type with `std::optional` for the token would be cleaner, and avoids a breaking change later.

foreign/cpp/include/iggy.hpp
```diff
@@ -252,4 +333,425 @@ class PollingStrategy final {
...
+     * Establishes the configured transport connection and starts heartbeat
+     * processing. If automatic login was configured, authentication is also
+     * performed. Calling this on an already connected client has no effect.
+     *
+     * @note HTTP is stateless; connecting initializes heartbeat processing but
+     *       does not open a persistent transport connection.
+     * @throws IggyException if the connection or automatic authentication
+     *         fails.
+     */
+    void Connect() const;
```
ethanlin01x: Nit: These methods are `const` but clearly mutate connection state (it only compiles because the state lives behind `client_`). Marking them non-const would better communicate the semantics, since `const` usually implies no observable state change and often thread-safety.

Applies to:
- `Connect()`
- `Disconnect()`
- `Shutdown()`
- `Login()`
- `Logout()`

ethanlin01x (commented):
A few suggestions

foreign/cpp/src/client.rs
slbotbm: Done

foreign/cpp/include/iggy.hpp
slbotbm: Done

foreign/cpp/include/iggy.hpp
slbotbm: tbh this is currently a workaround for cxx's limitations. I eventually plan to convert all structs defined in rust to cpp structs/enums and call them from rust. That'll give the users a better experience as compared to now. I'll remember this, and implement this when I make that change.

foreign/cpp/src/client.rs
krishvishal, line 71: This and `iggy.hpp:389-395` says destroying the client releases the connection. `IggyClient::connect` spawns a heartbeat holding its own `Arc` to the transport, exiting only on `ClientShutdown`, so the socket, the server session and the ping loop all survive. Same reason `:488` ("no effect when already connected") is false and the `:501-505` reuse cycle leaks a task per cycle. Worth filing upstream: store the `JoinHandle` discarded at `clients/client.rs:250` and abort it in a `Drop for IggyClient`.

foreign/cpp/include/iggy.hpp
```diff
@@ -252,4 +333,425 @@ class PollingStrategy final {
...
+     * @brief Connects to the configured Iggy server.
+     *
+     * Establishes the configured transport connection and starts heartbeat
+     * processing. If automatic login was configured, authentication is also
+     * performed. Calling this on an already connected client has no effect.
+     *
+     * @note HTTP is stateless; connecting initializes heartbeat processing but
+     *       does not open a persistent transport connection.
+     * @throws IggyException if the connection or automatic authentication
+     *         fails.
```
krishvishal: This promises `Connect()` throws on failure. With the documented default (unlimited retries) `tcp_client.rs:424-442` retries forever and `block_on` parks the caller indefinitely. It can only throw if `WithReconnectionMaxRetries(n)` was called.

foreign/cpp/src/client.rs
```diff
@@ -73,37 +75,85 @@ pub struct Client {
...
+                builder = builder.with_tls_domain(config.tls_domain);
+            }
+            if !config.tls_ca_file.is_empty() {
+                builder = builder.with_tls_ca_file(config.tls_ca_file);
+            }
+            if config.has_tls_validate_certificate {
+                builder = builder.with_tls_validate_certificate(config.tls_validate_certificate);
+            }
+        }
+    }
```
krishvishal: `has_tls_enabled`:  `Builder().WithTlsDomain(d).WithTlsCaFile(p).Build()` without `WithTlsEnabled(true)` silently drops all three TLS settings and connects in plaintext. The gate carries no information either, since `TcpClientConfig::default()` is already `tls_enabled: false`. Suggest deleting the field, calling `with_tls_enabled` unconditionally, and erroring when TLS material is set with TLS off. - `set_reconnection_max_retries` (`src/lib.rs:348`): dead. "Never set" and `WithoutReconnectionLimit()` are indistinguishable, and `{set_=false, has_=true, value=n}` is representable and silently ignored. `has_reconnection_max_retries` alone expresses `Option<u32>`.

foreign/cpp/include/iggy.hpp

foreign/cpp/src/client.rs
slbotbm: Done

ethanlin01x (approved):

foreign/cpp/src/client.rs
slbotbm: https://github.com/apache/iggy/pull/3852

slbotbm:
The one test failing is because of migrating to the new server. The test works like this: the root user creates a new user without the `read_user` permission. The code then logs in as the new user and tries to get the user's own details using `get_user` and asserts that this is possible since the user was requesting their own details, not someone else's. 

In the legacy server this was possible, but in the new server, this is not possible since `read_user` permission is checked before anything happens. 

Is this a regression or a new capability?

foreign/cpp/include/iggy.hpp
```diff
@@ -370,4 +449,441 @@ class TopicOption final {
...
+     * Any Rust client handle currently owned by this object is released first.
+     * Call Shutdown() before replacing a connected client. The moved-from
+     * client must not be used for client operations.
+     */
+    IggyBlockingClient &operator=(IggyBlockingClient &&other) noexcept;
+
+    /**
+     * @brief Releases the handle to the underlying Rust client.
+     *
+     * Destruction does not stop the heartbeat task started by Connect(). For
```
hubcio: also at lines 466, 614-615 and 632-634.

these lifecycle notes don't match the sdk: dropping the client aborts the heartbeat task (`impl Drop for IggyClient` in core/sdk/src/clients/client.rs), and `connect()` early-returns when a live handle exists, so repeated `Connect()` doesn't stack tasks and disconnect/reconnect cycles are fine - the suite's own `DisconnectThenReconnectWithoutRelogin` exercises exactly that cycle. the only true part is that `Disconnect()` keeps the existing task pinging, which with auto-login configured can silently reconnect - that's the note worth keeping.

foreign/cpp/include/iggy.hpp
```diff
@@ -370,4 +449,441 @@ class TopicOption final {
...
+ * @code{.cpp}
+ * auto client = iggy::IggyBlockingClient::Builder()
+ *                   .WithServerAddress("127.0.0.1:8090")
+ *                   .Build();
+ * client.Connect();
+ * client.Login("iggy", "iggy");
+ * client.Shutdown();
+ * @endcode
+ */
+class IggyBlockingClient final {
```
hubcio, line 507: as it stands the class can't do any actual work - the only operations are connect/login/logout, `client_` is private with no accessor, so no stream, topic, message or user call is reachable. it's also never instantiated by the test suite: all three builder tests throw inside `Build()` before construction, so the move ctor, move assign, destructor and all five methods have zero executed coverage. either add a handle accessor plus lifetime coverage, or keep the type out of the public header until it wraps real operations.

foreign/cpp/tests/e2e/client.cpp
```diff
@@ -146,6 +184,1387 @@ TEST_F(LowLevelE2E_Client, LogoutErrorsWhenCalledMoreThanOnce) {
...
+}
+
+TEST_F(LowLevelE2E_Client, UpdateUserAcceptsUsernameLengthBounds) {
+    RecordProperty("description", "Accepts exact three-byte and fifty-byte username boundaries.");
+    iggy::ffi::Client *client           = GetLoggedInClient();
+    const std::string first_username    = GetRandomName(50);
+    const std::string second_username   = GetRandomName(50);
+    const std::string first_replacement = GetRandomName(3);
+    std::string second_replacement      = GetRandomName(50);
+    second_replacement.resize(50);
```
hubcio: `resize(50)` pads with `\0`, so this tests a random 8..50 char name plus nul bytes instead of the 50-byte boundary the test name claims - only 1 in 43 draws is a clean 50. `resize(50, 'a')` plus `ASSERT_EQ(second_replacement.size(), 50u)`, like line 266 does. the `ASSERT_NE` on the next line compares a 3-byte vs 50-byte string, always true, can be dropped.

foreign/cpp/src/client.cpp
```diff
@@ -0,0 +1,216 @@
...
+    try {
+        return IggyBlockingClient(ffi::from_connection_string(connection_string));
+    } catch (const std::exception &error) {
+        throw IggyException(error.what());
+    }
+}
+
+void IggyBlockingClient::Connect() {
+    try {
+        client_->connect();
```
hubcio: all five methods deref `client_` unguarded; after a move it's nullptr and cxx passes it as a rust `&Client`, which is UB rather than a throw - and callers have no way to check (no `operator bool`, no accessor). a private `Handle()` that throws when null and routes all five turns use-after-move into a clean `IggyException`.

foreign/cpp/src/lib.rs
```diff
@@ -372,14 +372,58 @@ mod ffi {
         streams: Vec<StreamPermissionEntry>,
     }
 
+    struct UserInfo {
+        id: u32,
+        created_at: u64,
+        status: u8,
```
hubcio: also on `UserInfoDetails` and the `create_user`/`update_user` params.

`status: u8` means magic `1`/`2` at every call site (the e2e file has over a hundred of them), and C++ is the only sdk without a named type - java, C#, go, python and node all expose a `UserStatus` enum. a shared `#[repr(u8)] enum UserStatus { Active = 1, Inactive = 2 }` in the bridge gives C++ named constants exactly like the existing `HeaderKind` pattern: the field stays `u8`, no rust match change, no wire change.

foreign/cpp/src/lib.rs
```diff
@@ -372,14 +372,58 @@ mod ffi {
...
+    struct LoginInfo {
+        user_id: u32,
+        has_access_token: bool,
+        access_token: String,
+        access_token_expiry: u64,
+    }
+
+    struct IggyClientConfig {
+        server_address: String,
+        auto_login_kind: String,
```
hubcio: stringly-typed discriminant with two spellings for one state (`""` from aggregate init, `"disabled"` from the builder). a shared `#[repr(u8)] enum AutoLoginKind { Disabled = 0, UsernamePassword, PersonalAccessToken }` collapses that, drops three heap strings from the config, and gives C++ named constants; `Disabled = 0` keeps zero-init working.

foreign/cpp/src/client.rs
```diff
@@ -75,37 +77,86 @@ pub struct Client {
...
-        "" => RustIggyClientBuilder::new()
-            .with_tcp()
-            .build()
-            .map_err(|error| format!("Could not build default connection: {error}"))?,
-        s if s.starts_with("iggy://") || s.starts_with("iggy+") => {
-            RustIggyClient::from_connection_string(s)
-                .map_err(|error| format!("Could not parse connection string '{s}': {error}"))?
+pub fn new_connection(config: ffi::IggyClientConfig) -> Result<*mut Client, String> {
+    let mut builder = RustIggyClientBuilder::new().with_tcp();
+    if !config.server_address.is_empty() {
```
hubcio, line 82: gating on `is_empty()` makes empty strings silently mean unset, which contradicts the documented contract: `WithServerAddress("").Build()` quietly falls back to 127.0.0.1:8090 even though the doxygen says build throws on an invalid address (an unset env var ends up talking to localhost instead of failing loudly), `WithTlsDomain("")` / `WithTlsCaFile("")` slip past the tls-requires-tls check below, and `WithPersonalAccessToken("")` arms auto-login with credentials that can only fail later at `Connect()`. rejecting empty explicitly (or has_* flags like the other optionals) closes all three.

foreign/cpp/include/iggy.hpp
```diff
@@ -370,4 +449,441 @@ class TopicOption final {
     TopicOption() = delete;
```
hubcio, line 476: the PascalCase rename skipped this class - `segment_size`, `enforce_fsync`, `messages_required_to_save`, `size_of_messages_required_to_save` and `preallocate_segments` are still snake_case while every other option type in the header moved, so the public header ships two naming conventions for the same kind of type.

foreign/cpp/src/client.cpp
```diff
@@ -0,0 +1,216 @@
...
+        config.has_reestablish_after        = reestablish_after_micros_.has_value();
+        config.reestablish_after_micros     = reestablish_after_micros_.value_or(0);
+        config.tls_enabled                  = tls_enabled_;
+        config.tls_domain                   = tls_domain_;
+        config.tls_ca_file                  = tls_ca_file_;
+        config.has_tls_validate_certificate = tls_validate_certificate_.has_value();
+        config.tls_validate_certificate     = tls_validate_certificate_.value_or(false);
+        config.no_delay                     = no_delay_;
+        client                              = ffi::new_connection(std::move(config));
+        if (client == nullptr) {
```
hubcio: `new_connection` either throws or returns a non-null pointer (`Box::into_raw`), the move ctor is noexcept, and the private ctor already throws this exact message - so the null check and the catch-block rollback are unreachable. `return IggyBlockingClient(ffi::new_connection(std::move(config)));` inside the try, mirroring `FromConnectionString`, saves ~10 lines and the duplicated literal.

foreign/cpp/tests/e2e/client.cpp
```diff
@@ -146,6 +184,1387 @@ TEST_F(LowLevelE2E_Client, LogoutErrorsWhenCalledMoreThanOnce) {
...
+TEST_F(LowLevelE2E_Client, UpdateUserByUsernameChangesUsernameAndStatus) {
+    RecordProperty("description", "Updates a user by username and changes both username and status.");
+    iggy::ffi::Client *client     = GetLoggedInClient();
+    const std::string username    = GetRandomName(50);
+    const std::string replacement = GetRandomName(50);
+    iggy::ffi::UserInfoDetails created{};
+    ASSERT_NO_THROW({ created = CreateUser(client, username, "secret123", 1); });
+    ASSERT_NO_THROW(client->update_user(make_string_identifier(username), true, replacement, true, 2));
+    const auto tracked_user = std::find(tracked_user_names_.begin(), tracked_user_names_.end(), username);
+    ASSERT_NE(tracked_user, tracked_user_names_.end());
```
hubcio: this find/assert/if/assign block repeats 9 times (also at 716, 749, 775, 781, 830, 911, 951, 1019), and the trailing `if` is dead in each - `ASSERT_NE` already returns on failure. a `RenameTrackedUser(from, to)` helper next to `ForgetUser` saves ~40 lines and lets the fixture members stay private (the other four members widened to protected aren't touched by any test).

foreign/cpp/src/client.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+void IggyBlockingClient::Reset() noexcept {
+    if (client_ == nullptr) {
+        return;
+    }
+
+    ffi::Client *client = std::exchange(client_, nullptr);
+    try {
```
hubcio: `delete_connection` can only return `Ok(())`, and a rust panic aborts instead of throwing (cxx wraps every extern "Rust" shim in `prevent_unwind`), so this catch and the five like it in the tests are dead. changing the bridge fn to return unit instead of `Result<()>` removes them all and is source-compatible for callers.

foreign/cpp/tests/e2e/test_helpers.hpp
```diff
@@ -222,13 +245,61 @@ class E2ETestFixture : public ::testing::Test {
         }
     }
 
+    void CleanupUsers() {
```
hubcio: these two are the 5th and 6th copy of the same connect-as-root-and-loop body (streams and consumer groups each have a normal + best-effort pair too). a `RunAsRoot(fn)` plus a noexcept best-effort variant collapses all six to one-liners, ~70 lines less, and one root login per teardown instead of up to three - each login is an argon2 hash server-side, so it adds up across the suite. one constraint: keep the consumer-groups -> streams -> users order.

foreign/cpp/src/client.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+IggyBlockingClient::~IggyBlockingClient() {
+    Reset();
+}
+
+IggyBlockingClient IggyBlockingClient::FromConnectionString(std::string connection_string) {
+    try {
+        return IggyBlockingClient(ffi::from_connection_string(connection_string));
+    } catch (const std::exception &error) {
```
hubcio: same five-line catch/rethrow block appears 7 times in this file. a small rethrow helper would collapse them, though seven explicit blocks arguably read fine - take it or leave it. worth doing only if combined with the null-guarding `Handle()` so one wrapper does both.

hubcio (changes requested):

foreign/cpp/include/iggy.hpp
slbotbm: This was left over from the previous change. Fixed.

foreign/cpp/include/iggy.hpp
slbotbm: ~Moved into a `detail` namespace inside `iggy.hpp`.~

slbotbm: I'd like to leave this as it is. The next PR will expose all of the remaining functions + structs that are in rust right now. The PR after that will convert all tests to utilize this public API, and also add tests for cpp-only code.

foreign/cpp/tests/e2e/client.cpp
slbotbm: Done

foreign/cpp/src/client.cpp
slbotbm: done. added a unit test as well.

foreign/cpp/src/lib.rs
slbotbm: Done

foreign/cpp/src/lib.rs
slbotbm: Done

foreign/cpp/src/client.rs
slbotbm: Empty strings are rejected on cpp side now.

foreign/cpp/include/iggy.hpp
slbotbm: fixed

foreign/cpp/src/client.cpp
slbotbm: Done

foreign/cpp/tests/e2e/client.cpp
slbotbm: done

foreign/cpp/src/client.cpp
slbotbm: done

slbotbm:
I also consolidated the two test / e2e CI jobs for cpp into one since it was inefficient to run two separate jobs that were compiling the same code, one just without the server.

foreign/cpp/src/client.cpp
slbotbm: DOne

foreign/cpp/tests/e2e/test_helpers.hpp
slbotbm: Done

hubcio (approved):

krishvishal (approved):

PR #3745, opened 2026-07-24
feat(java): add cluster metadata and leader redirection to TCP client

mmodzelewski:
Java clients connected to a VSR follower had no way to discover
the cluster leader, so operations against a multi-node cluster
failed until the user manually pointed the client at the right
node.

Add the GET_CLUSTER_METADATA command with cluster roster models
and a binary deserializer. After login the async TCP client
fetches the roster and, when a healthy leader with an enabled
TCP endpoint lives elsewhere, reconnects to it and replays the
login. Redirections are capped at 3 consecutive hops so two
nodes both claiming leadership cannot ping-pong a client, and
every failure on the redirect path is non-fatal: the client
stays on the current node. Mirrors the Rust and Go SDKs.

Covered by a leader_redirection BDD scenario wired into the
cluster docker-compose setup.


numinnex (approved):

slbotbm (commented):
The changes look alright to me. One comment though:

Let us consider the following setup: users A and B, follower F, and leader L. users A and B login through follower using the following code:
```java
AsyncIggyTcpClient client = ...;
client.users().login("userA", "passwordA");
client.users().login("userB", "passwordB");
```
If the two users utilize the same client as above, the following happens:
1. A calls login(). Its initial authentication runs on F.
2. B calls login() concurrently. Its initial authentication also runs on F.
3. A enters the serialized redirection step, discovers L, switches the shared client connection to L, and replays A’s login there.
4. B’s redirection step runs next. It checks the current connection, which now points to L.
5. Because L is already the leader, B’s login is not replayed.
6. B’s future returns B’s identity from the earlier login on F, but the active connection to L is authenticated as A.

This happens because  the initial login happens before the serialization:
```java
loginWithoutRedirect(username, password)
      .thenCompose(identity -> redirectionHook.afterLogin(...));
```
If this is to be fixed, you should serialize the complete login-and-redirection operation. 

If this is not a concern, please feel free to merge the PR.

mmodzelewski:
@slbotbm Thanks for the review - the race you describe is real. I would merge it as is, as the scenario itself falls outside the client's contract: a TCP client wraps a single connection, and the server keeps one session per connection, so the client is inherently single-user. Even with the complete login-and-redirection operation serialized, the second login would still replace the connection's identity, and user A's caller would silently end up on a connection authenticated as user B. Multiple users need separate client instances.

numinnex (approved):

spetz (approved):

PR #3753, opened 2026-07-27
feat(consensus): rework client table for session resume across restarts

numinnex:


core/metadata/src/impls/metadata.rs
```diff
@@ -995,9 +1011,12 @@ where
...
             .expect("submit_register_in_process: consensus only exists on shard 0");
 
-        // Idempotent fast path: existing session skips pipeline + wire-reply.
-        if let Some(session) = self.client_table.borrow().get_session(client_id) {
-            return Ok(session);
+        // Idempotent fast path: existing entry skips pipeline + wire-reply.
+        // Returns the current epoch without bumping it; rebind-bumps-epoch
+        // (zombie fencing per reconnect) arrives with the stable-session-id
+        // auth flow, which commits a Register per bind.
+        if let Some(epoch) = self.client_table.borrow().get_epoch(client_id) {
```
hubcio: the fast path returns the existing entry's epoch without checking that the authenticated user matches `entry.user_id`, and it never proposes a Register, so `commit_register` - the only writer that refreshes `user_id` - never runs. the vsr key is caller-supplied on the wire (`let vsr_client_id = request.header().client;` in the login handler at dispatch.rs:2687), so a client authenticating with its own valid low-privilege credentials but presenting someone else's client id gets bound to that entry, and every later replicated op resolves its acting user from it. the frame is well formed (session 0, request 0) so header validation passes it cleanly.

scope is replicated metadata ops - partition ops resolve the user from `SessionManager` instead, so those stay on the attacker's own identity. it needs the victim's client id, which is a uuidv4 for sdk clients but is logged and sent cleartext on every frame, and is sequential for http.

the entry doc already claims "re-register refreshes it: the rebind re-authenticated", so the check looks intended and simply absent. requiring `entry.user_id == authenticated user_id` here closes this and the restart case below in one edit.

core/metadata/src/impls/recovery.rs
```diff
@@ -200,6 +211,26 @@ where
             continue;
         }
 
+        // Register/Logout mutate the client table and skip the state
+        // machine, mirroring the commit paths (`on_ack` / `commit_journal`).
+        // Replaying them from a fresh table in apply order re-mints the same
+        // epochs every replica derived live.
+        if header.operation == Operation::Register {
+            let reply = build_reply_message(header, &bytes::Bytes::new());
+            client_table.commit_register(header.client, header.user_id, reply, |_| false);
```
hubcio: this is the first change that makes the table survive a reboot: replay rebuilds entries keyed by the previous boot's client ids. but neither id minter is seeded from recovered state - `client_seq` in the shard coordinator is `Cell::new(1)` per process, and http uses that minted value directly as the vsr table key. both sequences restart at 1, so the first post-restart login can land on a live recovered entry owned by a different user, and the fast path above then hands it that entry without refreshing `user_id`.

two effects. replicated ops execute under the prior boot's authority, and that stamp goes into the replicated prepare header, so into the durable log and the next snapshot. and the new session numbers its requests from 1 against an inherited watermark, so its early writes read as duplicates and it gets the previous user's cached reply bytes decoded as an answer to a command it never sent, or `AlreadyApplied` and the writes silently vanish.

no attacker needed - this fires on the scenario the pr is built for. it needs the prior Register still inside the replayed window and a different login order across restarts. worth noting the exposure is inversely correlated with traffic: checkpointing is what drains the stale Register, so a quiet deployment stays exposed indefinitely.

core/server-ng/src/dispatch.rs
```diff
@@ -578,11 +578,142 @@ where
...
+            presented = header.session,
+            current = epoch,
+            "refusing session resume with mismatched epoch"
+        );
+        return None;
+    }
+
+    {
+        let mut sessions = sessions.borrow_mut();
+        if let Err(error) = sessions.login(transport_client_id, user_id) {
```
hubcio: resume binds an unbound transport to an existing table session and logs it in as the entry's cached `user_id` while taking no credential at all. no user-exists check, no `status == Active` (the normal login path does check it in auth.rs), no pat expiry. authority then resolves from the table, so the connection runs as the original registrant, and `bind_session` also evicts the legitimate holder.

the doc above treats `(client_id, epoch)` as a bearer token on the premise that the client id is a client-generated random u128. that holds for the rust sdk but not for http/quic/tcp-tls, which draw from the sequential per-process counter, and the epoch half contributes no entropy because nothing on the live path ever bumps it past 1. so for an http-originated entry the credential is `client=N, session=1`. the table also carries no plane or transport tag, so a raw tcp peer can bind to an http session.

separately exploitable with no guessing at all by a deactivated or deleted user: `remove_client` only runs on Logout apply and replay, so neither DeleteUser nor a status flip drops the session, and the authz layer never re-checks status. user slab ids are reused, so a stale entry later resolves to whoever occupies that index.

fix is to re-authenticate on resume and bind only when `entry.user_id` matches the authenticated user.

core/consensus/src/client_table.rs
```diff
@@ -260,55 +322,44 @@ impl ClientTable {
...
         } else {
             if self.index.len() >= self.slots.len() {
                 self.evict_oldest(&in_flight);
             }
             let slot_idx = self.first_free_slot().expect("eviction must free a slot");
+            let mut ring = VecDeque::with_capacity(REPLY_RING_CAPACITY);
+            ring.push_back(cached);
             self.slots[slot_idx] = Some(ClientEntry {
-                session,
+                epoch: 1,
```
hubcio: a fresh entry hardcodes `epoch: 1, watermark: REGISTER_REQUEST_ID`. so after any entry drop, a client re-registering under the same id gets a clean slate, and every request id it already committed is above 0, which means all of them read `New`, dispatch, and re-apply.

both pre-existing containments were removed together here: `session` used to be `prepare_header.op`, log-derived and non-regressing per key, and contiguity (`request == committed + 1`) capped how far a reset could reach. eviction and journal-window loss are local and unlogged, so nothing in the log tells a re-registering or recovering path what the entry used to hold.

the fence half is separately inert: no live proposing path bumps an epoch (the fast path returns early, and the preflight either replays or drops), so every entry a primary holds sits at epoch 1 and `Fenced`/`EpochAhead` never fire. zombie fencing, the stated design goal, does not happen in practice.

making the fence the register's op number restores a log-derived identity and fixes the replica-divergence problem below at the same time.

core/consensus/src/client_table.rs
```diff
@@ -345,33 +401,47 @@ impl ClientTable {
...
         assert!(
-            new_commit >= slot_commit,
-            "commit_reply: commit regression for client {client_id}: {slot_commit} -> {new_commit}",
+            new_commit >= latest_commit,
+            "commit_reply: commit regression for client {client_id}: {latest_commit} -> {new_commit}",
         );
         assert!(
-            new_request >= slot_request,
-            "commit_reply: request regression for client {client_id}: {slot_request} -> {new_request}",
+            new_request >= entry.watermark,
```
hubcio: the rebind branch deliberately preserves the watermark, and a live capacity eviction leaves no log record (`evict_oldest` only mutates `slots`/`index`). so for a wal shaped `Register(X)@a, app(X,req=5)@b, [evict], Register(X)@c, app(X,req=R<5)@d` the live run is fine, because the eviction reset the watermark, but replay starts from an empty table, never evicts, takes the rebind branch, keeps watermark 5, and this assert fires inside `recover()`. deterministic on the same wal, so the node cannot boot without wal surgery, while every other replay failure surfaces as a typed `RecoveryError`.

there is a variant that needs no reboot: a backup that did not evict X hits the same sequence in `commit_journal` and panics its shard pump. the `evict_oldest` divergence noted below is what produces that primary/backup split. `R < 5` is reachable in-tree since http restarts numbering at 1.

apply and replay paths should log-and-skip or return a typed error here, not assert.

core/consensus/src/metadata_helpers.rs
```diff
@@ -107,7 +113,7 @@ where
 
     let status = client_table
         .borrow()
-        .check_request(client_id, session, request);
+        .check_request(client_id, session, request, request_checksum);
     match status {
         // Frozen-backed cache -> refcount handoff to the home shard, no copy.
         RequestStatus::Duplicate(cached_reply) => {
```
hubcio, line 141: `CreatePersonalAccessToken` mints a fresh raw token and hash on every attempt, before any preflight (`rewrite_pat_request_for_user` in server-ng's pat.rs). when that request comes back a duplicate, this arm hands the cached committed reply to the caller and it flows straight into `build_raw_pat_reply`. both of that function's guards pass - the reply is a `Command2::Reply`, and a cached successful create carries result code 0 - so the newly minted raw token gets spliced into a reply whose committed hash belongs to the original token. the caller receives a credential that can never authenticate, and the real raw secret only ever existed at the original mint, so it is gone.

this was already reachable at `request == watermark` before, but the reply ring widens the replayable set from one id to every id the ring retains. fix is to carry a "this was a replay" flag out of the preflight and pass the cached reply through untouched - the same reasoning `build_raw_pat_reply` already applies to the eviction case.

core/server-ng/src/dispatch.rs
```diff
@@ -578,11 +578,142 @@ where
...
+) -> Option<(u64, u32)>
+where
+    B: ShellBus,
+    MJ: JournalHandle + 'static,
+    MJ::Target: Journal<MJ::Storage, Entry = Message<PrepareHeader>, Header = PrepareHeader>,
+    S: 'static,
+{
+    let metadata = shard.plane.metadata();
+    let consensus = metadata.consensus.as_ref()?;
+    if !consensus::is_caught_up_primary(consensus) {
```
hubcio: this collapses the transient "not a caught-up primary" refusal into the same `None` as the terminal "no table entry", so the caller answers with the empty Reply that the sdk is documented (a few hundred lines down) to surface as `InvalidCommand` fail-fast rather than `TransientNotCommitted`.

on a multi-replica node the window is guaranteed, not incidental: `is_caught_up_primary` requires `commit_max >= recovery_barrier`, and bootstrap arms the barrier whenever `commit_watermark < restored_op`, which is always true post-restart in the non-solo branch because `commit_watermark` is folded from prepare `commit` stamps and so can never reach the journal head. so the client that can resume is told terminally that it cannot.

make it tri-state (found / no entry / refused) and emit `TransientNotCommitted` for refused.

core/consensus/src/client_table.rs
```diff
@@ -412,9 +482,9 @@ impl ClientTable {
     /// state -> identical choice. `commit_journal` catch-up has empty pipeline,
```
hubcio: this determinism claim does not hold. the `in_flight` input reads the local pipeline, which only a primary populates - `push_prepare_entry` asserts `is_primary()` and `has_message_from_client` scans only local queues. same committed log, different victim on primary vs backup, so table membership and epoch counters diverge. the next sentence concedes the input is pipeline state and then draws the opposite conclusion from it: the prepare queue holds uncommitted primary-local prepares, so it is not agreed-log-derived.

more precisely it is false on any replica whose pipeline is empty, which is wider than just backups - a demoted primary keeps its queue until view change completes.

this split is what makes the replay panic above reachable on a backup. the change also drops the `warn!` that used to make table mismatches visible.

core/server-ng/src/dispatch.rs
```diff
@@ -2397,6 +2542,21 @@ fn submit_disconnect_logout<B, MJ, S>(
...
     const DISCONNECT_LOGOUT_REQUEST_ID: u64 = u64::MAX;
+
+    let is_group_member = !shard
+        .plane
+        .metadata()
+        .mux_stm
+        .streams()
+        .consumer_group_memberships(vsr_client_id)
+        .is_empty();
+    if !is_group_member {
```
hubcio: a non-group-member disconnect returns without submitting Logout, the heartbeat verifier only reaps group members, and there is no ttl (`ClientEntry` has no timestamp field; the table defers it). meanwhile the reference sdk abandons the key on every re-login - `begin_register` does `*self = Self::new()` and the sdk module doc spells out the consequence, that the old entry stays in the server `ClientTable` until evicted. before this change the disconnect Logout reclaimed the entry before the sdk abandoned the id; now the two halves compose and the slot leaks.

then `evict_oldest` ranks by last-write recency, which is anti-correlated with liveness: a live idle admin connection gets evicted in preference to a recently-leaked dead entry.

on its own this is bounded by the table size and an explicit Logout still frees a bound slot, so it is not an authz break by itself. what makes it matter is that it moves the eviction-per-register point from ~4096 simultaneous connections to ~4096 cumulative connects, which any cli-driven or health-probed cluster reaches on a timer - and eviction is exactly the unlogged drop that the findings above turn into re-execution and a replay panic.

core/metadata/src/impls/recovery.rs
```diff
@@ -188,6 +198,7 @@ where
             .fold(snapshot_floor, u64::max)
     };
 
+    let mut client_table = ClientTable::new(CLIENTS_TABLE_MAX);
```
hubcio, line 201: replay starts above the snapshot floor, `checkpoint` drains `0..=last_op`, the journal index is 1024 slots and its own doc warns that entries past a wrap become unreachable for recovery, and the table is not in the snapshot yet (serialization todo). so the advertised resume does not cover a session whose Register fell below the replay window.

the consequence here is a loud refusal rather than silent re-execution: a missing entry gives `NoSession`, which maps to `Evict(EvictionReason::NoSession)`. the silent part comes one step later, after the client re-registers and gets a fresh entry with watermark 0.

core/consensus/src/metadata_helpers.rs
```diff
@@ -116,29 +122,45 @@ where
...
+            tracing::error!(
+                client_id,
+                request,
+                "request_preflight: request id reused for a different operation (client bug)"
+            );
+            PreflightOutcome::Drop
+        }
+        // Applied once, original reply aged out of the ring: refuse
+        // re-execution, nothing to replay. Silent drop.
+        RequestStatus::AlreadyApplied { .. } | RequestStatus::AlreadyRegistered { .. } => {
```
hubcio: `AlreadyApplied` maps to a silent drop and its `{ request, watermark }` payload is thrown away by `{ .. }`, while all three sibling arms log. no reply frame goes out on either plane.

the convention this file and the dispatch path already settled on is terminal -> reply, transient -> silence; the comment on the unbound-transport case says outright that the sdk blocked until socket timeout, so emit an empty Reply and let it fail fast. this is terminal. server-ng does log something, but a `warn!` that labels it transient and carries no request or watermark, so the one log that exists cannot diagnose it.

the silence predates this change; what is new is that a correct client retrying an in-doubt id can land here.

core/server-ng/src/dispatch.rs
```diff
@@ -578,11 +578,142 @@ where
...
+) -> Option<(u128, u64)>
+where
+    B: ShellBus,
+    MJ: JournalHandle + 'static,
+    MJ::Target: Journal<MJ::Storage, Entry = Message<PrepareHeader>, Header = PrepareHeader>,
+    S: 'static,
+{
+    // A resumable frame carries the full old identity; anything less is a
+    // plain unbound request (e.g. a fresh SDK that has not registered yet).
+    if header.client == 0 || header.session == 0 || header.request == 0 {
```
hubcio: this guard filters nothing. ingress validation already enforces `client != 0` for all ops and `session > 0 && request > 0` for every non-Register replicated op, so the only frame it can still catch is `Operation::Reserved` - and the comment's "a fresh sdk that has not registered yet" describes a class that cannot reach here at all, since such a client's `session == 0` is rejected at ingress.

so every real unbound replicated frame reaches the lookup, which on shards other than 0 is a cross-shard message plus an await on the metadata owner shard, before any credential check, with attacker-chosen ids and no memoization. the bind also happens before the `is_partition()` branch, so data-plane frames pay it too. it is bounded and fails closed (`try_send` drops), so this is cost and surprise rather than a break.

either delete the guard or reduce it to a `debug_assert`, and fix the comment.

core/server-ng/src/dispatch.rs
```diff
@@ -578,11 +578,142 @@ where
...
+/// is a client-generated random u128, so presenting it proves the caller is
+/// (or eavesdropped on) the original registrant. `bind_session` evicts any
+/// previous connection bound to the same client id, which is the conflict
+/// rule for one session arriving on two connections.
+///
+/// Returns the `(client_id, session)` binding on success, `None` when the
+/// identity does not check out (caller falls back to the unbound-transport
+/// reply and the client re-registers).
+#[allow(clippy::future_not_send)]
+async fn try_resume_session<B, MJ, S>(
```
hubcio: resume skips the login preamble. `is_protocol_compatible` has exactly one call site, inside the login handler, and `record_sdk_info` two in auth.rs - resume calls neither. so after a server upgrade narrows the compatible window, a previously-registered client rebinds without re-presenting `ClientVersionInfo` and nothing else re-checks it. client listings also report sdk name and version as none for resumed sessions.

core/consensus/src/client_table.rs
```diff
@@ -424,8 +494,9 @@ impl ClientTable {
 
         for (idx, slot) in self.slots.iter().enumerate() {
             let Some(entry) = slot else { continue };
-            let commit = entry.reply.header().commit;
-            let client_id = entry.reply.header().client;
+            let latest = entry.latest().header();
```
hubcio: this walks every slot inside shard 0's no-await commit region, and per occupied slot it does two `bytemuck::checked::try_from_bytes` header casts plus `in_flight`, which itself runs two linear scans over the prepare and request queues. then `first_free_slot` walks the array again. with an 8192-slot table that is a lot of synchronous work on the commit loop, and it compounds with the slot leak: once the table is full this runs on every register.

if you cache `commit`/`client_id` on the entry to avoid the casts, ship it together with the determinism fix above - caching alone would freeze the replica-dependent victim choice in place while making this function look reviewed.

core/consensus/src/client_table.rs
```diff
@@ -156,52 +207,70 @@ impl ClientTable {
...
+        if request == entry.watermark
+            && entry.watermark_checksum != 0
+            && request_checksum != 0
+            && entry.watermark_checksum != request_checksum
+        {
+            return RequestStatus::ChecksumMismatch { request };
         }
 
-        RequestStatus::New
+        entry.find_cached(request).map_or(
```
hubcio: the checksum comparison above only runs for `request == entry.watermark`, so any below-watermark hit that finds a cached reply here returns it without ever comparing the stamp. a client that reused an old request id for a different operation gets the wrong operation's reply.

latent today rather than active: nothing writes a non-zero `request_checksum` - the sdk leaves it 0, `RequestHeader::default` zeroes it, and server-ng's wire layer says outright that the server does not validate it - so both `!= 0` conditions fail on every real request and the guard cannot fire at all. it becomes a real gap the moment integrity stamping lands, so worth closing the coverage now (compare at every `find_cached` hit) rather than after. keeping the field is the right call even though it is derivable from the cached reply - the ring is documented as lossy by design, and a correctness gate should not depend on a best-effort cache.

core/consensus/src/client_table.rs
```diff
@@ -319,18 +370,23 @@ impl ClientTable {
     /// ships; cache skipped; client gets `NoSession` next request.
     ///
     /// # Panics
-    /// On session mismatch or commit/request regression. Missing client
+    /// On epoch mismatch or commit/watermark regression. Missing client
     /// does NOT panic.
-    pub fn commit_reply(&mut self, client_id: u128, session: u64, reply: Message<ReplyHeader>) {
+    pub fn commit_reply(&mut self, client_id: u128, epoch: u64, reply: Message<ReplyHeader>) {
```
hubcio: the `epoch` param and its equality assert are unreachable by construction: all three production callers read the epoch from this same table one statement earlier with no await in between, on a single-threaded shard, and the boot installer cannot interleave. looking the epoch up inside `commit_reply` drops the param, the duplicated caller-side traces, one hash lookup and one RefCell borrow pair per committed metadata op, and one panic path off a `pub` api.

if a partition-plane caller later wants the tripwire back, it should return as an explicit skip rather than an assert.

core/metadata/src/impls/metadata.rs
```diff
@@ -1049,14 +1068,20 @@ where
                 return Err(MetadataSubmitError::PipelineFull);
             }
             return match receiver.await {
-                Ok(reply) => Ok(reply.header().commit),
+                // The commit's `commit_register` minted the epoch; read it
+                // from the table (the reply header does not carry it).
+                Ok(_reply) => self
```
hubcio: after the rework both arms here are byte-identical, and so are both arms of the match below. `let _ = <fut>.await;` followed by a single `get_epoch` read says the same thing. the two rationale comments are load-bearing and should stay.

core/consensus/src/client_table.rs
```diff
@@ -71,19 +71,48 @@ impl CachedReply {
...
+/// refcount bumps and are never persisted or transferred.
+const REPLY_RING_CAPACITY: usize = 5;
+
+/// Per-session entry: fence epoch + committed-request watermark + replies.
+///
+/// The key (`client_id` today, the stable `session_id` once SDK identity
+/// stability lands) is client-supplied; `epoch` is the server-minted fence
+/// that orders rebinds of that key.
 #[derive(Debug)]
 pub struct ClientEntry {
```
hubcio: `ClientEntry` is `pub` with all six fields private and no reference outside this file, and `lib.rs` re-exports only `CachedReply` and `ClientTable`. it is still reachable as `consensus::client_table::ClientEntry` through the `pub mod`, so making it private is a real surface reduction for one keyword.

core/metadata/src/impls/metadata.rs
```diff
@@ -546,6 +547,14 @@ impl<C, J, S, M> IggyMetadata<C, J, S, M> {
         *self.commit_notifier.borrow_mut() = notifier;
     }
 
+    /// Install the client table rebuilt by WAL-replay recovery
+    /// ([`crate::impls::recovery::recover`]). Boot-time only, on the owning
+    /// shard, before it serves traffic - replacing a live table would drop
+    /// committed session state.
+    pub fn install_client_table(&self, client_table: ClientTable) {
```
hubcio: worth guarding this against a non-empty table (`count()` exists on `ClientTable`) so a stray second install cannot silently drop committed session state - the doc says boot-time only but nothing enforces it. prefer a `debug_assert` or a logged refusal over `assert!`, since a boot-path panic is a worse failure than a refusal. cheaper than threading the table through `IggyMetadata::new`, which has ten call sites.

core/consensus/src/client_table.rs
```diff
@@ -453,21 +524,31 @@ impl ClientTable {
...
         let &slot_idx = self.index.get(&client_id)?;
-        self.slots[slot_idx].as_ref().map(|entry| entry.session)
+        self.slots[slot_idx].as_ref().map(|entry| entry.epoch)
+    }
+
+    /// Committed-request watermark for a registered client. A (re)bind reply
+    /// surfaces this so a restarted client resumes numbering at
+    /// `watermark + 1` instead of silently colliding below it.
+    #[must_use]
+    pub fn get_watermark(&self, client_id: u128) -> Option<u64> {
```
hubcio, line 604: `get_watermark` has no non-test callers, and the doc's promise that "a (re)bind reply surfaces this" is not implemented: `LoginRegisterResponse` is exactly `{user_id, session, server_protocol_version, server_version}` and `ReplyHeader.context` is hardcoded 0, so there is no side channel either.

that is the dx gap sitting under the whole feature - without the watermark in the bind reply, no sdk can implement the resume numbering this pr exists to enable. either wire it into the reply or drop the accessor and the doc claim.

hubcio (changes requested):
review of the client table rework (head 9244af7a).

the table rework itself is the right design and worth keeping: watermark over contiguity, deterministic reply bytes, log-derived eviction ordering, wal-replay recovery. what blocks merge is the identity model layered on top - a client-supplied, sometimes-sequential key treated as a bearer credential, plus a fence and a watermark derived from local history that unlogged entry drops silently reset.

minimum to unblock, roughly in the order i would do them:

1. the register fast path must require `entry.user_id == authenticated user_id`. closes the login collision and the restart re-mint in one edit.
2. gate resume on re-authentication and re-verify the user is active / the pat still live.
3. make the fence log-derived (the register's op number, the way `session` used to be). closes the fresh-entry reset and the epoch divergence together.
4. replace the apply-path asserts in `commit_reply` with log-and-skip or a typed error.
5. pass a replay flag out of the preflight so the pat reply builder cannot splice a freshly minted secret into a cached reply.
6. reclaim slots on disconnect (ttl or grace window) so capacity eviction stops being the practical reclaim path.
7. give `AlreadyApplied` a reply frame, and surface the watermark in the bind reply - without it no sdk can implement the resume this is built for.
8. stop seeding `client_seq` at 1.
9. distinguish the transient resume refusal from the terminal one, and advance the http request id unconditionally.

four more findings live in files this diff does not touch, so they cannot be line comments:

- `session_manager.rs:267` - "evict the older" is not terminal, and the cycle is specific to the resume path. `bind_session` demotes the loser to `ConnectionState::Connected`, which is the one state `login` accepts, and `try_resume_session` performs that transition itself with no credential, so the loser's next replicated frame re-resumes and steals back, unbounded, with no eviction frame in either direction. while demoted, its non-replicated reads get `Eviction(NoSession)`. a register-based second bind escapes this because it bumps the epoch.
- `http/submit.rs:236` - the http gate advances the request id only on `Command2::Reply`, leaving the id free after an unanswered or `Canceled` submit that `http/state.rs` itself documents as "may still commit cluster-wide". under contiguity that reuse was mandatory; under a watermark it is a wrong-answer hazard - if the earlier op did commit, the reused id is at or below the watermark, so the caller's next and different operation gets answered from the cache with the previous operation's reply and a 200/204. jumps above the watermark are free now, so the id can simply be burned. the doc block justifying the current policy cites `RequestGap`, which no longer exists.
- `http/session.rs:38-46` - the `MAX_HTTP_SESSIONS = CLIENTS_TABLE_MAX / 2` rationale is invalidated. it reasons that the half cap keeps combined steady state under the shared bound, which assumed non-http occupancy tracks live connections. post-change it does not, so the comment's own "residual eviction race, degrades gracefully" becomes the steady state - and eviction erases the watermark, which turns a documented rare case into a routine at-least-once window.
- `header.rs:155,161` - `session` and `request_checksum` still carry no doc while the adjacent `user_id` has a six-line contract. this change redefines `session` from register-commit-op to fence epoch and makes it a resume credential, and makes `request_checksum` load-bearing for dedup. six sdks read this struct as the wire contract.

stale docs the diff's own sweep missed. three of these are the stated justification for a design decision the new model no longer supports, so they are not cosmetic:

- `metadata.rs:989` and `:1001` - `submit_register_in_process`'s public doc still promises "session number (= commit op)" and claims a mid-window dispatch "produces two register entries and panics on replay". it now returns the table-minted epoch read back via `get_epoch`, and a double register bumps the epoch instead of panicking.
- surviving contiguity comments asserting the deleted `request == committed + 1` rule or naming deleted variants: `metadata.rs:2566`, `stm/stream.rs:552`, `dispatch.rs:893` (names `SessionMismatch`, which no longer exists), `dispatch.rs:2439`, `http/submit.rs:131`.

on tests: keep both specs red until the fixes land - a two-op wal on a single node is the one shape that dodges every finding above, and the 3-node arm added in 9244af7a is what actually exercises the recovery barrier. test 1's proof that a retry was replayed rather than re-applied is indirect: it rests on `CreateStream` rejecting a duplicate name, and `resume_request` never asserts the returned bytes are the cached ones. also worth pinning ingress validation (`RequestHeader::validate` reached through `Message::try_into_typed`) with a test - several conclusions in this review turned on whether it runs.

two follow-ups, not blocking: replayed `Logout` skips consumer-group member removal (byte-identical to master and documented by its own todo, but the snapshot bakes it in), and `Operation::Reserved` is exempt from both validate branches while its `is_client_allowed` rejection happens after `request_preflight`, so a bound client sending `Reserved, request = 0` can get its own register reply replayed. own bytes, no cross-user effect.

core/consensus/src/client_table.rs
numinnex: Remedy landed: neither apply nor replay asserts anymore. `commit_reply` returns a typed `CommitReply { Cached, NoEntry, SkippedRegression { stored, received } }` and the callers log it (`warn!` for the regression, trace for the missing entry). The wire reply ships either way, so a skip degrades one entry's dedup instead of taking down a shard pump or blocking boot.

The reachability chain does not hold though, which matters for whether this is a regression or a latent hazard.

The shape needs two `Register(X)` for the same X inside one replay window, and a second Register is only ever proposed when the table has no entry for X. That is the same fast path you point out on `submit_register_in_process`: entry present means the epoch is returned and nothing is proposed.

- Within one process, absence means `evict_oldest` dropped it, so X has to come back. The Rust SDK abandons the key on re-login (`*self = Self::new()`, as you note on dispatch.rs:2553) and the HTTP counter is monotonic per process, so no id repeats.
- Across a restart, absence means the entry was not recovered, which means `Register(X)@a` fell below the replay floor. Replay then never sees `@a`, so there is no rebind branch to take and the watermark starts at 0.

`R < 5` via HTTP renumbering needs the cross-restart case, which is the one that cannot also supply `@a`.

It is also not introduced here. On master the same WAL panics through a different door: `Register(X)@c` carries a different `commit`, so `commit_register` takes its session-mismatch warn-and-skip arm and leaves the old entry in place with `reply.request == 5`, and the next `commit_reply` for `R < 5` fails `assert!(new_request >= slot_request)`. Same input, same panic, before this PR.

Fixed regardless: asserting on state that diverges replica-locally was the wrong failure mode whether or not the current tree can produce the input, and the epoch param went with it (your client_table.rs:375 comment).

core/consensus/src/metadata_helpers.rs
numinnex: Payload is logged now: `warn!` with `client_id`, `request` and `watermark`, since a run of these means the reply ring is undersized for the client's retry latency.

The last line does not hold as scoped. A correct client has at most one request in flight per session, which is the invariant the ring capacity doc rests on, so the in-doubt id is by definition its own watermark. The watermark's reply is structurally the newest ring entry: `push_latest` pops the front, and only a strictly newer request triggers a pop. So the watermark's reply cannot age out while it is still the watermark, and retrying an in-doubt id resolves as `Duplicate`, never here. Reaching `AlreadyApplied` needs a request strictly below the watermark whose reply has aged out, which means either more than one op in flight or a return to an older id.

The path where a well-behaved client does land below an inherited watermark is the post-restart renumbering on recovery.rs:220. The cross-user half of that is closed now, `submit_register_in_process` requires the authenticated user to own the entry before binding. The same-user half remains: a fresh session that collides with its own prior entry inherits a watermark it has no way to read, which is the DX gap on client_table.rs:577.

Kept as a drop rather than an empty Reply. The requests that reach it did commit, so `InvalidCommand` would be a wrong answer rather than a fast one, and the clients that can reach it are misnumbering. Happy to emit the reply instead if you would rather have fail-fast there.

core/consensus/src/client_table.rs
numinnex: The determinism claim was wrong exactly as you describe, and the `warn!` point is fair too. Fixed by deleting `in_flight` from `evict_oldest` and from `commit_register`'s signature rather than by caching the header fields: the victim is now oldest-commit with a lowest-slot tiebreak and nothing local feeds the choice. The doc now states the constraint positively, that pipeline state must not be consulted because only a primary populates it. A `warn!` also comes back on the other side of the same divergence, `commit_reply` now returns `SkippedRegression { stored, received }` and the caller logs it.

Two provenance corrections though.

`in_flight`, its two queue scans and the "pipeline state derives from the agreed log" sentence are all present on master, and `evict_oldest` diffs semantically clean against it. So the divergence is not introduced by this PR. Worth fixing here since the PR is what makes the table durable, but it is not a regression.

The consequence you draw from it is also better under this PR than before it, not worse. Take the same primary/backup split and follow it to what the client sees:

- master: the backup takes `commit_register`'s session-mismatch arm, skips the update and keeps `session = A`, while the client holds the newer `M` from the primary. On failover `check_request` sees `expected(A) < received(M)`, which is the "newer than issued, client bug" arm: error log, no eviction, no reply. The client waits out its read timeout before it learns anything.
- here: the backup bumps to epoch 2 while the client holds 1. On failover that is `Fenced`, which maps to a terminal `Evict(SessionTooLow)`. Eviction tells the client to re-register, so it recovers on the next round trip.

Same divergence, and the fence turns a silent wedge into a frame that names the problem. So the spurious-fence symptom is new, but it is an improvement over the silent drop it replaced.

core/consensus/src/client_table.rs
numinnex: Diagnosis accepted, remedy declined.

The two observations hold. A fresh entry does reset epoch and watermark, and unlogged entry drops leave nothing in the log to reconstruct what the entry held. The fence is inert on the live path for the reason you give: the register fast path returns before proposing, so nothing bumps an epoch past 1 and `Fenced`/`EpochAhead` are unreachable there.

Making the fence the register's op number is the part I do not want. The table is deliberately free of any value derived from a commit op number, because one logical table spans both the metadata plane and the partition plane and those have independent op ranges. An op-derived fence is not comparable across them, so it buys log-derived identity for the metadata plane at the cost of the table no longer being one mechanism. That constraint is the reason `session` stopped being `prepare_header.op` in the first place, so restoring it is a design reversal rather than a fix.

The alternative that keeps the model is to make rebind actually propose a Register, i.e. delete the fast-path short-circuit, so the bump is real, logged and replica-identical without anything in the entry deriving from an op number. That is the change I would rather make, and it is not in this PR.

The "fixes the replica-divergence problem below at the same time" half is moot now regardless: the divergence is fixed at its source by removing pipeline state from the victim choice (see the reply on client_table.rs:482), so an op-derived fence is no longer load-bearing for it.

core/metadata/src/impls/metadata.rs
numinnex: Fixed: the fast path now requires the authenticated user to own the entry, and a mismatch is a terminal `MetadataSubmitError::ClientIdOwnedByAnotherUser` so the SDK fails hard instead of spinning on it. The resume path that shared the hole is deleted outright rather than gated.

Two corrections to the framing.

The mechanism is not introduced here. Master's fast path is the same early return with no owner check (`get_session` where this branch reads `get_epoch`), and `resolve_acting_user_id` is byte-identical on master. What this PR changes is reachability, not the missing check: recovery now rebuilds the table across a restart, which is what makes the collision you describe on recovery.rs:220 a routine event rather than a nonexistent one. Both still needed fixing here, since this is the PR that makes them reachable.

The scope paragraph understates it in your own favour. "partition ops resolve the user from `SessionManager` instead, so those stay on the attacker's own identity" is true for the fast path, but false for resume: `try_resume_session` called `SessionManager::login` with the `user_id` it read out of the table, so `dispatch_partition_request` resolved the victim's identity too. The data plane was in scope for that path, not just replicated metadata ops.

core/server-ng/src/dispatch.rs
numinnex: The first two paragraphs are right and are fixed: resume is deleted outright rather than gated, so there is no credential-free bind and no `(client_id, epoch)` bearer path. Both halves of the third paragraph need correcting though, because the deleted case does not hold.

The trace stops at identity resolution and does not follow the stamped `user_id` into the gate that consumes it. `DeleteUser`'s apply purges the permissioner:

```rust
// stm/user.rs
state.permissioner.delete_permissions_for_user(user_id as UserId);
```

and that purge is total: `permissioner/mod.rs` clears the global map, both all-streams sets and all three specific-streams indices. Every rule then fails closed on a missing entry, e.g. `permissioner_rules/streams.rs` is `if let Some(..) = users_permissions.get(&user_id) && .. { Ok(()) } else { Err(Unauthorized) }`, and `messages.rs::poll_messages` reaches the same `Err(Unauthorized)` after exhausting every path.

Both planes go through that same permissioner, metadata via `authz::authorize` -> `check`, partition via `dispatch/authz.rs::append_messages` and `dispatch.rs`'s `poll_messages`. So a deleted user's surviving entry holds a slot and a watermark and no authority: every gated op on either plane commits as an `Unauthorized` no-op. `remove_client` not firing on `DeleteUser` is not load-bearing, because deletion is already enforced per-op by the layer that actually decides.

The slab-reuse half narrows to almost nothing for the same reason. The freed id carries no inherited grants (the purge removed them), and a *reconnect* cannot exploit the reuse: the register fast path now requires the authenticated user to own the entry, and after reuse that id legitimately belongs to the new user, so whoever binds acts as themselves. What is left needs the deleted user's socket to stay open across the delete AND an admin creating a new user that lands on exactly the freed slab key. Root is structurally excluded, slab 0 is undeletable and `stm/user.rs` comments that exact reason.

The clause that does stand is the other one, "the authz layer never re-checks status", and it stands on its own without the session-purge framing. The distinction is which attributes the gate re-reads per op:

- grants: `UpdatePermissionsRequest::apply` calls `permissioner.update_permissions_for_user`, and `authorize` consults the permissioner on every op, so revocation is already live against bound sessions.
- existence: purged on delete, same per-op path, also live.
- status: `UpdateUserRequest::apply` sets `user.status` and never touches the permissioner, and `authz.rs` contains no `status` reference on this branch or on master.

So status is the single authorization-relevant attribute living outside the per-op path, which makes it the only one a bound session can outrun. That also points at a cheaper fix than purging sessions: check status inside `authorize` after the root short-circuit (`&Users` already exposes it in the same borrow), and in the partition path's `authorize_uid`. Deactivation then behaves exactly like permission revocation, which works today. Purging the client table instead would mean a `user_id -> client_ids` reverse scan over up to `CLIENTS_TABLE_MAX` slots inside shard 0's no-await commit region, which is the cost you object to on `evict_oldest`, and `SessionManager` is per-shard transport-local while `DeleteUser` applies on shard 0 only, so that half needs a cross-shard broadcast. Neither buys authority the permissioner does not already enforce.

This is also pre-existing rather than introduced here: master has the same two `remove_client` call sites, the same login-only status check in `auth.rs`, and no status check in the gate. Worth its own issue; happy to file it.

Unrelated to the above, your `dispatch.rs:2553` finding turned up one more thing while I was verifying this. The reclaim I added to fix the leak was itself leaky: `take_expired_reclaims` drains with `retain`, so the entry was gone before the spawned `Logout` submit ran, and a transient failure (`NotPrimary` mid view change, `PipelineFull`, `Canceled`) dropped it permanently, leaving exactly the slot-until-eviction case your comment describes. Now the sweeper re-arms the deadline before submitting and the entry is cleared only when the `Logout` commits, so a failed submit retries after the grace instead of leaking. A repeat is harmless since `submit_logout_*` short-circuits once the slot is gone. Pinned by a unit test covering both halves.

core/consensus/src/client_table.rs
numinnex: Correcting myself on both counts here: I declined the remedy, and I was wrong. It has landed.

`commit_register` now sets the entry's epoch to the register's commit op (`reply.header().commit`, stamped from the prepare's op by `build_reply_message`) instead of `1, +1 per rebind`.

What I missed when I argued for the counter: a counter resets. Drop an entry by capacity eviction and re-register the same key, and the fresh entry is back at epoch 1 -- indistinguishable from the zombie still holding epoch 1 from before the eviction, so the fence is blind exactly where it is most needed. An op-derived fence cannot regress, because ops do not. That is your "log-derived identity" point, and it is the property the counter was quietly trading away.

My "not comparable across planes" objection conflated the fence *value* with the table's *identity*. The key stays client-supplied; only the fence is op-derived. Register commits solely in the metadata group, so there is exactly one minting authority and the value is a plain `u64` for any group's slice to store and compare. Our own design note said as much before the counter drifted in ("Register is always a metadata-plane op, so there is exactly one minting authority. Monotonic within the metadata group = globally comparable everywhere") -- I argued against a decision that had already been made correctly.

The inertness half of your diagnosis is fixed alongside it, because deriving the value changes nothing if no bind ever commits. `submit_register_in_process` lost its fast-path short-circuit and `register_preflight` no longer consults the table (`check_register` and `RequestStatus::AlreadyRegistered` are deleted), so every bind proposes. The ownership gate stays where it was and still refuses another user's entry before dispatching; an owned entry now falls through to a real Register, whose rebind branch refences the entry while preserving its watermark and reply ring. `Fenced`/`EpochAhead` are reachable on the live path for the first time.

One thing that falls out for free: the reconnect-versus-reclaim-sweep race is closed. A sweep's in-flight `Logout` carries the pre-rebind epoch, and `submit_logout_in_process`'s epoch guard drops it rather than tearing down the session the client just rebound.

Cost is a consensus round trip per reconnect, which is the honest price of a bind being a logged event.

core/consensus/src/client_table.rs
numinnex: Withdrawing the second half of this reply. The determinism fix and the provenance note stand; the "it is an improvement, not a regression" argument does not, and it was reasoning from the wrong place.

The trace I gave was accurate as far as it went: master takes the "newer than issued (client bug)" arm and silently drops, this branch produces `Fenced` -> terminal eviction, and an eviction does recover faster than a read timeout. But comparing failure modes was the wrong exercise. The epoch was not supposed to be *diverging* on either branch. I used master's worse handling of a bad state to argue our bad state was fine, instead of asking why the state was reachable at all -- which is what your `client_table.rs:370` comment was pointing at, and which I also pushed back on there.

The actual answer is that a per-entry counter was the wrong derivation, and it is now the register's commit op (see the reply on client_table.rs:370). Under that scheme the spurious-fence case I was defending cannot arise: epochs come from log positions, so replicas that agree on the log agree on every entry's fence, whether or not they made the same local eviction decision. There is no divergent-epoch state left to characterize as better or worse.

The `in_flight` removal you asked for is what makes the victim choice agree in the first place, and that part shipped as described.

core/server-ng/src/dispatch.rs
```diff
@@ -112,6 +112,17 @@ use tracing::{debug, warn};
...
+/// How long a disconnected session's client-table slot is held open for the
+/// client to reconnect and resume onto it before the sweeper reclaims it.
+///
+/// Resume runs through the login path, so the window only has to cover a
+/// reconnect plus a re-login -- seconds. It is generous by an order of
+/// magnitude so a client riding out a view change still lands inside it, and
+/// still bounded so cumulative connects cannot push the client table to its
+/// capacity-eviction point (every eviction silently erases a watermark).
+/// Swept by `run_heartbeat_verifier`, whose tick is far shorter than this.
+const SESSION_RECLAIM_GRACE: std::time::Duration = std::time::Duration::from_mins(1);
```
hubcio: this sweep never runs. `take_expired_reclaims` has exactly one production caller (`run_heartbeat_verifier`), which is only spawned when `config.heartbeat.enabled` is true - and that ships `false` in both `core/server-ng/config.toml` and `core/server/config.toml`. there is no `true` anywhere in the repo.

so the deferral is armed and never collected: `pending_reclaims` grows for the process lifetime and the client-table slot is never released.

core/server-ng/src/dispatch.rs
```diff
@@ -2369,10 +2452,61 @@ where
...
+{
+    let is_group_member = !shard
+        .plane
+        .metadata()
+        .mux_stm
+        .streams()
+        .consumer_group_memberships(vsr_client_id)
+        .is_empty();
+    if is_group_member {
+        submit_session_logout(shard, sessions, vsr_client_id, session);
```
hubcio: this is a regression against master. on master `submit_disconnect_logout` had no `is_group_member` branch at all - it spawned the logout unconditionally, so every disconnect released the slot cluster-wide. now a non-consumer-group disconnect emits nothing, and the sweep meant to catch it is gated off by default (see the comment on `SESSION_RECLAIM_GRACE`).

with 8192 slots and one leaked per non-cg disconnect that reached `Bound`, a node fills up over hours to days depending on churn, and only a process restart resets it. after that every register evicts a live session, and each eviction silently erases that session's dedup watermark.

simplest fix is to delete the grace machinery and restore the unconditional logout - that also removes the `complete_reclaim` session-key bug and the demoted-primary retry loop below. worth a tracking note for when sdk-side stable client ids land, since the window becomes useful then and would need its own always-on timer rather than riding the heartbeat verifier.

please don't fix this by flipping `heartbeat.enabled` - `collect_stale` keys off the heartbeat interval, so ungating it would mass-evict every consumer-group member on a deployment that isn't pinging.

core/shard/src/coordinator.rs
```diff
@@ -143,7 +147,32 @@ impl ShardZeroCoordinator {
...
+    /// recovers client-table entries keyed by the previous boot's ids. Left
+    /// alone, the first logins after a restart re-mint ids that are already
+    /// taken: a different user's login is then refused outright (the register
+    /// ownership gate), and the same user's inherits a watermark it never
+    /// wrote. Seeding past the recovered high-water mark makes the collision
+    /// unreachable instead of handled.
+    ///
+    /// Boot-time only, on shard 0, before any listener accepts. Never lowers
+    /// the counter.
+    pub fn seed_client_sequence(&self, recovered_ids: impl Iterator<Item = u128>) {
```
hubcio: this only runs at boot - `bootstrap.rs` is the sole call site. but the client table is replicated, so a node keeps accumulating entries minted by other nodes' counters long after its own seed ran, and nothing reseeds on promotion.

after a failover the new primary mints from a counter that has no relationship to the ids its predecessor committed. the two are uncorrelated, so they collide whenever the promoted node's counter is at or below the old primary's high-water mark. `MAX_HTTP_SESSIONS` is 4096, so a lot of consecutive http logins can land on an occupied entry.

different user hits the ownership gate and gets a terminal refusal for a perfectly legitimate login. same user falls through and rebinds, inheriting a watermark written by a different session of that user.

reseeding on promotion as well as at boot removes the trigger outright, and it's the smallest of the fixes in this area.

separately: the mask folds every recovered id including caller-supplied ones, and `mint_client_id` never bounds `seq`. one login with a chosen `client_id` near the top of the 112-bit range plus a restart overflows into the shard tag, and `client_id_owning_shard` is the inter-shard reply routing key - so replies for new connections would forward to the wrong shard for the process lifetime. seeding only from ids where `(id >> 112) < total_shards` and bounding `seq` covers it.

core/server-ng/src/dispatch.rs
```diff
@@ -519,13 +535,14 @@ where
...
                 } => {
-                    let session = shard
+                    let bound = shard
                         .plane
                         .metadata()
                         .submit_register_in_process(vsr_client_id, user_id)
                         .await
-                        .ok();
-                    let _ = reply.try_send(session);
+                        .ok()
```
hubcio: `.ok()` throws the error away, so `ClientIdOwnedByAnotherUser` arrives as `None` and `submit_register_on_owner` turns it into `Canceled`. `LoginRegisterError::is_terminal` then reads `Canceled.is_transient()` as true and answers the login as transient.

that matters because login is special-cased twice in the sdk: `tcp_client.rs` gives it the full `RESPONSE_READ_TIMEOUT` instead of the 2 s transient window, and the failover arm explicitly excludes login codes. so the client replays against a 30 s budget at a 50 ms interval with no leader redirect - order of 600 attempts - and every one costs a full argon2 verify on the home shard's event loop (`Argon2::default()`, m=19456 KiB).

only shard 0 takes the in-process path, so this is every connection that round-robins elsewhere.

putting `BoundSession` plus the real error on this channel fixes it and drops the untyped `(u64, u64)` tuple at the same time.

core/server-ng/src/http/state.rs
```diff
@@ -253,14 +253,23 @@ impl HttpInner {
                     _ => AuthError::SessionUnavailable,
```
hubcio: `ClientIdOwnedByAnotherUser` falls into the `_` arm here, so it becomes `SessionUnavailable` and then a bare 503.

worth flagging because it means fixing the `.ok()` at the shard boundary changes nothing for http - both paths independently downgrade a permanent, deterministic refusal to a retryable status. and 503 is about the most auto-retried status there is: none of the foreign sdks have any 503 handling, so whatever the integrator's http stack does by default takes over.

the diagnostic story is rough too - the actionable line is the `iggy.metadata.diag` warn in `submit_register_in_process`, while the default log view shows `Canceled` here, which reads as transient.

a dedicated arm mapping to a terminal 4xx would fix it - 409 fits, the id is taken.

core/metadata/src/impls/metadata.rs
```diff
@@ -969,35 +1048,67 @@ where
...
+            {
+                warn!(
+                    target: "iggy.metadata.diag",
+                    client_id,
+                    authenticated_user = user_id,
+                    entry_owner = owner,
+                    "refusing register: client id is registered to a different user"
+                );
+                return Err(MetadataSubmitError::ClientIdOwnedByAnotherUser);
+            }
```
hubcio, line 1128: this gate reads the replicated table authoritatively but sits above the primary check below it, and there's no `is_caught_up_primary` anywhere before it.

both sibling readers of this table gate first - `request_preflight` and `register_preflight` both call the catch-up gate, and `request_preflight` has a SAFETY comment justifying it as what makes the replica authoritative for session truth. this one consults the same table with neither guard, so a backup answers the ownership refusal instead of `NotPrimary` and the client never learns to redirect.

moving it below the catch-up check makes it consistent with the other two and stops a lagging or diverged replica from issuing a terminal refusal.

core/metadata/src/impls/metadata.rs
```diff
@@ -1067,21 +1175,82 @@ where
...
-                // (`session = commit-op`, deterministic). Own surviving
-                // Register would have routed through `AlreadyRegistered`
-                // against the same entry, so no "this primary vs inherited
-                // primary" split.
-                self.client_table
-                    .borrow()
-                    .get_session(client_id)
-                    .ok_or(MetadataSubmitError::Canceled)
+        // Same commit/cancel collapse as the queued path above.
+        let _ = self.dispatch_prepare_and_await(consensus, prepare).await;
```
hubcio: `let _ =` discards the committed reply, and `bound_session` then re-reads the table. the reply header already carries this register's own epoch - `build_reply_message` stamps `commit` from the prepare's op.

this is a regression against master, which had `Ok(reply) => Ok(reply.header().commit)` on the queued arm and only re-read the table on cancel. the cancel path is where it bites: if the register is canceled by a view change but some earlier entry for that `client_id` survives, `bound_session` hands back the old epoch and the login succeeds with a fence that never moved.

not reachable today since nothing re-presents a client id, but it goes live the moment sdk identity stability lands, and it's silent when it fires - a stale epoch passes `check_request`'s equality test, so no `Fenced`, no `EpochAhead`, nothing.

taking `epoch` from `reply.header().commit` and reading only `watermark` from the table is the two-line fix.

core/server-ng/src/http/state.rs
```diff
@@ -253,14 +253,23 @@ impl HttpInner {
...
                 }
             })?;
+        // Number from above the entry's watermark, not from 1. The gateway is
+        // the one client whose request counter lives in the process that
+        // restarts: after a reboot the id minter can re-mint this client id
+        // for the same user, binding to the recovered entry -- numbering from
+        // FIRST_REQUEST_ID would then read as duplicates and answer this
+        // session's writes with the previous boot's cached replies. A fresh
+        // entry has watermark 0, so this is FIRST_REQUEST_ID there.
```
hubcio: this can only ever be `FIRST_REQUEST_ID` on a single node, which makes the whole `BoundSession` watermark path dead there.

the comment describes resume as the minter re-minting a client id onto a recovered entry - but that was never resume, it was an accidental collision, and `seed_client_sequence` was added in this same commit specifically to make it unreachable ("seeding past the recovered high-water mark makes the collision unreachable instead of handled"). so a fresh mint is provably above every recovered id, every register takes `commit_register`'s fresh-entry branch, and `watermark` is `REGISTER_REQUEST_ID`.

net effect is that two fixes in this commit cancel: the watermark plumbing added for resume, and the reseed that removed the only mechanism producing a non-zero watermark.

the exception is a cluster - after a failover a promoted node can mint onto an existing entry and a same-user collision rebinds, so the value is non-zero but inherited from a different session of that user. that's worse than dead, not better.

worth deciding whether the resume machinery lands now alongside the sdk half, or comes out and the pr description is adjusted.

core/sdk/src/vsr.rs
```diff
@@ -47,11 +47,10 @@ const NON_REPLICATED_CODE_RANGE: std::ops::Range<usize> = 0..4;
...
 // client/session, the retried request gets a new (client_id, request_id) tuple, so
 // server-side deduplication cannot match a mutation that may already have committed
-// before the transport failure. TigerBeetle avoids session resume and relies on
-// idempotency for requests retried from a new client session. For Iggy, either stop
-// transparent retries for replicated mutations after VSR session reset, add explicit
-// session resume/rebind semantics, or add a protocol-level idempotency key that is
-// independent of (client_id, request_id).
+// before the transport failure. The server now rebinds a transport that presents
+// its old (client, session) identity (`try_resume_session`), so the fix here is
+// to keep the ConsensusSession across reconnects and retry replicated writes
```
hubcio: this todo points the next implementer at `try_resume_session`, which this pr deletes as a pre-auth session takeover. so the guidance is now to build against a mechanism that was removed for security reasons.

awkward because the integration test module doc nominates "the `sdk/vsr.rs` retry todo" as the remaining client half - so the one file the pr names as the follow-up is the one file that misdirects it.

the advice underneath is still right (keep the session across reconnects, retry under the same identity), it just needs to say resume happens through the login path now.

core/consensus/src/metadata_helpers.rs
```diff
@@ -116,29 +129,55 @@ where
         // Session evicted under capacity pressure. SAFETY: catch-up gate makes
         // this replica authoritative for session truth.
         RequestStatus::NoSession => PreflightOutcome::Evict(EvictionReason::NoSession),
```
hubcio, line 147: this claim doesn't hold for a node whose table was rebuilt from a truncated wal.

every clause of `is_caught_up_primary` is about the log - applied what it holds, suffix re-earned quorum. none of it says anything about how the table was built. but `recover()` starts from the local snapshot floor, and checkpoints are node-local: `checkpoint_if_needed` fires on local journal occupancy, the floor is the local `commit_min`, and the margin is per-node config. so two replicas cross the boundary at different ops.

the apply paths are deterministic, the initial condition isn't, and this gate checks the wrong quantity to bridge them. either the comment needs correcting or the check needs adding - the second is the already-acknowledged table-in-snapshot work.

core/metadata/src/impls/recovery.rs
```diff
@@ -200,6 +211,26 @@ where
             continue;
         }
```
hubcio, line 213: the epochs part is true - they're op-derived so they replay identically. the watermark isn't.

if an earlier register for a client fell below this node's snapshot floor but a later one survived, replay takes `commit_register`'s fresh-entry branch and the entry comes back with a valid epoch and `watermark = 0`, where a peer that took the rebind branch preserved it. the fence passes, so nothing is evicted, and the same request id a peer answers `Duplicate` gets answered `New` and re-executed.

no shipping client can observe this today since none re-presents a recovered id, but it becomes a silent exactly-once to at-least-once violation once identity stability lands. the comment asserting cross-replica determinism here is what would stop a reader from asking.

the test suite can't catch it either - `client_table_restart.rs` uses a two-op wal, so the snapshot floor is always 0. one test that forces a checkpoint before the restart would turn this into a red/green fact.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -152,12 +152,22 @@ pub struct RequestHeader {
...
     pub request_checksum: u128,
     pub timestamp: u64,
     pub request: u64,
     pub operation: Operation,
     pub operation_padding: [u8; 7],
     pub namespace: u64,
+    /// Session fence epoch, handed to the client by its `Register` reply and
+    /// echoed on every subsequent request. Bumped by the server on each
+    /// committed re-register of the same `client`, so a request stamped with an
+    /// older value is a zombie from before a rebind and gets fenced. Unrelated
```
hubcio: last sentence is no longer true - `commit_register` sets the epoch to `reply.header().commit`, which is the register's commit op, so it derives from the log directly. `client_table.rs` says so in this same commit ("op-derived fence", "the commit number becomes the session number").

also "bumped by the server on each committed re-register" reads as a counter, which it was before this pr and isn't now.

worth getting exact since this struct is the wire contract.

core/server-ng/src/session_manager.rs
```diff
@@ -166,6 +181,66 @@ impl SessionManager {
...
+    }
+
+    /// Drop a pending reclaim because its `Logout` committed and the slot is
+    /// released. Separate from [`Self::cancel_reclaim`] only in intent: the
+    /// reclaim submit is fire-and-forget and can fail transiently, so the
+    /// sweeper re-arms the deadline before submitting and the entry is cleared
+    /// here on success. Without that pairing a failed submit would drop the
+    /// bookkeeping and leak the slot until capacity eviction, which silently
+    /// erases the client's dedup watermark.
+    pub fn complete_reclaim(&mut self, client_id: u128) {
```
hubcio: this removes by `client_id` alone, so it can delete a reclaim armed for a newer epoch.

the sequence: sweeper takes the reclaim for epoch A, re-arms, spawns the logout. client reconnects and `bind_session` cancels. client disconnects again, arming epoch B. the in-flight logout for A hits the epoch guard in `submit_logout_in_process`, which returns `Ok(commit_min)` - a success value for a logout that logged nothing out - and the spawned task calls `complete_reclaim`, dropping B's entry. slot then leaks until capacity eviction, which is exactly what the grace window exists to prevent.

not reachable while the sweep is gated off, but it goes live the moment that's fixed. compare-and-remove on `(client_id, session)` covers it.

same applies to the demoted-primary case - `submit_logout_in_process` returns `NotPrimary` there and the sweeper re-arms every tick forever, using the grace constant as the retry backoff.

core/consensus/src/client_table.rs
```diff
@@ -156,181 +236,190 @@ impl ClientTable {
         }
     }
 
-    /// Check request against table. Session first, then request progression.
-    /// For Register, use [`check_register`].
+    /// Check a request against the table. Epoch fence first, then the
+    /// watermark. For Register, use [`Self::check_register`].
```
hubcio: `check_register` was deleted in this commit, so this intra-doc link dangles. `cargo doc` in ci doesn't set `RUSTDOCFLAGS`, so it warns and still exits 0 - it'll just sit in the log.

there's a prose reference to `check_register` in `metadata_helpers.rs` too.

core/integration/tests/cluster/client_table_restart.rs
```diff
@@ -235,22 +344,64 @@ async fn commit_request(stream: &mut TcpStream, session: u64, request: u64, body
...
             continue;
         };
+        // Re-authenticate on the fresh connection. The rebind commits a
+        // Register, so the epoch strictly advances past the pre-restart one
+        // (op-derived; regression would mean the fence can be replayed into).
+        let resumed = match login_on(&mut stream).await {
+            Some(resumed) => {
+                assert!(
+                    resumed > session,
+                    "rebind must refence above the pre-restart epoch                      (old {session}, got {resumed})"
```
hubcio: looks like a botched line join - there's a run of spaces mid-message and the line runs past 100 columns.

core/server-ng/src/http/session.rs
```diff
@@ -37,10 +37,12 @@ use tokio::sync::Mutex;
...
-/// which LRU-evicts the oldest client when full. Capping HTTP at half that
-/// bound keeps this plane from crowding the others out and keeps the combined
-/// steady state under the shared bound, so a live idle HTTP session is not
-/// routinely evicted consensus-side. The residual eviction race (both planes
+/// which evicts the oldest-committed client when full. Capping HTTP at half
+/// that bound keeps this plane from crowding the others out and keeps the
+/// combined steady state under the shared bound, so a live idle HTTP session
+/// is not routinely evicted consensus-side. The other half stays honest
+/// because a dropped connection releases its slot once the reclaim grace
+/// expires, rather than holding it for the process lifetime. The residual eviction race (both planes
```
hubcio: this rationale doesn't hold as written. the added clause says a dropped connection releases its slot once the reclaim grace expires, but the sweep that would do that is behind `heartbeat.enabled`, which ships false - so a dropped non-consumer-group connection holds its slot for the process lifetime, which is the case the sentence says doesn't happen.

the line is also over 100 columns after the edit.

hubcio (changes requested):

hubcio (dismissed):

spetz (approved):

hubcio (approved):

krishvishal (approved):

PR #3754, opened 2026-07-27
fix(server-ng): legacy-parity consumer group errors, roster HTTP bind

hubcio:
Consumer group join/leave/delete on the ng metadata plane
committed silent OK no-ops (or collapsed every miss into one
code) when the stream, topic, group or membership was missing,
so clients saw success where the legacy server reports an
error. All three ops now resolve level by level and commit the
legacy codes (1009/2010/5000, and 5006 for leaving as an absent
member), pinned end to end by an integration test across TCP,
QUIC and WebSocket; HTTP cannot run the join/leave ladder since
its stateless sessions carry no member identity. Consumer group
ids become 0-based to match the legacy server, asserted on the
wire-visible create response id; server-ng has no stable-format
users yet.

The HTTP listener bound from the top-level [http].address even
in cluster mode, so one-host clusters collided on the shared
port. Cluster listeners now take their port from the node's
roster entry while keeping the bind interface from each
transport's own address, so a loopback-only or 0.0.0.0 host
survives cluster mode and the roster ip stays purely the
advertised address. Roster ports are mandatory for every
enabled transport, validated at startup instead of falling
back to a shared default port that two same-host nodes would
race for, letting every node run from byte-identical per-node
configs.

### DIFF
--- a/core/server-ng/src/bootstrap.rs
+++ b/core/server-ng/src/bootstrap.rs
@@ -481,6 +481,7 @@ struct TcpTopology {
     replica_listen_addr: Option<SocketAddr>,
     ws_listen_addr: Option<SocketAddr>,
     quic_listen_addr: Option<SocketAddr>,
+    http_listen_addr: Option<SocketAddr>,
     tcp_tls_listen_addr: Option<SocketAddr>,
     peers: Vec<(u8, SocketAddr)>,
 }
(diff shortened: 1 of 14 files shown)


core/server-ng/src/bootstrap.rs
```diff
@@ -2112,31 +2117,52 @@ fn resolve_optional_listener_addr(
...
         default_quic_addr,
         |ports| ports.quic,
     )?;
-    Ok((client_listen_addr, ws_listen_addr, quic_listen_addr))
+    let http = resolve_cluster_optional_addr(
+        self_node,
+        "cluster.nodes[*].ports.http",
+        default_http_addr,
+        |ports| ports.http,
+    )?;
```
krishvishal: Cluster mode discards the host from `[http].address`.
`bootstrap.rs:2154-2159` -> `resolve_cluster_optional_addr:2168-2179` takes only
`default_addr.port()`, then `socket_addr_from_parts:3014-3027` =
`SocketAddr::new(self_node.ip, port)`. Operator host dropped, not merged.

- `config.toml:41` = `127.0.0.1:3000`, documented `:40` "listens on localhost
  only"; example node `ip = "192.168.1.100"` (`:644`). So `cluster.enabled = true`
  silently moves JWT minting and `/ui` from loopback to the cluster interface.
  No log, no warning.
- `Dockerfile:58`, `helm/.../values.yaml:95`, `web/docker-compose.yml:24` all
  `0.0.0.0:3000` (helm `README.md:209` requires it). Those narrow to the roster
  IP, breaking loopback probes and sidecar scrapes.
- Validation misses both: `cluster.rs:257-276` guards `if let Some(port)`, so
  unset ports resolving to one shared endpoint are never compared.
  `TransportPorts.http` (`cluster.rs:160`) undocumented.

Fix: `SocketAddr::new(default_http_addr.ip(), roster_port_or_default)`. If
ws/quic symmetry is preferred, log the effective bind when it differs from
`http.address` and document it on `TransportPorts` + `[cluster]`.

hubcio: good catch,cluster now binds config-address ip + roster port; ports mandatory (fallback removed).

krishvishal (dismissed):

krishvishal (dismissed):

krishvishal (approved):

numinnex (approved):

PR #3755, opened 2026-07-27
feat(sdk): let raw binary requests declare their replication kind

spetz:
The raw command path only accepted codes the SDK already knew.
Under VSR, its closed operation table rejected unknown codes before
any bytes left the client, so the raw API was not actually open.

Make replication kind explicit across all nine SDKs in one deliberate
API break. Classic transports ignore the declaration and retain
byte-identical framing.

Rust now applies the declaration to VSR framing, and Python inherits
that behavior through its Rust bindings. Add native VSR framing to the
TypeScript SDK, selected at runtime with `protocol: "vsr"`.

The TypeScript client now implements registration, session and request
tracking, operation and namespace routing, committed reply decoding,
evictions, transient retries, leader redirects, and consumer-group
assignment synchronization.

Protocol tables remain authoritative. A shipped code keeps its class,
an unknown non-replicated code is forwarded for the server to judge,
and an unknown replicated code fails until a replicated extension
registry exists. Login and logout remain rejected before any I/O on
every transport.

### DIFF
--- a/bdd/cpp/features/step_definitions/raw_command_steps.cpp
+++ b/bdd/cpp/features/step_definitions/raw_command_steps.cpp
@@ -25,18 +25,35 @@
 
 #include <cstdint>
 #include <exception>
+#include <stdexcept>
 #include <string>
 
 #include "world.hpp"
 
-WHEN("^I send a raw command with code ([0-9]+) and an empty payload$") {
+namespace {
+
+iggy::ffi::BinaryRequestKind parse_request_kind(const std::string &kind) {
+    if (kind == "non_replicated") {
+        return iggy::ffi::BinaryRequestKind::NonReplicated;
+    }
+    if (kind == "replicated") {
+        return iggy::ffi::BinaryRequestKind::Replicated;
+    }
+    throw std::invalid_argument("unknown binary request kind: " + kind);
+}
+
+}  // namespace
+
+WHEN("^I send a raw (\\w+) command with code ([0-9]+) and an empty payload$") {
+    REGEX_PARAM(std::string, kind);
     REGEX_PARAM(std::uint32_t, code);
     cucumber::ScenarioScope<bdd::GlobalContext> context;
 
     context->raw_response.clear();
     context->raw_error.clear();
     try {
-        const auto response = context->client->send_binary_request(code, rust::Vec<std::uint8_t>());
+        const auto response =
+            context->client->send_binary_request(parse_request_kind(kind), code, rust::Vec<std::uint8_t>());
         context->raw_response.assign(response.begin(), response.end());
     } catch (const std::exception &error) {
         context->raw_error = error.what();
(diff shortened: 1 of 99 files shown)


bdd/cpp/features/step_definitions/raw_command_steps.cpp
```diff
@@ -25,18 +25,35 @@
...
+namespace {
+
+iggy::ffi::BinaryRequestKind parse_request_kind(const std::string &kind) {
+    if (kind == "non_replicated") {
+        return iggy::ffi::BinaryRequestKind::NonReplicated;
+    }
+    if (kind == "replicated") {
+        return iggy::ffi::BinaryRequestKind::Replicated;
+    }
+    throw std::invalid_argument("unknown binary request kind: " + kind);
```
slbotbm: Let's inline this since there is only 1 call site.

foreign/cpp/src/client.rs
```diff
@@ -37,6 +37,18 @@ use std::sync::Arc;
 /// reserve `u32::MAX` as the sentinel for `partition_id`.
 const ANY_PARTITION_ID: u32 = u32::MAX;
 
+/// A C++ `enum class` can hold any value of its underlying type, so cxx models
+/// a shared enum as an open repr and the catch-all arm is reachable.
+fn resolve_binary_request_kind(
+    kind: ffi::BinaryRequestKind,
+) -> Result<RustBinaryRequestKind, String> {
```
slbotbm: Let's convert this to `TryFrom<ffi::BinaryRequestKind> for RustBinaryRequestKind` in type_conversion.rs

foreign/cpp/Cargo.toml
```diff
@@ -17,7 +17,7 @@
 
 [package]
 name = "iggy-cpp"
-version = "0.1.1"
+version = "0.1.2"
```
slbotbm, line 20: It would be good to update the bazel version as well.

foreign/cpp/Cargo.toml
```diff
@@ -26,6 +26,12 @@ ignored = ["cxx-build"]
 [lib]
 crate-type = ["staticlib"]
 
+# Framing is a compile-time choice, so the static library is locked to whatever it
+# was built with. Opt in with `--features vsr` to build against the consensus
+# protocol; the default build stays on the classic protocol.
+[features]
+vsr = ["iggy/vsr", "iggy_common/vsr"]
```
slbotbm, line 33: We utilize bazel as the build system, and we'll need to add this feature flag there as well.

slbotbm (changes requested):

hubcio:
i'm against this PR as it stands - it's three PRs in one: a breaking API change across nine SDKs, a whole new VSR client for typescript, and an unrelated topic decoder fix.

main issue is the rust SDK isn't stable yet. we're still moving VSR framing around underneath it, so breaking eight other SDKs to match itlocks in an interface for a design that's still in progress.

spetz:
Ok, fair point, I'll split this PR into one for Rust + Python only and the second one for TS changes. All the other SDKs will follow the changes once VSR framing is implemented in them.

PR #3763, opened 2026-07-28
feat(js): add VSR framing to the Node SDK

spetz:
The Node SDK only supported classic binary framing, preventing it
from connecting to VSR servers.

Add runtime-selectable VSR framing with registration, session
recovery, request sequencing, bounded response decoding, transient
retries, leader redirection, and consumer-group support.

Keep the existing sendBinaryRequest(code, payload) API. Known commands
derive their VSR operation from the protocol registry. Codes unknown
to the SDK are sent as non-replicated requests with the command code
stored in RequestHeader.reserved, allowing the server to handle an
extension or return the authoritative error. Session-control commands
remain blocked before I/O.

### DIFF
--- a/.github/actions/node-npm/pre-merge/action.yml
+++ b/.github/actions/node-npm/pre-merge/action.yml
@@ -20,7 +20,7 @@ description: Node.js pre-merge testing github iggy actions
 
 inputs:
   task:
-    description: "Task to run (lint, test, build, e2e)"
+    description: "Task to run (lint, test, build, e2e, e2e-vsr)"
     required: true
 
 runs:
@@ -39,11 +39,11 @@ runs:
       shell: bash
 
     - name: Setup Rust with cache
-      if: inputs.task == 'e2e'
+      if: inputs.task == 'e2e' || inputs.task == 'e2e-vsr'
       uses: ./.github/actions/utils/setup-rust-with-cache
 
     - name: Install netcat
-      if: inputs.task == 'e2e'
+      if: inputs.task == 'e2e' || inputs.task == 'e2e-vsr'
       run: sudo apt-get update && sudo apt-get install -y netcat-openbsd
       shell: bash
 
@@ -82,8 +82,6 @@ runs:
           echo "Warning: dist directory not found after build"
         fi
       shell: bash
-
-
     - name: Test
       if: inputs.task == 'test'
       run: |
@@ -114,6 +112,47 @@ runs:
         IGGY_SERVER_HOST: 127.0.0.1
         IGGY_SERVER_TCP_PORT: 8090
       shell: bash
+
+    - name: Start Iggy VSR server
+      id: iggy-vsr
+      if: inputs.task == 'e2e-vsr'
+      uses: ./.github/actions/utils/server-start
+      with:
+        cargo-bin: iggy-server-ng
+        pid-file: ${{ runner.temp }}/iggy-node-vsr.pid
+        log-file: ${{ runner.temp }}/iggy-node-vsr.log
+        wait-timeout-seconds: "90"
+      env:
+        IGGY_SYSTEM_PATH: ${{ runner.temp }}/iggy-node-vsr-data
+
+    - name: VSR E2E tests
+      if: inputs.task == 'e2e-vsr'
+      run: |
+        cd foreign/node
+        mkdir -p ../../reports/node-coverage/e2e-vsr
+        npx c8 --reporter=lcov \
+          --reports-dir=../../reports/node-coverage/e2e-vsr \
+          npm run test:e2e:vsr
+      env:
+        IGGY_SERVER_HOST: 127.0.0.1
+        IGGY_SERVER_TCP_PORT: 8090
+        IGGY_TEST_PROTOCOL: vsr
+      shell: bash
+
+    - name: Stop Iggy VSR server
+      if: always() && inputs.task == 'e2e-vsr'
+      uses: ./.github/actions/utils/server-stop
+      with:
+        pid-file: ${{ steps.iggy-vsr.outputs.pid_file }}
+        log-file: ${{ steps.iggy-vsr.outputs.log_file }}
+
+    - name: Upload VSR server logs
+      if: always() && inputs.task == 'e2e-vsr'
+      uses: actions/upload-artifact@v7
+      with:
+        name: iggy-node-vsr-server-logs
+        path: ${{ steps.iggy-vsr.outputs.log_file }}
+        if-no-files-found: ignore
 
     - name: Stop Iggy server (plain)
       if: always() && inputs.task == 'e2e'
--- a/.github/actions/node-npm/pre-merge/action.yml
+++ b/.github/actions/node-npm/pre-merge/action.yml
@@ -113,17 +113,37 @@ runs:
         IGGY_SERVER_TCP_PORT: 8090
       shell: bash
 
-    - name: Start Iggy VSR server
-      id: iggy-vsr
+    - name: Start Iggy VSR server 0
+      id: iggy-vsr-0
       if: inputs.task == 'e2e-vsr'
       uses: ./.github/actions/utils/server-start
       with:
         cargo-bin: iggy-server-ng
-        pid-file: ${{ runner.temp }}/iggy-node-vsr.pid
-        log-file: ${{ runner.temp }}/iggy-node-vsr.log
+        cargo-features: vsr
+        replica-id: "0"
+        pid-file: ${{ runner.temp }}/iggy-node-vsr-0.pid
+        log-file: ${{ runner.temp }}/iggy-node-vsr-0.log
         wait-timeout-seconds: "90"
       env:
-        IGGY_SYSTEM_PATH: ${{ runner.temp }}/iggy-node-vsr-data
+        IGGY_CLUSTER_ENABLED: "true"
+        IGGY_SYSTEM_PATH: ${{ runner.temp }}/iggy-node-vsr-0-data
+
+    - name: Start Iggy VSR server 1
+      id: iggy-vsr-1
+      if: inputs.task == 'e2e-vsr'
+      uses: ./.github/actions/utils/server-start
+      with:
+        cargo-bin: iggy-server-ng
+        cargo-features: vsr
+        replica-id: "1"
+        tcp_address: 127.0.0.1:8091
+        http_address: 127.0.0.1:3001
+        pid-file: ${{ runner.temp }}/iggy-node-vsr-1.pid
+        log-file: ${{ runner.temp }}/iggy-node-vsr-1.log
+        wait-timeout-seconds: "90"
+      env:
+        IGGY_CLUSTER_ENABLED: "true"
+        IGGY_SYSTEM_PATH: ${{ runner.temp }}/iggy-node-vsr-1-data
 
     - name: VSR E2E tests
       if: inputs.task == 'e2e-vsr'
@@ -134,24 +154,31 @@ runs:
           --reports-dir=../../reports/node-coverage/e2e-vsr \
           npm run test:e2e:vsr
       env:
-        IGGY_SERVER_HOST: 127.0.0.1
-        IGGY_SERVER_TCP_PORT: 8090
         IGGY_TEST_PROTOCOL: vsr
       shell: bash
 
-    - name: Stop Iggy VSR server
+    - name: Stop Iggy VSR server 1
       if: always() && inputs.task == 'e2e-vsr'
       uses: ./.github/actions/utils/server-stop
       with:
-        pid-file: ${{ steps.iggy-vsr.outputs.pid_file }}
-        log-file: ${{ steps.iggy-vsr.outputs.log_file }}
+        pid-file: ${{ steps.iggy-vsr-1.outputs.pid_file }}
+        log-file: ${{ steps.iggy-vsr-1.outputs.log_file }}
+
+    - name: Stop Iggy VSR server 0
+      if: always() && inputs.task == 'e2e-vsr'
+      uses: ./.github/actions/utils/server-stop
+      with:
+        pid-file: ${{ steps.iggy-vsr-0.outputs.pid_file }}
+        log-file: ${{ steps.iggy-vsr-0.outputs.log_file }}
 
     - name: Upload VSR server logs
       if: always() && inputs.task == 'e2e-vsr'
       uses: actions/upload-artifact@v7
       with:
         name: iggy-node-vsr-server-logs
-        path: ${{ steps.iggy-vsr.outputs.log_file }}
+        path: |
+          ${{ steps.iggy-vsr-0.outputs.log_file }}
+          ${{ steps.iggy-vsr-1.outputs.log_file }}
         if-no-files-found: ignore
 
     - name: Stop Iggy server (plain)
--- a/codecov.yml
+++ b/codecov.yml
@@ -62,7 +62,7 @@ flag_management:
     - name: php
       paths:
         - foreign/php/
-    - name: node-test
+    - name: node
       paths:
         - foreign/node/
     - name: node-e2e
--- a/codecov.yml
+++ b/codecov.yml
@@ -62,7 +62,13 @@ flag_management:
     - name: php
       paths:
         - foreign/php/
-    - name: node
+    - name: node-test
+      paths:
+        - foreign/node/
+    - name: node-e2e
+      paths:
+        - foreign/node/
+    - name: node-e2e-vsr
       paths:
         - foreign/node/
     - name: go
--- a/foreign/node/README.md
+++ b/foreign/node/README.md
@@ -30,6 +30,89 @@ npm i --save apache-iggy
 
 ## basic usage
 
+### VSR framing
+
+Classic framing remains the default. Select VSR explicitly when connecting to
+an Iggy VSR server:
+
+```typescript
+import {
+  BinaryRequestKind,
+  SimpleClient,
+  getRawClient,
+} from "apache-iggy";
+
+const config = {
+  protocol: "vsr" as const,
+  transport: "TCP" as const,
+  options: { host: "127.0.0.1", port: 8090 },
+  credentials: { username: "iggy", password: "iggy" },
+};
+const client = new SimpleClient(getRawClient(config));
+const response = await client.sendBinaryRequest(
+  BinaryRequestKind.NonReplicated,
+  60_000,
+  Buffer.from("opaque mutation"),
+);
+```
+
+VSR is a runtime protocol choice in Node.js, not a build feature. Custom
+non-replicated codes use `Operation::NonReplicated`; custom replicated codes
+are rejected until a server-side replicated extension registry exists.
+
+The same npm package supports both framing modes. VSR currently supports TCP
+only and restricts `Client` to one pooled connection because authentication,
+request sequencing, and consumer-group assignments belong to one consensus
+session. Configurations requesting VSR over TLS or more than one pooled
+connection fail before a socket is opened.
+
+Response frames larger than `maxResponseFrameSize` (default 64 MiB) are
+rejected and close the connection under both framing modes. Raise the limit
+in the client configuration when polling very large batches.
+
+VSR authentication translates the existing password and personal-access-token
+login APIs into the register handshake required by the consensus protocol. A
+disconnect or eviction invalidates the session, and later work must register a
+new session. Transient not-committed responses retry the exact encoded request
+within one bounded deadline. A disconnected mutation is never replayed under a
+new session.
+
+`sendBinaryRequest` has one intentionally breaking signature:
+
+```typescript
+sendBinaryRequest(
+  kind: BinaryRequestKind,
+  code: number,
+  payload: Buffer,
+): Promise<Buffer>
+```
+
+Migrate calls from:
+
+```typescript
+await client.sendBinaryRequest(code, payload);
+```
+
+to:
+
+```typescript
+await client.sendBinaryRequest(
+  BinaryRequestKind.NonReplicated,
+  code,
+  payload,
+);
+```
+
+There is no compatibility overload or `sendBinaryRequestWithKind` method.
+Known command tables remain authoritative under VSR: a conflicting declaration
+is rejected, unknown non-replicated codes reach the server, and unknown
+replicated codes fail locally until the extension registry exists. The kind is
+not serialized by classic framing, so classic request bytes remain unchanged.
+
+The client includes its npm package version and the binary protocol crate
+version in VSR registration. An incompatible server rejects registration with
+a protocol-version error instead of accepting a mismatched wire contract.
+
 ```ts
 import { Client } from "apache-iggy";
 
@@ -60,7 +143,8 @@ npm run build
 
 ### test
 
-note: use env var `IGGY_TCP_ADDRESS="host:port"` to set server address for bdd and e2e tests.
+note: use env var `IGGY_TCP_ADDRESS="host:port"` to set the server
+address for bdd and e2e tests.
 
 #### unit tests
 
--- a/foreign/node/README.md
+++ b/foreign/node/README.md
@@ -30,6 +30,10 @@ npm i --save apache-iggy
 
 ## basic usage
 
+### Response frame limit
+
+**Compatibility note:** response frames larger than `maxResponseFrameSize` (default 64 MiB) are now rejected and close the connection under both framing modes. This is a behavior change for existing classic-framing clients. Raise the limit in the client configuration when polling very large batches.
+
 ### VSR framing
 
 Classic framing remains the default. Select VSR explicitly when connecting to
@@ -45,10 +49,7 @@ const config = {
   credentials: { username: "iggy", password: "iggy" },
 };
 const client = new SimpleClient(getRawClient(config));
-const response = await client.sendBinaryRequest(
-  60_000,
-  Buffer.from("opaque mutation"),
-);
+const stats = await client.system.getStats();
 ```
 
 VSR is a runtime protocol choice in Node.js, not a build feature. Codes absent
@@ -62,10 +63,6 @@ request sequencing, and consumer-group assignments belong to one consensus
 session. Configurations requesting VSR over TLS or more than one pooled
 connection fail before a socket is opened.
 
-Response frames larger than `maxResponseFrameSize` (default 64 MiB) are
-rejected and close the connection under both framing modes. Raise the limit
-in the client configuration when polling very large batches.
-
 VSR authentication translates the existing password and personal-access-token
 login APIs into the register handshake required by the consensus protocol. A
 disconnect or eviction invalidates the session, and later work must register a
@@ -73,10 +70,21 @@ new session. Transient not-committed responses retry the exact encoded request
 within one bounded deadline. A disconnected mutation is never replayed under a
 new session.
 
-`sendBinaryRequest(code, payload)` has the same signature under classic and
-VSR framing. Known replicated commands use their registered operation, while
-unknown codes reach the server as non-replicated requests. Classic request
-bytes remain unchanged.
+When the server's `[heartbeat]` eviction is enabled, configure the client's `heartbeatInterval` below the server heartbeat interval. Client heartbeats are disabled when `heartbeatInterval` is unset.
+
+`sendBinaryRequest(code, payload)` has the same signature under classic and VSR framing. Known replicated commands use their registered operation, while unknown codes reach the server as non-replicated requests and are rejected by servers that do not register them. Classic request bytes remain unchanged.
+
+```typescript
+import { ResponseError } from "apache-iggy";
+
+try {
+  await client.sendBinaryRequest(60_000, Buffer.from("opaque request"));
+} catch (error) {
+  if (error instanceof ResponseError) {
+    console.error(error.commandCode, error.errorCode);
+  }
+}
+```
 
 The client includes its npm package version and the binary protocol crate
 version in VSR registration. An incompatible server rejects registration with
--- a/foreign/node/README.md
+++ b/foreign/node/README.md
@@ -36,11 +36,7 @@ Classic framing remains the default. Select VSR explicitly when connecting to
 an Iggy VSR server:
 
 ```typescript
-import {
-  BinaryRequestKind,
-  SimpleClient,
-  getRawClient,
-} from "apache-iggy";
+import { SimpleClient, getRawClient } from "apache-iggy";
 
 const config = {
   protocol: "vsr" as const,
@@ -50,15 +46,15 @@ const config = {
 };
 const client = new SimpleClient(getRawClient(config));
 const response = await client.sendBinaryRequest(
-  BinaryRequestKind.NonReplicated,
   60_000,
   Buffer.from("opaque mutation"),
 );
 ```
 
-VSR is a runtime protocol choice in Node.js, not a build feature. Custom
-non-replicated codes use `Operation::NonReplicated`; custom replicated codes
-are rejected until a server-side replicated extension registry exists.
+VSR is a runtime protocol choice in Node.js, not a build feature. Codes absent
+from the SDK command table use `Operation::NonReplicated` and carry the command
+code in the request header's reserved field. The server remains authoritative
+for classifying or rejecting extension commands.
 
 The same npm package supports both framing modes. VSR currently supports TCP
 only and restricts `Client` to one pooled connection because authentication,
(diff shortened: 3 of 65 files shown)


foreign/node/src/wire/message/poll-messages.command.ts
```diff
@@ -67,7 +85,79 @@ export const POLL_MESSAGES = {
...
+    SYNC_GROUP.code,
+    SYNC_GROUP.serialize({
+      streamId: request.streamId,
+      topicId: request.topicId,
+      groupId: request.consumer.id,
+    }),
+  );
+  const assignment = SYNC_GROUP.deserialize(response);
+  if (assignment === null)
+    throw responseError(SYNC_GROUP.code, 5006);
```
numinnex: Error code 5006 is the right choice here: it matches the server's `ConsumerGroupMemberNotFound` (`core/common/src/error/iggy_error.rs`), and `core/common/src/traits/binary_impls/messages.rs` confirms a non-member must surface 5006.

However, the SDK's translation table in `src/wire/error.code.ts` still maps 5006 to "Invalid file size", so anyone hitting this path gets a misleading error message. Since this PR makes 5006 reachable from the SDK, it would be worth relabeling it to "Consumer group member not found" in the same change.

numinnex (dismissed):

foreign/node/src/client/client.socket.ts
```diff
@@ -167,25 +238,177 @@ export class CommandResponseStream extends EventEmitter {
...
+        this.vsrSession.reset();
+        this.emit('sessionReset');
+      }
+      if (error instanceof ResponseError)
+        throw responseError(command, error.errorCode);
+      throw error;
+    }
+  }
+
+  private _exchange(write: () => void, timeout?: number): Promise<Buffer> {
```
hubcio: `_exchange` is first-frame-wins - `once('response')` with no request/reply correlation, so whatever frame shows up next resolves whatever exchange is currently open.

the server's `evict_stale_client` (`core/server-ng/src/dispatch.rs:1588`) sends an unsolicited `Eviction` frame and never closes the socket. if one lands while a command is in flight it eats that exchange, and the genuine reply lands on the *next* one. from then on every command resolves the previous command's body as a success. silently, no error, forever.

reproduced against a real server: command B resolved with the reply for command A, and it chained through five commands.

it needs operator-enabled `[heartbeat]` (ships `enabled = false` in `core/server-ng/config.toml:359`) plus group membership plus a client without `heartbeatInterval` - which is the client default, and the README doesn't mention it. that's why the suite never gets near it.

cheapest fix is routing `Eviction` out of band in `_onData` at connection scope instead of through the exchange slot. that also gives idle eviction the event it currently lacks. `abort()` in the eviction branch works too. real request/operation correlation only helps replicated ops - `.client` is caller-dependent and NonReplicated/partition ops share `currentRequestId()` - so correlation alone can't cover reads.

foreign/node/src/wire/vsr/namespace.ts
```diff
@@ -0,0 +1,214 @@
...
+  const topic = peekIdentifier(metadata, offset);
+  offset += topic.length;
+
+  if (metadata.length < offset + 2)
+    throw responseError(COMMAND_CODE.SendMessages, INVALID_COMMAND);
+  const partitioningKind = metadata.readUInt8(offset);
+  const partitioningLength = metadata.readUInt8(offset + 1);
+  const PARTITIONING_PARTITION_ID = 2;
+  if (partitioningKind !== PARTITIONING_PARTITION_ID)
+    throw responseError(COMMAND_CODE.SendMessages, FEATURE_UNAVAILABLE);
```
hubcio, line 169: default typed produce doesn't work under vsr. `message.send({streamId, topicId, messages})` leaves `partition` unset (optional in `send-messages.command.ts:37`), `partitioning.utils.ts:134-138` fills in Balanced, and this line rejects it with code 5 `Feature is unavailable` before anything reaches the wire. `MessageKey` hits the same wall.

the restriction itself is not node-specific - `core/sdk/src/vsr.rs:420` rejects everything but `PartitionId` too, and `IggyProducer` defaults to `Partitioning::balanced()` (`core/sdk/src/clients/producer.rs:467`), so rust fails the same way unless you wire a custom partitioner. what's node-specific is that there's no way out: no partition-count API to build a resolver on, no producer layer, and `message.send` is the only send API there is. the PR's own e2e quietly swaps to `PartitionId` in exactly the suites that would have caught this, and the README says nothing.

`offset.store` on a group with `partitionId: null` has the same shape - throws code 6 even though the JSDoc at `offset.utils.ts:93` promises null means "the group's assigned partition". `pollConsumerGroup` resolves the partition internally and never surfaces it, so a vsr group app can poll but cannot commit the documented way.

minimum: document the constraint and fix that JSDoc. better: surface the cursor's partitionId so commits have something to pass.

foreign/node/src/wire/message/poll-messages.command.ts
```diff
@@ -67,7 +85,79 @@ export const POLL_MESSAGES = {
...
+
+  let cursors = groupCursors.get(client);
+  if (!cursors) {
+    cursors = new Map();
+    groupCursors.set(client, cursors);
+    client.once('sessionReset', () => groupCursors.delete(client));
```
hubcio: this is the only `sessionReset` reaction anywhere in the SDK, and all it does is drop the cursor.

any disconnect fires it (`client.socket.ts:129-137`), re-register mints a fresh clientId, and the server keys group membership by that id (`core/server-ng/src/dispatch.rs:1032` - "`vsr_client_id` keys the consumer-group offset fence (the member id)"). so `SyncGroup` comes back empty forever, then 5006, and the stream `Readable` errors permanently.

`groupConsumerStream` joins once at `stream/consumer-stream.ts:119` and never rejoins, so one routine TCP blip kills the feature's main consumer API with no recovery path. reproduced end to end.

rust covers both halves - rejoin on `SignedIn` after reconnect (`core/sdk/src/clients/consumer.rs:612-646`) and a 5006-triggered rejoin (`:810`). node needs the same: rejoin on sessionReset, or off 5006.

foreign/node/src/client/client.frame.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    frames.push(buffer.subarray(offset, offset + declaredSize));
+    offset += declaredSize;
+  }
+
+  return {
+    frames,
+    remainder: offset === buffer.length
+      ? Buffer.alloc(0)
+      : Buffer.from(buffer.subarray(offset))
```
hubcio: `Buffer.from(...)` copies the entire pending remainder on every socket read, so a frame arriving in many chunks gets recopied each time. `buffer.subarray(offset)` is the same value with no copy - safe here since the source buffer isn't reused after extraction. measured roughly X.Xx slower than master on a large blocked read.

the quadratic `Buffer.concat` accumulator behind it (`client.connection.ts:318-329`) is pre-existing and master is strictly worse (uncapped), so that can be a follow-up - a chunk list kills it outright. worth noting the stall also feeds the eviction desync above: a merely busy client can cross the stale window.

foreign/node/src/client/client.config.ts
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import type { ClientConfig, Protocol } from './client.type.js';
+
+export const DEFAULT_MAX_RESPONSE_FRAME_SIZE = 64 * 1024 * 1024;
```
hubcio, line 20: this cap plus destroy-on-oversize now governs the classic path too - master had no cap and never destroyed the socket. the cap is right (uncapped is an OOM vector), but it's a behavior change for existing classic users shipped under a patch bump with no BREAKING note, and the README explains it under `### VSR framing` where a classic user won't look. move that paragraph and call it out in the PR body.

foreign/node/src/wire/message/poll-messages.command.ts
```diff
@@ -67,7 +85,79 @@ export const POLL_MESSAGES = {
...
+  cursors.set(key, cursor);
+  return cursor;
+};
+
+const pollConsumerGroup = async (
+  client: RawClient,
+  request: PollMessages,
+): Promise<PollMessagesResponse> => {
+  for (let attempt = 0; attempt < GROUP_POLL_MAX_ATTEMPTS; attempt += 1) {
+    const cursor = await syncAssignment(client, request);
```
hubcio: `syncAssignment` runs a SyncGroup RPC before every single group poll - 2 round trips per poll, where rust only syncs on cache miss, fence, or heartbeat refresh.

the catch is the unconditional sync is load-bearing right now, since node has no other refresh trigger. dropping straight to a cache would strand widened assignments after a rebalance. so it needs the bundle: assignment cache, a refresh trigger, exposing the cursor's partitionId (same plumbing the offset-commit fix wants), and a kind-tagged `groupKey`.

foreign/node/src/client/client.socket.ts
```diff
@@ -167,25 +238,177 @@ export class CommandResponseStream extends EventEmitter {
...
+        this.emit('sessionReset');
+      }
+      return parsed;
+    } catch (error) {
+      if (error instanceof VsrEvictionError) {
+        this.isAuthenticated = false;
+        this.userId = undefined;
+        this.vsrSession.reset();
+        this.emit('sessionReset');
+      } else if (!(error instanceof ResponseError)) {
```
hubcio: this branch decides to reset the session by sniffing the error type instead of tracking whether anything was actually written. a transient-deadline exit throws a plain `Error`, not `ResponseError(57)`, so a 30s deadline on a slow command tears down the session and mints a fresh clientId - which kills group membership. rust keeps the typed 57 and the session (`core/sdk/src/tcp/tcp_client.rs`, its reconnect list omits 57). an encode `RangeError` from `wire/vsr/index.ts:67-68` also resets, even though it fires before any request id is spent.

the comment is also wrong. `core/server-ng/src/client_table.rs:185-187` says it outright: "Watermark, not contiguity... There is no RequestGap". same stale rationale is repeated at `wire/vsr/session.ts:26-32`. the wording exists elsewhere in the repo so it wasn't invented here, but it shouldn't get propagated into a new module.

fix: a `written` flag in `_exchange`, a typed deadline error, and rewrite both comments.

foreign/node/src/wire/message/poll-messages.command.ts
```diff
@@ -67,7 +85,79 @@ export const POLL_MESSAGES = {
...
+    if (cursor.partitions.length === 0)
+      return { partitionId: 0, currentOffset: 0n, count: 0, messages: [] };
```
hubcio: returning `{partitionId: 0, count: 0}` for "no partitions assigned" is a fine wire shape - rust's `PolledMessages::empty()` looks the same. the problem is the stream layer can't tell it apart from "drained": `stream/consumer-stream.ts:42-55` treats count 0 as end of data, so a member legitimately holding zero partitions mid-rebalance ends an `endOnLastOffset` stream on its very first poll. partitionId 0 being a real partition id makes it worse. needs a way to distinguish no-assignment from drained - a sentinel, or the cursor's real partition.

foreign/node/scripts/check-vsr-protocol.mjs
```diff
@@ -0,0 +1,301 @@
...
+  nodeOperation,
+  /^\s+([A-Za-z0-9]+):\s*([0-9]+),?$/gm
+);
+assert.deepEqual(
+  nodeOperations,
+  rustOperations,
+  'Node Operation differs from the Rust consensus enum'
+);
+
+const rustReplicated = new Set(
```
hubcio: two holes, both confirmed by mutating the rust sources and watching the gate stay green.

first, the gate never reads `core/binary_protocol/src/namespace.rs`. rust derives the shifts from `MAX_PARTITIONS` via `bits_required`, node hardcodes `STREAM_SHIFT = 32n` / `TOPIC_SHIFT = 20n` (`wire/vsr/namespace.ts:39-40`). bump `MAX_PARTITIONS` and every partition write goes to the wrong shard while the gate reports fine. recompute the shifts from namespace.rs instead of hardcoding them.

second, this replicated check compares sets of operation *names*, not the code to operation pairing. swap CreateStream and DeleteStream and it passes. compare numeric `{codeValue -> operation}` maps instead - const names don't map 1:1 to the node keys, so a name-based comparison structurally can't catch mis-pairing.

foreign/node/scripts/check-vsr-protocol.mjs
```diff
@@ -0,0 +1,301 @@
...
+assert.ok(protocolVersion, 'binary protocol crate version is missing');
+const packedVersion =
+  Number(protocolVersion[1]) << 20 |
+  Number(protocolVersion[2]) << 10 |
+  Number(protocolVersion[3]);
+const nodePackedVersion = nodeRegister.match(
+  /export const IGGY_PROTOCOL_VERSION =\s*\(([0-9]+) << 20\) \| \(([0-9]+) << 10\) \| ([0-9]+);/
+);
+assert.ok(nodePackedVersion, 'Node packed protocol version source changed');
+assert.equal(
```
hubcio: exact patch equality is stricter than the actual contract. rust masks the patch out of the minimum (`core/binary_protocol/src/version.rs:117`) and ignores it for compatibility (`:129-132`), so a patch bump on `iggy_binary_protocol` never changes the wire but does red-build foreign/node. compare major.minor only.

keep minor strict though - on 0.x a minor bump can legitimately break the wire, which `version.rs:113-116` calls out.

foreign/node/README.md
```diff
@@ -30,6 +30,58 @@ npm i --save apache-iggy
...
+import { SimpleClient, getRawClient } from "apache-iggy";
+
+const config = {
+  protocol: "vsr" as const,
+  transport: "TCP" as const,
+  options: { host: "127.0.0.1", port: 8090 },
+  credentials: { username: "iggy", password: "iggy" },
+};
+const client = new SimpleClient(getRawClient(config));
+const response = await client.sendBinaryRequest(
```
hubcio: this headline example always throws. `core/server-ng/src/responses.rs:585-589` fail-closes unknown codes, and the PR's own e2e asserts exactly that rejection. leading the vsr section with a snippet that can't run is rough - use a typed example here and keep the raw one below as an explicit "this gets rejected" demo.

foreign/node/src/wire/error.utils.ts
```diff
@@ -19,10 +19,24 @@
 import { translateCommandCode } from './command.code.js';
 import { translateErrorCode } from './error.code.js';
 
-export const responseError = (cmdCode: number, errCode: number) => new Error(
-  `command: { code: ${cmdCode}, name: ${translateCommandCode(cmdCode)} } ` +
-  `error: {code: ${errCode}, message: ${translateErrorCode(errCode)} }`
-);
+export class ResponseError extends Error {
```
hubcio, line 22: `ResponseError`, `DeserializeError`, `VsrEvictionError` and `ProtocolFrameError` aren't reachable from the package root - `src/index.ts` re-exports `client/index.js` and `stream/index.js`, and neither pulls in the error classes. the whole vsr error story is "catch it and match on it", so users want `instanceof`.

not fatal since there's no exports map (deep dist imports work) and `err.name` / `err.errorCode` duck-typing works, but re-exporting them is a one-liner.

foreign/node/src/wire/error.code.ts
```diff
@@ -189,7 +189,7 @@ export const translateErrorCode = (code: number): string => {
     case '5003': return "Failed to close file";
     case '5004': return "Failed to delete file";
     case '5005': return "Cannot read file";
-    case '5006': return "Invalid file size";
+    case '5006': return "Consumer group member not found";
     case '5007': return "Cannot create file";
     case '5008': return "Cannot rename file";
     case '5009': return "Cannot get file info";
```
hubcio: the whole 5000 block is wrong. rust's 5000-5009 range is consumer group errors, not file/storage - 5001 doesn't exist in rust at all, and 5010-5020 are invented here. this PR fixed 5006 but left the rest, so the resync path it adds renders 5009 `ConsumerGroupPartitionNotOwned` as "Cannot get file info". diagnostics only and mostly pre-existing, but it's specifically the fence this PR depends on that misreports.

foreign/node/src/wire/command-set.ts
```diff
@@ -165,6 +166,7 @@ const groupAPI = (c: ClientProvider) => ({
   create: createGroup(c),
   join: joinGroup(c),
   leave: leaveGroup(c),
+  sync: syncGroup(c),
```
hubcio: `group.sync()` is new public API that always fails on classic - `core/server/src/binary/dispatch.rs` has arms for get/create/delete/join/leave consumer group but none for 606, so it falls through to `_ => InvalidCommand` at `:451-453`. either gate it on protocol or keep it internal.

foreign/node/src/client/client.connection.ts
```diff
@@ -204,8 +243,45 @@ export class IggyConnection extends EventEmitter {
...
+    this.socket.removeAllListeners();
+    this.socket.destroy();
+    this.connected = false;
+    this.connecting = false;
+    this.connectPromise = undefined;
+    this._endResponseWait();
+    // The old socket's close handler was just detached, so surface the drop
+    // to any in-flight exchange and queued work ourselves.
+    this.emit('disconnected', false);
+    this.config.options = { ...this.config.options, host, port };
```
hubcio: config is committed before `await connect()`, and this is the only production write to `config.options`. if the redirect connect fails, the configured seed endpoint is gone for good - `reconnect()` then re-dials the dead ex-leader 12 times, and `isConnectedTo`'s config fallback reports the wrong thing (the test at `client.connection.test.ts:85-95` canonizes that behavior). commit the config only after connect resolves, and reset `reconnectCount` on success.

foreign/node/src/client/client.socket.ts
```diff
@@ -104,54 +143,86 @@ export class CommandResponseStream extends EventEmitter {
...
-      await this.connection.connect()
+      if (!this.connection.connected)
+        await this.connection.connect()
 
-    if (!this.isAuthenticated && !UNLOGGED_COMMAND_CODE.includes(command))
-      await this.authenticate(this.options.credentials);
+      if (this.options.protocol === 'vsr' &&
+          isLoginCommand(command) &&
+          this.isAuthenticated)
+        await this.sendCommand(
```
hubcio: this LOGOUT is unshifted to the front of the queue while the LOGIN that follows goes through `wrapCommand` and lands at the back, so the whole pre-existing queue runs between them. under an already-reset session every one of those comes back `Unauthenticated`. keep the pair adjacent.

foreign/node/src/client/client.ts
```diff
@@ -43,23 +44,25 @@ const createPoolFactory = (config: ClientConfig) => ({
...
-    const c = await pool.acquire();
+const createPooledClientProvider = (config: ClientConfig) => {
+  const minPoolSize = config.poolSize?.min || 1;
+  const maxPoolSize = config.poolSize?.max || 4;
+  const pool = createPool(createPoolFactory(config), {
+    min: minPoolSize,
+    max: maxPoolSize
+  });
+  const clientProvider = async () => {
+    const client = await pool.acquire();
```
hubcio, line 57: if anything throws between `pool.acquire()` and the `finishQueue` release, the connection never comes back - there's no try/catch, and `wrapCommand` serializes *after* acquire, so a serializer throw is enough. on classic that's a slow leak. under vsr the pool is pinned to max 1, so the first throw deadlocks the Client permanently. reachable through `serializeGetOffset`'s single/null guard. either release on throw, or serialize before acquiring.

foreign/node/src/wire/message/poll-messages.command.ts
```diff
@@ -67,7 +85,79 @@ export const POLL_MESSAGES = {
...
+
+    const partitionId = cursor.partitions[cursor.position];
+    cursor.position = (cursor.position + 1) % cursor.partitions.length;
+    const response = await client.sendCommand(
```
hubcio: the group poll spans several `sendCommand` calls but `finishQueue` fires between them - `pendingSubmissions` only covers nested submissions, and it fires twice for two sequential commands. so the pool releases the client mid-operation. bounded blast radius, but real under the vsr pool of 1: a session-reset command can land in the middle of the sequence, and the cursor is shared.

.github/actions/node-npm/pre-merge/action.yml
```diff
@@ -115,6 +113,47 @@ runs:
         IGGY_SERVER_TCP_PORT: 8090
       shell: bash
 
+    - name: Start Iggy VSR server
+      id: iggy-vsr
+      if: inputs.task == 'e2e-vsr'
+      uses: ./.github/actions/utils/server-start
```
hubcio, line 119: the vsr server starts with no cluster config - the plain e2e step sets `replica-id` and `IGGY_CLUSTER_ENABLED`, this one sets neither. so `_ensureVsrLeader`, `redirect()`, `isConnectedTo` and the no-leader throw path get zero real-server coverage, which is exactly the code flagged elsewhere in this review. worth either a cluster variant or enabling it here with a 2-node expectation.

the build is also missing `--features vsr`. inert today since the feature gates nothing in server-ng, but it's a latent false-green plus a rust-cache flavor mismatch waiting to happen - `_test_bdd.yml:54-60` does it right.

minor: `IGGY_SERVER_HOST` / `IGGY_SERVER_TCP_PORT` on the step above are dead, only `IGGY_TCP_ADDRESS` is read. copied from the existing step, not introduced here.

foreign/node/src/wire/vsr/reply.ts
```diff
@@ -0,0 +1,165 @@
...
+    (operation === Operation.Register && body.length > 0);
+  if (!resultFramed) return body;
+
+  if (body.length < RESULT_COUNT_LEN)
+    throw responseError(0, INVALID_COMMAND);
+  const count = body.readUInt32LE(0);
+  const sectionLength = RESULT_COUNT_LEN + count * RESULT_ENTRY_LEN;
+  if (body.length < sectionLength)
+    throw responseError(0, INVALID_COMMAND);
+  if (count === 0)
```
hubcio: this throws on any count >= 1, but rust treats a first result of 0 as success and strips the full `4 + count*8` section - see `result_code` in `core/binary_protocol/src/consensus/reply_result.rs:55-65` and `split_metadata_result` in `core/sdk/src/vsr.rs:293-301`. node turns that same body into a throw. (the count == 0 path does agree, since the section is exactly 4 bytes there.)

unreachable today - the server only emits count <= 1, rejection only - and it fails closed, so this is forward-compat rather than a live bug. still worth a `code === 0` success arm. same for threading the real command code: right now the failure carries `errorCode` 0, which `translateErrorCode` renders as the literal string "error".

foreign/node/src/client/client.connection.ts
```diff
@@ -125,54 +129,86 @@ export class IggyConnection extends EventEmitter {
...
     });
+    return socket;
+  }
+
+  /**
+   * Establishes the connection to the server.
+   *
+   * @returns Promise that resolves when connected
+   */
+  connect(): Promise<this> {
```
hubcio, line 209: `connect()` never settles if the socket is already open. the constructor opens the socket at `:140`, but the 'connect' listener is only registered in here - if the event already fired, nothing resolves the promise, and `connectPromise` caches the hang for every later caller. a pooled Client wedged this way never resolves `destroy()` or `drain()` either.

master has the same shape so this isn't introduced here and shouldn't block the PR, but `connect()` was rewritten wholesale and the fix is one guard: resolve immediately when `!socket.connecting && readyState === 'open'`, or register 'connect' in `_installSocket` alongside the other lifecycle listeners. CI passing is iteration luck - a microtask boundary is safe, a `setTimeout(0)` or an I/O await hangs, and both were observed in a single run.

codecov.yml
```diff
@@ -62,7 +62,13 @@ flag_management:
     - name: php
       paths:
         - foreign/php/
-    - name: node
+    - name: node-test
```
hubcio: renaming the `node` flag to `node-test` orphans the old one - with `carryforward: true` it sticks around in the codecov UI reporting stale coverage until someone deletes it by hand. worth a note in the PR body.

foreign/node/src/client/client.socket.ts
```diff
@@ -93,8 +123,17 @@ export class CommandResponseStream extends EventEmitter {
    */
   _init() {
     this.heartbeat(this.options.heartbeatInterval);
-    this.connection.on('disconnected', async () => {
+    this.connection.on('error', (error: Error) => {
+      this._failQueue(error);
+    });
+    this.connection.on('disconnected', () => {
```
hubcio, line 147: the same four statements - clear auth, clear userId, reset the vsr session, emit sessionReset - appear here and twice more at `:309-314` and `:316-331`. one `_resetSession()` saves about 14 lines. one caveat: if the eviction branch grows an `abort()` per the correlation fix, keep that one split out.

foreign/node/src/client/client.frame.ts
```diff
@@ -0,0 +1,77 @@
...
+    const available = buffer.length - offset;
+    const declaredSize = protocol === 'vsr'
+      ? readVsrSize(buffer.subarray(offset, offset + headerSize))
+      : CLASSIC_HEADER_SIZE + buffer.readUInt32LE(offset + 4);
+
+    if (declaredSize < headerSize)
+      throw new ProtocolFrameError(
+        `declared ${protocol} frame size ${declaredSize} is below header size`
+      );
+    if (!Number.isSafeInteger(declaredSize) ||
```
hubcio: `Number.isSafeInteger(declaredSize)` can't fail - `declaredSize` comes from `readUInt32LE`, so it's bounded by 2^32. the adjacent bound check does the real work. same dead guard at `wire/vsr/index.ts:67` and `wire/consumer-group/sync-group.command.ts:50`.

keep the `partitionsLength > data.length - 12` half of the sync-group one, two tests assert its message.

foreign/node/src/wire/vsr/index.ts
```diff
@@ -0,0 +1,151 @@
...
+      if (this.state.session === null)
+        throw responseError(command, UNAUTHENTICATED);
+      request = isPartition(operation)
+        ? this.state.currentRequestId()
+        : this.state.nextRequestId();
+      session = this.state.session;
+    }
+
+    const header = encodeRequestHeader({
+      size: HEADER_SIZE + payload.length,
```
hubcio: `size` was already computed and validated at `:66` - pass it instead of recomputing `HEADER_SIZE + payload.length`. and the conditional spread at `:95-97` can be `nonReplicatedCode: operation === Operation.NonReplicated ? command : undefined`, which compiles clean without `exactOptionalPropertyTypes` and drops an intermediate object per request.

foreign/node/src/client/client.connection.ts
```diff
@@ -239,47 +315,28 @@ export class IggyConnection extends EventEmitter {
...
+      : data;
+
+    try {
+      const extracted = extractResponseFrames(
+        this.config.protocol ?? 'classic',
+        buffered,
+        this.config.maxResponseFrameSize ?? DEFAULT_MAX_RESPONSE_FRAME_SIZE
+      );
+      this.readBuffers = extracted.remainder;
+      this.waitingResponseEnd = extracted.remainder.length > 0;
```
hubcio: `waitingResponseEnd` is a pure derivative of `readBuffers.length > 0` - set from `extracted.remainder.length > 0`, and only ever read alongside the same condition. drop the field. moot anyway if the accumulator becomes a chunk list.

foreign/node/scripts/check-vsr-protocol.mjs
```diff
@@ -0,0 +1,301 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import assert from 'node:assert/strict';
+import { access, readFile } from 'node:fs/promises';
+import { resolve } from 'node:path';
+import { pathToFileURL } from 'node:url';
+
+const rootCandidates = [
```
hubcio: the first candidate resolves to `foreign/`, which can never hold `core/binary_protocol/src/codes.rs`. the comment describes an isolated test layout that doesn't exist in this tree. drop it and keep the monorepo root.

foreign/node/src/wire/consumer-group/group.utils.ts
```diff
@@ -18,6 +18,13 @@
 
 import { serializeIdentifier, type Id } from '../identifier.utils.js';
 
+/** Stream, topic, and consumer-group identifiers. */
+export type TargetGroup = {
```
hubcio, line 22: `TargetGroup` landed next to the four identical inline shapes instead of replacing them - `join-group`, `leave-group`, `delete-group` and `get-group` all spell out the same three fields. collapsing saves about 16 lines. keep the old names as exported aliases (`export type JoinGroup = TargetGroup`) so it isn't a breaking change.

hubcio (changes requested):

foreign/node/src/client/client.connection.ts
```diff
@@ -125,53 +130,124 @@ export class IggyConnection extends EventEmitter {
...
+        return;
+      debug('socket/close event', hadError);
       this.connected = false;
+      this.connecting = false;
+      this.connectPromise = undefined;
+      this._endResponseWait();
       this.emit('disconnected', hadError);
-      this.reconnect();
+      if (!this.ending)
+        void this.reconnect();
```
hubcio: `void this.reconnect()` drops a promise that can reject. on re-entry `reconnect()` returns `this.reconnectPromise` directly (line 264), and since it is an `async` method the promise it hands back adopts the shared one, so it rejects too. only the first caller wraps the await in try/catch (lines 282-292), so every re-entry through this close handler leaves an unhandled rejection behind.

re-entry happens on each retry: `_reconnectUntilConnected` installs a new socket, the dial fails, that socket's close handler lands here, `reconnectPromise` is already set, and the adopted promise gets voided. with the defaults at lines 76-80 (enabled, 5000 ms, 12 retries) an outage that outlives the whole retry budget leaves 11 unhandled rejections, and node terminates the process on the first one. the unit tests miss it because the only dead-endpoint case (`client.socket.test.ts:380`) uses `maxRetries: 1`, which never re-enters.

one-line fix: `void this.reconnect().catch(() => undefined);`

foreign/node/src/client/client.connection.ts
```diff
@@ -191,37 +272,116 @@ export class IggyConnection extends EventEmitter {
...
+    this.connecting = false;
+    this.connectPromise = undefined;
+    this.reconnectPromise = undefined;
+    this._endResponseWait();
+    // The old socket's close handler was just detached, so surface the drop
+    // to any in-flight exchange and queued work ourselves.
+    this.emit('disconnected', false);
+    this.socket = this._installSocket(getTransport(redirectedConfig));
+    await this.connect();
+    this.config.options = redirectedOptions;
```
hubcio, line 377: committing `config.options` after a successful redirect throws away the seed endpoint for good. once the client has followed the leader here, `this.config` only knows that leader, so when it dies `_reconnectUntilConnected` builds every socket from `getTransport(this.config)` and dials the dead ex-leader until maxRetries, then gives up - while the rest of the quorum is sitting there healthy. that is the failover path vsr exists for.

keep the configured endpoint as a seed and fall back to it in the reconnect loop (or hold a list and rotate) instead of overwriting the only endpoint the client knows. moving the write after `connect()` fixed the failed-redirect half of this, but the successful-redirect half is the one that costs you failover.

foreign/node/src/client/client.connection.ts
```diff
@@ -191,37 +272,116 @@ export class IggyConnection extends EventEmitter {
...
+    );
+  }
+
+  async redirect(host: string, port: number) {
+    const redirectedOptions = { ...this.config.options, host, port };
+    const redirectedConfig = {
+      ...this.config,
+      options: redirectedOptions
+    };
+    this.socket.removeAllListeners();
```
hubcio: `removeAllListeners()` also strips the listeners `_waitForConnection` registered on this same socket, so a dial in flight never settles - and there is no connect timeout anywhere, so that promise and everything awaiting it hangs for good.

it also hides a second problem: the abandoned `_reconnectUntilConnected` loop is not fenced to the socket it installed. today the orphaned await wedges that loop so it cannot clobber the socket the redirect installs, but drop this line alone and the clobber goes live. fix both together - remove the blanket listener wipe (the `this.socket !== socket` guards already keep the old socket inert, and the manual `disconnected` emit below stays), and have the loop bail out when `this.socket` is no longer the socket it created.

hubcio (changes requested):

foreign/node/src/client/client.connection.ts
hubcio: correcting one clause in my earlier comment on this line - there are two reconnect tests with a dead endpoint, and the one i named as the only case is not the relevant one.

`client.connection.test.ts:242` (`maxRetries: 1`) points the *redirect target* at a closed port, so the retry dial goes back to the live seed and succeeds - no retry ever fails there. `client.socket.test.ts:380` (`maxRetries: 1`, server closed before the client connects) is the only test where a retry dial actually fails, and with one retry it never re-enters. so the real gap is that no test ever has two consecutive failed attempts, which is the minimum this needs.

the rest of that comment stands: the leak count is `maxRetries - 1`, so 11 at the defaults.

foreign/node/src/client/client.connection.ts
hubcio, line 377: two things to add to my earlier comment on this line.

both dial sites read the mutated config - `_reconnectUntilConnected` at line 314 and `connect()` at line 216 - and nothing routes around the dead endpoint: `redirect()` is only reachable from `_ensureVsrLeader`, which needs a working connection to fetch the cluster metadata first, and vsr pins the pool to a single connection (`client.config.ts:54`) while `createPool` gets no `validate` or eviction options (`client.ts:52`), so the broken client is never replaced.

also these retry dials feed the unhandled rejection flagged at line 196, so with the defaults the process can die before the retry budget even runs out.

hubcio (changes requested):

hubcio (approved):

numinnex (approved):

PR #3765, opened 2026-07-28
feat(metadata): state transfer for snapshot and client table

numinnex:


core/shard/src/lib.rs
```diff
@@ -2749,6 +2964,790 @@ where
...
+                target_accepted: false,
+                attempts,
+                idle_ticks: 0,
+            });
+            self.send_request_state_transfer(consensus, peer, nonce)
+                .await;
+            return;
+        };
+
+        consensus.set_state_transfer_stage(consensus::StateTransferStage::Installing);
```
hubcio, line 3971: this `set_state_transfer_stage(Installing)` is the only unguarded stage set in the shard, and `Idle -> Installing` is an illegal transition, so it's an `assert!` panic that takes down shard 0. reachable: the probe-exhausted fallback clears the stage to `Idle` (legal from `Fetching`, so silent) but lives in the consensus crate and can't touch `metadata_transfer`, so the session survives with its nonce intact; none of the receive handlers (`on_state_transfer_target`, `on_state_chunk`, `request_pending_state_chunk`) checks the stage, so the pull completes into this assert. `on_state_transfer_target` also sets `target_accepted = true` unconditionally while moving the stage only from `AwaitingTarget` - same missing invariant from the other side. fix needs both halves: bail out of `on_transfer_progress` when the stage isn't `Fetching`, and drop the session wherever the stage gets cleared.

core/metadata/src/stm/mod.rs
```diff
@@ -434,6 +440,10 @@ macro_rules! collect_handlers {
                                 $crate::stm::StateHandler::apply(payload, self, *ts)
                             },
                         )*
+                        [<$state Command>]::RestoreSnapshot(snapshot) => {
+                            *self = Self::inner_from_snapshot(snapshot.clone());
```
hubcio: `inner_from_snapshot` mints a fresh `Arc::new(StatsRegistry::default())` per call, and this arm runs once per left-right buffer (`absorb_first` and `absorb_second` both dispatch it), so after an install the two buffers hold different registries. boot is only safe because `from_snapshot` builds one inner and `new_from_empty` clones it. worse, `inner_from_snapshot` never repopulates the registry's `partitions` map, so every `Arc<PartitionStats>` the data plane registered at bootstrap/reconcile is orphaned - `partition_get` returns none and partition stats read (0,0,0,0) permanently on any node that state-transferred, while stream/topic aggregates alternate between the two arc sets by publish parity. the comment three lines above the `Arc::new` in `stream.rs` states exactly the invariant this breaks.

fix: pass the existing shared registry into the rebuild (this arm has `self` to source it) and prune it to the snapshot's key set while keeping survivors' partition entries - slab keys get reused, so a stale entry would hand a new stream a dead one's counters. note the only test pinning exact counts is `cfg(not(feature = "vsr"))`, so CI can't see this.

core/metadata/src/impls/metadata.rs
```diff
@@ -1308,6 +1366,248 @@ where
...
+    /// stamps read `commit_min` inside one synchronous region, so they are
+    /// mutually consistent.
+    #[must_use]
+    pub fn state_transfer_offer(&self) -> Option<StateTransferOffer> {
+        let consensus = self.consensus.as_ref()?;
+        if !is_caught_up_primary(consensus) {
+            return None;
+        }
+        let coordinator = self.coordinator.as_ref()?;
+        let snapshot = std::fs::read(coordinator.snapshot_path()).ok()?;
```
hubcio: three problems in this function, one rewrite fixes all of them.

the raw `fs::read` ships the file including its integrity trailer, and neither end ever runs `split_trailer`: install decodes and re-persists the bytes raw, so `write_durably` appends a second trailer (+20 B per transfer generation, tolerated only because rmp-serde ignores trailing bytes - any future strict decode bricks every transferred node at boot). worse, corruption-at-rest that survives the msgpack decode (a flipped bit inside a u64 or a string body) gets re-sealed on the receiver with a fresh valid trailer and a matching pairing: a fault the source node would refuse boot over becomes undetectable on the second node.

both `.ok()?` here also swallow the error, and the shard-side refusal log then says "not a caught-up primary, or no snapshot persisted" - actively wrong when the real cause is an unreadable or corrupt snapshot.bin.

and the whole thing (read + full msgpack decode of every stream/topic/user to get one u64 + full table encode) runs synchronously on shard 0's pump, rebuilt on every `RequestStateTransfer` including stall retries.

suggested shape: `fs::read` with the error logged, `split_trailer` as verification, keep the decode for `sequence_number`, ship the stripped payload (the installed file is then byte-shape-identical to a local checkpoint), and serve repeat requests from the cached offer instead of rebuilding. deriving the seq from `last_checkpoint().0` instead does not work: `write_durably` renames before the parent-dir fsync, so a DirSync failure leaves the new file live with the cell stale, and the offer would under-advertise the frontier.

core/shard/src/lib.rs
```diff
@@ -2749,6 +2964,790 @@ where
...
+                let start = header.offset as usize;
+                let end = start.saturating_add(header.len as usize);
+                // Tail of the final artifact: the receiver now holds everything
+                // the manifest promised, so the offer only has to outlive a
+                // possible re-request of this very chunk.
+                if header.artifact as usize == last_artifact && end >= artifact_bytes.len() {
+                    served.fully_served = true;
+                }
+                artifact_bytes
+                    .get(start..end.min(artifact_bytes.len()))
```
hubcio: `get(start..end.min(artifact_bytes.len()))` returns `Some(&[])` when `start == len`, so this answers a fully-consumed artifact with an empty chunk instead of dropping the request - and the handler already reset `served.idle_ticks` above, so the offer never expires. the receiver accepts the empty payload (offset matches, no overrun), extends nothing, resets its own `idle_ticks`, and immediately re-requests the same offset. both liveness counters pinned at zero, `attempts` never increments: the transfer has no bound at all, the two nodes ping-pong empty frames at network rate, and the rejoining replica withholds `PrepareOk` for the life of the process - a 3-node cluster is permanently down to 2 acking replicas.

reachable via the offer rebuild: a stall-retry `RequestStateTransfer` reuses the session nonce, and `on_request_state_transfer` rebuilds the offer and replaces the cache entry under that same nonce. the client table is encoded live, so if the first descriptor was delayed rather than lost, the receiver holds manifest A while chunks come from offer B. `B.len > A.len` is caught by the artifact checksum; `B.len < A.len` (a client logged out between the two builds) lands the cursor exactly on `B.len` and enters this loop.

fix wants all three: refuse chunk requests with `start >= len`, don't reset either idle counter on a zero-byte payload, and make the cached offer first-wins per `(requester, nonce)` so a rebuild can't replace a manifest the receiver already accepted.

core/shard/src/lib.rs
```diff
@@ -2161,7 +2311,34 @@ where
             && consensus.namespace() == header.namespace
         {
             match consensus.handle_commit(&header) {
-                CommitOutcome::Advanced => planes.0.commit_journal().await,
+                CommitOutcome::Advanced => {
+                    // Mid-transfer the pre-install STM must not walk: the
+                    // snapshot being installed already contains those ops.
+                    // `commit_max` still advanced inside `handle_commit`, so
+                    // the post-install repair targets the right frontier.
+                    if !consensus.is_transferring() {
```
hubcio, line 2404: this gate is right, but the same walk is reachable ungated from `on_start_view` (line 2225) and `on_do_view_change` (line 2128) - the `AwaitingTarget` early-return only covers that one stage, so a StartView arriving during `Fetching` walks the pre-install STM. two consequences: the walk can advance `commit_min` past the incoming `snapshot_seq`, flipping `snapshot_ahead` to false so the install silently degrades to table-only (no STM restore, no persist, no pairing) while still logging success; and if the advance lands inside the install's superblock await, `set_commit_floor`'s anti-rewind assert panics the shard. same one-line gate belongs on both paths.

core/consensus/src/client_table.rs
```diff
@@ -944,6 +952,200 @@ impl ClientTable {
...
+            table.slots[slot_idx] = Some(ClientEntry {
+                epoch,
+                user_id,
+                watermark,
+                watermark_checksum,
+                ring,
+                client_id,
+                latest_commit,
+            });
+            table.index.insert(client_id, slot_idx);
```
hubcio: the wire decode skips three validations `from_snapshot` treats as mandatory, on strictly less trusted input. duplicate `client_id`: this `index.insert` silently overwrites, leaving an occupied-but-unindexed slot - with `count == max_clients` plus one duplicate, `commit_register` later sees `index.len() >= slots.len()` as false (skips eviction), `first_free_slot()` finds nothing, and the `.expect("eviction must free a slot")` panics shard 0. `ring_len`: accepted up to 255 with no `REPLY_RING_CAPACITY` bound - `push_latest` only evicts on `==`, so an over-capacity ring grows by one on every later `commit_reply` for that client, and `encode`'s `ring.len() as u8` wraps at 256, making the table permanently un-transferable. the artifact checksum only proves transit, so a corrupt or buggy peer's table reaches this codepath - exactly the threat model the sibling decoder documents (`DuplicateClientId`, `DuplicateSlot`, `SlotOutOfRange`). worth porting all three bounds.

separately: entries compact into `0..count` here, so the encode doc's "every caught-up replica encodes identical bytes" stops being true after one transfer (new registrations land in different slots). eviction victims are unaffected, so just fix the doc.

core/shard/src/lib.rs
```diff
@@ -2563,17 +2740,55 @@ where
...
+                        *self.metadata_repair.borrow_mut() = None;
+                        consensus.begin_state_transfer_await();
+                        let nonce = iggy_common::random_id::get_uuid();
+                        *self.metadata_transfer.borrow_mut() = Some(MetadataTransferSession {
+                            nonce,
+                            peer: header.replica,
+                            commit_op: 0,
+                            artifacts: Vec::new(),
+                            target_accepted: false,
+                            attempts: 0,
```
hubcio: the retry budget lives on the session, but three of the four writers create a fresh session, so no loop is actually bounded. this re-arm and the StartView arm both mint `attempts: 0`; only the decode-failure path carries it forward. a permanent failure (undecodable table, capacity mismatch) therefore cycles abandon -> repair -> `RangeEvicted` -> re-arm at zero forever, re-pulling the whole snapshot each round - the comment on the decode path claims the opposite. and in the other direction `attempts` is never reset on progress (chunk arrival only resets `idle_ticks`), so 5 scattered transient stalls abandon a transfer that was nearly done.

the three near-identical session literals (here, the StartView arm, the decode retry) are already drifting on exactly this field - one `arm_metadata_transfer(consensus, peer, attempts)` helper would remove the drift and make the budget a real bound (hoisted off the session, reset on progress).

core/shard/src/lib.rs
```diff
@@ -2749,6 +2964,790 @@ where
...
+                    damaged = true;
+                    break;
+                }
+            }
+        }
+
+        let decoded = if damaged {
+            None
+        } else if let (Some(snapshot), Some((table_bytes, table_frontier))) = (snapshot, table) {
+            match consensus::ClientTable::decode(&table_bytes, self.clients_table_max.get()) {
```
hubcio: decoding against the local `clients_table_max` livelocks on any capacity skew: the serving primary's live table can legitimately hold more entries than this node's configured cap - heterogeneous config, or homogeneous config after a cap reduction, since `from_snapshot` sizes capacity to `max(min_slots, highest_slot + 1)`. result is a deterministic `TooManyEntries` every round, feeding the unbounded re-arm loop above, log-only - a permanent silent join failure. decoding against the live table's capacity (needs a `capacity()` accessor) kills every route, and lets the duplicated `IggyShard::clients_table_max` cell + setter + bootstrap call go away - the same config value is already plumbed into `IggyMetadata::set_clients_table_max`.

core/shard/src/lib.rs
```diff
@@ -2749,6 +2964,790 @@ where
...
+            served.map_or(Some(ChunkReply::UnknownOffer), |served| {
+                // Serving a chunk is the only liveness signal the offer gets;
+                // the expiry sweep drops it once these stop arriving.
+                served.idle_ticks = 0;
+                // Manifest-index addressing: an index past the offer is a
+                // requester bug (or a stale frame) and is dropped below.
+                let last_artifact = served.offer.payloads.len().saturating_sub(1);
+                let artifact_bytes = served.offer.payloads.get(header.artifact as usize)?;
+                let start = header.offset as usize;
+                let end = start.saturating_add(header.len as usize);
```
hubcio: `header.len` comes off the wire and only `len == 0` is rejected, so a peer asking for `u32::MAX` gets one frame of the whole remaining artifact. anything above the configured `message_bus.max_message_size` is rejected by the receiving transport, which tears down the entire replica connection - logged at `debug!` only, so at default level the operator just sees a link flap with no reason. the reverse config also breaks: `max_message_size` below `STATE_CHUNK_LEN` + header makes any artifact bigger than the cap untransferable (the const's doc says the bus cap is "far above this", but nothing checks it). clamp the served len to `min(STATE_CHUNK_LEN, max_message_size - HEADER_SIZE)`, derive the requester's chunk size from the same value, and validate the relation at boot.

core/metadata/src/stm/mux.rs
```diff
@@ -188,6 +188,26 @@ where
     }
 }
 
+impl<SnapshotData, Head, Tail> RestoreSnapshotInPlace<SnapshotData> for variadic!(Head, ...Tail)
+where
+    Head: RestoreSnapshotInPlace<SnapshotData>,
+    Tail: RestoreSnapshotInPlace<SnapshotData>,
+{
+    fn restore_snapshot_in_place(&self, snapshot: &SnapshotData) -> Result<(), SnapshotError> {
+        self.0.restore_snapshot_in_place(snapshot)?;
```
hubcio, line 206: the two restores run sequentially with `?` and no rollback: if users restores and streams fails, the halves are left at different states, the snapshot was already persisted and the pairing seeded, and the caller logs and keeps running - the follow-up `commit_journal` then replays whatever contiguous suffix the local WAL holds into the half that did restore. the same malformed input fail-stops the boot path (`restore_snapshot?` aborts recovery). validating all sections are present before mutating anything - or fail-stopping on a partial restore - restores the symmetry.

core/metadata/src/impls/metadata.rs
```diff
@@ -1308,6 +1366,248 @@ where
...
+        // other, which is precisely the torn pairing the superblock exists to
+        // detect. Serializing against `checkpoint_if_needed` makes that
+        // impossible by construction rather than by argument (a transferring
+        // replica withholds acks, so it should not be committing, but "should"
+        // is not an invariant this path can rely on).
+        //
+        // Deadlock-free: nothing between here and the write awaits, and
+        // `write_superblock` takes no lock of its own -- the checkpoint path
+        // calls it under this same gate.
+        let _superblock_gate = if snapshot_ahead && self.superblock.is_some() {
```
hubcio: the "impossible by construction" claim doesn't hold: `checkpoint_if_needed` holds `checkpoint_lock` for its whole sequence and takes `superblock_lock` only around the pairing write, while `persist_snapshot`'s snapshot.bin rewrite runs under `checkpoint_lock` alone - so this gate serializes nothing against the checkpoint's file rewrite. checkpoints run on spawned tasks, so a prepare that passed preflight before the transfer armed can drive one during this install's superblock await; the reachable damage is a `CheckpointChecksumMismatch` boot refusal if the node crashes inside that window (same op, different bytes). taking `checkpoint_lock` then `superblock_lock` here - the checkpoint's own order, so no inversion - makes the comment true.

core/metadata/src/impls/metadata.rs
```diff
@@ -1308,6 +1366,248 @@ where
...
+                let handle = journal.handle();
+                if snapshot_seq > handle.snapshot_op() {
+                    handle.set_snapshot_op(snapshot_seq);
+                }
+            }
+
+            // The snapshot IS ops `..=snapshot_seq` applied: jump the applied
+            // frontier (this is the op-jump the tail repair resumes from) and
+            // let the announced commit point pull the walk target forward.
+            consensus.set_commit_floor(snapshot_seq);
```
hubcio, line 1807: flagging as a follow-up rather than for this PR: `set_commit_floor(snapshot_seq)` hands the commit walk a floor it never verified, and the walk matches entries by op number only - no view or hash-chain check - while `set_snapshot_op` leaves everything above the floor resident. a node with a pre-crash prepared-but-uncommitted suffix that a view change reassigned cluster-side would apply those stale bodies as committed (the table is shielded by the frontier fence, the STM is not). the `TODO(suffix-truncation)` in consensus names the same root cause and the missing durable truncate-from-op primitive; this floor jump widens exposure to it on exactly the nodes guaranteed to have a stale log.

core/shard/src/lib.rs
```diff
@@ -2749,6 +2964,790 @@ where
...
+            // superblock write (view change or checkpoint) records it. Until then a
+            // crash recovers the PREVIOUS checkpoint and this replica transfers again,
+            // which is correct, just wasted work. Continue as a success.
+            Err(metadata::stm::snapshot::SnapshotError::SuperblockNotDurable { op }) => {
+                consensus.set_state_transfer_stage(consensus::StateTransferStage::Idle);
+                tracing::warn!(
+                    shard = self.id,
+                    snapshot_seq = op,
+                    commit_op,
+                    "metadata state transfer installed but its checkpoint pairing is \
```
hubcio: two related papercuts. `SuperblockNotDurable` is an `Err` for an outcome that is a success - the install fully landed (snapshot, table, frontiers, commit point) before it's returned, and this sole caller special-cases it back; any future caller treating err as failure would lose a completed install. and this degraded log shares the prefix "metadata state transfer installed" with the success line while `stdout_occurrences` matches substrings, so 6 of the 7 new specs pass on the degraded path (only `assert_pairing_matches_install` still discriminates). `Result<InstallOutcome, SnapshotError>` plus distinct log strings fixes both.

core/journal/src/lib.rs
```diff
@@ -60,6 +60,17 @@ where
     ) -> impl Future<Output = io::Result<Vec<Self::Entry>>> {
         async { Ok(Vec::new()) }
     }
+
+    /// Snapshot watermark: entries at or below it are evictable. `0` for
+    /// journals without snapshot bookkeeping.
+    fn snapshot_op(&self) -> u64 {
```
hubcio: these defaults are a trap through `JournalState<J>`, which forwards only `header`/`previous_header`/`append`/`entry` - through the wrapper `set_snapshot_op` is a silent no-op and `snapshot_op()` returns 0 (the pre-existing `remaining_capacity` default already makes `should_checkpoint` permanently false there too). latent today because the metadata plane holds `PrepareJournal` directly, but the moment partition-plane transfer lands on the wrapper this breaks silently - or panics, since forwarding `set_snapshot_op` while inheriting the `snapshot_op() -> 0` default turns the install's monotonicity guard into an assert trip. making both methods required (or forwarding everything in `JournalState`) closes it in both directions.

core/shard/src/lib.rs
```diff
@@ -2749,6 +2964,790 @@ where
...
+                // Manifest-index addressing: an index past the offer is a
+                // requester bug (or a stale frame) and is dropped below.
+                let last_artifact = served.offer.payloads.len().saturating_sub(1);
+                let artifact_bytes = served.offer.payloads.get(header.artifact as usize)?;
+                let start = header.offset as usize;
+                let end = start.saturating_add(header.len as usize);
+                // Tail of the final artifact: the receiver now holds everything
+                // the manifest promised, so the offer only has to outlive a
+                // possible re-request of this very chunk.
+                if header.artifact as usize == last_artifact && end >= artifact_bytes.len() {
```
hubcio, line 3671: `fully_served` is set before the bounds check decides whether anything is actually served, so an out-of-bounds or stale frame that ships zero bytes still flips a live offer onto the short expiry. and the short expiry itself is one `repair_retry_ticks` - the same interval as the receiver's stall re-request, plus a network hop - so the re-request-the-last-chunk window this grace exists for is roughly a coin flip, and losing it costs a full re-pull (expired offer -> `UnknownOffer` -> session dropped with everything already downloaded). move the flag inside the map that produces the payload, and give it at least 2 retry intervals.

core/metadata/src/impls/metadata.rs
```diff
@@ -1296,6 +1336,24 @@ where
...
+pub struct StateTransferOffer {
+    /// Serving primary's applied frontier when the offer was built; the
+    /// receiver's tail repair targets past this.
+    pub commit_op: u64,
+    /// Manifest entries, index-aligned with `payloads`. Metadata plane:
+    /// `[METADATA_SNAPSHOT (frontier = sequence_number),
+    ///   CLIENT_TABLE (frontier = commit_min at encode)]`.
+    pub artifacts: Vec<consensus::StateArtifact>,
+    /// Artifact bytes, chunk-served by `(manifest index, offset)`.
+    pub payloads: Vec<Vec<u8>>,
```
hubcio: `artifacts` and `payloads` are index-aligned by comment only, and the two halves are consumed in different files (manifest encoded from one, chunks served from the other). `Vec<(StateArtifact, Vec<u8>)>` makes the desync unrepresentable.

core/sdk/src/quic/quic_client.rs
```diff
@@ -177,6 +177,13 @@ impl BinaryTransport for QuicClient {
         #[cfg(feature = "vsr")]
         if skip_auto_login {
             *self.skip_auto_login_once.lock().await = true;
+            // The replayed login/register must mint a fresh Register: the failed
+            // attempt already consumed the one-shot register request id and may
+            // have half-bound the session.
+            *self
```
hubcio: this block can go: `self.disconnect()` a few lines up already calls `reset_vsr_session()`, and `begin_register` re-mints on exactly the two conditions this comment cites, on every login encode. tcp and ws reconnect paths don't have it and are fine. it's also the one session mutation that runs outside the connection lock - in the path where disconnect early-returns (retry-exhausted `connect` sets `Disconnected` without a reset), this can race an in-flight request's encode and turn it into a spurious `Unauthenticated`. deleting it restores "all session mutation happens under the connection lock".

core/consensus/src/client_table.rs
```diff
@@ -944,6 +952,200 @@ impl ClientTable {
...
+    /// slot order `client(u128) epoch(u64) user_id(u32) watermark(u64)
+    /// watermark_checksum(u128) ring_len(u8) [reply_len(u32) reply_bytes]*`,
+    /// terminated by an `XxHash3_64(8)` over everything before it.
+    ///
+    /// Slot order makes the bytes deterministic across caught-up replicas:
+    /// slots are apply-derived state.
+    #[must_use]
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn encode(&self) -> Vec<u8> {
+        let mut out = Vec::with_capacity(64 * self.index.len() + 16);
```
hubcio: `64 * self.index.len() + 16` under-reserves badly: the fixed fields alone are 53 bytes per entry and each cached reply is a full wire message (>= 256 bytes) plus a 4-byte length, ring up to 5. at the default 8192-client cap a full table encodes to roughly 11 MB grown from a half-MB reservation - several reallocs of a multi-MB buffer on the serving primary's pump, per offer build. summing the actual ring byte lengths first makes it one allocation.

core/shard/src/lib.rs
```diff
@@ -743,6 +745,96 @@ struct MetadataRepairSession {
...
+    /// Ticks with no frame progress; at the configured repair-retry
+    /// threshold the missing piece is re-requested.
+    idle_ticks: u32,
+}
+
+/// A cached serving-side state-transfer offer (shard 0 of the serving
+/// primary). Keyed by requester replica id so a rebooted requester's fresh
+/// nonce replaces the stale offer; chunks must all come from ONE offer or
+/// the artifact checksums cannot hold.
+struct ServedStateTransfer {
```
hubcio, line 870: each cached offer owns a private copy of the snapshot plus the encoded table, keyed per requester - n-1 simultaneous rejoiners pin n-1 multi-MB copies on shard 0 until the sweep. building the offer once per snapshot generation and sharing `Rc<Vec<u8>>` payloads (the shard is single-threaded) makes the marginal requester free; `Rc` alone doesn't help while every request re-builds.

core/consensus/src/state_manifest.rs
```diff
@@ -0,0 +1,341 @@
...
+        });
+    }
+    if !reader.bytes.is_empty() {
+        return Err(StateManifestError::Truncated);
+    }
+    Ok(artifacts)
+}
+
+/// Little-endian cursor over the encoded manifest content.
+struct ManifestReader<'a> {
```
hubcio: `ManifestReader` here and `CodecReader` in client_table.rs are the same LE cursor (take/u8/u32/u64) duplicated within one crate, and the two codecs open-code the `XxHash3_64` new/write/finish triple four times when `state_artifact_checksum` already exists. one shared cursor (common `Truncated` error + `From` into each codec's enum) plus routing the hashing through the existing fn drops ~55 lines.

core/shard/src/lib.rs
```diff
@@ -1440,7 +1561,10 @@ where
                 Input = Message<PrepareHeader>,
                 Output = metadata::stm::result::ApplyReply,
                 Error = iggy_common::IggyError,
-            > + StreamsFrontend,
+            > + StreamsFrontend
+            + metadata::stm::snapshot::RestoreSnapshotInPlace<
```
hubcio, line 1648: this same 3-part bound (`StateMachine<...> + StreamsFrontend + RestoreSnapshotInPlace<MetadataSnapshot>`) is now pasted 11 times across this file and router.rs. a `trait MetadataStm: ... {}` with a blanket impl collapses all of them with no api change.

core/server-ng/src/responses.rs
```diff
@@ -1340,6 +1340,14 @@ pub(crate) fn build_raw_pat_reply(
         return Ok(committed);
     }
     let header_len = std::mem::size_of::<ReplyHeader>();
+    // A `Reply` whose result section is nonzero is not a successful commit but a
+    // committed business rejection or a `TransientNotCommitted` retry frame, both
+    // with no payload and no token to ship. Pass it through so the client decodes
+    // the typed result (and, for a transient, replays) instead of having a raw
+    // token grafted onto a rejection body.
+    if result_code(&committed.as_slice()[header_len..]) != Some(0) {
```
hubcio: this early return is dominated by the existing `result_code` gate a few lines down - same leading bytes, and the pre-existing one is better bounded (`get(header_len..size)` vs this unbounded slice that can read past the declared size into buffer padding). every frame this catches, that one catches. if this was added for an observed bug, the old gate already covered that case, which would mean the actual bug is elsewhere - worth a note on what motivated it. otherwise just drop this one.

core/integration/tests/cluster/metadata_state_transfer.rs
```diff
@@ -0,0 +1,419 @@
...
+
+/// Assert `node`'s durable superblock pairs with the checkpoint its state-transfer
+/// install recorded.
+async fn assert_pairing_matches_install(harness: &TestHarness, node: usize) {
+    const PAIRING_MARKER: &str = "state transfer recorded its checkpoint pairing";
+
+    // The server colors its tracing fields, so `checkpoint_op=193` reaches the log
+    // as `checkpoint_op\x1b[0m\x1b[2m=\x1b[0m193`. Strip escapes before matching,
+    // the same reason `ServerHandle::stdout_occurrences` does.
+    let (stdout, _stderr) = harness.node(node).collect_logs();
```
hubcio: verbatim copy of the harness's private `strip_ansi`. this caller needs the stripped text (it parses `checkpoint_op=` out of it), so `stdout_occurrences` can't serve it - export `strip_ansi` or add a `ServerHandle::stdout_plain()` and delete the copy.

core/consensus/src/plane_helpers.rs
```diff
@@ -40,7 +40,10 @@ pub async fn pipeline_prepare_common<C, F>(
 {
     assert!(!consensus.is_follower(), "on_request: primary only");
     assert!(consensus.is_normal(), "on_request: status must be normal");
-    assert!(!consensus.is_syncing(), "on_request: must not be syncing");
+    assert!(
+        !consensus.is_transferring(),
+        "on_request: must not be syncing"
```
hubcio: assert text still says "must not be syncing" after the rename to `is_transferring` - same stale string at `metadata.rs:2759` and `iggy_partition.rs:1374` (that one's predicate changed one line above).

core/consensus/src/lib.rs
```diff
@@ -149,8 +149,13 @@ where
 
 pub mod client_table;
 pub use client_table::{
-    CachedReply, ClientEntrySnapshot, ClientTable, ClientTableDecodeError, ClientTableSnapshot,
-    CommitReply,
+    CachedReply, ClientEntrySnapshot, ClientTable, ClientTableCodecError, ClientTableDecodeError,
```
hubcio: `ClientTableCodecError` next to the pre-existing `ClientTableDecodeError` - neither name says which format (msgpack checkpoint vs transfer wire), and "codec" oversells it since `encode` is infallible. `ClientTableWireError` or similar would disambiguate.

core/shard/src/lib.rs
```diff
@@ -2035,6 +2179,40 @@ where
...
+            // anyway.
+            //
+            // Outside the superblock gate above: that gate fail-closes the VSR
+            // actions this replica would VOUCH with (notably `PrepareOk`) until
+            // the adopted view is durable. Requesting a transfer vouches for
+            // nothing -- it only pulls state -- and a transferring replica
+            // withholds `PrepareOk` on its own (`is_transferring`). Gating it
+            // would also wedge the one path that repairs a replica whose gap
+            // sits below every peer's floor.
+            if consensus.state_transfer_stage() == consensus::StateTransferStage::AwaitingTarget {
```
hubcio: this arms on stage alone and ignores whether `handle_start_view` accepted anything - it returns an empty action set for wrong-primary / old-view / below-floor, but the arm still mints a fresh nonce, zeroes `attempts`, and fires a request at a possibly-non-primary. it self-heals via `available = 0`, but it resets the retry budget, and a stream of stale StartViews re-mints the nonce each time so the in-flight descriptor gets dropped by the nonce filter. gate on the StartView actually being adopted.

core/integration/src/harness/handle/server.rs
```diff
@@ -978,6 +999,35 @@ impl Restartable for ServerHandle {
...
+    /// node that rejoins the cluster with no on-disk history, as a fresh
+    /// operator-provisioned replacement or a wiped disk does.
+    ///
+    /// Distinct from [`Restartable::restart`], which preserves the WAL and so
+    /// exercises the recovery-with-history path. The empty data directory is
+    /// what forces the state-transfer join: local recovery has nothing, and
+    /// the committed prefix the node missed no longer exists as WAL entries on
+    /// the peers that checkpointed it.
+    pub fn restart_from_clean_slate(&mut self) -> Result<(), TestBinaryError> {
+        let cleanup = self.config.cleanup;
```
hubcio: the cleanup save/restore dance is dead - `ServerConfig::cleanup` is never read for behavior anywhere in the harness (the real gate is `TestContext::cleanup`), and this fn wipes the data dir itself two lines later. copied from the same dead pattern in `restart`.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -1309,6 +1309,277 @@ impl ConsensusHeader for RepairRangeReplyHeader {
...
+        self.command
+    }
+    fn size(&self) -> u32 {
+        self.size
+    }
+
+    fn validate(&self) -> Result<(), ConsensusError> {
+        if self.command != Command2::RequestStateTransfer {
+            return Err(ConsensusError::InvalidCommand {
+                expected: Command2::RequestStateTransfer,
```
hubcio, line 1371: the new header-only frames (`RequestStateTransferHeader`, `RequestStateChunkHeader`) don't size-check in `validate` while `EvictionHeader` does, and `StateTransferTargetHeader` constrains size only on the `available == 0` branch. the generic `Message::try_from` bound makes this safe either way, but adding `== HEADER_SIZE` to the two header-only frames (and a comment that the `available == 1` body is bounded downstream by the manifest caps) keeps the validate surface consistent.

core/integration/tests/data_integrity/mod.rs
```diff
@@ -15,13 +15,10 @@
...
-// Requires state transfer: the 5 MB bench fill is thousands of ops while the
-// partition journal's evicted ring retains only the last 4096, so a restarted
-// replica's rejoin window exceeds what journal repair can serve. The commit
-// floor lets recovered segments stand in for the evicted prefix, but the
-// sub-floor stats/offset seeding this test asserts (exact messages_count /
-// size_bytes across the restart) is state transfer's job.
-#[cfg(not(feature = "vsr"))]
+// Partially vsr-gated inside the module: the bench-fill test requires
+// PARTITION-plane state transfer (sub-floor stats/offset seeding), which is
+// not implemented yet; the metadata-only deletion/restart test runs under
```
hubcio, line 20: dropping the module-level gate puts `should_handle_resource_deletion_and_restart` into the vsr lane for the first time, and CI doesn't run that lane - green here doesn't prove it passes. worth pasting a local `--features vsr` run of it in the PR.

hubcio (changes requested):
two notes with no diff line to hang them on: `recovery.rs:437` still carries `TODO(state-transfer): there is no metadata state transfer yet` - this PR is its trigger condition.

core/shard/src/lib.rs
```diff
@@ -2749,135 +3014,1102 @@ where
...
+            // unbounded empty-frame ping-pong. The serving side refuses to
+            // produce these now; the guard stays because a peer running an
+            // older build still can.
+            if payload.is_empty() {
+                return;
+            }
+            artifact.buf.extend_from_slice(payload);
+            session.idle_ticks = 0;
+        }
+        self.note_metadata_transfer_progress();
```
hubcio, line 3792: this reset fires on every accepted chunk, and a decode failure is only reachable after the last artifact completes - right after a chunk zeroed the counter - so for any manifest with a non-empty artifact `attempts` oscillates 0/1 and never reaches the budget. an undecodable table, an unknown artifact kind, a checksum mismatch and a capacity mismatch all take this path: burn, `1 > 5` is false, re-arm, whole snapshot pulled again, forever. the doc on `metadata_transfer_attempts` and the block comment in the decode-failure branch both claim this cycle is now bounded - neither holds, and that branch's own next sentence concedes no other bound covers it ("frames ARE flowing here, so `idle_ticks` never accumulates and that sweep can never fire"). the stall sweep still bounds silent peers; decode failures need a counter that only a successful install clears, with chunk arrival resetting the stall counter alone.

core/shard/src/lib.rs
```diff
@@ -2749,135 +3014,1102 @@ where
...
+            None
+        } else if let (Some(snapshot), Some((table_bytes, table_frontier))) = (snapshot, table) {
+            // Decode against the LIVE table's capacity, not a separately
+            // plumbed config cell. The serving primary's table can legitimately
+            // hold more entries than this node's configured cap -- heterogeneous
+            // config, or homogeneous config after a cap reduction, since
+            // `from_snapshot` sizes capacity to `max(min_slots, highest_slot +
+            // 1)`. Against the raw config value that is a deterministic
+            // `TooManyEntries` every round: a permanent, log-only join failure.
+            let capacity = planes.0.client_table_capacity();
```
hubcio, line 3914: live capacity closes the skew only for a receiver that recovered a checkpoint. a cold-boot receiver (fresh join, wiped disk) builds its table straight from config - `ClientTable::new(clients_table_max)` in recovery - so `capacity()` is the raw config value there. a serving primary holding more live sessions than that cap - because its own table grew via `from_snapshot`'s `max(min_slots, highest_slot + 1)`, or because it simply runs a larger cap - encodes `count = index.len()` above it, and decode still returns `TooManyEntries` deterministically. so the cap-reduction join failure the comment above describes survives for the cold-boot receiver, and with the reset at line 3715 the retry budget never triggers: a permanent, log-only loop re-pulling the snapshot every round. needs a decode-side change, since `count` is read inside `decode`: size the table to `min(max(capacity, count), CLIENTS_TABLE_SLOT_MAX)` - mirroring from_snapshot's slot ceiling, so a hostile count cannot drive the allocation.

core/metadata/src/impls/metadata.rs
```diff
@@ -1308,6 +1499,337 @@ where
...
+        // below the primary it later replaces). Restoring such a snapshot
+        // would rewind the STM below `commit_min` with no way back: the
+        // commit walk never revisits ops it already counted as applied, so
+        // the rewound-over effects would be lost until the next transfer.
+        // Keep the local STM (it is a superset) and let tail repair cover
+        // `(commit_min, commit_op]`. The client table still installs below:
+        // it comes from the serving primary's LIVE state at `table_frontier
+        // == commit_op`, which is never behind this replica.
+        let local_applied = consensus.commit_min();
+        let snapshot_ahead = snapshot_seq > local_applied;
```
hubcio: `local_applied` and `snapshot_ahead` are read here, before the gates below are awaited, and never re-read - every branch after the awaits acts on possibly stale values, and the block comment concedes commits can run during the wait ("should not be committing" is not an invariant this path can rest on). two live outcomes once `commit_min` moves past `snapshot_seq` during the await. a plain commit leaves the pairing alone, but `set_commit_floor(snapshot_seq)` then trips the "commit floor may not rewind commit_min" assert - plain `assert!`, so release panics too, and only after `write_durably`, `seed_last_checkpoint` and the in-place restore already ran. a checkpoint completing past `snapshot_seq` additionally overwrites the newer snapshot.bin and regresses the pairing before that same panic, with the STM rewound below `commit_min` - exactly the harm the rewind comment a few lines up exists to prevent. the read-then-await shape is inherited, but the old code waited on `superblock_lock`, which a checkpoint holds only around its pairing write; waiting on `checkpoint_lock` widens the window to a full checkpoint, WAL drain included. re-read `commit_min` and `last_checkpoint()` after the gates close so the decision matches the state it guards.

hubcio (changes requested):

spetz (approved):

mmodzelewski (approved):

hubcio (approved):

PR #3766, opened 2026-07-28
fix(sdk): forward unknown command codes instead of rejecting them

hubcio:
An application built on this SDK cannot reach a command its SDK build
has never heard of. operation_for_code ended None => InvalidCommand,
so a code absent from COMMAND_TABLE failed at encode time, before any
byte left the process, and a server that does implement it never got
the chance to answer. COMMAND_TABLE is a protocol registry, not a
per-server capability list, so the server is the authority on whether
a code exists: an unknown code now ships as non-replicated, carrying
the code in RequestHeader.reserved, and the server answers with a
proper error if it does not know it.

The SDK no longer second-guesses that classification either. It used
to refuse a code the registry knows as replicated, but the server
re-derives the class from the same registry and denies a mismatch on
a path that commits nothing, so the client check duplicated a guard
it does not own. It was unreachable besides, since every replicated
entry resolves through from_command_code first, and it was the sole
construction of an UnknownReplicatedCommand variant nothing in the
tree consumed. operation_for_code is now infallible, the raw-command
documentation no longer explains consensus to callers, and error code
14002 is retired rather than reused.

### DIFF
--- a/core/common/src/error/iggy_error.rs
+++ b/core/common/src/error/iggy_error.rs
@@ -522,8 +522,6 @@ pub enum IggyError {
     AlreadyAuthenticated = 14000,
     #[error("VSR session value {0} is invalid (must be non-zero)")]
     InvalidSession(u64) = 14001,
-    #[error("Replicated command with unknown code {0}")]
-    UnknownReplicatedCommand(u32) = 14002,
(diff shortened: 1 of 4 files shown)


spetz:
Looks good, this should actually close #3758 and simplify a bit #3763

spetz (approved):

thepatrykk (approved):

PR #3767, opened 2026-07-28
feat(consensus): persist metadata VSR state in a durable superblock

krishvishal:
A metadata replica kept its consensus numbers only in RAM. After a crash it
  rebuilt at `view = 0` and inferred a view from the last WAL prepare, so it could
  re-enter a superseded view and re-vote, splitting the log. Sessions and the
  at-most-once dedup cache were volatile too, so a returning client hit
  `NoSession` and a retried request re-executed. And once a checkpoint drained the
  WAL prefix, nothing durably identified the snapshot it folded into.

  ## What

  **`journal::superblock`** (new). `SuperblockStore` trait plus `PingPongSuperblock`:
  two files written alternately, each replaced `temp, fsync, rename, dir-fsync`,
  highest valid `sequence` wins. Rename is atomic, so an update can never destroy
  the last good record. `read_latest` returns three outcomes, kept distinct because
  collapsing them into an `Option` is exactly what lets a lost superblock
  masquerade as a fresh deployment:

  | Outcome | Meaning | Response |
  | --- | --- | --- |
  | `Empty` | never written | fresh boot |
  | `Present(payload)` | newest checksum-clean record | recover from it |
  | `Unreadable { version }` | written, now torn / foreign / unknown version | refuse boot |

  **`consensus::VsrState`** (new). The 58-byte little-endian payload: cluster,
  replica id and count, `view`, `log_view`, `commit_max`, and the paired
  `checkpoint_op` / `checkpoint_checksum`. Offsets pinned by test.

  **Durable-before-send gate.** `VsrConsensus` tracks `view_durable` /
  `log_view_durable`, exposing `needs_superblock_persist()` and
  `mark_superblock_durable(view, log_view)`, which takes the values written rather
  than re-reading `self.view` (the in-memory view can advance across the write's
  await while a checkpoint holds the lock). Every metadata dispatch site (`SVC` /
  `DVC` / `StartView` / `RequestStartView`, the tick, the stale-heartbeat reply,
  `send_prepare_ok`) persists first and withholds on failure. A `debug_assertions`
  tripwire in `dispatch_vsr_actions` asserts callers did so. `RequestStartView` is
  exempt: it is a probe asking to *learn* the view.

  **Three-phase checkpoint.** Persist snapshot, record the pairing, then drain the
  WAL. A crash before the pairing write recovers the prior checkpoint with the WAL
  intact; after it, the new one. One `LocalGate` serializes checkpoints against
  view-change persists, since both drive the single superblock, whose `write` picks
  its slot before it awaits, and in-process submits each run on their own task.

  **Client-table durability.** `ClientTableSnapshot` folds into `MetadataSnapshot`
  under `#[serde(default)]`, preserving slot positions so eviction order survives.
  Boot restores it from the snapshot and replays the committed suffix through the
  same `apply_committed_prepare` the live commit walk uses.

  **Verify-then-decode recovery.** Recovery cross-checks the superblock against the
  on-disk snapshot before trusting its contents. New typed refusals: superblock
  ahead of snapshot, checksum mismatch at the same op, undecodable payload,
  unreadable on every slot. A superblock *lagging* a newer atomically-complete
  snapshot is accepted (`commit_max` is only a recovery lower bound). An absent
  superblock stays a fresh boot; refusing there would brick healthy nodes.

  **WAL body integrity.** The primary seals `checksum_body` once and replicates it
  verbatim. The scan truncates a genuinely torn tail but refuses boot on interior
  bit-rot rather than discarding the committed entries behind it.

  **Incarnation nonce.** `RequestStartView` carries a per-boot `u128`; the answering
  `StartView` echoes it. While `Recovering`, a replica only adopts a `StartView`
  provably post-restart.

  ## Wire compatibility

  `StartViewHeader` and `RequestStartViewHeader` gain `incarnation: u128`, carved
  from the tail of their existing `reserved` regions. Placed **last** and
  16-aligned, so `op` / `commit` / `namespace` keep their offsets and the structs
  stay padding-free (static asserts enforce both). A peer that predates the field
  sends zeros, decoding as `incarnation == 0`, inert per the adoption guard. Header
  size unchanged, so a mixed-version rolling upgrade is safe.

  ## Not in scope

  - Partition-plane consensus is not superblock-backed and is exempt from the gate.
  - No metadata state transfer, so a corrupt checkpoint refuses boot rather than
    repairing from a peer. Marked `TODO(state-transfer)` where the fetch belongs.
  - Header `checksum` and its `parent` chain stay unsealed: activating them needs
    the retransmit path to re-seal a re-stamped header.
  - `checksum_body` has no format-version gate, so a WAL written before this change
    refuses boot. Fail-safe, but a hard upgrade break; `TODO(wal-integrity)`.

### DIFF
--- a/core/binary_protocol/src/consensus/header.rs
+++ b/core/binary_protocol/src/consensus/header.rs
@@ -1056,9 +1056,10 @@ pub struct StartViewHeader {
     ///
     /// Carved from the tail of the former `reserved` region and placed LAST so it
     /// lands 16-aligned with no padding WITHOUT moving `op`/`commit`/`namespace`.
-    /// A peer that predates it sends zeros, decoding as `incarnation == 0`, inert
-    /// per the `handle_start_view` guard, so a mixed-version rolling upgrade is
-    /// wire-compatible.
+    /// A peer that predates it sends zeros, decoding as `incarnation == 0`, which
+    /// the `handle_start_view` guard treats as no claim rather than as a foreign
+    /// one, so a mixed-version rolling upgrade is wire-compatible: the pre-upgrade
+    /// peer's `StartView` is judged by the view checks alone, as before the field.
     pub incarnation: u128,
 }
 const _: () = {
--- a/core/binary_protocol/src/consensus/header.rs
+++ b/core/binary_protocol/src/consensus/header.rs
@@ -1048,15 +1048,29 @@ pub struct StartViewHeader {
     /// max(commit) from all DVCs.
     pub commit: u64,
     pub namespace: u64,
-    pub reserved: [u8; 104],
+    pub reserved: [u8; 88],
+    /// Sender's incarnation, echoed from the `RequestStartView` this answers so a
+    /// recovering replica can prove the reply post-dates its restart (see
+    /// `RequestStartViewHeader::incarnation`). `0` on an unsolicited `StartView`
+    /// (a normal view-change completion), which carries no freshness claim.
+    ///
+    /// Carved from the tail of the former `reserved` region and placed LAST so it
+    /// lands 16-aligned with no padding WITHOUT moving `op`/`commit`/`namespace`.
+    /// A peer that predates it sends zeros, decoding as `incarnation == 0`, inert
+    /// per the `handle_start_view` guard, so a mixed-version rolling upgrade is
+    /// wire-compatible.
+    pub incarnation: u128,
 }
 const _: () = {
     assert!(size_of::<StartViewHeader>() == HEADER_SIZE);
+    // op/commit/namespace keep their pre-incarnation offsets.
     assert!(
         offset_of!(StartViewHeader, op)
             == offset_of!(StartViewHeader, reserved_frame) + size_of::<[u8; 66]>()
     );
-    assert!(offset_of!(StartViewHeader, reserved) + size_of::<[u8; 104]>() == HEADER_SIZE);
+    // `incarnation` is last and 16-aligned, so the struct has no padding.
+    assert!(offset_of!(StartViewHeader, incarnation) + size_of::<u128>() == HEADER_SIZE);
+    assert!(offset_of!(StartViewHeader, incarnation) % 16 == 0);
 };
 
 impl ConsensusHeader for StartViewHeader {
@@ -1097,8 +1111,11 @@ impl ConsensusHeader for StartViewHeader {
 /// Recovering replica -> all replicas: resend me the current `StartView`.
 ///
 /// Header-only; only the current view's primary answers, with a targeted
-/// `StartView` (adoption is fenced by the receiver's view monotonicity and
-/// the sender-is-primary check, so no nonce is needed).
+/// `StartView`. Adoption is fenced by the receiver's view monotonicity and the
+/// sender-is-primary check; `incarnation` additionally proves the reply
+/// post-dates this replica's restart, so a `StartView` from a previous
+/// incarnation still in flight cannot be adopted (see the `handle_start_view`
+/// recovering-status guard).
 #[derive(Debug, Clone, Copy, PartialEq, Eq, CheckedBitPattern, NoUninit)]
 #[repr(C)]
 pub struct RequestStartViewHeader {
@@ -1113,15 +1130,26 @@ pub struct RequestStartViewHeader {
     pub reserved_frame: [u8; 66],
 
     pub namespace: u64,
-    pub reserved: [u8; 120],
+    pub reserved: [u8; 104],
+    /// The requester's per-boot incarnation, echoed back in the answering
+    /// `StartView` so a reply from a previous incarnation is detectable.
+    ///
+    /// Carved from the tail of the former `reserved` region and placed LAST so it
+    /// lands 16-aligned with no padding WITHOUT moving `namespace`. A peer that
+    /// predates it sends zeros, decoding as `incarnation == 0`, so a mixed-version
+    /// rolling upgrade is wire-compatible.
+    pub incarnation: u128,
 }
 const _: () = {
     assert!(size_of::<RequestStartViewHeader>() == HEADER_SIZE);
+    // namespace keeps its pre-incarnation offset.
     assert!(
         offset_of!(RequestStartViewHeader, namespace)
             == offset_of!(RequestStartViewHeader, reserved_frame) + size_of::<[u8; 66]>()
     );
-    assert!(offset_of!(RequestStartViewHeader, reserved) + size_of::<[u8; 120]>() == HEADER_SIZE);
+    // `incarnation` is last and 16-aligned, so the struct has no padding.
+    assert!(offset_of!(RequestStartViewHeader, incarnation) + size_of::<u128>() == HEADER_SIZE);
+    assert!(offset_of!(RequestStartViewHeader, incarnation) % 16 == 0);
 };
 
 impl ConsensusHeader for RequestStartViewHeader {
--- a/core/consensus/src/client_table.rs
+++ b/core/consensus/src/client_table.rs
@@ -80,6 +80,13 @@ impl CachedReply {
 /// Real requests start at 1 (header validation enforces `request > 0`).
 pub const REGISTER_REQUEST_ID: u64 = 0;
 
+/// Exclusive ceiling on a checkpointed slot index.
+///
+/// Bounds the table [`ClientTable::from_snapshot`] allocates from an index it read off
+/// disk. Mirrors the config's `MAX_METADATA_CLIENTS_TABLE_MAX`, the largest capacity an
+/// operator can configure, so no valid checkpoint can carry an index at or above it.
+pub const CLIENTS_TABLE_SLOT_MAX: usize = 1 << 16;
+
 /// Committed replies retained per entry, newest at the back.
 ///
 /// The back is the latest committed reply and is structurally safe:
@@ -156,17 +163,72 @@ pub struct ClientEntrySnapshot {
     pub watermark_checksum: u128,
     /// Wire bytes of the entry's latest committed reply, round-tripped through
     /// [`CachedReply::from_message`]. Never empty: registration seeds the ring.
+    ///
+    /// Serialized as a msgpack `bin` blob, not the integer array a plain `Vec<u8>`
+    /// produces, which spends 2 bytes on every byte >= 0x80 and runs a checkpoint's
+    /// reply payload up to roughly double on disk.
+    #[serde(with = "reply_bytes")]
     pub reply: Vec<u8>,
 }
 
-/// Serializable [`ClientTable`].
+/// Serializable [`ClientTable`]: the occupied slots, each with its index.
 ///
-/// Slots keep their position, preserving deterministic eviction order across a
-/// checkpoint; [`ClientTable::from_snapshot`] rebuilds the index from the
-/// occupied slots' `client_id`s.
+/// Slot positions are carried explicitly rather than by array position, so the
+/// encoded form is proportional to live clients instead of to configured capacity.
+/// The alternative (a full `Vec<Option<_>>`) makes the slot count self-perpetuating:
+/// [`ClientTable::from_snapshot`] would have to honour the array's length, so
+/// lowering `clients_table_max` could never take effect, and every checkpoint would
+/// serde-walk `clients_table_max` entries while shard 0 is blocked on fsyncs.
+///
+/// Deterministic eviction order is unaffected: the index is what places each entry,
+/// and it survives here verbatim.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ClientTableSnapshot {
-    pub slots: Vec<Option<ClientEntrySnapshot>>,
+    pub slots: Vec<(u32, ClientEntrySnapshot)>,
+}
+
+/// Serializes reply bytes as a msgpack `bin` blob. See [`ClientEntrySnapshot::reply`].
+mod reply_bytes {
+    use serde::de::{Error, SeqAccess, Visitor};
+    use serde::{Deserializer, Serializer};
+    use std::fmt;
+
+    pub fn serialize<S: Serializer>(bytes: &[u8], serializer: S) -> Result<S::Ok, S::Error> {
+        serializer.serialize_bytes(bytes)
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<Vec<u8>, D::Error> {
+        deserializer.deserialize_byte_buf(BytesVisitor)
+    }
+
+    struct BytesVisitor;
+
+    impl<'de> Visitor<'de> for BytesVisitor {
+        type Value = Vec<u8>;
+
+        fn expecting(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            f.write_str("reply bytes")
+        }
+
+        fn visit_bytes<E: Error>(self, bytes: &[u8]) -> Result<Self::Value, E> {
+            Ok(bytes.to_vec())
+        }
+
+        fn visit_byte_buf<E: Error>(self, bytes: Vec<u8>) -> Result<Self::Value, E> {
+            Ok(bytes)
+        }
+
+        /// A checkpoint written before the `bin` encoding holds an integer array.
+        /// Accepting it keeps this a read-compatible change rather than one that
+        /// refuses a checkpoint it could decode.
+        fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
+            let mut bytes = Vec::with_capacity(seq.size_hint().unwrap_or_default());
+            while let Some(byte) = seq.next_element()? {
+                bytes.push(byte);
+            }
+            Ok(bytes)
+        }
+    }
 }
 
 /// A [`ClientTableSnapshot`] could not be decoded into a [`ClientTable`], so a
@@ -192,6 +254,20 @@ pub enum ClientTableDecodeError {
         /// The duplicated client id.
         client_id: u128,
     },
+    /// Two entries claim the same slot index. The second would overwrite the first,
+    /// leaving that client indexed onto another's state, so the decode is rejected.
+    DuplicateSlot {
+        /// The repeated slot index.
+        slot: usize,
+    },
+    /// A slot index is past what any configured capacity can produce, so honouring
+    /// it would size the table from a corrupt length.
+    SlotOutOfRange {
+        /// The out-of-range slot index.
+        slot: usize,
+        /// Exclusive ceiling on a slot index.
+        max: usize,
+    },
 }
 
 impl fmt::Display for ClientTableDecodeError {
@@ -210,6 +286,14 @@ impl fmt::Display for ClientTableDecodeError {
                 "client-table checkpoint slot {slot} repeats client_id {client_id} already in \
                  slot {first_slot}"
             ),
+            Self::DuplicateSlot { slot } => write!(
+                f,
+                "client-table checkpoint holds two entries for slot {slot}"
+            ),
+            Self::SlotOutOfRange { slot, max } => write!(
+                f,
+                "client-table checkpoint slot index {slot} is past the {max}-slot ceiling"
+            ),
         }
     }
 }
@@ -218,7 +302,9 @@ impl std::error::Error for ClientTableDecodeError {
     fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
         match self {
             Self::InvalidReply { source, .. } => Some(source),
-            Self::DuplicateClientId { .. } => None,
+            Self::DuplicateClientId { .. }
+            | Self::DuplicateSlot { .. }
+            | Self::SlotOutOfRange { .. } => None,
         }
     }
 }
@@ -357,35 +443,33 @@ impl ClientTable {
         *self = Self::new(max_clients);
     }
 
-    /// Check a request against the table. Epoch fence first, then the
-    /// watermark. Register does not come through here: every bind proposes
-    /// unconditionally so its fence actually moves, see
-    /// [`Self::commit_register`].
+    /// Snapshot the table for the metadata checkpoint: every occupied slot with its
+    /// index, so positions (and with them deterministic eviction order) survive, plus
+    /// each entry's `client_id` so the index rebuilds on decode.
     ///
-    /// Snapshot the table for the metadata checkpoint, preserving slot positions
-    /// (deterministic eviction order) and carrying each occupied slot's
-    /// `client_id` so the index rebuilds on decode.
-    ///
-    /// An occupied slot with no index entry is an index/slots desync, already
-    /// unreachable through the live table since every lookup goes via the index.
-    /// It is logged and dropped rather than panicking: this runs on shard 0's
-    /// checkpoint task under the durability lock, where an unwind would take down
-    /// the owner shard mid-checkpoint.
+    /// Walks only occupied slots. This runs on shard 0's checkpoint task, inside the
+    /// section where that core is already blocked on the snapshot's fsyncs, so it must
+    /// not also allocate and serde-walk one element per configured slot.
     #[must_use]
     pub fn to_snapshot(&self) -> ClientTableSnapshot {
         let slots = self
             .slots
             .iter()
-            .map(|slot| {
+            .enumerate()
+            .filter_map(|(slot_idx, slot)| {
                 let entry = slot.as_ref()?;
-                Some(ClientEntrySnapshot {
-                    client_id: entry.client_id,
-                    epoch: entry.epoch,
-                    user_id: entry.user_id,
-                    watermark: entry.watermark,
-                    watermark_checksum: entry.watermark_checksum,
-                    reply: entry.latest().as_bytes().to_vec(),
-                })
+                let slot_idx = u32::try_from(slot_idx).ok()?;
+                Some((
+                    slot_idx,
+                    ClientEntrySnapshot {
+                        client_id: entry.client_id,
+                        epoch: entry.epoch,
+                        user_id: entry.user_id,
+                        watermark: entry.watermark,
+                        watermark_checksum: entry.watermark_checksum,
+                        reply: entry.latest().as_bytes().to_vec(),
+                    },
+                ))
             })
             .collect();
         ClientTableSnapshot { slots }
@@ -399,30 +483,44 @@ impl ClientTable {
     /// older retransmits degrade from [`RequestStatus::Duplicate`] to
     /// [`RequestStatus::AlreadyApplied`].
     ///
-    /// `min_slots` is the configured capacity, padded on as free slots when the
-    /// checkpoint was taken under a smaller one. Slot positions are preserved
-    /// either way, so eviction order is unchanged. A capacity *lowered* below the
-    /// checkpoint's cannot be honoured here without dropping recovered sessions,
-    /// so the larger count stands until those entries drain.
+    /// `min_slots` is the configured capacity. The rebuilt table is that, or enough
+    /// to hold the highest occupied index when the checkpoint was taken under a
+    /// larger capacity: a capacity *lowered* below a live entry's slot cannot be
+    /// honoured without dropping a recovered session, so the larger count stands
+    /// until those entries drain, and a later checkpoint then rebuilds at the
+    /// configured size. Slot positions are preserved either way, so eviction order is
+    /// unchanged.
     ///
     /// # Errors
     /// [`ClientTableDecodeError`] if a slot's reply bytes are not a valid reply
-    /// message, or if two slots share a `client_id` (a corrupt, torn, or foreign
-    /// checkpoint). Surfaced rather than panicked so a bad checkpoint refuses boot
-    /// instead of unwinding the shard. Callers verify the checkpoint's checksum
+    /// message, if two slots share a `client_id`, if two entries claim the same slot,
+    /// or if a slot index is past [`CLIENTS_TABLE_SLOT_MAX`] (a corrupt, torn, or
+    /// foreign checkpoint). Surfaced rather than panicked so a bad checkpoint refuses
+    /// boot instead of unwinding the shard. Callers verify the checkpoint's checksum
     /// against the superblock first, so a correct boot never hits this.
     pub fn from_snapshot(
         snapshot: ClientTableSnapshot,
         min_slots: usize,
     ) -> Result<Self, ClientTableDecodeError> {
-        let capacity = snapshot.slots.len().max(min_slots);
-        let mut index = HashMap::with_capacity(capacity);
+        // Bound the capacity on a slot index read off disk before allocating from it,
+        // as the superblock and WAL do with their length fields.
+        let mut capacity = min_slots;
+        for (slot_idx, _) in &snapshot.slots {
+            let slot = *slot_idx as usize;
+            if slot >= CLIENTS_TABLE_SLOT_MAX {
+                return Err(ClientTableDecodeError::SlotOutOfRange {
+                    slot,
+                    max: CLIENTS_TABLE_SLOT_MAX,
+                });
+            }
+            capacity = capacity.max(slot + 1);
+        }
+
+        let mut index = HashMap::with_capacity(snapshot.slots.len());
         let mut slots = Vec::with_capacity(capacity);
-        for (slot_idx, slot) in snapshot.slots.into_iter().enumerate() {
-            let Some(entry) = slot else {
-                slots.push(None);
-                continue;
-            };
+        slots.resize_with(capacity, || None);
+        for (slot_idx, entry) in snapshot.slots {
+            let slot_idx = slot_idx as usize;
             let reply = Message::<ReplyHeader>::try_from(Owned::<MESSAGE_ALIGN>::copy_from_slice(
                 &entry.reply,
             ))
@@ -444,7 +542,12 @@ impl ClientTable {
             let latest_commit = reply.header().commit;
             let mut ring = VecDeque::with_capacity(REPLY_RING_CAPACITY);
             ring.push_back(CachedReply::from_message(reply));
-            slots.push(Some(ClientEntry {
+            // Two entries claiming one slot would silently drop the first, leaving it
+            // indexed but pointing at another client's state.
+            if slots[slot_idx].is_some() {
+                return Err(ClientTableDecodeError::DuplicateSlot { slot: slot_idx });
+            }
+            slots[slot_idx] = Some(ClientEntry {
                 epoch: entry.epoch,
                 user_id: entry.user_id,
                 watermark: entry.watermark,
@@ -452,9 +555,8 @@ impl ClientTable {
                 ring,
                 client_id: entry.client_id,
                 latest_commit,
-            }));
+            });
         }
-        slots.resize_with(capacity, || None);
         Ok(Self { slots, index })
     }
 
@@ -1014,6 +1116,73 @@ mod tests {
         assert_eq!(restored.get_epoch(3), Some(30));
     }
 
+    // A LOWERED `clients_table_max` must take effect too. It cannot while the encoded
+    // form is one element per configured slot, since the rebuilt table has to be at
+    // least as long as that array.
+    #[test]
+    fn from_snapshot_shrinks_to_the_configured_capacity() {
+        let mut table = ClientTable::new(8);
+        table.commit_register(1, TEST_USER_ID, make_register_reply(1, 10));
+
+        let mut restored = ClientTable::from_snapshot(table.to_snapshot(), 2).unwrap();
+
+        // Capacity 2: the recovered session plus one, and the third register evicts.
+        restored.commit_register(2, TEST_USER_ID, make_register_reply(2, 20));
+        restored.commit_register(3, TEST_USER_ID, make_register_reply(3, 30));
+        assert_eq!(restored.count(), 2);
+        assert_eq!(
+            restored.get_epoch(1),
+            None,
+            "the oldest committed entry is the eviction victim at the lowered capacity"
+        );
+    }
+
+    // A capacity lowered below a live entry's slot cannot be honoured without dropping
+    // a recovered session, so the table keeps room for it.
+    #[test]
+    fn from_snapshot_keeps_a_slot_above_the_configured_capacity() {
+        let mut table = ClientTable::new(8);
+        table.commit_register(1, TEST_USER_ID, make_register_reply(1, 10));
+        let mut snapshot = table.to_snapshot();
+        snapshot.slots[0].0 = 5;
+
+        let restored = ClientTable::from_snapshot(snapshot, 2).unwrap();
+        assert_eq!(
+            restored.get_epoch(1),
+            Some(10),
+            "an entry above the configured capacity must survive, not be dropped"
+        );
+    }
+
+    #[test]
+    fn from_snapshot_rejects_two_entries_in_one_slot() {
+        let mut table = ClientTable::new(2);
+        table.commit_register(1, TEST_USER_ID, make_register_reply(1, 10));
+        table.commit_register(2, TEST_USER_ID, make_register_reply(2, 20));
+        let mut snapshot = table.to_snapshot();
+        snapshot.slots[1].0 = snapshot.slots[0].0;
+
+        assert!(matches!(
+            ClientTable::from_snapshot(snapshot, 0),
+            Err(ClientTableDecodeError::DuplicateSlot { slot: 0 })
+        ));
+    }
+
+    #[test]
+    fn from_snapshot_rejects_a_slot_index_past_the_ceiling() {
+        // The capacity is allocated from this index, so an out-of-range one must be
+        // refused before the allocation rather than sized from.
+        let mut table = ClientTable::new(2);
+        table.commit_register(1, TEST_USER_ID, make_register_reply(1, 10));
+        let mut snapshot = table.to_snapshot();
+        snapshot.slots[0].0 = u32::MAX;
+
+        assert!(matches!(
+            ClientTable::from_snapshot(snapshot, 0),
+            Err(ClientTableDecodeError::SlotOutOfRange { .. })
+        ));
+    }
+
     // A raised `clients_table_max` must take effect on the next boot rather than
     // staying inert behind the checkpoint's slot count.
     #[test]
@@ -1037,7 +1206,7 @@ mod tests {
         table.commit_register(1, TEST_USER_ID, make_register_reply(1, 10));
         table.commit_register(2, TEST_USER_ID, make_register_reply(2, 20));
         let mut snapshot = table.to_snapshot();
-        snapshot.slots[1] = snapshot.slots[0].clone();
+        snapshot.slots[1].1 = snapshot.slots[0].1.clone();
 
         assert!(matches!(
             ClientTable::from_snapshot(snapshot, 0),
@@ -1054,7 +1223,7 @@ mod tests {
         let mut table = ClientTable::new(2);
         table.commit_register(1, TEST_USER_ID, make_register_reply(1, 10));
         let mut snapshot = table.to_snapshot();
-        snapshot.slots[0].as_mut().expect("slot 0 occupied").reply = vec![0xff; 8];
+        snapshot.slots[0].1.reply = vec![0xff; 8];
 
         assert!(matches!(
             ClientTable::from_snapshot(snapshot, 0),
--- a/core/consensus/src/client_table.rs
+++ b/core/consensus/src/client_table.rs
@@ -16,8 +16,11 @@
 // under the License.
 
 use iggy_binary_protocol::ReplyHeader;
-use server_common::{MESSAGE_ALIGN, Message, iobuf::Frozen};
+use iggy_binary_protocol::consensus::ConsensusError;
+use serde::{Deserialize, Serialize};
+use server_common::{MESSAGE_ALIGN, Message, iobuf::Frozen, iobuf::Owned};
 use std::collections::{HashMap, VecDeque};
+use std::fmt;
 use std::mem::size_of;
 use tracing::trace;
 
@@ -64,6 +67,12 @@ impl CachedReply {
         Self {
             bytes: msg.into_generic().into_frozen(),
         }
+    }
+
+    /// Raw reply bytes for checkpoint serialization, round-tripped through
+    /// [`Self::from_message`] on decode.
+    fn as_bytes(&self) -> &[u8] {
+        self.bytes.as_slice()
     }
 }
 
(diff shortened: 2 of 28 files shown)


core/journal/src/prepare_journal.rs
```diff
@@ -344,26 +350,68 @@ impl PrepareJournal {
...
+            let body_len = (entry_size - HEADER_SIZE as u64) as usize;
+            // `read_at` (read_exact_at) fills the buffer to capacity, so it must
+            // hold exactly `body_len`. A prior buffer of the same length is reused
+            // as-is; any size change replaces it, since capacity cannot shrink in
+            // place and an oversized buffer would read past the entry.
+            if body_buf.len() != body_len {
+                body_buf = vec![0u8; body_len];
+            }
+            body_buf = storage.read_at(pos + HEADER_SIZE as u64, body_buf).await?;
+            if u128::from(XxHash3_64::oneshot(&body_buf)) != header.checksum_body {
```
numinnex: **Blocker: no format gate on `checksum_body`, so any pre-existing WAL refuses boot.**

Every entry written before this change carries `checksum_body == 0`, so this check fails all of them and the interior-corruption branch below refuses boot. It also breaks mixed-version clusters: an old-version primary sends zero-checksum prepares, a new-version backup journals them verbatim, and that backup's next restart refuses boot, so a rolling upgrade is unsafe in that direction.

The `TODO(wal-integrity)` above already names this gap, but I think it needs to be part of this PR rather than a follow-up: treat `checksum_body == 0` as unsealed and skip verification for that entry, or gate on a WAL entry format version.

core/journal/src/superblock.rs
```diff
@@ -0,0 +1,579 @@
...
+            SlotClass::Valid {
+                sequence: seq_a,
+                payload: payload_a,
+            },
+            SlotClass::Valid {
+                sequence: seq_b,
+                payload: payload_b,
+            },
+        ) => SuperblockContents::Present(if seq_a >= seq_b { payload_a } else { payload_b }),
+        (SlotClass::Valid { payload, .. }, _) | (_, SlotClass::Valid { payload, .. }) => {
```
numinnex: **Blocker: `(Valid, Corrupt)` silently regresses durable state on newest-slot bit-rot.**

A torn write dies in the `.tmp` (rename is atomic), so a corrupt non-temp slot can only be bit-rot. When that hits the newest slot, this arm falls back to the older generation: older `view`/`log_view`, and the replica can re-vote in a view it already acted in, which is exactly the split-brain this PR exists to prevent.

Since the corrupt slot's sequence is unreadable, there is no way to tell whether it was the newer one, so I believe the sound response for `Valid + Corrupt` is refusing boot (`Unreadable`), the same as both-corrupt.

core/journal/src/superblock.rs
```diff
@@ -0,0 +1,579 @@
...
+        (SlotClass::Absent, SlotClass::Absent) => SuperblockContents::Empty,
+        (SlotClass::UnsupportedVersion(version), _)
+        | (_, SlotClass::UnsupportedVersion(version)) => SuperblockContents::Unreadable {
+            version: Some(version),
+        },
+        _ => SuperblockContents::Unreadable { version: None },
+    }
+}
+
+async fn read_sequence(path: &Path) -> io::Result<Option<u64>> {
```
numinnex: Compounding the `combine_slots` fallback above: `read_sequence` maps both Absent and Corrupt to `None`, so `open` aims the next write at the corrupt (possibly newest) slot and overwrites the evidence with `sequence = valid + 1`, retroactively legitimizing the regression. Distinguishing Absent from Corrupt here would fall out of the same fix.

numinnex (commented):
Two blockers, tagged inline (reviewed in the context of the upcoming metadata state transfer work, which builds on this).

core/server-ng/src/bootstrap.rs
```diff
@@ -1899,12 +1939,29 @@ fn restore_metadata_consensus(
...
+    // `else` means it is genuinely absent: a fresh node, or one that took writes but
+    // never checkpointed or changed view. There, inferring the view from the last WAL
+    // prepare is safe, since the persist-before-send gate guarantees this replica
+    // never externalized a view beyond what a re-probe re-derives, and it re-probes
+    // as a backup below. log_view cannot be inferred and stays 0 until the next
+    // superblock write.
+    if let Some(state) = recovered_state {
+        consensus.set_view(state.view);
+        consensus.set_log_view(state.log_view);
+        consensus.mark_superblock_durable(state.view, state.log_view);
```
hubcio, line 2003: the probe gate at line 1979 is still `replica_count > 1 && restored_op > 0`, but this block just made `view` recoverable independently of the WAL. a node with an empty WAL and a non-zero recovered view now falls into the `else`: `init()`, `Status::Normal`, `is_primary()` true for that view, no probe, `ceded_primaryship` left false. it resumes primaryship in a view the rest of the cluster may have moved past.

reachable on any multi-node cluster that changes view before its first metadata write and then bounces. view changes persist here without touching the WAL, so `restored_op > 0` no longer means "had a prior life" - the superblock broke the proxy this gate relied on.

`given_advanced_view_when_metadata_replica_restarts_should_recover_view_from_superblock` drives exactly this path (5 replicas, zero traffic, every WAL empty) and passes, so the flagship test demonstrates it while asserting success. same gate in `core/simulator/src/replica.rs:208`.

one condition: `restored_op > 0 || recovered_state.is_some()`, in both places.

core/metadata/src/impls/metadata.rs
```diff
@@ -239,63 +282,71 @@ impl<M> SnapshotCoordinator<M> {
...
+        let mut snapshot = (self.create_snapshot)(stm, commit_op, created_at)?;
+        // Fold in the client table, which `create_snapshot` does not see since it is
+        // not a state machine. Recovery restores it as the reconstruction floor for
+        // the drained WAL prefix.
+        snapshot.snapshot.client_table = client_table;
+        // Encode once: the checksum and the on-disk record share one buffer, so the
+        // snapshot is not serialized twice while the durability lock freezes this
+        // core.
+        let encoded = snapshot.encode()?;
+        let checksum = u128::from(calculate_checksum(&encoded));
```
hubcio: the pairing checksum is `hash(encoded_bytes)` here, but recovery recomputes it as `hash(encode(decode(disk)))` (`recovery.rs:319` via `IggySnapshot::checksum`). those agree only while decode-then-encode is byte-identical to what is on disk.

`rmp_serde::to_vec` writes positional arrays, so the next trailing `#[serde(default)]` field breaks it: an old file decodes fine (the default fills the missing element) but re-encodes with one MORE array element, the checksums diverge, and every checkpointed node hits `CheckpointChecksumMismatch` and refuses boot - with the WAL prefix already drained.

that append is the migration this repo mandates. `stm/consumer_group.rs:844-847` documents trailing `serde(default)` as THE forward-compatible pattern, and it is already used across the snapshot tree, including by this diff.

it escapes today only because the superblock is new in the same commit: a master-written dir has no slots, `recovered_state` is `None`, and `verify_checkpoint_pairing` short-circuits. armed here, fires on the next field.

fix is to hash the raw bytes `IggySnapshot::load` already reads and throws away. note that both new byte-stability tests are same-build round trips, so no CI can catch this - it needs two builds and a file that outlives one.

core/server-ng/src/bootstrap.rs
```diff
@@ -1899,12 +1939,29 @@ fn restore_metadata_consensus(
...
+    // unreadable superblock already refused boot in `recover()`, so reaching the
+    // `else` means it is genuinely absent: a fresh node, or one that took writes but
+    // never checkpointed or changed view. There, inferring the view from the last WAL
+    // prepare is safe, since the persist-before-send gate guarantees this replica
+    // never externalized a view beyond what a re-probe re-derives, and it re-probes
+    // as a backup below. log_view cannot be inferred and stays 0 until the next
+    // superblock write.
+    if let Some(state) = recovered_state {
+        consensus.set_view(state.view);
+        consensus.set_log_view(state.log_view);
```
hubcio, line 2002: restoring `log_view` durably while the head still comes from the WAL pairs a view number with a log that view never covered.

`handle_start_view` moves the sequencer down to the adopted head (`consensus/src/impls.rs:2446`) but nothing truncates the WAL suffix - no such primitive exists. so a replica that adopted a StartView at view 6 with head 105, while its own WAL still ends at 120, comes back with `log_view = 6` and `restored_op = 120`.

the equal-view skip at `consensus/src/impls.rs:2410` (`msg_view == log_view && msg_op < current_sequence()`) then drops the primary's StartView, including the probe reply that echoes our own incarnation and passes the new guard. after `probe_attempts_max` probes it falls back to an election, and its DVC carries `(log_view=6, op=120)`, which beats the real primary's `(6, 105)` in `dvc_select_winner`'s `max_by(log_view, then op)`. ops that view 6 durably discarded come back as view-7 entries and get pushed cluster-wide over committed bodies.

this was unreachable before: `log_view` was never restored, so it stayed 0 and the skip could not fire after a restart. the new fixture says so out loud at `impls.rs:3348` - "log_view stays 0 so the equal-view-old-op skip does not fire". the guard's own test runs in a state production restarts no longer produce.

cheapest fix that works: compare `msg_op < self.commit_min()` at `impls.rs:2410` instead of `< current_sequence()`. that also kills the DVC inversion, since adoption then drops the sequencer. the correct fix is a real suffix-truncation primitive. adding a `VsrState` field is not an option - `try_from` is exact-length, so every existing record would fail `WrongLength` and refuse boot with no migration path.

`cluster_view_durability_vsr` is one setup step from covering this: it already restarts a StartView adopter, it just settles first so no divergent suffix exists.

core/consensus/src/vsr_state.rs
```diff
@@ -0,0 +1,165 @@
...
+/// + `checkpoint_checksum`(16).
+pub const ENCODED_LEN: usize = 58;
+
+/// The durable consensus state of one replica for one consensus group.
+///
+/// A view a replica acted in must survive a crash, or it can re-participate in
+/// an old view and split the log.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct VsrState {
+    /// Cluster this superblock belongs to. Recovery rejects a mismatch (a copied
```
hubcio: this says recovery rejects a cluster mismatch, and line 46 says the same for `replica_id`. nothing does. `cluster`, `replica_id`, `replica_count` and `commit_max` have zero readers - `recovered_state` is consumed in exactly three places (`bootstrap.rs:1960`, `simulator/src/replica.rs:195`, `recovery.rs:322`) and all three touch only `view`, `log_view`, `checkpoint_op` and `checkpoint_checksum`. 4 of 8 durable fields are write-only.

this matters more than usual because the superblock is the only on-disk identity in the metadata dir, and the wire handshake cannot stand in: both peers derive `cluster` from `auth::cluster_domain_id(&config.cluster.name)`, so it never sees a foreign DISK. and the format freezes at version 1 with no migration path, so adding the check later needs a version bump that invalidates every existing record.

scope it honestly though - identity fields catch misplacement only, not staleness. an older backup of the node's own dir has the right `replica_id` and passes.

related, and cheap while you are here: `try_from` validates length only, so it accepts `log_view > view` against the invariant on line 53. `DoViewChangeHeader::validate` rejects that at the receiver, so such a record would make every peer drop this replica's DVCs while it looks healthy locally.

implement the check or delete the promise. shipping the claim without the code is the option to avoid.

core/consensus/src/impls.rs
```diff
@@ -2704,6 +2826,18 @@ where
...
+        // Seal the body integrity field over the payload past the 256-byte header
+        // (the client request bytes, carried through the transmute unchanged).
+        // Computed once by the primary and replicated verbatim, so every backup
+        // stores the same value and the scan verifies it after a crash. The body is
+        // never re-stamped (`restamp_prepare_view` patches only `view`), so this
+        // survives view-change retransmits. The header `checksum` and its `parent`
+        // chain stay `0`: activating them needs the retransmit path to re-seal a
+        // re-stamped header, a separate change.
+        let checksum_body = u128::from(calculate_checksum(
+            &self.as_slice()[size_of::<PrepareHeader>()..],
```
hubcio: `Project::project` is one generic impl with no namespace gate, so partition produce prepares (`partitions/src/iggy_partition.rs:1298,:1356`) pay an unconditional full-payload XxHash3 pass on top of `batch_checksum`, which is already computed and already verified.

worse, it is provably wrong there: `stamp_prepare_for_persistence` (`server_common/src/send_messages2.rs:706-710`) rewrites the command header INSIDE the sealed region, so every partition `checksum_body` describes pre-stamp bytes.

harmless today only because nothing on that plane reads it - dedup keys on `header.checksum` (still 0), and `PrepareOkHeader.prepare_checksum` copies `checksum`, not `checksum_body`. it turns into a hard failure the moment partitions get a durable `PrepareJournal`: every entry then hits the interior-corruption branch at `prepare_journal.rs:418-430`, which has no truncation path.

gate on `METADATA_CONSENSUS_NAMESPACE`. partition prepares then fall back to `checksum_body == 0`, which is the designed "nothing to verify" sentinel, so that is strictly better than deleting the seal. worth doing in this diff, because after merge the landmine is invisible.

core/journal/src/prepare_journal.rs
```diff
@@ -344,26 +368,72 @@ impl PrepareJournal {
...
+            // `commit_watermark = max(header.commit)`, so a flipped `commit` makes it
+            // apply prepared-but-uncommitted ops as committed, the very ops a view
+            // change may have truncated cluster-wide, diverging this replica. Seal
+            // and verify the header checksum + parent chain.
+            if header.checksum_body == CHECKSUM_BODY_UNSEALED {
+                // Skip the body read too, so a WAL written entirely by a pre-sealing
+                // build scans without touching its payload.
+                unsealed_entries += 1;
+            } else {
+                let body_len = (entry_size - HEADER_SIZE as u64) as usize;
```
hubcio, line 493: the producer seals `[HEADER..]` over the whole buffer (`consensus/src/impls.rs:2837`), but the scan verifies `[HEADER..header.size]`. `Message::try_from` explicitly permits `len >= size` (`server_common/src/consensus_message.rs:443`), so the two disagree whenever a buffer carries slack. latent right now, since every reachable producer is exact-length.

do NOT fix it by narrowing the seal to `header.size`. `append` writes the whole buffer (`entry.into_owned()` at line 833, no slice), so the slack lands on disk either way. with a narrowed seal the checksum passes, `pos += entry_size` lands in the slack, the header parse fails, and control reaches `truncate_or_fail`, which silently truncates and fsyncs for anything under 64 MiB. entry A of size 320 in a 384-byte buffer followed by an exact entry B: today that is a loud interior-corruption refusal with B intact, narrowed it is a truncate to 320 and B is gone unrecoverably.

the whole-buffer seal is accidentally load-bearing as an over-length tripwire. if you want to narrow it, `append` has to reject `buffer.len() != header.size` before `write_append` first, next to the existing pre-write slot-collision check - same rationale.

core/metadata/src/impls/recovery.rs
```diff
@@ -163,6 +269,58 @@ where
...
+    //
+    // TODO(state-transfer): there is no metadata state transfer yet, so refusing boot
+    // is the only sound response. Once it lands, fetch the checkpoint from a healthy
+    // replica here instead of erroring.
+    let snapshot_op = snapshot.as_ref().map_or(0, IggySnapshot::sequence_number);
+    let snapshot_checksum = match snapshot.as_ref() {
+        Some(snapshot) => snapshot.checksum()?,
+        None => 0,
+    };
+    verify_checkpoint_pairing(recovered_state.as_ref(), snapshot_op, snapshot_checksum)?;
```
hubcio: "verify before decode" is not the ordering. `IggySnapshot::load` at line 268 decodes the whole snapshot 54 lines before this runs, and `RecoveredMetadata.superblock`'s doc at line 177 makes the same claim.

consequence is diagnostic rather than unsafe - the STM restore and `from_snapshot` do run after the verify - but a corrupt snapshot surfaces as a generic msgpack error instead of the typed `CheckpointChecksumMismatch` this diff built for it.

core/metadata/src/impls/metadata.rs
```diff
@@ -2236,31 +2467,73 @@ where
...
-            .checkpoint_if_needed(&self.mux_stm, journal, snap_op, created_at)
-            .await
-        {
-            Ok(true) => {
-                debug!(
+
+        // Durability ordering, must not be reordered: persist the snapshot, durably
+        // record the (checkpoint_op, checksum, commit_max) pairing in the superblock,
+        // THEN drain the snapshotted prefix from the WAL. A crash before the
+        // superblock write recovers the prior checkpoint with the WAL intact; a crash
```
hubcio, line 2650: this is not true. phase 1 renames the new snapshot over the single `snapshot.bin` (`write_durably`, line 152), so by the time the superblock write runs there is no prior checkpoint left to recover.

a crash in that window boots into `verify_checkpoint_pairing`'s `checkpoint_op < snapshot_op` arm (`recovery.rs:517-529`), which accepts and decodes the snapshot with no integrity check at all - and `snapshot.bin` carries no internal checksum of its own. so that arm is the normal outcome of a crash inside a checkpoint, not the corner case it is written up as.

either give `snapshot.bin` a trailer or sidecar checksum, or keep two generations (`snapshot.<op>.bin`) and let the superblock pick. do not reorder the phases - that just makes `CheckpointAheadOfSnapshot` the normal crash outcome instead.

core/consensus/src/vsr_state.rs
```diff
@@ -0,0 +1,165 @@
...
+    pub replica_id: u8,
+    /// Cluster size at write time.
+    pub replica_count: u8,
+    /// Current view.
+    pub view: u32,
+    /// Latest view in which this replica changed its head (adopted a `StartView`
+    /// as a backup, or completed a DVC quorum as primary). `log_view <= view`.
+    pub log_view: u32,
+    /// Highest op known committed by the cluster at write time.
+    pub commit_max: u64,
```
hubcio, line 66: `commit_max` is written on every superblock write and read by nothing. recovery still folds `max(header.commit)` over the WAL (`recovery.rs:366-376`), over a field that `prepare_journal.rs:386-393` admits is unsealed and bit-flippable.

it is also not quite the value the TODOs want - `bootstrap.rs:2006-2011` asks for a commit-path watermark, and this is a coarser thing wired to neither. either wire it or drop it now, because it is frozen by `SUPERBLOCK_VERSION` plus the offset-pinning test, and a bump turns every existing record `Unreadable`, so "wire it later" is a downtime migration.

core/journal/src/prepare_journal.rs
```diff
@@ -344,26 +368,72 @@ impl PrepareJournal {
...
+                                "interior WAL corruption at pos {pos} (op {}, operation {:?}): \
+                                 prepare body checksum mismatch with {} bytes of entries \
+                                 following; refusing to truncate and discard the committed suffix",
+                                header.op,
+                                header.operation,
+                                file_len - (pos + entry_size),
+                            ),
+                        )));
+                    }
+                    truncate_or_fail(&storage, pos, "prepare body checksum mismatch at tail")
```
hubcio, line 525: two problems with routing here.

`truncate_or_fail` has zero tracing, and `append` fsyncs before `send_prepare_ok` returns - so an entry that was durable, acked and quorum-committed can be silently discarded at the next boot. asymmetric with the interior branch right above, which is loud.

and the two branches ask different questions. the interior one asks "did anything complete follow?", which is right. `truncate_or_fail` asks "is total trailing <= 64 MiB?" and infers torn-tail from that, which is sound only if damage is always at EOF. it cannot tell 100 bytes of torn tail from 100 bytes of damage followed by intact committed entries. its own comment at lines 206-209 states the narrower one-entry premise the code never tests. the right test is the one the interior branch already uses: probe forward for a parseable entry and refuse if one exists.

core/metadata/src/impls/metadata.rs
```diff
@@ -2202,6 +2350,75 @@ where
...
+            .as_ref()
+            .map_or((0, 0), SnapshotCoordinator::last_checkpoint);
+        let state = consensus.vsr_state(checkpoint_op, checkpoint_checksum);
+        match superblock.dyn_write(&state.to_bytes()).await {
+            Ok(()) => {
+                consensus.mark_superblock_durable(state.view, state.log_view);
+                true
+            }
+            Err(error) => {
+                tracing::error!(
```
hubcio: a persistent write error (ENOSPC, EIO) never clears `needs_superblock_persist()`, so every 10 ms consensus tick (`CONSENSUS_TICK_INTERVAL` in `shard/src/router.rs`) runs a full `atomic_replace` - create, write, fsync, rename, dir fsync - plus this `error!`. no backoff, no cap, no metric, no self-fence, on the shard-0 executor that also serves partition traffic. `degraded` cannot short-circuit it either: it is computed once in `open` from bad bytes, never from a write error.

worse, the caller drops the WHOLE action batch on `false` (`shard/src/lib.rs:2894-2898`), so a failed persist also suppresses `SendCommit`, `RetransmitPrepares` and the probe. the replica goes quiet and provokes the view changes that make the gate harder to clear. same shape at the other gated sites. keep the whole-batch withholding - filtering by "view-scoped" would only release `SendRequestStartView`, which unsticks nothing and turns a fail-closed branch into a classification that has to stay exhaustive forever.

a retry deadline on the tick path, plus a metric and a stated terminal policy, would cover it.

core/metadata/src/impls/metadata.rs
```diff
@@ -2236,31 +2467,73 @@ where
...
+        // record the (checkpoint_op, checksum, commit_max) pairing in the superblock,
+        // THEN drain the snapshotted prefix from the WAL. A crash before the
+        // superblock write recovers the prior checkpoint with the WAL intact; a crash
+        // after it recovers the new one. Draining before the superblock points at the
+        // new snapshot could strand committed ops on a crash. Each fallible step
+        // withholds the rest and returns early, leaving the WAL undrained, so
+        // `should_checkpoint` stays true and the next tick retries from the top at the
+        // then-current commit_min. The prepare being replicated appends regardless
+        // (see the phantom-op comment at the call site).
+        let client_table = self.client_table.borrow().to_snapshot();
```
hubcio, line 2657: this widens a blocking-in-async window that already existed. `write_durably` is blocking `std::fs` with two fsyncs (lines 130-170) and it runs on the compio shard-0 executor while `durability_lock` is held. the base `persist` did the same, but the section now also carries `to_snapshot()` at full configured slot count (default `clients_table_max = 8192`), the extra msgpack, and the superblock write - 2 fsyncs become 4 in one critical section. the sibling superblock writer in the same section already uses `compio::fs`.

second order: acks that arrive with a pending view persist now queue behind the whole checkpoint, including the async WAL drain where the executor is free but the lock is not. the lock-free fast path spares the steady state, but the ack path never waited on checkpoint I/O before this. holding the lock only around the superblock write would fix that half.

core/journal/src/superblock.rs
```diff
@@ -0,0 +1,698 @@
...
+        Err(e) if e.kind() == io::ErrorKind::NotFound => return Ok(SlotClass::Absent),
+        Err(e) => return Err(e),
+    };
+    let len = file.metadata().await?.len();
+    if len == 0 {
+        return Ok(SlotClass::Absent);
+    }
+    let len = usize::try_from(len)
+        .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "superblock file too large"))?;
+    let buf = vec![0u8; len];
```
hubcio, line 512: `vec![0u8; len]` straight off `file.metadata().len()` with no ceiling, and the `usize::try_from` above it is dead on 64-bit. the record is exactly 94 bytes and `parse_record` demands an exact length, so any oversized file this writer could have produced is necessarily corrupt - the allocation is spent to conclude "corrupt" before any validation runs. realistic trigger is a foreign or leftover file, not bit-rot, since the value is `st_size`.

`prepare_journal.rs:32-37` already has `MAX_ENTRY_SIZE` with a comment citing this exact hazard. same treatment here, checked before the allocation. worth bounding `payload_len` in `parse_record` too, since `decode_slots` is public and takes bytes directly.

core/journal/src/superblock.rs
```diff
@@ -0,0 +1,698 @@
...
+    result?;
+    // A checksum/magic/version failure is not an I/O error: the other slot may
+    // still be good, so classify the bytes and let `combine_slots` decide.
+    Ok(classify(&buf))
+}
+
+/// Replace `dir/file_name` atomically: write a temp, fsync it, rename over the
+/// target, then fsync the directory so the rename itself is durable.
+async fn atomic_replace(dir: &Path, file_name: &str, bytes: Vec<u8>) -> io::Result<()> {
+    let tmp_path = dir.join(format!("{file_name}.tmp"));
```
hubcio, line 523: `atomic_replace` leaks `superblock.{a,b}.tmp` on any failure before the rename. `PrepareJournal::drain` has `TmpFileGuard` (`prepare_journal.rs:242-263`) for exactly this.

not correctness - `File::create` truncates, and `next_sequence`/`next_slot` correctly stay un-advanced so a retry re-targets the same slot. hygiene.

core/journal/src/superblock.rs
```diff
@@ -0,0 +1,698 @@
...
+        // writers would target the same slot and could tear it while both return
+        // `Ok`. The ping-pong guarantee, that the non-written slot is always an
+        // intact prior generation, holds only with one write in flight.
+        let sequence = self.next_sequence.get();
+        let slot = self.next_slot.get();
+        // Built before the first await so the `payload` borrow does not span it.
+        let record = build_record(sequence, payload)?;
+        #[cfg(debug_assertions)]
+        assert!(
+            !self.writing.replace(true),
```
hubcio: the tripwire is set before the `.await` and cleared only on the completion path, so a dropped write future latches it for the store's lifetime and every later write panics with "called concurrently; writes must be externally serialized" - a message that sends whoever debugs it off to audit a lock that was never violated.

no confirmed live drop point today: every entry point is detached and the pump's select arm body is not cancellable, so this is cancel-unsafe by construction rather than reachable. still worth three lines - build an RAII guard after the assert instead of setting the flag by hand.

core/consensus/src/client_table.rs
```diff
@@ -251,6 +357,107 @@ impl ClientTable {
...
+    /// [`ClientTableDecodeError`] if a slot's reply bytes are not a valid reply
+    /// message, or if two slots share a `client_id` (a corrupt, torn, or foreign
+    /// checkpoint). Surfaced rather than panicked so a bad checkpoint refuses boot
+    /// instead of unwinding the shard. Callers verify the checkpoint's checksum
+    /// against the superblock first, so a correct boot never hits this.
+    pub fn from_snapshot(
+        snapshot: ClientTableSnapshot,
+        min_slots: usize,
+    ) -> Result<Self, ClientTableDecodeError> {
+        let capacity = snapshot.slots.len().max(min_slots);
```
hubcio: `capacity = snapshot.slots.len().max(min_slots)` plus a `to_snapshot` that emits trailing `None`s makes the slot count self-perpetuating across checkpoints. lowering `[metadata] clients_table_max` is permanently inert while `snapshot.bin` exists, and the rustdoc at line 406 ("the larger count stands until those entries drain") is false - draining occupied slots never shortens the Vec. a config knob that silently does nothing.

the fix is also the fix for the snapshot cost: make `slots` a `Vec<(u32, ClientEntrySnapshot)>` of occupied entries only. the slot index is already carried explicitly, so eviction order and the index rebuild are unchanged, and `to_snapshot` stops allocating and serde-walking a full 8192-entry Vec inside the window where shard 0 is blocked on fsyncs. worth doing now, since the format freezes at release.

core/shard/src/lib.rs
```diff
@@ -3024,6 +3048,34 @@ async fn dispatch_vsr_actions<B, P, J>(
...
+    // partition consensus has no superblock, so its `needs_superblock_persist` is
+    // always dirty and it is exempt. `RequestStartView` is exempt too, being a probe
+    // that asks to LEARN the view rather than advertise it.
+    #[cfg(debug_assertions)]
+    if consensus.namespace() == METADATA_CONSENSUS_NAMESPACE {
+        for action in actions {
+            let advertises_view = matches!(
+                action,
+                VsrAction::SendStartViewChange { .. }
+                    | VsrAction::SendDoViewChange { .. }
```
hubcio, line 3074: `VsrAction::SendCommit` is missing from this set. it is view-scoped: a backup drops it on `header.view != self.view` (`consensus/src/impls.rs:2534`) and a primary answers `header.view < self.view` with `RespondStartView` (`:2513`). gated today only because its sole emitter rides the tick, which already persists first.

core/shard/src/lib.rs
```diff
@@ -3024,6 +3048,34 @@ async fn dispatch_vsr_actions<B, P, J>(
         }
     };
 
+    // Centralized durable-before-send tripwire: a view-scoped message must never
+    // advertise a (view, log_view) the superblock has not recorded, or a crash could
+    // recover an older view than one a peer already saw, splitting the brain or
+    // losing a commit. Every metadata caller persists first (the view-change dispatch
+    // sites and the on_replicate / on_commit send gates), so this asserts they did
+    // rather than letting a future bypass through silently. Metadata plane only:
+    // partition consensus has no superblock, so its `needs_superblock_persist` is
```
hubcio: partition consensus is not always dirty. `needs_superblock_persist()` is `view != view_durable || log_view != log_view_durable`, which is CLEAN at view 0. the namespace test on the next line does all the actual work, so this rationale does not hold.

core/shard/src/lib.rs
```diff
@@ -3087,6 +3144,7 @@ async fn dispatch_vsr_actions<B, P, J>(
                         h.view = *view;
                         h.op = *op;
                         h.commit = *commit;
+                        h.incarnation = *incarnation;
```
hubcio, line 3161: the probe-answering `StartView` goes out via `broadcast` below, so the echoed incarnation reaches every peer and a concurrently-recovering peer rejects a perfectly fresh `StartView` through the new guard. that is a recovery-latency regression for simultaneous restarts, and it makes `RequestStartViewHeader`'s rustdoc claim of a "targeted `StartView`" false against this dispatcher.

`send(header.replica, ..)` for the echo arm, keep `broadcast` for the unsolicited case.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -1048,15 +1048,29 @@ pub struct StartViewHeader {
...
+    pub reserved: [u8; 88],
+    /// Sender's incarnation, echoed from the `RequestStartView` this answers so a
+    /// recovering replica can prove the reply post-dates its restart (see
+    /// `RequestStartViewHeader::incarnation`). `0` on an unsolicited `StartView`
+    /// (a normal view-change completion), which carries no freshness claim.
+    ///
+    /// Carved from the tail of the former `reserved` region and placed LAST so it
+    /// lands 16-aligned with no padding WITHOUT moving `op`/`commit`/`namespace`.
+    /// A peer that predates it sends zeros, decoding as `incarnation == 0`, inert
+    /// per the `handle_start_view` guard, so a mixed-version rolling upgrade is
```
hubcio: not inert on the metadata plane. the `handle_start_view` guard only ignores incarnation when `self_incarnation == 0`, and production always sets it non-zero (`rand::random::<u128>() | 1`, `bootstrap.rs:1946`). a zero-incarnation `StartView` from a pre-upgrade peer has `header.incarnation != self_incarnation`, so it gets classified stale whenever the other three conditions hold.

the escape hatches cover the common failover shapes so reachability is low, but the claim as written is not right. treating `header.incarnation == 0` as "no claim" and falling through would make it true.

core/metadata/src/impls/metadata.rs
```diff
@@ -2202,6 +2350,75 @@ where
...
 }
 
+impl<B, P, J, S, M> IggyMetadata<VsrConsensus<B, P>, J, S, M>
+where
+    B: MessageBus,
+    P: Pipeline<Entry = PipelineEntry>,
+{
+    /// Persist the current VSR state to the superblock when the view changed since
+    /// the last write. The split-brain gate: callers MUST invoke this before
+    /// dispatching any view-scoped VSR message, so a replica that acted in a view can
```
hubcio, line 2466: the gate fences the SEND, not the ACT. handlers mutate view, log_view, status, sequencer and pipeline before this ever runs, and `commit_journal()` sits outside the `if` at `shard/src/lib.rs:1977-1982` and `:2040-2045`, so a failed persist still applies committed ops and moves the state machine in a view that was never recorded.

durable-before-externalize is the right VSR fence and the code does implement it. the wording just invites a reader to assume local state is fenced too.

core/integration/tests/server/mod.rs
```diff
@@ -44,6 +44,14 @@ mod http_tls;
...
 #[cfg(feature = "vsr")]
 mod cluster_metadata_vsr;
+// A metadata view change must persist the advanced view and recover it from disk
+// across a replica restart.
+#[cfg(feature = "vsr")]
+mod cluster_view_durability_vsr;
+// A metadata checkpoint must drain the WAL and recover from the snapshot fold plus
+// the WAL suffix across a restart.
+#[cfg(feature = "vsr")]
+mod metadata_checkpoint_recovery_vsr;
```
hubcio, line 54: both of these are `#[cfg(feature = "vsr")]`, and CI runs `cargo nextest run --locked --no-fail-fast --profile ci` with no `--features vsr` and no `--all-features`, while `integration` has `default = ["login-session"]`. so all of it compiles under the `--all-features` clippy/check pass and never executes anywhere. the durability argument this PR rests on has no running coverage.

same root cause hides the release side: `-D warnings` rides a clippy invocation, which builds debug, so the release-only unused-import warning is never observed and the `cfg(debug_assertions)` tripwire has no release backstop.

core/simulator/src/deps.rs
```diff
@@ -242,8 +250,121 @@ impl JournalHandle for SimJournal<MemStorage> {
...
 #[derive(Debug, Default)]
 pub struct SimSnapshot {}
 
+/// In-memory superblock for the simulator.
+///
+/// Held by the harness behind an `Rc` so its bytes survive a replica being dropped
+/// and rebuilt across a restart. RAM has no torn writes, so it holds only the latest
+/// payload, with none of the framing or checksum `PingPongSuperblock` adds.
+#[derive(Debug, Default)]
+pub struct SimSuperblock {
```
hubcio, line 322: `SimSuperblock` is a bare last-write-wins RAM cell - no framing, no sequence, no ping-pong, no `Unreadable`. so the sim never exercises `PingPongSuperblock` at all: not slot selection, not the degraded refusal, not `combine_slots`.

with the vsr integration tests not running in CI either, everything rests on the 7 superblock unit tests, and none of those covers a crash mid-`atomic_replace` or the crash-between-superblock-write-and-drain window. `SLOT_FILE_NAMES` and `decode_slots` were made public precisely so a test could do this.

core/simulator/src/replica.rs
```diff
@@ -169,20 +185,80 @@ pub fn new_shard(
...
         );
-        consensus.init();
+        // Seed-derived incarnation, bumped per restart by the harness, so a StartView
+        // from a previous incarnation is ignored while replay stays deterministic; an
+        // OS-random nonce would break byte-identical replay.
+        consensus.set_incarnation(incarnation);
+        // View/log_view come from the durable superblock; op, commit, and the
+        // last-prepare markers come from the retained WAL. Independent inputs,
+        // mirroring server-ng's restore_metadata_consensus.
+        if let Some(state) = recovered_state {
```
hubcio, line 198: this drops the `else if let Some(header) = last_header { set_view(header.view) }` fallback production has at `bootstrap.rs:1964-1966`, so the `None`-superblock branch - the upgrade path every existing deployment takes on first boot - has zero sim coverage.

also, none of the restart tests calls `sim.step()` after `replica_restart`, so no message ever reaches a restarted replica and every chain starts at the first StartView delivered post-restart. the harness is otherwise faithful on the WAL-head derivation, so it could drive these paths, it just never does.

core/integration/tests/server/metadata_checkpoint_recovery_vsr.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    // Crossing CHECKPOINT_MARGIN must have driven the coordinator to persist a snapshot
+    // and drain the WAL prefix behind it.
+    let snapshot_path = harness
+        .node(0)
+        .data_path()
+        .join("metadata")
+        .join("snapshot.bin");
+    let snapshot_len = std::fs::metadata(&snapshot_path).map(|m| m.len()).ok();
+    assert!(
```
hubcio, line 79: one-way subset. the premise below is that stream 0 was drained and can only come back from the snapshot, but the only evidence collected is that `snapshot.bin` is non-empty. phase 3 of the checkpoint returns early on failure (`metadata.rs:2518-2527`), and if the drain never ran the WAL still holds the op and both assertions still pass.

assert that `journal.wal` shrank, or rename it aside before the restart.

core/simulator/src/lib.rs
```diff
@@ -891,6 +1032,306 @@ mod tests {
...
+                !elected,
+                "replica {replica} must not be elected primary while a survivor's \
+                 superblock write fails: the persist gate withholds its view-scoped \
+                 votes, so quorum cannot form"
+            );
+        }
+
+        // The faulted replica never durably recorded a new view, so a restart recovers
+        // its old view, never one it might have voted in.
+        assert_eq!(
```
hubcio, line 1377: this assertion cannot fail. `set_fail_writes()` runs before any possible write in a test with no traffic, so `read_latest_sync()` is trivially `None`.

the test also passes if the gate over-blocks, since nothing asserts the in-memory view actually advanced. adding `consensus.view() > 0` alongside the durable absence would make it prove what it claims.

core/metadata/src/impls/recovery.rs
```diff
@@ -43,6 +83,43 @@ impl fmt::Display for RecoveryError {
             Self::Journal(e) => write!(f, "recovery journal error: {e}"),
             Self::StateMachine(e) => write!(f, "recovery state machine error: {e}"),
             Self::Io(e) => write!(f, "recovery I/O error: {e}"),
+            Self::Superblock(e) => write!(f, "recovery superblock error: {e}"),
```
hubcio: none of the boot-refusal errors name a path or a slot, and `compio::fs::File::open` attaches none either, so the operator's chain terminates in "failed to recover metadata snapshot and journal" plus a pathless variant. worth threading the metadata dir and slot name through the boot-path variants. the write-refusal string is unreachable in the server (boot fails first on `Unreadable`), so it does not need the same treatment.

core/journal/src/superblock.rs
```diff
@@ -0,0 +1,698 @@
...
+    if version != SUPERBLOCK_VERSION {
+        // Magic matched, so a superblock writer produced this, but the framing
+        // past the version field may differ, so neither the payload length nor
+        // the checksum can validate it. Surface the version so recovery can name
+        // it (a downgrade).
+        return SlotClass::UnsupportedVersion(version);
+    }
+    // `parse_record` re-checks magic/version (cheap) and applies this build's
+    // length + checksum validation.
+    match parse_record(bytes) {
```
hubcio: `parse_record` re-runs the length, magic and version checks `classify` just did - the comment above says so. folding them into one `SlotClass`-returning function drops a double parse per slot and roughly 25 lines.

while in here: `SlotClass::Corrupt` and `UnsupportedVersion(u16)` differ only in what `combine_slots` reports, so `Corrupt { version: Option<u16> }` collapses them - as long as the current precedence survives (unrecognized version reported over generic corruption when both slots are bad), else the downgrade diagnostic is lost.

core/consensus/src/client_table.rs
```diff
@@ -124,6 +133,96 @@ struct ClientEntry {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ClientEntrySnapshot {
+    pub client_id: u128,
+    pub epoch: u64,
+    pub user_id: u32,
+    pub watermark: u64,
+    pub watermark_checksum: u128,
+    /// Wire bytes of the entry's latest committed reply, round-tripped through
+    /// [`CachedReply::from_message`]. Never empty: registration seeds the ring.
+    pub reply: Vec<u8>,
```
hubcio, line 171: `Vec<u8>` serializes as a msgpack array of integers, not `bin`, so every byte >= 0x80 costs 2 bytes and reply blobs run up to about 2x on disk. `#[serde(with = "serde_bytes")]` fixes it, and it is much cheaper to do before the format freezes.

core/consensus/src/client_table.rs
```diff
@@ -251,6 +357,107 @@ impl ClientTable {
         *self = Self::new(max_clients);
     }
 
+    /// Check a request against the table. Epoch fence first, then the
+    /// watermark. Register does not come through here: every bind proposes
+    /// unconditionally so its fence actually moves, see
```
hubcio: this doc block belongs to `check_request` - it was copied over and the real `to_snapshot` description starts in the second paragraph. the last paragraph is also wrong: the body does not detect or log an index/slots desync, it just maps over the slots.

core/server-ng/src/bootstrap.rs
```diff
@@ -935,6 +937,9 @@ async fn shard_main(
                     recovered.last_applied_op,
                     recovered.last_journaled_op,
                     recovered.client_table,
+                    recovered.superblock,
```
hubcio: `owner_state` is an anonymous positional 8-tuple rebuilt from `RecoveredMetadata`'s own fields and destructured by position 30 lines down, and `restore_metadata_consensus` is now 18 positional params, 8 of them `u64` ticks. passing `RecoveredMetadata` straight through, or a named struct, removes a whole class of silent misbinding. `ReplicaTopology` (`shard/src/lib.rs:100-109`) is the existing precedent for this.

core/metadata/src/impls/metadata.rs
```diff
@@ -3000,6 +3256,60 @@ mod tests {
...
+            "a populated snapshot must re-encode byte-identically after a decode; an \
+             unordered collection in the serialized form would break the checkpoint \
+             checksum cross-check and refuse boot on a healthy node"
+        );
+        assert_eq!(
+            snapshot.checksum().unwrap(),
+            decoded.checksum().unwrap(),
+            "persist-time and boot-time checkpoint checksums must agree"
+        );
+        assert_ne!(
```
hubcio, line 3489: this `assert_ne!` cannot fail - the populated snapshot and `IggySnapshot::new(7)` differ by construction. and neither new determinism test populates `users`, so the half of the invariant most likely to regress (someone swapping `PermissionerSnapshot`'s `Vec`s or `Permissions.streams`' `BTreeMap` for a `HashMap`) has no guard, with a hard boot refusal on a healthy node as the failure mode.

core/journal/src/prepare_journal.rs
```diff
@@ -35,6 +36,19 @@ const HEADER_SIZE: usize = size_of::<PrepareHeader>();
 /// multi-GiB allocation during the WAL scan.
 const MAX_ENTRY_SIZE: u64 = 64 * 1024 * 1024;
 
+/// `checksum_body` of an entry no producer sealed: written by a build predating
+/// body sealing, or replicated from a primary predating it (backups journal the
+/// header verbatim, so the zero travels along). Indistinguishable from
+/// sealed-then-corrupted, so the scan skips verification and counts these rather
+/// than reject a WAL it cannot prove is damaged. Safe as a sentinel: `XxHash3_64`
+/// never returns `0`, not even for an empty body, which a sealed zero-length entry
```
hubcio: "`XxHash3_64` never returns `0`" is not a property of a 64-bit hash - some input maps to zero, we just do not know which. the consequence is benign here since it fails open into the unsealed path, but the doc should state the probability rather than an impossibility.

core/integration/Cargo.toml
```diff
@@ -58,6 +61,9 @@ iggy_common = { workspace = true }
 # `build_label` function — keeping the test and production label format in lock-step.
 iggy_connector_doris_sink = { path = "../connectors/sinks/doris_sink" }
 iggy_connector_sdk = { workspace = true, features = ["api"] }
+# vsr-only: locate and decode the on-disk superblock slot files in the recovery
+# test (`SLOT_FILE_NAMES`, `decode_latest_payload`).
```
hubcio: the export is `decode_slots`, not `decode_latest_payload`.

core/shard/src/lib.rs
```diff
@@ -54,7 +54,9 @@ use metadata::impls::metadata::StreamsFrontend;
 use metadata::stm::StateMachine;
 use metadata::{BoundSession, MetadataSubmitError};
 use partitions::{IggyPartition, IggyPartitions, PollFragments, PollingArgs, PollingConsumer};
-use server_common::sharding::{IggyNamespace, PartitionLocation, ShardId};
+use server_common::sharding::{
+    IggyNamespace, METADATA_CONSENSUS_NAMESPACE, PartitionLocation, ShardId,
```
hubcio: `METADATA_CONSENSUS_NAMESPACE` is imported unconditionally but only used inside the `cfg(debug_assertions)` block at line 3061, so release builds warn. green only because `-D warnings` rides a clippy invocation, which builds debug.

hubcio (changes requested):
three things worth filing separately - all outside this diff's hunks, so they cannot be line comments:

- `core/metadata/src/impls/metadata.rs:988-1003` - the "journal already holds this op, re-forward and re-ack it" fast path compares nothing between the stored and the incoming header. it forwards the NEW bytes, keeps the OLD bytes locally, and acks the NEW op. `replicate_preflight` passes on equal view, `panic_if_hash_chain_would_break_in_same_view` is bypassed, and `commit_journal` then applies the stale body as committed with the vote counted toward quorum. `fence_old_prepare_by_commit` fences on `commit_min`, not `commit_max`, so it is narrower than it looks, and `send_prepare_ok`'s `persisted` check is the same op-only lookup, so the ack asserts durability of the new op while the WAL holds the old bytes. live on master today and independent of this PR, but materially more reachable once `log_view` is durable, since that is exactly the path that makes an un-truncated local suffix survive a restart into a live view. this diff supplies the fix for free - compare `checksum_body` (letting `CHECKSUM_BODY_UNSEALED` fall through) before re-acking.

- the metadata WAL has no uncommitted-suffix truncation primitive. that is the root cause of both the `log_view` comment on `bootstrap.rs:1962` and the re-ack path above, and it deserves its own issue.

- `core/consensus/src/observability.rs:347` - `SendStartView { op, commit, .. }` drops `incarnation` from the control-action record, so the sim oracle is blind to exactly the probe-reply-vs-unsolicited distinction the new guard rests on.

the core design holds up, for what it is worth: gate coverage is complete across every metadata emitter of SVC/DVC/StartView/PrepareOk, the three-phase ordering is right, lock discipline has no inversion, `view`/`log_view` are monotone so the durable value cannot regress, and a pre-change WAL still boots through the sentinel. placing `client_table` LAST in `MetadataSnapshot` is what makes the forward decode work at all - a mid-struct insert would have silently misassigned fields by position. every blocker above sits in the recovery half, not the gate.

numinnex (approved):

spetz (approved):

hubcio (approved):

PR #3696, opened 2026-07-18
feat(python): expose structured IggyError

mmmmxa:
## Which issue does this PR address?

Closes #3618 

## Rationale

Introduce `IggyError` class in Python SDK to fully represent Iggy-specific errors from Rust. (See issue for more info)

## What changed?

Python was exposing Iggy failures as plain RuntimeError or ValueError with only a message, while no code and snake case error name was provided.

The changes introduce the `IggyError` Python class that represent `IggyError` in Rust 1 to 1. Thus, it has code, message and the snake case error name, which gives more transparency in debugging and more information for clean exception handling.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT) - Claude
2. Scope of usage? (e.g., autocomplete, generated functions, entire implementation) - autocomplete, mass updates of some test cases, documentation generation.
3. How did you verify the generated code works correctly? - The generated code is either tested or a part of tests, which all pass. I have also performed multiple self-reviews.
4. Can you explain every line of the code if asked? - Yes.

PR #3776, opened 2026-07-29
feat(python): expose TCP client configuration

### LINKED ISSUE #3742: Python SDK: `IggyClient` exposes no client configuration — no reconnection or auto-login, so session recovery is unreachable (Go/C# already expose it)
### Description

The Python binding's `IggyClient` accepts only a connection string, so **no client
configuration can be set from Python** — in particular neither reconnection nor auto-login.
On `master`, `foreign/python/src/client.rs`:

```rust
#[new]
#[pyo3(signature = (conn=None))]
fn new(conn: Option<String>) -> PyResult<Self>

let client = IggyClientBuilder::new()
    .with_tcp()
    .with_server_address(conn.unwrap_or("127.0.0.1:8090".to_string()))
    .build()
```

The generated stub confirms the surface: the only exported types are `AutoCommit`,
`AutoCommitAfter`, `AutoCommitWhen`, `IggyClient`, `IggyConsumer`, `PollingStrategy`,
`ReceiveMessage`, `SendMessage`, `StreamDetails`, `TopicDetails`. There is no config or
builder type.

### Why this matters — the SDK's own session recovery becomes unreachable from Python

`#2880` added `Unauthenticated`, `Disconnected`, `EmptyResponse` and `StaleClient` to the
reconnect match in `send_raw_with_response`, so the client reconnects **and replays the
stored login**. That recovery is gated on the client having `reconnection` enabled and an
`AutoLogin` credential to replay — and `#3651` documents the other half:

> *"`login_user` / `logout_user` … never touch the state. The only writer of the session
> values is `connect()`'s auto-login block, so a manual `connect()` + `login_user()` flow
> stays at `Connected`"*

A Python user has **only** the manual `connect()` + `login_user()` flow available. So the
combination is: no auto-login to replay, and a flow that does not own session state. The
practical result is that a session lost shortly after login is not recovered — it surfaces as
`Unauthenticated` on the *next* call, and the application must handle it by hand.

### Observed impact

A long-running Python consumer on a restart loop, against `iggy-server` 0.8.0 over TCP
loopback. `login_user()` returned successfully; the very next call raised:

```
File ".../core/iggy_bus.py", line 229, in ensure_topology
    if await self._client.get_stream(self._stream) is None:
RuntimeError: Unauthenticated
```

Ten seconds later the same startup failed differently, then succeeded on the third attempt
and ran for ~2h48m:

```
File ".../core/iggy_bus.py", line 206, in connect
    await self._client.login_user(...)
RuntimeError: Disconnected
```

Both are errors `#2880` already classifies as reconnectable — but with no reconnection/
auto-login configurable from Python, they terminate the process instead, leaving restart
supervision (systemd) as the only recovery.

### Workaround a Python caller has to hand-roll today

Because `login_user()` returning is not sufficient evidence of an authenticated session, we
now re-verify it explicitly and re-login on failure — effectively reimplementing `auto_login`
plus reconnection in Python:

```python
for attempt in range(MAX_ATTEMPTS):
    try:
        await client.connect()
        await client.login_user(username, password)
        await client.get_stream(stream)   # probe: raises Unauthenticated if the session is dead
    except Exception:
        await sleep_with_backoff(attempt)  # then retry the whole connect+login
    else:
        break
```

### Parity — other foreign SDKs already expose this

- Go: `foreign/go/client/tcp/tcp_core.go` — `autoLogin AutoLogin`
- C#: `foreign/csharp/Iggy_SDK/Configuration/AutoLoginSettings.cs`

Python is the outlier.

### Suggested direction

Everything needed already exists on the Rust side; this is about exposing it:

- `core/common/src/types/configuration/tcp_config/tcp_client_config.rs` — `TcpClientConfig`
- `.../tcp_client_config_builder.rs` — `with_auto_sign_in(AutoLogin)`
- `.../tcp_client_reconnection_config.rs` — `TcpClientReconnectionConfig`
- `core/common/src/types/configuration/auth_config/auto_login.rs` — `AutoLogin`
- `core/sdk/src/clients/client_builder.rs` — `IggyClientBuilder::with_auto_sign_in`

Minimal version: optional keyword arguments on `IggyClient(...)` for reconnection settings
and auto-login credentials, passed through to `IggyClientBuilder`. Fuller version: expose a
config/builder object mirroring `TcpClientConfigBuilder`, which would also give the Python
SDK somewhere to hang the QUIC/HTTP/WebSocket options from `#2835`.

Happy to open a PR if a maintainer indicates the preferred shape (kwargs vs. exposed builder).

### Affected area / component

Python SDK (`foreign/python`)

### Related

- `#3651` — SDK client state conflates transport and session dimensions (explains why the
  manual `connect()` + `login_user()` flow does not own session state). Related, not a
  duplicate: that issue is about the state enum being unreliable; this one is about the
  configuration surface being absent in Python.
- `#2880` — added the reconnect-and-replay-login behavior that this config would unlock.
- `#1453` — `with_auto_sign_in` for the Rust builders.
- `#2835` — Python SDK transports; a config object would be the natural home for those too.

### Versions

Server `0.8.0` (DockerHub `apache/iggy:0.8.0`), `apache-iggy` 0.8.0 (PyPI wheel, cp312,
manylinux), Python 3.12.13, TCP on loopback. Python constructor and builder call quoted above
were read on `master`, so the gap is present there too.

### Scope note

This issue is deliberately limited to the **missing configuration surface in the Python
binding**. Whether a session can still be lost immediately after a successful `login_user()`
*even with* auto-login and reconnection enabled is a separate Rust/server question that we
have not isolated, and it is not claimed here.

ethanlin01x:

Closes #3742

## Rationale

The Python binding accepts only a bare server address, so reconnection and auto-login cannot be configured from Python. The SDK's session recovery is therefore unreachable: a server restart surfaces as `Unauthenticated` on the next call.

## What changed?

`IggyClient(...)` took only `host:port`, with `AutoLogin::Disabled` hardcoded and the reconnection policy untunable.

It now also accepts a keyword-only `TcpConfig` mirroring the Rust `TcpClientConfig` (`auto_login`, `reconnection`, `heartbeat_interval`, TLS, `nodelay`). Unset fields fall back to the Rust defaults, and durations are validated `datetime.timedelta`. The bare-address constructor and `from_connection_string` are unchanged.

One behavior change: a negative `timedelta` on `create_topic` / `update_topic` (`message_expiry`), the `consumer(...)` intervals, or `AutoCommit.Interval(...)` became a near-`u64::MAX` duration and now raises `ValueError`.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? Claude
2. Scope of usage? help generate and review this PR
3. How did you verify the generated code works correctly?  Integration tests against a real server prove credentials are replayed on connect without a manual `login_user()`.


examples/python/client-configuration/main.py
slbotbm, line 1: There's no need for an extra example for the new config. Let's fold this into the getting-started/ example

examples/python/README.md
slbotbm, line 1: No need for this extra change if the above example gets folded in getting-started.

foreign/python/README.md
```diff
@@ -58,6 +58,44 @@ maturin develop
...
+    await client.connect()
+
+
+asyncio.run(main())
+```
+
+`TcpConfig` also carries `tls_enabled`, `tls_domain`, `tls_ca_file`,
+`tls_validate_certificate` and `nodelay`. Every field is keyword-only and defaults to the
+same value the Rust SDK uses. `IggyClient.from_connection_string(...)` remains available
+for the same settings in string form.
```
slbotbm: You could show these as commented out options in the above config instead of mentioning them here. No need to mention anything about `from_connection_string` here.

foreign/python/README.md
```diff
@@ -58,6 +58,44 @@ maturin develop
 pytest tests/ -v # Run tests (requires iggy-server running)
 ```
 
+## Client Configuration
+
+`IggyClient` takes either a server address or a `TcpConfig`. Configuring `auto_login`
+lets the SDK replay the credentials whenever it reconnects, so a session dropped by a
+server restart is recovered instead of surfacing as `Unauthenticated`:
```
slbotbm: No need to mention the `auto_login` option specially. That is for the API comment docs.

slbotbm (changes requested):
Looks good. Mostly cosmetic changes. One thing though: in the docs, you are declaring the thrown errors as `PyValueError` and similar types. These are rust types which the python user will not see. Also, there are references to the rust sdk in public docs. Please remove them. Our modelled users are python users, who would not know anything about rust.

foreign/python/src/duration.rs
```diff
@@ -0,0 +1,52 @@
...
+    py: Python<'_>,
+    duration: IggyDuration,
+) -> PyResult<Bound<'_, PyDelta>> {
+    let micros = duration.as_micros();
+    let total_seconds = micros / 1_000_000;
+    let days = i32::try_from(total_seconds / 86_400).map_err(|_| {
+        PyErr::new::<pyo3::exceptions::PyOverflowError, _>(
+            "duration does not fit into a datetime.timedelta",
+        )
+    })?;
```
numinnex: `IggyDuration::as_micros()` is a truncating cast (`core/common/src/utils/duration.rs:94`):

```rust
pub fn as_micros(&self) -> u64 {
    self.duration.as_micros() as u64
}
```

`py_delta_to_iggy_duration` accepts `timedelta(days=999_999_999)` — 8.64e13 seconds, which fits a `Duration` — but that is ~8.64e19 µs against a `u64::MAX` of ~1.84e19, so the value wraps here and the getter returns a wrong duration rather than raising.

That also leaves the `PyOverflowError` below unreachable: a post-wrap value caps at ~213,503 days, well under `i32::MAX`, so the `try_from` can never fail.

Reading the u128 directly fixes both:

```rust
let micros = duration.get_duration().as_micros();
```

The existing `try_from` guards start doing real work once the input isn't pre-wrapped.

foreign/python/tests/test_client_config.py
```diff
@@ -0,0 +1,303 @@
...
+        [timedelta(microseconds=-1), timedelta(seconds=-1), timedelta(days=-1)],
+    )
+    def test_negative_duration_is_rejected(
+        self,
+        construct: Callable[[timedelta], TcpReconnectionConfig],
+        negative: timedelta,
+    ):
+        """Test that a negative duration fails at construction, not at connect."""
+        with pytest.raises(ValueError, match="negative"):
+            construct(negative)
```
numinnex, line 128: The negative-duration rejection also changes methods that already shipped: `create_topic` / `update_topic` (`message_expiry`) and `consumer(poll_interval=…, polling_retry_interval=…, init_retry_interval=…)`, plus `AutoCommit.Interval(…)`. Those previously accepted a negative `timedelta` and produced a near-`u64::MAX` duration; they now raise `ValueError`.

That is the right fix, but every negative-duration case here targets the three new classes, so the pre-existing surface has no coverage of the new behavior. Worth one case on it, e.g. `create_topic(..., message_expiry=timedelta(seconds=-1))` raising `ValueError`.

Also worth a line in the PR description: it currently says the conversion is fixed, but not that previously-accepted input now raises.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,379 @@
...
+    ///     server_address: `host:port` of the Iggy server. Defaults to `127.0.0.1:8090`.
+    ///     auto_login: Credentials replayed on every connect. Defaults to `AutoLogin.disabled()`.
+    ///     reconnection: Reconnection policy. Defaults to `TcpReconnectionConfig()`.
+    ///     heartbeat_interval: Interval of heartbeats sent by the client. Defaults to 5 seconds.
+    ///     tls_enabled: Whether to connect over TLS. Defaults to disabled.
+    ///     tls_domain: Domain to validate the certificate against. Empty means it is
+    ///         taken from `server_address`.
+    ///     tls_ca_file: Path to the CA file for TLS.
+    ///     tls_validate_certificate: Whether to validate the server certificate.
+    ///         Defaults to validating.
```
numinnex: Mirroring `TcpClientConfigBuilder::with_tls_validate_certificate` is in scope, and the connection-string path hardcoding `true` (`core/common/src/types/configuration/tcp_config/tcp_client_config.rs:72-73`) guards a different case — strings arriving from config files or user input, not a config built in code. So no objection to exposing it.

The docstring is neutral for a flag that accepts any certificate the server presents, though. One sentence on the cost would help, e.g. "Disabling this accepts any certificate the server presents, including self-signed and mismatched ones; intended for local development only."

examples/python/README.md
ethanlin01x: Reverted in 4fb0ff33a along with the example.

examples/python/client-configuration/main.py
ethanlin01x: Folded in 4fb0ff33a. Once the Python SDK matures we may want a more complete example again.

foreign/python/README.md
ethanlin01x: Addressed in 12da3bc0d

foreign/python/README.md
ethanlin01x: Dropped in 12da3bc0d

ethanlin01x:
Hi @slbotbm 
Thanks for the review. Renamed to the Python exception names in 1e5561dbe, and 42ad9ab63 clears the remaining Rust the module docstring.  The RuntimeError references will need another pass once #3696 lands and Iggy failures become IggyError

foreign/python/src/duration.rs
ethanlin01x: Fixed in 92048cb8f by reading the std Duration directly, plus a timedelta(days=999_999_999) round-trip test pinning it.

foreign/python/tests/test_client_config.py
ethanlin01x: Added in dc180e5ac.  The PR description now states the behavior change.

foreign/python/src/config.rs
ethanlin01x: Added in ab97c0b8d

slbotbm:
@ethanlin01x could you resolve the conflicts?

/author

slbotbm (dismissed):

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+        // remaining fields are assigned directly so every unset argument falls
+        // back to the Rust `TcpClientConfig::default()` value instead of a
+        // literal duplicated here.
+        let mut inner = TcpClientConfigBuilder::new()
+            .with_server_address(server_address.unwrap_or(defaults.server_address))
+            .build()
+            .map_err(|e| PyErr::new::<pyo3::exceptions::PyValueError, _>(e.to_string()))?;
+        inner.auto_login = auto_login.inner.clone();
+        inner.reconnection = reconnection.inner.clone();
+        inner.heartbeat_interval = heartbeat_interval
```
hubcio: `heartbeat_interval=timedelta(0)` passes validation (only negatives are rejected) and lands in the sdk heartbeat loop at `core/sdk/src/clients/client.rs:269`, which does `sleep(heartbeat_interval.get_duration())` in an unbounded loop - zero means pinging as fast as round trips complete, for as long as the client lives. nothing downstream reads zero as "disabled". worth rejecting zero here with the same `ValueError` shape as negatives.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+        interval: Option<Py<PyDelta>>,
+        #[gen_stub(override_type(type_repr = "datetime.timedelta | None", imports=("datetime")))]
+        reestablish_after: Option<Py<PyDelta>>,
+    ) -> PyResult<Self> {
+        let defaults = RustTcpClientReconnectionConfig::default();
+        Ok(Self {
+            inner: RustTcpClientReconnectionConfig {
+                enabled: enabled.unwrap_or(defaults.enabled),
+                max_retries,
+                interval: interval
```
hubcio: `interval=timedelta(0)` combined with the default `max_retries=None` (unlimited) turns the reconnect loop at `core/sdk/src/tcp/tcp_client.rs:434-441` into a tight `TcpStream::connect` spin with one `info!` line per attempt while the server is down. zero with bounded retries is a legitimate fast-retry policy, so rejecting just the zero + unlimited combination is enough.

foreign/python/tests/test_client_config.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def test_zero_interval_is_allowed(self):
```
hubcio: this asserts the zero-interval value from the reconnect-spin problem is legal, so the test will fight the fix later. `reestablish_after` is the one duration where zero is genuinely meaningful (skips the cooldown - the guard at `tcp_client.rs:394` is `if elapsed < interval`), so retargeting the test there keeps the coverage.

foreign/python/src/client.rs
```diff
@@ -1045,7 +1069,7 @@ impl IggyClient {
         {
             builder = builder.init_retries(
                 init_retries,
-                py_delta_to_iggy_duration(&init_retry_interval),
+                py_delta_to_iggy_duration(&init_retry_interval)?,
```
hubcio: `init_retry_interval=timedelta(0)` reaches `time::interval(interval.get_duration())` at `core/sdk/src/clients/consumer.rs:323`, and tokio asserts `period must be non-zero` - a rust panic that surfaces in python as `pyo3_async_runtimes.RustPanic`, naming neither the argument nor the class. the interval is constructed unconditionally, so it fires every time, even when the stream and topic already exist. since this line now validates the sign, rejecting zero here too is one line.

foreign/python/src/client.rs
```diff
@@ -1020,16 +1044,16 @@ impl IggyClient {
...
         if let Some(poll_interval) = poll_interval {
-            builder = builder.poll_interval(py_delta_to_iggy_duration(&poll_interval))
+            builder = builder.poll_interval(py_delta_to_iggy_duration(&poll_interval)?)
         } else {
             builder = builder.without_poll_interval()
         };
         if let Some(polling_retry_interval) = polling_retry_interval {
             builder =
-                builder.polling_retry_interval(py_delta_to_iggy_duration(&polling_retry_interval))
+                builder.polling_retry_interval(py_delta_to_iggy_duration(&polling_retry_interval)?)
```
hubcio: same zero hole: `polling_retry_interval=timedelta(0)` becomes the retry sleep in `core/sdk/src/clients/consumer.rs:690-699` (`while !can_poll || ... { trace!(); sleep(...) }`) - no syscall in the loop body, so it burns a core, and with `auto_join_consumer_group=False` the join flag never flips and the spin is permanent. the field is renamed mid-chain to `reconnection_retry_interval` (`consumer_builder.rs:247`), in case you grep for it.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+    ///
+    /// Args:
+    ///     server_address: `host:port` of the Iggy server. Defaults to `127.0.0.1:8090`.
+    ///     auto_login: Credentials replayed on every connect. Defaults to `AutoLogin.disabled()`.
+    ///     reconnection: Reconnection policy. Defaults to `TcpReconnectionConfig()`.
+    ///     heartbeat_interval: Interval of heartbeats sent by the client. Defaults to 5 seconds.
+    ///     tls_enabled: Whether to connect over TLS. Defaults to disabled.
+    ///     tls_domain: Domain to validate the certificate against. Empty means it is
+    ///         taken from `server_address`.
+    ///     tls_ca_file: Path to the CA file for TLS.
```
hubcio: `tls_ca_file` is silently ignored when `tls_validate_certificate=False` - `tcp_client.rs:475` only reads the ca inside the validating branch, the other branch installs a blanket no-verification resolver. this config is the first python surface that can express that combination (connection strings hardcode validation on), and `TcpConfig(tls_enabled=True, tls_ca_file="ca.pem", tls_validate_certificate=False)` reads as "pin to my ca" while accepting any certificate. same silent ignore when `tls_enabled=False`, which both examples hit (they pass `tls_ca_file` regardless of `--tls`). document the precedence here, or reject the contradictory combo.

foreign/python/src/client.rs
```diff
@@ -438,7 +462,7 @@ impl IggyClient {
         };
 
         let expiry = match message_expiry {
-            Some(delta) => IggyExpiry::ExpireDuration(py_delta_to_iggy_duration(&delta)),
+            Some(delta) => IggyExpiry::ExpireDuration(py_delta_to_iggy_duration(&delta)?),
```
hubcio: behavior change worth documenting: a negative `message_expiry` used to silently become a near-`u64::MAX` expiry (the old converter did i32 math and cast to u64), now it raises `ValueError` at call time, before the awaitable is created. the docstrings of `create_topic`, `update_topic` and `consumer_group` still list `RuntimeError` only - the new `TcpConfig`/`TcpReconnectionConfig` docstrings document the `ValueError`, these three should too. for `consumer_group` that covers four timedelta params plus the `AutoCommit` interval variants.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+#[gen_stub_pymethods]
+#[pymethods]
+impl TcpReconnectionConfig {
+    /// Constructs a reconnection policy.
+    ///
+    /// Args:
+    ///     enabled: Whether to reconnect at all. Defaults to enabled.
+    ///     max_retries: Attempts before giving up, or `None` for unlimited.
```
hubcio, line 124: the docstring doesn't say that `None` is the default, and that with the server down an awaited call then never returns - the retry loop sits inside `connect()` (`tcp_client.rs:364`), which the send path re-enters, so `connect()`/`send_messages`/`poll_messages` all block indefinitely. both getting-started examples ship exactly this policy (reconnection configured, `max_retries` unset). worth one sentence here, plus suggesting a finite `max_retries` for request/reply style usage.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+    fn tls_validate_certificate(&self) -> bool {
+        self.inner.tls_validate_certificate
+    }
+
+    #[getter]
+    fn nodelay(&self) -> bool {
+        self.inner.nodelay
+    }
+
+    fn __repr__(&self) -> String {
```
hubcio, line 392: repr prints 5 of the 9 fields and drops exactly the ones you'd debug a tls handshake with: `tls_domain`, `tls_ca_file`, `tls_validate_certificate`, `nodelay`. a config with certificate validation off prints identically to a validating one, and that's the field whose own docstring says "local development only". also `heartbeat_interval=5s` isn't valid python, so the constructor-shaped output can't be pasted back (`TcpReconnectionConfig.__repr__` has the same `interval=1s` issue). no test pins the repr, so this is free to fix.

foreign/python/tests/test_client_config.py
```diff
@@ -0,0 +1,325 @@
...
+
+        assert reconnection.interval == timedelta(0)
+
+    def test_very_long_interval_round_trips(self):
+        """Test that an interval beyond 68 years survives the i32 boundary."""
+        reconnection = TcpReconnectionConfig(interval=timedelta(days=30_000))
+
+        assert reconnection.interval == timedelta(days=30_000)
+
+    def test_maximum_interval_round_trips(self):
```
hubcio, line 169: the docstring claims the value "survives the u64-microsecond boundary", but the test only proves the getter round-trips - the getter reads `get_duration().as_micros()` as u128. nothing on the `interval` path ever narrows to u64 (`tcp_client.rs:440` sleeps on the full `Duration`), so the boundary named here isn't exercised and doesn't exist for this field. the sibling `reestablish_after` is the one consumed via the truncating `as_micros()` cast (`tcp_client.rs:389`), where `timedelta(days=999_999_999)` actually wraps. suggest rewording to the i32-days boundary in the getter conversion, which is what this covers (the test above at line 127 already words it right).

foreign/python/tests/test_client_config.py
```diff
@@ -0,0 +1,325 @@
...
+
+        client = IggyClient(TcpConfig(server_address=f"{host}:{port}"))
+        await client.connect()
+        await wait_for_ping(client)
+
+        with pytest.raises(RuntimeError):
+            await client.create_stream(unique_name())
+
+    @pytest.mark.asyncio
+    async def test_config_and_connection_string_are_equivalent(self, unique_name):
```
hubcio: name promises equivalence, body proves both clients authenticate - the reconnection params configured on either side are never observed (`IggyClient` exposes no config getter, and `get_stream(...) is None` is the only assertion). rename to something like `test_config_and_connection_string_both_authenticate`.

foreign/python/src/client.rs
```diff
@@ -56,17 +57,41 @@ pub struct IggyClient {
...
+        >,
     ) -> PyResult<Self> {
+        let config = match conn {
+            Some(PyClientConfig::Config(config)) => config.client_config(),
+            Some(PyClientConfig::ServerAddress(server_address)) => Arc::new(
+                TcpClientConfigBuilder::new()
+                    .with_server_address(server_address)
+                    .build()
+                    .map_err(|e| {
+                        PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string())
```
hubcio, line 116: same invalid address now raises two different exception types: `IggyClient("bad")` keeps the historical `RuntimeError`, `TcpConfig(server_address="bad")` raises `ValueError` - and neither is a subclass of the other, so one `except` can't cover "bad address". keeping the str path as-is is right (changing it would break existing `except RuntimeError` handlers), but the `conn` docstring should name both, since which one you get depends on which form you passed.

foreign/python/Cargo.toml
```diff
@@ -45,4 +45,5 @@ pyo3-async-runtimes = { version = "0.29.0", features = [
     "tokio-runtime",
 ] }
 pyo3-stub-gen = "0.23.0"
+secrecy = "0.10"
```
hubcio, line 50: the committed `Cargo.lock` wasn't regenerated - the `apache-iggy` dependency list in it has no `secrecy` edge, so `cargo metadata --locked` fails in this directory. to be fair it was already stale before this PR (`paste` from #3613 is missing too) and nothing in CI passes `--locked` for this crate, so nothing breaks visibly - but a `cargo check` here re-syncs it and picks up both.

foreign/python/src/duration.rs
```diff
@@ -0,0 +1,55 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use iggy::prelude::IggyDuration;
+use pyo3::prelude::*;
+use pyo3::types::{PyDelta, PyDeltaAccess};
+use std::time::Duration;
+
+pub fn py_delta_to_iggy_duration(delta: &Py<PyDelta>) -> PyResult<IggyDuration> {
```
hubcio, line 24: pyo3 0.29 already ships both directions of this conversion: `FromPyObject`/`IntoPyObject` for `std::time::Duration` do the same days/seconds/micros arithmetic, reject negative timedeltas with a `ValueError` whose message contains "negative" (so the `match="negative"` tests keep passing), and pyo3-stub-gen maps `Duration` to `datetime.timedelta` natively. taking `Option<Duration>` in the constructors and converting with `IggyDuration::from` deletes this whole file. the `Option<...>` parameter stub overrides have to stay either way (stub-gen renders bare `Option<T>` as `typing.Optional[...]`). side note: the overflow branch below is unreachable - python caps `timedelta` at 999,999,999 days, which fits i32, and the rust-side defaults are a few seconds.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+}
+
+/// Configuration for the TCP transport, accepted by `IggyClient(...)`.
+///
+/// Every field is keyword-only and optional.
+#[gen_stub_pyclass]
+#[pyclass(from_py_object)]
+#[derive(Clone)]
+pub struct TcpConfig {
+    auto_login: AutoLogin,
```
hubcio: `auto_login` and `reconnection` are stored twice - as wrapper fields here and inside `inner` (both assigned in `__new__`). nothing can make them diverge (no setters), but it's two copies of one truth a reader has to prove safe, and it keeps an extra live copy of the password (`SecretString`) per config object. both `inner` fields are `pub`, so the getters can rebuild the wrappers on demand (`AutoLogin { inner: self.inner.auto_login.clone() }`), which also makes `impl Default for AutoLogin` and the `Default` derive on `TcpReconnectionConfig` dead.

foreign/python/src/config.rs
```diff
@@ -0,0 +1,378 @@
...
+        #[gen_stub(override_type(type_repr = "datetime.timedelta | None", imports=("datetime")))]
+        heartbeat_interval: Option<Py<PyDelta>>,
+        #[gen_stub(override_type(type_repr = "builtins.bool | None"))] tls_enabled: Option<bool>,
+        #[gen_stub(override_type(type_repr = "builtins.str | None"))] tls_domain: Option<String>,
+        #[gen_stub(override_type(type_repr = "builtins.str | None"))] tls_ca_file: Option<String>,
+        #[gen_stub(override_type(type_repr = "builtins.bool | None"))]
+        tls_validate_certificate: Option<bool>,
+        #[gen_stub(override_type(type_repr = "builtins.bool | None"))] nodelay: Option<bool>,
+    ) -> PyResult<Self> {
+        let defaults = RustTcpClientConfig::default();
```
hubcio: `TcpClientConfigBuilder` is `#[derive(Default)]` over `TcpClientConfig`, so the config coming out of `build()` already carries every default - this `defaults` is a second identical construction and the `unwrap_or(defaults.x)` tails re-apply values that are already there. assigning only on `Some` (or building one exhaustive struct literal sourcing untouched fields from the builder output) drops ~10 lines. two constraints if you do: keep the trimmed `server_address` from the builder output (`build()` trims it), and leave `max_retries` as the raw pass-through it is - `None` there is a real user value meaning unlimited, not "unset".

foreign/python/src/client.rs
```diff
@@ -56,17 +57,41 @@ pub struct IggyClient {
...
+                    })?,
+            ),
+            None => Arc::new(TcpClientConfig::default()),
+        };
+        let tcp_client = TcpClient::create(config)
+            .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
         let client = IggyClientBuilder::new()
```
hubcio: `IggyClientBuilder::new().with_client(..).build()` plus the `map_err` guards an error that can't happen - `build()` only fails when no client was set. `RustIggyClient::new(ClientWrapper::Tcp(tcp_client))` is public, infallible and identical (partitioner/encryptor default to `None`), so this collapses to one line without the dead error path.

examples/python/getting-started/consumer.py
```diff
@@ -91,34 +99,30 @@ def parse_args() -> ArgNamespace:
...
+            interval=timedelta(seconds=1),
+        ),
+        tls_enabled=args.tls,
+        tls_ca_file=args.tls_ca_file or None,
+    )
 
 
 async def main():
     args: ArgNamespace = parse_args()
+    config = build_config(args)
```
hubcio: `build_config()` can now raise `ValueError` (address validation moved to `TcpConfig`) and it sits outside the `try` below - `--tcp-server-address 127.0.0.1` (no port) passes the argparse url check and produces a raw traceback. same in producer.py, which has no try at all.

foreign/python/README.md
```diff
@@ -58,6 +58,42 @@ maturin develop
...
+            reconnection=TcpReconnectionConfig(
+                enabled=True,
+                max_retries=10,
+                interval=timedelta(seconds=2),
+                reestablish_after=timedelta(seconds=30),
+            ),
+            heartbeat_interval=timedelta(seconds=5),
+            # tls_enabled=True,
+            # tls_domain="localhost",
+            # tls_ca_file="core/certs/iggy_ca_cert.pem",
```
hubcio: this path is relative to the repo root, but a reader of this readme runs from `foreign/python` - the same file is `../../core/certs/iggy_ca_cert.pem` from a sibling dir (that's what the examples readme uses).

hubcio (changes requested):
two things that don't fit on diff lines:

- pre-existing, separate PR: the six sync methods on `IggyConsumer` (`consumer.rs:58-92`) call `blocking_lock()` while holding the GIL; `consume_messages` holds the same mutex for its whole run, so calling e.g. `consumer.name()` during consumption deadlocks the interpreter (or panics if called from a sync callback body). untouched by this PR, just surfaced while tracing it.
- follow-up issue material: the zero-duration hazards below aren't python-specific. `TcpClient::create` is the one choke point all `TcpClientConfig` construction sites funnel through (only 2 of 10 go via the builder), and e.g. `--tcp-heartbeat-interval none` on the CLI already maps to zero today (`IggyDuration::from_str` treats `0`/`none`/`disabled`/`unlimited` the same). a fence there closes every surface at once.

foreign/python/src/config.rs
ethanlin01x: Fixed in acd01f68e — zero now raises the same ValueError as a negative value.

foreign/python/src/config.rs
ethanlin01x: Fixed in acd01f68e. Rejects only the zero + unlimited combination.

foreign/python/tests/test_client_config.py
ethanlin01x: Retargeted to `reestablish_after` in acd01f68e, as suggested. Split into three: zero `reestablish_after`, zero `interval` with bounded retries, and zero `interval` with reconnection disabled — the three places zero is meaningful.

foreign/python/src/client.rs
ethanlin01x: Fixed in acd01f68e — rejected on the same line that validates the sign.

foreign/python/src/client.rs
ethanlin01x: Fixed in acd01f68e.

foreign/python/src/config.rs
ethanlin01x: Documented the precedence in 54e80df41 rather than rejecting the combination, to keep the binding from being stricter than the Rust SDK. `tls_ca_file` now says it is read only when `tls_enabled` and `tls_validate_certificate` are both on, and `tls_validate_certificate` says it takes precedence.

foreign/python/src/client.rs
ethanlin01x: Partly superseded: #3623 landed on master while this was in review, so `message_expiry` is now an `IggyExpiry` object rather than a timedelta and documents its own ValueError. `consumer_group` still needed it and has it in 54e80df41, covering the intervals and the AutoCommit variants.

foreign/python/src/config.rs
ethanlin01x: Documented in 54e80df41, including that an awaited call never returns while the server is down, and a suggestion to set a finite value for request/reply usage.

foreign/python/src/config.rs
ethanlin01x: Fixed in 72025b394 — all nine fields, and durations print as `datetime.timedelta(seconds=N)`. Both reprs now parse as Python; there is a test pinning it.

foreign/python/tests/test_client_config.py
ethanlin01x: Reworded in 7ae1f0058

foreign/python/tests/test_client_config.py
ethanlin01x: Renamed in 7ae1f0058.

foreign/python/src/client.rs
ethanlin01x: Named both in 54e80df41, under `conn` rather than `Raises:` — `TcpConfig("bad")` raises before `IggyClient` is ever called, so the constructor itself cannot raise it. Also dropped "or if the client cannot be built": that was the builder, gone in acd01f68e.

foreign/python/Cargo.toml
ethanlin01x: There is no committed Cargo.lock for this crate to regenerate — b6c1cffdb (#1896) deleted it and added `Cargo.lock` to foreign/python/.gitignore in the same commit. `git show HEAD:foreign/python/Cargo.lock` reports it is not in the tree. The stale local lock you hit is real, but nothing in the repo is out of date. Happy to commit one if the crate should be locking again.

foreign/python/src/duration.rs
ethanlin01x: Adopted in acd01f68e — the hand-rolled arithmetic and the unreachable overflow branch are gone. Kept a thin wrapper so the negative message stays ours: pyo3's names a Rust type, which slbotbm asked to avoid earlier in this PR.

foreign/python/src/config.rs
ethanlin01x: Fixed in acd01f68e — the wrapper fields are gone and the getters rebuild from `inner`. `impl Default for AutoLogin` and the `Default` derive on `TcpReconnectionConfig` went with them.

foreign/python/src/config.rs
ethanlin01x: Fixed in acd01f68e. The trimmed address and the raw `max_retries` pass-through are preserved.

foreign/python/src/client.rs
ethanlin01x: Fixed in acd01f68e.

examples/python/getting-started/consumer.py
ethanlin01x: Fixed in 72025b394 — both examples catch the ValueError and print the message instead of a traceback.

foreign/python/README.md
ethanlin01x: Fixed in 71e3936da.

ethanlin01x:
@hubcio 
Thanks for the thorough review — the pointers into the SDK internals taught me a lot.

On the two that didn't fit on diff lines, the `blocking_lock()` deadlock and the zero-duration fence at `TcpClient::create`: want me to open them as follow-up issues?

hubcio:
@ethanlin01x you can fix them without creation of issue, just mention that it was found in #3776.

hubcio (approved):

slbotbm (approved):

PR #3783, opened 2026-07-31
fix(server-ng): bind cluster TCP to [tcp].address, not the roster ip

mmodzelewski:
In cluster mode the TCP listener bound the roster entry's 'ip'
directly while ws/quic/http kept the interface from their own
'address' config. The roster ip is an advertised address, so a node
advertising 10.0.0.5 stopped answering on loopback, leaving every
co-located dialer (sidecars, health probes, on-host consumers) on
ECONNREFUSED, and an operator's explicit [tcp].address was silently
discarded.

TCP now goes through the same merge as the other transports: the
roster supplies the port, the transport config supplies the bind
interface. A bind that cannot serve the advertised ip still warns at
startup rather than being rebound, and the warning stays quiet for a
wildcard bind or a hostname roster entry, neither of which is a
misconfiguration.

### DIFF
--- a/core/server-ng/src/bootstrap.rs
+++ b/core/server-ng/src/bootstrap.rs
@@ -2329,7 +2329,7 @@ struct ClusterClientAddrs {
 
 fn resolve_cluster_client_addrs(
     self_node: &configs::ng_cluster::ClusterNodeConfig,
-    default_client_addr: SocketAddr,
+    default_tcp_addr: SocketAddr,
     default_ws_addr: Option<SocketAddr>,
     default_quic_addr: Option<SocketAddr>,
     default_http_addr: Option<SocketAddr>,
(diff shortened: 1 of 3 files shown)


core/server-ng/src/bootstrap.rs
```diff
@@ -2328,6 +2329,7 @@ struct ClusterClientAddrs {
 
 fn resolve_cluster_client_addrs(
     self_node: &configs::ng_cluster::ClusterNodeConfig,
+    default_client_addr: SocketAddr,
```
hubcio: default_tcp_addr to keep concistency with the rest

core/server-ng/src/bootstrap.rs
```diff
@@ -2371,25 +2374,48 @@ fn resolve_cluster_optional_addr(
...
+    port: u16,
+) -> SocketAddr {
+    let listen_addr = SocketAddr::new(bind_addr.ip(), port);
+    if roster_ip_unreachable_from_bind_addr(roster_ip, listen_addr) {
         warn!(
-            "{transport} listener binds {listen_addr} but the roster advertises {}:{port}; \
-             peers and clients dialing the advertised endpoint will not reach this node",
-            self_node.ip
+            "{transport} listener binds {listen_addr} but the roster advertises {roster_ip}:{port}; \
+             peers and clients dialing the advertised endpoint will not reach this node"
```
hubcio: will not -> may not

hubcio (changes requested):

hubcio (approved):

spetz (approved):

Issue #3764, opened 2026-07-28
feat(connectors): Iggy Source/Sink Connector: Native Cluster-to-Cluster Replication


kparisa:
### Description

**Summary**

Implement an Iggy Source/Sink Connector that consumes data from an upstream Iggy cluster and publishes it into a downstream Iggy cluster.

This enables seamless cluster-to-cluster replication and provides the foundation for multi-region deployments, disaster recovery, active-active architectures, edge-to-cloud streaming, and hierarchical streaming topologies.

Unlike traditional application-level replication, this connector allows operators to replicate selected streams, partitions, or namespaces without requiring producers or consumers to change.

**Motivation**

Many production deployments require data to exist in multiple locations.

Today, applications would need to publish to multiple Iggy clusters directly, increasing complexity and coupling replication logic into business applications.

A native Iggy Source/Sink Connector would make replication a first-class capability by continuously consuming from one cluster and publishing to another while preserving ordering and delivery guarantees.

             Producers
                 │
                 ▼
        +------------------+
        |  Iggy Cluster A  |
        |   (Primary)      |
        +------------------+
                 │
      Iggy Source/Sink Connector
                 │
                 ▼
        +------------------+
        |  Iggy Cluster B  |
        |  (Replica)       |
        +------------------+
                 │
          Local Consumers

The connector should behave like any other Iggy connector:

* Source = upstream Iggy
* Sink = downstream Iggy

Aspects to cover (ideal, not a comprehensive list, can be broken down into follow-up tasks after the basic version is in place):
* Iggy → Iggy replication
* configurable source streams
* configurable destination streams
* partition-aware replication
* offset checkpointing
* resumable replication
* configurable batching
* configurable compression
* TLS support
* authentication support
* metrics
* retry with backoff
* configurable parallelism

### Affected area / component

Connectors

### Contribution

- [ ] I'm willing to submit a pull request to implement this feature

### Good first issue

- [x] I think this could be a good first issue for a new contributor

raman20:
hey @kparisa i would like to work on this...

kparisa:
@raman20 Thank you! Assigned to you.
This will be a great feature!

You can implement it as a source or sink connector.

ryerraguntla:
@raman20 - When you raise PR, please raise separate PR for Sink and source. Please ping me on discord, I want to share a work in progress blog article for developing connectors for Iggy.

raman20:
iggy_sink pr is out.... 
[...]

PR #3784, opened 2026-07-31
feat(connectors): add iggy_sink connector plugin

raman20:

Relates to #3764

## Rationale

The Apache Iggy connectors subsystem needed a native `iggy_sink` plugin to enable streaming message
replication and forwarding between Iggy topics and external Iggy clusters.

## What changed?

The connectors runtime lacked a plugin to sink consumed messages into another Iggy stream or cluster.

This PR implements `iggy_sink` as a `cdylib` plugin loaded dynamically by the connectors runtime. It forwards
messages to a target Iggy endpoint, supports optional stream and topic overrides (`stream_id`, `topic_id`),
preserves message IDs for idempotent retries, and maintains partition-aware routing.

## Local Execution

- Passed
- Pre-commit hooks ran

Verified locally inside devcontainer:
- `cargo fmt --all --check`
- `cargo clippy -p iggy_connector_iggy_sink --all-targets -- -D warnings`
- `cargo test -p iggy_connector_iggy_sink`
- Live end-to-end message replication test in devcontainer with `iggy-server` and `iggy-connectors`.

## AI Usage

1. Which tools? Google Antigravity
2. Scope of usage? Learning data-flow and implementation of `iggy_sink` crate, unit tests, integration test suite, and
plugin documentation.
3. How did you verify the generated code works correctly? Verified with `cargo test`, zero-warning clippy
check, and live end-to-end message streaming replication inside devcontainer.

### DIFF
--- a/core/connectors/sinks/iggy_sink/README.md
+++ b/core/connectors/sinks/iggy_sink/README.md
@@ -1,0 +1,61 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Iggy Sink Connector
+
+The Iggy sink connector consumes messages from an upstream Iggy stream/topic and sinks them directly into a target downstream Iggy cluster, stream, or topic.
+
+## Features
+
+- **Iggy-to-Iggy Replication**: Replicate or forward streaming messages across topics or clusters.
+- **Dynamic Topic Overrides**: Optional stream and topic overrides; if omitted, messages retain their original stream and topic names.
+- **Header & Payload Preservation**: Forward message payloads and metadata seamlessly.
+- **Zero-Clone Hot Path**: Efficient batch forwarding without unnecessary memory copies.
+
+## Configuration
+
+```toml
+[[streams]]
+stream = "source_stream"
+topics = ["source_topic"]
+schema = "json"
+batch_length = 100
+poll_interval = "5ms"
+consumer_group = "iggy_sink"
+
+[plugin_config]
+server_address = "iggy://iggy:iggy@127.0.0.1:8090"
+stream_id = "target_stream"
+topic_id = "target_topic"
+```
+
+## Configuration Options
+
+| Option | Type | Default | Description |
+| ------ | ---- | ------- | ----------- |
+| `server_address` | string | required | Address of the target Iggy cluster (e.g. `iggy://user:pass@127.0.0.1:8090` or `127.0.0.1:8090`) |
+| `stream_id` | string (optional) | `None` | Optional target stream override. If omitted, uses the source stream name. |
+| `topic_id` | string (optional) | `None` | Optional target topic override. If omitted, uses the source topic name. |
+
+## Production Guarantees
+
+- **Transient Failures:** Underneath, the Iggy Rust SDK client retries connection attempts and reconnects automatically.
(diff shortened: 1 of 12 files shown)


core/connectors/sinks/iggy_sink/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+        let stream_id: Identifier = stream
+            .as_str()
+            .try_into()
+            .map_err(|_| Error::InvalidConfigValue(format!("Invalid stream: {stream}")))?;
+        let topic_id: Identifier = topic
+            .as_str()
+            .try_into()
+            .map_err(|_| Error::InvalidConfigValue(format!("Invalid topic: {topic}")))?;
+
+        let partitioning = Partitioning::partition_id(messages_metadata.partition_id);
```
krishvishal, line 109: Multi-partition source topics can have their messages silently misrouted:  (Partitioning::partition_id(messages_metadata.partition_id)) forwards the last message's source partition ID as the routing key for the entire flushed batch. Traced in runtime/src/sink.rs's consume_messages: partition_id is reassigned every loop iteration from whatever partition the last poll returned, and a single sink consumer (one consumer-group member) round-robins across all partitions assigned to it. A batch can end up containing messages from partitions 0 and 1, tagged and routed entirely under partition 1's ID. This label-only field is harmless for every other sink (none of them route by it); iggy_sink is the first
  to turn it into an actual write-path decision. 

Reachability: any source topic with >1 partition (the norm), not exercised by this PR's own
  single-partition example/test config. Fix: group ConsumedMessages by originating partition before building the target batch (one send_messages call per source partition) — sink-local, reduces batching efficiency slightly when a poll straddles
  partitions. A structural fix (guarantee single-partition batches at the runtime level) affects every sink and needs separate sign-off.

core/connectors/sinks/iggy_sink/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+    ConsumedMessage, Error, MessagesMetadata, Payload, Sink, TopicMetadata, sink_connector,
+};
+use serde::{Deserialize, Serialize};
+use tracing::{debug, info};
+
+sink_connector!(IggySink);
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct IggySinkConfig {
+    pub server_address: String,
```
krishvishal, line 34: Credentials logged in plaintext; not `SecretString`. logged unredacted at :59-62 `(info!("Opened iggy_sink with ID: {}, connecting to: {}", self.id,  self.config.server_address))`. Every config shipped in this PR embeds credentials (iggy://iggy:iggy@127.0.0.1:8090).

core/connectors/sinks/iggy_sink/README.md
```diff
@@ -0,0 +1,61 @@
...
+| Option | Type | Default | Description |
+| ------ | ---- | ------- | ----------- |
+| `server_address` | string | required | Address of the target Iggy cluster (e.g. `iggy://user:pass@127.0.0.1:8090` or `127.0.0.1:8090`) |
+| `stream_id` | string (optional) | `None` | Optional target stream override. If omitted, uses the source stream name. |
+| `topic_id` | string (optional) | `None` | Optional target topic override. If omitted, uses the source topic name. |
+
+## Production Guarantees
+
+- **Transient Failures:** Underneath, the Iggy Rust SDK client retries connection attempts and reconnects automatically.
+- **Delivery Semantics:** At-least-once delivery; messages are flushed per batch.
```
krishvishal, line 59: This claims "at-least-once" semantics, with no retry logic to back it up. Fix: correct the README to reflect the correct behavior of this connector.

core/connectors/sinks/iggy_sink/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        client
+            .send_messages(&stream_id, &topic_id, &partitioning, &mut iggy_messages)
+            .await
+            .map_err(|e| Error::Connection(e.to_string()))?;
+
+        Ok(())
+    }
+
+    async fn close(&mut self) -> Result<(), Error> {
```
krishvishal, line 145: `close()` never disconnects the client. This just logs and returns `Ok(())` without calling `client.disconnect()/shutdown()`. `IggyClient::connect()` spawns a background heartbeat task holding its own Arc clone that keeps the TCP connection alive independent of the outer handle `.shutdown()` (not `.disconnect()`) actually terminates it. Every close/restart leaks a connection + task; since the failure modes above push operators toward repeated manual restarts, this accumulates. Fix: call `.shutdown().await` in `close()`, matching the pattern already used in runtime/src/main.rs.

PR #3786, opened 2026-07-31
feat(server-ng): close the partition materialisation race at the owner

krishvishal:
## Problem

`create_topic` returns on metadata commit; the owning shard publishes the partition only after `build_partition_fresh` (mkdir, segment open, fallocate). A produce issued right after races materialisation, on every shard.

`partition_reconciler` carried a TODO for a materialisation barrier: stop non-owners seeding `shards_table` rows, have the owner push its row to peers, so row presence implies the partition exists.

Rejected. It turns a deterministic hash cache into a distributed readiness protocol whose deltas need ordering, retry and repair, and it gives up level-triggered repair (a pass re-derives every row from committed metadata, so a lost row is rewritten). The owner can decide all of it locally.

## Approach

Resolve at the owner; `shards_table` stays a hint.

* `router::route_typed` falls back to `calculate_shard_assignment` on a missing row, so an early frame reaches the future owner.
* `park_if_unmaterialised` holds it until the matching `InsertOwned`, then re-queues onto that shard's inbox.
* `serves_committed_incarnation` refuses a namespace whose committed `created_revision` disagrees with the row epoch, so a mid-teardown request cannot be acked against an incarnation about to be erased.

Tombstoned namespace, park overflow, and a namespace the shard gave up building all reply with a retriable status. No silence, so no lockstep transport waits out its read timeout.


core/server-ng/src/partition_reconciler.rs
```diff
@@ -512,6 +589,78 @@ async fn reconcile_additions(
...
+fn reconcile_parked_frames(ctx: &ReconcilerCtx, counters: &mut PassCounters) {
+    let parked = ctx.shard.parked_namespaces();
+    if parked.is_empty() {
+        return;
+    }
+    let partitions = ctx.shard.plane.partitions();
+    let total_shards = u32::from(ctx.total_shards);
+    let now = Instant::now();
+    for ns in parked {
+        if partitions.contains(&ns) {
```
numinnex: A frame re-parked after a refused re-dispatch is stranded for the process lifetime once its namespace is materialised, and this `continue` is where the last exit closes.

The sequence:

1. Frames park while the partition is absent, then its `InsertOwned` applies and `redispatch_parked_frames` runs.
2. The shard's own inbox is full (the comment on `redispatch_parked_frames` itself notes that a pass materialising many namespaces at once can overrun it), so the frame is re-parked via `repark_partition_frames`.
3. The namespace is now in `partitions`, and nothing drains that entry again. The only `redispatch_parked_frames` call site is the `InsertOwned` apply, which cannot fire twice for a materialised namespace, and this sweep skips the namespace right here, so the frames are neither re-dispatched, nor aged, nor reclaimed.

The doc comment on `redispatch_parked_frames` says re-parking "keeps the frame for the next pass", but no pass looks at a parked namespace that is already materialised.

Effects beyond the stuck frame itself:

- The client waits out its full response read-timeout, which is the silence this PR sets out to eliminate.
- The re-parked bytes stay counted against the shard-wide park budget forever, so every other namespace starts shedding earlier than it should.
- `has_parked_partition_frames()` stays true, so the fast-skip never arms again and the reconciler runs a full diff every tick from then on.

The same skip also covers tombstoned namespaces, so frames caught this way survive into a later teardown and are only answered once its `ConfirmRemove` finally drains, which a wedged disk delete can postpone indefinitely.

A possible fix inside this loop: when `partitions.contains(&ns)` and frames are still parked, re-drive the re-dispatch with the epoch from the routing row (it is idempotent, and the stale-stamp check already guards mismatched incarnations), or at least age the frames so `MAX_PARKED_PASSES` bounds their residency again. The inbox-full re-park path currently has no test, so nothing pins whichever behaviour is intended.

numinnex (dismissed):

core/shard/src/lib.rs
```diff
@@ -1609,8 +1685,106 @@ enum ParkOutcome<H> {
...
+/// reaches 30s, which would put four passes at 120s -- four times the SDK's
+/// response read-timeout, so the client times out first and the bound stops
+/// being the thing that answers it. Commit-tick wakes collapse it the other way,
+/// to tens of milliseconds. It bounds residency in units the simulator's virtual
+/// clock governs; it is not a latency guarantee.
+///
+/// TODO(krishna): derive this from `reconcile_periodic_interval` and the SDK
+/// response timeout so the bound tracks the configured interval instead of
+/// assuming one.
+const MAX_PARKED_PASSES: u32 = 3;
```
hubcio, line 1834: `MAX_PARKED_PASSES` also ages out replicated prepares, and that is a new loss path: on master a parked frame was held until materialisation, so a lagging backup's prepare was eventually served. a prepare has no client (`deny_parked_client_request` returns false) and nothing recovers the op afterwards - `retransmit_targets` skips anything with `ok_quorum_received`, and the partition plane opens a repair session only in `on_start_view`. the module TODO covers shed/refused prepares but not aged ones.

passes are commit-driven (`MetadataCommitTick` fires per partition-shaping commit, and a non-empty park buffer defeats the fast-skip), so 4 passes elapse in milliseconds during a create burst - and the sweep ages every parked namespace on each pass while `staged_this_pass` exempts only the one built that pass, so unrelated commits burn a lagging namespace's budget. worst shape is replica_count=2: the primary self-acks to quorum alone, so every prepare the backup ages out is permanently lost and the backup stays behind `commit_max` until an unrelated view change.

prepares should not be aged at all - the byte budget already bounds residency.

core/shard/src/lib.rs
```diff
@@ -1841,23 +2331,74 @@ where
...
+        });
+        // The per-namespace byte cap is waived for the first frame, or a frame
+        // bigger than the cap could never park at all; see
+        // [`MAX_PARKED_BYTES_PER_NAMESPACE`]. The shard-wide budget is not
+        // waived, so residency is unchanged.
+        let over_namespace_budget = parked_len > 0
+            && namespace_bytes.saturating_add(frame_cost) > MAX_PARKED_BYTES_PER_NAMESPACE;
+        if parked_len >= MAX_PARKED_PER_NAMESPACE
+            || over_namespace_budget
+            || parked_bytes.saturating_add(frame_cost) > MAX_PARKED_BYTES
```
hubcio: the shard-wide check has no empty-buffer waiver, while the per-namespace check one line up does (`parked_len > 0`). so a partition frame with footprint over 16 MiB can never park, on any attempt - `message_bus.max_message_size` ships at 64 MiB and nothing narrows partition frames. for a client request that is a transient deny plus a retry; for a prepare it is destroyed at the `Overflow(_) => {}` arm, whose comment says retransmit redelivers - contradicted by this PR's own TODO in partition_reconciler.rs (retransmit skips quorum-acked ops; repair only opens in `on_start_view`).

the const's own TODO admits the >16 MiB gap, but shipping it turns a bound change into committed-data loss on a backup: on master the only cap was 128 frames, so an oversize prepare parked and was served. and the loss door does not even need an oversize frame - `MAX_PARKED_BYTES_PER_NAMESPACE` sheds the 4th ordinary 1 MiB frame (footprint 1052672 x 4 > 4 MiB) where master allowed 128.

cleanest fix: do not shed prepares on the byte budgets at all - allow a one-frame overshoot (worst case = budget + one max-size frame, which is what the TODO proposes anyway). shedding only requests makes the loss class structurally impossible instead of raising a ceiling.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -510,6 +599,113 @@ async fn reconcile_additions(
...
+/// `ConfirmRemove` and `RemoveRouted`. Neither can name a namespace that was
+/// never built: it is absent from `IggyPartitions` (so `reconcile_removals`
+/// sees no owned ghost) and absent from `shards_table` (the owner seeds a row
+/// only via `InsertOwned`, and emits `InsertRouted` only for namespaces it does
+/// NOT own). So without this sweep the frames are held for the process
+/// lifetime and every waiting client burns its full response read-timeout.
+///
+/// Immediate reclaim needs positive evidence that the build will not finish. Two
+/// signals carry it: `build_partition_fresh` failed (ENOSPC, EPERM) and is backed
+/// off -- the backoff clamps at 60s, well past the client's 30s read timeout, so
```
hubcio: this justification does not hold for the case that actually fires: `next_backoff(1)` is 1s (`BACKOFF_BASE`, shift 0) - 60s needs 7 consecutive failures. `record_failure` stamps attempts 0 -> 1 and the sweep's `now` postdates the stamp, so the reclaim below fires on the very first failed build. one transient ENOSPC destroys every parked prepare for the namespace while the rebuild succeeds a second later, and a reclaimed prepare is unrecoverable (no client to answer, retransmit skips quorum-acked ops).

reclaim only when the backoff actually exceeds the remaining age budget - or better, never reclaim prepares on local-convergence grounds.

core/shard/src/lib.rs
```diff
@@ -1781,29 +1958,340 @@ where
...
+                continue;
+            };
+            let passes = frame.passes;
+            let parked_epoch = frame.epoch;
+            let Err(error) = sender.try_send(ShardFrame::consensus(self.id, frame.message)) else {
+                continue;
+            };
+            self.metrics.record_frame_drop(
+                crate::metrics::frame_drop_variant::PARTITION,
+                crate::coordinator::classify_try_send_err(&error),
```
hubcio: two compounding issues in this loop. the drop is recorded before the `Full` vs `Disconnected` match, so a frame that gets re-parked (retained, not lost) still bumps `frame_drops_total{variant=partition,reason=full}` - the counter metrics.rs documents as the only stranded-backup signal, and the one the simulator's `assert_no_frame_drops` pins at zero. and the loop keeps calling `try_send` after the first `Full`, which is provably futile: no await in the loop, and the only consumer of `senders[self.id]` is the pump task executing this very call, so nothing can free a slot mid-loop - up to 127 wasted sends per call, re-driven by `retry_reparked_frames` on every pump iteration.

one edit fixes both: on `Full`, push the frame, `extend` with the rest of the iterator, `break`; record a drop only on the `Disconnected` arm. (the unreachable `let ShardFrame::Consensus .. else` destructure and the inline duplicate of `deny_parked_client_request` on the disconnected path can fold into the same edit.)

core/shard/src/lib.rs
```diff
@@ -1383,6 +1429,29 @@ where
...
+    /// Exists for out-of-crate tests: the paths that hand work back to the pump
+    /// (`stage_transient_deny`, the parked-frame re-dispatch) index
+    /// `senders[self.id]`, so without it they silently no-op and a test asserting
+    /// on them proves nothing. The caller must keep the paired receiver alive;
+    /// dropping it turns every `try_send` into `Disconnected`.
+    ///
+    /// # Panics
+    /// If `sender` is not tagged for this shard, which would route this shard's
+    /// own frames to a peer.
+    pub fn attach_self_sender(&mut self, sender: TaggedSender) {
```
hubcio: `senders = vec![sender]` puts the sender at index 0, but every consumer indexes `senders[self.id]` - so for any shard id > 0 the re-dispatch and `stage_transient_deny` silently no-op, which is the exact failure the doc says this prevents. it also breaks the `senders[i].shard_id() == i` invariant both ctors validate, `forward_metadata_submit` indexes `senders[0]` directly (metadata submits would land in shard id's inbox instead of shard 0's), and `without_inbox` pins `shard_count: 1` which this never updates. latent today since every caller passes 0, but a future multi-shard test would silently prove nothing.

sturdier shape: `attach_senders(&mut self, Vec<TaggedSender>)` run through the existing `validate_sender_ordering`, update `shard_count` to match, and gate it `#[cfg(any(test, feature = "simulator"))]` like `parked_frame_count`.

core/shard/src/lib.rs
```diff
@@ -1781,29 +1958,340 @@ where
...
+    /// Returns `true` when a reply was staged. Callers are synchronous
+    /// (`apply_reconcile_ops`, the reconciler sweep), so the deny rides the
+    /// pump's outbound lifecycle path instead of an inline bus send.
+    ///
+    /// A prepare gets no reply because there is no client on this node to send
+    /// one to, but "no reply" must not mean "no record": the primary only
+    /// retransmits an op that has not reached quorum, so a destroyed prepare on
+    /// a lagging backup is invisible loss. The `false` return makes callers
+    /// count it under `frame_drops_total{variant=partition,reason=park_dropped}`.
+    fn deny_parked_client_request(&self, frame: ParkedFrame) -> bool {
```
hubcio, line 2328: returns true for any request even when nothing was staged: `stage_transient_deny` returns `()` and silently bails when `senders[self.id]` is missing, so callers log the frame as answered and skip the `park_dropped` bump for a request that was destroyed unanswered - while the doc above says the false return is exactly what makes callers count it. same count-before-delivery pattern this PR carefully fixes in `deny_partition_request_transient` and in `stage_transient_deny`'s own counter. make `stage_transient_deny` return bool and propagate it.

core/integration/tests/cluster/multi_shard_partition_convergence.rs
```diff
@@ -0,0 +1,201 @@
...
+
+/// Delete + recreate reuses the freed slab keys, so the namespace is
+/// byte-identical across incarnations and only `created_revision` separates
+/// them. This pins the observable outcome across that transition on a
+/// multi-shard node: the recreated topic serves its own data and none of the
+/// dead incarnation's.
+///
+/// It does NOT exercise the incarnation fence. Each step here is a completed
+/// round trip, so the reconciler converges before the next one starts and
+/// `serves_committed_incarnation` never denies (verified: zero "unverified
```
hubcio: this parenthetical cannot have verified anything: the unverified-incarnation deny logs at `debug!` and the harness runs the server at the default info level, so the line cannot appear in the captured log regardless of whether denials happened.

the harness already has `ServerHandle::stdout_occurrences`, and all three degraded-park modes log at `warn` ("park buffer at capacity", "retiring parked partition frames", "rejecting parked partition frame from a prior incarnation") - asserting those are zero would pin the mechanism positively, and is the cheapest regression net for the park paths given the assertions here cannot tell park + redispatch apart from deny + SDK replay.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -2104,4 +2423,918 @@ mod tests {
...
+            "a non-empty park buffer must defeat the fast-skip so the sweep can run"
+        );
+    }
+
+    /// Same hole, reached the other way: the namespace stays committed but
+    /// `build_partition_fresh` keeps failing. The `FailureCause::Add` backoff
+    /// clamps at 60s, twice the client's read timeout, so holding the frames
+    /// cannot help - answer them and let the client re-issue.
+    #[compio::test]
+    async fn parked_frames_are_reclaimed_while_the_build_is_backed_off() {
```
hubcio: this test pins the first-failure reclaim as correct (see the comment on the sweep - `next_backoff(1)` is 1s, not the 60s the doc claims), and it uses the inbox-less `build_test_shard`, so it cannot observe whether the client was ever answered: `stage_transient_deny` bails at the missing sender while `deny_parked_client_request` still returns true. the sibling test `parked_frames_are_answered_once_they_outlive_their_admission_window` spells out this exact trap and uses `build_test_shard_with_inbox` + `drain_staged_client_sends`. also nothing in this branch covers a parked prepare.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -2104,4 +2423,918 @@ mod tests {
...
+        );
+    }
+
+    /// The frame cap bounds count, not residency: `Message::into_generic` is a
+    /// retag, so each parked entry keeps its whole buffer -- up to 64 MiB. With
+    /// only a frame cap, one namespace could pin 128 x 64 MiB and nothing capped
+    /// the namespace count. The shard-wide byte budget is what actually bounds
+    /// it, so large frames must shed well before the frame cap.
+    #[compio::test]
+    async fn park_byte_budget_sheds_large_frames_before_the_frame_cap() {
```
hubcio: this trips the per-namespace cap, not the shard-wide budget it is named and asserted for: all frames share one namespace, footprint per 1 MiB body is 1052672, so frame 4 sheds at 4210688 > `MAX_PARKED_BYTES_PER_NAMESPACE` (4 MiB) - nowhere near 16 MiB. both asserts stay green for the wrong reason (4 < 128). net effect: `MAX_PARKED_BYTES` is never the binding constraint in any test in the repo, and its only single-namespace-reachable behaviour is the first-frame-waiver path where the >16 MiB gap lives.

to actually bind the shard-wide budget: 5 namespaces x 3 frames all fit (15790080), then the first frame into a 6th namespace crosses 16 MiB while per-namespace-waived - unambiguously the shard-wide shed. and rename this to the cap it currently proves.

core/shard/src/lib.rs
```diff
@@ -1841,23 +2331,74 @@ where
...
         }
+        // Read the committed revision before taking the borrow below: the frame
+        // is stamped with the incarnation it was addressed to, so a later drain
+        // can tell it apart from a same-key replacement.
+        let epoch = self
+            .plane
+            .metadata()
+            .mux_stm
+            .streams()
+            .created_revision_for_namespace(namespace);
```
hubcio, line 2426: the stamp is not stable across re-entry: a re-dispatched frame still sitting in the inbox when a delete + recreate completes (`ConfirmRemove` removes and untombstones in the same arm, then the rebuild lands) re-parks here stamped with the new incarnation's revision - read fresh from committed metadata - and with `passes: 0`, so it is then served against the replacement. that is the exact write the stamp exists to block, and a third door beyond the two the module docs admit (`epoch: None`, and the wire-level prepare gap). narrow window (the full delete + recreate has to finish while the frame waits in the inbox), but worth carrying the original stamp across re-entry and making `passes` monotonic - or at least softening the module doc's absolute about stale drains.

core/shard/src/lib.rs
```diff
@@ -1781,29 +1958,340 @@ where
...
+            count = frames.len(),
+            epoch,
+            "re-dispatching parked partition frames after materialisation"
+        );
+        let mut refused_frames: Vec<ParkedFrame> = Vec::new();
+        for frame in frames {
+            // Only a stamp that exists and disagrees is evidence of a prior
+            // incarnation; see this function's docs on why `None` is not.
+            if let Some(parked_epoch) = frame.epoch
+                && parked_epoch != epoch
```
hubcio, line 2160: fence is direction-blind: park stamps the live committed revision, but `InsertOwned` carries the epoch snapshotted at build time, so a delete + recreate committing between staging and the pump apply gives the parked frame a stamp newer than `epoch` - and it gets rejected with the "prior incarnation" warn and counted on `partition_frames_rejected_stale_total`, whose doc says non-zero is a caught correctness anomaly, not routine churn. false positive for anyone alerting on it.

note the tempting one-liner `parked_epoch < epoch` is worse: a newer-stamped prepare would then be applied into the old incarnation's partition, which the staleness teardown erases along with its disk data. keep rejecting both directions - just split the log line and the label by direction so the anomaly counter keeps its contract.

core/shard/src/lib.rs
```diff
@@ -995,9 +996,50 @@ where
...
+    /// order would vary run to run for identical committed state, making the
+    /// simulator's deny ordering unreproducible for a fixed seed -- the same
+    /// hazard `router.rs` documents as its reason for `select_biased!`.
+    pending_partition_frames: RefCell<BTreeMap<IggyNamespace, Vec<ParkedFrame>>>,
+
+    /// Running sum of [`parked_footprint`] over every frame in
+    /// [`Self::pending_partition_frames`], maintained at each mutation site.
+    ///
+    /// Recomputing it per arriving frame is O(all parked frames): the budget
+    /// admits ~262k entries at 256 bytes each, so filling the buffer would cost
```
hubcio: math is off here: the budget is charged on `parked_footprint`, which floors at 4096, so the ceiling is 16 MiB / 4 KiB = 4096 entries and the worst-case fill is around 8.4M visits - not 262k entries / 10^10. (262144 is 64 MiB / 256, looks keyed off `MAX_MESSAGE_SIZE` instead of `MAX_PARKED_BYTES`; even on logical bytes 16 MiB / 256 is 65536.) the conclusion stands, the running total is the right call.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -280,6 +344,9 @@ struct PassCounters {
     /// blocked by a consumer barrier or by a rejoin whose offsets land via
     /// journal repair, and neither unblocking bumps `Streams::revision`.
     trims_pending: usize,
+    /// Namespaces whose parked frames were answered because this shard is not
```
hubcio: doc matches 2 of the 4 bump sites - the aged branches below also count namespaces that are materialised or still building, not just ones this shard "is not going to materialise". and "answered" is wrong for prepares, which are destroyed with no reply; that answered/destroyed conflation makes the loss paths easy to misread.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -510,6 +599,113 @@ async fn reconcile_additions(
...
+/// deleted -- is aged instead. [`shard::IggyShard::age_parked_partition_frames`]
+/// answers frames past `MAX_PARKED_PASSES`, so residency stays bounded and no
+/// client waits out its read timeout; the deleted case simply takes a few passes
+/// rather than one. The bound is residency only -- the SDK replays the identical
+/// request, so answering a late frame does not stop its operation from being
+/// applied late (see `ParkedFrame::passes`).
+///
+/// A namespace in `staged_this_pass` is exempt: its `ReconcileOp::InsertOwned`
+/// is queued for the pump but not applied, so it reads as un-materialised here
+/// and would burn one of its three passes on the pass that built it.
```
hubcio: understates what the exemption does: `reconcile_additions` has no cross-pass guard against a queued-but-unapplied `InsertOwned` (line 475 tests `partitions.contains`, false while the op is queued), so it re-stages the namespace on every pass until the pump drains - the set protects parked frames across arbitrary pump lag, not one pass. worth saying outright: the one-pass framing here is what makes removing this set look safe, and removing it would age frames on every commit-driven pass while the pump lags.

core/shard/src/lib.rs
```diff
@@ -1841,23 +2331,74 @@ where
...
+            || parked_bytes.saturating_add(frame_cost) > MAX_PARKED_BYTES
+        {
+            self.metrics.record_frame_drop(
+                crate::metrics::frame_drop_variant::PARTITION,
+                crate::metrics::frame_drop_reason::PARK_OVERFLOW,
+            );
+            // Shedding drops a frame, so it is logged at `warn` like every other
+            // drop site: the counter is not registered with a scrape endpoint yet
+            // (see the TODO in `crate::metrics`), so these logs are the only
+            // alertable signal. Rate-limited by the buffer being full -- once it
```
hubcio: "rate-limited by the buffer being full" is not rate limiting - the full buffer is the trigger, and this emits one formatted warn per arriving frame while the map borrow is held. with the per-namespace cap biting at 4 ordinary 1 MiB frames the storm is easy to reach, and the non-blocking appender is lossy under pressure, so it can drop unrelated lines. log the empty -> full transition once and let the counter carry the rest.

core/shard/src/metrics.rs
```diff
@@ -94,16 +113,26 @@ pub mod frame_drop_variant {
...
     pub const UNROUTABLE: &str = "unroutable";
     pub const DELIVERY_FAILED: &str = "delivery_failed";
     pub const MISROUTED: &str = "misrouted";
+    pub const PARK_OVERFLOW: &str = "park_overflow";
+    pub const PARK_DROPPED: &str = "park_dropped";
 }
 
 const VARIANT_COUNT: usize = 7;
-const REASON_COUNT: usize = 5;
+const REASON_COUNT: usize = 7;
```
hubcio, line 139: 5 -> 7 mints the full 7x7 counter cross product per shard; `park_overflow` / `park_dropped` are only ever bumped with `variant=partition`, so 12 pairs are unreachable. free today with no registry - noting for when scraping lands.

core/server-ng/Cargo.toml
```diff
@@ -184,6 +184,13 @@ vergen-git2 = { workspace = true }
 assert_cmd = { workspace = true }
 bytemuck = { workspace = true }
 iggy = { workspace = true }
+# The reconciler's unit tests assert on `ShardMetrics` snapshots and
+# `IggyShard::parked_frame_count`, which are gated to test/simulator builds so
+# they cannot grow production callers. `shard`'s own `cfg(test)` is false when it
+# is compiled as our dependency, so the feature is how those accessors become
+# visible here. Dev-only: a production `cargo build -p iggy-server-ng` does not
+# resolve dev-dependencies, so nothing extra is compiled in.
+shard = { path = "../shard", features = ["simulator"] }
```
hubcio: workspace already has a `shard` entry - `{ workspace = true, features = ["simulator"] }` like the sibling dev-deps. also the comment's mechanism is slightly off: dev-dependencies are always resolved (they are in the lockfile); what the resolver guarantees is that their feature selections are not unified into non-test targets.

core/shard/src/lib.rs
```diff
@@ -1781,29 +1958,340 @@ where
...
+    /// parked, so re-admitting them cannot grow the total past what it held a
+    /// moment ago, and shedding here would answer a frame the inbox merely
+    /// deferred.
+    ///
+    /// Arming [`Self::reparked_partition_namespaces`] is what makes the deferral
+    /// a deferral. Every other exit from the park map is closed for a
+    /// materialised namespace: the reconciler sweep only ages one it has not
+    /// built, and `reconcile_additions` never stages a second `InsertOwned` for
+    /// one already in `IggyPartitions`.
+    fn repark_partition_frames(&self, namespace: IggyNamespace, frames: Vec<ParkedFrame>) {
```
hubcio, line 2236: the per-namespace byte total is re-summed over the whole entry at four sites (park admission, take, repark, age) while the shard-wide total is an incremental cell. a map value of `ParkEntry { frames: Vec<ParkedFrame>, bytes: usize }` kills the rescans, and reading emptiness from the map (`!pending.borrow().is_empty()`) instead of the byte cell removes the bytes-as-emptiness-oracle invariant - which would also make the "single place entries leave the map" claim on `take_parked_partition_frames` true (aging currently removes entries directly and never clears the retry set).

core/shard/src/lib.rs
```diff
@@ -1781,29 +1958,340 @@ where
...
+        self.pending_partition_frames
+            .borrow()
+            .get(&namespace)
+            .map_or(0, Vec::len)
+    }
+
+    /// Answer every frame parked under `namespace` and drop the entry, without
+    /// touching the routing table. Used by the reconciler for a namespace it has
+    /// given up on materialising this pass.
+    pub fn reclaim_parked_partition_frames(&self, namespace: IggyNamespace) {
```
hubcio: pure alias of `discard_parked_partition_frames` - make that one pub and delete this; the park vocabulary is already at park / shed / repark / redispatch / age / retire / discard / reclaim.

core/shard/src/lib.rs
```diff
@@ -1781,29 +1958,340 @@ where
...
-            .borrow_mut()
-            .remove(&namespace)
+        // Bound the borrow to this statement: the guard in an `if let`
+        // scrutinee otherwise lives to the end of the then-block, holding a
+        // shard-global map locked across the outbound sends below.
+        let parked = self.take_parked_partition_frames(namespace);
+        if let Some(frames) = parked
             && !frames.is_empty()
         {
+            let total = frames.len();
```
hubcio: this retire loop (deny -> answered, else `park_dropped`) is duplicated verbatim in `age_parked_partition_frames`. one helper returning (answered, dropped) would also single-site the accounting fix needed for `deny_parked_client_request`'s return value.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -969,6 +1171,83 @@ mod tests {
...
+    async fn park_one_prepare(shard: &TestShard, namespace: IggyNamespace, op: u64) {
+        shard
+            .on_message(build_partition_prepare(namespace, op))
+            .await;
+    }
+
+    /// Build a partition-plane client `Request` for `namespace`, as the pump
+    /// receives it off the wire. Only the routing fields matter: parking reads
+    /// `operation` + `namespace` and never touches the body.
+    fn build_partition_request(namespace: IggyNamespace) -> Message<GenericHeader> {
```
hubcio, line 1212: byte-identical to `build_partition_request_sized(namespace, 0)` - delegate.

hubcio (changes requested):
three paths destroy a quorum-acked prepare with no repair driver (aged out after `MAX_PARKED_PASSES` commit-driven passes; shed by the byte caps; reclaimed on the first build failure). one root cause: a prepare is treated as sheddable like a request, and only a request can retry. the barrier design itself - hash fallback, owner-side park, incarnation fence - looks right; the damage is confined to the residency bounds. details on the inline comments.

besides that:
- `reconcile_additions` (line 475) has no cross-pass guard against a queued-but-unapplied `InsertOwned`, so a pass firing in that window re-runs the whole `build_partition_fresh` (mkdir, four file opens, a full consensus pipeline) and the apply-site guard discards the result - wasted multi-ms io on the create-burst path, and the discarded op also skips the `shards_table` epoch refresh, leaving the row on the older epoch for a couple of passes after a delete + recreate. pre-existing, non-blocking. note the sound guard is checking `reconcile_queue` for a pending op - an `epoch_for` check cannot distinguish never-built from built-and-queued (the row is written by the same apply) and would wedge materialisation.

core/shard/src/lib.rs
krishvishal: Fixed. `age_parked_partition_frames` now expires client requests only; the extract predicate skips `is_replicated()`. Prepare residency is bounded by the byte budgets instead. The only path that still destroys a parked prepare is `discard_parked_partition_frames`, where the namespace is unreachable from this shard (tombstoned or not hashing here). Covered by `aging_answers_requests_and_never_expires_a_prepare`.

core/shard/src/lib.rs
krishvishal: Fixed the way you suggested: prepares are not shed on a byte budget crossing at all. A request is refused the moment admitting it would cross a budget; a prepare only once a budget is already spent, which gives one frame of overshoot (worst case `MAX_PARKED_BYTES` + `max_message_size`, 80 MiB per shard) instead of loss. That also removes the >16 MiB gap, so the TODO is gone. Covered by `a_frame_over_the_namespace_byte_cap_still_parks_into_an_empty_entry`.

core/server-ng/src/partition_reconciler.rs
krishvishal: You are right, the 60s justification was wrong: `next_backoff(1)` is 1s and the clamp needs seven consecutive failures. The `is_backed_off` branch is gone from the sweep, so a backed-off namespace is aged like any other (requests get `MAX_PARKED_PASSES`, prepares are retained). Immediate discard is now only tombstoned or not-ours. Covered by `a_backed_off_build_ages_requests_and_retains_prepares`.

core/shard/src/lib.rs
krishvishal: Fixed in one edit as suggested. On `Full`: push the frame, `extend` with the rest of the iterator, `break`, no counter. The drop is recorded only on the `Disconnected` arm. The unreachable destructure is now `unreachable!`, and the inline duplicate of `deny_parked_client_request` folds into the shared `retire_parked_frames`. Covered by `a_re_parked_frame_is_re_dispatched_once_the_inbox_drains`.

core/shard/src/lib.rs
krishvishal: Fixed with the shape you proposed: `attach_senders(&mut self, Vec<TaggedSender>)`, run through `validate_sender_ordering`, updating `shard_count`, gated `#[cfg(any(test, feature = "simulator"))]`. The test builder now wires `0..=shard_id` and drops the peer receivers, so a frame misrouted to one fails as `Disconnected` instead of landing in this shard's inbox and reading as success.

core/shard/src/lib.rs
krishvishal: Fixed. `stage_transient_deny` returns `bool` and `deny_parked_client_request` forwards it, so a request whose deny was never staged is counted under `park_dropped` rather than logged as answered.

core/integration/tests/cluster/multi_shard_partition_convergence.rs
krishvishal: Agreed, that parenthetical could not have verified anything. Removed the claim; the doc now states the sequencing argument and says outright that the deny logs at `debug` while the harness runs at `info`, so its absence proves nothing. Added `assert_no_degraded_park_paths`, counting the three `warn` markers via `stdout_occurrences` at the end of both tests. Both pass at zero, so the park path is now positively pinned.

core/server-ng/src/partition_reconciler.rs
krishvishal: Rewritten as `a_backed_off_build_ages_requests_and_retains_prepares`: `build_test_shard_with_inbox` + `drain_staged_client_sends` so the answer is observed rather than assumed, both frame classes parked, and it asserts that one failed build destroys nothing. The behaviour it used to pin is gone with the backoff-reclaim fix.

core/server-ng/src/partition_reconciler.rs
krishvishal: Correct, that was the per-namespace cap. Renamed to `park_namespace_byte_budget_sheds_large_frames_before_the_frame_cap` and made the assertion exact (`NAMESPACE_BUDGET / MIB_FOOTPRINT` frames, then one shed). Added `park_shard_wide_byte_budget_sheds_a_namespace_that_would_cross_it` with the shape you described, so `MAX_PARKED_BYTES` is bound by a test for the first time.

core/shard/src/lib.rs
krishvishal: Fixed as suggested, and both directions still reject. Ahead now bumps a separate `partition_frames_rejected_ahead_total` and logs at `debug`; behind keeps `partition_frames_rejected_stale_total` and the `warn`, so the anomaly counter keeps its contract instead of firing on ordinary delete + recreate churn.

core/shard/src/lib.rs
krishvishal: Fixed. The doc now says 4096 entries and ~8.4M visits, keyed off `MAX_PARKED_BYTES` and the `MESSAGE_ALIGN` floor rather than `MAX_MESSAGE_SIZE`.

core/server-ng/src/partition_reconciler.rs
krishvishal: Fixed. The doc now covers all bump sites and separates the two outcomes: aging answers client requests, discarding also destroys prepares. It also says it counts namespaces, not frames.

core/server-ng/src/partition_reconciler.rs
krishvishal: Fixed, and stated outright: `reconcile_additions` has no cross-pass guard against a queued-but-unapplied `InsertOwned`, so it re-stages every pass until the pump drains. The exemption therefore covers arbitrary pump lag, and removing it would age frames on every commit-driven pass the pump falls behind.

core/shard/src/lib.rs
krishvishal: Fixed. `ParkEntry` carries a `shed` count; the transition into shedding warns once per namespace and everything after logs at `debug`. The counter carries the volume.

core/shard/src/metrics.rs
krishvishal: Noted in a comment beside the constants: 12 pairs are unreachable, and once a registry lands the counters should be minted per drop site rather than by cross product so the scrape carries no permanent zeroes.

core/server-ng/Cargo.toml
krishvishal: Fixed both. Now `shard = { workspace = true, features = ["simulator"] }`, and the comment says the resolver keeps a dev-dependency's feature selection out of non-test targets, rather than claiming dev-dependencies are not resolved.

core/shard/src/lib.rs
krishvishal: Done. `ParkEntry { frames, bytes, shed }` removes the rescans at all four sites, `has_parked_partition_frames` reads the map instead of using the byte cell as an emptiness oracle, and aging now routes its removal through `take_parked_partition_frames`, which clears the retry set with the entry. The single-removal-site claim is true now.

core/shard/src/lib.rs
krishvishal: Deleted. `discard_parked_partition_frames` is `pub` and callers use it directly.

core/shard/src/lib.rs
krishvishal: Extracted `retire_parked_frames(frames) -> (answered, dropped)`, used by both sites. That also single-sited the accounting fix for `deny_parked_client_request`'s return value.

core/server-ng/src/partition_reconciler.rs
krishvishal: Delegated to `build_partition_request_sized(namespace, 0)`.

core/shard/src/lib.rs
krishvishal: Confirmed, and the window is reachable exactly as you describe. I took the second option for now: recorded as a `TODO(krishna)` in the module docs and softened the bullet's absolute, since the stamp is re-derived on re-entry rather than carried. A real fix needs the frame to carry its provenance through the inbox instead of re-reading committed metadata on arrival, which is more than this PR should take on. Leaving this thread open rather than resolving it, since the gap is documented but not closed.

core/shard/src/lib.rs
```diff
@@ -1841,23 +2391,98 @@ where
...
+                namespace_bytes >= MAX_PARKED_BYTES_PER_NAMESPACE
+            } else {
+                namespace_bytes.saturating_add(frame_cost) > MAX_PARKED_BYTES_PER_NAMESPACE
+            };
+        let shard_budget_spent = if replicated {
+            parked_bytes >= MAX_PARKED_BYTES
+        } else {
+            parked_bytes.saturating_add(frame_cost) > MAX_PARKED_BYTES
+        };
+        if parked_len >= MAX_PARKED_PER_NAMESPACE || namespace_budget_spent || shard_budget_spent {
```
hubcio: the frame cap still sheds prepares: `parked_len >= MAX_PARKED_PER_NAMESPACE` is checked unconditionally, before the class-split budget checks get a say. for footprints under 32 KiB (4 MiB / 128) the 128-frame cap fires long before the per-namespace byte budget ever could - reproduced: 128 header-only prepares park, the 129th is shed with `park_overflow` bumped, at 512 KiB of a 4 MiB budget. the byte budgets never fire. so small-append prepares are destroyed exactly as before this fix, and the backoff change widens the window: prepares are now retained across a failing build (up to 60s), and 128 small prepares in 60s is 2/sec. reachable at every replica count, not just 2 - quorum always has at least one surplus backup (n=2 primary self-acks alone, n=3 quorum is 2 of 3), so one lagging backup loses shed prepares with nobody noticing.

this also contradicts the new docs in four places ("byte budgets bound those instead", the three-axes paragraph, "a shed frame must never grow it past that"), and the module doc's known-gaps TODO says exposure is narrowed to two cases - the frame cap is an undocumented third.

fix is the same shape as the budget split: `parked_len >= MAX_PARKED_PER_NAMESPACE && !replicated`. verified against this tree: prepares then admit to exactly 1024 per namespace (4 MiB) before the byte budget sheds, requests still cap at 128, all existing tests stay green, and every drain path still works (re-dispatch, discard, age, stale-reject; a backup's replicate path never touches the primary-side prepare queue, so no assert hazard). two things must land with it: the four doc sites above flip stale the moment the cap becomes request-only, and a test is mandatory - the one-liner breaks no existing test, so it would land uncovered. shape: >128 small prepares retained with the byte budget as the shedder, plus a request into the same deep entry still shed at 128. worth knowing: with 1024 parkable prepares one namespace's re-dispatch can now fill the default 1024-slot inbox where it was 1/8th - the first-`Full` re-park handles it, other namespaces just converge a bit later.

core/shard/src/lib.rs
```diff
@@ -1841,23 +2391,98 @@ where
...
+                crate::metrics::frame_drop_variant::PARTITION,
+                crate::metrics::frame_drop_reason::PARK_OVERFLOW,
             );
-            return ParkOutcome::Parked;
+            // Warn once per namespace on the transition into shedding, then
+            // `debug`. A full buffer is this branch's trigger, not a rate limit:
+            // every later frame lands here too, and one formatted `warn` apiece
+            // is enough for the non-blocking appender to shed unrelated lines.
+            // The counter carries the volume.
+            let first_shed = existing.is_none_or(|entry| {
```
hubcio: `is_none_or` returns true every time the namespace has no entry, and only the shard-wide budget can shed with `existing == None` (the other two conditions require a non-empty entry) - so on that path every frame warns and `shed` is never recorded. that is exactly the load `MAX_PARKED_BYTES` was sized for per its own doc, a burst against many un-materialised namespaces, so under the intended scenario this warns per frame again: the log-loss the warn-once change exists to prevent. the existing `park_shard_wide_byte_budget_sheds_a_namespace_that_would_cross_it` test already drives this path if you want to assert on it.

needs a shard-level gate for the entryless case - not `entry().or_default()`, which would reintroduce the empty-entry-reads-as-parked problem the comment above deliberately avoids.

hubcio (changes requested):

hubcio (approved):

numinnex (approved):

spetz (approved):

PR #3792, opened 2026-08-01
fix(ci): make README CLI preflight execute its commands

### LINKED ISSUE #3791: CI: Rust and C# preflight never executes any README CLI command
### Bug description

The examples CI job reports success for the Rust and C# preflight without executing a single CLI command from the root README.

Both preflights (`scripts/run-examples-from-readme.sh:188`, `:350`) extract CLI commands with:

```bash
run_readme_commands "README.md" '^\`cargo r --bin iggy -- '
```

Two independent bugs stack here:

1. **Stale literal.** The README uses `cargo run --bin iggy -- ...`, not `cargo r`, so the pattern matches nothing on any grep.
2. **GNU grep anchor.** In GNU grep ERE, `` \` `` is a buffer-start anchor, not a literal backtick. Even with the literal fixed, `` ^\` `` requires line start *and* buffer start, so only line 1 could ever match: the corrected pattern matches 9 lines with BSD grep but 0 with GNU grep — and CI runs on Linux.

`run_readme_commands` (`scripts/utils.sh:305`) silently returns on zero matches, so both preflights pass green. As a result the 9 CLI commands in the README (stream create/list/get, topic create/list/get, message send ×2, message poll) are never exercised, and no other CI check covers them.

### Affected area / component

CI / build / tooling

### Deployment

Not applicable

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

```bash
# GNU grep (Linux/CI): current pattern and literal-fixed pattern both match 0
$ docker run --rm -v "$PWD/README.md:/README.md:ro" debian:stable-slim bash -c "
    grep -cE '^\\\`cargo r --bin iggy -- ' /README.md
    grep -cE '^\\\`cargo run --bin iggy -- ' /README.md
    grep -cE '^[\\\`]cargo run --bin iggy -- ' /README.md"
0
0
9
```

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

Run `./scripts/run-examples-from-readme.sh --language rust --skip-tls` on Linux: the preflight completes without printing a single `Checking command from README.md:` line for the CLI commands.

### Contribution

### Good first issue

ethanlin01x:

Closes #3791

## Rationale

The Rust and C# preflights reported success without executing a single CLI command from the root README; see the issue for the full analysis.

## What changed?

The preflight grep still targeted `cargo r` (the README now uses `cargo run`), and GNU grep treats `` \` `` in an ERE as a buffer-start anchor, so the extraction was always empty and the runner silently skipped it.

The pattern now matches the backtick via a bracket expression, and the transform fills in the README's credential placeholders so the commands survive `eval`. 
A separate commit restarts the server after the preflight: leftover streams shift the auto-assigned IDs and hang `getting-started-consumer`, which caused the first CI run's 60-minute timeout. 
The C# preflight is removed as it duplicated the Rust job, the only one that builds the CLI. 
A final docs commit normalizes `cargo r` to `cargo run` in the example docs (none of those lines run in CI).

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage

1. Which tools? Claude code
2. Scope of usage? help to analysis the issue and generate git description
3. How did you verify the generated code works correctly?  Ran the fixed preflight end to end against a real local server


ethanlin01x:
Sorry for the noise from the repeated re-drafts. Local verification passed, but the full CI run still timed out: the preflight leaves state behind that breaks the examples' fresh-server assumption. The restart commit fixes it.

slbotbm (dismissed):
make `_rust_preflight` fail when zero commands match. `run_readme_commands` still silently succeeds on an empty extraction, so a future README formatting change could recreate the same false-positive CI behavior.

hubcio (approved):

ethanlin01x:
Hi @slbotbm 
Addressed! 8e81bb0b282bd58caba3d686ceea8bfca03379de
Added a check in `_rust_preflight` to ensure it fails when zero commands match.

spetz (approved):

thepatrykk (approved):

numinnex (approved):

hubcio (approved):

PR #3795, opened 2026-08-01
fix(connectors): bound source forwarding channel with backpressure

mlevkov:
## Summary

The channel between a source plugin's send callback and the runtime's forwarding loop was `flume::unbounded()`, so a slow or hung Iggy meant batches accumulated in memory without bound instead of propagating backpressure into the plugin's polling loop. This is the prerequisite runtime fix requested in the HTTP source discussion (#3039), and it applies to every source connector, including the source PRs currently in flight.

## What changed

- The forwarding channel is now a bounded crossfire channel (`crossfire::mpsc::bounded_blocking_async`), the same shape `shard` and `server-ng` use. flume is no longer a runtime dependency.
- Capacity comes from a new optional `SourceConfig` field, `channel_capacity`, counted in batches (a single batch can be megabytes), defaulting to 1024 and clamped to [1, 65536] since crossfire eagerly allocates the ring and asserts capacity < 2^31. The existing `ConfigEnv` derive provides `IGGY_CONNECTORS_SOURCE_<KEY>_CHANNEL_CAPACITY`; configs without the field behave as before apart from the bound.
- The FFI send callback applies backpressure with a `try_send` fast path and a `send_timeout(10ms)` retry loop that re-reads a per-instance shutdown flag between waits. The manager sets that flag before `iggy_source_close` so a hung Iggy cannot deadlock the close. Process shutdown sets every instance's flag (`signal_shutdown_all`) before the sequential stops, because instances loaded from one plugin library share a single tokio runtime and a wedged sibling would otherwise hold a worker an earlier close needs.
- A full channel logs one `warn!` per backpressure episode (latched, cleared on genuine recovery). A batch that still cannot be enqueued after the stop signal is dropped and counted in `iggy_connector_errors_total`.
- Unit tests pin the behavior shutdown relies on: buffered batches drain after the senders drop (crossfire's docs do not promise this), the retry loop unblocks when the flag flips mid-backoff, and shutdown drops are counted without being enqueued.
- Docs updated across the runtime README, sources README, and the connector skills, which still described flume and the pre-#3321 shutdown order.

## One correction to the discussion notes

@hubcio the spec assumed crossfire's blocking sender has no `send_timeout`. It does: `blocking_tx.rs:288` on `Tx`, reachable from `MTx` via `Deref`. The loop is built on it instead of `try_send` plus sleep, so the sender wakes as soon as capacity frees while shutdown latency stays bounded by the retry interval.

## Known limitation

Stopping a single connector via the runtime API while enough same-library sibling instances are saturated can delay that close until the siblings drain, because the callback parks a worker of the shared plugin runtime. The code comment and the connector skill document this. The complete fix is an SDK-side worker handoff (`tokio::task::block_in_place` around the callback invocation); happy to file it as a follow-up issue.

## Test plan

- `cargo clippy -p iggy-connectors --all-targets -- -D warnings` clean
- `cargo test -p iggy-connectors`: 128 passed, including the new channel and shutdown tests
- `cargo build -p iggy_connector_stdout_sink -p iggy_connector_random_source`
- `cargo test -p integration -- connectors::runtime::` could not run on this machine (`hwlocality-sys` needs `pkg-config`); relying on CI for the integration suite


core/connectors/runtime/src/source.rs
```diff
@@ -729,23 +793,102 @@ pub(crate) extern "C" fn handle_produced_messages(
...
+fn send_with_backpressure(
+    plugin_id: u32,
+    sender: &BatchSender,
+    shutdown: &AtomicBool,
+    backpressure_active: &AtomicBool,
+    error_counter: &Counter,
+    messages: ProducedMessages,
+) {
+    let mut messages = match sender.try_send(messages) {
+        Ok(()) => {
```
hubcio, line 848: the `Ok(())` arm never reads `shutdown`, which makes the shutdown drop below non-terminal: drop batch N, the forwarding loop frees a slot, batch N+1 enqueues, ships, and persists its state - the saved cursor now covers N. sources advance the cursor at poll time and snapshot it into every batch (postgres `tracking_offsets`, same shape in the other sources), so a restart resumes past the hole. silent mid-stream loss, not tail truncation.

the same supersession already exists on master via the `producer.send()` err path (skips the save, continues), but that one flips the connector to Error status; this one only logs + counts, and fires on routine paths - SIGTERM (`signal_shutdown_all` arms every source for the whole sequential-stop window) and connector restart via the api.

fix is a per-instance `dropped` latch: after the first drop, drop every later batch too, so the persisted cursor can never pass the gap (ring is fifo, buffered older batches still flush). use the same flag to latch the `error!` in `drop_during_shutdown`, otherwise after latching it fires once per poll; keep the counter bumping per batch.

core/connectors/runtime/src/source.rs
```diff
@@ -729,23 +793,102 @@ pub(crate) extern "C" fn handle_produced_messages(
...
+            }
+            return;
+        }
+        Err(TrySendError::Full(returned)) => returned,
+        Err(TrySendError::Disconnected(returned)) => {
+            log_channel_closed(plugin_id, returned.messages.len(), error_counter);
+            return;
+        }
+    };
+    if shutdown.load(Ordering::Acquire) {
```
hubcio: zero grace: the first `Full` after the flag drops immediately, while the forwarding loop keeps draining until `cleanup_sender`. one bounded `send_timeout` round before the first drop (skipped once latched) lets most in-flight batches land - the parked sender is woken on drain, so the wait is the drain period, not the full timeout. anything larger only pays off after `iggy_source_close` moves to `spawn_blocking`, since it currently blocks a host worker for the duration.

core/connectors/runtime/src/source.rs
```diff
@@ -729,23 +793,102 @@ pub(crate) extern "C" fn handle_produced_messages(
...
+            }
+            Err(SendTimeoutError::Disconnected(returned)) => {
+                log_channel_closed(plugin_id, returned.messages.len(), error_counter);
+                return;
+            }
+        }
+    }
+}
+
+fn drop_during_shutdown(plugin_id: u32, message_count: usize, error_counter: &Counter) {
```
hubcio: the log reports a message count but the counter moves by 1 per batch, into the same `iggy_connector_errors_total` series as decode/send/save failures. this is the only signal for a permanently lost batch - a dedicated counter (with `inc_by(message_count)` where the count exists) keeps loss distinguishable from ordinary errors.

core/connectors/runtime/src/source.rs
```diff
@@ -729,23 +793,102 @@ pub(crate) extern "C" fn handle_produced_messages(
...
                     "Failed to deserialize produced messages for source connector with ID: {plugin_id}. {err}"
                 );
-                entry.error_counter.inc();
+                error_counter.inc();
             }
         }
     }
 }
 
+// Parks a worker thread of the plugin library's shared tokio runtime - a
```
hubcio, line 826: this covers the close-delay case but not steady state: the park removes a worker from the plugin library's shared runtime (one runtime per .so, workers = available_parallelism), so saturated instances degrade every sibling from the same .so, and a 1-2 vcpu container can wedge on one instance. worth a sentence here and in the skill doc. also, a runtime-side `tokio::task::block_in_place` would be a silent no-op on these threads (they belong to the plugin's tokio, not ours), so the handoff fix really does have to be sdk-side.

core/connectors/runtime/src/source.rs
```diff
@@ -768,3 +911,314 @@ fn build_iggy_message(
...
+        );
+        assert!(
+            receiver.try_recv().is_err(),
+            "shutdown-dropped batch must not have been enqueued"
+        );
+    }
+
+    #[test]
+    fn given_backoff_in_progress_when_shutdown_signaled_should_unblock_and_drop() {
+        let (sender, _receiver) = bounded_channel(1);
```
hubcio: this test and `given_full_channel_when_receiver_frees_capacity_should_deliver_batch` are the only ones that enter the backoff loop, and both run capacity 1, which crossfire routes to a different queue impl (`OneMpsc`) and a different backoff regime (`large = capacity >= 10` gates yield-vs-spin). prod default 1024 is `ArrayMpsc` with `large = true`, so the park/wake path being pinned isn't the shipped one. capacity >= 10 here fixes both. the capacity-4 drain test is fine - `try_send`/`recv` never enter the backoff.

core/connectors/runtime/src/source.rs
```diff
@@ -50,11 +53,28 @@ use iggy_connector_sdk::api::ConnectorStatus;
 use prometheus_client::metrics::counter::Counter;
 use tokio::task::JoinHandle;
 
+pub(crate) const DEFAULT_CHANNEL_CAPACITY: usize = 1024;
```
hubcio: 1024 batches is a weak bound in bytes - postgres defaults to 1000 rows per batch, so this admits about a million messages plus payloads before backpressure engages. something like 64 would still be generous. sequencing note: land any smaller default after the shutdown-drop fix, since a smaller channel makes `Full` (and drops) more frequent.

core/connectors/runtime/src/source.rs
```diff
@@ -50,11 +53,28 @@ use iggy_connector_sdk::api::ConnectorStatus;
...
+
+// crossfire eagerly allocates the whole ring and asserts capacity < 2^31;
+// the cap keeps a config typo from panicking the process or committing
+// gigabytes up front.
+const MAX_CHANNEL_CAPACITY: usize = 65_536;
+
+const SEND_RETRY_INTERVAL: Duration = Duration::from_millis(10);
```
hubcio, line 67: future-tuning note: lowering this buys no latency - a parked sender is woken by the receiver the moment a slot frees, the interval only bounds shutdown-flag observation. smaller values just multiply yield/park churn.

core/connectors/runtime/src/source.rs
```diff
@@ -50,11 +53,28 @@ use iggy_connector_sdk::api::ConnectorStatus;
...
+
+pub(crate) type BatchSender = MTx<Array<ProducedMessages>>;
```
hubcio: `DEFAULT_CHANNEL_CAPACITY`, `BatchSender` and `BatchReceiver` have no uses outside this file and can all drop `pub(crate)` - the private-interfaces lint checks the resolved type (public crossfire types), not the alias, so private aliases in `pub(crate)` signatures compile clean.

core/connectors/runtime/src/source.rs
```diff
@@ -719,7 +772,18 @@ pub(crate) extern "C" fn handle_produced_messages(
     unsafe {
         // Entry missing = SOURCE_SENDERS cleaned up at shutdown; benign race
         // expected on stop/restart. No metric (would conflate with real failures).
-        let Some(entry) = SOURCE_SENDERS.get(&plugin_id) else {
+        // Clone out and drop the guard: the backoff loop below may run for a
+        // while, and holding the shard guard would block cleanup_sender.
+        let Some((sender, error_counter, shutdown, backpressure_active)) =
```
hubcio: only `from_raw_parts` needs the unsafe block - `let messages = unsafe { std::slice::from_raw_parts(messages_ptr, messages_len) };` and everything else moves out. the wide block predates this pr, but it grew here.

core/connectors/runtime/src/source.rs
```diff
@@ -719,7 +772,18 @@ pub(crate) extern "C" fn handle_produced_messages(
...
+            SOURCE_SENDERS.get(&plugin_id).map(|entry| {
```
hubcio: simpler shape for this: `DashMap<u32, Arc<SourceSenderEntry>>`, clone one arc, pass `&SourceSenderEntry` down (`send_with_backpressure` goes 6 params -> 3), and `shutdown`/`backpressure_active` become plain `AtomicBool`. disconnect timing is unchanged - the arc keeps the `MTx` alive exactly like today's clone. if you take it, write `Arc::clone(&entry)` explicitly; dashmap's `Ref` has no `Clone`, so a bare `.clone()` only works via deref and reads as a guard clone.

core/connectors/runtime/src/source.rs
```diff
@@ -768,3 +911,314 @@ fn build_iggy_message(
...
+            state: None,
+        }
+    }
+
+    fn bounded_channel(capacity: usize) -> (BatchSender, BatchReceiver) {
+        crossfire::mpsc::bounded_blocking_async(capacity)
+    }
+
+    // Shutdown relies on buffered batches surviving sender drop; crossfire's
+    // docs don't promise it, so this pins the behavior against upgrades.
```
hubcio: crossfire does document drain-before-disconnect for `try_recv` and `recv_timeout` ("...and the channel is empty") - it's `recv()` whose doc omits the qualifier. reword to name `recv()` so the comment survives someone checking the docs; the test is worth keeping regardless.

core/connectors/runtime/src/source.rs
```diff
@@ -768,3 +911,314 @@ fn build_iggy_message(
...
+        blocked.join().expect("blocked thread panicked");
+        assert_eq!(
+            error_counter.get(),
+            1,
+            "batch dropped after mid-backoff shutdown must count as an error"
+        );
+    }
+
+    #[test]
+    fn given_registered_entry_when_signal_shutdown_called_should_set_flag() {
```
hubcio: this can pass even if `signal_shutdown` were a no-op: the sibling test's `signal_shutdown_all()` sets every entry's flag, including this one, and id partitioning doesn't help against a whole-map op. one merged test with two entries - target set, sibling not set, then `signal_shutdown_all` sets both - is hermetic. only plain `cargo test` is affected (nextest is process-per-test), but that's the documented local flow.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -147,8 +147,11 @@ impl SourceManager {
...
-        // Order: close FFI (stops callbacks) -> drop sender (unblocks
-        // recv_async) -> await tasks. Reversing risks an abort mid-save.
+        // Order: signal shutdown (unwedges a callback stuck in its
+        // full-channel backoff, which iggy_source_close waits on) -> close
+        // FFI (stops callbacks) -> drop sender (unblocks recv) -> await
+        // tasks. Reversing risks a deadlocked close or an abort mid-save.
+        source::signal_shutdown(plugin_id);
         if let Some(container) = &container {
             info!("Closing source connector with ID: {plugin_id} for plugin: {key}");
             (container.iggy_source_close)(plugin_id);
```
hubcio: return code dropped, and the new ordering's safety argument leans on close actually stopping callbacks - the sdk returns -1 when it can't join the polling task, and then callbacks can outlive the close and land on the removed-entry branch that drops with no metric. `init` checks the same call (`if close_result != 0 { warn! }`) - mirror it here.

core/connectors/runtime/README.md
```diff
@@ -219,6 +219,21 @@ Emitted fields:
 
 Filter the stream via `RUST_LOG=iggy_connectors::benchmark=info`. The corresponding stage durations are also recorded in the `iggy_connector_stage_duration_seconds` histogram regardless of this flag, so Prometheus dashboards remain available without enabling text events.
 
+## Source Channel Capacity
+
+Each source configuration accepts an optional `channel_capacity` setting that bounds the channel between the plugin's send callback and the runtime's forwarding loop. Capacity is counted in batches (one `poll()` result each, potentially megabytes), not messages or bytes. The default is 1024 batches.
+
+When the channel is full (Iggy accepts messages more slowly than the plugin produces them), the send callback backs off and retries instead of buffering without bound, so backpressure propagates into the plugin's polling loop. During shutdown, a batch that still cannot be enqueued after the stop signal is dropped and counted in `iggy_connector_errors_total`, so a saturated source may report errors at SIGTERM. Values outside `[1, 65536]` are clamped with a warning.
```
hubcio: this undersells the failure: the drop isn't just an error count - a later batch can still enqueue and persist its state, moving the saved position past the dropped batch, so the data is silently gone (see the send-callback comment). it also happens on connector restart via the api, not only at SIGTERM. once the fix lands this should state the guarantee: replay from the last delivered batch's state - duplicates for offset-mode sources, permanent loss for `delete_after_read` / `processed_column` ones. same wording lives in the connector-runtime skill doc.

core/connectors/runtime/README.md
```diff
@@ -219,6 +219,21 @@ Emitted fields:
...
+
+```toml
+type = "source"
+key = "postgres"
+# ... other fields ...
+channel_capacity = 1024
+```
+
+Environment override: `IGGY_CONNECTORS_SOURCE_<KEY>_CHANNEL_CAPACITY`.
```
hubcio: only the local config provider wires env overrides - with `config_type = "http"` this var is silently ignored, and the unknown-var warning is suppressed for the `IGGY_CONNECTORS_SOURCE_` prefix, so nothing surfaces it. qualify with "local config provider only".

core/connectors/sources/README.md
```diff
@@ -72,6 +73,7 @@ path = "libiggy_connector_random_source" # Path to the source connector
 config_format = "toml"
```
hubcio: pre-existing, two lines from your change: the example key is `config_format` but the real field is `plugin_config_format` - no serde alias and no `deny_unknown_fields`, so the documented key is silently ignored.

hubcio (changes requested):
a few things that don't fit a diff line:

- `manager/source.rs:164` - the 5s task-await is per handle (there are two), fixed, and `channel_capacity` now scales the post-`cleanup_sender` drain against it (buffered batches still drain after the sender drops, up to capacity send+save iterations). abort mid-drain is a clean tail truncation since the state save is atomic, so it costs lost position, not corruption - but the interaction deserves a line in the readme: raising capacity lengthens shutdown.
- `source.rs:627` - `iggy_source_handle`'s i32 is discarded too, same family as the close-code comment; only reachable in a start-then-stop race today, so hygiene.
- elasticsearch_source with `[state] enabled = true` persists its own cursor at close and overrides the runtime state at open, so a runtime-side latch can't cover it. opt-in and off by default; follow-up issue.
- the pre-existing `producer.send()` err path has the same cursor-supersession shape (skip save, continue, next batch persists) - the latch here doesn't close that one; separate follow-up.
- worth a regression test once the latch lands: saturate the channel, stop the connector, restart, assert no row gap - `state_persists_across_connector_restart` in the postgres integration suite is a natural template.
- `spawn_source_handler` is at 11 params; passing `&SourceConfig` instead needs the resolved-path/version split on `SourceConnectorPlugin` first, so follow-up sized.

mlevkov:
Thanks - the shutdown-drop finding is a real bug and I had it characterised wrong in the PR description. Pushed `8c4100e62` on top of your branch update.

**The silent supersession.** Traced it and you are right: the drop returns, the forwarding loop keeps draining so a slot frees, the next batch enqueues and ships, and the save that follows carries a cursor covering the batch that was dropped. Calling that tail truncation was wrong; it is mid-stream loss, and it fires on SIGTERM and on API restart, not on some exotic path.

Implemented the per-instance latch as you described. First drop sets `dropped`, every later batch from that instance is dropped too, so the persisted cursor can never pass the gap. Queued batches still flush since the ring is FIFO and they all predate the drop. The `error!` latches with the flag; the counter keeps moving per batch.

Mutation-checked: removing the latch check fails the new test with "a batch after the gap must be dropped even with capacity free, or the persisted cursor would advance past the batch that was lost".

**Grace round** before the first drop, as suggested - one bounded `send_timeout`, skipped once latched. Its test fails if the round is removed.

**Dedicated counter.** `iggy_connector_messages_dropped_total`, `inc_by(message_count)`, separate from `errors` so permanent loss is not mixed in with decode and send failures that get retried.

**Capacity default 1024 -> 64**, landed together with the fix rather than before it, per your sequencing note. READMEs and the skill doc follow.

**Tests.** Both backoff tests move to capacity 16, since capacity 1 routes to `OneMpsc` and a different backoff regime than the shipped default - the park/wake path being pinned was not the one that ships. The two shutdown-signal tests are merged into one hermetic test with two entries (target set, sibling not, then `signal_shutdown_all` sets both); you were right that the sibling's whole-map call could have covered for `signal_shutdown` being a no-op.

**Also taken:** close return code checked in `stop_connector` the way `init` already does it; `DashMap<u32, Arc<SourceSenderEntry>>` with `&SourceSenderEntry` passed down (6 params -> 3, and the flags become plain `AtomicBool`); unsafe narrowed to `from_raw_parts`; `DEFAULT_CHANNEL_CAPACITY`/`BatchSender`/`BatchReceiver` dropped to private; `recv()` named in the drain comment; the steady-state worker-starvation point added to the code comment and the skill doc; README now states the guarantee (resume from the last delivered batch's state, duplicates for offset sources and permanent loss for `delete_after_read`/`processed_column`) and qualifies the env override as local-provider-only; `config_format` -> `plugin_config_format` in the sources README.

**Left for follow-ups, as you framed them:** the elasticsearch_source cursor override, the `producer.send()` err path with the same supersession shape, the `spawn_source_handler` parameter count (needs the resolved-path/version split first), and the `iggy_source_handle` return code. Happy to file those as issues so they do not evaporate - say the word and I will, or I can take any of them here.

The saturate-stop-restart-assert-no-gap regression test you suggested is the one I would most like to add, but it belongs in the postgres integration suite next to `state_persists_across_connector_restart` rather than in this diff. Also happy to do that as a follow-up PR.

Note on the branch: you had updated it with a merge from master, so I rebased my commit on top of that rather than force-pushing, and this was a normal fast-forward push - your merge commit is intact.

Gate: fmt, sort, clippy `-D warnings`, `cargo test -p iggy-connectors` 134 passing, taplo, typos, license headers (via `hawkeye check`; `scripts/ci/license-headers.sh` now requires bash >= 4.2 after #3837 and will not run on macOS's 3.2).

mlevkov:
Follow-ups from your review are filed so they survive this PR merging:

- #3898 - `producer.send()` failure lets a later batch supersede the skipped save. Same supersession shape, not closed by the latch here since that is armed only by a drop.
- #3899 - elasticsearch_source persists its own cursor and overrides the runtime state at open, so a runtime-side guarantee cannot cover it.
- #3900 - `spawn_source_handler` at 11 params, blocked on the resolved-path/version split on `SourceConnectorPlugin`.

One I took here instead of filing: the discarded `iggy_source_handle` return code (`f2db3b48d`, now `5de9283fa` after the rebase). It is the twin of the close-code check already in this PR, so fixing one and deferring the other seemed worse than the small extra diff. Non-zero means the SDK could not register the handler, which leaves an instance reporting Running while producing nothing.

**On the regression test** - the saturate, stop, restart, assert-no-row-gap one against `state_persists_across_connector_restart`: I read "once the latch lands" as meaning a follow-up, so I have not written it. But it is the direct regression test for the fix in this PR, so I would rather ask than guess. Do you want it here, or as a separate PR once this merges? If here, it goes in the postgres integration suite rather than the runtime unit tests, so it will need Docker in the loop.

Also rebased onto current master per your note, along with my other three. Gate green on all four; this branch is 134 passing.

core/connectors/runtime/src/configs/connectors.rs
```diff
@@ -137,6 +137,8 @@ pub struct CreateSourceConfig {
     pub verbose: bool,
     #[serde(default)]
     pub benchmark: bool,
+    /// Forwarding channel capacity in batches; defaults to 1024.
```
hubcio, line 140: still says 1024 - `DEFAULT_CHANNEL_CAPACITY` is 64 and the readme says 64. same at line 183.

core/connectors/runtime/src/source.rs
```diff
@@ -768,3 +944,336 @@ fn build_iggy_message(
         (None, None) => IggyMessage::builder().payload(payload.into()).build(),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use iggy_connector_sdk::ProducedMessage;
+
+    // Prod default (1024) is crossfire's `ArrayMpsc` with `large = true`.
```
hubcio, line 953: prod default is 64 now, not 1024. the substance holds - 64 is still `ArrayMpsc` with `large = true` - but the number is stale.

hubcio (changes requested):

mlevkov:
I reassessed this against `af9ce9548` (#3855, source batch acknowledgments) instead of rebasing it, because the ack contract changes the premise this PR rests on.

Three things hold on master `c0c74931b`:

1. The forwarding channel is `flume::unbounded()` (`runtime/src/source.rs:625`) and the loop is `while let Ok(batch) = receiver.recv_async().await`. flume yields queued items before it reports disconnect, so every queued batch is drained once the sender goes away. There is no drop path on master today.
2. `handle_messages` (`sdk/src/source.rs:278`) keeps `pending_batch: Option<PendingBatch>` and awaits a oneshot for the batch result before it increments `batch_id` and polls again. One batch in flight, by construction.
3. The state save runs inside the send-success branch, and `batch_result = Ack` is set only when that save also succeeded.

That retires the bug your review found. "Drop batch N, the loop keeps draining, N+1 enqueues and persists a cursor covering N" needs two batches in flight, and (2) forbids it. It also means the bound this PR adds can never fill, since the producer stops after one batch until the runtime acks. So `channel_capacity`, the `dropped` latch, the grace `send_timeout` round and `iggy_connector_messages_dropped_total` all guard a path that can no longer execute, and the PR would be introducing the drop path in order to defend it. Your two comments about the stale 1024 in the doc strings are moot along with the rest.

One gap does survive, and it is the piece I would keep. `BATCH_RESULT_TIMEOUT` is 30s (`sdk/src/source.rs:54`). When the forwarding loop lags past it the SDK nacks, re-polls and enqueues a new batch while the timed-out one is still queued, so depth grows by about one batch per 30s per source. `MAX_CONSECUTIVE_NACKS = 5` (`:57`) looks like a ceiling, but any successful ack resets the counter, so an alternating ack/timeout pattern never trips it. Ordering stays safe (FIFO plus a sequential loop), so this is memory pressure rather than cursor corruption. Worth noting that a bounded channel only helps here if the callback blocks. Dropping would reintroduce exactly the loss this PR set out to prevent.

One more thing I noticed while reading: a batch that times out and then sends successfully has its Ack refused by the plugin on the `take_pending_batch` id mismatch (`sdk/src/source.rs:466`), so the runtime logs an error and sets connector error state for a batch that actually shipped fine.

Two questions, since the call is yours:

- Close this and file the 30s timeout growth separately, or reshape this PR around it as a blocking handoff with no drop?
- Does the same reasoning retire #3898? It needs a later batch to ship while an earlier one failed, which (2) also forbids.

#3797 hangs off this decision: it wraps the send callback in `block_in_place` for backpressure that only exists if this bound lands. I have left both unrebased pending your answer. #3798 and #3804 are rebased onto master and green.

mlevkov:
Closing this per your call on Discord.

To leave the reasoning on the record: #3855 retires the premise. One batch is in flight at a time, master's forwarding channel is `flume::unbounded()` so there is no drop path to begin with, and the cursor is persisted only after a successful send. The `dropped` latch here would be defending a drop path that this PR itself introduces.

Two things from this work carry forward rather than disappearing with the PR:

- The 30s `BATCH_RESULT_TIMEOUT` queue-growth case. You said the timeout stays and Rohan is settling what it was defending against, so I have not filed anything for it. Say the word if you want it tracked once he answers.
- A batch that times out and then sends successfully has its Ack refused on the `take_pending_batch` id mismatch, so the runtime logs an error and sets connector error state for a batch that actually shipped. Small, separate, happy to file it.

#3899 came out of this review and stays open on its own, since the elasticsearch_source cursor override at open is independent of acks.

mlevkov:
Following up so this does not end here.

Closing this was right: #3855's one-batch-in-flight rule means the forwarding channel cannot grow, which was the whole premise. But that only holds while the runtime answers inside the SDK's 30s `BATCH_RESULT_TIMEOUT`. Past it `handle_messages` stops waiting, NACKs, increments the batch id and polls again, so the timed-out batch stays in the channel and the next one joins it. `MAX_CONSECUTIVE_NACKS = 5` looks like the ceiling, but `apply_batch_result` resets the counter on any `Ack`, so an alternating ack/timeout run never trips it.

That is the half of this PR that outlived its premise, and it was tracked nowhere once this closed. It is now #3981, together with the second-order finding from the same review: a batch that times out and then sends successfully has its ack refused on a `take_pending_batch` id mismatch, so the runtime logs a delivery failure, bumps the error counter, and marks the connector errored for a batch that reached Iggy intact.

#3941 covers the other end of the same policy, the five-NACK stop and its invisibility. If `BatchPolicy` becomes configurable there, `result_timeout` is the third knob.

PR #3797, opened 2026-08-02
fix(connectors): hand off SDK worker during blocking send callbacks

### LINKED ISSUE #3796: SDK: hand off the plugin-runtime worker during blocking source send callbacks
## Context

#3795 bounded the source forwarding channel. When the channel is full, the runtime's send callback (`handle_produced_messages` -> `send_with_backpressure` in `core/connectors/runtime/src/source.rs`) parks in bounded `send_timeout` waits until capacity frees, the connector shuts down, or the channel disconnects. That park is the backpressure mechanism and is intentional.

The callback runs synchronously inside the SDK's polling loop (`core/connectors/sdk/src/source.rs`, the `callback(plugin_id, messages.as_ptr(), messages.len())` call in `handle_messages`), which is spawned on the plugin library's process-global tokio runtime (`core/connectors/sdk/src/lib.rs`, `static RUNTIME: OnceLock<Runtime>`). Every instance loaded from the same `.so` shares that runtime's workers.

## Problem

A parked callback holds one worker for the duration of its backpressure episode. With enough saturated instances of one plugin library, all workers can be occupied, and `iggy_source_close` for a sibling instance is delayed until a channel drains: the close blocks on the sibling's polling task, which needs a free worker to observe its shutdown signal. `signal_shutdown_all()` covers process shutdown (added in #3795), but the runtime API path that stops a single connector has no bound when its same-library siblings are saturated. The limitation is documented in the `send_with_backpressure` comment and in the connector-runtime skill.

## Proposed fix

Hand the worker off in the SDK before invoking the callback:

```rust
// core/connectors/sdk/src/source.rs, inside handle_messages
tokio::task::block_in_place(|| callback(plugin_id, messages.as_ptr(), messages.len()));
```

Notes:

- The wrap must live in the SDK, not the runtime: `block_in_place` consults the calling thread's tokio context, which is only set on the plugin runtime's own worker threads.
- The SDK runtime is multi-thread (`Runtime::new()`), which `block_in_place` requires.
- No FFI or ABI change. Existing plugin binaries keep working and pick the fix up when rebuilt against the updated SDK.

## Alternatives considered (in #3795 review)

- Signaling all same-library siblings during a single-connector stop: unblocks the close but drops sibling batches that are merely backpressured, punishing healthy instances.
- A wall-clock deadline on the retry loop: bounds the park but converts sustained backpressure into data loss, which defeats the purpose of the bounded channel.

Refs: #3795, discussion #3039 (bounded-channel spec).

mlevkov:
## Summary

Closes #3796. The send callback runs synchronously inside the SDK's polling task (`handle_messages` in `core/connectors/sdk/src/source.rs`), on the tokio runtime shared by every connector instance loaded from the same plugin library (`static RUNTIME: OnceLock<Runtime>` in `lib.rs`). A callback that blocks pins one worker for the duration. With #3795's bounded forwarding channel, a callback legitimately blocks for backpressure while the channel is full, so enough saturated instances of one library can occupy all workers, and `iggy_source_close` for a sibling instance then waits behind them: the close blocks on the sibling's polling task getting scheduled to observe its shutdown signal.

## Change

One call site: wrap the callback invocation in `tokio::task::block_in_place`, so the worker is handed off before the callback runs and the runtime keeps scheduling sibling tasks regardless of how long the callback blocks.

Notes from the analysis in #3796:

- The wrap has to live in the SDK, not the runtime: `block_in_place` consults the calling thread's tokio context, which is only set on the plugin runtime's own worker threads.
- The SDK runtime is `Runtime::new()`, i.e. multi-threaded, which `block_in_place` requires.
- No FFI or ABI change. Existing plugin binaries keep working and pick the fix up when rebuilt against the updated SDK.

This composes with #3795 (bounded channel + shutdown signaling) but does not depend on it: any long-blocking callback benefits.

## Test plan

- `cargo clippy -p iggy_connector_sdk --all-targets --all-features -- -D warnings` clean
- `cargo test -p iggy_connector_sdk --all-features` passes
- `cargo build -p iggy_connector_stdout_sink -p iggy_connector_random_source -p iggy-connectors` (macro consumers + runtime rebuild cleanly)
- Behavior under saturation is exercised end to end by the integration suite in CI

### DIFF
--- a/core/connectors/sdk/src/source.rs
+++ b/core/connectors/sdk/src/source.rs
@@ -196,7 +196,15 @@ async fn handle_messages<T: Source>(
                     }
                 };
 
-                callback(plugin_id, messages.as_ptr(), messages.len());
+                // The callback can block for backpressure (the runtime parks
+                // it while its bounded forwarding channel is full), and this
+                // task runs on the runtime shared by every instance loaded
+                // from this library. Hand the worker off so a saturated
+                // instance cannot pin it and stall a sibling's close, which
+                // waits on this task exiting.
+                tokio::task::block_in_place(|| {
+                    callback(plugin_id, messages.as_ptr(), messages.len());
+                });
             }
         }
     }


mlevkov:
Holding this pending a decision on #3795, because `af9ce9548` (#3855) removed its justification.

The comment this PR adds says the callback can block "while its bounded forwarding channel is full". On master `c0c74931b` that channel is `flume::unbounded()` (`runtime/src/source.rs:625`), so the send never parks. `handle_messages` also awaits the batch-result oneshot immediately after the callback returns (`sdk/src/source.rs:278`), which yields the worker at that point anyway, so `block_in_place` buys nothing on top of it.

The backpressure this protects against exists only if #3795's bounded channel lands, and I have argued over there that the bound is no longer needed. The `SendCallback` signature also gained `batch_id`, which is the conflict showing here.

Separately, the discipline this PR was arguing for has already landed in the direction that genuinely blocks: the runtime invokes `batch_result_callback` inside `tokio::task::spawn_blocking` (`runtime/src/source.rs:559`) because the plugin runs its async `on_batch_result` to completion inside that FFI call.

So my read is that this closes along with #3795 unless you want the bound, and I am happy to reopen if the bounded channel comes back. Should #3796 close too, or be reframed around the 30s timeout case I described in #3795?

mlevkov:
Closing this alongside #3795, per your call on Discord.

Without the bounded channel there is no backpressure for `block_in_place` to hand off: master's `flume::unbounded()` send never parks, and `handle_messages` awaits the batch-result oneshot immediately after the callback returns, which yields the worker at that point anyway. The discipline this PR was arguing for also already landed in the direction that genuinely blocks, since the runtime invokes `batch_result_callback` inside `tokio::task::spawn_blocking` because the plugin runs its async `on_batch_result` to completion inside that FFI call.

#3796 is still open. Happy to close it too, or reframe it around the 30s timeout case, whichever you prefer.

Issue #3688, opened 2026-07-17
add apache Fluss source connector


ryerraguntla:
### Description

Add a **source connector** that reads from [Apache Fluss](https://fluss.apache.org/) tables and publishes records into Iggy streams/topics.

Fluss is streaming storage for real-time analytics and lakehouse hot tiers. Its official Rust client ([`fluss-rs`](https://crates.io/crates/fluss-rs)) uses **Apache Arrow** (`RecordBatch` / `GenericRow`) end-to-end. That aligns with Iggy's connectors subsystem and with **Arrow-friendly** downstream sinks (for example Iceberg).

Use cases:

- Bridge Fluss log-table changelogs into Iggy for fan-out to multiple consumers or SDKs.
- Ingest Fluss primary-key table CDC-style changes into Iggy for event-driven apps.
- Unify Fluss hot-tier data with Iggy-based routing, transforms, and connector egress.

### Affected area / component

Connectors

### Proposed solution

Implement `iggy_connector_fluss_source` as a `cdylib` plugin implementing `iggy_connector_sdk::Source`, following conventions from `postgres_source` and `influxdb_source`.

### Client dependency

Use [`fluss-rs`](https://github.com/apache/fluss-rust) (Tokio async, Arrow-native). Pin a released version in the workspace `Cargo.toml` when adding the crate.

### Table modes (phased)

| Phase | Fluss table type | Read API | Notes |
| ----- | ---------------- | -------- | ----- |
| 1 | Log table | `subscribe` + `poll` / batch scanner | Offset-ordered append-only; primary use case |
| 2 | Primary-key table | Changelog / scan APIs as exposed by `fluss-rs` | Upsert/delete events; confirm API stability in 0.1.x |

### Message mapping

Default: decode Fluss rows to **`Schema::Json`** via existing connector JSON encoders (one object per row, column names from Fluss schema).

Optional (config flag): emit **`Schema::Raw`** Arrow IPC bytes per `RecordBatch` for zero-copy analytics pipelines (consumers decode Arrow client-side).

### Offset / state

Persist Fluss log offsets (and partition/bucket identifiers if applicable) in connector state via the runtime state API (`msgpack`), same pattern as other sources. On restart, resume from last committed offset.

### Draft configuration

```toml
[plugin_config]
bootstrap_servers = "localhost:9123"
database = "mydb"
table = "events"
table_type = "log"          # "log" | "primary_key"
starting_offset = "earliest" # "earliest" | "latest" | explicit offset
[...]

PR #3799, opened 2026-08-02
feat(connectors): add Apache Fluss source connector

seokjin0414:
## What

Adds a source connector that reads an [Apache Fluss](https://fluss.apache.org/)
log table and publishes each row into an Apache Iggy stream as JSON. This is
phase 1 of #3688: log tables and `Schema::Json`.

## How it works

The connector resolves the table schema while opening, so a missing table, a
primary-key table, a partitioned table or a column type with no JSON form all
fail at startup rather than once per batch. The log scanner is created once in
`open()` and reused, which works because it owns its client state rather than
borrowing the connection.

Offsets are tracked per bucket and returned in every `ProducedMessages`,
including empty polls, so the runtime can persist them. Since the batch
acknowledgment change (#3855) they are staged during `poll` and only become the
committed position once the runtime delivers an `Ack` through
`on_batch_result`. A `Nack` discards the staged offsets and rewinds the scanner
to the last acknowledged ones, because the scanner has already advanced past
the rejected records internally and would otherwise skip them until a restart;
fetches buffered from the old position are discarded by the scanner's own
expected-offset check. On restart, buckets already present in the state keep
their offset and only new ones fall back to `starting_offset`, which keeps a
widened bucket count from rewinding buckets that were already consumed. Each
message carries an id derived from its bucket and offset so Apache Iggy can
deduplicate after an at-least-once replay.

Column projection is pushed down to the server when `columns` is set. Temporal
values keep their Fluss-native integer form, because Fluss carries no timezone
that a formatted string could honour.

## The packaging question I asked in #3688

`fluss-rs` compiles its protocol definitions in a build script, so the workspace
now needs a system `protoc`. I raised this in
https://github.com/apache/iggy/issues/3688#issuecomment-5156005392 and said I
leaned towards registering the crate as a regular workspace member, so that is
what the last two commits do: `protobuf-compiler` goes into the shared Rust
setup action, which every Rust workflow already uses, and CONTRIBUTING gets the
prerequisite.

If you would rather keep it out of the workspace build via `exclude`, only
`build: register the Apache Fluss source in the workspace` and
`ci: install protoc ...` need reworking. The connector and its test do not
change. Happy to flip it.

Update: the upstream fix landed - apache/fluss#3874 checks the generated code
into `fluss-rs`, so the next crates.io release drops the `protoc` requirement
entirely. Once that release exists, the two commits above disappear and this PR
becomes a plain workspace member with no system dependency.

## Testing

Unit tests cover config validation, the state round-trip, offset resolution and
the row-to-JSON mapping (including nulls, base64 for binary and non-finite
floats).

The integration test appends rows to a real Fluss log table and asserts on what
comes back out of the Apache Iggy topic, offsets and bucket included. The
cluster runs as a single container: the image ships `local-cluster.sh`, which
starts the same embedded ZooKeeper, coordinator and tablet server, but it
rewrites the tablet server's bind port to 0 so it cannot collide with the
coordinator, and a random port inside the container cannot be published to the
host. Giving the tablet server an explicit second port keeps both reachable and
lets both advertise `localhost`, which then resolves the same way inside the
container and from the test process.

## Notes on scope

Two limits come from the released `fluss-rs` 0.1.0 rather than from the
connector, and each is rejected at startup with the reason rather than silently
ignored:

- Primary-key changelog scanning is not in the release, so `table_type =
  "primary_key"` is refused.
- `payload_format = "arrow_ipc"` is doable (0.1.0 does expose an Arrow
  `RecordBatch` scanner) but it is a separate scanner with its own
  offset-tracking path, so I left it for a follow-up.

`starting_offset = "latest"` is supported: buckets without persisted state get
their tail offset from `FlussAdmin::list_offsets(.., OffsetSpec::Latest)` at
open time (verified against a real 0.9.1 cluster: an empty bucket reports 0 and
the offset moves to the row count after appends, so subscribing there yields
only new records).

Refs #3688

### DIFF
--- a/.github/actions/utils/setup-rust-with-cache/action.yml
+++ b/.github/actions/utils/setup-rust-with-cache/action.yml
@@ -64,7 +64,9 @@ runs:
 
     - name: Install system dependencies (Linux)
       if: runner.os == 'Linux' && inputs.install-system-dependencies == 'true'
-      run: ./scripts/ci/apt-install.sh libhwloc-dev pkg-config libudev-dev
+      # protobuf-compiler: fluss-rs compiles its protocol definitions in a build
+      # script, so the Apache Fluss connector needs protoc at build time.
+      run: ./scripts/ci/apt-install.sh libhwloc-dev pkg-config libudev-dev protobuf-compiler
       shell: bash
 
     - name: Install system dependencies (macOS)
@@ -76,6 +78,9 @@ runs:
         curl -fsSL https://raw.githubusercontent.com/Homebrew/homebrew-core/bb1e23f8e5eacf4d31acd489f6079c8a53ebd690/Formula/h/hwloc.rb \
           -o "$(brew --repository iggy/local-hwloc)/Formula/hwloc.rb"
         brew install iggy/local-hwloc/hwloc
+        # protobuf-compiler equivalent: fluss-rs compiles its protocol definitions in
+        # a build script, so the Apache Fluss connector needs protoc at build time.
+        brew install protobuf
       shell: bash
 
     - name: Setup Rust toolchain
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,6 +38,18 @@ for review.
 
 One PR = one thing. Bug fix, refactor, feature - separate PRs. Mixed PRs will be closed.
 
+### System Dependencies
+
+Building the workspace needs `protoc` on the `PATH`, because the Apache Fluss
+connector's client compiles its protocol definitions in a build script.
+
+```bash
+sudo apt-get install -y protobuf-compiler   # Debian / Ubuntu
+brew install protobuf                       # macOS
+```
+
+Alternatively, point `PROTOC` at an existing binary.
+
 ### Quality Checks
 
 For Rust code:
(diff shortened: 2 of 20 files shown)


slbotbm:
Would be good to convert this to a draft until upstream changes are merged. This will get autoclosed otherwise

seokjin0414:
Rebased onto latest master and adapted to the source batch acknowledgments (#3855): offsets are now staged in `poll` and committed on `Ack`; a `Nack` rewinds the scanner to the last acknowledged offsets so a rejected batch is read again instead of skipped. Unit + Docker e2e pass locally, CI is green.

Upstream apache/fluss#3874 landed, so the next fluss-rs release drops the protoc requirement. Keeping this as a draft until then; once it ships I'll drop the two protoc commits and mark this ready for review.

seokjin0414:
fluss-rs 1.0.0 is out with the generated protobuf code from apache/fluss#3874, so the protoc commits are gone and this is now a plain workspace member with no CI change.

Rebased onto master and picked up the newer source rules on the way (no state on empty polls, a bounded retry on the rewind). Two fixes came out of that pass: the start offsets resolved for `latest` are now persisted before the first row arrives, and a batch that fails to build rewinds the scanner instead of skipping its rows. Both have a new integration test against Fluss 1.0.0, and CI is green.

@hubcio this is the one the upstream change was for, ready for review when you have time.

PR #3800, opened 2026-08-02
feat(csharp): migrate to VSR wire protocol

lukaszzborek:
The .NET SDK previously spoke classic TCP framing, which the clustered
next-generation server does not understand. Classic framing is now
dropped entirely: the SDK speaks VSR framing only, so every connection
goes through the register handshake, with session lifetime tied to the
connection, request sequencing, bounded reply decoding, transient
replays, leader redirection over the cluster roster, and
assignment-based consumer groups.

The IIggyClient surface is unchanged. Partitioning moves client-side
(balanced and message-key, hash byte-compatible with the Rust SDK),
credentials are bounds-checked before framing, and a request written
without a server verdict surfaces as VsrRequestOutcomeUnknownException
rather than being silently replayed on a new session, since that would
bypass server-side deduplication.


numinnex (dismissed):

spetz (dismissed):

hubcio:
please remove `e2e-vsr` and make `e2e` use `iggy-server-ng` - we're dropping support for `iggy-server`.

mmodzelewski (approved):

hubcio (approved):

Issue #3802, opened 2026-08-02
fix(connectors): runtime config API returns plugin credentials with auth disabled by default


mlevkov:
## Summary

The connectors runtime control API returns plugin configuration verbatim, credentials included, and its authentication is disabled by default. The default bind is loopback, so this is a hardening gap rather than a remote disclosure — but nothing warns an operator who moves the listener off loopback, which is a normal thing to do when running the runtime in a container.

Filing publicly rather than to the security list because of the loopback default. Happy to move it if maintainers judge otherwise.

## Configuration returns secrets

`core/connectors/runtime/src/api/source.rs::get_source_plugin_config` serves `source.config.plugin_config` as-is:

```rust
let (content_type, config) = map_connector_config(config, format)?;
```

`plugin_config` is the raw `serde_json::Value` parsed from TOML, so every credential an operator configured comes back in the clear. `GET /sources/{key}/configs` and `/configs/{version}` and `/configs/active` return the enclosing `SourceConfig`, which carries the same field. The sink routes mirror all of it. A repo-wide grep finds no redaction anywhere in `core/connectors/runtime/src` — the only `redact` hits are log-line truncation in `stream.rs`.

For most connectors that means a database connection string. For `iggy_connector_http_source` (#3798) it also means a `management_token`, which mints webhook endpoints, and each endpoint's HMAC secret.

## Authentication is off by default

`core/connectors/runtime/src/api/auth.rs::resolve_api_key`:

```rust
if context.api_key.expose_secret().is_empty() {
    return Ok(next.run(request).await);
}
```

`core/connectors/runtime/config.toml` ships:

```toml
address = "127.0.0.1:8081"
api_key = "" # Optional API key for authentication to be passed as `api-key` header
```

So out of the box the API is unauthenticated. Loopback confines that to local processes, which is a defensible posture for an admin API — the gap is that the two defaults compose into "any local process can read every connector credential", and an operator who changes `address` to `0.0.0.0` to reach the API from outside a container gets an unauthenticated endpoint serving credentials with no warning at any layer.

## Suggested fixes, roughly in order of value

1. Redact credential-bearing fields in the config responses, or gate those specific routes behind a configured `api_key` regardless of the global default.
2. Log a warning at startup when `api_key` is empty and `address` is not loopback.
3. Document in the runtime README that the control API returns credentials in plaintext and should be treated as privileged.

Related: #3801, which is why the plugin-side `SecretString` annotations do not help here.

[...]

PR #3803, opened 2026-08-02
docs(connectors): fix the inverted serde_secret redaction claim

### LINKED ISSUE #3801: docs(connectors): serde_secret helpers expose secrets, but the connectors guidance says they redact
## Summary

The connectors guidance tells plugin authors that annotating a `SecretString` field with `iggy_common::serde_secret::serialize_secret` gives them redaction. It does the opposite: the helper calls `expose_secret()` and writes the value in plaintext. Nine connector plugins follow that guidance today.

## The claim

`.claude/skills/connectors-overview/SKILL.md`, "Secrets":

> `secrecy::SecretString` + `iggy_common::serde_secret::serialize_secret` | `sinks/postgres_sink::PostgresSinkConfig::connection_string` | **Auto-redact on Debug/Display + serialization**

## The code

`core/common/src/utils/serde_secret.rs`:

```rust
pub fn serialize_secret<S: serde::Serializer>(
    secret: &SecretString,
    serializer: S,
) -> Result<S::Ok, S::Error> {
    serializer.serialize_str(secret.expose_secret())
}
```

Its own module doc says the opposite of the skill:

> These helpers are for fields that **must** be serialized (e.g., wire protocol payloads, persisted TOML configs, API responses that already expose credentials by design).
>
> Do **not** add `serialize_with` to fields that should remain redacted in serialized output — rely on `SecretString`'s default behavior instead.

And its test `serialize_optional_secret_with_some_value` asserts the output is `{"token":"tok_123"}`.

The `Debug` half of the claim is true — `SecretBox`'s `Debug` prints `[REDACTED]`. Only the serialization half is inverted.

## Why it matters

`SecretString` deliberately has no `Serialize` impl, so a config struct holding one cannot derive `Serialize` at all. Adding `serialize_with` is what *unblocks* the derive — it converts a compile-time guarantee into plaintext output, while the documentation says it does the reverse.

Nine plugins carry the annotation on credential fields:

`sinks/{postgres,s3,influxdb,elasticsearch,mongodb,surrealdb}_sink`, `sources/{postgres,elasticsearch,influxdb}_source`.

In practice the attribute appears inert: the runtime keeps plugin configuration as `serde_json::Value` and never deserializes into a plugin's config struct, so nothing calls these serializers today. The problem is that the protection authors believe they have does not exist, and the first caller that serializes a plugin config leaks every credential in it.

## Suggested fix

1. Correct the SKILL.md row — the helper exposes; `SecretString`'s default (no `Serialize` impl) is what redacts.
2. Consider dropping `Serialize` from plugin config structs that do not need it. `iggy_connector_http_source` (#3798) does this, which makes the property compiler-enforced rather than convention-enforced.
3. If a config struct genuinely needs `Serialize`, a `serialize_redacted` helper alongside the existing one would give authors the thing the docs currently promise.

Found while preparing #3798.

mlevkov:
Closes #3801.

## What was wrong

`.claude/skills/connectors-overview/SKILL.md` told plugin authors that annotating
a `SecretString` field with `iggy_common::serde_secret::serialize_secret` made
`Debug` **and serialization** redact. Only the `Debug` half is true. The helper
calls `expose_secret()` and writes the plaintext:

```rust
pub fn serialize_secret<S: serde::Serializer>(
    secret: &SecretString,
    serializer: S,
) -> Result<S::Ok, S::Error> {
    serializer.serialize_str(secret.expose_secret())
}
```

`serde_secret.rs`'s own module doc already said the opposite of the skill
("Do **not** add `serialize_with` to fields that should remain redacted"), so the
two documents contradicted each other and the skill is the one plugin authors
read.

The inversion matters more than a typo would, because `SecretString` has no
`Serialize` impl *by design* — a struct holding one cannot derive `Serialize` at
all. Adding the attribute is what **unblocks** the derive. The guidance therefore
recommended the exact step that converts a compile-time guarantee into plaintext
output, while describing it as protection.

Nine plugins carry the annotation on credential fields. It is inert today (the
runtime keeps plugin config as `serde_json::Value` and never deserializes into a
plugin's config struct, so nothing calls these serializers), so this is a
correctness-of-documentation fix rather than a live leak — but the protection
those authors believe they have does not exist.

## What this changes

**`docs(connectors)`** — the Secrets section now says which half of the claim was
true, and gives the default: *do not derive `Serialize` on a plugin config
struct at all*, since nothing needs it and leaving it off makes the property
compiler-enforced rather than convention-enforced.
`iggy_connector_http_source` (#3798) does exactly that.

It also notes that none of this protects the credential from the runtime's own
control API, which returns plugin configuration verbatim — that is #3802, and it
is not addressable from the plugin side.

While in there: the "In-tree uses" list named `delta_sink`, which does not use
these helpers, and omitted `s3_sink` and `surrealdb_sink`, which do.

**`feat(common)`** — adds `serialize_redacted` and `serialize_optional_redacted`
so the corrected guidance has something to point at. Suggestion 3 in the issue.
Without them, an author who genuinely needs `Serialize` has no redacting option
and reaches for the exposing one, which is the trap. The optional form keeps
`Some` distinguishable from `None`: whether a credential is configured is not
itself secret, and collapsing it to `null` would report a configured field as
unset. Documented as not round-tripping, so nobody feeds redacted output back
into a config loader.

## Deliberately not in scope

Suggestion 2 in the issue — dropping `Serialize` from the nine plugin config
structs that do not need it. It is the right follow-up and it is safe, but it
touches nine crates and each needs checking for a real serializing caller, so it
does not belong in the same review as the documentation fix. Happy to do it as a
separate PR; say the word and I will.

## Verification

`cargo fmt --all --check`, `cargo sort --check --no-format --workspace`,
`cargo clippy -p iggy_common --all-features --all-targets -- -D warnings`,
`cargo test -p iggy_common serde_secret` (6 pass), `taplo fmt --check`,
`hawkeye check`, `typos`, `markdownlint`, trailing whitespace/newline — all exit
0.


.claude/skills/connectors-overview/SKILL.md
```diff
@@ -96,14 +96,21 @@ The connectors codebase is intentionally repetitive across plugins. Cross-plugin
...
+
+**`serde_secret::serialize_secret` EXPOSES the secret. It does not redact.** It calls `expose_secret()` and writes the plaintext. `SecretString` deliberately has no `Serialize` impl, and that absence is the protection - so adding `serialize_with` is what *unblocks* the derive and turns a compile-time guarantee into plaintext output. Use it only where the plaintext is the point: a wire payload, a persisted config, an API response that exposes credentials by design.
+
+So the default for a plugin config struct is **do not derive `Serialize` at all**. The runtime keeps plugin configuration as the `serde_json::Value` it parsed from TOML and never deserializes into a plugin's config struct, so nothing needs the impl. Leaving it off makes the property compiler-enforced instead of convention-enforced (`sources/http_source/src/lib.rs::HttpSourceConfig` does this, and comments the omission so nobody adds it back).
```
hubcio: two wording nits in this sentence. the plugin itself does deserialize into its config struct - the sdk glue calls `serde_json::from_str::<C>` with a `DeserializeOwned` bound, so `Deserialize` stays required and "never deserializes" reads wrong; the load-bearing fact for dropping the impl is that nothing ever re-serializes the struct. and `plugin_config` doesn't only come from TOML - the control API accepts it as JSON and env vars can inject it too.

.claude/skills/connectors-overview/SKILL.md
```diff
@@ -113,7 +120,11 @@ let pool = PgPoolOptions::new()
     .await?;
 ```
 
-In-tree uses: `sinks/{postgres,mongodb,elasticsearch,influxdb,delta}_sink`, `sources/{postgres,elasticsearch,influxdb}_source`.
+If a config struct genuinely needs `Serialize`, `serde_secret::serialize_redacted` (and `serialize_optional_redacted`) write `[REDACTED]` in place of the value. Reach for `serialize_secret` only when the caller must get the real thing back. The sinks and sources listed below predate that helper and use the exposing one; the annotation is inert today, but it is not the protection it looks like.
+
+Note that none of this protects the credential from the runtime's own control API, which returns plugin configuration verbatim - see #3802. Plugin-side annotations are inert there because the runtime never routes through them.
+
+In-tree uses of the exposing helpers: `sinks/{postgres,mongodb,elasticsearch,influxdb,s3,surrealdb}_sink`, `sources/{postgres,elasticsearch,influxdb}_source`.
```
hubcio: this reads as an exhaustive inventory but only covers plugins. `runtime/src/api/config.rs` also puts `serialize_secret` on `HttpConfig::api_key` (equally inert - nothing serializes the runtime config today), and a handful of core/common wire-payload types (login, create-user, change-password, PAT) use the exposing helpers by design. scoping the sentence to plugin-side uses is probably the cleanest fix.

core/common/src/utils/serde_secret.rs
```diff
@@ -17,22 +17,35 @@
...
 //! #[serde(serialize_with = "crate::utils::serde_secret::serialize_secret")]
 //! pub password: SecretString,
 //! ```
 //!
-//! Do **not** add `serialize_with` to fields that should remain redacted in
-//! serialized output — rely on `SecretString`'s default behavior instead.
+//! [`serialize_redacted`] and [`serialize_optional_redacted`] write
+//! [`REDACTED`] in place of the value, for a struct that must be serializable
+//! for unrelated reasons but whose credential no reader is entitled to.
+//! Redacted output does not round-trip: deserializing it yields the literal
```
hubcio: worth a paired `deserialize_with` that rejects the `[REDACTED]` literal, so redacted output fed back into a config loader fails loudly instead of becoming the actual secret? nothing in-tree can hit this today - the redacting helpers have no consumers yet and the only persist/reload path round-trips the raw `serde_json::Value` - so a doc warning may be enough for now. question is whether to make it mechanical now or when the first consumer shows up.

hubcio (changes requested):
follow-up outside this diff: `runtime/src/api/config.rs` Debug impl hardcodes `"[REDACTED]"`, which now duplicates the new `REDACTED` const. same literal is also hardcoded in core/common auth credentials, server state models and the s3_sink URL redaction - worth a small sweep to the const later.

hubcio (changes requested):
-

mlevkov:
Thanks, both wording findings were correct and the second one was load-bearing rather than cosmetic. Pushed `3cd136fe3`.

**`Deserialize` vs re-serialize.** You are right, and my sentence would have led an author to drop the wrong derive. `sdk/src/{sink,source}.rs` do `serde_json::from_str::<C>` under a `DeserializeOwned` bound, so the plugin's struct is exactly what gets deserialized into. The guidance now says derive `Deserialize` but not `Serialize`, and names the property that actually carries the argument: nothing ever re-serializes the struct. Also corrected that `plugin_config` reaches the runtime as TOML, as JSON posted to the control API, or from env.

**The inventory.** Scoped to plugin-side callers, and it now says explicitly that the others are not all mistakes, naming `HttpConfig::api_key` and the `core/common` login / create-user / change-password / PAT payloads where the credential is the payload by design. Framing those as oversights was the worse half of that sentence.

**On the paired `deserialize_with`:** I have left it as a documented warning, strengthened to name the failure directly rather than implying it, and I would rather add the mechanism with its first consumer. Reasoning, in case you disagree:

A `deserialize_with` guard is itself opt-in, so it does not close the hole mechanically. An author who writes `serialize_with = serialize_redacted` and forgets the matching `deserialize_with` is precisely the case it claims to cover, and they get the silent placeholder-as-secret anyway. The shape that cannot be half-applied is a newtype owning both directions, and I would rather design that against a real call site than guess at one. This PR already adds two helpers with no consumers; a third feels like the wrong direction until something needs it.

Happy to add either form now if you would rather have it mechanical.

**On the `REDACTED` const sweep** (`api/config.rs` Debug impl, core/common auth credentials, server state models, s3_sink URL redaction): agreed, and I read your note as out of scope here. Say the word and I will open it as its own PR rather than widen this one.

Gate: fmt, sort, clippy `-D warnings`, `cargo test -p iggy_common serde_secret` (6 pass), taplo, hawkeye, typos, markdownlint all exit 0.

mlevkov:
The red on this PR after the rebase is a Maven Central rate limit, not the diff. Flagging it so it does not read as a real failure.

`BDD • bdd-java` failed while Gradle resolved its own plugin classpath:

```
> Could not resolve org.jetbrains.kotlin:kotlin-stdlib:2.3.21.
   > Could not GET 'https://repo.maven.apache.org/maven2/org/jetbrains/kotlin/
     kotlin-gradle-plugin-api/2.3.21/kotlin-gradle-plugin-api-2.3.21.pom'.
     Received status code 429 from server: Too Many Requests
BUILD FAILED in 26s
```

It never reached a step; the failure is in `:java:buildSrc` configuration. The five other BDD jobs are `CANCELLED` by fail-fast rather than failed on their own, and `finalize_pr` fails downstream of those. Everything else on the run is green: 91 passing, 1 skipped.

This PR touches `.claude/skills/connectors-overview/SKILL.md` and adds two additive `pub fn`s plus a doc comment to `iggy_common::utils::serde_secret`. No Java, no wire format, no server behaviour. The same rebase onto `01a64b2e0` produced fully green runs on my other four PRs, which went through the same BDD matrix minutes earlier.

I cannot re-run it myself (`Must have admin rights to Repository`). A re-run should clear it. If you would rather not spend the click, say so and I will force-push an identical tree to re-trigger, though that would change the SHA I quoted above.

Local gate on the rebased branch is unchanged and green: fmt, sort, clippy `-D warnings`, `cargo test -p iggy_common` 271 passing, license-headers, typos, taplo, all exit 0.

.claude/skills/connectors-overview/SKILL.md
mlevkov: Fixed in `3cd136fe3`, carried through the rebase. Verified the SDK glue you described: `sdk/src/{sink,source}.rs` call `serde_json::from_str::<C>` under a `DeserializeOwned` bound, so the guidance now says derive `Deserialize` but not `Serialize`, and names the property that actually carries the argument (nothing re-serializes the struct). Also corrected that `plugin_config` arrives as TOML, as JSON posted to the control API, or from env.

.claude/skills/connectors-overview/SKILL.md
mlevkov: Fixed in `3cd136fe3`. Scoped to plugin-side callers, and the sentence now says the other uses are not mistakes, naming `HttpConfig::api_key` and the `core/common` login / create-user / change-password / PAT payloads where the credential is the payload by design.

core/common/src/utils/serde_secret.rs
mlevkov: Left as a doc warning for now, strengthened to name the failure rather than imply it. My reasoning is in the PR comment: a paired `deserialize_with` is itself opt-in, so an author who forgets it hits exactly the case it claims to cover. A newtype owning both directions is the only shape that cannot be half-applied, and I would rather design that against a real consumer than guess. Happy to add either form if you disagree.

hubcio (approved):

numinnex (approved):

PR #3809, opened 2026-08-03
chore(docs): core/sdk lib.rs and IggyClient docs

### LINKED ISSUE #3306: Extend Rust SDK Docs: lib.rs and high-level API
### Description

1. Add a lib.rs level docs to onboard (new) users.
2. Extend docs for high-level API (IggyClient, IggyConsumer, IggyProducer)

### Component

Rust SDK

### Proposed solution

_No response_

### Alternatives considered

_No response_

haubur:
Draft PR as discussed here: https://discord.com/channels/1144142576266530928/1144142825630474360/1507361396382240839

Would be nice to get some feedback on:
- language (tone towards the user)
- understandability (level of detail)
- guidance (easy to follow examples)

If the general direction is good and supported, happy to clean and merge.
Would suggest to continue with IggyConsumer and IggyProducer in a next PR.

Closes #3306

## Rationale
Adding User facing docs for the Rust SDK.

## What changed?
- proposal for crate level docs in lib.rs
- docs for the IggyClient type

## Local Execution
- prek run (specifically typo and rust fmt)

## AI Usage
Everything is manually written or at least re-written in cases I had claude suggest a formulation.


core/common/src/traits/partitioner.rs
```diff
@@ -21,8 +21,18 @@ use crate::types::message::IggyMessage;
 use std::fmt::Debug;
 
 /// The trait represent the logic responsible for calculating the partition ID and is used by the `IggyClient`.
-/// This might be especially useful when the partition ID is not constant and might be calculated based on the stream ID, topic ID and other parameters.
+///
+/// Iggy uses a hierarchical model for append-only logs. A stream contains topics which hold partitions. Each partition is an append-only log.[^note]
+/// A producer of messages such as an [`IggyProducer`], that appends messages to the log, might want to choose to which partition to write the messages.
```
slbotbm: >  might want to choose to which partition to write the messages.

might want to choose which partition to write the messages into.

core/common/src/traits/partitioner.rs
```diff
@@ -21,8 +21,18 @@ use crate::types::message::IggyMessage;
...
+/// To do that, a producer can take a type that implements this trait.
+/// This might be especially useful when computing the partition ID requires some client side info, i.e. stream ID, topic ID and [`IggyMessage`] attributes.
```
slbotbm: > This might be especially useful

This may be especially useful

slbotbm: > stream ID, topic ID and [`IggyMessage`] attributes.

stream ID, topic ID or [`IggyMessage`] attributes.

core/common/src/traits/partitioner.rs
```diff
@@ -21,8 +21,18 @@ use crate::types::message::IggyMessage;
...
+///
+/// Note, that the [`Partitioning`] of a producer defines what _partitioning strategy_ is triggered on the server.
+/// Using a [`Partitioner`] in a producer sets the strategy in to request a specific partition [`PartitioningKind::PartitionID`] calculated with [`Partitioner::calculate_partition_id()`].
```
slbotbm: The usage of [`Partitioner`] in a producer allows the user to set a strategy that will request a specific partition [...] calculated using [...].


While I think the above is a better formulation, I think it would be better to completely change the sentence since it is difficult to understand and is missing information

core/common/src/types/message/partitioning.rs
```diff
@@ -25,11 +25,17 @@ use std::{
...
+/// A type that defines a what strategy the server should choose to partition the messages.
+///
+/// Iggy uses a hierarchical model for append-only logs. A stream contains topics which hold partitions. Each partition is an append-only log.[^note]
+/// A producer of messages such as an [`IggyProducer`], that appends messages to the log can choose between three partitioning strategies.
 /// - `Balanced` - the partition ID is calculated by the server using the round-robin algorithm.
-/// - `PartitionId` - the partition ID is provided by the client.
 /// - `MessagesKey` - the partition ID is calculated by the server using the hash of the provided messages key.
+/// - `PartitionId` - the partition ID is provided by the client.
+///
+/// Note, that using a [`Partitioner`] on top of [`Partitioning`] sets the strategy to [`PartitioningKind::PartitionId`] and the value to whatever [`Partitioner::calculate_partition_id()`] computes.
```
slbotbm: "to whatever [`Partitioner::calculate_partition_id()`] computes." implies that the calculation is non-deterministic. It would be better to phrase this in a way that allows the user to understand how `calculate_partition_id` calculates the partition

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// `IggyClient` wraps a transport-specific low-level **client** ([`ClientWrapper`]) and
+/// **provides access to the full server API**.
+/// Iggy comes with four options for client-server communication: TCP, QUIC, WebSocket and HTTP.
+/// The `IggyClient` is configured with one of these transport modes, hence abstracting
+/// transport specific implementations away.
+///
+/// The [`ClientWrapper`] lives behind an [`IggyRwLock`] so that one connection
+/// can be shared safely. You create a single client and use it from many tasks
+/// at once (producers, consumers, the background heartbeat), and they all talk
+/// over the same connection. The lock keeps that safe. Many operations can read
```
slbotbm: "The lock keeps that safe" is not required.

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// from the connection at the same time, while actions that reshape it, like
+/// connecting, reconnecting, or logging in, briefly take exclusive access. It is
+/// also async-aware, so holding it across a call to the server never blocks the
+/// runtime.
```
slbotbm: I'd ideally rephrase this paragraph. There is a lot of repetition here and the paragraph can be shortened.

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// A [`Partitioner`] and a client-side [`EncryptorKind`] are optional, and both
+/// default to disabled. The [`Partitioner`] computes on the client-side the target
+/// partition for messages published without an explicit partition. Hence, routing
+/// can depend on the stream, topic, and/ or message contents.
+///
+/// The [`EncryptorKind`] encrypts each message payload before it leaves the client and decrypts it on
+/// the way back, keeping payloads opaque to the server. Attach either through
+/// [`create`].
+///
+/// # What you can do
```
slbotbm, line 82: What you can do -> What can be done

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// - [`StreamClient`]: create, get, update, delete, and purge streams.
+/// - [`TopicClient`]: create, get, update, delete, and purge topics within a stream.
+/// - [`PartitionClient`]: add and remove partitions on a topic.
+/// - [`SegmentClient`]: delete closed segments from a partition.
+/// - [`ConsumerGroupClient`]: create, get, delete, and join or leave consumer groups.
+/// - [`ConsumerOffsetClient`]: store, read, and delete consumer offsets.
+/// - [`MessageClient`]: send and poll messages, and flush the unsaved buffer.
+///
+/// Additionally, you can bypass invoking methods from these traits and directly talk binary to the server with [`send_binary_request`] and [`send_http_request`] for http.
+/// Both are essentially backdoors for low-level control. You need to know the server codes and the wire format.
```
slbotbm: > /// Additionally, you can bypass invoking methods from these traits and directly talk binary to the server with [`send_binary_request`] and [`send_http_request`] for http.

Additionally, you can bypass invoking methods from these traits and directly talk to the server with [`send_binary_request`] and [`send_http_request`] for http.

> /// Both are essentially backdoors for low-level control. You need to know the server codes and the wire format.

The "backdoor" framing looks wrong. Maybe think of another way of saying this?

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///
+/// # Usage
+///
+/// The typical lifecycle of an [`IggyClient`] is construct [`connect`], use, and finally [`shutdown`].
```
slbotbm: > construct [`connect`], use, and finally [`shutdown`].

construct, [`connect`], use, and finally [`shutdown`].

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///
+/// 1. Construct a client from a connection string ([`from_connection_string`]),
+///    from the [`builder`], or by wrapping an existing transport client with
+///    [`new`] / [`create`].
+/// 2. Call [`connect`] to establish the connection. If the transport was
+///    configured with auto-login, this also authenticates. Otherwise call
+///    [`UserClient::login_user`] afterwards.
```
slbotbm: > Otherwise call [`UserClient::login_user`] afterwards.

"Otherwise call [`login_user`] afterwards." seems better

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// 3. Spawn [`IggyConsumer`]s and [`IggyProducers`] to write and consume messages
+///    to and from the server.
```
slbotbm: > 3. Spawn [`IggyConsumer`]s and [`IggyProducers`] to write and consume messages to and from the server.

Spawn [`IggyConsumer`]s and [`IggyProducer`]s to write to, and consume messages from, the server.

core/sdk/src/clients/client.rs
```diff
@@ -55,9 +55,159 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// 4. Call [`shutdown`] (or drop the client) to release resources; drop also
+///    leaves any consumer groups the client had joined.
```
slbotbm: You talk about drop leaving any consumer groups behind here, but do not talk about what shutdown does to consumer groups.

slbotbm (changes requested):
Will read more later

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
 // under the License.
 
+//! Apache Iggy is a high-performance, persistent message streaming platform written in Rust,
+//! capable of processing millions of messages per second with ultra-low latency.
+//! It is part of the [`Apache Incubating Program`] of the [`Apache Software Foundation`] (ASF).
+//!
+//! **This library is the Apache Iggy SDK.**
+//! It exposes a low-level and a high-level API for the Apache Iggy message streaming infrastructure for the Rust programming language.
+//! SDKs for other programming languages can be found in [`core/foreign`] of the root repository on GitHub.
+//! Note, most of them wrap this SDK. Hence, newer features might be delayed in other languages.
```
slbotbm: This line does not seem required here.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! The core of the Iggy server is a persisted append-only log data structure.
+//! It is concerned with allowing read and writes in the most efficient way.
+//! Reading and writing to the server is the domain of this SDK.
+//! The server exposes *commands* that can be triggered to change it's state.
```
slbotbm: "that can be triggered to change it's state." -> "that can be triggered to change its state."

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//! These commands allow administrative tasks, such as handling users, permissions and setting up streams and topics
+//! or writing and reading messages from the log.
+//! A comprehensive overview of commands can be found in the [`schema spec`] on the website or checking the [`server command enum`] within the source code.
```
slbotbm: I think it would be better to also give links here. "schema spec" and "server command enum" do not help the reader understand where the information is located.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! The SDK provides tools to build production ready message-streaming applications.
+//! It exposes its functionality at two levels. The [high-level API](#high-level-api)
+//! is transport-agnostic and ships with the batching, retry, offset-tracking, and
+//! connection-management machinery a production application needs. The
+//! [low-level API](#low-level-api) is the set of concrete transport clients that
+//! speak the wire protocol directly and that the high-level API is built on top of.
+//! Start with the high-level API unless you have a specific reason not to.
```
slbotbm: "Start with the high-level API unless you have a specific reason not to." -> "It is recommended to start with the high-level API, and utilize the low level API in case the high-level API cannot satisfy your requirements.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! # High-level API
+//!
+//! The high-level API is most likely what you are looking for, especially if you are new to building
+//! message-streaming applications with Iggy.
+//! Clients provided by the high-level API already provide common message-streaming features, that
+//! you would otherwise need to build yourself.
```
slbotbm: "common message-streaming features, that" -> "common message-streaming features that"

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! # Choosing a high-level client
+//!
+//! There are three client types, layered from low to high level. They are not
+//! alternatives to pick between so much as a control plane and two data-plane
+//! helpers built on top of it.
```
slbotbm: These two sentences are hard to understand -- rephrasing them would be optimal

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//! from an [`IggyClient`] with [`IggyClient::producer`] and
+//! [`IggyClient::consumer`] so they share its connection.
+//!
+//! ## When to use each
+//!
+//! Reach for [`IggyClient`] directly for administrative tasks such as
+//! creating streams, topics, users, and consumer groups, reading or storing
+//! offsets, or sending and polling a handful of messages in a script. Anything
+//! [`IggyClient`] cannot do, neither can the producer or consumer, since both
+//! are built on to of it.
```
slbotbm: "Anything
//! [`IggyClient`] cannot do, neither can the producer or consumer, since both
//! are built on to of it."

is not required, I think

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! Reach for [`IggyProducer`] and [`IggyConsumer`] when producing and consuming messages.
+//! You could use the [`IggyClient`] for that, however the former two come with some
+//! additional features already implemented that are frequentlly required when building messages
```
slbotbm: frequently

slbotbm: I think the line "features already implemented that are frequently required when building messages streaming applications" is repeated from above. Would be good to de-duplicate

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! - A [`futures::Stream`] implementation, so a `while let Some(message) =
+//!   consumer.next().await` loop drives polling, paging, and the poll interval
+//!   for you.
+//! - A **polling strategy** (`next`, `offset`, or `timestamp`) that tracks
+//!   position instead of taking an offset on every call.
+//! - **Auto-commit** and offset storage on an interval or after a number of
+//!   messages, so a restart resumes where it left off.
+//! - **Auto-join** of consumer groups, assignment refresh and reconnection
+//!   handling, and payload **decryption**.
```
slbotbm: Rephrasing this would be better, difficult to understand.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//!
+//! # Stream builder API
+//!
+//! The stream builder API is a convenient way to use the high-level API.
+//! [`IggyStream`], [`IggyStreamProducer`], and
+//! [`IggyStreamConsumer`] are construct everything at once.
```
slbotbm: Did not understand what you're trying to convey here.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//! by the same [`IggyClient`].
+//!
+//! ## When to use it
+//!
+//! Reach for the stream builder when you want a producer and consumer wired up
+//! with the least ceremony, especially a matched pair on one topic, or when you
+//! prefer to keep stream configuration in one declarative object instead of
+//! spread across imperative builder calls. When you need finer control over
+//! construction, drop down to [`IggyClient::producer`] and
+//! [`IggyClient::consumer`] and configure the builders directly. The stream
```
slbotbm: "drop down to [`IggyClient::producer`] and [`IggyClient::consumer`] and configure the builders directly" -> "the usage of [`IggyClient::producer`] and [`IggyClient::consumer`] is recommended"

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,260 @@
...
+//! builder offers nothing the high-level API cannot, since it is built entirely
+//! on top of it.
```
slbotbm: Would be better to replace this negative sentence with a positive one

core/sdk/src/lib.rs
slbotbm, line 1: There seems to be a lot of repetition of stuff here. Would be better to dedupe.

slbotbm (dismissed):

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,237 @@
...
+//! - A [`futures::Stream`] implementation, so a `while let Some(message) =
+//!   consumer.next().await` loop drives polling, paging, and the poll interval
+//!   for you.
+//! - A **polling strategy** (`next`, `offset`, or `timestamp`) that tracks
+//!   position instead of taking an offset on every call.
+//! - **Auto-commit** and offset storage on an interval or after a number of
+//!   messages, so a restart resumes where it left off.
+//! - **Auto-join** of consumer groupsa and assignment refresh should the server have
+//!   assigned the consumer another partition
+//! - Reconnection handling should the client disconnect
```
slbotbm: groups -> groups

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,237 @@
...
+//! [`IggyStreamConsumer`] construct everything at once.
+//! You can pass an [`IggyClient`] (or just a connection string) together with a config,
+//! and they hand back a ready, connected [`IggyProducer`] / [`IggyConsumer`].
+//! Compared to the **high-level API**, it changes how you construct
+//! producers and consumers, not what they can do. Instead of chaining an
+//! [`IggyProducerBuilder`] / [`IggyConsumerBuilder`] and setting each option
+//! with a method call, you describe the whole setup once in an
+//! [`IggyStreamConfig`] and build from it. The result is
+//! the same [`IggyProducer`] and [`IggyConsumer`] the builders produce, backed
+//! by the same [`IggyClient`].
```
slbotbm: change to
```rust
//! Compared to the **high-level API**, the stream builder simplifies construction
//! but exposes a smaller configuration surface. It returns the same
//! [`IggyProducer`] and [`IggyConsumer`] types, backed by the same [`IggyClient`],
//! but configures the producer in direct mode. Use [`IggyProducerBuilder`] when
//! you need background batching, sharding, backpressure, or in-flight controls.
```

slbotbm (commented):
other than two comments, lgtm

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// - [`PersonalAccessTokenClient`]: create, list, and delete personal access tokens, log in with one.
+/// - [`StreamClient`]: create, get, update, delete, and purge streams.
+/// - [`TopicClient`]: create, get, update, delete, and purge topics within a stream.
+/// - [`PartitionClient`]: add and remove partitions on a topic.
+/// - [`SegmentClient`]: delete closed segments from a partition.
+/// - [`ConsumerGroupClient`]: create, get, delete, and join or leave consumer groups.
+/// - [`ConsumerOffsetClient`]: store, read, and delete consumer offsets.
+/// - [`MessageClient`]: send and poll messages, and flush the unsaved buffer.
+///
+/// Additionally, you can bypass invoking methods from these traits and directly talk to the server with [`send_binary_request`] and [`send_http_request`] for http.
```
hubcio, line 105: also at lines 199, 457, 488, 587, 594, 650, 659, 717, 719, and partitioner.rs:26,30,31, partitioning.rs:31,36,37.

none of these links resolve, so `cargo doc` with `-D warnings` (the CI doctest job) fails on both crates. the variant is `PartitionId`, `iggy_common` can't link `IggyProducer`, and link defs don't carry across doc blocks.

core/sdk/src/clients/client.rs
```diff
@@ -182,7 +707,55 @@ impl IggyClient {
         ))
     }
 
-    /// Returns the builder for the producer.
+    /// Returns an [`IggyProducerBuilder`].
+    ///
+    /// Copies the client and the encryptor registered with the [`IggyClient`]
+    /// into a [`IggyProducerBuilder`] and returns it.
+    /// The [`Partitioner`] is applied by the client when messages are sent without
```
hubcio: also at lines 74-77, 485-488, 501 and partitioner.rs:23.

this partitioner never runs - `producer()` passes `None` at line 767 and `send_messages` never reads the field. either say it's inert and point at `IggyProducerBuilder::partitioner`, or drop these paragraphs.

core/common/src/traits/partitioner.rs
```diff
@@ -21,8 +21,19 @@ use crate::types::message::IggyMessage;
...
 /// The trait represent the logic responsible for calculating the partition ID and is used by the `IggyClient`.
-/// This might be especially useful when the partition ID is not constant and might be calculated based on the stream ID, topic ID and other parameters.
+///
+/// Iggy uses a hierarchical model for append-only logs. A stream contains topics which hold partitions. Each partition is an append-only log.[^note]
+/// A producer of messages such as an [`IggyProducer`], that appends messages to the log, may want to choose which partition to write the messages into.
+/// To do that, a producer can take a type that implements this trait.
+/// This is especially useful when computing the partition ID requires some client side info, i.e. stream ID, topic ID and/ or [`IggyMessage`] attributes.
+///
+/// Note the difference between [`Partitioning`] and [`Partitioner`]. [`Partitioning`] is a type used to set the _partitioning strategy_ for a producer.
+/// If you use both, the [`Partitioner`] overwrites the strategy, sets it to [`PartitioningKind::PartitionID`] and the partition ID is
```
hubcio: a producer-level partitioner also overrides the `Partitioning` passed to `send_with_partitioning` / `send_to` (`producer.rs:356-359` ignores the argument), not only the builder one. say 'overrides any partitioning, including per send'.

core/sdk/src/clients/client.rs
```diff
@@ -163,7 +643,52 @@ impl IggyClient {
...
 
-    /// Returns the builder for the consumer group.
+    /// Returns an [`IggyConsumerBuilder`] for a member of a consumer group.
+    ///
+    /// Copies the client and the encryptor registered with the [`IggyClient`]
+    /// into a [`IggyConsumerBuilder`] and returns it.
+    /// Sets the consumer to [`ConsumerKind::ConsumerGroup`], i.e. a member of a
+    /// load-balanced group. The provided name identifies the group.
+    /// Registers the member for every partition of `topic` in `stream`. The
+    /// group then balances those partitions across its members, so each message is
```
hubcio, line 667: delivery is at-least-once, not exactly one member - a rebalance moves polled-but-uncommitted partitions to another member, which re-reads them (`consumer_group.rs:141-148` spells this out). say 'each partition is served by one member at a time; uncommitted messages can be redelivered'.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///
+/// # Usage
+///
+/// The typical lifecycle of an `IggyClient` is construct, [`connect()`], use, and finally shutdown.
+///
+/// 1. Construct a client from a connection string ([`from_connection_string()`]),
+///    from the [`builder()`], or by wrapping an existing transport client with
+///    [`new()`] / [`create()`].
+/// 2. Call [`connect()`] to establish the connection. If the transport was
+///    configured with auto-login, this also authenticates. Otherwise call
```
hubcio, line 116: also at lines 350-353 and 875-891 (both http examples).

http ignores the connection-string credentials - `HttpClientConfig::from` drops them and `connect()` is a no-op - so both http examples send an empty bearer and get unauthorized. call `login_user()` first, and say http needs it.

core/sdk/src/clients/client.rs
```diff
@@ -75,19 +232,209 @@ impl Default for IggyClient {
...
+    /// # }
+    /// ```
+    ///
+    /// ## WebSocket
+    ///
+    /// - `heartbeat_interval`: [`IggyDuration`]. Client heartbeat period. Default: `5s`.
+    /// - `reconnection_retries`: "unlimited" or u32. Number of attempts to connect. Default: `unlimited`.
+    /// - `reconnection_interval`: [`IggyDuration`]. Wait between reconnection attempts. Default: `1s`.
+    /// - `reestablish_after`: [`IggyDuration`]. Grace period before reconnecting. Default: `5s`.
+    /// - `read_buffer_size`: usize. Size of the read buffer in bytes. Default: unset.
```
hubcio: these buffer sizes aren't unset, the tungstenite defaults apply: 131072 read/write, `usize::MAX` max write, 67108864 message, 16777216 frame. the example at line 382 sets `max_write_buffer_size=131072`, which panics tungstenite on connect - drop that key.

core/sdk/src/clients/client.rs
```diff
@@ -75,19 +232,209 @@ impl Default for IggyClient {
...
+    /// - `tls_domain`: string. Server name to validate the certificate against. Default: unset.
+    /// - `tls_ca_file`: filesystem path. Extra CA certificate to trust. Default: unset.
+    /// - `tls_validate_certificate`: bool. Whether to verify the server certificate. Default: `false`.
+    ///
+    /// ```no_run
+    /// use iggy::prelude::*;
+    ///
+    /// # fn run() -> Result<(), IggyError> {
+    /// let client = IggyClient::builder_from_connection_string(
+    ///     "iggy+ws://user:secret@localhost:8090\
```
hubcio: ws default port is 8092, 8090 is tcp.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///     .build();
+/// producer.init().await?;
+/// producer
+///     .send(vec![IggyMessage::from_str("our-first-message")?])
+///     .await?;
+///
+/// // A consumer pinned to one partition, committing its offset
+/// // automatically so a restart resumes where it left off.
+/// let mut consumer = client
+///     .consumer("consumer_name", "stream_name", "topic_name", 1)?
```
hubcio, line 173: the producer above writes `orders`/`created` but this consumer reads `stream_name`/`topic_name`, so `init()` fails with `StreamNameNotFound`. use the same stream and topic.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///     .build();
+/// consumer.init().await?;
+///
+/// while let Some(message) = consumer.next().await {
+///     let message = message?;
+///     // Handle `message.message.payload` here however required
+///     break;
+/// }
+///
+/// client.shutdown().await?;
```
hubcio, line 192: background producer with only `client.shutdown()` skips the drain step 4 above requires. add `consumer.shutdown().await?; producer.shutdown().await;` before this.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///         IggyExpiry::ServerDefault,
+///         MaxTopicSize::ServerDefault,
+///     )
+///     .build();
+/// producer.init().await?;
+/// producer
+///     .send(vec![IggyMessage::from_str("our-first-message")?])
+///     .await?;
+///
+/// // A consumer pinned to one partition, committing its offset
```
hubcio, line 170: also at lines 600, 613, 665, 678 and lib.rs:116-118.

`PollingMessages` commits on the poll request, before the app sees the batch, so a crash mid-batch skips those messages - that isn't 'resume where it left off'. use `ConsumingAllMessages` or say it's at-most-once; the builder default behaves the same.

core/sdk/src/clients/client.rs
```diff
@@ -75,19 +232,209 @@ impl Default for IggyClient {
...
+    ///
+    /// # Examples
+    ///
+    /// Each example lists configuration options per transport mode and shows
+    /// one concrete example.
+    ///
+    /// If the value of an option is
+    /// - a duration pass a `humantime` string such as `5s`, `500ms`, or `1h 1m 1s`.
+    ///   These are cast into an [`IggyDuration`].
+    ///   Note, `unlimited`, `none`, `disabled`, and `0` parse to zero.
```
hubcio: also at lines 283, 285, 306, 308, 343, 360, 362.

`heartbeat_interval` and `reconnection_interval` are `NonZeroIggyDuration` on every transport, so `0`/`none`/`unlimited` fail with `InvalidConnectionString` instead of parsing to zero. only `reestablish_after` takes zero.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,284 @@
...
+//! The high-level API is most likely what you are looking for, especially if you are new
+//! to building message-streaming applications with Iggy.
+//! High-level API clients already provide common message-streaming features that
+//! you would otherwise need to build yourself.
+//!
+//! There are three client types:
+//! - [`IggyClient`] is the entry point and the full API surface. It owns the
+//!   connection and implements every domain trait, including [`MessageClient`]
+//!   with the raw [`send_messages`] and [`poll_messages`] primitives. For both, each call
+//!   is a single, stateless request, i.e. there is no batching, retries, offset tracking,
```
hubcio: not stateless either - http retries transient failures (default 3, can resend a `POST /messages`), tcp reconnects and replays safe requests, and `auto_commit: true` stores the offset server-side. say 'no producer/consumer-level policy' instead.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,284 @@
...
+//!
+//! ## Differences to the high-level API
+//!
+//! - **Transport is fixed at compile time.** You name a concrete type
+//!   ([`TcpClient`], [`QuicClient`], and so on) instead of configuring a
+//!   transport-agnostic [`IggyClient`]. Swapping transports means swapping the
+//!   type, not changing a connection-string scheme.
+//! - **No managed connection.** [`IggyClient`] owns a shared connection and
+//!   spawns a heartbeat task to keep it alive. A transport client does neither.
+//!   You own the connection lifecycle and must ping the server
```
hubcio: transport clients do manage the connection: `TcpClient` (quic and ws too) reconnects with unlimited retries by default and re-logs-in. without `IggyClient` you lose the heartbeat task and the shared handle, nothing else.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,284 @@
...
+//! everything at once. You can pass an [`IggyClient`] (or just a connection string)
+//! together with a config, and they hand back a ready, connected
+//! [`IggyProducer`] / [`IggyConsumer`].
+//! Compared to the **high-level API**, it changes how you construct
+//! producers and consumers, not what they can do. Instead of chaining an
+//! [`IggyProducerBuilder`] / [`IggyConsumerBuilder`] and setting each option
+//! with a method call, you describe the whole setup once in an
+//! [`IggyStreamConfig`] (or in a single [`IggyProducerConfig`] /
+//! [`IggyConsumerConfig`] when you only need one side) and build from it. The result is
+//! the same [`IggyProducer`] and [`IggyConsumer`] the builders produce, backed
```
hubcio: the stream builder does limit what you get: `IggyProducerConfig` has no background mode, partitioner, or topic expiry/size, and `IggyConsumerConfig` has no `allow_replay` or `offset_drain_timeout`. say it covers a subset.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,284 @@
...
+//! Additionally it provides the following features:
+//!
+//! - **Retries** with a configurable count and interval.
+//! - A pluggable **partitioning strategy** ([`Partitioner`]).
+//! - **Encryption** of message payloads before they leave the process.
+//! - **Auto-creation** of the stream and topic if they do not exist yet.
+//!
+//! Only in **direct** mode:
+//! - **Chunking** splits an input larger than `batch_length` into several
+//!   requests, and `linger_time` sets a minimum delay between them.
```
hubcio: `linger_time` spaces consecutive `send()` calls, not chunks - the wait runs once before the first chunk and the chunks of one send go back to back (`producer.rs:429-457`).

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,284 @@
...
+//!
+//! Only in **background** mode:
+//! - **Batching** collects messages until the batch size in bytes, the message
+//!   count, or the linger interval is reached, whichever comes first.
```
hubcio: `batch_length` counts buffered `send()` calls, not messages - `producer_sharding.rs:182` compares `buffer.len()` and each entry holds a whole `Vec<IggyMessage>`. say 'number of buffered sends'.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
-/// It also provides the additional builders for the standalone consumer, consumer group, and producer.
+/// `IggyClient` wraps a transport-specific low-level **client** ([`ClientWrapper`]) and
+/// **provides access to the full server API**.
+/// Iggy comes with four options for client-server communication: TCP, QUIC, WebSocket and HTTP.
+/// The `IggyClient` is configured with one of these transport modes, hence abstracting
+/// transport specific implementations away.
+///
+/// The [`ClientWrapper`] lives behind an [`IggyRwLock`] so that the connection
+/// can be shared safely. You create a single client and use it from many tasks
+/// at once (e.g. producers, consumers). Many operations can read
```
hubcio: nothing takes the write lock - `connect`, `login_user` and `shutdown` all go through `read()`, the transports guard their own state. say every operation takes a shared read guard.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///    [`new()`] / [`create()`].
+/// 2. Call [`connect()`] to establish the connection. If the transport was
+///    configured with auto-login, this also authenticates. Otherwise call
+///    [`login_user()`] afterwards.
+/// 3. Spawn [`IggyConsumer`]s and [`IggyProducer`]s to write to, and consume messages
+///    from, the server.
+/// 4. To shut everything down, call [`IggyConsumer::shutdown()`] on each consumer to store their
+///    final offset and leave consumer groups. Then, call [`IggyProducer::shutdown()`] on each
+///    [`IggyProducer`] so that _background_ producers flush the latest state. Finally,
+///    call [`shutdown()`] on the [`IggyClient`] which closes the connection and stops the heartbeat.
```
hubcio: `shutdown()` doesn't stop the heartbeat, the task exits on its next tick when `ping` returns `ClientShutdown`. on http `shutdown` is a no-op and it runs until drop.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,166 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+///    Use [`disconnect()`] rather than [`shutdown()`] to close the connection but keep the client usable, as a
+///    client that has been shut down cannot reconnect.
```
hubcio: with auto-login the heartbeat reconnects and re-logs-in within one interval, so an explicit `disconnect()` gets undone. say so, or point readers at `shutdown()` (tcp/quic/ws only, http `shutdown` does nothing).

core/common/src/types/message/partitioning.rs
```diff
@@ -160,6 +167,9 @@ impl Partitioning {
     }
 
     /// Maximum size of the Partitioning struct
+    /// NOTE(haubur) I think this can be removed as it seems to be superseded
```
hubcio: personal note in shipped rustdoc, and the reason is off - `get_size_bytes` is a per-instance size, it doesn't replace a max bound. drop the note; if you want the fn gone, delete it in its own PR (it's pub).

hubcio (changes requested):

core/common/src/traits/partitioner.rs
haubur: I will highlight this again in the producer docs, specifically `send_with_partitioning` and `send_to`.

core/sdk/src/clients/client.rs
```diff
@@ -56,9 +56,173 @@ const SESSION_CONTROL_CODES: [u32; 5] = [
...
+/// transport specific implementations away.
+///
+/// The [`ClientWrapper`] lives behind an [`IggyRwLock`] so that the connection
+/// can be shared safely. You create a single client and use it from many tasks
+/// at once (e.g. producers, consumers). Every operation takes a shared read
+/// guard.
+///
+/// A [`Partitioner`] and a client-side [`EncryptorKind`] are optional, and both
+/// default to disabled. The [`Partitioner`] computes on the client-side the target
+/// partition for messages published without an explicit partition. Hence, you can
```
hubcio, line 74: also at lines 495-498 and `partitioner.rs:23`.

the client-level partitioner never runs - `producer()` passes `None` at line 787, so only `IggyProducerBuilder::partitioner()` takes effect.

core/sdk/src/clients/client.rs
```diff
@@ -75,19 +239,211 @@ impl Default for IggyClient {
...
+    ///
+    /// # Examples
+    ///
+    /// Each example lists configuration options per transport mode and shows
+    /// one concrete example.
+    ///
+    /// If the value of an option is
+    /// - a duration pass a `humantime` string such as `5s`, `500ms`, or `1h 1m 1s`.
+    ///   These are cast into an [`IggyDuration`]/[`NonZeroIggyDuration`].
+    ///   Note, `unlimited`, `none`, `disabled`, and `0` parse to zero.
```
hubcio: line 276 also names `NonZeroIggyDuration`, and there `0`, `none`, `disabled` and `unlimited` are rejected with `NonZeroDurationError::Zero`. only plain `IggyDuration` maps them to zero.

core/sdk/src/clients/client.rs
```diff
@@ -75,19 +239,211 @@ impl Default for IggyClient {
...
+    /// )?
+    /// .build()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// ## QUIC
+    ///
+    /// - `validate_certificate`: bool. Verify the server certificate. Default: `false`.
+    /// - `heartbeat_interval`: [`IggyDuration`]. Client heartbeat period. Default: `5s`.
```
hubcio: also at lines 315, 350, 368, 370.

these are `NonZeroIggyDuration` too, same as the tcp entries at 290 and 292.

core/sdk/src/clients/client.rs
```diff
@@ -75,19 +239,211 @@ impl Default for IggyClient {
...
+    /// - `tls_validate_certificate`: bool. Whether to verify the server certificate. Default: `false`.
+    ///
+    /// ```no_run
+    /// use iggy::prelude::*;
+    ///
+    /// # fn run() -> Result<(), IggyError> {
+    /// let client = IggyClient::builder_from_connection_string(
+    ///     "iggy+ws://user:secret@localhost:8092\
+    ///      ?heartbeat_interval=5s&reconnection_retries=unlimited&reconnection_interval=1s&reestablish_after=5s\
+    ///      &read_buffer_size=131072&write_buffer_size=131072&\
```
hubcio: the trailing `&` here plus the leading one on the next line gives `131072&&max_message_size`, which `websocket_connection_string_options.rs:109` rejects as an empty segment.

core/sdk/src/clients/client.rs
```diff
@@ -138,12 +566,78 @@ impl IggyClient {
...
+    /// To get builder for a load-balanced consumer group, use
+    /// [`consumer_group`](IggyClient::consumer_group) instead.
+    ///
+    /// Refer to the [`IggyConsumer`] type and the [`IggyConsumerBuilder`]
+    /// for details on how a consumer can be configured.
+    ///
+    /// # Examples
+    ///
+    /// Connect a client, build a consumer pinned to partition 1, and configure
+    /// it to auto-commit its offset so a restart resumes where it left off.
```
hubcio: also at line 683 and `lib.rs:116-118`.

`AutoCommitWhen::PollingMessages` commits inside the poll request, before the app sees the messages - a crash drops them rather than resuming where it left off.

core/sdk/src/lib.rs
```diff
@@ -15,6 +15,281 @@
...
+//! - **Encryption** of message payloads before they leave the process.
+//! - **Auto-creation** of the stream and topic if they do not exist yet.
+//!
+//! Only in **direct** mode:
+//! - **Chunking** splits an input larger than `batch_length` into several
+//!   requests.
+//! - **Spacing** applies `linger_time` between consecutive sends.
+//!
+//! Only in **background** mode:
+//! - **Batching** collects number of buffered sends until the batch size in bytes, the message
```
hubcio: `batch_length` counts buffered `send()` calls, not messages. the sentence is also missing a word - `collects number of buffered sends`.

core/common/src/types/message/partitioning.rs
```diff
@@ -160,6 +170,8 @@ impl Partitioning {
     }
 
     /// Maximum size of the Partitioning struct
+    /// by implementing Sizeable/ get_size_bytes().
```
hubcio: leftover fragment - `by implementing Sizeable/ get_size_bytes()` no longer attaches to anything.

hubcio (changes requested):

core/sdk/src/clients/client.rs
haubur: Solved with `self.partitioner.clone()` in producer() (see below).

hubcio (approved):

numinnex (approved):

hubcio (approved):

Issue #3747, opened 2026-07-24
feat(connectors): Implement RabbitMQ / AMQP source & sink


amr8t:
### Description

As part of the connector ecosystem roadmap: [#2753](https://github.com/apache/iggy/issues/2753) I want to add the RabbitMQ connector source + sink

### Affected area / component

Connectors

### Proposed solution

Add a native Rust source & sink connector for RabbitMQ (AMQP 0.9.1), enabling Iggy streams to publish messages to RabbitMQ exchanges (sink) and consume messages from RabbitMQ queues into Iggy streams (source).

### Contribution

- [x] I'm willing to submit a pull request to implement this feature

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

ryerraguntla:
@amr8t Thanks for opening the issue and owning up. While you implement, please produce two separate PRs - one for source and one for sink. Two separate PRs help reviewers. @hubcio @spetz

PR #3811, opened 2026-08-03
feat(connectors): add RabbitMQ sink

amr8t:

Relates to #3747 

## Sumary
This change adds the RabbitMQ sink connector via the lapin client. As requested, doing only sink connector in this PR. Source will be a separate one.

Adds Configurable exchange and various types (topic, direct, fanout).
Adds Configurable Retries with exponential backoff
Added Integration tests against RabbitMQ container covering topic, fanout and direct exchange behavior.

## Local Execution
Completed below based on https://github.com/apache/iggy/blob/master/CONTRIBUTING.md
- Quality Checks
- Typos Checks
- License Header Checks
- Pre-commit Hooks

AI Usage

    Which tools? opencode
    Scope of usage? investigation, code suggestions
    How did you verify the generated code works correctly? Read through, no agentic flow but write via refer, execute, write tests
     Can you explain every line of the code if asked? Yes


core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+#[derive(Debug)]
+struct RabbitMqState {
+    connection: Connection,
+    channel: Channel,
+}
+
+#[derive(Debug)]
+pub struct RabbitMQSink {
+    id: u32,
+    amqp_url: String,
```
slbotbm, line 46: (also applies to `RabbitMQSinkConfig::amqp_url`)

AMQP URLs normally include the username and password. This field is a `String` on a `Serialize` config type, so the sink's derived `Debug` output and any serialization of this type contain the URL. The runtime also exposes the raw `plugin_config` through its sink-plugin-config endpoint.

Store the URL as `secrecy::SecretString`, add `iggy_common::serde_secret::serialize_secret`, and use `ExposeSecret` only at the `Connection::connect` call sites, following the existing Postgres and MongoDB sinks. This prevents plugin-side logging and serialization leaks, but does not redact the runtime's raw `plugin_config`; that endpoint also needs a general redaction mechanism or access restriction.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+            ));
+            warn!(
+                "Transient RabbitMQ publish error for connector ID: {} (attempt {attempts}/{}): {error}. Retrying in {:?}.",
+                self.id, self.max_retries, delay
+            );
+            tokio::time::sleep(delay).await;
+        }
+    }
+
+    async fn reconnect(&self) -> Result<(), Error> {
```
slbotbm, line 229: (also applies to `RabbitMQSink::open`)

`ExchangeDeclareOptions::default()` declares a non-durable exchange. RabbitMQ rejects a declaration when an existing exchange with the same name has different durability attributes, closing the channel with `PRECONDITION_FAILED`. Consequently, a normal pre-created durable exchange cannot be used with this connector; the fixture masks this by declaring the same non-durable exchange.

Expose the declaration properties in the connector configuration with safe defaults, or declare an operator-managed exchange passively. Cover an existing durable exchange in the integration tests.

core/connectors/sinks/rabbitmq_sink/README.md
slbotbm, line 1: The README documents only the connection and routing fields. It omits `include_metadata`, `verbose_logging`, `max_retries`, `retry_delay_secs`, and `max_retry_delay_secs`, even though all are public plugin configuration. Operators therefore cannot discover how to disable generated headers or control retry behavior.

Document each supported field, its type, default, and behavior in the configuration table and TOML example.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
slbotbm, line 1: The sink sets `BasicPublishOptions::mandatory = true`, but increments `published` for every `Ok(_)` publisher confirmation. RabbitMQ acknowledges an unroutable mandatory publish with `Confirmation::Ack(Some(returned_message))`; the message has been returned, not routed to a queue. This code therefore returns `Ok(())` and lets the consumed Iggy message advance even though RabbitMQ delivered it nowhere.

Match the confirmation explicitly: only `Ack(None)` is success. Treat `Ack(Some(_))` and `Nack(_)` as errors, and add an integration test with a routing key that has no matching binding.

slbotbm: `BasicProperties::default()` leaves the AMQP delivery mode unset, which RabbitMQ treats as non-persistent. There is no configuration to request persistent delivery. Even if the exchange and queue are durable, RabbitMQ can discard a publisher-confirmed message on broker restart, while the sink has already reported it as successfully published.

Set persistent delivery mode by default or make it an explicit, documented configuration option. Test restart behavior with a durable exchange and queue.

slbotbm: The sink constructs a new `FieldTable` containing only a few generated metadata values and never reads `message.headers`. User headers are therefore lost on every publish. A headers exchange can route on the generated `iggy_*` values when metadata is enabled, but it cannot route on the original user-supplied headers.

Encode representable `message.headers` values into AMQP headers, using `ByteArray` for raw binary values rather than a lossy string conversion, then add an integration test that publishes a message with a custom header and routes it through a headers exchange.

slbotbm: `ConsumedMessage::offset` is `u64`, but the code converts it to `u32` and silently substitutes `u32::MAX` on overflow. Long-lived topics will consequently publish an incorrect offset header for every message after the first 4,294,967,295 offsets.

`lapin` does not expose an `AMQPValue::LongLongUInt`, so encode the full `u64` offset as a decimal `LongString` rather than narrowing it. Cover an offset above `u32::MAX` in a unit test.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+                        &self.routing_key,
+                        lapin::options::BasicPublishOptions {
+                            mandatory: true,
+                            ..Default::default()
+                        },
+                        &body,
+                        props,
+                    )
+                    .await
+                    .map_err(|e| Error::CannotStoreData(e.to_string()))?;
```
slbotbm, line 180: An error from `channel.basic_publish(...).await` is converted with `?` and returned immediately. It never sets `last_error`, reconnects, or uses the configured retry delay. Only errors while awaiting a publisher confirmation reach the retry path.

Route immediate publish errors through the same retry flow as confirmation errors. Preserve the index of the first unconfirmed message when retrying so this fix does not republish earlier confirmed messages.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+                    .map(|s| s.channel.clone())
+                    .ok_or_else(|| Error::Connection("RabbitMQ not connected".into()))?
+            };
+
+            let mut last_error: Option<Error> = None;
+            let mut published: u64 = 0;
+            for message in messages {
+                let body = message.payload.clone().try_into_vec()?;
+                let mut props = BasicProperties::default();
+                if self.include_metadata {
```
slbotbm, line 147: When a later message fails, `published` records how many earlier messages were confirmed, but the next loop iterates over the whole `messages` slice again. A transient error on the final message of a 100-message batch therefore republishes the first 99 confirmed messages. This duplication is introduced by the sink’s own retry loop, independent of any runtime retry behavior.

Resume at the first unconfirmed message after reconnecting, and document the remaining at-least-once case where connection loss makes the final publish outcome unknowable. Add a test that forces a failure after at least one confirmation.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+                guard
+                    .as_ref()
+                    .map(|s| s.channel.clone())
+                    .ok_or_else(|| Error::Connection("RabbitMQ not connected".into()))?
+            };
+
+            let mut last_error: Option<Error> = None;
+            let mut published: u64 = 0;
+            for message in messages {
+                let body = message.payload.clone().try_into_vec()?;
```
slbotbm, line 145: `consume` only has a borrowed message, but this clones the complete `Payload` before converting it into bytes. For `Payload::Json`, that needlessly deep-clones the `simd_json::OwnedValue` tree before serializing it, adding allocation and CPU cost to every published JSON message.

Use `message.payload.try_to_bytes()` to serialize JSON directly from the borrowed payload.

slbotbm (changes requested):

hubcio:
closing due to inactivity of author

amr8t:
Apologies for the delay, I was out for a week @hubcio are you able to open this?

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Good idea

amr8t:
replaced by #3973

PR #3824, opened 2026-08-05
fix(consensus): stop view change discarding committed ops

krishvishal:
## What this fixes

A committed, client-acknowledged op survived a primary crash only if the round-robin new primary happened to hold it.

`DoViewChange` carried scalars only, so the new primary took the winner's op number without its log, rebuilt its pipeline from its own journal, hit the local gap, and truncated the range as "decided lost". The op was journaled on a quorum, committed, and already replied to; the next client op reused the number and collided with the stale entry on an up-to-date backup. The comment justifying the truncation cited a DVC merge that did not exist.

`DoViewChange` now carries the sender's uncommitted suffix, a nack bit and a present bit per entry, and the new primary merges the quorum's headers. Discarding requires a nack quorum, which quorum intersection makes unreachable for anything that may have committed. Otherwise the view does not start: unavailable and saying so beats available and short a reply.

## Changes

Independent defects on the same path.

**Prepare identity.** `PrepareHeader.checksum` was never set, so `parent` chained zeros and every merge comparison was `0 == 0`. Now sealed on both planes and verified at all four ingress points, not just metadata repair: a frame corrupted in flight was journaled as-is and re-served, which the interior-corruption boot refusal turns into an unbootable node. The WAL scan checks it and the parent chain, since a flipped `commit` reaches recovery's `max(header.commit)` watermark. `view` is excluded, because `restamp_prepare_view` rewrites it in place.

**Zero, enforced two ways.** For prepares `CHECKSUM_UNSEALED` is 0 and means "skip", so existing WALs replay. For control headers a zero is a `FrameChecksumMismatch`: keying on "looks sealed" lets one flipped bit disable the layer by clearing the field that gates the check. Tolerant on disk, strict on the wire.

**Control frame integrity.** Canonical selection took the first sender scanned and read a differing header as an implicit nack, so one bit flipped in that sender's suffix turned every honest sender's correct header into a nack against the garbage: a nack quorum on three replicas, above `commit_max` where the refusal does not fire. Control headers now carry a checksum over every byte past it, verified on the typed parse before `validate` reads a field. Body verification keys on body presence, not on whether `checksum_body` looks sealed.

**Merge correctness.** Canonical selection consults every canonical sender and treats disagreement as undecidable, since senders at the canonical `log_view` cannot legitimately disagree when a primary prepares one thing per op. Only a sender behind that `log_view` may nack implicitly, a constraint the comment stated and nothing tested. The decoder recomputes each identity and checks view and timestamp monotonicity and the hash chain. A tripwire that panicked on same-`log_view` disagreement is gone; it fired on remote bytes.

**Quorum sizing.** Splitting the replication and view-change quorums to define the nack quorum surfaced a latent bug: at `replica_count` 4 both were 2, so they could be disjoint.

**`log_view` timing.** Raised when the merge parked, before the merged head was installed, so a primary-elect superseded or crashed mid-repair held a durable claim to a view whose headers it never had, then carried its own stale head as sole canonical sender of the next view change.

**Parked-log roles.** A backup's parked log is a verification reference, not a repair window. It doubled as the view's repair scope, so once traffic passed the view's opening head every repaired op was discarded unnamed and the backup never converged.

**Latched quorum flag.** Set on reaching a view-change quorum, not on deciding a log, so every non-Ready outcome was terminal: an `AwaitingRepair` waiting for more `DoViewChange` messages had guaranteed it would ignore them.

**Repair serving.** Bounded by the local frontier again. `to_op` comes from a peer with only `from_op <= to_op` validated and the skip loop walks op by op with no `await`, so on the single-threaded shard pump an unclamped `u64::MAX` ends the shard.

**Divergence reconciliation.** `Journal::truncate_from` drops a diverging uncommitted suffix and leaves the snapshot watermark alone. Not `drain`, which advances the watermark past what it removed: correct for a committed prefix, ruinous for a suffix, marking the ops that must stay refillable as evictable.

**Data at rest.** Disk polls verify each batch against its `batch_checksum` and fail closed. Consumer offset files gain a checksum, since the offset is a cursor reloaded unchanged on every restart and a flipped bit silently rewinds or skips a consumer. Bare pre-checksum files still load and upgrade on the next write, and the legacy server stays compatible both ways by reading the first eight bytes and stopping. Separable from the consensus work.


core/shard/src/lib.rs
```diff
@@ -5057,10 +5523,387 @@ where
...
+            view_headers.and_then(|headers| view_header_at(headers, entry_op))
+        {
+            // Held from the adopted view rather than from the journal, so the
+            // header is reported and the op is NOT nacked: this replica knows
+            // the op exists and simply cannot serve its body yet. No present
+            // bit for the same reason.
+            headers.push(*header);
+        } else {
+            headers.push(dvc_blank(entry_op));
+            if entry_op > commit {
```
numinnex, line 8435: `build_dvc_suffix` pushes a blank for every op it cannot resolve, but only nacks it when `entry_op > commit`. At `entry_op == commit` the entry goes out blank *and* un-nacked.

The merge scans inclusive of `commit_max` ("the commit point is scanned so the adopted log is anchored on a servable header"), so that entry is exactly the one it needs a verdict on. It gets neither a canonical header nor a nack, and falls into the `verdict.canonical.is_none() || verdict.copies == 0` branch in `dvc_merge.rs`:

```
2 of 3 senders blank at the commit point -> AwaitingRepair { undecided_op: 5 }
3 of 3 senders blank at the commit point -> Deadlocked    { undecided_op: 5 }
```

Measured against this branch with a scratch test recording `dvc(n, 1, 5, 5, DvcSuffix::new(vec![dvc_blank(5)], 0, 0))` for each sender.

`Deadlocked` is terminal, so once every replica has reported, the view never starts.

This is not the `op_head == commit_max` case in general: with a real suffix the commit-point header is present, and the merge returns `Ready`. It needs `header_at(commit)` to actually miss.

Whether that is reachable comes down to two things: can journal trimming advance the snapshot watermark to `commit_max` itself, and can a partition plane recover a non-zero `commit_max` with an empty in-memory journal? If the commit point is structurally pinned, the invariant is worth an assertion and a test, since nothing here enforces it. If it is not pinned, this un-nacked blank wedges the group.

krishvishal: Not pinned, both planes drop it, and the partition case doesn't even need a restart.  

**Metadata:** not the watermark, the drain. `SnapshotCoordinator::drain` was
  `drain(0..=snap_op)` with `snap_op = commit_min`, so the commit point was removed
  outright. Checkpoints fire on journal occupancy, which is symmetric across replicas.

  **Partitions:** `commit_messages` reads `committed_prefix(commit_max)`, it breaks on
  `header.op > commit_max`, inclusive — and `evict_prefix` clears those headers. A
  flushed replica has no resident header at its commit point with the process still
  running. The entry is still servable from the evicted ring; only the header lookup
  missed it, so the present bit was wrong too.

  A lagging sender saves the view (its floor op is above its own compaction point).
  The wedge needs the survivors to agree on the commit point.

core/shard/src/lib.rs
```diff
@@ -4322,137 +4434,517 @@ where
...
+            return;
+        }
+        let Some(pending) = consensus.pending_view_log() else {
+            return;
+        };
+        let Some(journal) = metadata.journal.as_ref() else {
+            return;
+        };
+
+        let held = |op: u64| {
```
hubcio: `held` answers by presence only - `journal.header(slot).is_some()` - never by identity. a primary-elect below the canonical `log_view` holding a *different* prepare at an op the merge kept sees `held() == true`, skips repair, and starts the view: `StartView` ships the merged (canonical) headers while the pipeline rebuild reads the local journal and self-acks the divergent entry. a backup that already holds the canonical header skips the append on retransmit (`on_replicate`'s already-journaled path) and acks the *received* header's checksum, and `handle_prepare_ok` compares that against the primary's own pipeline entry - both sides of the gate are the primary's header, so quorum forms and the divergent op commits while the up-to-date backup applies the canonical one. two replicas apply different operations at the same committed, client-acked op, silently.

reachable 2-of-3: the primary-elect contributes exactly one implicit nack, short of the nack quorum, so the canonical header survives the merge and nothing ever asks the primary to repair. the backup path below does the identity compare (`local.checksum == canonical.checksum`) that this closure omits, and `reconcile_metadata_view_divergence` returns early for the primary. `held` has to match the journaled header's checksum against `pending.headers`/`committed_elsewhere`, and a mismatching primary-elect needs the same truncate + repair a backup gets. the partition twin has the same presence-only check and doesn't consult `committed_elsewhere` at all.

core/consensus/src/view_change_quorum.rs
```diff
@@ -64,43 +239,196 @@ pub fn dvc_count(array: &DvcQuorumArray) -> usize {
...
+            if header.checksum != CHECKSUM_UNSEALED && header.identity_checksum() != header.checksum
+            {
+                return Err(DvcSuffixError::ChecksumMismatch { index });
+            }
+            if let Some(child) = child {
+                // Views never go backwards down the log, timestamps never forwards,
+                // and consecutive entries hash-chain. A frame breaking any of these
+                // describes a log that cannot exist.
+                if header.view > child.view {
+                    return Err(DvcSuffixError::ViewRegressed { index });
```
hubcio: the `ViewRegressed` check rejects a log this codebase legitimately produces. `restamp_prepare_view` rewrites `view` on the wire copy only, and `on_replicate` skips the append when the journal already holds the op - so a backup that held op 12 from view 1 but missed 11 and 13 appends those two at view 2 and keeps 12 at view 1. its next suffix decodes [13:v2, 12:v1, 11:v2], trips this check, and the whole `DoViewChange` is dropped before the view advance; every retransmit is byte-identical, so it drops every time. if that replica is needed for the view-change quorum, the view change stalls until something else moves.

gating on consecutiveness like `ChainBreak` below doesn't help - 11 and 12 are adjacent in that scenario. the premise "views never go backwards down the log" doesn't hold once skip-kept entries exist; this check has to go or be rethought. `TimestampNotDecreasing` next to it can fire across a skip-kept entry the same way - worth checking with the same scenario.

core/partitions/src/offset_storage.rs
```diff
@@ -98,17 +178,43 @@ async fn read_persisted_offset(path: &str) -> Result<Option<u64>, IggyError> {
...
+        } => {
+            tracing::error!(
+                path,
+                offset,
+                expected,
+                found,
+                "consumer offset file failed its checksum; refusing to fold a value that may \
+                 rewind or skip the consumer"
+            );
+            Err(IggyError::CannotReadConsumerOffsets(path.to_owned()))
```
hubcio: this `Err` ends in a shard panic. `persist_offset_max` propagates it, the commit walk turns the failed consumer-offset commit into the "replica is divergent from cluster commit; restart required" `panic!`, and the boot loader skips a corrupt file *without unlinking it* - while the in-memory offset map starts empty every boot, so the first auto-commit touching that key re-reads the file and panics again. one corrupt offset file is a persistent landmine: crash, restart, crash. only the auto-commit read path is affected (an explicit store blind-writes and repairs the file).

the two readers should agree: boot already treats corrupt as absent and keeps going. do the same here - log it, fold from `max_committed` alone, and never route a data-file checksum failure into the commit-walk fatal path.

core/partitions/src/poll_plan.rs
```diff
@@ -533,14 +536,23 @@ impl DiskReadPlan {
...
                 };
-                let consumed = walk_disk_chunk(
+                let ChunkWalk { consumed, corrupt } = walk_disk_chunk(
                     &chunk,
                     query,
                     count,
                     &mut matched,
                     &mut fragments,
                     &mut last_matching_offset,
+                    self.validate_checksum,
```
hubcio: `validate_checksum` ships `false`, and that flips the disk poll from fail-closed to fail-open. on master this path went through `decode_batch_slice`, which verified `batch_checksum` unconditionally and never served a failing batch; with the shipped default only the layout check runs and a corrupt batch is decoded and handed to the consumer. `read_disk_faults_closed_on_batch_checksum_mismatch` even asserts `Matched` with the knob off - a test asserting corrupt data is served. the PR text says "disk polls verify each batch against its `batch_checksum` and fail closed", which isn't what ships. flipping the default restores master's behavior and master's cost - the hash is on the order of the memcpy the poll already performs.

three adjacent things. `Faulted` maps to an empty poll with no error surfaced, and at-rest corruption never "reads again", so with the knob *on* the consumer stalls forever, indistinguishable from caught-up - wants an error to the client, independently of the default. the comment above the knob in server-ng's config.toml still calls it inert and claims boot warns when set (that warn got deleted). and the same key now means a one-time boot sweep on the legacy server vs a per-poll hash forever on ng - operator-facing text should say so.

core/consensus/src/dvc_merge.rs
```diff
@@ -0,0 +1,911 @@
...
+/// header repair walks the chain backwards and stops at a gap, so an op missing
+/// below the new primary's commit point can never be repaired into place and
+/// refusing it here strands the log permanently. The claim is already
+/// quorum-backed, since a sender cannot report an op committed unless a
+/// replication quorum held it.
+///
+/// `None` when two senders report *different* prepares committed at one op. These
+/// install unconditionally, so guessing is least affordable here; as in
+/// [`canonical_header_at`], the caller treats it as undecidable.
+fn committed_elsewhere(
```
hubcio, line 313: this takes one stale sender's word for it - no second sender, no nack quorum, no chain check against the canonical range. `dvc_commit` reports `commit_max.min(op)`, `commit_max` advances from the primary's claim without checking the local log matches, and reconcile deliberately leaves a divergent entry at or below `applied_floor` in place - so a replica can report commit >= X while holding a header that never committed, and ship it here on the next view change. nothing installs these headers into the log, but they poison the repair gate: the genuine repaired prepare then mismatches `expected.checksum` and gets discarded, and the view change stalls at that op. wants corroboration or a chain check before trusting the entry.

core/consensus/src/impls.rs
```diff
@@ -3062,15 +3485,25 @@ where
...
+        //
+        // So a partition prepare's `checksum` identifies its header alone, and two
+        // such prepares at one op with matching header fields are indistinguishable
+        // to the view-change merge. Closing that wants `checksum_body` here to BE
+        // the batch checksum, recomputed after stamping, so `checksum` covers the
+        // body for free by hashing this field.
+        //
+        // Bounded by `size`, the range every verifier re-reads; the prepare
+        // inherits it verbatim below.
         let checksum_body = if consensus.namespace == METADATA_CONSENSUS_NAMESPACE {
```
hubcio, line 3538: partition prepares still seal `checksum_body = 0`, so `checksum` covers the header alone and the batch bytes ride unverified. the comment above already says it, but the merge consequence is worth spelling out: the dvc merge counts a replica whose batch *bytes* differ as a servable copy of the canonical header, and partition repair ingest has no merged-log identity gate, so the body branch of `verify_prepare_integrity` short-circuits on the zero and a divergent batch passes on the serving peer's word. the fix the comment names (make this the batch checksum, recomputed after stamping) closes it - "sealed on both planes" in the PR text is true of the field, not yet of the coverage.

core/shard/src/lib.rs
```diff
@@ -4322,137 +4434,517 @@ where
...
-    #[allow(
-        clippy::future_not_send,
-        clippy::cast_possible_truncation,
+        // Truncation is safe only above what this replica has *applied*, which is
+        // not the view's commit point: `pending.commit_max` is the new primary's
+        // number and a backup can sit above it. Splitting on the view's number
+        // would drop already-executed ops with no rollback, and silently.
+        let applied_floor = pending.commit_max.max(consensus.commit_min());
+
+        let mut repairable_from: Option<u64> = None;
```
hubcio, line 4764: the reconcile only walks ops named in `pending.headers`, which spans commit..head - a local WAL suffix *above* the announced head is never truncated, while `set_sequence(announced)` drops the head under it. the next prepare at `op_head+1` then hits the append slot-collision error (it fires even when `existing.op == header.op`) and the backup wedges. that's the case the TODO in `handle_start_view` describes, and `truncate_from` is the tool for it - just not wired to this case. the subtle part: `repairable_from` only arms on an in-window checksum disagreement, so the miss is precisely the log that *agrees* in-window and is only stale above the head - the restart/re-adopt shape. needs a `journal last_op > pending.op_head` check, which wants a trait accessor (`last_op` is inherent on `PrepareJournal`).

core/shard/src/lib.rs
```diff
@@ -3476,7 +3515,16 @@ where
...
+        let Some(suffix_body) = control_suffix_body_verified(&msg, header.checksum_body) else {
+            tracing::warn!(
+                shard = self.id,
+                from_replica = header.replica,
+                view = header.view,
+                "dropping start_view whose body failed its checksum"
+            );
+            return;
+        };
+        let actions = consensus.handle_start_view(PlaneKind::Partitions, &header, suffix_body);
```
hubcio, line 3567: partition StartView adoption has no divergence reconciliation (the metadata arm calls it right after adopting), and `PartitionJournal` doesn't implement `truncate_from`, so it inherits the `Unsupported` default. the failure mode here is quieter and worse than the metadata plane's wedge: partition `append` has no slot-collision check - a re-prepared op pushes a duplicate header and rewrites `op_to_storage_offset` - and `committed_prefix` walks headers positionally, so the *stale* pre-divergence entry is what `evict_prefix` flushes to the segment. durable divergent bytes, no error anywhere. `clear_all` already exists for the state-transfer install; StartView adoption needs the equivalent.

core/journal/src/lib.rs
```diff
@@ -46,6 +46,28 @@ where
...
+    /// it removed, which would mark the removed ops evictable when a suffix
+    /// truncation needs them refillable.
+    ///
+    /// Defaults to `Unsupported` rather than a silent zero: "removed nothing" and
+    /// "cannot remove anything" demand different responses from the caller.
+    ///
+    /// # Errors
+    /// I/O error if the rewrite fails, or `Unsupported` if the implementation cannot
+    /// truncate.
+    fn truncate_from(&self, _from_op: u64) -> impl Future<Output = io::Result<usize>> {
```
hubcio: the `Unsupported` default hides missing impls. `JournalState` hand-forwards every other method - its own doc says to forward EVERY method because a wrapper that silently substitutes a default for its inner journal is a trap - but `truncate_from` isn't in its list, and `PartitionJournal` doesn't implement it either. dropping the default turns both into compile errors instead of a runtime `Unsupported` in the middle of a view change. two one-line impls, and the crate isn't published.

core/configs/src/server_ng_config/metadata.rs
```diff
@@ -60,10 +60,15 @@ pub const DEFAULT_METADATA_JOURNAL_SLOTS: usize = 1024;
...
-pub const MAX_METADATA_PREPARE_QUEUE_DEPTH: usize = 4096;
+/// Upper bound on `prepare_queue_depth`.
+///
+/// Pinned by the view-change wire format, not by memory: a `DoViewChange` carries
+/// the sender's uncommitted suffix plus one nack bit and one present bit per entry,
+/// each bitset a single `u128` (`consensus::DVC_HEADERS_MAX` = 128). The suffix
+/// spans `commit_max..=op`, which this depth bounds, so a deeper queue produces
+/// entries the new primary can neither adopt nor prove dead. The reserved head slot
+/// leaves room for the head op.
+pub const MAX_METADATA_PREPARE_QUEUE_DEPTH: usize = 127;
```
hubcio, line 71: the drop to 127 is right and load-bearing - `merge_commit_max` treats `dvc.op - ceiling` as proof of commit, so a peer configured deeper would manufacture a commit point it never reached. don't relax this under operator pressure later. but a deployment tuned above 127 now refuses to boot with "exceeds the maximum (127)" and no hint why; worth naming the cause in the validation message plus a release note. defaults are 32, so nobody hits it out of the box.

core/binary_protocol/src/consensus/error.rs
```diff
@@ -29,6 +29,15 @@ pub enum ConsensusError {
     #[error("invalid checksum")]
     InvalidChecksum,
 
+    #[error(
+        "{command:?}: header checksum {found:#034x} does not cover the frame (expected {expected:#034x})"
```
hubcio: a pre-seal peer's frame surfaces with `found` printing as 0x000...0, and the drop gets logged as "dropping message with invalid command" (router and shard both) even though the command parsed fine. mixed-version is a declared hard break, so the operator diagnosing it deserves better: retitle the drop log, and when `found == 0` add a hint that this looks like a peer predating the frame seal - that's the signature every pre-seal node sends.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -78,12 +105,82 @@ pub trait ConsensusHeader: Sized + CheckedBitPattern + NoUninit {
...
+    fn frame_checksum(&self) -> u128 {
+        let bytes: &[u8; HEADER_SIZE] = bytemuck::bytes_of(self)
+            .try_into()
+            .expect("every consensus header is HEADER_SIZE bytes");
+        frame_checksum_bytes(bytes)
+    }
+
+    /// Stamp [`Self::frame_checksum`]. Call last when building a frame: it covers
+    /// every other field, `checksum_body` included, so later writes are uncovered.
+    fn seal(&mut self) {
```
hubcio, line 155: calling `seal()` on a `FRAME_SEALED = false` type (a prepare) would silently overwrite the identity checksum with a frame checksum in release - the `debug_assert` is the only guard. no prepare goes through `seal()` today (`seal_prepare_checksum` writes the field directly), but making `FRAME_SEALED` a required const with no trait default turns that mistake into a compile error for free.

core/consensus/src/view_change_quorum.rs
```diff
@@ -64,43 +239,196 @@ pub fn dvc_count(array: &DvcQuorumArray) -> usize {
...
+        let beyond = !((1u128 << count) - 1);
+        if nack_bitset & beyond != 0 || present_bitset & beyond != 0 {
+            return Err(DvcSuffixError::BitsetBeyondSuffix { count });
+        }
+    }
+
+    let mut headers = Vec::with_capacity(count);
+    // The entry above the current one, skipping blanks: high-to-low, so the child.
+    let mut child: Option<PrepareHeader> = None;
+    for index in 0..count {
```
hubcio, line 377: none of these entries go through `PrepareHeader::validate` - the reserved-must-be-zero rule (one of the two reasons stated on `validate` is exactly the `dvc_blank` exact-equality classification) is never enforced at the one place that indexes blanks. an entry with a dirty reserved byte and `checksum == 0` also skips the identity recompute below, so it reads as a Valid header, conflicts with every honest sender's, and the merge refuses to pick a canonical header - the view change stalls and repeats on every escalation for as long as the entry survives. the DVC body is sealed and verified in transit, so the exposure is a crafted frame from an authenticated peer or a corrupt-at-rest unsealed header re-sealed on send - but the rule exists for exactly that class. same story on the repair door: `RepairPrepareHeader::validate` checks the command only. both ingress points want the full validate, or at least the reserved-zero check.

core/shard/src/lib.rs
```diff
@@ -7438,10 +7914,411 @@ where
...
+                    .any(|expected| expected.op == op)
+        })
+}
+
+/// Ceiling on the op range a repair request may ask this replica to walk.
+///
+/// Not `commit_max` alone: a new primary repairing toward a merged log needs the
+/// uncommitted suffix the view change kept, which sits above every commit point.
+///
+/// Bounded by the local frontier all the same. `RequestPreparesHeader::validate`
```
hubcio, line 8305: the doc here already names the weak `validate` and the no-await walk - this covers the half it doesn't: the ceiling bounds `to_op`, but nothing floors `from_op`. `RequestPreparesHeader::validate` only requires `1 <= from_op <= to_op`, and the metadata skip loop in `on_request_prepares` walks op by op from whatever the peer sent. probes are cheap but the loop is unbounded - `from_op = 1` against a big compacted frontier is a long non-yielding stretch on the pump, repeatable at will, against a 10 ms consensus tick. the partition arm already jumps straight to `retained_from`, which is exactly the shape: floor the metadata side at `snapshot_op() + 1` (and/or bound `to_op - from_op` in validate). the `to_op` raise to `max(commit_max, head)` is required - the merged suffix above commit_max has to stay servable - so the fix is the floor only.

core/shard/src/lib.rs
```diff
@@ -7438,10 +7914,411 @@ where
...
+            clamped - 1
+        );
+        low = clamped;
+    }
+
+    let len = usize::try_from(op - low + 1).unwrap_or(DVC_HEADERS_MAX);
+    let mut headers = Vec::with_capacity(len);
+    let mut nack_bitset = 0u128;
+    let mut present_bitset = 0u128;
+    for (index, entry_op) in (low..=op).rev().enumerate() {
```
hubcio, line 8419: the partition arm's `header_at` is `repair_header` - a linear `header_by_op` scan plus a linear walk of the evicted ring - probed per op across the window, so up to 128 x resident-header-count comparisons per rebuild. `repaired_window_shape`'s doc warns about exactly this shape, and the tick-gate comment nearby bounds the *frequency*, not the per-rebuild cost: this reruns on every SVC/DVC arrival and every non-Normal tick per partition (`local_dvc_suffix_stale` keys on `(op, commit)`, and repair advances it), so an active view change burns it repeatedly on the pump. a one-pass window builder over headers + ring, same shape as `repaired_window_shape`, kills the quadratic.

core/consensus/src/impls.rs
```diff
@@ -2818,6 +3137,110 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+            // config says nothing about how deep a peer's pipeline is.
+            prepare_queue_ceiling: PREPARE_QUEUE_CEILING as u64,
+        }
+    }
+
+    /// The merged log this replica is repairing toward, if a view change is
+    /// mid-transition. The shard reads it for the op range it must cover before the
+    /// view can start, and for which peers offered the bodies.
+    #[must_use]
+    pub fn pending_view_log(&self) -> Option<MergedLog> {
```
hubcio, line 3182: `pending_view_log()` clones the whole `MergedLog` (two `Vec<PrepareHeader>`) on every call, and most call sites want `.is_some()` (every metadata tick) or two u64s; the repair path clones once per repaired prepare. an `is_some()` accessor + a bounds accessor + a with-ref closure covers all but the two sites that hold the value across an await or across `start_pending_view` (which takes the cell and would trip the RefCell inside a closure).

core/consensus/src/impls.rs
```diff
@@ -2818,6 +3137,110 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+
+    /// Finish the parked view change: this replica's journal now covers the merged
+    /// log, so it can serve any op it is about to announce.
+    ///
+    /// Called by the shard after repair progress. No-op when nothing is parked.
+    ///
+    /// # Panics
+    /// If the merged uncommitted range exceeds pipeline capacity, which needs a head
+    /// more than one pipeline depth above the proven commit point.
+    pub fn start_pending_view(&self, plane: PlaneKind) -> Vec<VsrAction> {
```
hubcio, line 3227: this is the only exit from a view change that doesn't reset the DVC quorum - the other five paths all go through `reset_view_change_state()`. the winning primary therefore keeps `do_view_change_from_all_replicas`, now carrying a suffix `Vec` per sender, for its whole primaryship, per partition it leads. nothing reads the array after the view starts (a late same-view DVC returns at the status gate before touching it; a higher-view one resets anyway), so freeing the suffix storage here (`dvc_reset` + `invalidate_local_dvc_suffix`) is safe - but not `reset_dvc_quorum`, which also clears the `do_view_change_quorum` latch that now means "log decided".

core/shard/src/lib.rs
```diff
@@ -4322,137 +4434,517 @@ where
...
+                );
+                continue;
+            }
+            repairable_from = Some(repairable_from.map_or(canonical.op, |op| op.min(canonical.op)));
+        }
+
+        let Some(from_op) = repairable_from else {
+            return;
+        };
+        match journal.handle().truncate_from(from_op).await {
```
hubcio, line 4817: `truncate_from` shares the drain guard, so it's excluded against `drain` and other rewrites - but nothing excludes it against an append. that exclusion is metadata's `journal_gate`, a private field this crate can't reach, and `on_repair_prepare`'s append is in the same position. it appears to hold today by call-site placement (backups-only caller, submits gated on primaryship), which nothing documents or asserts - and detached submit tasks are explicitly named as concurrent drivers of the gated section, so the serialization isn't self-evident. worth routing both shard-side journal mutations through metadata-crate methods that take the gate, or at minimum a comment here recording why no append can interleave - the reasoning isn't visible at the call site.

core/consensus/src/plane_helpers.rs
```diff
@@ -169,6 +202,41 @@ where
...
+/// different checksum per replica, and the merge would read those as competing
+/// prepares nacking each other. Not merely a workaround either: two frames
+/// differing only in `view` ARE the same op, and everything that makes them
+/// genuinely different (client, request, timestamp, parent, operation, or the body
+/// via `checksum_body`) stays covered.
+///
+/// `checksum_body` must already be set, since it is how the body reaches this
+/// value. A prepare left unsealed there gets an identity over its header alone.
+#[must_use]
+pub fn prepare_identity_checksum(header: &PrepareHeader) -> u128 {
```
hubcio: one-line pass-through to `header.identity_checksum()` - `seal_prepare_checksum` ten lines down and the tests are the only users. the `restamp_prepare_view` rationale in the doc is the part worth keeping; it can live on `identity_checksum` itself and the wrapper can go.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -1023,10 +1297,52 @@ impl ConsensusHeader for DoViewChangeHeader {
...
+impl DoViewChangeHeader {
+    /// Number of `PrepareHeader`s in the body.
+    ///
+    /// Zero is valid and means "no suffix": a replica with nothing uncommitted, or a
+    /// peer predating the suffix. Both contribute numbers only.
+    ///
+    /// # Errors
+    /// [`ConsensusError::InvalidField`] when `size` is short of the header, is not a
+    /// whole number of headers, or exceeds what the bitsets can address.
+    pub fn suffix_len(&self) -> Result<usize, ConsensusError> {
```
hubcio, line 1351: body is token-identical to `StartViewHeader::suffix_len` except the label in the three messages - one free fn taking the frame name, two thin wrappers, and the bounds logic can't drift apart.

core/partitions/src/offset_storage.rs
```diff
@@ -98,17 +178,43 @@ async fn read_persisted_offset(path: &str) -> Result<Option<u64>, IggyError> {
...
-    let compio::BufResult(read, buf) = file.read_exact_at(buf, 0).await;
-    match read {
-        Ok(()) => {}
-        Err(error) if error.kind() == std::io::ErrorKind::UnexpectedEof => return Ok(None),
+    // Read the whole record, falling back to a bare offset: a pre-checksum file is
+    // exactly `OFFSET_SIZE` long, so the first read reports EOF rather than failing.
+    let compio::BufResult(read, buf) = file.read_exact_at(vec![0u8; OFFSET_RECORD_SIZE], 0).await;
+    let bytes = match read {
+        Ok(()) => buf,
+        Err(error) if error.kind() == std::io::ErrorKind::UnexpectedEof => {
```
hubcio: the 16-then-8 fallback costs a second syscall + alloc per legacy-file read just to learn the length. `decode_offset_record` already classifies any length, so one non-exact `read_at` into a 16-byte buffer and decoding `&buf[..n]` does it. the `..n` matters: decoding the zero-padded buffer would turn a legacy 8-byte file into `Corrupt`, and that's a hard `Err` out of `read_persisted_offset`, not a `None`.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -1023,10 +1297,52 @@ impl ConsensusHeader for DoViewChangeHeader {
...
+        }
         Ok(())
     }
 }
 
+impl DoViewChangeHeader {
+    /// Number of `PrepareHeader`s in the body.
+    ///
+    /// Zero is valid and means "no suffix": a replica with nothing uncommitted, or a
+    /// peer predating the suffix. Both contribute numbers only.
```
hubcio: "a peer predating the suffix" can't reach this decode anymore - the frame seal rejects a pre-seal peer's frame before any field is read (this file's module doc: replicas upgraded together, cluster down). same stale pre-seal-peer claims on `StartViewHeader::incarnation`, `RequestStartViewHeader::incarnation`, `StartViewHeader::suffix_len`, `view_change_quorum`'s decode comment, and the shard's suffix-verify comments. worth striking them all and pointing at the module doc - keeping zero-tolerance wording only where it's real, the on-disk `CHECKSUM_UNSEALED` path (which also covers partition-plane prepares, so that skip stays).

core/consensus/src/impls.rs
```diff
@@ -1885,16 +2051,9 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
-        let action = VsrAction::SendDoViewChange {
-            view: self.view.get(),
-            target: self.primary_index(self.view.get()),
-            log_view: self.log_view.get(),
-            op: current_op,
-            // commit_max clamped to op: see `handle_start_view_change`.
-            commit: self.commit_max.get().min(current_op),
-            namespace: self.namespace,
-        };
+        // The same snapshot the first send used. A retransmit that re-derived the
```
hubcio: "the same snapshot the first send used" - it isn't: `build_do_view_change` calls `local_dvc_suffix()` fresh each time, and the shard refreshes the suffix on every SVC/DVC handler and non-Normal tick. worse, on a tick-driven retransmit the `(op, commit)` tag can have moved with no refresh in between, and `local_dvc_suffix()` then returns an *empty* suffix - the retransmit silently retracts every nack and body offer. harmless today only because `dvc_record` drops duplicate senders. either pin the snapshot for real or fix the comment.

core/server_common/src/send_messages2.rs
```diff
@@ -678,6 +678,24 @@ fn transcode_legacy_request(
...
+/// `verify_checksum` exists for the disk-poll path, whose operator knob decides
+/// whether a read pays for a full re-hash of every batch. The layout checks are not
+/// optional either way: a short or self-inconsistent record is rejected regardless,
+/// because the caller would otherwise index past it.
+///
+/// # Errors
+/// [`IggyError::InvalidCommand`] for a short or inconsistent record, and
+/// [`IggyError::InvalidBatchChecksum`] when verification is on and the batch does not
+/// match. Callers that must tell corruption from a partial tail need both.
+pub fn decode_batch_slice_verified(
```
hubcio: the name promises "verified" but the second argument turns verification off - poll_plan passes the operator knob straight through, so the common disk-poll read runs `..._verified(.., false)`. rename, or take a two-variant enum instead of the bool so the call site reads honestly.

core/simulator/src/deps.rs
```diff
@@ -180,6 +180,33 @@ impl<S: Storage<Buffer = Vec<u8>>> Journal<S> for SimJournal<S> {
     /// ever superseded by a snapshot. Answered explicitly (the trait has no
     /// default) so a simulated state transfer has to opt into a watermark
     /// rather than silently inherit one that never moves.
+    /// Drop the suffix, so a simulated backup whose entries disagree with a started
+    /// view reconciles the way a real one does. Mirrors
+    /// `PrepareJournal::truncate_from`, whose watermark stays put; here it never moves.
+    async fn truncate_from(&self, from_op: u64) -> std::io::Result<usize> {
```
hubcio, line 186: this landed between `snapshot_op`'s doc comment and `snapshot_op` itself - the doc above now describes the wrong method and `snapshot_op` lost its doc. move it below `set_snapshot_op`.

core/partitions/src/offset_storage.rs
```diff
@@ -89,6 +165,10 @@ pub async fn persist_offset_max(
 /// files, so the commit-path reader must agree or a torn file turns every
 /// later commit-apply into an error. Real I/O errors still propagate: mapping
 /// them to `None` would silently rewind a valid higher offset.
+///
+/// A checksum mismatch is an error, not `None`. `None` means "no offset recorded",
+/// which the caller folds as `max(absent, incoming)` and overwrites; doing that to a
+/// failed-checksum file discards a cursor that may have been far ahead.
 async fn read_persisted_offset(path: &str) -> Result<Option<u64>, IggyError> {
     if !Path::new(path).exists() {
```
hubcio, line 259: blocking `Path::exists()` stat on the pump - and `persist_offset` above documents removing exactly this pattern. it's also redundant: map `open`'s `NotFound` to `Ok(None)` instead (keep that arm, or every cold key becomes a hard error).

hubcio (changes requested):
one finding with no diff line to attach to: legacy `persist_offset` (core/server/src/streaming/partitions/storage.rs:146) opens without `.truncate(true)` and writes 8 bytes, so a downgrade over a 16-byte checksummed record leaves the stale checksum in bytes 8..16, and the next upgrade reads it as `Corrupt` - which lands in the offset-panic chain flagged on offset_storage.rs. the "legacy server stays compatible both ways" claim in offset_recovery.rs holds for reads, not writes. one-line fix on the legacy side, or scope the claim.

core/consensus/src/impls.rs
krishvishal: Not fixing this in the review pass, deliberately. The comment is now a TODO spelling out the merge consequence you named, plus both closures with their hazards:

1. Recomputing after `stamp_prepare_for_persistence` changes the identity at stamp time and folds in `base_offset`, desyncing the pipeline entry from the journaled one that `handle_prepare_ok` compares.
2. Hashing the stamp-invariant cover (everything past the command header) is safe here, but costs a produce-path pass and retires the sentinel that lets existing WALs replay.

Not sure which way to take.

core/partitions/src/poll_plan.rs
krishvishal: On the `Faulted` sub-point: not adding a client error in this pass.

The poll reply has no error channel at all, `resolve_poll_request` failures already flatten to an empty body, so adding one is an SDK-visible change on every client. Recorded at the fault site instead, including why an empty poll is a fair answer for a transient IO fault and a wrong one for at-rest corruption that never recovers.

numinnex (dismissed):

hubcio (approved):

numinnex (approved):

PR #3825, opened 2026-08-05
refactor(python): readme improvements

### LINKED ISSUE #3750: Python SDK: Developer Experience and Development Environment Improvements
### Description

The current documentation and testing configurations in Python SDK have a number of problems:

- README file doesn't offer a reliable and easy to follow guide for development workflow. We are presented with a list of commands that get some of the jobs done, some of them with side effects. More on that in next items.
- Using `uv run` without `--no-sync` is performing implicit syncing that overwrites the installation made by `maturin develop`, reverting the newest changes that were just introduced by installing the package via `maturin develop`. 
- Command to generate stubs is not documented.
- Formatting commands required for the hooks to let your commits and pushes through specifically for Python code are not documented. 
- Docker compose and the Dockerfile seem to not have been updated in a while. Iggy server build is failing without seccomp:unconfined option, the overall compose command is failing due to problems with connectivity between containers, how this connectivity is established or checked (I'm still in the process of figuring all the root issues out here).
- Dockerfile can be improved by adding caching and other changes.
- Any other fixes for the development features that Python SDK is providing but which actually don't work out of the box.

In general, this set of changes is intended to make sure that a developer can follow the instructions which cover all necessary steps and all of which work as intended and serve a clear purpose.

### Affected area / component

Python SDK

### Proposed solution

- [x] Provide a development workflow in README, covering all steps of developing every feature and providing commands that are most reliable and easiest to use. Important action items as a part of this:
  - [x] Add `--no-sync` to some `uv run` commands to get rid of undesired side effects.
  - [x] Document the command to generate stubs
  - [x] Provide `ruff` formatting commands that are required by hooks.
- [ ] ~~Update docker-related files to make sure docker-compose runs tests out of the box.~~
  - [ ] ~~Improve test Dockerfile by switching to Rust-based image, installing uv and a specific python version. This will pin specific versions as arguments and will allow to perform checks for different python or rust versions if needed. Add caching for uv and `--locked` option for the `uv sync` so that specifically the container build could trigger a developer to intentionally update the lock file. (The latter is subject to change in the process of developing the workflow).~~
  - [ ] ~~Fix issues with entrypoint script of the image and / or the configurations inside of the docker-compose file (there are interconnected and present as a group of issues).~~
  - [ ] ~~Resolve the problem that `test_tls.py` introduces -- to be updated~~
- [ ] ~~Any other fixes for the development features that Python SDK is providing but which actually don't work out of the box and / or for the documentation.~~

I have already implemented almost all of these locally and the list may grow further as I discover the algorithm that gets a developer through the whole development process without issues.

UPD: The docker part will be eventually moved into its own issue and is still under discussion.
### Alternatives considered

_No response_

### Contribution

### Good first issue

mmmmxa:

Closes #3750 

## Rationale

The documentation provided in Python SDK does not provide a golden path for development workflow. The absence of golden path adds the cognitive load on most devs, especially the new ones. 

## What changed?

README for Python SDK now provides a clear instruction on how to build and test it and gives basic advice on troubleshooting for precommit and prepush hooks. The new path is tested by hand and has fixed a number of unwanted side effects described in the linked issue.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? - Claude
2. Scope of usage? - autocomplete, editing
3. How did you verify the generated code works correctly? - ran the proposed workflows from the readme manually on my local computer
4. Can you explain every line of the code if asked? - yes


mmmmxa:
@ethanlin01x @jiengup Hi guys, it is optional but would be helpful if you test these instructions on your machine and see whether it works. @ethanlin01x you mentioned you're on MacOS and you're willing to test it, your input on what could be added here will be valuable. Thanks.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   > Activate the venv:
+   >
+   > ```bash
+   > source .venv/bin/activate
+   > ```
+   >
+   > Install the dependencies with `pip`:
+   >
+   > ```bash
+   > pip install -e ".[all]"
```
hubcio: pip path lost its rebuild step - master had `maturin develop` after `pip install -e ".[all]"`, this rewrite dropped it. there's no pure-python source here; the module is the compiled cdylib, and the editable install copies the built `.so` into site-packages (`direct_url.json` says editable, but there's no `.pth` redirect and no import hook - the artifact is a snapshot). so after any rust edit, `pytest` silently tests the old binary. a new symbol at least fails loud with `AttributeError`, but changed behavior on an existing method just runs stale code, and the fix-a-bug-then-retest loop is actively misleading - the test keeps failing and you start "fixing" code that was already correct. re-add an explicit rebuild command to the pip path.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   > ```
+
+2. Run the server to be able to run the tests
+
+   ```bash
+   cargo run --bin iggy-server -- --with-default-root-credentials --fresh
```
hubcio, line 84: step 2 only works from repo root - `foreign/python` is excluded from the root workspace, so from the sdk dir cargo errors with `no bin target named 'iggy-server'`. meanwhile steps 1/3/4 only work from `foreign/python`. no step says where to run it from, so the numbered sequence can't be followed from any single directory - a per-command cwd note would fix the whole thing. two more small things here: the server runs foreground (steps 3-5 need a second terminal), and `--fresh` wipes `local_data/` on every start (the flag's own help says "THIS WILL DELETE ALL DATA!") - worth a word of warning now that it's the only documented way to start the server.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   > Sync the environment without updating it:
+   >
+   > ```bash
+   > uv sync --frozen --all-extras --no-install-project
+   > ```
+   >
+   > Build the project - this runs cargo build and performs an editable install:
+   >
+   > ```bash
+   > uv run maturin develop
```
hubcio: missing `--no-sync`. bare `uv run` implicitly syncs, which builds and installs the project via the pep517 backend (release profile) right before `maturin develop` builds it again in debug - two full cargo builds per iteration, one thrown away. it also undoes the `--no-install-project` added on line 55, so that flag currently buys nothing. line 95 already does this correctly, and CI uses `uv run --no-sync maturin develop` in `.github/actions/python-maturin/pre-merge/action.yml`.

foreign/python/README.md
```diff
@@ -24,6 +24,7 @@ Apache Iggy is the persistent message streaming platform written in Rust, suppor
 
 ```bash
 # Using uv
+uv venv # if not already created
```
hubcio: `uv venv` isn't the missing prerequisite here - in a bare directory `uv add apache-iggy` fails with `No pyproject.toml found in current directory or any parent directory`, venv or not (`uv add` needs a project). since this readme is also the pypi landing page, better to show `uv pip install apache-iggy` for the standalone flow and drop this line. `uv add` is fine when the reader already has a project.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+4. To update the stubs, use
+
+   ```bash
+   cargo run --bin stub_gen
+   ```
+
+5. Before committing, test the pre-commit and pre-push hooks. `prek` only inspects staged content, so stage your work first:
+
+   ```bash
+   git add -A
```
hubcio, line 112: this sequence wedges when any fixer hook modifies a file: fixes land in the worktree, the index keeps the unfixed blob, `prek run` exits 1, and the pre-push run just stashes the fix and re-checks the stale index (`Hook changes conflicted with the saved unstaged changes. Reverting the hook changes`) - zero progress until you re-stage. add a note to re-run `git add -A` after any run where hooks changed files. (`prek run --all-files` is not a substitute - it reads the worktree and passes while the staged blob is still broken, and it feeds the whole repo to the `typos -w` fixer.)

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   prek run # runs pre-commit hooks
+   prek run --hook-stage pre-push
+   ```
+
+   These are some of the essential commands prek is running, so it's recommended to run them manually before  running prek / committing / pushing. This list is not exhaustive and other hook failures are possible.
+
+   ```bash
+   ruff format .
```
hubcio: also line 125. bare `ruff` uses whatever is on PATH, and on the uv path the venv is never activated so it may not exist at all. the versions in play differ (PATH vs venv vs lock vs the prek pin), and format output changes across minors, so this can fight the hook. `uv run --no-sync ruff format .` / `uv run --no-sync ruff check --fix .` matches CI and the synced venv.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   ```bash
+   ruff check --fix .
+   ```
+
+   ```bash
+   cargo fmt --all
+   ```
+
+   ```bash
+   cargo clippy --all-targets --all-features -- -D warnings
```
hubcio: also line 129. neither command covers this crate from repo root - `foreign/python` is excluded from the workspace, so `cargo fmt --all` is a silent no-op for the sdk and this clippy line builds the whole root workspace while checking none of the sdk's rust. the prek hooks have the same blind spot (they run against the root manifest), so while the text above is accurate about what prek runs, the pair still never lints this crate. CI does it right: `cd foreign/python` then `cargo fmt --manifest-path Cargo.toml` / `cargo clippy --manifest-path Cargo.toml --all-targets --all-features -- -D warnings`. worth using those forms here and noting they run from `foreign/python`.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   ```
+
+   ```bash
+   ./scripts/ci/markdownlint.sh --fix # read the diff after applying this, sometimes it gives unwanted results, e.g. messing up enumerations
```
hubcio: without a path argument the script falls back to `**/*.md`, so `--fix` rewrites every markdown file in the repo (74 tracked files) - much wider than the prek hook, which only gets staged filenames. given this line's own warning about the fixer messing up enumerations, scope it: `./scripts/ci/markdownlint.sh --fix foreign/python/README.md`. it also only exists relative to repo root.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   > uv venv
+   > ```
+   >
+   > Sync the environment without updating it:
+   >
+   > ```bash
+   > uv sync --frozen --all-extras --no-install-project
+   > ```
+   >
+   > Build the project - this runs cargo build and performs an editable install:
```
hubcio: "editable install" oversells it - there's no python source to edit; `maturin develop` copies a compiled module into the venv, so every rust change needs a re-run. also worth noting the two paths land on different build profiles: `maturin develop` builds debug while the pip path's pep517 build is release, and with the 30s per-test timeout the debug extension is the more timeout-prone one. suggestion: "builds the rust extension into the venv (debug profile) - re-run after any rust change".

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
-uv run pytest tests/ -v # Run tests (requires iggy-server running)
-
-# Using pip:
-python3 -m venv .venv
-source .venv/bin/activate
-pip install -e ".[all]"
-maturin develop
-pytest tests/ -v # Run tests (requires iggy-server running)
-```
+1. Build a project for development
```
hubcio, line 49: prerequisites are understated - the only stated requirement in this file is python 3.10+, but every path needs a rust toolchain from step 1 onward (this step literally runs cargo build, and the pip variant compiles the extension too), plus uv and prek. and since the docker option was dropped, there's no toolchain-free way to get a server anymore. a short prerequisites line plus a pointer to the prek install section of the root CONTRIBUTING.md would help.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   `pip`:
+
+   ```bash
+   pytest tests/ -v # make sure iggy-server is running and the venv is activated
+   ```
+
+4. To update the stubs, use
+
+   ```bash
+   cargo run --bin stub_gen
```
hubcio, line 105: two things: the command only resolves from `foreign/python` (the crate is not in the root workspace), and it's worth saying when to run it - only after changing the pyo3 surface; nothing in CI checks stub freshness, so unconditional regen just invites `.pyi` churn. separate but related: `stub_gen.rs` builds its license-prepend path from `file!()`, which resolves against the runtime cwd - run this from any subdirectory of `foreign/python` and it rewrites the tracked `apache_iggy.pyi` without the apache header before failing with a bare ENOENT. the CI license gate catches it, but `Path::new(env!("CARGO_MANIFEST_DIR"))` in that binary would make step 4 cwd-independent. happy to see that land separately since the code is outside this diff.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
-source .venv/bin/activate
-pip install -e ".[all]"
-maturin develop
-pytest tests/ -v # Run tests (requires iggy-server running)
-```
+1. Build a project for development
+
+   With `uv`:
+
+   > Create a venv:
```
hubcio: style: blockquote-wrapped code fences are new to this repo - every other readme (and steps 2-4 below) uses plain fences, and `>` elsewhere means the ASF disclaimer. collapsing each install path to one fence with `#` comments reads the same and drops ~40 lines.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   ```
 
 ## Examples
 
 Refer to the [examples/python/](https://github.com/apache/iggy/tree/master/examples/python) directory for usage examples.
 
 ## Contributing
 
-See [CONTRIBUTING.md](https://github.com/apache/iggy/blob/master/foreign/python/CONTRIBUTING.md) for development setup and guidelines.
+See [CONTRIBUTING.md](https://github.com/apache/iggy/blob/master/CONTRIBUTING.md) for development setup and guidelines.
```
hubcio: the retarget itself is right (`foreign/python/CONTRIBUTING.md` doesn't exist, so the old link was a 404), but the root CONTRIBUTING.md delegates language-specific setup back to this very readme, so "for development setup" now sends the reader in a circle. "for contribution guidelines" fits better.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+
+5. Before committing, test the pre-commit and pre-push hooks. `prek` only inspects staged content, so stage your work first:
+
+   ```bash
+   git add -A
+   prek run # runs pre-commit hooks
+   prek run --hook-stage pre-push
+   ```
+
+   These are some of the essential commands prek is running, so it's recommended to run them manually before  running prek / committing / pushing. This list is not exhaustive and other hook failures are possible.
```
hubcio: double space in "before  running". this line is also 212 chars - worth wrapping.

hubcio (changes requested):

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+   ```bash
+   cargo run --bin iggy-server -- --with-default-root-credentials --fresh
+   ```
+
+3. Run the tests
+
+   `uv`:
+
+   ```bash
+   uv run --no-sync pytest tests/ -v
```
ethanlin01x, line 92: step 3 does not mention docker. not asking for the compose file back - #3750 defers that - but `tests/test_tls.py` needs a daemon regardless: it starts `apache/iggy:edge` through testcontainers with no skip guard, so all 6 tests there error at fixture setup without one. `--all-extras` on line 55 pulls `testcontainers` in, so this is the default path.

worth listing docker under prerequisites, or giving the no-docker variant `uv run --no-sync pytest tests/ -v --ignore=tests/test_tls.py`. same for line 101.

foreign/python/README.md
```diff
@@ -38,33 +39,111 @@ pip install apache-iggy
...
+
+   ```bash
+   cargo run --bin stub_gen
+   ```
+
+5. Before committing, test the pre-commit and pre-push hooks. `prek` only inspects staged content, so stage your work first:
+
+   ```bash
+   git add -A
+   prek run # runs pre-commit hooks
```
ethanlin01x, line 113: same `--no-sync` issue as line 61, but in the hook and with the opposite outcome. `.pre-commit-config.yaml:45` runs `uv run pyrefly check` with no `--no-sync`, on any staged `foreign/python/**.{py,pyi}` - which step 4's regenerated `.pyi` plus the `git add -A` above makes routine. on line 61 the sync lands before `maturin develop`, so the release build is the one thrown away and you still end up with debug. here it lands after step 1, so the release build wins and installs over the debug `.so` you were testing against.

the real fix is in the hook, which is outside this diff and wants its own pr. for now a note under step 5 would do: "`prek run` reinstalls the package as a side effect, re-run `uv run --no-sync maturin develop` before going back to step 3" - removable once the hook is fixed. happy to open an issue for it.

foreign/python/README.md
mmmmxa: `uv pip install apache-iggy` fails the same in a directory without a project or a venv. Since `uv` is practically always used in a managed project, I've just left the add command and changed the comment to specify that the command requires a project.

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: Specified that default path to run the commands from is `foreign/python` unless specified otherwise. This applies to the comments of the same kind as well. Added the note about wiping.

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: Added a TODO note with a link to this comment.

foreign/python/README.md
mmmmxa: The intention to keep these separate was to make it easier for the readers of the rendered file to copy the commands by clicking the button to copy the code in the code block. I find copying and pasting from a bigger block much less convenient. About the blockquotes -- that was a problem with linted which messed up enumeration in the file because it considered enumeration to be separate and put all 1's instead of enumeration I wanted. I have applied the changes you requested in order to get this PR through quicker, but if you agree with my idea about ease of copying, I'd return it back. If you don't have time for this -- let's just move on and leave it at that.

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: Done, now Docker is listed under prerequisites.

foreign/python/README.md
mmmmxa: This sounds worth looking into -- please open an issue for this yourself -- I have already spent way more time on this README and dev tools that I'd like :smiling_face_with_tear:

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: done

foreign/python/README.md
mmmmxa: done

hubcio (approved):

numinnex (approved):

ethanlin01x (commented):
Just finished running through every command in the new README on macOS and they all work. A bit late since this is already merged, but leaving a note anyway.

PR #3828, opened 2026-08-06
fix(server-ng): close server-ng SDK-visible behavior gaps

hubcio:
The Python SDK suite pinned twelve server-ng divergences from the
legacy server as strict xfails; this closes the eleven that were
server bugs (get_topics on a missing stream erroring is intended).

Timestamp polls now filter on the broker append time replies
report; a missing partition id denies with PartitionNotFound
instead of an empty success; an unmaterialized partition reports
its initial one-empty-segment state. Pre-consensus validation
shared by TCP and HTTP denies TooManyPartitions (create_topic,
create_partitions, delete_partitions) and InvalidTopicSize before
a rejected request burns a replicated log entry. Updates store
ServerDefault sentinels verbatim and reads echo the stored size
and expiry instead of freezing the node default into state.

Purge never reached partition data. Purge commits now wake the
reconciler; the applied generation is durable per partition
(purge.gen, always fsynced last, and an I/O error reading it back
fails boot instead of silently re-purging); a purge floor fences
every journal-apply path, surviving the poll-index rebuild of
eviction re-appends; repair serves only above the floor and defers
while a committed purge is not yet locally applied. Repair-vs-purge
ordering across a StartView stays open, documented as TODO.

### DIFF
--- a/core/metadata/src/impls/metadata.rs
+++ b/core/metadata/src/impls/metadata.rs
@@ -41,7 +41,6 @@ use iggy_binary_protocol::requests::partitions::CreatePartitionsRequest as WireC
 use iggy_binary_protocol::requests::partitions::CreatePartitionsWithAssignmentsRequest as PersistedCreatePartitionsRequest;
 use iggy_binary_protocol::requests::topics::CreateTopicRequest as WireCreateTopicRequest;
 use iggy_binary_protocol::requests::topics::CreateTopicWithAssignmentsRequest as PersistedCreateTopicRequest;
-use iggy_binary_protocol::requests::topics::UpdateTopicRequest as WireUpdateTopicRequest;
 use iggy_binary_protocol::{
     Command2, ConsensusHeader, EvictionReason, GenericHeader, Operation, PrepareHeader,
     PrepareOkHeader, ReplyHeader, RequestHeader, WireDecode, WireEncode, WireName,
@@ -892,23 +891,11 @@ impl<C, J, S, M, SB> IggyMetadata<C, J, S, M, SB> {
         self.client_table.borrow_mut().set_capacity(max_clients);
     }
 
-    /// Resolved byte value for `MaxTopicSize::ServerDefault`.
-    #[must_use]
-    pub const fn default_max_topic_size(&self) -> u64 {
-        self.default_max_topic_size.get()
-    }
-
     /// Install the resolved micros value used for `IggyExpiry::ServerDefault`.
-    /// Server-ng bootstrap calls this with `system.topic.message_expiry` on every
-    /// shard (responses read it too); only shard 0's copy feeds admission.
+    /// Server-ng bootstrap calls this with `system.topic.message_expiry`; only
+    /// shard 0's copy feeds admission.
     pub fn set_default_message_expiry(&self, message_expiry_micros: u64) {
         self.default_message_expiry.set(message_expiry_micros);
-    }
-
-    /// Resolved micros value for `IggyExpiry::ServerDefault`.
-    #[must_use]
-    pub const fn default_message_expiry(&self) -> u64 {
-        self.default_message_expiry.get()
     }
 
     /// Fire post-commit notifier. Clones the `Rc` out under a short
@@ -3321,30 +3308,10 @@ where
                     &body,
                 ))
             }
-            Operation::UpdateTopic => {
-                let mut request = WireUpdateTopicRequest::decode_from(body)
-                    .map_err(|_| IggyError::InvalidCommand)?;
-                // Same `ServerDefault` resolution as `CreateTopic` above; rebuild
-                // the prepare only if a sentinel actually needs stamping, else
-                // project the untouched buffer zero-copy.
-                let needs_rewrite = request.max_topic_size == 0 || request.message_expiry == 0;
-                if request.max_topic_size == 0 {
-                    request.max_topic_size = self.default_max_topic_size.get();
-                }
-                if request.message_expiry == 0 {
-                    request.message_expiry = self.default_message_expiry.get();
-                }
-                if needs_rewrite {
-                    let body = request.to_bytes();
-                    return Ok(build_prepare_message(
-                        consensus,
-                        &header,
-                        Operation::UpdateTopic,
-                        &body,
-                    ));
-                }
-                Ok(message.project(consensus))
-            }
+            // `UpdateTopic` deliberately takes the default arm: unlike create,
+            // an update stores `ServerDefault` sentinels verbatim (legacy
+            // parity), so a later get echoes `ServerDefault` instead of the
+            // node default frozen at update time.
             _ => Ok(message.project(consensus)),
         }
     }
@@ -3735,9 +3702,9 @@ where
     // `created_at` on every CreateStream/CreateTopic/CreatePartitions. The
     // in-process callers that bypass `Project::project` build their prepare
     // through this helper directly (the CreateTopic/CreatePartitions
-    // assignment rewrites, the UpdateTopic default-size rewrite, and the
-    // PAT-cleaner delete); the stamp is load-bearing for the creates and inert
-    // for the UpdateTopic rewrite and the delete, whose applies ignore it.
+    // assignment rewrites and the PAT-cleaner delete); the stamp is
+    // load-bearing for the creates and inert for the delete, whose apply
+    // ignores it.
     // Shared `next_monotonic_timestamp` keeps the in-process path on the same
     // monotonic-clock guard as the wire path.
     let timestamp = consensus.next_monotonic_timestamp();
@@ -3762,9 +3729,9 @@ where
         // Carry the acting user id so the in-apply RBAC gate sees the same
         // identity on every replica. The default projection copies it (see
         // `Project::project`); this helper builds prepares for the ops it
-        // rewrites (the CreateTopic/CreatePartitions assignment rewrites, the
-        // UpdateTopic default-size rewrite, and the PAT-cleaner delete), which
-        // would otherwise reset it to 0 via `..Default::default()`.
+        // rewrites (the CreateTopic/CreatePartitions assignment rewrites and
+        // the PAT-cleaner delete), which would otherwise reset it to 0 via
+        // `..Default::default()`.
         user_id: request.user_id,
         // Seal the body integrity field over the rewritten body, exactly as
         // `Project::project` does for wire-projected prepares. This helper builds
--- a/core/metadata/src/impls/metadata.rs
+++ b/core/metadata/src/impls/metadata.rs
@@ -873,6 +873,16 @@ impl<C, J, S, M, SB> IggyMetadata<C, J, S, M, SB> {
         self.default_max_topic_size.set(max_topic_size_bytes);
     }
 
+    /// Byte value a stored `MaxTopicSize::ServerDefault` resolves to on this
+    /// node. Read by the per-shard segment cleaner, which enforces retention
+    /// locally and so must resolve the sentinel at enforcement time: create
+    /// admission rewrites it before replication, but an UPDATE back to
+    /// `ServerDefault` leaves the sentinel in committed state.
+    #[must_use]
+    pub const fn default_max_topic_size(&self) -> u64 {
+        self.default_max_topic_size.get()
+    }
+
     /// Raise the forced-checkpoint margin to cover a configured
     /// prepare-queue depth (`[metadata] prepare_queue_depth`). Clamped to
     /// the built-in floor by the coordinator; no-op on shards without a
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -3927,9 +3927,21 @@ where
             .map_err(PurgeError::Unserviceable)?;
         // Make the unlinks AND the replanted dirent durable together: without
         // this a crash can resurrect pre-purge segments until the boot re-purge
-        // fires. Bounded and self-healing, but the fsync is one call.
-        if let Some(partition_dir) = self.partition_dir.clone() {
-            let _ = crate::state_transfer::fsync_dir(&partition_dir).await;
+        // fires. Bounded and self-healing, so a failure is logged, not fenced:
+        // the generation write below has not run yet, so a crash after a failed
+        // fsync re-purges at boot anyway.
+        if let Some(partition_dir) = self.partition_dir.clone()
+            && let Err(error) = crate::state_transfer::fsync_dir(&partition_dir).await
+        {
+            warn!(
+                target: "iggy.partitions.diag",
+                plane = "partitions",
+                namespace_raw = namespace.inner(),
+                generation,
+                %error,
+                "purge could not fsync the partition dir; a crash before the \
+                 generation record re-purges at boot"
+            );
         }
         // The boot-time durable line marks recovered bytes that must not be
         // re-persisted, but the purge just deleted those bytes and offsets
@@ -3977,14 +3989,28 @@ where
         // crash right after the purge otherwise resurrects the offset files at
         // boot, and while recovery clamps a resurrected offset down to the
         // rebuilt head, "consumed through 0" is not the intended "no entry at
-        // all" -- that consumer skips the first post-purge message.
+        // all" -- that consumer skips the first post-purge message. Logged on
+        // failure, sharper than the partition-dir fsync above: the generation
+        // write below still runs, so a crash would resurrect these files with
+        // no boot re-purge left to clear them.
         for dir in self
             .consumer_offsets_path
             .clone()
             .into_iter()
             .chain(self.consumer_group_offsets_path.clone())
         {
-            let _ = crate::state_transfer::fsync_dir(&dir).await;
+            if let Err(error) = crate::state_transfer::fsync_dir(&dir).await {
+                warn!(
+                    target: "iggy.partitions.diag",
+                    plane = "partitions",
+                    namespace_raw = namespace.inner(),
+                    generation,
+                    dir = %dir,
+                    %error,
+                    "purge could not fsync an offsets dir; a crash may resurrect \
+                     deleted offset files with the purge already recorded"
+                );
+            }
         }
         // The persisted-offset tracker mirrors the files unlinked above; a
         // stale entry would make a post-purge auto-commit skip its write and
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -20,7 +20,10 @@ use crate::journal::{MessageLookup, PartitionJournal, PartitionJournalMemStorage
 use crate::log::JournalInfo;
 use crate::log::SegmentedLog;
 use crate::messages_writer::MessagesWriter;
-use crate::offset_storage::{delete_persisted_offset, persist_offset, persist_offset_max};
+use crate::offset_storage::{
+    PURGE_GENERATION_FILE, delete_persisted_offset, persist_offset, persist_offset_max,
+    persist_purge_generation, read_purge_generation,
+};
 use crate::poll_plan::{
     AutoCommitCtx, AutoCommitTarget, DiskReadPlan, DiskSegment, LastPolledCtx,
     PartitionDirResolution, PollPlan, PollTier, ResidentTailSnapshot,
@@ -158,6 +161,15 @@ where
     /// generation against this and resets only when it advances, so a redundant
     /// reconcile pass never re-wipes a partition already at this generation.
     pub(crate) applied_purge_generation: u64,
+    /// Highest consensus op assigned when the last purge ran. INVARIANT: every
+    /// journal-apply path must no-op entries with `op <= purge_floor_op`. The
+    /// purge keeps journal entries resident (consensus history for backups,
+    /// repair and retransmission) while wiping the segments, so without the
+    /// floor a pre-purge op committing after the purge would flush purged
+    /// bytes back into a fresh segment or re-advance the reset offset. Not
+    /// persisted: the in-memory journal dies with the process, so no resident
+    /// pre-purge entry survives a restart.
+    purge_floor_op: u64,
     /// Durable superblock for this partition's consensus group, recording
     /// `(view, log_view)` across a crash so this replica can never
     /// re-participate in a view older than one it advertised. `None` for
@@ -316,6 +328,17 @@ pub enum PurgeError {
     /// `ENOSPC` / `EIO` is logged by the superblock writer on the first failure
     /// and at every power-of-two thereafter.
     FrontierNotRecorded,
+    /// The wipe ran and the fresh chain is planted, but the applied purge
+    /// generation could not be recorded durably (`purge.gen`), so
+    /// `applied_purge_generation` stays at its pre-purge value and the
+    /// reconciler re-issues the purge. Retry, do not fence: the partition is
+    /// serviceable and re-purging an already-empty chain is cheap.
+    ///
+    /// Sets [`Self::purge_deferred`] for the same reason as
+    /// [`Self::FrontierNotRecorded`]: an op acked between this failure and the
+    /// retry would be wiped by that retry while every peer that recorded the
+    /// generation keeps it.
+    GenerationNotRecorded(IggyError),
     /// A step after the drain failed, so the partition holds no serviceable
     /// segment chain and its next append would panic on `active_segment()`.
     /// The caller must fence this group for rebuild.
@@ -328,6 +351,11 @@ impl fmt::Display for PurgeError {
             Self::FrontierNotRecorded => write!(
                 f,
                 "could not record the purge's offset-frontier reset; nothing was mutated"
+            ),
+            Self::GenerationNotRecorded(source) => write!(
+                f,
+                "purge reset the partition but could not record its applied generation; \
+                 the purge will be re-issued: {source}"
             ),
             Self::Unserviceable(source) => write!(
                 f,
@@ -436,6 +464,7 @@ where
             persisted_offsets: RefCell::new(HashMap::new()),
             observed_view,
             applied_purge_generation: 0,
+            purge_floor_op: 0,
             superblock: None,
             superblock_lock: LocalGate::new(),
             superblock_write_failures: Cell::new(0),
@@ -460,6 +489,34 @@ where
     #[must_use]
     pub const fn applied_purge_generation(&self) -> u64 {
         self.applied_purge_generation
+    }
+
+    /// See [`Self::purge_floor_op` field docs](#structfield.purge_floor_op).
+    /// Exposed for the repair-serving path: a peer must not serve entries at
+    /// or below this replica's floor.
+    #[must_use]
+    pub const fn purge_floor_op(&self) -> u64 {
+        self.purge_floor_op
+    }
+
+    /// Seed [`Self::applied_purge_generation`] from the partition dir's
+    /// `purge.gen` file at build time (both fresh create and recovery walk
+    /// this). Absent file reads 0, so a partition that never purged and a
+    /// repair-rebuilt dir both start below any committed generation and the
+    /// reconciler re-applies the purge; a crash AFTER a purge's durable
+    /// generation write correctly skips the re-wipe, keeping messages
+    /// appended since. No-op without a partition dir (in-memory storage).
+    ///
+    /// # Errors
+    /// Propagates a real I/O failure reading `purge.gen`: booting with the
+    /// sentinel 0 instead would make the reconciler silently re-purge and
+    /// destroy post-purge messages, so the boot fails loud.
+    pub async fn hydrate_applied_purge_generation(&mut self) -> Result<(), IggyError> {
+        if let Some(dir) = self.partition_dir() {
+            let path = format!("{dir}/{PURGE_GENERATION_FILE}");
+            self.applied_purge_generation = read_purge_generation(&path).await?;
+        }
+        Ok(())
     }
 
     #[must_use]
@@ -2595,6 +2652,13 @@ where
                         }
                         continue;
                     }
+                    // Purge floor: a pre-purge batch committing after the
+                    // purge must not flush its (purged) bytes into the fresh
+                    // segment. It still counts into `chunk_len`, so it joins
+                    // the evictable prefix and commit_min advances normally.
+                    if peek_op(&entry) <= self.purge_floor_op {
+                        continue;
+                    }
                     // Resident committed SendMessages entry: this node stamped it
                     // in `append_messages` (recomputing the batch checksum over these
                     // exact bytes), so a validating re-decode would only re-hash ~1
@@ -2752,7 +2816,13 @@ where
         }
         let retained = self.log.journal().inner.evict_prefix(count).await;
         let mut retained_info = JournalInfo::default();
-        for (_, meta) in &retained {
+        for (entry, meta) in &retained {
+            // Purge floor: a retained pre-purge batch must not fold its
+            // accounting back into `journal.info`, or the info would re-adopt
+            // a pre-purge `current_offset` the purge just reset.
+            if peek_op(entry) <= self.purge_floor_op {
+                continue;
+            }
             if let Some(meta) = meta {
                 accumulate_committed_info(
                     &mut retained_info,
@@ -2911,6 +2981,14 @@ where
             .iter()
             .map(|entry| {
                 if entry.header.operation != Operation::SendMessages {
+                    return None;
+                }
+                // Purge floor: a pre-purge send committing after the purge
+                // reports no visible offsets ("send without confirmation", the
+                // established degradation), which also keeps
+                // `commit_partition_entry` from re-advancing the reset offset
+                // and stats with pre-purge values.
+                if entry.header.op <= self.purge_floor_op {
                     return None;
                 }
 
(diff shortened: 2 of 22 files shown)


core/partitions/src/iggy_partition.rs
```diff
@@ -3153,6 +3227,33 @@ where
         self.stats.zero_out_all();
         self.stats.increment_segments_count(1);
 
+        // Fence the resident journal instead of clearing it: entries are
+        // consensus history (backup commit walks, repair, retransmission), so
+        // they stay, but every journal-apply path no-ops ops at or below this
+        // floor (see `purge_floor_op`). The write lock held here is the same
+        // one appends take, and the pump is single-threaded, so no op can be
+        // assigned between reading the sequence and installing the floor.
+        self.purge_floor_op = self.consensus.sequencer().current_sequence();
```
numinnex, line 4036: **Critical.** The fence is derived from the *local* sequencer at *local apply* time, which is strictly above the purge's own position in the partition op order. The purge is asynchronous: `PurgeTopicRequest::apply` bumps the generation and replies Ok (`core/metadata/src/stm/stream.rs:1785`), the reconciler stages on a later pass (`partition_reconciler.rs:1096`), and the pump applies here (`shard/src/router.rs:664`). Every op the partition sequencer assigns in that window lands `<= purge_floor_op`.

So a send the client issued *after* `purge_topic()` returned Ok is either already flushed and unlinked by the loop at `:3151-3166`, or resident and fenced at `:2032`, or fenced at `:2364` and answered as success. An acknowledged write is destroyed with no crash, on a single node. Legacy has no equivalent window: it purges inline before replying (`core/server/src/shard/execution.rs:348-354`).

Second symptom from the same line: because each replica reads its own sequencer at its own reconciler timing, replicas fence *different* op sets. Ops in `(min_floor, max_floor]` flush on some and not others, so `offset`/`dirty_offset` diverge and the next live batch is stamped at a different `base_offset` per replica — the invariant documented at `:2463-2474` says that cannot happen. The `TODO` at `partition_reconciler.rs:1075` names the skew but not the acked-write loss.

A replicated fence value fixes the divergence but not the wipe, since the unlink loop still runs at reconciler timing. Suggest routing the purge through the *partition* consensus log as a `PurgePartition` prepare, so its own op is the fence: ops below it deleted, ops above kept, identical on every replica, and ordered against repair by the same walk. Note a fence carried in the `PurgeTopic` body cannot work — that is one metadata-group op, while each partition has its own `VsrConsensus`, sequencer and primary (`bootstrap.rs:2156-2185`), so one number is wrong for N-1 partitions.

core/partitions/src/iggy_partition.rs
```diff
@@ -3153,6 +3227,33 @@ where
...
+
+        // Last durable step: record the applied generation before the
+        // in-memory marker advances. On a write failure the marker stays old,
+        // the error propagates, and the reconciler retries the whole purge
+        // (idempotent). The reverse order would ack a purge that a crash
+        // then silently undoes: restart would hydrate the old generation,
+        // yet the reconciler believes the purge applied.
+        if let Some(dir) = self.partition_dir() {
+            let path = format!("{dir}/{PURGE_GENERATION_FILE}");
+            persist_purge_generation(&path, generation).await?;
```
numinnex: **Critical.** The comment above claims the retry is idempotent. It is not idempotent with respect to messages: on `Err` the segments are already wiped, the floor is installed, `write_lock` is released and `applied_purge_generation` is unchanged, so the partition goes on accepting and acking sends at offset 0 until the next reconciler pass re-purges and deletes them.

Worse, the loss ratchets. `reconcile_partition_purges` has no backoff gate (compare `reconcile_partition_teardowns` at `partition_reconciler.rs:800-803`) and `request_purge_partition` is a bare `try_send` (`shard/src/lib.rs:1537-1541`), so every ~1s pass re-runs the whole purge and re-reads `current_sequence()` at `:3236`. The fenced set grows one step per pass for as long as `persist_purge_generation` keeps failing, destroying whatever was acked since the previous attempt.

Suggest refusing appends until the generation is durable, and routing retries through the existing `failure_state` backoff with an attempt cap and a `warn!`.

Separately on durability ordering: `persist_purge_generation` only `sync_data()`s the file, and `purge()` fsyncs no directory. The dirent for a first `purge.gen` is therefore not durable, and the segment unlinks are not ordered before the applied mark. A power loss can leave `purge.gen = N` durable with the deletions lost, in which case boot hydrates `applied == committed`, skips the re-purge, and `load_persisted_segments` recovers the purged segments permanently. Write-then-mark needs the deletion durable first — suggest fsyncing the partition dir after the unlink loop and after `install_empty_segment`, before writing the generation.

core/shard/src/lib.rs
```diff
@@ -3311,8 +3312,47 @@ where
...
-        let retained_from = partition.log.journal().inner.repair_retained_from();
+        // Purge fence: never serve entries at or below this replica's purge
+        // floor. The journal keeps them (own commit walk), but a rejoiner's
+        // floor died with its process, so served pre-purge batches would
+        // flush right back into its freshly reset segments. Reporting the
+        // floor as the retention start rides the normal `RangeEvicted` path:
+        // the rejoiner moves its commit floor to the purge point instead.
+        let purge_floor = partition.purge_floor_op();
+        let retained_from = match partition.log.journal().inner.repair_retained_from() {
+            Some(op) => Some(op.max(purge_floor + 1)),
```
numinnex: **Critical — permanent rejoin wedge.** When a purge is the newest partition op, `retained_from` becomes `purge_floor + 1 > to_op` and the served window is empty.

Full chain on a quiescent purged partition, where the serving peer's `purge_floor == sequencer == commit_max == to_op`: `handle_start_view` arms repair with `from_op=1, to_op=commit_max` (`:3040-3055`, `to_op` frozen there) → this line raises `retained_from` → `RangeEvicted(purge_floor+1)` sets `repair.floor = Some(purge_floor)` (`:3613`, `saturating_sub(1)`) → the serve loop runs zero iterations so `RepairDone` follows → `complete_repair` clamps `floor.min(commit_max) == to_op` and `repaired_window_is_offsets_only` returns false on `floor >= to_op` (`core/partitions/src/iggy_partition.rs:3430`) → refused. `commit_min` never leaves 0.

There is no exit. `complete_repair` returns at `:3390` with `self.repair` still armed; the stall sweep reuses the frozen `session.to_op`, never a refreshed `commit_max` (`:4811-4818`), so the identical window is re-asked every `repair_retry_ticks` forever; the session clears only on `commit_min >= session.to_op` (`:3406-3409`); and the partition `RangeEvicted` arm has no `arm_metadata_transfer` equivalent — that call is metadata-only (`:3571-3582`). Live traffic does not rescue it either, since the rejoiner's backup gap check drops op `to_op+1`. The cost is a group running one replica short indefinitely, so every commit's latency becomes the surviving backup's p99 with no hedging.

The comment at `:3348-3350` says "the rejoiner moves its commit floor to the purge point instead" — the code refuses to.

On the fix: please do not simply exempt the purge floor from the empty-window rule, since that deletes the guard at `iggy_partition.rs:3368-3375` and lets a rejoiner still holding pre-purge segments claim them as committed state and then serve them. Gate acceptance on the rejoiner having applied the purge itself (`applied_purge_generation >= committed`), which sets `recovered_durable_offset = None` and wipes the segments, making the empty window true rather than assumed. Also worth mirroring the metadata branch's explicit `from_op > to_op` reply (`:3234-3259`) — the partition branch reaching `RangeEvicted` + `RepairDone` here is incidental, not intended.

core/partitions/src/iggy_partition.rs
```diff
@@ -2302,6 +2356,14 @@ where
                 if entry.header.operation != Operation::SendMessages {
                     return None;
                 }
+                // Purge floor: a pre-purge send committing after the purge
+                // reports no visible offsets ("send without confirmation", the
+                // established degradation), which also keeps
+                // `commit_partition_entry` from re-advancing the reset offset
+                // and stats with pre-purge values.
+                if entry.header.op <= self.purge_floor_op {
```
numinnex, line 3019: **Critical.** A fenced `SendMessages` returns the empty body at `:3592` (`Bytes::from_static(&[0,0,0,0])`), which `SendMessagesResponse::decode` turns into `confirmations: vec![]` (`core/binary_protocol/src/responses/messages/send_messages.rs:121-136`) — indistinguishable from success at the SDK. The client is told Ok for a write the server discarded.

The root problem is a semantic overload: before this change an empty confirmation list meant "committed, offsets not resolvable" (`:2235-2239`), and the wire doc already flags the ambiguity (`send_messages.rs:94-96`). This adds "silently discarded" to the same shape.

Suggest replying a nonzero reply-header `status = IggyError::TransientNotCommitted` (57) instead. It is compat-safe (status is pre-existing wire, not a body change), all three transports already replay it on the same connection with the same request id (`core/sdk/src/tcp/tcp_client.rs:976-990`, `quic_client.rs:776-790`, `websocket_client.rs:777`), and it converges because the replayed send gets a fresh op above the floor. There is no double-apply risk here: the fenced batch was never stored, so the replay is not a duplicate. If the floor is climbing (see the retry comment on `:3255`), the replay exhausts its budget into a visible timeout rather than a silent success, which is still the better failure.

Note this closes the fenced-at-commit window only. The variant where the send committed and was answered with a real offset-bearing `SendMessagesConfirmationResponse` *before* the purge applied, whose bytes the unlink loop then deletes, cannot be reached by any reply-status change — that one needs the ordering fix on `:3236`.

core/partitions/src/iggy_partition.rs
```diff
@@ -1984,6 +2025,13 @@ where
                         }
                         continue;
                     }
+                    // Purge floor: a pre-purge batch committing after the
+                    // purge must not flush its (purged) bytes into the fresh
+                    // segment. It still counts into `chunk_len`, so it joins
+                    // the evictable prefix and commit_min advances normally.
+                    if peek_op(&entry) <= self.purge_floor_op {
```
numinnex, line 2677: **Critical (receiving side, distinct from the fence itself).** Pre-existing mechanism, newly reachable, and it is the state a rejoiner escapes into once a fresh view change lifts `to_op` above the purge floor.

A replica that has not yet applied the purge boots with `recovered_durable_offset = Some(D)` for a large D (`core/server-ng/src/bootstrap.rs:2242-2249`). A peer that *has* purged serves only post-purge sends, whose `base_offset` restarts at 0, so `complete_repair` takes the `(Some(0), Some(D))` arm, `0 <= D+1` holds, and the floor is accepted (`:3366`). The flush loop just below then drops every one of those batches unwritten at `:2062-2066` (`batch_end <= recovered_durable_offset`), and `commit_partition_entry` skips the offset/stats advance at `:2422-2424`. Because `commit_min` has advanced past them, nothing ever refetches; when that replica finally purges, `recovered_durable_offset` becomes `None` and the new floor covers those ops, fencing them permanently.

Net effect: committed post-purge messages never land on that replica, and until it purges it serves pre-purge data — including as primary. Suggest requiring local purge-apply before accepting a repair floor while a committed purge is outstanding, and naming this mechanism in the `TODO(hubcio)` at `partition_reconciler.rs:1075`, which currently covers only the ordering, not this loss path.

core/partitions/src/iggy_partition.rs
```diff
@@ -143,6 +146,15 @@ where
     /// generation against this and resets only when it advances, so a redundant
     /// reconcile pass never re-wipes a partition already at this generation.
     applied_purge_generation: u64,
+    /// Highest consensus op assigned when the last purge ran. INVARIANT: every
```
numinnex, line 171: The invariant stated here ("every journal-apply path must no-op entries with `op <= purge_floor_op`") does not hold: `append_repaired_send_messages` (`:3451`) is a journal-apply path with no guard. A repaired op at or below the floor still writes `dirty_offset`, `journal.info.current_offset` and `segment.current_position` from its *pre-purge* stamps, so post-purge appends resume at a stale high offset while `segment.start_offset == 0`.

It is currently masked only externally — the `header_by_op` early return at `:3288` plus the peer-side fence at `shard/src/lib.rs:3352` — so it breaks the moment either changes. Suggest `if header.op <= self.purge_floor_op { return Ok(None); }` to match the four sites that did get the guard.

This is also the reachable route into the resident-poll gap (see the comment on `journal.rs:693`): the unguarded append lands out of order, `append_with_meta`'s `op > poll_floor` check keeps it out of the index, but `resident_entries()` still hands it to `select_resident`. Fixing this line alone does not close that one, and vice versa.

core/partitions/src/journal.rs
```diff
@@ -662,6 +679,25 @@ where
...
+    /// falls back to the on-disk segments. Called by a partition purge, which
+    /// wipes the segments but must KEEP the journal entries themselves:
+    /// headers, storage, `op_to_storage_offset` and the evicted ring are
+    /// consensus history that backups, repair and retransmission still walk.
+    /// Clearing those would wedge `commit_min` until a view change.
+    ///
+    /// `floor` (the purge's fence op) makes the seal survive eviction:
+    /// `evict_prefix` re-appends the retained tail, and without the floor
+    /// that re-append would re-index the pre-purge entries just cleared.
+    pub fn clear_poll_index(&self, floor: u64) {
```
numinnex, line 810: The seal reaches `offset_to_op`/`timestamp_to_op` and the `append` path, but not the resident tier itself. `resident_entries()` (`:619`) returns fenced pre-purge entries, `select_resident` (`:969`) filters on the query only, and the `DiskReadOutcome::Empty` retention-recovery branch at `core/partitions/src/poll_plan.rs:283` runs with no `oldest_resident` gate at all. So a poll can serve purged bytes; today that is masked by tier selection rather than by the invariant, which is a thin thing to rest on.

Suggest filtering in `resident_entries()` rather than in `select_resident`: the positionally-paired `headers` vec (length-lock invariant at `:531-537`, already relied on by `committed_prefix` at `:441`) lets the filter read `header.op` out of a dense `Vec<PrepareHeader>` with no payload touch, whereas filtering in `select_resident` costs a header cast plus a cold cache line into each ~1 MiB buffer per entry per poll. A `poll_floor == 0` short-circuit keeps the never-purged path byte-identical.

Related, on this function's own contract: zeroing `journal.info` in `purge()` (`iggy_partition.rs:3241`) makes the `messages_count == 0` gate at `:1925` short-circuit before `evict_committed_prefix`, and the shutdown force-flush hits that same gate and only logs. The fenced entries are bounded in bytes but have no eviction driver in time, so an idle purged partition pins them until the process exits — on the operation users run to reclaim space. If you evict inside `purge()` to fix that, it must be `committed_prefix(commit_min).len()`, not `commit_max`: evicting ops in `(commit_min, commit_max]` clears their `headers` (`:477`) and holes the backup's `collect_committable_from_journal` walk (`iggy_partition.rs:1778-1788`), wedging `commit_min`. `header_by_op` is also what `RetransmitPrepares` resolves (`iggy_partition.rs:1822-1828`) and the evicted ring is invisible to it, so uncommitted fenced ops must stay resident.

core/metadata/src/impls/metadata.rs
```diff
@@ -3394,30 +3381,10 @@ where
...
-                    return Ok(build_prepare_message(
-                        consensus,
-                        &header,
-                        Operation::UpdateTopic,
-                        &body,
-                    ));
-                }
-                Ok(message.project(consensus))
-            }
+            // `UpdateTopic` deliberately takes the default arm: unlike create,
```
numinnex, line 3321: Dropping this arm is right for the wire echo but it silently disables retention enforcement. With the sentinel now stored, `update_topic(ServerDefault)` leaves `core/server-ng/src/segment_cleaner.rs:88` mapping it to no size cap, while `CreateTopic` still stamps the node default at `:3320-3325`. The same topic therefore gets enforced retention if the value came from create and unbounded growth if it came from an update, and the config validator proves these are real values, not sentinels (`server_ng_config/validators.rs:108-128`).

For the record on legacy, since the commit message frames this as parity: legacy does not resolve at update either — `get_topic_config` returns the raw stored pair (`core/server/src/metadata/reader.rs:722-734`), resolution is create-only (`core/server/src/shard/system/topics.rs:49-50`). But its cleaner then gates on `!matches!(max_topic_size, Unlimited)` (`core/server/src/shard/system/segments.rs:65`), so a stored `ServerDefault` enters the sized branch with `max_bytes = 0` (`core/common/src/utils/topic_size.rs:52`), `threshold = 0`, and `current_size < 0` is never true — it trims *every* sealed segment. So the wire echo is genuinely legacy-identical, but enforcement now diverges in the opposite direction from legacy and neither side honors `system.topic.max_size`. The expiry half is a real parity win (`core/common/src/types/segment.rs:74` and `segment_cleaner.rs:76` agree).

Suggest resolving `ServerDefault` against node config in `topic_retention_config`/the cleaner and keeping the verbatim wire echo — read-back shape and enforcement are separate concerns. Please also split the "matching the legacy cleaner" comment at `segment_cleaner.rs:71-73`, which is true for expiry and false for size. One caveat to accept explicitly: a resolved sentinel makes local GC config-dependent where a stored `Custom` is not, so replicas would need matching `system.topic.max_size`.

Also, this behavior change lives in metadata but its only guard is a TCP integration test. There is a create-side sibling unit test at `:4379`; a mirror asserting the `UpdateTopic` prepare now carries the sentinel verbatim would be worth adding.

core/server-ng/src/bootstrap.rs
```diff
@@ -3465,6 +3475,9 @@ const fn operation_triggers_partition_reconcile(op: Operation) -> bool {
             | Operation::DeleteTopic
             | Operation::DeleteStream
             | Operation::DeletePartitions
+            | Operation::PurgeStream
+            | Operation::PurgeTopic
+            | Operation::TruncatePartition
```
numinnex, line 3643: Adding `TruncatePartition` here shortens the fuse on a pre-existing data-loss path. `PurgeTopic`/`PurgeStream` bump `purge_generation` but never reset `deleted_up_to_offset` (`core/metadata/src/stm/stream.rs:1778-1781`), and `reconcile_segment_truncations` re-stages `request_truncate_partition(ns, watermark)` on every pass whenever `watermark != 0`, with no applied-guard (`partition_reconciler.rs:1044-1065`) — unlike purge, which at least has a generation guard.

So after a `delete_segments` followed by a `purge`, post-purge offsets restart at 0 and the first post-purge segment to seal below the stale pre-purge watermark is removed by `remove_sealed_segments_up_to`. Its only brake is `min_committed_offset()` (`iggy_partition.rs:2944`, `:2951-2956`), which the purge just cleared. Waking the reconciler on the commit tick plus `purges_pending` keeping passes running both make this land sooner.

Fix looks small and self-contained: zero `deleted_up_to_offset` in both purge applies. It is a deterministic STM mutation on every replica, cannot un-delete anything, and is semantically forced — a watermark in the pre-purge offset space is meaningless after the reset.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -1077,6 +1094,7 @@ fn reconcile_partition_purges(ctx: &ReconcilerCtx) {
             .map_or(0, partitions::IggyPartition::applied_purge_generation);
         if committed > applied {
             ctx.shard.request_purge_partition(namespace, committed);
+            counters.purges_pending += 1;
```
numinnex: `purges_pending` correctly disarms the fast-skip, but there is no backoff behind it, unlike `reconcile_partition_teardowns` at `:800-803`. With a persistently failing purge (ENOSPC/EROFS/EIO on `persist_purge_generation`) this restages a frame per stuck partition on every ~1s pass, and each retry re-runs the entire `purge()` on the pump: unlink every segment file, two file creates in `install_empty_segment`, fsync. On a 1000-partition topic that is a data-plane stall on a single-threaded pump.

It also re-reads `current_sequence()` at `iggy_partition.rs:3236` on every attempt, so `purge_floor_op` climbs monotonically and the fenced set ratchets over sends acked since the previous attempt. That turns the cross-replica divergence from a one-shot skew bounded by reconciler timing into a window that grows for the duration of the fault, since a failing replica's floor climbs while healthy replicas' floors stay put.

Please keep the counter — it is the only self-heal for a stranded purge, and before this PR a failed purge armed the fast-skip and stayed stranded until an unrelated commit. The ask is `failure_state` backoff plus an attempt cap and a `warn!`.

Two related gaps while here: `purges_pending` and the pre-existing `trims_pending` feed `total()` but are absent from the pass-complete `debug!` (`:473-486`), so an operator cannot see why passes stopped fast-skipping. And a failing purge makes the repair-serve defer at `shard/src/lib.rs:3333` drop every repair request silently and forever — it is `debug!`-only with no attempt counter or metric, so the cluster loses redundancy invisibly.

core/server-ng/src/partition_helpers.rs
```diff
@@ -536,6 +536,11 @@ pub async fn build_partition_fresh(
         topic_id,
         partition_id,
     ));
+    // Fresh dirs read generation 0; a dir surviving from a crashed process
+    // (this "fresh" build races repair re-materialization) reads the last
+    // durably-applied purge so the reconciler does not re-wipe messages
+    // appended after it.
+    partition.hydrate_applied_purge_generation().await?;
```
numinnex, line 692: Two problems with hydrating here.

First, this sets `applied_purge_generation = N` while `purge_floor_op` stays 0, so the replica claims the purge is applied yet carries no fence. Its repair-serve then skips the defer at `shard/src/lib.rs:3333` and raises no floor, which is exactly the state the fence was added to prevent.

Second, a `purge.gen` that survives a failed teardown (`partition_reconciler.rs:817-836`, with the delete-then-recreate race already named at `:782`) gets hydrated by a partition recreated on the reused slab ids, whose `purge_generation` restarts at 0 (`core/metadata/src/stm/stream.rs:1615`, `:1853`). Then `committed > applied` is false at `partition_reconciler.rs:1095`, the wipe never runs, and the client still gets Ok for the purge.

On the fix: unlinking `purge.gen` on a "fresh" build is not safe, because this function cannot distinguish a true fresh create from a restart materialization — its only signal is `restarted = dir exists && replica_count > 1` (`:470-476`), which is unconditionally false single-node, so unlinking would re-arm the re-purge on a genuine restart and destroy post-purge data. Comparing `committed != applied` is the minimal safe change, but it must be applied at *both* sites: the router's own gate is `partition.applied_purge_generation() < generation` (`core/shard/src/router.rs:662`), and `N < 1` is false, so changing only the reconciler leaves the purge still swallowed. Keying the file on `Partition::created_revision` (`stm/stream.rs:1849`) is the durable version.

core/server-ng/src/dispatch.rs
```diff
@@ -724,6 +730,80 @@ fn pop_next_client_request(
     message
 }
 
+/// Per-request partitions-count cap shared by create-topic, create-partitions
+/// and delete-partitions admission. Runs pre-consensus like
+/// [`validate_topic_bounds`]: an oversized count must not burn a replicated
+/// log entry (create-partitions admission would also allocate that many
+/// consensus-group ids before replicating).
+pub(crate) const fn validate_partitions_count(partitions_count: u32) -> Result<(), IggyError> {
+    if partitions_count > MAX_PARTITIONS_PER_REQUEST {
```
numinnex, line 753: The cap is checked but zero is not, and legacy TCP rejects zero on both partitions ops: `core/server/src/binary/handlers/partitions/create_partitions_handler.rs:43` (`!(1..=MAX)`) and `delete_partitions_handler.rs:42` (`== 0`). HTTP already rejects it too (`core/common/src/http/partitions/create_partitions.rs:53`, `delete_partitions.rs:53`), so this also splits TCP from HTTP within server-ng, and the new test only probes `MAX+1`.

It is not a harmless no-op reply on the create side: `create_partitions(0)` passes validation, `allocate_many(0)` yields an empty vec, and `CreatePartitionsWithAssignmentsRequest::apply` bumps `state.revision` unconditionally and calls `rebalance_consumer_groups()` (`core/metadata/src/stm/stream.rs:1838-1858`). Each such request burns a replicated log entry, forces a consumer-group rebalance, and disarms the reconciler fast-skip into a full O(N) diff plus truncation and purge sweeps on every shard — client-driven amplification from a call that creates nothing. `delete_partitions(0)` is a clean no-op by comparison (`:1884-1898`), costing one wasted log entry.

Suggest `1..=MAX_PARTITIONS_PER_REQUEST` for both partitions ops, keeping `0..=MAX` for create_topic since legacy allows it there (`create_topic_handler.rs:47`).

Also worth folding in while this area is open: the typed `PartitionNotFound` this PR adds is TCP-only. HTTP poll collapses it into a generic `ResourceNotFound` (`http/handlers.rs:1041`) where legacy HTTP propagates the typed error with a `partition_id` field (`core/server/src/http/error.rs:50,90`); `get_consumer_offset` still replies an empty body for a bad partition id (`:1971-1978`) where legacy returns `PartitionNotFound` (`core/server/src/shard/system/consumer_offsets.rs:124`); and send / store|delete-consumer-offset / delete-segments hardcode `ResourceNotFound` at `:1218`, discarding the resolver's typed error, where legacy returns 3007. The last two are pre-existing, but they are the same "empty success a consumer reads as end-of-data" class this PR set out to close.

core/server-ng/src/responses.rs
```diff
@@ -1193,13 +1176,18 @@ fn partition_response(
...
+    // initial state instead of a zero a client would read as "no storage".
     let stats = streams
         .stats_registry
         .partition_get(stream_id, topic_id, partition.id);
     let (segments_count, current_offset, size_bytes, messages_count) =
-        stats.map_or((0, 0, 0, 0), |stats| {
+        stats.map_or((1, 0, 0, 0), |stats| {
             (
-                stats.segments_count_inconsistent(),
+                stats.segments_count_inconsistent().max(1),
```
numinnex, line 1219: The `.max(1)` on the materialized read hides a real zero. A partition whose owning shard actually has zero segments — a failed `install_empty_segment` inside `purge()` (`core/partitions/src/iggy_partition.rs:3141-3171`) leaves exactly that state, and it is also the state that then panics the shard via the `segments().len() - 1` underflow at `:1008` or `active_segment()`'s `expect` at `core/partitions/src/log.rs:338` — reports as one healthy empty segment. That is the one case where monitoring most needs the truth. Legacy never fabricates the count (`core/server/src/binary/handlers/topics/get_topic_handler.rs:64`); it just happens to be 1 because create increments synchronously.

The registry-miss `1` on the line above is more defensible as the deterministic committed state, but the comment at `:1180-1183` is factually wrong about when it applies: `partition_response` iterates the metadata STM, whose row outlives a shard-side tombstone and a mid-teardown, so the miss window is not only "before that first materialization" — a torn-down partition also reports one empty segment.

Suggest dropping `.max(1)` and correcting the comment. Worth noting the fabricated value silently defeated the suite's only materialization check, whose convergence loop this PR removed (`core/integration/tests/server/scenarios/system_scenario.rs:175`).

numinnex (changes requested):

core/partitions/src/iggy_partition.rs
hubcio: agreed - the floor is read off the local sequencer at local apply time, so replicas fence different in-flight sets and an ack in the apply window can die. a real fix needs the purge ordered against the partition plane (checkpoint barrier or a replicated purge op), which is a protocol change, not something to smuggle into this PR. the TODO at the reconciler now names both skew modes; tracking it as a follow-up.

core/partitions/src/iggy_partition.rs
hubcio: both halves are gone after the rebase onto the state-transfer machinery. the frontier reset is recorded before anything is mutated, and a failed generation record now sets `purge_deferred`, which withholds `PrepareOk` for the group until the re-issued purge lands - so nothing gets acked at offset 0 inside the retry window (single node included, the commit rides the loopback ack). ordering: the partition dir is fsynced after the unlinks and the empty plant, before the generation write, so `purge.gen` can never be durable while the deletions are not.

core/shard/src/lib.rs
hubcio: resolved by the rebase. an empty repaired window (floor >= to_op) now counts as complete, refusal is definitive (session dropped, `FloorRefused`), and the shard converts it to a partition state transfer instead of retrying forever. the serve side also defers while a committed purge is not yet locally applied, so the window you describe cannot even be served pre-purge.

core/partitions/src/iggy_partition.rs
hubcio: kept the empty-confirmations shape, and documented why at the site: the op did commit cluster-wide (the fence only stops the local flush of purged bytes), so a typed transient status would invite a client retry that duplicates the send into the post-purge offset space. the reply is byte-identical to a send without confirmation, which is the established degradation for offsets-unavailable.

core/partitions/src/iggy_partition.rs
hubcio: fixed. the receive side now mirrors the serve gate: `on_repair_range_reply` defers the whole reply (floor install included) while the committed purge generation is ahead of the local applied one, so the stale durable line is cleared by the local purge before any floor can be accepted. deferring both arms beats gating just completion - a deferred `RangeEvicted` is re-derived by the stall retry.

core/partitions/src/iggy_partition.rs
hubcio: fixed - the guard is in. the entry is still journaled (dropping it would leave a hole and wedge `commit_min` on the walk), but a repaired send at or below the floor no longer touches `dirty_offset`, segment position or `journal.info`, and returns no offset so it cannot anchor `first_batch_offset` in the floor-connect check.

core/partitions/src/journal.rs
hubcio: fixed on both halves. `resident_entries()` (the snapshot accessor the poll walk splices on; the index path is fenced separately by the same floor at insert) filters ops at or below the floor, so a post-purge re-index cannot re-expose fenced entries. and the purge now hands the fenced prefix to `evict_committed_prefix` bounded at `commit_min` - not `commit_max`, evicting an op the commit walk has not reached would delete its header and wedge the walk on the hole. the ring still serves repair, and the serve path clamps `retained_from` above the floor anyway.

core/metadata/src/impls/metadata.rs
hubcio: fixed. the cleaner now resolves `ServerDefault` against the node default at enforcement time (new `default_max_topic_size()` accessor + an extracted `per_partition_size_budget` so it is unit-testable). explicit sizes ignore the default; a node default of unlimited still means no cap, so it cannot turn into a trim-everything budget.

core/server-ng/src/bootstrap.rs
hubcio: fixed. both purge applies (`PurgeStream` and `PurgeTopic`) now zero `deleted_up_to_offset` - replicated state, so every replica converges on the same reset and the reconciler has nothing stale to re-stage. `PartitionSnapshot` already carried the field, so the snapshot shape is unchanged.

core/server-ng/src/partition_reconciler.rs
hubcio: took the observability half: `purges_staged` and `trims_pending` are in the pass-complete debug line now, and both defer sites (serve + receive) bump a new `partition_repair_serves_deferred_purge_total` counter, plus a per-attempt warn on each deferred purge. left the backoff out deliberately - staging is a `try_send` of one frame and the pump re-checks the generation, so a restage per pass is the pacing, and the frontier-record failure path already respects the superblock writer's own backoff and mutates nothing when it refuses.

core/server-ng/src/partition_helpers.rs
hubcio: the incarnation hole is fixed: `purge.gen` now records `[generation][created_revision]` (the STM row's epoch), and hydration treats a mismatched incarnation as 0, so a file surviving a failed teardown cannot swallow the next topic's purges. left the `committed > applied` gate alone - `!=` would restage forever whenever a stale applied ran ahead of committed, which pins the reconciler scan. on the first problem: a fresh build carries no segments and the resident journal dies with the process, so there is nothing pre-purge for the hydrated generation to mis-fence - the debug_assert on the build path pins that.

core/server-ng/src/dispatch.rs
hubcio: checked legacy first and it splits: create_topic admits zero (`0..=MAX` in the shared validator), while create/delete partitions reject it - so the zero gate went into a separate `validate_partitions_change_count` used only by the two change paths, keeping create_topic parity. also took the two typed-error gaps while in there: tcp `get_consumer_offset` now denies typed on `PartitionNotFound` (other resolver errors keep the empty body so "no offset yet" still reads as none), and the http poll passes `PartitionNotFound` through instead of a generic 404.

core/server-ng/src/responses.rs
hubcio: looked hard at this one and the premise does not hold: the registry entry is minted get-or-create by the reconciler before `build_partition_fresh` runs, so "registered with zero segments" is the ordinary mid-build window, not the fenced state. i tried dropping the clamp - it deterministically breaks the get-topic-right-after-create contract (one empty segment, no convergence loop, which the system scenario pins). a fenced partition also reading one empty segment is the accepted cost and the comment now says so explicitly. separating the two needs a materialization signal the stats registry does not carry today, so that is a design change rather than a tweak here.

mmodzelewski (approved):

piotr-laserdata (approved):

spetz (approved):

numinnex (approved):

krishvishal (approved):

PR #3832, opened 2026-08-07
test(integration): default vsr harness to single node

hubcio:
Building with the vsr feature forced every test that does not
pin cluster_nodes onto a 3-node cluster, tripling server spawns
for tests that do not exercise cluster behavior.

Default to a single node regardless of features. The existing
IGGY_TEST_CLUSTER_NODES env var still scales the whole suite
(e.g. =3 to probe replication), and explicit cluster_nodes
attributes keep winning over both.

Also drop the stress produce/consume hammer scenarios; they
add little coverage over the existing produce/consume suites.


hubcio:
closing this, too early for this PR

PR #3834, opened 2026-08-07
feat(go): migrate SDK to the VSR wire protocol

spetz:
The Go SDK still used classic framing and could not communicate
with the VSR server or follow consensus leader changes safely.

Replace classic framing with VSR envelopes, session registration,
metadata-based leader discovery, bounded replay, and TCP/TLS
failover. Add confirmations, partition routing, group synchronization,
raw commands, and VSR-only CI coverage.

Keep VSR as the only Go wire protocol to avoid duplicate compatibility
paths and test it through race, TLS, e2e, BDD, and example suites.


thepatrykk (dismissed):

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
+		c.mtx.Lock()
+		c.skipAutoLoginOnce = true
+		c.mtx.Unlock()
+	}
+
+	c.logger.Info("Reconnecting to the server...",
+		slog.String("server_address", c.currentServerAddress),
+		slog.Any("error", err))
+
+	if reconnectErr := c.Connect(ctx); reconnectErr != nil {
```
numinnex, line 517: **critical — reconnect + replay re-applies a committed write**

This replays the request after a reconnect for *any* reconnectable error, and `Connect` at line 845 calls `c.session.Reset()`, which mints a fresh u128 client id and resets `requestCounter` to 1. `sendFrame` then re-enters with `stamped := false` (line 478), so the frame is re-stamped under the new identity.

The server dedups on `(client_id, request)`. Both halves changed, so `client_table.rs:613` sees `request=1 > watermark=0` and returns `RequestStatus::New` — the request is dispatched again. The trigger is the ordinary "write landed, reply lost" case: `readReplyLocked` maps a failed read to `ErrDisconnected` (line 642), which `isReconnectable` accepts.

Consequences for non-idempotent replicated ops:
- `CreatePartitions` is documented additive (`contracts/client.go:212-214`), so the replay lands 2N partitions.
- `CreatePersonalAccessToken` mints a second token the caller never learns, because the reply that carried the first one was lost.
- `SendMessages` duplicates the batch (disclosed for sends, but not for the metadata ops).

The Rust SDK deliberately refuses this: `core/sdk/src/tcp/tcp_client.rs:775-782` returns `Err(IggyError::Disconnected)` instead of replaying, and `core/sdk/src/vsr.rs:44-63` carries a standing TODO naming this exact hazard — "the retried request gets a new (client_id, request_id) tuple, so server-side deduplication cannot match a mutation that may already have committed" — with client-id-preserving session resume as the prerequisite.

Suggested fix: make the replay per-operation rather than blanket. Replay `ErrTransientNotAccepted` (never admitted), non-replicated reads, and `Register`; return `ErrDisconnected` for anything `OperationForCode` maps to a replicated operation. A blanket removal would break the deliberate login replay at lines 418-419.

Note `tcp_connect_test.go:262-270` currently asserts this behaviour as intended ("the replay signs in again before it repeats the request"), so that test needs rewriting alongside the fix.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -608,11 +840,78 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
...
+// establishSession points the connection at the leader and signs in when
+// auto-login is configured.
+//
+// Leadership is settled before the sign-in, and it is settled even when
+// auto-login is off: register is a consensus operation that a backup answers
+// transiently, so signing in against a follower replays for the whole request
+// budget instead of failing over. Cluster metadata is sessionless and works on
+// the unauthenticated connection.
+func (c *IggyTcpClient) establishSession(ctx context.Context, skipAutoLogin bool) error {
+	redirect, err := c.HandleLeaderRedirection(ctx)
```
numinnex: **critical — unbounded reentrant `Connect` through the request path**

`establishSession` issues a request (`HandleLeaderRedirection` → `util.CheckAndRedirectToLeader` → `c.GetClusterMetadata`) while it is itself running inside `Connect`. That request goes through `do` → `exchange`, and `exchange` reconnects on a reconnectable error, so the cycle closes:

```
exchange:424 disconnect()          -> transportState = Disconnected
exchange:437 Connect()
Connect:749  default:              -> Connecting, dial
Connect:841                        -> Connected
Connect:851  establishSession()
     :889    HandleLeaderRedirection()
             -> GetClusterMetadata -> do -> exchange
exchange     read fails            -> ErrDisconnected (isReconnectable)
exchange:424 disconnect()          -> recurse
```

The already-connected short-circuit at line 740 never fires, because `exchange` calls `disconnect()` (and `invalidateConnLocked` at line 706 does the same on the read failure) before the nested `Connect` is entered, so every level observes `Disconnected` and takes the `default:` branch.

There is no depth counter on this path — `leaderRedirectionState.CanRedirect()` bounds the redirect arm, not the reconnect arm, and `leader_aware.go:44-48` only swallows the metadata error *after* the recursion would have unwound. Each level opens a fresh TCP connection. With `context.Background()` it recurses to stack exhaustion, which is a fatal unrecoverable runtime error in Go; with `maxRetries: 0` (the default) each level's dial retries forever instead of unwinding.

Reachable with nothing but `WithAutoLogin` (the documented setup) plus a dial that succeeds and a first read that fails reproducibly: accept-then-close under a connection limit, an LB draining, or a plaintext client pointed at a TLS port.

Suggested fix: suppress the reconnect path for requests issued from inside `Connect`/`establishSession` (a `connecting` flag, or a no-reconnect variant of `do`), and bound reconnect depth.

foreign/go/client/tcp/tcp_group_polling.go
```diff
@@ -0,0 +1,228 @@
...
+		synced, err = c.SyncConsumerGroup(ctx, streamId, topicId, groupId)
+		if err != nil {
+			return groupAssignment{}, err
+		}
+		if synced == nil {
+			return groupAssignment{}, ierror.ErrConsumerGroupMemberNotFound
+		}
+	}
+
+	assignment := groupAssignment{
```
numinnex, line 270: **critical — assignment refresh resets the round-robin cursor, starving partitions**

This builds a fresh `groupAssignment` with the implicit zero `cursor` and `put`s it on *every* `assignmentRefreshInterval` expiry (line 196-199, 5s), including when `synced.Generation` equals the cached generation. `put` at line 83 stores `&assignment`, so the zero cursor is what `pollGroup` reads at line 156.

Concrete: 4 owned partitions, polling every 2s → p0, p1, p2, then the 6s refresh resets the cursor → p0 again. **p3 is never polled and its lag grows without bound.** At a poll interval of 5s or slower, only `partitions[0]` is ever read — permanent non-delivery for every other owned partition, not just skew.

The Rust reference resets the cursor only on a real rebalance:

```rust
// core/common/src/consumer_group_client_state.rs:72-84
if entry.generation != generation { entry.cursor = 0; }
```

Node does the same, preserving `position` on a same-generation re-sync (`foreign/node/src/wire/message/poll-messages.command.ts:172-177`).

The comment at lines 31-34 ("it only bounds the drift of an idle member") does not hold — the reset hits an actively polling member on every window.

Suggested fix: carry the previous cursor forward when `generation` is unchanged. No shrink clamp is needed, because line 156 already does `cursor % len(partitions)` at read time. Worth doing together with moving the refresh out of band (the inline 5s sync also costs 2 RTTs per group per window; Rust drives it from the heartbeat, `messages.rs:110`).

Nothing covers `pollGroup` at unit level, and `foreign/go/tests/e2e_test.go:167` polls in a tight sub-5s loop, so the suite passes through this.

foreign/go/client/tcp/tcp_group_polling.go
```diff
@@ -0,0 +1,228 @@
...
+		time.Since(cached.fetchedAt) < assignmentRefreshInterval {
+		return cached, nil
+	}
+
+	synced, err := c.SyncConsumerGroup(ctx, streamId, topicId, groupId)
+	if err != nil {
+		return groupAssignment{}, err
+	}
+	if synced == nil {
+		if err := c.JoinConsumerGroup(ctx, streamId, topicId, groupId); err != nil {
```
numinnex: **critical — an explicit `LeaveConsumerGroup` is silently undone by the next poll**

`JoinConsumerGroup`, `LeaveConsumerGroup` and `DeleteConsumerGroup` (`tcp_consumer_group_management.go:89`, `:100`, `:78`) never touch `c.groups`, so the assignment cached here survives the client leaving. The next `PollMessages` then walks straight back into the group:

```
pollGroup:145  ensureAssignment  -> cache hit (<5s), stale assignment
          :160 pollPartition     -> fenced (member no longer owns it)
          :163 isAssignmentStale -> true
          :166 groups.drop
          :167 continue
          :145 ensureAssignment  -> miss
          :201 SyncConsumerGroup -> nil (not a member)
          :206 JoinConsumerGroup -> re-joins the group the caller just left
```

The implicit join here is reasonable for a first-time poll, but combined with the missing invalidation it reverses an explicit user action with no error returned. Rust invalidates *and* deregisters for exactly this reason (`core/common/src/traits/binary_impls/consumer_groups.rs:162-164` on join, `:188-193` on leave).

Suggested fix: drop the group cache entry in `JoinConsumerGroup`/`LeaveConsumerGroup`/`DeleteConsumerGroup`, and gate the implicit join at line 206 on not having explicitly left.

foreign/go/client/tcp/tcp_messaging.go
```diff
@@ -32,22 +34,96 @@ func (c *IggyTcpClient) SendMessages(
...
-		Partitioning: partitioning,
+		Partitioning: resolved,
 		Messages:     messages,
 	})
-	return err
+	if err != nil {
+		return nil, err
+	}
+
+	confirmations, err := binaryserialization.DeserializeSendMessagesConfirmations(response)
```
numinnex, line 72: **critical — an empty status-0 reply is reported as a successful send**

A zero-length reply body reaches this call as `nil` with no error, and comes back out as an empty-but-successful `SendMessagesResponse`:

```
tcp_core.go:654                bodyLength == 0            -> nil body
vsr/reply.go:74-77             0 < 0 false                -> passes
vsr/reply.go:81                status == 0                -> no error
vsr/reply.go:99                IsResultFramed(160) false  -> passthrough
binary_response_deserializer.go:634  len == 0             -> &SendMessagesResponse{}, nil
```

The caller sees success. Nothing was written.

Server-ng emits precisely that frame — `build_empty_reply` (`core/server-ng/src/responses.rs:1259-1265`, which leaves `status = 0`; only `build_deny_reply` sets it) for any replicated request arriving on a transport whose session is gone while the client still believes it is bound (`core/server-ng/src/dispatch.rs:836-866`). Its own comment there states the intent: the empty body exists so "the typed decoder downstream rejects the empty body with `InvalidCommand`" rather than hanging. That holds for every result-framed operation (`vsr/reply.go:105-106`) and fails for `SendMessages`, the one replicated operation with `IsResultFramed == false`.

For the record on origin: the Rust SDK has the identical hole — `decode_send_confirmations` (`core/common/src/traits/binary_impls/messages.rs:265-272`) accepts empty → `Ok` unconditionally, with no `cfg(vsr)` gate. So this is a protocol-contract gap shared with the reference, not Go drift. But the doc rationale there is explicitly legacy-scoped ("the legacy server answers that way"), and this SDK dropped legacy framing entirely, so the guard is free here.

Suggested fix, both sides: `dispatch.rs` should use `build_deny_reply` with `NoSession` instead of `build_empty_reply`; and this SDK should reject a zero-length body for `OperationSendMessages` as defence in depth.

Separately: the swallow below (lines 66-71) is correct for a *non-empty* body that fails to parse — it matches `committed_send_confirmations` (`messages.rs:277-286`) and surfacing that error would turn one committed write into N retries. Only the empty-body case should change.

foreign/go/client/tcp/tcp_group_polling.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+		if len(assignment.partitions) == 0 {
+			return &iggcon.PolledMessage{
+				PartitionId: iggcon.NoAssignedPartition,
+				Messages:    []iggcon.IggyMessage{},
+			}, nil
+		}
+
+		cursor := assignment.cursor % len(assignment.partitions)
+		partitionId := assignment.partitions[cursor]
```
numinnex: **warning — the cursor read-modify-write is not atomic**

`c.groups.get` (line 67) returns a *copy* of the entry, and `advance` (line 88) then writes an absolute cursor derived from that copy, in a second critical section. Two concurrent `PollMessages` calls on the same group both read cursor N, both poll the same partition, and both write N+1: one partition is polled twice, another is skipped, and the cursor can move backwards. With `autoCommit` the second poll re-reads already-committed offsets.

The cache is deliberately guarded by its own lock rather than the connection lock (lines 60-61), so concurrent entry is the designed case, not a misuse.

`topicCache.nextBalanced` (`tcp_topic_cache.go:84-96`) already does this correctly under a single lock, and Rust does read+increment under one lock too (`core/common/src/consumer_group_client_state.rs:95-110`).

Suggested fix: one `nextPartition(key) (uint32, bool)` on `groupAssignmentCache` that does refresh-if-stale, carries the cursor forward on an unchanged generation, and then reads and advances — all inside one `mtx` hold. Folding the whole operation into a single method also removes the three separate ways (`get`/`put`/`advance`) to reintroduce this. This is necessary but not sufficient on its own: the cursor reset at line 218 is an independent defect.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -608,11 +840,78 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
 	c.conn = conn
 	c.transportState = iggcon.TransportStateConnected
 	c.connectedAt = time.Now()
+	// The server fence does not survive the old socket, so the new connection
+	// starts from a fresh client identity.
+	c.session.Reset()
+	skipAutoLogin := c.skipAutoLoginOnce
```
numinnex, line 983: **warning — `skipAutoLoginOnce` is consumed by the wrong `Connect`, causing a double `Register`**

The flag is read and cleared here, but `establishSession` recurses into `c.Connect(ctx)` on a leader redirect (line 894). The nested `Connect` reads the flag as already false and performs the automatic sign-in the flag exists to suppress. Then `exchange` line 445 replays the original login frame, whose `BeginRegister` (`vsr/session.go:95-98`) resets the just-bound session and mints another client id.

Result: two `Register` commits, and the first registration is orphaned in the server's client table until it is fenced or evicted for capacity.

Suggested fix: pass `skipAutoLogin` through the redirect recursion instead of re-reading the field at the nested level.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
 }
 
-func finalizeWireRequest(buf []byte) []byte {
-	binary.LittleEndian.PutUint32(buf[0:4], uint32(len(buf)-4))
-	return buf
+// isReconnectable reports whether a fresh connection can recover the failure.
+func isReconnectable(err error) bool {
+	reconnectable := []error{
+		ierror.ErrDisconnected,
+		ierror.ErrEmptyResponse,
```
numinnex, line 562: **warning — `ErrUnauthenticated` here reverses an explicit logout and tears down healthy connections**

Two problems follow from treating this sentinel as reconnectable.

1. An explicit `LogoutUser` is silently undone under `WithAutoLogin`:

```
tcp_session_management.go:110  session.Reset()
next replicated call
vsr/envelope.go:50-52          ErrUnauthenticated
exchange:407                   isReconnectable -> true
exchange:437                   Connect
establishSession:911           LoginUser        -> signed back in
```

The caller asked to log out, gets no error, and the client re-presents the stored credentials.

2. `vsr.StampRequestHeader` returns this error *locally, before any byte is written* (`vsr/envelope.go:50-52`), for a replicated operation on an unbound session. So a purely local precondition failure closes a healthy TCP connection and re-dials.

Suggested fix: distinguish a local pre-write precondition failure from a server-side `Unauthenticated` before consulting `isReconnectable`, and keep an explicit-logout flag that suppresses the auto-login reconnect.

foreign/go/internal/vsr/reply.go
```diff
@@ -0,0 +1,164 @@
...
+			}
+		}
+	case EvictionMalformedLogin:
+		mapped = ierror.ErrInvalidFormat
+	case EvictionReserved, EvictionClientReleaseTooLow, EvictionClientReleaseTooHigh,
+		EvictionInvalidRequestOperation, EvictionInvalidRequestBody,
+		EvictionInvalidRequestBodySize:
+		mapped = ierror.ErrInvalidCommand
+	default:
+		mapped = ierror.ErrUnauthenticated
```
numinnex: **warning — the fallback diverges from the mapping this function claims to mirror, and it is reconnectable**

The comment at lines 123-124 says this is "the same mapping as eviction_reason_to_error in core/common". Every enumerated arm does match; only the fallthrough drifts — `core/common/src/error/eviction.rs:67` maps `_ => IggyError::InvalidCommand`, not `Unauthenticated`.

That difference has teeth rather than being cosmetic: `ErrUnauthenticated` is in `isReconnectable` (`client/tcp/tcp_core.go:452`), so a single unrecognized reason byte from a newer server silently drives disconnect → re-dial → re-login instead of surfacing. It is also a live feeder for the two criticals on this PR (the replay double-apply and the reentrant `Connect`), which both need a reconnectable error to start.

This is the cheapest high-leverage fix in the change: `default: mapped = ierror.ErrInvalidCommand`.

`protocol_parity_test.go` asserts the eviction reason *discriminants* but never this mapping, which is why the drift is unguarded — worth adding as a parity case.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
+	ctx context.Context,
+	code uint32,
+	frame []byte,
+	transientDeadline, readDeadline time.Time,
+) ([]byte, error) {
+	for {
+		c.logger.Debug("Sending a TCP request",
+			slog.Int("frame_length", len(frame)), slog.Int("code", int(code)))
+		if _, err := c.write(frame); err != nil {
+			c.invalidateConnLocked()
```
numinnex, line 718: **warning — write failures are not normalized, unlike read failures**

This returns `c.write`'s raw error (`*net.OpError`, or `io.ErrNoProgress`), while both read paths normalize to `ierror.ErrDisconnected` (lines 642 and 663). `isReconnectable` (lines 450-458) matches only `ierror` sentinels, so an ECONNRESET on write is neither reconnectable nor matchable by a caller doing `errors.Is(err, ierror.ErrDisconnected)` — the documented way to detect a dropped connection.

Suggested fix: map write failures to `ierror.ErrDisconnected` (or wrap in `ErrTcpError`) so classification is uniform on both directions.

One ordering caveat: this asymmetry is currently what *limits* the reentrant-`Connect` recursion (a write failure terminates instead of recursing). Land the recursion guard first, or fixing this in isolation widens that bug.

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -22,58 +22,95 @@ import (
...
 		return nil, err
 	}
-	if shouldRedirect {
-		if err = c.Connect(ctx); err != nil {
-			return nil, err
-		}
-		return c.login(ctx, loginCmd)
+
+	c.mtx.Lock()
+	err = c.session.Bind(registered.Session)
```
numinnex, line 85: **warning — the register transaction spans two critical sections**

`attempt` releases `c.mtx` when it returns (`tcp_core.go:527-528`), so `BeginRegister` — which mints the client id that goes on the wire, inside `StampRequestHeader` — and this `Bind` sit in different critical sections. `vsr/session.go:48-50` states that the transport "serializes access through its exchange lock", which is not true for this pair.

Interleaving with two concurrent sign-ins (an app-issued `LoginUser` racing a reconnect's auto-login):

```
G1 stamps -> client A, server returns fence F1
G2 stamps -> BeginRegister sees registerConsumed -> reset() -> client B, fence F2
G1 Bind(F1) succeeds (G2's reset cleared `bound`)   -> local state = (client B, fence F1)
G2 Bind(F2) -> ErrSessionAlreadyBound
```

The stale fence itself is inert — `core/server-ng/src/dispatch.rs:897-906` overwrites the wire `client`/`session` from the server-side transport binding for replicated metadata ops, and header validation only requires `session > 0`. The observable damage is client A's committed `Register` orphaned in the `ClientTable` until capacity eviction, plus a spurious `ErrSessionAlreadyBound` on the losing caller.

Also: a `Bind` failure here returns with the connection still up and the local session unbound, so every later replicated call fails `ErrUnauthenticated` — which is reconnectable, so it degenerates into the auto-relogin loop.

Suggested fix: hold `c.mtx` across the whole register transaction (or single-flight `register`), and call `invalidateConnLocked` on a `Bind` error, the way `handleReplyFailureLocked` already treats a terminal session error.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
+		}
 
-	c.logger.Debug("Sent a TCP request, waiting for a response...", slog.Int("code", int(commandCode)))
+		response, err := vsr.DecodeReply(&c.respHeader, body)
+		switch {
+		case errors.Is(err, ierror.ErrTransientNotCommitted) && time.Now().Before(readDeadline):
+			// The outcome is unknown, so only a replay of the same request id
+			// on this session is safe. The server's client table answers from
+			// its reply cache if the request did commit.
+			if waitErr := waitBeforeReplay(ctx, readDeadline); waitErr != nil {
```
numinnex: **warning — `c.mtx` is held across the replay loop, including these sleeps**

`attempt` takes `c.mtx` with `defer` (lines 527-528) and the hold spans all of `exchangeLocked`, so this 50 ms wait and every subsequent one happen with the client mutex held — up to the full 30 s `responseReadTimeout` on a `TransientNotCommitted` sequence. `GetConnectionInfo`, `disconnect`, `Connect` and every other caller's request block for that whole window.

Worse, `shutdown` (line 983) takes the same mutex, so `Close()` cannot interrupt an in-flight replay and blocks up to 30 s while the loop keeps writing to the connection the caller asked to close. The `context.AfterFunc` escape at lines 557-569 does not help, because `Close()` takes no context.

Suggested fix, in this order:
1. Add a `closed chan struct{}` closed by `shutdown()`, and select on it in `waitBeforeReplay` alongside `ctx.Done()`. This is what actually fixes the `Close()` case, independently of lock granularity.
2. If the lock hold is also shortened (release around the wait, or hoist the loop into `sendFrame`): after reacquiring, abort unless `c.conn` is still the snapshotted `conn` from line 554. Otherwise `Connect` can install a new connection *and* call `session.Reset()`, and the replay writes bytes stamped with the dead session's client id and fence to a fresh socket, with `stamped` already true so nothing re-stamps them.
3. On such an abort, do **not** clear `stamped` for `ErrTransientNotCommitted` — the outcome is unknown, so only a same-request-id replay on the same session is sound, and re-stamping would manufacture exactly the double-apply from the `exchange` critical. Clearing `stamped` is sound only for `ErrTransientNotAccepted`, which `sendFrame` line 510 already does on redirect.

foreign/go/internal/vsr/reply.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, ierror.ErrInvalidCommand
+	}
+
+	// A pre-commit denial always ships an empty body. The status channel and
+	// the committed result section are mutually exclusive.
+	if status := ReadStatus(header); status != 0 {
+		return nil, ierror.FromCode(ierror.Code(status))
+	}
+
+	operation := ReadReplyOperation(header)
```
numinnex, line 85: **warning — no reply correlation; any Reply frame satisfies the pending request**

`DecodeReply` allowlists the operation byte but never checks that the reply's `client` (offset 160) and `request` (offset 200) echo the request just sent. Soundness therefore rests entirely on "exactly one Reply per write", plus the unconditional `invalidateConnLocked` on partial reads (`client/tcp/tcp_core.go:641`, `:662`).

That invariant is weaker after this change than before it: `exchangeLocked` (`tcp_core.go:598-631`) now writes several frames inside one read budget on the replay path, so a server that ever emits two replies for one request desyncs the stream permanently, with no detection — every later request reads the previous request's answer.

The Rust SDK has the same gap (`core/sdk/src/vsr.rs:242-262` reads status and operation only), so this is protocol-level rather than a Go regression. But the fix is two lines and free: the 8 bytes are already sitting in `c.respHeader`, so reading `request` at offset 200 and dropping the connection on a mismatch costs no syscall and no allocation, and converts a whole class of silent mis-pairing into a detectable error.

foreign/go/client/tcp/tcp_messaging.go
```diff
@@ -32,22 +34,96 @@ func (c *IggyTcpClient) SendMessages(
...
+
+	// Routing needs an explicit partition: the broker never picks one under
+	// consensus. Balanced and key-based strategies resolve here against the
+	// topic's partition count.
+	resolved, err := c.resolvePartitioning(ctx, streamId, topicId, partitioning)
+	if err != nil {
+		return nil, err
+	}
+
+	response, err := c.do(ctx, &command.SendMessages{
```
numinnex, line 55: **warning — S2 compression on this path panics (pre-existing, but this is the only caller)**

`c.MessageCompression` reaches `internal/command/message.go`, where the compression loop writes the header length to the range *copy*:

```go
// internal/command/message.go:56-57 (and :62-63, :68-69)
s.Messages[i].Payload = s2.Encode(...)                       // compressed, via index
message.Header.PayloadLength = uint32(len(message.Payload))   // loop copy, and reads the
                                                             // still-uncompressed payload
```

So after the loop `s.Messages[i].Payload` is compressed while `s.Messages[i].Header.PayloadLength` still holds the uncompressed length. `calculateMessageBytesCount` (line 149) then sizes `bytes` from `len(msg.Payload)` — the compressed length — while the write loop at lines 130-131 slices and advances by the uncompressed `Header.PayloadLength`. Every message contributes `uncompressed - compressed` of overshoot against a buffer with zero slack, so a single-message batch panics on the first iteration with `slice bounds out of range`.

Reachable from the exported `IggyTcpClient.MessageCompression` field consumed on this line. The zero value is empty, so default sends are safe — but it is a library panic on an opt-in public field.

Suggested fix: assign through the index, `s.Messages[i].Header.PayloadLength = uint32(len(s.Messages[i].Payload))`, and derive both the buffer size and the copy length from the same field.

Ordering note: land this before adding an `AppendBinary` for `SendMessages`. A rewrite that sizes and copies from `len(message.Payload)` throughout makes the panic disappear while still emitting a header whose `PayloadLength` claims the uncompressed size — which mis-frames every message boundary in the batch. Silent wire corruption is strictly worse than the panic.

foreign/go/client/tcp/tcp_group_polling.go
```diff
@@ -0,0 +1,228 @@
...
+			if !isAssignmentStale(err) {
+				return nil, err
+			}
+			c.groups.drop(key)
+			continue
+		}
+
+		// The server marks a stale assignment by answering an empty batch on
+		// the resync sentinel partition.
+		if polled.PartitionId == iggcon.ResyncRequiredPartition && polled.MessageCount == 0 {
```
numinnex: **warning — the resync sentinel gates on the server-declared count, which can exceed the decoded messages**

`polled.MessageCount` comes straight from the count the server declared (`binary_serialization/binary_response_deserializer.go:193`), while the message loop in that deserializer `break`s silently on a short body (lines 148-151 and 158-160) and still returns success. So `len(Messages) < MessageCount` is a reachable state, and a truncated sentinel reply passes this gate with a nonzero `MessageCount` — the assignment is never re-synced.

The Rust reference gates on the decoded messages instead: `messages.is_empty()` (`core/common/src/traits/binary_impls/messages.rs:219-220`).

The same silent truncation is a problem on the ordinary poll path independently of this gate: a short reply returns fewer messages than declared with `err == nil`, and a consumer that commits `CurrentOffset` skips the missing ones. Related: line 164 of that deserializer slices by `header.UserHeaderLength` with no bounds check, so a truncated user-header tail panics rather than breaking.

Suggested fix: gate here on `len(polled.Messages) == 0`, and in the deserializer return an error rather than `break` when the declared count is not satisfied.

foreign/go/internal/vsr/protocol_parity_test.go
```diff
@@ -0,0 +1,521 @@
...
+
+	workDir, err := os.Getwd()
+	require.NoError(t, err)
+	root := filepath.Clean(filepath.Join(workDir, "..", "..", "..", ".."))
+
+	sources := make(map[string]string, len(rustSources))
+	for name, relative := range rustSources {
+		content, err := os.ReadFile(filepath.Join(root, relative))
+		if err != nil {
+			t.Skipf("protocol sources are not available at %s: %v", root, err)
```
numinnex: **warning — one missing Rust source silently skips the entire parity suite**

This `t.Skipf` fires when *any* of the seven Rust sources cannot be read, and the message blames the repo root. So moving or renaming a single file on the Rust side turns every parity assertion in this file into a skip, and CI stays green through exactly the refactor this test exists to catch.

Suggested fix: probe one sentinel path to answer "is this a full checkout", skip only on that, then `require.NoError` on each individual read.

While here, two coverage gaps worth closing, both places where the Go side already diverges or could diverge unnoticed:
- the eviction reason → `IggyError` mapping in `reply.go:126` (the fallback is currently drifted from `core/common/src/error/eviction.rs:67`)
- `NamespaceForRequest`'s routing decisions against `core/sdk/src/vsr.rs:345-361`

The rest of the file is solid — header offsets recomputed from the `repr(C)` text with a fill-the-header assert, operation discriminants by name, band and `is_metadata`/`is_result_framed` predicates derived from the Rust `matches!` arms, eviction discriminants, command codes, namespace bits and the protocol version all verify, and would catch drift.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
-		}
-		return finalizeWireRequest(out), nil
+// isRegisterCode reports whether the code carries the sign-in handshake.
+func isRegisterCode(code uint32) bool {
+	return code == uint32(command.LoginRegisterCode) ||
+		code == uint32(command.LoginRegisterWithPATCode)
+}
+
+// appendCommandFrame reserves the request prologue in buf and appends the
+// encoded command payload after it, so the frame is a single allocation the
```
numinnex: **warning — this comment is false for the one command that matters**

"keeps the frame a single allocation" holds only when the command implements `encoding.BinaryAppender`. Repo-wide, only `Identifier` and `PollMessages` (`internal/command/message.go:168`) do — `SendMessages` does not, so the hottest command takes the `MarshalBinary` fallback at lines 396-400: a full `make([]byte, totalSize)` (~1.08 MB for the 1000×1000 B batch the benchmark uses), then `append(buf, body...)` reallocates and copies the whole batch a second time. On top of that, `message.Header.ToBytes()` (`message.go:129`) allocates 64 B per message and discards it after the copy, so a 1000-message batch does ~1010 allocations and roughly double the necessary memory traffic.

Suggested fix: implement `AppendBinary` on `SendMessages` (append headers in place, no `ToBytes`) and keep `MarshalBinary` as `AppendBinary(nil)`. See the compression note on `tcp_messaging.go:54` — the `PayloadLength` index fix must land first or the rewrite converts that panic into silent wire corruption.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
 }
 
 // invalidateConnLocked closes the connection and marks it as disconnected
 func (c *IggyTcpClient) invalidateConnLocked() {
 	_ = c.closeConnLocked()
 	c.transportState = iggcon.TransportStateDisconnected
 	c.sessionState = iggcon.SessionStateUnauthenticated
+	c.session.Reset()
+	c.groups.clear()
+	c.topics.clear()
```
numinnex: **warning — clearing the topic cache on reconnect resets every producer's balanced cursor**

`topics.clear()` wipes both maps (`tcp_topic_cache.go:99-104`), including `balancedCursors`. After a reconnect, `nextBalanced` reads a missing key as 0, so the first post-failover batch from every producer in a fleet lands on partition 0 — a hot shard created at exactly the moment the cluster is already stressed. The balanced cursor is pure client-side fairness state with no dependency on cluster metadata, so dropping it buys nothing. The Rust equivalent (`ConsumerGroupClientState`) has no reset on reconnect at all.

`groups.clear()` on the other hand is correct and should stay: a new connection registers a new client identity, so nothing the server assigned to the old one still holds.

The `partitionsCounts` half is more subtle and should *not* simply be kept: `topicKey` is the encoded identifier (`tcp_topic_cache.go:33-43`), so a topic deleted and recreated during the disconnect maps to the same key, and nothing invalidates a count on a send rejection — only this client's own `CreatePartitions`/`DeletePartitions`/`DeleteTopic`. A recreated-larger topic would strand the new partitions forever, and a recreated-smaller one would fail every send.

Suggested fix: split `clear()` — keep `balancedCursors` across reconnect, keep dropping `partitionsCounts`, and additionally invalidate a count when a send returns the partition-not-found error class. A TTL on the count (it currently has none, so another client's partition change is never observed while the connection lives) then becomes a nice-to-have rather than a correctness requirement.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
+
+		switch {
+		case err == nil:
+			return response, nil
+		case errors.Is(err, ierror.ErrTransientNotAccepted) &&
+			!isRegisterCode(code) && time.Now().Before(deadline):
+			// The server never admitted the request, so re-issuing it cannot
+			// double-apply. A same-connection replay keeps the stamped request
+			// id; a redirect registers again, so the frame is stamped afresh.
+			redirect, redirectErr := c.HandleLeaderRedirection(ctx)
```
numinnex, line 612: **warning — this puts a blocking, context-free DNS resolution on the request path**

`HandleLeaderRedirection` now runs on every `ErrTransientNotAccepted` round (so roughly every `failoverCheckInterval`, 2 s, during a transient storm). It reaches `util.CheckAndRedirectToLeader` → `processClusterMetadata` → `isSameAddress` → `parseAddress`, which calls `net.ResolveTCPAddr` up to twice per comparison (`internal/util/leader_aware.go:166`, `:170`).

`net.ResolveTCPAddr` takes no context. IP literals hit Go's fast path, but `currentServerAddress` can be a hostname straight from `config.serverAddress`, and then a slow or unreachable resolver blocks inside the 30 s request budget with neither the caller's context nor that budget able to interrupt it. That makes it a liveness issue rather than only a cost.

The comparison logic is pre-existing; this change is what moved it onto the hot failover path.

Suggested fix: compare host:port literally (the addresses come from the same cluster-metadata roster on both sides), or resolve once at `Connect` and cache the result.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -293,16 +339,15 @@ func (c *IggyTcpClient) write(payload []byte) (int, error) {
 // the appender interface encode directly into a pooled buffer.
 func (c *IggyTcpClient) do(ctx context.Context, cmd command.Command) ([]byte, error) {
 	bp := acquireRequestBuf()
-	buf, err := encodeWireRequest(*bp, cmd)
+	defer releaseRequestBuf(bp)
+
+	frame, err := appendCommandFrame(*bp, cmd)
```
numinnex, line 377: **warning — the buffer pool no longer serves the send path, and lost its tests in the same commit**

Three issues in the pooling code around this call, none currently covered by a test:

1. `releaseRequestBuf`'s `maxPooled = 64 * 1024` (line 288) means any frame above 64 KiB is never returned to the pool. Every real producer batch therefore takes a cap-256 buffer from the pool, grows it to MB scale, and throws it away — the pool only ever serves small metadata commands.
2. The pool's `New` returns `cap 256` (line 278), exactly `vsr.HeaderSize`, so the very first payload byte forces a reallocation on every freshly minted buffer. This was harmless under the old 8-byte prologue; it is not under a 256-byte one.
3. On this line's error path, `*bp` is never updated before returning, so a buffer that already grew goes back to the pool at its old capacity.

The reason all three went unnoticed is that `tcp_core_encode_test.go` was deleted in this commit, taking `TestRequestBufPool_AcquireReleaseRoundTrip`, `TestRequestBufPool_OversizedBufferDropped`, `TestEncodeWireRequest_FallbackPropagatesMarshalError` and `TestEncodeWireRequest_GrowsUndersizedBuffer` with it. Nothing remaining references `appendCommandFrame`, `acquireRequestBuf` or `releaseRequestBuf`, so the 64 KiB drop path and the `MarshalBinary` fallback are exercised only incidentally through `do` — in the same change that grew the prologue from 8 to 256 bytes.

Suggested fix: raise or remove the ceiling (or add a second size class for batch frames), start the pool at `HeaderSize + 512`, assign `*bp` before the error return, and port the four deleted tests forward.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -334,153 +379,335 @@ func isSessionControlCode(code uint32) bool {
...
+func isRegisterCode(code uint32) bool {
+	return code == uint32(command.LoginRegisterCode) ||
+		code == uint32(command.LoginRegisterWithPATCode)
+}
+
+// appendCommandFrame reserves the request prologue in buf and appends the
+// encoded command payload after it, so the frame is a single allocation the
+// header can be stamped into once the payload length is known.
+func appendCommandFrame(buf []byte, cmd command.Command) ([]byte, error) {
+	buf = append(buf[:0], requestPrologue[:]...)
```
numinnex: **warning — the 256-byte prologue is zeroed twice per request**

This appends 256 zero bytes, and then `EncodeRequestHeader` does `clear(dst[:])` over the same range (`internal/vsr/header.go:137`). That is 512 bytes written and 256 read per request, four cache lines of pure waste, on every RPC. Same double-zero at line 363 and at `tcp_session_management.go:63`.

Keep the `clear` — it is the documented guard (`header.go:135-136`) against pooled bytes leaking into the reserved header regions, and it must not be removed. Drop the zero-append instead:

```go
buf = slices.Grow(buf[:0], vsr.HeaderSize)[:vsr.HeaderSize]
```

and let `EncodeRequestHeader` own the zeroing.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -271,6 +311,9 @@ func (c *IggyTcpClient) readInto(buf []byte) (int, error) {
 		if err != nil {
 			return totalRead, err
 		}
+		if n == 0 {
```
numinnex, line 347: **warning — reply bodies are allocated per reply and aliased by the poll decoder**

`read` (lines 296-302) allocates `make([]byte, expectedSize)` for every reply body and never pools it, and `DeserializeFetchMessagesResponse` hands out subslices of that buffer (`binary_serialization/binary_response_deserializer.go:161`, `:166`). So a single retained `IggyMessage` pins the entire MB-scale reply buffer for as long as the caller holds it — a retention hazard, not just allocation churn.

Separately, every reply costs at least two read syscalls (256-byte header, then body) with no `bufio.Reader` in front of the connection.

Both predate this change, but the poll path is the one this PR now routes group polling through repeatedly.

Suggested fix: put a buffered reader in front of the header read, and either pool the body buffer with an explicit copy-out for retained payloads, or document the aliasing on the public poll API.

Related, same file: `binary_response_deserializer.go:144-146` reads `messagesCount` and then allocates `make([]iggcon.IggyMessage, 0)` with no capacity, so a 1000-message poll does roughly 11 grow reallocations and ~200 KB of element copying. Worth fixing — but cap the capacity hint by `len(payload)/MessageHeaderSize` rather than trusting the declared count, which is server-controlled.

foreign/go/client/tcp/tcp_core.go
```diff
@@ -556,8 +785,10 @@ func (c *IggyTcpClient) Connect(ctx context.Context) error {
 		}),
 	).Do(
 		func() error {
-			c.logger.Info("Iggy client is connecting to server...", slog.String("server_address", c.currentServerAddress))
-			connection, err := (&net.Dialer{}).DialContext(ctx, "tcp", c.currentServerAddress)
+			address := candidates[candidateIndex%len(candidates)]
+			candidateIndex++
+			c.logger.Info("Iggy client is connecting to server...", slog.String("server_address", address))
+			connection, err := (&net.Dialer{}).DialContext(ctx, "tcp", address)
```
numinnex, line 927: **warning — Nagle is enabled on a lockstep client, with no option to turn it off**

Line 798 calls `tc.SetNoDelay(c.config.noDelay)`, and `noDelay` defaults to `false` (line 101) with no `WithNoDelay` option anywhere in the module. So the client explicitly *enables* Nagle's algorithm — Go's own default for a `TCPConn` is `NoDelay` on, i.e. Nagle off — on a strictly serial request/response client where a multi-segment send frame stalls its trailing partial segment waiting for an ACK.

Pre-existing (master has the identical `noDelay bool` / `noDelay: false` / `SetNoDelay(c.config.noDelay)` at lines 79, 89 and 567), so not introduced here, but the VSR framing makes frames larger and the lockstep exchange makes the stall directly visible in request latency.

Suggested fix: default `noDelay: true` and expose `WithNoDelay` so an operator can choose.

foreign/go/internal/vsr/session.go
```diff
@@ -0,0 +1,168 @@
...
+	if s.requestCounter == math.MaxUint64 {
+		return 0, ErrRequestCounterExhausted
+	}
+	id := s.requestCounter
+	s.requestCounter++
+	return id, nil
+}
+
+// CurrentRequestID returns the watermark without advancing it. Non-replicated
+// and partition-plane requests use it: the server's client table tracks ids
```
numinnex: **warning — the stated reason for not advancing the counter is factually inverted**

The comment says consuming a request id here "would gap the next metadata request and the primary would drop it". The server does the opposite: `core/consensus/src/client_table.rs:613` returns `RequestStatus::New` for *any* `request > watermark`, and that is pinned by `check_request_jump_above_watermark_is_new` (`client_table.rs:1628`), whose own comment reads "No contiguity requirement".

The behaviour is correct; only the justification is wrong. The real reason `CurrentRequestID` is safe is that the partition plane never consults the client table at all (`core/consensus/src/metadata_helpers.rs:18-20`). That distinction matters for a future reader: someone "fixing the gap" on the basis of this comment would break routing, and the actual design has a consequence worth naming — every partition request on a session carries the identical id, which is also the id the next metadata operation will claim.

The same wrong rationale appears at `envelope.go:74-78`, in `client/tcp/tcp_core.go:614-616` (which claims the client table answers a partition-plane replay from its reply cache — it does not, so that replay is at-least-once and the comment is what makes it look safe), and in `core/sdk/src/vsr.rs:96-104`.

Suggested fix: replace all four with the real reason and name the invariant it depends on.

foreign/go/client/tcp/tcp_session_management.go
```diff
@@ -22,58 +22,95 @@ import (
...
+		slog.String("server_version", registered.ServerVersion))
+	return &iggcon.IdentityInfo{UserId: registered.UserID}, nil
+}
+
+// endBoundSession logs out a live session before a re-login, so the server
+// drops its client-table entry instead of leaving it to be fenced.
+func (c *IggyTcpClient) endBoundSession(ctx context.Context) error {
+	c.mtx.Lock()
+	bound := c.session.Bound()
+	c.mtx.Unlock()
```
numinnex, line 111: **warning — this makes every re-login traverse a server path that can answer with no frame at all**

`endBoundSession` now sends a `Logout` before each re-login. On the server, `handle_logout_request` (`core/server-ng/src/dispatch.rs:2645-2660`) returns *silently* on two paths — an unbound session, and a `submit_logout_on_owner` error — sending no reply frame.

This client is lockstep, so `readReplyLocked` blocks on the socket for the full 30 s `responseReadTimeout`, and because `attempt` holds `c.mtx` for the whole exchange, every other caller blocks with it. Reachable on an ordinary re-login against a non-primary.

The same server file explicitly rejects this pattern elsewhere (`dispatch.rs:919-921`: "A silent drop would wedge every later request on the connection until the socket read timeout").

This is a server-side defect rather than an SDK one, but this line is what puts it on a routine path, so the two should be fixed together: `handle_logout_request` should answer denied or transient instead of returning.

foreign/go/contracts/client.go
```diff
@@ -93,13 +93,18 @@ type Client interface {
...
+	// under a fresh client identity, so the server cannot deduplicate it
+	// against the original and the batch may commit twice.
 	SendMessages(
 		ctx context.Context,
 		streamId Identifier,
 		topicId Identifier,
 		partitioning Partitioning,
 		messages []IggyMessage,
-	) error
+	) (*SendMessagesResponse, error)
```
numinnex, line 109: **warning — `PollMessages` gained three caller-visible outcomes, none documented**

The doc comment is unchanged, but the behaviour is not. A group poll that names no partition is now orchestrated client-side (`client/tcp/tcp_group_polling.go`) and can end in three different ways:

1. `err == nil` with `PartitionId == NoAssignedPartition` (0xFFFFFFFE) and an empty batch, when the member owns no partition (`tcp_group_polling.go:150-153`).
2. `err == ErrConsumerGroupPartitionNotOwned` (5009) when the attempt budget is exhausted (`:179`) — an error no other SDK raises for this case.
3. Real messages at a real partition id.

Only the sentinel constant is explained anywhere, and only in `contracts/consumer_group_assignment.go:20-28`.

Suggested contract to document here, which also resolves the 5009 divergence:
- `ErrConsumerGroupMemberNotFound` → the caller must rejoin. This is what keeps "broken" distinguishable from "idle".
- `err == nil` with `NoAssignedPartition` → nothing to read right now, poll again. Both the zero-partition case and the exhausted-attempts case should return this shape.
- anything else → real data.

A caller cannot act on "rebalance storm" differently from "idle" in any case, because the SDK has already re-synced the assignment before returning; the only remedy in both is to poll again. Rust says exactly this at `core/common/src/traits/binary_impls/messages.rs:248-253`, and Node returns the empty shape from both paths (`foreign/node/src/wire/message/poll-messages.command.ts:204-209`, `:234-239`).

foreign/go/client/tcp/tcp_group_polling.go
```diff
@@ -0,0 +1,228 @@
...
+			c.groups.drop(key)
+			continue
+		}
+		return polled, nil
+	}
+
+	return nil, ierror.ErrConsumerGroupPartitionNotOwned
```
numinnex: **warning — a hard 5009 here makes a normal rebalance look like a failure**

Exhausting `groupPollAttempts` returns `ErrConsumerGroupPartitionNotOwned`. Both other SDKs return an empty poll for exactly this case, and the Rust reference spells out why at `core/common/src/traits/binary_impls/messages.rs:248-253`: a normal rebalance must not look like a hard error to a consumer-group application that does not special-case 5009. Node does the same (`foreign/node/src/wire/message/poll-messages.command.ts:234-239`).

As it stands, Go consumers get an error on a benign back-to-back rebalance that no other SDK raises.

Suggested fix: return `&iggcon.PolledMessage{PartitionId: iggcon.NoAssignedPartition, Messages: []iggcon.IggyMessage{}}, nil`, matching line 150-153, and document the three outcomes on the `Client` interface. Worth landing after the cursor fixes above, so that an empty return signals a rebalance rather than masking a starved cursor.

foreign/go/contracts/client.go
```diff
@@ -194,6 +199,17 @@ type Client interface {
 		groupId Identifier,
 	) error
 
+	// SyncConsumerGroup fetch the partitions this client currently owns in a consumer group,
+	// together with the group generation, for the given stream and topic by unique IDs or names.
+	// Returns nil when the client is not a member of the group.
+	// Authentication is required, and the permission to read the streams or topics.
```
numinnex, line 223: **warning — `(nil, nil)` for a non-member is a footgun and diverges from every sibling getter**

Returning a nil assignment with a nil error is unique on this interface. Every other getter maps an empty body to a typed error: `tcp_topic_management.go:43-45` → `ErrTopicIdNotFound`, `tcp_stream_management.go:45-47` → `ErrStreamIdNotFound`, `tcp_consumer_group_management.go:52-54` → `ErrConsumerGroupIdNotFound`. `ErrConsumerGroupMemberNotFound` (5006) already exists in `errors_gen.go` and is what the Rust SDK surfaces here.

Any external caller that follows the interface's own convention will nil-deref on the one method that breaks it.

Suggested fix: return `ErrConsumerGroupMemberNotFound` for a non-member, and keep a member holding zero partitions as a non-nil assignment with an empty `Partitions` slice — the two states need opposite handling and only membership distinguishes them.

Coupling note: the `(nil, nil)` originates at `binary_serialization/binary_response_deserializer.go:670-672`, and `client/tcp/tcp_group_polling.go:205-215` currently depends on nil meaning "not a member" to drive its implicit join, so the typed error has to land together with a rework of that branch.

foreign/go/client/tcp/tcp_topic_management.go
```diff
@@ -118,5 +118,8 @@ func (c *IggyTcpClient) UpdateTopic(
 
 func (c *IggyTcpClient) DeleteTopic(ctx context.Context, streamId, topicId iggcon.Identifier) error {
 	_, err := c.do(ctx, &command.DeleteTopic{StreamId: streamId, TopicId: topicId})
-	return err
+	if err != nil {
+		return err
+	}
+	return c.dropTopicCache(streamId, topicId)
```
numinnex: **warning — `DeleteStream` is missing the equivalent of this cache drop**

`DeleteTopic` correctly drops the topic cache here, and `CreatePartitions`/`DeletePartitions` invalidate the count (`tcp_partition_management.go:33-36`, `:45-48`). `DeleteStream` (`tcp_stream_management.go:81-84`) does neither, so deleting and recreating a stream leaves `Balanced` and `MessageKey` sends resolving against the dead topic's partition count.

Suggested fix: sweep every `topicKey` whose stream half matches, since `topicKey` is `(stream, topic)` and a stream delete invalidates every topic under it.

Minor, same neighbourhood: `invalidateTopicCache`/`dropTopicCache` (`tcp_partition_management.go:53-69`) return their `newTopicKey` error, so `CreatePartitions`, `DeletePartitions` and `DeleteTopic` can report failure *after* the server already applied the change. It is unreachable — `Identifier.AppendBinary` (`contracts/identifier.go:117-119`) never returns an error, and the same identifiers already marshalled successfully inside `c.do` — which is the argument for dropping the error from `newTopicKey` rather than propagating it on a post-success path.

foreign/go/client/tcp/tcp_connect_test.go
```diff
@@ -0,0 +1,469 @@
...
+	assert.Equal(t, vsr.OperationLogout, recorded[2].operation(),
+		"a re-login logs the live session out first")
+	assert.Equal(t, vsr.OperationRegister, recorded[3].operation())
+	assert.NotEqual(t, first, recorded[3].clientID(),
+		"the re-login presents a new client identity")
+	assert.True(t, client.session.Bound())
+}
+
+func TestConnect_ExchangesOverTLS(t *testing.T) {
+	certificate, err := tls.LoadX509KeyPair(certPath("iggy_cert.pem"), certPath("iggy_key.pem"))
```
numinnex: **warning — this pulls a dated filesystem dependency into the unit-test lane**

`certPath` reaches `core/certs/iggy_cert.pem`, whose `notAfter` is `Nov 17 09:21:08 2026 GMT`, and certificate validation is on by default. The unit lane runs `go test $(go list ./... | grep -v /tests)` (`.github/actions/go/pre-merge/action.yml:73`), which includes `client/tcp` — so unlike the deleted `foreign/go/tests/tls_test.go`, this test is no longer excluded by that `grep -v /tests`.

Two consequences: the unit lane starts failing on cert expiry, and it cannot run against a `foreign/go`-only checkout that has no `core/certs`.

Suggested fix: generate a throwaway self-signed cert in-test, or skip when the cert file is absent.

numinnex (dismissed):

numinnex (dismissed):

hubcio (dismissed):

hubcio (approved):

mmodzelewski (approved):

PR #3836, opened 2026-08-07
refactor(consensus): remove namespace from client headers

numinnex:
Removes the client-supplied `namespace` word from the VSR client request and reply headers. The server derives the routing target itself: the plane from `operation`, the partition group from the request body. At the admission boundary a request is promoted to the server-internal `RoutedRequestHeader`, which carries the resolved consensus group in the tail of the client header's reserved area (bytes 248..256).

This is a behavior change, not only a refactor. Name-addressed sends previously shipped headers pointing at partition `(0, 0, 0)`, and the Rust default producer config could not send over VSR at all. With the namespace derived server-side, balanced and message-key partitioning now work over VSR for the Rust and Node SDKs.

Review follow-ups included here:

- `RoutedRequestHeader::validate` now enforces the same field rules as the client boundary (shared `validate_request_fields`), so a peer-wire `Command2::Request` frame cannot carry `client = 0` into the client table's hard assert or replay a cached register reply via `operation = Reserved`.
- The client-to-routed promotion is an in-place retype (`Message::into_routed`) that zeroes only the eight `group` bytes instead of rebuilding the 256-byte header; the terminal validate stays.
- Cached client-table replies persist as raw `ReplyHeader` bytes, and this PR moves `status` from 224 to 216. Both persistence surfaces are versioned: the state-transfer artifact magic is bumped to `ICT2`, and `MetadataSnapshot::decode` refuses snapshots whose `version` differs from the current format version (bumped to 2).
- Layout pins: `ReplyHeader::status` at offset 216, `RoutedRequestHeader::group` at 248..256, and the data-bearing reserved prefix surviving promotion.


mmodzelewski (dismissed):

spetz (dismissed):

mmodzelewski (dismissed):

spetz (dismissed):

core/server_common/src/consensus_message.rs
```diff
@@ -606,7 +606,9 @@ where
 
         match command {
             Command2::Prepare => Ok(Self::Prepare(value.try_into_typed::<PrepareHeader>()?)),
-            Command2::Request => Ok(Self::Request(value.try_into_typed::<RequestHeader>()?)),
+            Command2::Request => Ok(Self::Request(
```
hubcio, line 637: this decodes every peer-wire `Command2::Request` as `RoutedRequestHeader`, whose `validate()` only checks the command byte - the old `RequestHeader` decode here rejected `client == 0`. a peer can now forge a request with `client = 0` and a metadata op that reaches `check_request` and trips the hard `assert!(client_id != 0)` in `client_table.rs`, aborting the metadata primary in a release build. replica auth is off by default, so any host that can reach the replica port does it, and the elected successor is equally killable. the same weakened validate also lets `operation == Reserved` through, which replays a client's cached register reply. fix: give `RoutedRequestHeader::validate` the same field rules as `RequestHeader::validate` (share one `validate_request_fields`), or reject `Command2::Request` in `MessageBag::try_from` since nothing puts one on the replica wire.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -304,9 +303,88 @@ const _: () = {
...
+impl Default for RoutedRequestHeader {
+    fn default() -> Self {
+        Self {
+            checksum: 0,
+            checksum_body: 0,
+            cluster: 0,
+            size: 0,
+            view: 0,
+            release: 0,
+            command: Command2::Request,
```
hubcio: every other header's `Default` sets `command: Command2::Reserved` so a half-filled header fails `validate()`; this one sets `Request`, so a fully-defaulted `RoutedRequestHeader` passes validate carrying `client = 0` and `operation = Reserved`. set it to `Reserved` too. the only builder that omits `command` (the prepare template near line 2314) never validates, so it's safe. note this is hygiene, not the security fix - the peer path reinterprets wire bytes and never goes through `Default`.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -325,11 +403,84 @@ impl Default for RequestHeader {
...
+}
+
+impl RoutedRequestHeader {
+    /// Promote a client-wire [`RequestHeader`] to the server-internal routed
+    /// shape, stamping the resolved consensus `group`. Every shared field is
+    /// copied verbatim; this is the ONLY sanctioned crossing between the two
+    /// layouts (transmute-based reads across them would alias `group` with
+    /// reserved bytes).
+    #[must_use]
+    pub const fn from_request(header: &RequestHeader, group: u64) -> Self {
```
hubcio: `from_request` is the only sanctioned crossing between the two 256-byte layouts and has no test - nothing in the test module names it or `RoutedRequestHeader`. add one pinning that `reserved[0..52]` survives and `group` lands at bytes 248..256, so a later reshuffle of the reserved carve can't silently move the non-replicated op code that lives in `reserved[0..4]`.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -304,9 +303,88 @@ const _: () = {
...
+/// A [`RequestHeader`] AFTER the receiving node resolved its target.
+///
+/// The server-internal shape a client request travels in between shards
+/// (and follower to primary), never on the client wire. `group` carries the
+/// resolved consensus group so the owner shard can route, park, and fence
+/// WITHOUT re-decoding the payload -- clients no longer send any namespace
+/// (it is derived: plane from `operation`, partition group from the body),
+/// so this is where the derivation result lives for the internal hop.
+///
+/// Layout: identical to [`RequestHeader`] with `group` claiming the first
```
hubcio: this says `group` claims the first eight reserved bytes, but the field doc just below and the actual layout put it at the tail (248..256, the last 8). an SDK author porting from this doc places `group` at 196 and shifts every field after it. point the struct doc at the tail.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -2389,11 +2539,7 @@ mod tests {
         use std::mem::offset_of;
         assert_eq!(size_of::<ReplyHeader>(), HEADER_SIZE);
         assert_eq!(
-            offset_of!(ReplyHeader, status),
-            offset_of!(ReplyHeader, namespace) + size_of::<u64>()
-        );
-        assert_eq!(
-            offset_of!(ReplyHeader, reserved) + size_of::<[u8; 28]>(),
+            offset_of!(ReplyHeader, reserved) + size_of::<[u8; 36]>(),
```
hubcio, line 2519: this test lost its status-offset assertion. it used to pin `status` relative to `namespace`; with `namespace` gone that assert was removed instead of re-anchored, so both remaining lines just restate the compile-time `const _` block and a test named `..._status_offset_..._pinned` pins no status offset. the reply funnel reads `status` at 216 and four SDKs hardcode it. add `assert_eq!(offset_of!(ReplyHeader, status), 216);`.

core/metadata/src/impls/metadata.rs
```diff
@@ -1893,7 +1896,7 @@ where
         }
 
         let request = build_register_request_message(consensus, client_id, user_id);
-        // Wire path runs `RequestHeader::validate` at network boundary;
+        // Wire path runs `RoutedRequestHeader::validate` at network boundary;
         // in-process skips it. debug_assert pins drift.
         debug_assert!(
```
hubcio, line 1901: this debug_assert used to prove the register builder emits a header the wire path would accept, but `request` is now `Message<RoutedRequestHeader>` and its `validate()` only checks the command byte, so it no longer pins `client != 0` / `session == 0` / `request == 0`. the comment above was updated to name `RoutedRequestHeader::validate`, which is exactly the validate that lost the rules. the logout builder assert has the same downgrade. restoring the field rules on `RoutedRequestHeader::validate` makes both real again.

core/server-ng/src/dispatch.rs
```diff
@@ -861,6 +863,13 @@ async fn handle_client_request<B, MJ, S, SB>(
             return;
         }
     };
+    // Promote to the server-internal routed shape at the boundary: the
+    // client wire carries no group (it is derived -- plane from `operation`,
+    // partition target from the payload), so it starts unset here and the
+    // resolution sites below stamp it before anything routes on it.
+    let request = request.transmute_header(|header, new_header: &mut RoutedRequestHeader| {
```
hubcio: this rebuilds the whole 256-byte header on every client request, including ping and poll, to stamp `group = 0` - and nothing reads `group` before the metadata and partition paths overwrite it a few lines down. an in-place retype that zeroes only the 8 group bytes is byte-for-byte equivalent and lets `from_request` go away. best done after the validate fix, and keep the terminal `validate` inside the retype.

core/binary_protocol/src/version.rs
```diff
@@ -63,7 +63,7 @@
 //! `ClientVersionInfo` is the leading bytes of the login-register request
 //! *body*, which itself rides inside a 256-byte VSR `RequestHeader` (see
 //! `consensus::header`): `command` = `Command2::Request`, `operation` =
-//! `Operation::Register`, `namespace` = `METADATA_CONSENSUS_NAMESPACE`,
+//! `Operation::Register`, `namespace` = `METADATA_GROUP`,
```
hubcio: this still tells foreign SDK authors the login header carries `namespace = METADATA_GROUP`, but the client header no longer has that field. an author following it writes `1<<63` into what is now `session` and the register is rejected. drop the `namespace` clause - the client sends no group.

hubcio (changes requested):
- cached client-table replies persist as raw `ReplyHeader` bytes in the checkpoint and the state-transfer artifact, and this PR moved `status` 224 -> 216. correct today only by coincidence: every cached reply is metadata-plane so the old `namespace` word was `1<<63` with a zero low half, which the relocated `status` reads as 0. the state-transfer artifact has a magic tag but the msgpack snapshot reply bytes have no version marker at all. version both before this can bite.

- `IggyNamespace::new` masks (`& STREAM_MASK`) instead of rejecting and nothing caps live stream/topic count, so 4097 concurrently-live streams gives slab key 4096 which masks to stream 0 - a send to it lands in stream 0's partition and journals there. pre-existing (the resolve path is unchanged and string ids always masked server-side) so not a blocker here, but it's a silent cross-tenant write and this removes the last client-side fail-fast. worth its own issue: a fallible `checked_new` at the few sites that pack a runtime-derived key.

- worth noting in the description: this also enables balanced and message-key partitioning over VSR for the rust and node SDKs (the rust default producer config couldn't send over VSR before), so it's a behavior change, not only a refactor.

the core change is good - the deleted SDK packers re-derived a value the server computed anyway, and name-addressed sends were shipping headers pointed at partition (0,0,0). the gap is that the client boundary got hardened while the peer boundary stayed wire-trusting.

hubcio (approved):

spetz (approved):

PR #3837, opened 2026-08-07
feat(ci): require bash >= 4.2 in the pre-commit hook scripts

ethanlin01x:

Related: #3694 

## Rationale

Without a declared minimum, every hook script has to be written against bash 3.2 and carry portable fallbacks (#3694). Checking the version once, up front, means anyone writing a hook script from now on can use modern bash without thinking about it. 

## What changed?

1. `scripts/ci/lib/init.sh` checks for bash >= 4.2 and is sourced by the 13 script-backed hooks, so a future check reaches all of them from one place. 
2. The 16 scripts on `#!/bin/bash` are normalized to `#!/usr/bin/env bash` first, since on macOS the former is always the system bash 3.2 and no `brew install bash` can override it.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? Claude Code
2. Scope of usage?  Implementation
3. How did you verify the generated code works correctly? Ran the hooks under both bash 3.2 and bash 5,


jiengup (approved):
LGTM

hubcio (approved):

hubcio (approved):

mmodzelewski (approved):

spetz (approved):

PR #3841, opened 2026-08-07
feat(java): migrate the TCP client to the VSR wire protocol

mmodzelewski:
The Java SDK spoke only the legacy TCP framing, so it could not
talk to server-ng, which frames every request in the VSR
(Viewstamped Replication) consensus protocol.

The Netty pipeline now encodes requests with VSR headers from a
per-connection consensus session (client id, request counter,
fence epoch) and decodes VSR replies, replacing the legacy frame
codec. The server binds the session to one transport connection,
so the pool pins a single channel and the connectionPoolSize
builder option is gone. Transient denials replay the same encoded
frame so the server dedup sees one request id: not-committed for
the full 30s budget (outcome unknown), not-accepted for 2s only
(refused outright, typically a demoted primary). A failed pool
acquire starts a background redial that alternates the current
node with the seed and replays the builder credentials.

The VSR broker routes explicit partitions only, so balanced and
key-based sends resolve to a concrete partition client-side
(round-robin cursor, xxh32(key) % count) and group polls pick the
partition from the member's assignment, synced through the new
consumer-group SYNC command and fenced by generation, matching
the Rust SDK. Cached routing facts refresh past a 5s staleness
budget. sendMessages returns the commit confirmation (partition
and base offset) instead of Void.

Java CI, BDD, and examples lanes run against iggy-server-ng with
the vsr feature; the SDK version bumps to 0.9.0-SNAPSHOT. On the
server side, a login whose credentials were rejected no longer
falls through to the malformed-payload rejection but reports
InvalidCredentials.


foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncTcpConnection.java
```diff
@@ -193,100 +234,225 @@ public CompletableFuture<ByteBuf> send(CommandCode commandCode, ByteBuf payload)
...
+            long deadlineNanos = nowNanos + TRANSIENT_RETRY_BUDGET.toNanos();
+            long notAcceptedDeadlineNanos =
+                    isLoginCode(commandCode) ? deadlineNanos : nowNanos + NOT_ACCEPTED_RETRY_BUDGET.toNanos();
+            writeVsrFrame(channel, handler, frame, responseFuture, deadlineNanos, notAcceptedDeadlineNanos);
         } catch (RuntimeException e) {
             responseFuture.completeExceptionally(e);
         } finally {
             payload.release();
         }
     }
```
slbotbm, line 668: When a former leader remains reachable after an election, it can reject writes with `TransientNotAccepted`. `AsyncTcpConnection` retries the same frame on the same channel until `notAcceptedDeadlineNanos` and then completes the request exceptionally. `AsyncIggyTcpClient.onConnectionFailure()` only handles connection exceptions from failed pool acquisition, so this server response never triggers leader discovery and a healthy but demoted node is never replaced. Every later metadata write repeats the same failure until the application explicitly logs in again. Consider handling this status like the Rust VSR transport by fetching the current leader, reconnecting, re-registering, and reissuing the request; reissuing it elsewhere is safe because the server guarantees that it was not admitted.

Also applies to `AsyncIggyTcpClient.java:448-462`

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
slbotbm, line 1: The migrated client never starts a heartbeat task. With server-ng's `[heartbeat] enabled = true`, only `PING` updates a connection's heartbeat timestamp; ordinary polls and writes do not. The verifier evicts a consumer-group member after roughly `1.2 * heartbeat.interval`, so even an actively polling Java consumer loses its VSR session and group membership after a few seconds. Consider starting a periodic ping for the lifetime of each published connection, using a configurable interval, and canceling or restarting it on close and retarget.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncTcpConnection.java
```diff
@@ -108,16 +146,19 @@ public AsyncTcpConnection(
...
+                new PoolChannelHandler(host, port, enableTls, sslContext, consensusSession, this::onSessionEvicted),
                 ChannelHealthChecker.ACTIVE,
                 FixedChannelPool.AcquireTimeoutAction.FAIL,
                 poolConfig.getAcquireTimeoutMillis(),
-                poolConfig.getMaxConnections(),
+                1,
                 poolConfig.getMaxPendingAcquires());
 
-        log.info("Connection pool initialized with max connections: {}", poolConfig.getMaxConnections());
+        log.info("Connection pool initialized with a single VSR-pinned connection");
```
slbotbm, line 190: This changes the pool from five configurable connections to exactly one while retaining `AcquireTimeoutAction.FAIL`. The same class defines the default acquire timeout as three seconds at line 595 and the transient retry budget as 30 seconds at line 88. Because the channel is held until the response finishes, one slow or transiently retried operation makes every concurrent request fail acquisition after three seconds. This is a direct regression for the async client's concurrent callers and would also starve any heartbeat task. Consider preserving session pinning without a short pending-acquire deadline, multiplexing requests on the pinned channel, or using a separate VSR session per pool channel.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrResponseHandler.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        ByteBuf body;
+        try {
+            body = decodeReply(msg);
+        } catch (RuntimeException error) {
+            future.completeExceptionally(error);
+            return;
+        }
+        future.complete(body);
+    }
```
slbotbm, line 173: The handler polls `responseQueue` and returns before inspecting the consensus command. Evictions are session-level, one-way frames and can arrive while the connection is idle, notably from the stale-client verifier. Such a frame is currently discarded, so the session remains bound, the authentication generation is not advanced, and cached consumer-group assignments survive after the server has removed them. Consider inspecting and applying eviction effects before requiring a request future and failing any pending requests after resetting the session.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrNamespace.java
```diff
@@ -0,0 +1,153 @@
...
+        if (operation == VsrOperation.NON_REPLICATED || VsrOperation.isMetadata(operation)) {
+            return 0;
+        }
+        return switch (commandCode) {
+            case SEND_MESSAGES_CODE -> fromSendMessages(payload);
+            case STORE_CONSUMER_OFFSET_CODE,
+                    DELETE_CONSUMER_OFFSET_CODE,
+                    STORE_CONSUMER_OFFSET_2_CODE,
+                    DELETE_CONSUMER_OFFSET_2_CODE -> fromConsumerOffset(payload);
+            default -> throw IggyServerException.fromTcpResponse(VsrHeaders.ERROR_FEATURE_UNAVAILABLE, new byte[0]);
```
slbotbm, line 73: `VsrOperation.java:105` maps command 503 to `DELETE_SEGMENTS`, but that operation is deliberately neither metadata nor partition-classified. It therefore reaches this switch and falls into `FeatureUnavailable` for every request, so `sendBinaryRequest(503, ...)` can never reach server-ng. Consider adding the Rust SDK's equivalent case, which decodes the delete-segments request and packs its stream, topic, and partition into the namespace.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrFrameDecoder.java
```diff
@@ -0,0 +1,53 @@
...
+ * frame size (header included) sits at byte offset 48, followed by an
+ * optional body. There is no length delimiter outside the header.
+ */
+public class VsrFrameDecoder extends ByteToMessageDecoder {
+
+    /** Matches the server's default {@code max_message_size} (64 MB). */
+    static final long MAX_FRAME_SIZE = 64L * 1024 * 1024;
+
+    @Override
+    protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) {
```
slbotbm, line 54: The decoder treats the server's 64 MiB default as a protocol maximum, while `core/server-ng/config.toml:1000` exposes `message_bus.max_message_size` as a deployment setting rather than a fixed wire limit. A server configured for larger batches can therefore send a valid poll or metadata response that this client rejects as a desynchronized connection. The previous Java decoder had no such fixed ceiling. Consider making the client limit configurable and ensuring that it agrees with the target deployment instead of hard-coding the server default.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesDeserializer.java
```diff
@@ -177,13 +180,45 @@ public static ConsumerGroup readConsumerGroup(ByteBuf response) {
...
+    public static ConsumerGroupAssignment readConsumerGroupAssignment(ByteBuf response) {
+        // The generation is a monotonic rebalance counter compared only for
+        // equality, so reading the u64 as a signed long is safe.
+        var generation = response.readLongLE();
+        var partitionsCount = response.readUnsignedIntLE();
+        List<Long> partitions = new ArrayList<>(toInt(partitionsCount));
+        for (long i = 0; i < partitionsCount; i++) {
+            partitions.add(response.readUnsignedIntLE());
+        }
+        return new ConsumerGroupAssignment(generation, partitions);
```
slbotbm, line 199: Both the consumer-group assignment decoder here and the send-confirmation decoder at line 207 allocate an `ArrayList` directly from a wire-provided `u32` count before checking whether the body contains that many fixed-size entries. A short malformed response with a count near `Integer.MAX_VALUE` attempts a multi-gigabyte allocation and can terminate the JVM with `OutOfMemoryError`; `Math.toIntExact` only rejects values above the signed range. Consider bounding the count by `readableBytes() / 4` for assignments and `readableBytes() / 20` for confirmations before allocating the validated capacity.

slbotbm: ALso applies to BytesDeserializer.java:202-219

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/ClientRoutingState.java
```diff
@@ -0,0 +1,148 @@
...
+    private final Map<String, AtomicInteger> balancedCursors = new ConcurrentHashMap<>();
+    private final Map<String, GroupAssignment> assignments = new ConcurrentHashMap<>();
+
+    static String topicKey(StreamId streamId, TopicId topicId) {
+        return streamId + "|" + topicId;
+    }
+
+    static String groupKey(StreamId streamId, TopicId topicId, ConsumerId groupId) {
+        return streamId + "|" + topicId + "|" + groupId;
+    }
```
slbotbm: The `streamId + "|" + topicId` and `streamId + "|" + topicId + "|" + groupId` concatenations implicitly convert each identifier by calling the implementation at `identifier/Identifier.java:50-56`. That method returns only the name or numeric value and omits the identifier kind, while names can also contain the `|` delimiter. For example, numeric stream/topic IDs `1` and `2` produce the same key as named identifiers `"1"` and `"2"`, even though they can resolve to different resources; delimiter-containing names create further collisions. The cache can then reuse another topic's partition count or another group's assignment, causing sends to be routed to an invalid partition and polls to use an unrelated assignment. Consider keying these maps with records containing each identifier's kind and value rather than a formatted string.

slbotbm (dismissed):

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncTcpConnection.java
slbotbm: This was resolved.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesDeserializer.java
slbotbm: Resolved

mmodzelewski:
@slbotbm Thanks for the review. I've addressed all the comments.

/ready

numinnex (dismissed):

hubcio (dismissed):

numinnex (dismissed):

spetz (approved):

numinnex (approved):

PR #3846, opened 2026-08-08
fix(server-ng): stop redundant partition rebuilds resetting offsets

numinnex:


core/server-ng/src/partition_reconciler.rs
```diff
@@ -1595,61 +1627,122 @@ mod tests {
...
+            "the redundant build must be discarded, not adopted: a second insert \
+             overwrites the ns -> idx entry and orphans the first partition, \
+             leaking its VSR group and segment writers"
+        );
+        assert_eq!(
+            shard.shards_table().shard_for(ns),
+            Some(0),
+            "the discarded op must not repoint the routing row"
+        );
+        assert_eq!(
```
hubcio, line 1749: stays green even with all three `publish_current_offset()` calls deleted - every build here adopts at offset 0 and the hand `set_current_offset(3)` lands after the pass. nothing else covers the publish half (vsr-off CI spawns the legacy binary). fix: `partition.offset.store(N, Ordering::Release)` before enqueuing `InsertOwned`, then assert `stats.current_offset() == N` - must fail when the publish is removed.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -1595,61 +1627,122 @@ mod tests {
...
+
+        assert_eq!(
+            shard.plane.partitions().len(),
+            1,
+            "the redundant build must be discarded, not adopted: a second insert \
+             overwrites the ns -> idx entry and orphans the first partition, \
+             leaking its VSR group and segment writers"
+        );
+        assert_eq!(
+            shard.shards_table().shard_for(ns),
```
hubcio: can't fail - single shard, adopt would write `ShardId::new(0)` too. capture `epoch_before` after the first pass, stage the op with `epoch_before + 1`, assert `epoch_for(ns) == epoch_before`.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -599,18 +606,20 @@ async fn reconcile_additions(
...
-                counters.routed += 1;
-            }
+            stage_routing_row(ctx, ns, epoch, owning_shard, counters);
+            continue;
+        }
+
+        // An earlier pass already built this one and the pump has not applied it
+        // yet, so the `contains` test above reads false for finished work.
+        // Rebuilding is not a wasted-effort question: the second incarnation
+        // shares the namespace's `PartitionStats` with the live one and re-opens
```
hubcio: nothing is live here (`contains` was false) - the sibling is queued for adoption. same wording in the test doc at 1632; the copy at shard/lib.rs:1922 is fine.

core/shard/src/lib.rs
```diff
@@ -1882,21 +1910,26 @@ where
...
+                    //
+                    // A backstop, not the mechanism. `reconcile_additions`
+                    // skips a namespace whose `InsertOwned` is already staged
+                    // ([`Self::has_staged_insert_owned`]), so a second op for a
+                    // live namespace should not be built at all. Dropping one
+                    // here is damage control rather than a free no-op: the
+                    // build already planted its initial segment over the live
+                    // incarnation's path and folded that into the namespace's
+                    // shared stats.
                     if partitions.contains(&namespace) {
```
hubcio, line 1927: this drop is silent, but per the comment above it only fires after the build truncated the live writer's file - worth `tracing::error!` + a metric.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -1053,6 +1059,31 @@ fn fetch_partition_stats(
...
 }
 
+/// Point this shard's routing row for `ns` at its owner, so a frame that lands
+/// here reaches the shard holding the partition.
+///
+/// Compares the epoch, not just presence: a delete + recreate recycles the slab
+/// keys, so the row survives with the DEAD incarnation's `created_revision`. A
+/// presence-only gate never refreshes it, and nothing else writes a non-owner's
+/// row.
+fn stage_routing_row(
```
hubcio: single caller, wrapping `shards_table_has_epoch` (also single caller), while 593 spells the same predicate inline. inline this, keep `shards_table_has_epoch` and use it at 593.

core/server-ng/src/partition_reconciler.rs
```diff
@@ -1053,6 +1059,31 @@ fn fetch_partition_stats(
...
+    ctx: &ReconcilerCtx,
+    ns: IggyNamespace,
+    epoch: u64,
+    owning_shard: u16,
+    counters: &mut PassCounters,
+) {
+    if shards_table_has_epoch(ctx, ns, epoch) {
```
hubcio: no mirror of the `InsertOwned` guard, so a lagging pump gets one duplicate `InsertRouted` per namespace per pass and the fast-skip stays disarmed. harmless - just worth a doc line that the asymmetry is deliberate (a queue scan here would go quadratic).

core/server-ng/src/partition_reconciler.rs
```diff
@@ -599,18 +606,20 @@ async fn reconcile_additions(
...
+        // An earlier pass already built this one and the pump has not applied it
+        // yet, so the `contains` test above reads false for finished work.
+        // Rebuilding is not a wasted-effort question: the second incarnation
+        // shares the namespace's `PartitionStats` with the live one and re-opens
+        // segment 0 with `file_exists = false`, which truncates. It joins
+        // `staged` so the parked-frame sweep still reads the namespace as
+        // building instead of aging its frames.
+        if ctx.shard.has_staged_insert_owned(ns) {
+            counters.already_staged += 1;
+            staged.insert(ns);
```
hubcio: `staged` duplicates what `has_staged_insert_owned` answers and goes stale across the `reconcile_removals` awaits. drop the set + `staged_unapplied` param and query the queue per parked namespace; `parked` early-returns when empty.

core/partitions/src/iggy_partition.rs
```diff
@@ -770,7 +770,23 @@ where
...
+    ///
+    /// Call it when the instance BECOMES the addressable one, never while
+    /// building it. The stats registry keys on the namespace, not the
+    /// incarnation, so every build of a namespace holds the same `Arc` as
+    /// whatever is already serving it -- and a build is not guaranteed to be
+    /// adopted. Seeding from the build instead leaves a zeroed `current_offset`
+    /// on the live incarnation, which then rejects every
+    /// `store_consumer_offset` above 0 with `InvalidOffset` until the next send
+    /// re-seeds it.
+    pub fn publish_current_offset(&self) {
```
hubcio: all three call sites funnel through `IggyPartitions::insert` - moving the publish inside makes the contract unforgettable, guards the unit-unreachable boot site through the same test, and lets this go crate-private.

hubcio (changes requested):
fix itself checks out. one blocker in the comments: the publish half has no test that can fail.

two pre-existing issues found while tracing, filing separately:

- bootstrap.rs:1794: the failed-quarantine tombstone doesn't fence the reconciler (`contains` reads only `namespace_map`), so the namespace gets rebuilt - truncating the refused chain's segment 0 when the superblock frontier is 0 - and lands still-tombstoned with no `untombstone` path: unreachable forever, fast-skip pinned off. `fence_partition_for_rebuild`'s failure arm hits the same wedge at runtime.

hubcio (approved):

mmodzelewski (approved):

Issue #3573, opened 2026-06-27
Draft MSSQL Connector


lsabi:
### Description

As mentioned in the discussion regarding `connectors` (see ([https://github.com/apache/iggy/issues/2753#issuecomment-4750883206](https://github.com/apache/iggy/issues/2753#issuecomment-4750883206)), I can share the source code for a project that has been developed with the goal of capturing changes from a `MSQL` database and push them to `Iggy`. It took the changes, grouped them in case of update, so that these were processed atomically on a different system.

Of course it has to be adapted to comply with the standards of this organization, but a good part can be used.

[https://github.com/lsabi/msql_cdc/](https://github.com/lsabi/msql_cdc/)

In case of doubts, feel free to ask, as at the moment I don't have time to work on the development of a proper, production-ready, connector, so I'm sharing it with the hope that someone has more time than me

### Affected area / component

Connectors

### Contribution

- [X] I'm willing to submit a pull request to implement this feature
`Edit 11/07/2026 I'm working on the source connector`

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

lsabi:
I'll try to start (and finish) working on it...at least the `source`

lsabi:
Draft is ready to be tested (will require adjustments). How can I test it? I only have a production MSSQL instance that would not like to use for testing

See [https://github.com/lsabi/iggy/tree/lsabi/mssql-source](https://github.com/lsabi/iggy/tree/lsabi/mssql-source)

hubcio:
issue unassigned from @lsabi, looking for contributor that wants to continue work based on https://github.com/lsabi/iggy/tree/lsabi/mssql-source
<img width="1796" height="579" alt="Image" src="https://github.com/user-attachments/assets/18441229-481e-4b81-a67e-dd37e62c3e16" />

mattp5657:
I can take.

hubcio:
@mattp5657 assigned

PR #3851, opened 2026-08-09
MSSQL Source Connector

lsabi:

Partially Closes # [https://github.com/apache/iggy/issues/3573](https://github.com/apache/iggy/issues/3573)
Relates to #

## Rationale

This is just the implementation of the source connector for the `cdc`. I'm also working on the table polling, but it'll require more time.

I don't have a database instance to test it, but tests are passing. If someone, who has a test `MSSQL` instance can try it out, it would be great.

## What changed?

Added the

## Local Execution

- Unit Tests pass. Not tested with a real database instance, as I don't have an instance with dummy data to ruin (only production)
- Pre-commit hooks not ran

## AI Usage

None


lukaszzborek:
@lsabi please make this pr as draft or close it and add test, then reopen
there is docker image [microsoft/mssql-server](https://hub.docker.com/r/microsoft/mssql-server) which you can use

lsabi:
@lukaszzborek I'm afraid but I can't test it on my locale. I've spent a day understanding the docs, as it's not super clear how one should do in order to build their own connector.

I've then let my laptop run for almost a day and it did not finish compiling, just excessive heating.

I managed to compile `mssql_source` in a couple of hours, but not the `runtime` nor even the whole project. If you want, you can test it, but I've discovered that `iggy` and `runtime` are separate and require two different processes in order to run, which is what I did not want (I though the dynamic runtime could start up libraries on the fly and was part of the "core").

lukaszzborek:
@lsabi, quite intresting. Did you join to our discord? there are many people which can help. You can find discord invite in readme. Also you can merge current master to yours, because you have old version of server. And please write on discord what machine do you have (where you compile code)

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+						WHERE t.[__$start_lsn] = t_json.[__$start_lsn]
+							AND t.[__$update_mask] = t_json.[__$update_mask]
+							AND t.[__$command_id]= t_json.[__$command_id]
+						ORDER BY t_json.[__$command_id] ASC
+						FOR JSON AUTO
+					) as data
+				FROM [{db}].cdc.[{table}] t
+				WHERE {query_filter}
+				GROUP BY [__$start_lsn], [__$command_id], [__$seqval], [__$end_lsn], [__$update_mask]
+				ORDER BY [__$start_lsn] ASC, [__$command_id] ASC;";
```
hubcio: leftover `;";` inside the sql string - every poll fails with error 105 "unclosed quotation mark after the character string". ran the connector against sql server 2022 with cdc enabled: as shipped it never produces a single message.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+				"DELETE" => vec![1],
+				e => {
+					warn!("Invalid capture operation {e}. This will be ignored");
+					vec![]
+				}
+			})
+			.flatten()
+			.collect::<Vec<_>>();
+
+		let tables: Vec<String> = self.config.capture_table_columns.values().cloned().collect();
```
hubcio: this iterates the map's values (the column lists) as table names, so the query becomes `FROM [testdb].cdc.[id,name,active]` and fails with error 208 "invalid object name".

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+					@role_name	 = {cdc_role}"#);
+
+				self.query(Query::new(query)).await?;
+			}
+		}
+
+		// For each monitored table, get the capture columns
+		let table_columns = self.query(Query::new(
+			format!(r#"
+				SELECT OBJECT_NAME(ct.object_id, DB_ID('{database}')) as table_name, c.object_id, ct.capture_instance, ct.start_lsn, c.columns FROM [{database}].[cdc].[change_tables] as ct JOIN (SELECT object_id, STRING_AGG(column_name, ',') as columns FROM [{database}].[cdc].[captured_columns] GROUP BY object_id) as c ON c.object_id = ct.object_id;"#
```
hubcio: `OBJECT_NAME(ct.object_id, DB_ID(...))` names the change table (`dbo_users_CT`), not the source table - in `cdc.change_tables`, `object_id` is the change table's id and `source_object_id` is the source table's. so `capture_table_columns` ends up keyed by change-table names and a user-provided entry never matches the lookup on line 411. use `ct.source_object_id`.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+		let default_columns = "*".to_string();
+		for table in tables {
+			let columns = match self.config.capture_table_columns.get::<String>(&table) {
+				Some(x) => x,
+				None => &default_columns
+			};
+			let changes_query = format!(
+				r#"SELECT 
+						TODATETIMEOFFSET(
+						[#db].sys.fn_cdc_map_lsn_to_time([__$start_lsn]),
```
hubcio: `#db` is only substituted in the max-lsn query; here it reaches the server literally and fails with error 4121. the surrounding query already interpolates `{db}` via `format!`.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+
+		let mut messages = Vec::new();
+		for (table_name, row) in result {
+			for r in row {
+				let data: String = match r.get::<&str, _>("data") {
+					Some(x) => x.to_string(),
+					None => return Err(Error::InvalidRecord)
+				};
+
+				let operation_type = match r.get::<i64, _>("[__$operation]").unwrap_or(0) {
```
hubcio: this panics instead of erroring: the result column is `__$operation` (no brackets) and it's an `int`, so both the name and the `i64` type are wrong. tiberius `get` is `try_get(..).unwrap()`, and `try_get` errors on an unknown column name as well as on a failed conversion - so `unwrap_or(0)` is unreachable. the panic lands inside the task the sdk spawns for `poll()`, killing the polling task silently: the connector still reports as running and never produces again. use `try_get` with `i32` and return an error.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+						2 => "INSERT".to_string(),
+						3 => "UPDATE".to_string(),
+						4 => "UPDATE".to_string(),
+						e => {
+							error!("Operation type not recognized! {e}");
+							return Err(Error::InvalidRecord)
+						}
+					};
+
+				let timestamp = match r.get::<&str, _>("event_timestamp") {
```
hubcio, line 554: another panic: `event_timestamp` is a `datetimeoffset`, so `get::<&str>` blows up. even with a string conversion the format doesn't match `%Y-%m-%d %H:%M:%S %:z`, and the parsed value is discarded anyway - `DatabaseRecord.timestamp` below uses `Utc::now()`.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+						Ok(y) => y.with_timezone(&Utc),
+						Err(y) => {
+							warn!("Cannot parse datetime! {y}");
+							Utc::now()
+						}
+					}
+					None => Utc::now()
+				};
+
+				let parsed_data = parse_record_data(&data);
```
hubcio: `data` is json (`FOR JSON AUTO`) but this runs it through the text-format splitter, producing mangled output like `"data":{"":"1,\"name\":\"alice\",...}]"}`. parse it with `serde_json::from_str`. update events also arrive with before+after images concatenated in one array while `old_data` stays null.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+					@source_schema = N'{schema}',
+					@source_name   = N'{table}',
+					@role_name	 = {cdc_role}"#);
+
+				self.query(Query::new(query)).await?;
+			}
+		}
+
+		// For each monitored table, get the capture columns
+		let table_columns = self.query(Query::new(
```
hubcio, line 384: no filter on the configured tables here, so the connector streams every cdc-enabled table in the database - with `tables = ["users"]` configured, rows from an unconfigured `orders` table came through.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+				.await?;
+			res
+				.into_iter()
+				.map(|r| r.get::<&str, _>("name").unwrap().to_string())
+				.collect()
+		} else {
+		   self.config.tables.clone()
+		}.join(",");
+
+		let mut query = Query::new("SELECT name, is_tracked_by_cdc FROM sys.tables WHERE is_tracked_by_cdc = 0 AND name IN (@P1);");
```
hubcio: binding the comma-joined list as one `@P1` makes `name IN (@P1)` compare against the literal string `"t1,t2"` - matches nothing for more than one table, so the auto-enable path never fires on multi-table configs.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+
+		match self.config.mode.as_str() {
+			"cdc" => {
+				self.setup_cdc().await?;
+				info!(
+					"MSSQL CDC mode enabled for connector ID: {}",
+					self.id
+				);
+			}
+			"polling" => {
```
hubcio, line 209: "polling" is accepted here but `poll()` returns `Err(InvalidConfig)` forever since `poll_tables` is commented out - reject the mode in `open()` until it exists.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+
+		let n_lsn = lsn_to_hex(&new_lsn);
+		let query_filter = if state.last_poll_lsn.is_some() {
+			let last_poll_lsn = lsn_to_hex(&state.last_poll_lsn.unwrap());
+			format!("[__$start_lsn] > {last_poll_lsn} AND [__$start_lsn] <= {n_lsn}")
+		} else {
+			format!("[__$start_lsn] <= {n_lsn}")
+		};
+
+		let capture_ops = self
```
hubcio, line 465: `capture_ops` is built but never used - the `capture_operations` filter is silently ignored and every operation type gets streamed.

core/connectors/sources/mssql_source/src/lib.rs
```diff
@@ -0,0 +1,2103 @@
...
+		if self.verbose {
+			info!("CDC: Fetched {} change records", messages.len());
+		} else {
+			debug!("CDC: Fetched {} change records", messages.len());
+		}
+		Ok(messages)
+	}
+
+
+	/*
```
hubcio: over half the file is commented-out postgres source code (poll_tables, sqlx type mapping, nearly all tests) - please delete it, git keeps history. the crate also fails `cargo fmt --check` (tabs) and `cargo clippy -- -D warnings` (21 errors), both are ci gates.

core/connectors/sources/mssql_source/Cargo.toml
```diff
@@ -0,0 +1,60 @@
...
+[lib]
+crate-type = ["cdylib", "lib"]
+
+# [features]
+# Placeholder feature for future CDC using Supabase's ETL framework
+# cdc_pg_replicate = []
+
+[dependencies]
+async-trait = { workspace = true }
+base64 = { workspace = true }
```
hubcio: `base64` is only referenced inside commented-out code - dead dependency, drop it. machete won't catch this one: its heuristic matches the crate name inside comments, so the dead code keeps it looking used.

Cargo.toml
```diff
@@ -50,6 +50,7 @@ members = [
     "core/connectors/sources/elasticsearch_source",
     "core/connectors/sources/influxdb_source",
     "core/connectors/sources/postgres_source",
+    "core/connectors/sources/mssql_source",
```
hubcio, line 54: `Cargo.lock` isn't part of the PR - building adds ~238 lines (tiberius, async-native-tls, connection-string, ...). commit the lock update.

hubcio (changes requested):
took this for a spin against a real sql server 2022 container with cdc enabled. `open()` connects fine, but as shipped `poll()` never returns a single message - it took five local fixes to get the first record through, and the payloads were still mangled. detailed comments inline.

core/connectors/sources/mssql_source/src/lib.rs
lsabi: Since MSSQL has the source table and the change table with different names (by default `_ct` get's appended), either the user inserts the change table name or, there's need to make an internal conversion (while keeping a mapping of the names). Which one is the preferred one?

hubcio:
closing this, new assignment in https://github.com/apache/iggy/issues/3573

Issue #3635, opened 2026-07-09
connectors: no send acknowledgment from runtime to source plugins — state advances and rows are deleted before delivery is confirmed


tusharagrahari:
## Description

The connector runtime gives source plugins no way to know whether the messages they produced were actually delivered to Iggy. The runtime polls the source, sends the batch via the producer, and persists the source's saved state, but the send result is never fed back to the plugin, and offset advancement is not conditional on a successful send.

This is the source-side counterpart of #2927 (sink `consume()` return value discarded) and #2928 (offsets committed before sink processing). Same root cause: no delivery acknowledgment across the runtime/plugin boundary, just in the opposite direction.

Raised during review of the MySQL source (#3445), where it shows up twice:

1. In `delete` tracking mode, rows are deleted (or marked) in the source database as part of producing the batch before the runtime has sent anything. If the send fails, the row is gone from MySQL and never reached Iggy. Permanent data loss.
2. The tracking offset advances once the batch is handed to the runtime. A send failure doesn't roll it back, so the next poll starts past the failed batch.

`postgres_source` has the same behavior — this is not specific to the MySQL connector.

## Impact

At-least-once delivery isn't achievable for source connectors right now — same conclusion as #2928 on the sink side. For sources doing a destructive read (delete mode), this isn't "duplicates possible" anymore, it's actual data loss: the record's gone from the source DB and never landed in Iggy, and nothing gets logged about it.
Affects any polling source that tracks its own position, not just the SQL connectors.

## Related

- #2927 — sink consume() return value discarded (same gap, sink direction)
- #2928 — offsets committed before sink processing
- #2940 — replay-safe progress contract discussion
- #3445 — MySQL polling source PR where this was flagged in review

### Affected area / component

Connectors

### Deployment

None

### Contribution

- [ ] I'm willing to submit a pull request to fix this bug

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

rohankumardubey:
Hi @hubcio , I’m interested in helping with this issue. I noticed that #3568 and #3498 reference it but leave the runtime-to-source acknowledgment problem unresolved. Is anyone currently coordinating or implementing the underlying fix?
If it is available, I propose starting with a short design for a backward-compatible acknowledgment flow between the source plugin and runtime, together with a deterministic regression test for an Iggy send failure. I would not change the source FFI contract or connector SDK version without approval. Is this a useful scope?

hubcio:
AFAIK no one is working on it now.

[...]

PR #3855, opened 2026-08-10
feat(connectors): add source batch acknowledgments

rohankumardubey:

Relates to #3635

## Rationale

Establishes the generic source acknowledgment contract before migrating PostgreSQL source behavior and adding the kill-server regression test in a follow-up PR.

## What changed?

Source plugins could poll another batch without knowing whether the previous batch was delivered, allowing cursors and persisted state to advance after failed sends.

The source FFI now carries batch IDs and supports ACK/NACK results. The SDK permits one in-flight batch, while the runtime ACKs only after both the Iggy send and state persistence succeed. Send failures, timeouts, and state-save failures produce a NACK.

## Local Execution

- Passed `cargo fmt --all`
- Passed clippy with warnings denied for the connector SDK, runtime, and all source plugins
- Passed all connector SDK, runtime, and source-plugin tests
- Built the runtime and all source dynamic libraries successfully
- Confirmed the new `iggy_source_batch_result` symbol is exported
- `cargo-sort`, `markdownlint`, and `taplo` were not available locally
- Pre-commit hooks did not run because no hook was installed


core/connectors/runtime/src/source.rs
```diff
@@ -514,6 +527,20 @@ pub(crate) async fn source_forwarding_loop(
             } else {
                 debug!("No state provided for source connector with ID: {plugin_id}");
             }
+
+            if state_saved {
+                batch_result = SourceBatchResult::Ack;
```
hubcio, line 554: ack fires even when messages were dropped before the send - decode failures above and transform/encode/build failures inside `process_messages` shrink the batch, but nothing compares `sent_count` to `count`. extreme case: every message drops, `send(vec![])` returns ok without touching the server, still ack. under this contract ack tells the plugin to commit destructive work, so dropped messages are gone at the source.

`process_messages` already counts its errors - return that and nack when `decode_errors + error_count > 0` (keep `filtered_count` out, those drops are intentional). one catch: a plain nack on a message that can never decode means infinite redelivery, so this also needs a drop-after-n-attempts or dlq policy.

core/connectors/runtime/src/source.rs
```diff
@@ -461,6 +471,7 @@ pub(crate) async fn source_forwarding_loop(
 
         // Total histogram + emit (below) run regardless of send outcome.
         let mut state_save_us: Option<u64> = None;
+        let mut batch_result = SourceBatchResult::Nack;
         if let Err(error) = send_result {
```
hubcio, line 491: the error here already says which chunks committed (`ProducerSendFailed { committed, failed, .. }`) but it gets formatted into a string and the whole batch is nacked - redelivery then duplicates the committed prefix. resending just `failed` before deciding ack/nack would avoid that; at minimum log `committed.len()`.

core/connectors/runtime/src/source.rs
```diff
@@ -514,6 +527,20 @@ pub(crate) async fn source_forwarding_loop(
...
                 debug!("No state provided for source connector with ID: {plugin_id}");
             }
+
+            if state_saved {
+                batch_result = SourceBatchResult::Ack;
+            }
+        }
+
+        let result_code = batch_result_callback(plugin_id, batch_id, batch_result as u8);
+        if result_code != 0 {
```
hubcio, line 564: every stop with a batch in flight ends up here: `iggy_source_close` removes the INSTANCES entry before `cleanup_sender` runs, so the callback returns -1 and this logs an error, bumps the counter and sets connector error for a benign shutdown race. `handle_produced_messages` below treats the mirror case as trace-only for exactly this reason. side effect: the loop exit then sees status Error, so `update_status(Stopped)` skips the `sources_running` decrement and the gauge leaks on every restart.

core/connectors/runtime/src/source.rs
```diff
@@ -489,11 +500,13 @@ pub(crate) async fn source_forwarding_loop(
                 );
             }
 
+            let mut state_saved = true;
```
hubcio, line 525: state is saved (two fsyncs) every batch even when nothing moved. letting a source return `state: None` when nothing changed would skip it - the runtime already acks on None - though postgres would first have to stop re-stamping `last_poll_time` on every poll.

core/connectors/sdk/src/source.rs
```diff
@@ -157,22 +205,50 @@ impl<T: Source + std::fmt::Debug + 'static> SourceContainer<T> {
...
         });
 
         self.shutdown = Some(shutdown_tx);
         self.task = Some(handle);
         0
     }
+
+    #[doc(hidden)]
+    pub fn complete_batch(&self, batch_id: u64, result: u8) -> i32 {
+        let Ok(result) = SourceBatchResult::try_from(result) else {
```
hubcio: an unrecognized code returns -1 without completing the pending batch - the plugin then waits on `result_receiver` forever and the runtime waits on `recv_async`, a permanent stall on both sides. safer to treat unknown as nack and complete the batch. unreachable today, but it arms the moment `SourceBatchResult` grows a third variant, so worth fixing before that ever happens.

core/connectors/sdk/src/source.rs
```diff
@@ -192,16 +268,111 @@ async fn handle_messages<T: Source>(
...
+                if callback(plugin_id, batch_id, messages.as_ptr(), messages.len()) != 0 {
+                    _ = complete_pending_batch(
+                        &pending_batch,
+                        batch_id,
+                        SourceBatchResult::Nack,
+                        plugin_id,
+                    );
+                }
+
+                let (result, shutting_down) = tokio::select! {
```
hubcio: no timeout on this wait. the oneshot sender lives on the plugin side, so if the runtime's forwarding task dies between receiving the batch and calling the callback (say a panicking transform), nothing is ever sent and the `unwrap_or(Nack)` can't fire - the plugin parks forever with status still Running. bound the wait and nack on expiry.

core/connectors/sdk/src/source.rs
```diff
@@ -192,16 +268,111 @@ async fn handle_messages<T: Source>(
...
+                    biased;
+                    result = result_receiver => {
+                        (result.unwrap_or(SourceBatchResult::Nack), false)
+                    },
+                    _ = shutdown.changed() => {
+                        _ = complete_pending_batch(
```
hubcio: by the time this arm runs the select has already dropped `result_receiver`, so this send always fails and logs "Failed to deliver batch result..." on every clean shutdown with a batch in flight - deterministic, not a race. clearing the slot (`*pending = None`) is all that's needed here.

core/connectors/sdk/src/source.rs
```diff
@@ -192,16 +268,111 @@ async fn handle_messages<T: Source>(
...
+                            &pending_batch,
+                            batch_id,
+                            SourceBatchResult::Nack,
+                            plugin_id,
+                        );
+                        (SourceBatchResult::Nack, true)
+                    }
+                };
+                if !notify_source(&source, result, plugin_id).await {
```
hubcio: two things: when `on_batch_result` errors, the loop breaks and the runtime never finds out - status stays Running, no metric, the forwarding loop keeps waiting. and a nack re-polls immediately with no backoff; the sdk already ships `retry::exponential_backoff`, it's just not used here.

core/connectors/sdk/src/source.rs
```diff
@@ -34,23 +37,68 @@ pub struct RawMessage {
...
+        }
+    }
+}
+
+struct PendingBatch {
+    id: u64,
+    result_sender: oneshot::Sender<SourceBatchResult>,
+}
+
+impl std::fmt::Debug for PendingBatch {
```
hubcio: `#[derive(Debug)]` works here - `oneshot::Sender` is Debug.

core/connectors/sdk/src/source.rs
```diff
@@ -192,16 +268,111 @@ async fn handle_messages<T: Source>(
...
+                if !notify_source(&source, result, plugin_id).await {
+                    break;
+                }
+
+                if shutting_down {
+                    info!("Shutting down source connector with ID: {plugin_id}");
+                    break;
+                }
+
+                batch_id = batch_id.wrapping_add(1);
```
hubcio: 0 is not a sentinel anywhere and u64 won't wrap in practice - `batch_id += 1` is enough.

core/connectors/sdk/src/source.rs
```diff
@@ -192,16 +268,111 @@ async fn handle_messages<T: Source>(
...
+    };
+    if current.id != batch_id {
+        error!(
+            "Batch result ID mismatch for source connector with ID: {plugin_id}. Expected: {}, received: {batch_id}",
+            current.id
+        );
+        return -1;
+    }
+
+    let Some(current) = pending.take() else {
```
hubcio: this else-arm is unreachable (the guard above already proved Some) and returns -1 with no log. `Option::take_if` collapses the check and the take - just keep the expected id in the error message, a mismatch is the only visible sign of counter desync.

core/connectors/sdk/src/source.rs
```diff
@@ -192,16 +268,111 @@ async fn handle_messages<T: Source>(
...
+                let (result_sender, result_receiver) = oneshot::channel();
+                {
+                    let mut pending = lock_pending_batch(&pending_batch);
+                    *pending = Some(PendingBatch {
+                        id: batch_id,
+                        result_sender,
+                    });
+                }
+
+                if callback(plugin_id, batch_id, messages.as_ptr(), messages.len()) != 0 {
```
hubcio: `messages` stays alive across the whole result wait, but the runtime copies the bytes out before the callback returns - dropping it right after the callback frees the duplicate buffer for the whole send window.

core/connectors/sdk/src/lib.rs
```diff
@@ -111,6 +111,17 @@ pub trait Source: Send + Sync {
...
     async fn poll(&self) -> Result<ProducedMessages, Error>;
 
+    /// Invoked after the runtime has finished processing the most recently polled batch.
+    ///
+    /// Sources that track cursors or perform destructive operations should stage those changes
+    /// in [`Source::poll`] and apply them only after receiving [`source::SourceBatchResult::Ack`].
+    /// A [`source::SourceBatchResult::Nack`] means the staged changes must be discarded so the
+    /// batch can be polled again. The SDK allows only one batch to be in flight at a time and
+    /// stops polling if this method returns an error.
+    async fn on_batch_result(&self, _result: source::SourceBatchResult) -> Result<(), Error> {
```
hubcio, line 122: nothing in-tree implements this - every shipped source keeps the no-op default, and postgres still deletes/marks rows and advances its cursor inside `poll()`, so a nack rolls back nothing. worse, the loss becomes durable: after a failed send the next poll is empty (rows already consumed) but still carries the advanced cursor, `send(vec![])` short-circuits ok, and the state file gets saved. the postgres migration needs to land with this, or the readme claims scoped down until an adopter exists.

core/connectors/sdk/README.md
```diff
@@ -4,6 +4,33 @@ SDK provides the commonly used structs and traits such as `Sink` and `Source`, a
...
+
+Source connectors use a one-in-flight-batch contract between the plugin and the runtime:
+
+1. `Source::poll()` returns messages and candidate state without committing cursor changes or destructive operations.
+2. The runtime sends the batch to Iggy and waits for the producer result.
+3. After a successful send, the runtime persists the candidate state.
+4. The runtime reports `SourceBatchResult::Ack` to the plugin. A send or state-save failure reports `SourceBatchResult::Nack` instead.
+5. `Source::on_batch_result()` commits or discards the plugin's staged work before the next poll starts.
+
+An empty batch follows the same handshake. This prevents a successful no-op send from persisting state left over from an earlier failed delivery. Producer errors, including request timeouts, report a NACK. A successful send from the legacy Iggy server is still an ACK even though that server returns an empty confirmation list.
```
hubcio: this sentence only holds if the plugin rolled back on the earlier nack - no shipped plugin does, so today the empty poll is exactly what persists the stale cursor (see postgres). worth rewording until an adopter exists.

core/connectors/sdk/README.md
```diff
@@ -4,6 +4,33 @@ SDK provides the commonly used structs and traits such as `Sink` and `Source`, a
...
+
+The crash behavior is intentionally at-least-once:
+
+| Crash point | Recovery behavior |
```
hubcio, line 21: the table misses a server-side row: a batch is confirmed once committed in server memory, not fsynced, so ack -> destructive cleanup -> server crash before fsync loses the data on both ends. worth a row here plus a note on `Ack` that it means committed-in-memory (durability rides on the server's `enforce_fsync`, which ships off).

rohankumardubey:
@hubcio This PR is ready for review.

hubcio (changes requested):
i actually had comments ready yesterday, just as pending, forgot to post them. here a few things that don't sit on changed lines:

- no test exercises the failure path - nothing forces a send or state-save failure and checks that the nack arrives and the state file stays put. and no integration test ever waits for a second batch, so a handshake that wedges after batch 1 still passes the suite.
- `manager/source.rs:150` comment "close FFI (stops callbacks)" is stale now - close doesn't stop the runtime->plugin result callback, it guarantees it fails.
- the connector authoring docs and templates don't mention `on_batch_result`, so the next source written from them lands with the no-op default too.

hubcio:
ignore failing java examples, we'll check it

rohankumardubey:
Thanks for the detailed review. I agree with the issues you identified and will address the ABI mismatch, shutdown and timeout handling, callback error propagation, NACK backoff, docs/templates, and missing regression tests.

For processing failures, I propose NACK with capped backoff, eventually stopping the connector without ACKing instead of losing data. For partial producer failures, I’ll report committed/failed counts and retry only the failed tail before NACKing.

As agreed in #3635, I’ll keep PostgreSQL migration in the follow-up PR, use a small generic source to test this contract, and narrow the README claims until PostgreSQL is migrated. Does this look good?

hubcio:
@rohankumardubey looks good.

hubcio:
looks like test is failing @rohankumardubey 
```
TRY 4 FAIL [  16.782s] (553/692) integration::mod connectors::random::random_source::send_failure_preserves_state_and_source_recovers
```

https://github.com/apache/iggy/actions/runs/31840994243/job/94897663656?pr=3855

rohankumardubey:
> looks like test is failing @rohankumardubey
> 
> ```
> TRY 4 FAIL [  16.782s] (553/692) integration::mod connectors::random::random_source::send_failure_preserves_state_and_source_recovers
> ```
> 
> https://github.com/apache/iggy/actions/runs/31840994243/job/94897663656?pr=3855

The test was stopping only node 0 while running against a three-node cluster. After the first failure, the producer failed over to another node, so the test timed out waiting for a second failure. I’ve changed this regression to use a single-node cluster, which keeps the outage deterministic. I’ll push the fix after the local checks are complete.

hubcio:
@rohankumardubey looks good, we'll merge this once github will start working properly again

hubcio (approved):

rohankumardubey:
> @rohankumardubey looks good, we'll merge this once github will start working properly again

Thanks for the review! I’ve already started working on the PostgreSQL follow-up from #3635, including the staged cursor changes and deterministic kill-server regression test we discussed.

spetz (approved):

hubcio (approved):

hubcio:
@rohankumardubey have you joined our discord? if yes, then whats your handle?

rohankumardubey:
> @rohankumardubey have you joined our discord? if yes, then whats your handle?

@hubcio its kratos_4497 , and i have joined Iggy discord.

PR #3856, opened 2026-08-10
feat(server)!: promote server-ng to the Apache Iggy server

hubcio:
Maintaining two servers meant every protocol change, config knob
and test lane got built twice, and the vsr feature let the
replicated paths drop out of a default build silently. Every SDK
and CI lane already targets the VSR server, so the fork can go.

VSR is now unconditional: the feature is gone, classic TCP, QUIC
and WebSocket framing is deleted, and the integration harness
always spawns the VSR server with a 3-node default cluster. The
legacy server, its compat lane, bdd compose flavor and config
modules are removed, and server-ng takes over the server package,
the iggy-server binary, the configs tree and the systemd
sd-notify integration. The published Dockerfile gains the
license generation stage ASF requires.

Restart scenarios are pinned to one node because a 3-node
topology trips a pre-existing partitions-plane view-change stall,
proven present at the base commit and tracked separately. The
cutover is also the only free moment to rotate the JWT
key-derivation context, so it is renamed here before the string
re-freezes.

BREAKING CHANGE: the wire protocol changed, so older SDKs and
CLIs cannot connect. The --fresh and
--with-default-root-credentials flags are gone, leaving
--replica-id. The -p server-ng and --bin iggy-server-ng cargo
targets no longer exist, and the SDK crate no longer has a vsr
feature.


scripts/run-examples-from-readme.sh
```diff
@@ -244,8 +238,7 @@ run_node_examples() {
 # shellcheck disable=SC2329
 run_go_examples() {
     # The Go SDK speaks only the VSR wire protocol.
-    # TODO: change to iggy-server once legacy server is removed (core/server has VSR support)
-    resolve_server_binary "${TARGET}" "iggy-server-ng" "vsr"
+    resolve_server_binary "${TARGET}" "iggy-server"
     # The VSR server logs no startup line, so readiness is a connect poll.
```
hubcio: comment is false - the server does log "server client listeners started" at info, and `run_java_examples` relies on exactly that pattern at line 358. drop the comment and the tcp-probe override, or fix the rationale.

core/server/src/args.rs
```diff
@@ -21,118 +21,14 @@ use clap::Parser;
...
-    ///
-    /// Supplying the identity on the command line lets operators ship a
-    /// single byte-identical `config.toml` to every node in the cluster
-    /// and differ only in this CLI flag.
-    ///
-    /// Examples:
-    ///   iggy-server --replica-id 0                       # This node is replica 0
+    /// one `cluster.nodes[*].replica_id` entry in the loaded configuration.
     #[arg(long, verbatim_doc_comment)]
     pub replica_id: Option<u8>,
```
numinnex, line 125: `--fresh` and `--with-default-root-credentials` are removed here, but several docs still teach them against `iggy-server`, so the first commands newcomers copy-paste now fail with an unexpected-argument error:

- `README.md` lines 276-282 (the file is touched by this PR, this section was missed)
- `examples/rust/README.md` lines 12, 37, 44, 47
- `examples/php/README.md` line 11
- `foreign/php/README.md` line 73
- `foreign/python/README.md` lines 80, 84
- `examples/rust/src/message-headers/message-compression/README.md` line 14
- `examples/rust/src/message-headers/message-compression/producer/main.rs` line 40 (a runtime error message that tells the user to restart the server with `--fresh`)

Worth cleaning these up in this PR, since this is the change that makes the break.

hubcio: i will add it back, i actually have it ready on my branch but didnt push it yet, waiting for yours https://github.com/apache/iggy/pull/3836 to be merged

mmodzelewski (approved):

numinnex (approved):

PR #3858, opened 2026-08-10
refactor(csharp): cleanup tcp connection after vsr implementation

lukaszzborek:
Follow-up cleanup of the .NET SDK TCP transport after the VSR support landed


lukaszzborek:
If possible, merge https://github.com/apache/iggy/pull/3836 first. Then I will remove things with namespace which I refactoring 😅

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
```diff
@@ -471,32 +439,52 @@ private async Task<bool> RedirectAsync(CancellationToken token)
...
         }
-        finally
+    }
+
+    /// <summary>
+    ///     Reads the roster without following redirects: the probe is what redirects are decided from, so a
+    ///     probe that redirected or reconnected would reenter the very loop that called it. A probe the current
+    ///     node keeps refusing simply fails, and the caller stays where it is.
+    /// </summary>
+    private async Task<ClusterMetadata?> ReadClusterMetadataNoRedirectAsync(CancellationToken token)
```
hubcio, line 461: if the probe's socket dies, `DropVsrConnectionLocked` has already reset the session and set Disconnected, but the catch in `GetCurrentLeaderNodeAsync` turns that into "no leader" and callers carry on: `LoginRegisterAsync` returns a success `AuthResponse` on a disconnected, unbound client (with auto-login off nothing ever re-authenticates), and the send loop keeps retrying the dead connection and surfaces UNAUTHENTICATED with no reconnect. let connection exceptions propagate from that catch (`e is not OperationCanceledException && !VsrConnection.IsConnectionException(e)`) - a dead socket is not "no leader" - and re-check state before `LoginRegisterAsync` returns success.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
```diff
@@ -1015,12 +920,12 @@ private async Task TryEstablishConnectionAsync(bool autoLogin, CancellationToken
...
         {
-            // The sending semaphore owns every write to _stream, so an in-flight request never observes the
-            // field changing between its write and its reply.
+            // The sending semaphore owns every write to _connection, so an in-flight request never observes
+            // the field changing between its write and its reply.
             await _sendingSemaphore.WaitAsync(token);
             try
             {
-                _stream?.Dispose();
+                _connection?.Dispose();
```
hubcio: `_connection` is never set back to null, here or in `DropVsrConnectionLocked`, so a disposed connection stays reachable behind a `Connected` state for the whole reconnect window, and a request hitting it can surface `VsrRequestOutcomeUnknownException` for bytes that never left the process. it also leaves the `connection is null` branch in `SendVsrAttemptAsync` dead. set `_connection = null` after dispose/close at both sites (still under the sending semaphore) - that branch then turns these into `NotConnectedException`, which the reconnect path already handles.

foreign/csharp/Iggy_SDK/Vsr/VsrNamespace.cs
```diff
@@ -56,9 +56,38 @@ internal static class VsrNamespace
...
 
     /// <summary>
-    ///     The namespace a request header must carry, peeking into the classic payload for the ops the server
-    ///     routes by partition. Named (string) stream or topic identifiers resolve to 0; the server resolves
-    ///     them itself.
+    ///     The namespace a request header carries for its operation. Register and logout ride the metadata
+    ///     consensus group; non-replicated and metadata operations route by 0; partition-plane operations carry
+    ///     the namespace the caller packed from its typed identifiers.
+    /// </summary>
+    internal static ulong ForOperation(VsrOperation operation, ulong partitionNamespace)
```
hubcio: `ForOperation` returns whatever the caller passed, while the old resolver threw FEATURE_UNAVAILABLE for a partition op it could not route. a call site that forgets to pass `ns` now silently sends namespace 0 - a replicated write routed to the wrong shard. all current sites pass the right value, but the safety net is gone. making the parameter `ulong?` and throwing on null for partition-plane ops keeps explicit 0 valid for named identifiers (the server resolves those) while catching the forgotten case.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -505,6 +519,38 @@ private async Task OnClientConnectionStateChangedAsync(ConnectionStateChangedEve
...
+                    return;
+                }
+
+                if (e.CurrentState != ConnectionState.Authenticated)
+                {
+                    // Polling under a dropped session would be refused by the server for as long as the
+                    // reconnect takes to re-authenticate, so the membership is surrendered up front.
+                    if (e.CurrentState == ConnectionState.Disconnected)
+                    {
+                        _joinedConsumerGroup = false;
```
hubcio: this clear is still edge-based inside the branch that exists to stop depending on edges. state events can be published concurrently (`SetConnectionStateAsync` is a plain read-modify-write with the publish after the write), so a late Disconnected can land after the Authenticated arm already rejoined and stamped the new epoch - the flag stays false and polling silently stops until the next Authenticated event, which on a healthy connection never comes. gating the poll on `_joinedSessionEpoch == provider.SessionEpoch` for group consumers removes the dependence on event arrival; plain consumers must keep the early return above (they never re-stamp), and the epoch read then happens outside `_connectionStateSemaphore`, so access it with `Interlocked.Read`/`Exchange(ref ulong)`.

foreign/csharp/Iggy_SDK_Tests/ConsumerTests/ConsumerSessionGenerationTests.cs
```diff
@@ -0,0 +1,197 @@
...
+        var client = new EpochClient();
+        var consumer = new IggyConsumer(client.Object, BuildGroupConfig(), NullLoggerFactory.Instance);
+        await consumer.InitAsync(TestContext.Current.CancellationToken);
+        Assert.Equal(1, client.JoinCount);
+
+        // The session the membership was established under is gone.
+        client.SessionEpoch++;
+        await client.RaiseAsync(ConnectionState.Connected, ConnectionState.Authenticated);
+
+        Assert.Equal(2, client.JoinCount);
```
hubcio, line 69: nothing pins that the rejoin re-stamps `_joinedSessionEpoch` - delete that stamp in the consumer and all four tests stay green while every reconnect triggers a rejoin (a group-wide rebalance each time). raise a second `(Connected, Authenticated)` here and assert `JoinCount` is still 2.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
```diff
@@ -996,16 +875,42 @@ private static ReadOnlySpan<Message> AsSpan(IList<Message> messages)
...
-        BinaryPrimitives.WriteInt32LittleEndian(buffer[..4], messageBufferSize + 4);
-        BinaryPrimitives.WriteInt32LittleEndian(buffer[4..8], CommandCodes.SEND_MESSAGES_CODE);
-        return messageBufferSize;
+        if (partitioning.Kind != Enums.Partitioning.PartitionId)
+        {
+            throw VsrError.Exception(VsrError.FEATURE_UNAVAILABLE,
+                "Under VSR the partition must be resolved by the client; balanced and message-key partitioning cannot be sent on the wire.");
+        }
+
+        if (partitioning.Value.Length < 4)
```
hubcio: this checks the backing array, but the wire frames by `partitioning.Length` - `{ Length = 8, Value = 8 bytes }` passes here while the old path rejected anything but 4. the server refuses such a frame anyway, so the gain is a typed client-side error before a wasted round trip: `partitioning.Length != 4 || partitioning.Value.Length != 4` covers both mismatch directions.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
```diff
@@ -528,10 +509,12 @@ private async Task<IMemoryOwner<byte>> SendRawVsrAsync(ReadOnlyMemory<byte> payl
                         StatusCode: VsrError.TRANSIENT_NOT_ACCEPTED, FromServer: true
                     }
                     && !isLoginRegister
+                    && allowRedirect
                     && Environment.TickCount64 < overallDeadline)
                 {
-                    if (await RedirectAsync(token))
+                    if (redirects < VsrMaxLeaderRedirects && await RedirectAsync(token))
```
hubcio, line 527: the old redirect path logged "Maximum leader redirections reached, continuing on {Address}" when the budget ran out. now exhaustion is silent and the loop just spins replay windows until the 30s deadline - the login and connect paths kept their logs, this is the only dark one. log once when the budget runs out.

foreign/csharp/Iggy_SDK/Vsr/VsrNamespace.cs
```diff
@@ -175,6 +204,22 @@ private static ulong FromDeleteSegments(ReadOnlySpan<byte> payload)
         return FromPartition(streamId, topicId, ReadUInt32(payload, position));
     }
 
+    /// <summary>Numeric identifier value, or <c>null</c> for a named identifier the server has to resolve.</summary>
+    private static uint? NumericValue(Identifier identifier)
```
hubcio: a bad `Identifier` here throws INVALID_COMMAND with "Request payload is too short to resolve a namespace." - there is no payload on this path, and `Validate` below uses INVALID_IDENTIFIER for the same kind of defect. same code plus a message about the identifier would fit better.

foreign/csharp/Iggy_SDK/Iggy_SDK.csproj
```diff
@@ -68,13 +68,15 @@
...
+        <PackageReference Include="System.IO.Hashing"/>
     </ItemGroup>
 
     <ItemGroup>
-        <InternalsVisibleTo Include="Apache.Iggy.Tests" />
-        <InternalsVisibleTo Include="Apache.Iggy.Tests.Integrations" />
+        <InternalsVisibleTo Include="Apache.Iggy.Tests"/>
+        <InternalsVisibleTo Include="Apache.Iggy.Tests.Integrations"/>
+        <!-- Lets the test mocking framework proxy internal interfaces such as ISessionEpochProvider. -->
+        <InternalsVisibleTo Include="DynamicProxyGenAssembly2"/>
```
hubcio: the assembly is not signed, so this grant is name-only in the shipped package, and it exists for one `mock.As<ISessionEpochProvider>()` in tests. making `ISessionEpochProvider` public would drop the need for it and also let out-of-tree `IIggyClient` implementations opt into the epoch rejoin path, which they currently cannot.

foreign/csharp/Iggy_SDK_Tests/UtilityTests/SlicedMemoryOwnerTests.cs
```diff
@@ -50,18 +50,18 @@ public void Dispose_IsIdempotent()
     [Fact]
     public void FinalizerIsDeclared()
     {
-        Type sliced = typeof(ArrayPoolHelper)
+        var sliced = typeof(ArrayPoolHelper)
             .GetNestedType("SlicedMemoryOwner", BindingFlags.NonPublic)!;
 
-        MethodInfo? finalizer = sliced.GetMethod("Finalize", BindingFlags.NonPublic | BindingFlags.Instance);
+        var finalizer = sliced.GetMethod("Finalize", BindingFlags.NonPublic | BindingFlags.Instance);
```
hubcio: `GetMethod("Finalize", NonPublic | Instance)` without `DeclaredOnly` finds the inherited `Object.Finalize`, so this assert passes for any type - and `SlicedMemoryOwner` declares no finalizer. the test below only proves the GC collects an unreferenced object. do not add a finalizer to fix it (the pooled buffer is aliased by live `Memory<byte>` slices on the decrypt path, so returning it from a finalizer would be use-after-return) - add `DeclaredOnly` and delete both tests.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
```diff
@@ -138,16 +124,13 @@ public sealed partial class TcpMessageStream
...
-
             SetConnectionStateAsync(ConnectionState.Authenticating);
 
             LoginRegisterResponse response;
             try
             {
-                Interlocked.Exchange(ref _skipAutoLoginOnce, 1);
-                using IMemoryOwner<byte> responseBuffer = await SendWithResponseAsync(payload, token);
+                using IMemoryOwner<byte> responseBuffer =
+                    await SendWithResponseAsync(code, 0, message, token, false);
```
hubcio: bare trailing `false` here means `autoLoginOnReconnect`, and the same shape in `ReadClusterMetadataNoRedirectAsync` means `allowRedirect` - two different things, invisible at the call site. named arguments would help.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
```diff
@@ -873,27 +651,27 @@ private async ValueTask<bool> TryEnterSendingSemaphoreAsync()
...
     {
-        if (!ReferenceEquals(_stream, stream))
+        if (connection is null || !ReferenceEquals(_connection, connection))
         {
             return;
         }
 
         ResetConsensusSession();
-        _stream?.Close();
+        connection.Close();
```
hubcio: `Close()` sits between the session reset and the Disconnected publish, so a throwing close skips the state event and leaves `_state` Authenticated over a dead socket. neither stream can throw here today, but any wrapping stream that flushes on dispose would make it routine - this method only runs after a connection failure. publishing the state change first is free.

foreign/csharp/Iggy_SDK/Vsr/VsrConnection.cs
```diff
@@ -0,0 +1,355 @@
...
+                    // failure is reported as an outcome the server alone knows, which for a replicated write
+                    // tells the caller its request may have committed twice.
+                    token.ThrowIfCancellationRequested();
+                    requestStarted = true;
+
+                    // Header and body go out as two writes, straight from the buffers they were encoded into:
+                    // no frame-sized rent and no copy. The server frames by the announced size, and the
+                    // lockstep protocol already pays a full round trip per request, so the extra segment a
+                    // small request costs (with Nagle disabled) is noise against the reply wait.
+                    await _stream.WriteAsync(header, token);
```
hubcio: every request now pays a second write - with NoDelay that is a second segment (and a second TLS record) even for the 38-byte poll body. the comment argues the round trip dominates, which is plausible but unmeasured, and the old <= 4 KiB single-frame path was deleted on that reasoning. worth a produce/consume bench before locking this in.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
```diff
@@ -471,32 +439,52 @@ private async Task<bool> RedirectAsync(CancellationToken token)
...
+        ObjectDisposedException.ThrowIf(_disposed, this);
+
+        if (_state is ConnectionState.Disconnected or ConnectionState.Connecting)
+        {
+            throw new NotConnectedException();
+        }
+
         var isLoginRegister = code is CommandCodes.LOGIN_REGISTER_CODE or CommandCodes.LOGIN_REGISTER_WITH_PAT_CODE;
         var overallDeadline = Environment.TickCount64 + VsrRequestTimeoutMs;
         var headerBuffer = ArrayPool<byte>.Shared.Rent(VsrHeader.HEADER_SIZE);
```
hubcio: the request header is rented per request and threaded through two signatures, while the reply header is already a per-connection field. a `_requestHeaderBuffer` field on `VsrConnection` next to `_replyHeaderBuffer` drops the rent/return per request and a parameter off both signatures - it is re-encoded at the top of every attempt under the sending lock, so the reuse is safe.

foreign/csharp/Iggy_SDK/Vsr/VsrConnection.cs
```diff
@@ -0,0 +1,355 @@
...
+    ///     transport's job to ignore the call when a reconnect already replaced this connection.
+    /// </summary>
+    private readonly Action<VsrConnection> _onDropped;
+
+    private readonly byte[] _replyHeaderBuffer = new byte[VsrHeader.HEADER_SIZE];
+    private readonly int _requestTimeoutMs;
+    private readonly Stream _stream;
+
+    /// <summary>The session identity requests on this connection encode from.</summary>
+    internal ConsensusSession Session { get; }
```
hubcio: `Session` is only read inside this class - a private readonly field is enough.

foreign/csharp/Iggy_SDK/Vsr/VsrConnection.cs
```diff
@@ -0,0 +1,355 @@
...
+        _onDropped = onDropped;
+        _logger = logger;
+    }
+
+    public void Dispose()
+    {
+        _stream.Dispose();
+    }
+
+    internal void Close()
```
hubcio: `Close()` and `Dispose()` are the same operation (`Stream.Close()` is `Dispose()`), and `TcpMessageStream.Dispose()` calls both back to back. one of them can go.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
```diff
@@ -996,16 +875,42 @@ private static ReadOnlySpan<Message> AsSpan(IList<Message> messages)
...
-    [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private static int FillSendMessagesPayload(Span<byte> buffer, int maxMessageBufferSize,
-        Identifier streamId, Identifier topicId, Partitioning partitioning, ReadOnlySpan<Message> messages,
-        IMessageEncryptor? encryptor)
+    /// <summary>
+    ///     The routing namespace for a send: callers resolve balanced and message-key partitioning to an
+    ///     explicit partition before framing, so anything else reaching this point is a bug surfaced as the
+    ///     same error the server-side router would raise.
+    /// </summary>
+    private static ulong SendMessagesNamespace(Identifier streamId, Identifier topicId, Partitioning partitioning)
```
hubcio: `SendMessagesNamespace` and `ConsumerOffsetNamespace` duplicate `VsrNamespace`'s rules and two of its error strings verbatim, and as private statics they are unreachable from tests - the old null-partition reject coverage was lost with the deleted header test. moving both into `VsrNamespace` as internal statics gives the routing rules one home and makes them testable again.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -505,6 +519,38 @@ private async Task OnClientConnectionStateChangedAsync(ConnectionStateChangedEve
...
+                    return;
+                }
+
+                _joinedConsumerGroup = false;
+                await RejoinConsumerGroupOnReconnectionAsync();
+
+                return;
+            }
+
             if (e.CurrentState == ConnectionState.Disconnected)
```
hubcio, line 605: this fallback is unreachable in-tree: `HttpMessageStream.SubscribeConnectionEvents` has an empty body and never publishes, and an out-of-tree `IIggyClient` cannot implement the internal `ISessionEpochProvider` to take the branch above. the `ISessionEpochProvider` doc also says HTTP consumers fall back to connection-state edges, which is not true - they get no events at all. either make the interface public, or delete this and document group rejoin as built-in-TCP-only; if it stays, it needs the same plain-consumer carve-out as the epoch arm.

hubcio (changes requested):
solid direction overall - classic framing gone, the `VsrConnection` extraction, typed namespaces and the epoch rejoin are real improvements. blocking items are the probe losing its reconnect, the never-nulled `_connection`, the fail-open namespace and the edge-based Disconnected clear; the rest is small.

one thing outside the diff: `SetConnectionStateAsync` (TcpMessageStream.cs:1153) is a non-atomic check-then-set with the publish after the write, reachable from callers holding different locks - concurrent transitions can publish duplicates or a `PreviousState` that never actually preceded `CurrentState`. it predates this PR, and fixing it properly means moving the publish behind an ordered queue (holding a lock across the publish is not an option - the handler chain runs synchronously into socket I/O when the semaphores are uncontended, so a sync write fault re-enters the same lock). follow-up material, but it is the substrate under the Disconnected-clear race flagged inline.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
lukaszzborek: namespace was removed

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
lukaszzborek: namespace was removed

foreign/csharp/Iggy_SDK/Vsr/VsrConnection.cs
lukaszzborek: restored single send for <= 4KiB, tests show it's a bit faster in that case

hubcio (approved):

spetz (approved):

PR #3859, opened 2026-08-11
fix(server): preserve producer batches across replication

spetz:
Producer replication copied payloads after journaling and let each
replica stamp offsets from local state. This increased producer
latency and could leave repaired or retransmitted prepares with
divergent offsets after a view change.

Stamp each batch once on the primary, freeze it once, and share the
journaled allocation with the next replica. Backups validate the
exact batch and stamp before journal, forward, or ACK. View
reconciliation truncates divergent suffixes and restores append
cursors before retransmission.

Buffered segment growth also exposed filesystem allocation stalls in
producer tail latency. Add configurable local segment preallocation
that preserves logical file length and falls back when unsupported.

Keep segment persistence synchronous so client replies retain the
existing commit, storage, and ACK ordering.


core/partitions/src/messages_writer.rs
```diff
@@ -148,6 +156,39 @@ impl MessagesWriter {
...
+            target: "iggy.partitions.storage",
+            file = file_path,
+            preallocate_len = len,
+            "file preallocation size is unsupported, using buffered allocation"
+        );
+        return;
+    };
+
+    // KEEP_SIZE reserves extents without changing recovery-visible file length.
+    if let Err(error) = fallocate(file, FallocateFlags::FALLOC_FL_KEEP_SIZE, 0, len) {
```
numinnex: `fallocate()` runs synchronously on the compio shard thread, so with thread-per-core it blocks the whole shard rather than a single task. `FALLOC_FL_KEEP_SIZE` is metadata-only and normally sub-ms, but on network or otherwise slow filesystems this can stall at segment rotation. If that trade (pay once at rotation instead of at the append tail) is deliberate, a short comment here would help; otherwise it may be worth offloading if rotation-time stalls show up in benches.

core/consensus/src/plane_helpers.rs
```diff
@@ -95,27 +98,86 @@ where
...
+    consensus: &VsrConsensus<B, P>,
+    message: Frozen<MESSAGE_ALIGN>,
+) -> Result<(), SendError>
+where
+    B: MessageBus,
+    P: Pipeline<Entry = PipelineEntry>,
+{
+    let header =
+        *bytemuck::checked::try_from_bytes::<PrepareHeader>(&message[..size_of::<PrepareHeader>()])
+            .expect("validated prepare must begin with a valid prepare header");
```
numinnex: This `.expect()` (together with the `assert!(header.op > commit_min)` in `replication_target`) panics the shard on a malformed buffer. The panic contract is documented and both current call sites in `iggy_partition.rs` pass post-validation frozen buffers, so the invariant holds today, but it is fragile if a future caller feeds an unvalidated buffer. Fine if the assert-heavy VSR style is intended; otherwise consider returning an error instead.

numinnex (approved):

thepatrykk (approved):

PR #3860, opened 2026-08-11
fix(connector): iceberg dummy partition for every data

### LINKED ISSUE #3853: fix(connector-iceberg): partition logic hardcodes dummy values and ignores table spec
### Bug description

Description of Bug
The Apache Iceberg sink routing logic (write_data in router/mod.rs) contains severe bugs in partition key construction:

Hardcoded Zero/Dummy Partition Values: primitive_type_to_literal returns static zero/empty values (PrimitiveLiteral::Int(0), PrimitiveLiteral::String(String::new()), etc.). When passed into get_partition_type_value and PartitionKey::new(...), every batch is written under a single epoch-zero/dummy partition regardless of payload contents.

Empty PartitionSpec Construction: The code calls PartitionSpec::builder(table.current_schema_ref()).build(), which creates an unpartitioned spec (Spec ID 0) rather than retrieving the active spec via table.metadata().default_partition_spec().

Missing Record Fanout: DataFileWriter expects a single PartitionKey per writer instance. There is currently no logic to evaluate partition transforms against incoming RecordBatch data or group records across target partition writers.

Area
Connectors (iggy-connector-sdk / Iceberg connector) — High-Risk Area

Steps to Reproduce
Point the Iceberg sink to an Iceberg table partitioned by a field (e.g., identity(date) or day(event_timestamp)).

Produce JSON messages with varying timestamp/date payload fields into Iggy.

Observe sink output: all records end up in epoch 0 / default fallback partition paths.

### Affected area / component

Connectors

### Deployment

Docker (self-built image)

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

### Contribution

- [ ] I'm willing to submit a pull request to fix this bug

### Good first issue

ashudeqode:
Closes #3853 

Rationale
The Iceberg sink's [write_data](vscode-file://vscode-app/snap/code/254/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) function had three critical bugs that caused all records to land in a single fake partition, regardless of the table's actual partition spec. Partition values were hardcoded to dummy zeros (e.g., Int(0), String("")), the partition spec was built from scratch as empty instead of read from table metadata, and there was no fan-out logic to route records from a mixed-partition batch to separate data files.

What changed?
The sink was populating a PartitionKey with static zero-value literals and pairing it with a newly-built empty PartitionSpec, so every record — regardless of its actual partition column values — was written under a single dummy partition (e.g., year=1970).

Now the sink reads the real partition spec from [table.metadata().default_partition_spec()](vscode-file://vscode-app/snap/code/254/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). For unpartitioned tables, the writer is built with [None](vscode-file://vscode-app/snap/code/254/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) partition key (unchanged write path, zero overhead). For partitioned tables, a [RecordBatchPartitionSplitter](vscode-file://vscode-app/snap/code/254/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) evaluates the table's partition transforms (day(), bucket(), identity(), etc.) against each Arrow RecordBatch, and a [FanoutWriter](vscode-file://vscode-app/snap/code/254/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) fans records out to per-partition DataFileWriter instances. primitive_type_to_literal and get_partition_type_value were removed as dead code.

Local Execution
Passed: tested with a local iggy-connect runtime writing to a partitioned Iceberg table (REST catalog)
Verified that records with different partition values land in separate Parquet files under correct partition paths
Verified unpartitioned tables still work with no behavioral change
Pre-commit hooks ran
AI Usage
GitHub Copilot (VS Code)
API exploration of the [iceberg](vscode-file://vscode-app/snap/code/254/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) 0.9.1 crate, code review of the fix, and generating this description
Compiled with cargo check -p iggy_connector_iceberg_sink, then built via Docker and tested against a live local Iceberg catalog
Yes


hubcio:
@EdgarModesto23 could you please check this?

hubcio:
@ashudeqode please dont remove the PR template, fill it so we can see reasoning behind this change.

ashudeqode:
> @ashudeqode please dont remove the PR template, fill it so we can see reasoning behind this change.

@hubcio Updated the description

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -201,42 +137,109 @@ async fn write_data(
...
-                Error::InvalidRecordValue(chain)
-            })?;
-            writer.write(batch_data).await.map_err(|err| {
-                let chain = format_error_chain(&err);
-                error!("Error while writing record batch: {}", chain);
-                Error::WriteFailure(chain)
-            })?;
+    let partition_spec = table.metadata().default_partition_spec();
+
+    let data_files = if partition_spec.is_unpartitioned() {
```
EdgarModesto23, line 142: Could we extract the shared batch-processing and writer-closing logic here? The partitioned and unpartitioned branches are nearly identical, with only the write operation differing. A small writer abstraction or helper could encapsulate that variance and make this easier to maintain.

EdgarModesto23 (changes requested):

core/connectors/sinks/iceberg_sink/src/router/mod.rs
ashudeqode: sure, I'll do it.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -201,42 +137,109 @@ async fn write_data(
...
+            Ok(())
+        }
+        .await;
+
+        if let Err(e) = &write_result {
+            error!(
+                "Batch loop failed ({}), closing writer to release resources",
+                e
+            );
+            if let Err(close_err) = fanout_writer.close().await {
```
mattp5657, line 228: On a write failure we're possibly leaving orphaned files in the Iceberg table location. `close()` still finalizes whatever partitions already succeeded, but we never commit them and just throw away the result. Could we grab the files from `close()`'s return and delete them before returning the error?

core/connectors/sinks/iceberg_sink/src/router/mod.rs
```diff
@@ -201,42 +137,109 @@ async fn write_data(
...
-            e
-        );
-        if let Err(close_err) = writer.close().await {
-            error!("Failed to close writer after batch error: {}", close_err);
+        if let Err(e) = &write_result {
+            error!(
+                "Batch loop failed ({}), closing writer to release resources",
+                e
+            );
+            if let Err(close_err) = writer.close().await {
```
mattp5657, line 170: Same thing here as above comment: 

> On a write failure we're possibly leaving orphaned files in the Iceberg table location. close() still finalizes whatever partitions already succeeded, but we never commit them and just throw away the result. Could we grab the files from close()'s return and delete them before returning the error?

mattp5657 (commented):
@EdgarModesto23 I'm still pretty new here, but wanted to jump in and help review. Let me know if you agree with the comments below.

Also think we should probably add some regression testing here.

core/connectors/sinks/iceberg_sink/src/router/mod.rs
EdgarModesto23: You are correct! This has been noted before as seen in https://github.com/apache/iggy/pull/3194#discussion_r3273392351 and it's currently documented under https://github.com/apache/iggy/blob/master/core/connectors/sdk/src/lib.rs#L429. It's not an issue from this PR specifically so I wouldn't block for it, tho we would love to see this fixed either here or on a separate issue for sure :) Thank you for bringing this up! ❤️

core/connectors/sinks/iceberg_sink/src/router/mod.rs
ashudeqode: well  in iceberg data warehouse we have to run compaction for tables as streaming data make tons of parquet files so we have to run compaction time to time which combines multiple files and leave the other files as it is and then we have to expire snapshots and then remove the orphaned file time to time so it is a three step process so this will cover in it but in this PR our main concern is that the data should go in correct partition

hubcio:
could you please rebase and regenerate lockfile?

hubcio:
since author of this PR is inactive i'm closing this. fix itself is valuable so I ported it to a fresh one with the review feedback applied (shared write path, cleanup on write failure, regression tests). continued in #3969

PR #3870, opened 2026-08-12
fix(consensus): close metadata and consensus plane TODOs

krishvishal:
Clears ten `TODO` markers across the metadata, consensus, and shard planes. One commit per item.

## Correctness

**Roll back the sequencer on a WAL append failure.** `push_prepare_entry` claims an op before the journal append, so a failed append left the sequencer one ahead of the WAL forever and the next prepare chained over a hole no repair path refills. `rollback_pipelined_prepare` hands the op back and drops the pipeline entry. Refused when a sibling prepare was already projected off the failed op, since unwinding there would hand a live op number out twice.

**Snapshot format versioning.** `version` was a hardcoded `1` with no read-side check. Now exact-equality with fail-stop, plus a `release_format` provenance stamp, TigerBeetle's pair. The version is peeked off the msgpack prefix before the payload is deserialized, so a foreign layout is refused by version rather than by whichever field misparsed. Covers boot recovery and both state-transfer paths.

## Performance

**Zero-alloc op construction in STM apply.** `Bytes::copy_from_slice` per op body, then borrowed as `&[u8]` and dropped. `WireDecode::decode_from` takes a slice, so the `Bytes` bought nothing. Decodes off the backing buffer via a new `Message::body()`. Also drops a 256-byte header copy per op.

**One frame parse per hop.** Frames were classified twice, once in the router for routing and once in the shard for dispatch. `ShardFrame::Consensus` carries the `MessageBag`; `MessageBag::routing()` / `into_generic()` replace the router's 60-line `extract_routing`. `ShardFrame` stays 160 B, pinned by a `const` assert.

## Structure

**Stop leaking `&RefCell<Pipeline>`.** 49 call sites move to named accessors plus `with_pipeline` / `with_pipeline_mut`, whose `FnOnce` bound makes a borrow-across-`.await` unrepresentable. Same shape as `IggyPartitions::with_partition`.

**Prepare-timeout lifecycle.** "Ticking iff the pipeline is non-empty" now holds: `sync_prepare_timeout` disarms on empty and restarts otherwise, so a remaining head is timed from now instead of inheriting the drained entry's elapsed ticks. `init()` no longer arms it; a fresh primary has nothing to retransmit.

**One result-section framing.** `build_login_register_reply` and `build_raw_pat_reply` each hand-assembled `[u32 count=0][payload]`; the raw-PAT reply shipped once without the prefix and broke SDK decoding. One helper is now the only way to emit a success reply with a body.

**Stamp request-queue arrival.** `received_at` populated at park from `clock_realtime_micros()`, not `next_monotonic_timestamp()`. The latter advances the prepare-stamping sequence, so parking a request would perturb timestamps replicated to every backup.

**Fold the duplicated view-change entry.** Three sites inlined the same 12-line sequence, differing only in `reason`. Extracted to `enter_view_change`.

## Tests

**Pump realloc under a live partition borrow.** The borrow tripwire and deterministic executor already covered a concurrent `remove`; this adds the realloc half, where a growing `push` relocates every element. Guarded against passing vacuously by a new `buffer_addr()` and an `assert_ne!` across the insert, and falsification-checked. Swapping the bad read for `with_partition` makes it fail with "tripwire did not fire".

Also: a real-`PrepareJournal` append-failure rollback test, four rollback-outcome tests, two timer-lifecycle tests, a park-stamp test, snapshot version/peek/refusal tests, and a login-register framing test.


core/consensus/src/impls.rs
```diff
@@ -1484,6 +1659,47 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+    /// than colliding with a live entry. Dropping the entry drops its reply
+    /// sender, so a waiting client observes `Canceled` instead of hanging until
+    /// the request times out.
+    ///
+    /// The observed prepare timestamp is deliberately NOT rolled back:
+    /// [`Self::next_monotonic_timestamp`] only ever needs a lower bound, and
+    /// lowering it back could re-stamp a value a peer already observed.
+    ///
+    /// See [`PrepareRollback`] for the outcomes.
+    pub fn rollback_pipelined_prepare(&self, header: &PrepareHeader) -> PrepareRollback {
```
numinnex, line 1724: **[blocking]** `rollback_pipelined_prepare` runs after the append await with no view or status fence, and the tail match is by op alone. If a view change completes while the append is parked and this replica is re-elected, `start_pending_view` rewinds the sequencer to the merged head, so a new request can re-project the same op number with a different checksum. The old append's failure then passes both guards here: it pops the new view's live entry, rewinds the sequencer and parent chain under it, the live entry's acks die at `pipeline_holds_entry`, the op number is minted again, and backups panic in `panic_if_hash_chain_would_break_in_same_view`. Suggest refusing unless `header.view == self.view()` and status is `Normal`, and matching the tail by `(op, checksum)`.

core/consensus/src/impls.rs
```diff
@@ -1484,6 +1659,47 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+        if !self.is_primary() {
+            return PrepareRollback::NotPreAdvanced;
+        }
+        let sequence = self.sequencer.current_sequence();
+        if sequence != header.op {
+            return PrepareRollback::Overtaken { sequence };
+        }
+        let removed = self.pipeline.borrow_mut().remove_tail(header.op);
+        debug_assert!(
```
numinnex: **[blocking]** In release builds a `None` from `remove_tail` falls through this `debug_assert` and the sequencer and parent-chain rewind below still run, mutating state exactly when the invariant they rely on is violated. Returning a refusal on `None` without mutating keeps release behavior aligned with the assertion.

core/metadata/src/stm/snapshot.rs
```diff
@@ -15,12 +15,52 @@
...
+///
+/// Bumping it invalidates every `snapshot.bin` already on disk, and a node whose
+/// snapshot is refused refuses boot. That is deliberate. The metadata plane is
+/// pre-production, so the cost is clearing a data directory; once it ships, a bump
+/// needs an explicit translation path added here alongside it.
+///
+/// Version 2: `status` sits at reply-header offset 216 (version 1 carried a
+/// `namespace` word before it), which the client table's cached replies embed as raw
+/// wire bytes msgpack cannot introspect.
+pub const SNAPSHOT_FORMAT_VERSION: u32 = 2;
```
numinnex, line 47: **[blocking]** The serialized shape changes in this PR (`release_format` appended, `#[serde(default)]` dropped from `client_table`) but the version stays 2, and master already writes 6-element version-2 files. On the first cross-build boot the peek passes and `rmp_serde` fails with an invalid-length error, so the refusal surfaces as `SnapshotError::Deserialize`, the corruption-shaped diagnosis this gate exists to prevent. Both upgrade and downgrade directions misreport. This violates the bump rule documented above; bump to 3.

core/metadata/src/impls/metadata.rs
```diff
@@ -1198,28 +1198,35 @@ where
...
                 "journal append failed"
             );
+            if let PrepareRollback::Overtaken { sequence } = rollback {
+                error!(
+                    target: "iggy.metadata.diag",
+                    plane = "metadata",
+                    replica_id = consensus.replica(),
+                    op = header.op,
+                    sequence,
+                    "journal append failed after a sibling prepare was projected off this op; \
```
numinnex: **[warning]** Nothing triggers the view change this log waits on. The failed op was never broadcast, backups gap-drop the sibling at the hole check, and this primary keeps sending Commit heartbeats that reset the backups' election timers, so the group wedges permanently while looking healthy. A solo group has no peer at all. The wedge itself predates this PR (any append failure did this), but the log text is new and promises repair that cannot come. Either escalate here (cede primaryship or fail-stop, per the removed TODO) or reword to state that no election will follow without operator action.

core/partitions/src/iggy_partition.rs
```diff
@@ -1996,11 +1996,9 @@ where
...
                 // request room -> buffer; both full -> drop+warn (client retries
                 // via read-timeout).
-                if consensus.pipeline().borrow().is_full() {
-                    let push_result = consensus
-                        .pipeline()
-                        .borrow_mut()
-                        .push_request(consensus::RequestEntry::new(message));
+                if consensus.pipeline_is_full() {
+                    let push_result =
+                        consensus.push_queued_request(consensus::RequestEntry::new(message));
```
numinnex, line 2001: **[warning]** Parity gap: the partitions-plane primary has the same pre-advance in `push_prepare_entry`, but its append-failure arm still warns and returns without `rollback_pipelined_prepare`, leaving the sequencer ahead of the WAL on the data plane. The new rollback is plane-agnostic and the failure arm has no half-applied state, so mirroring the metadata call looks safe. Fine as a follow-up since it predates this PR, but worth a tracking issue before merge.

core/consensus/src/impls.rs
```diff
@@ -1418,17 +1493,117 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+    /// the current oldest entry from now. Exposed for the plane-side drains that
+    /// pop through [`Pipeline`] directly (`drain_committable_prefix`) rather than
+    /// through [`Self::pop_committed_prepare`].
+    pub fn sync_prepare_timeout(&self) {
+        let empty = self.pipeline.borrow().is_empty();
+        let mut timeouts = self.timeouts.borrow_mut();
+        if empty {
+            timeouts.stop(TimeoutKind::Prepare);
+        } else {
+            timeouts.reset(TimeoutKind::Prepare);
```
numinnex: **[nit]** `Timeout::reset` never sets `ticking`, so on a stopped timer with a non-empty pipeline this branch leaves it armed-but-dead and the doc's "restarts it" does not hold. Every current caller happens to be preceded by a start-arm (`advance_commit_max`, become-primary), but the bootstrap re-pipeline gap flagged separately would land on this branch. `start` when not ticking, `reset` otherwise; same trap `advance_commit_max` documents.

core/consensus/src/impls.rs
```diff
@@ -1484,6 +1659,47 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
+        let removed = self.pipeline.borrow_mut().remove_tail(header.op);
+        debug_assert!(
+            removed.is_some(),
+            "sequencer at op {} but the pipeline tail is not that op",
+            header.op
+        );
+        drop(removed);
+        self.sequencer.set_sequence(header.op.saturating_sub(1));
+        self.set_last_prepare_checksum(header.parent);
+        PrepareRollback::Unwound
```
numinnex, line 1756: **[nit]** The `Unwound` path pops the tail without `sync_prepare_timeout`, so unwinding the sole in-flight entry leaves the Prepare timer ticking on an empty pipeline, against the invariant this PR introduces. The backstop in `handle_prepare_timeout` caps the cost at one spurious fire.

core/metadata/src/impls/recovery.rs
```diff
@@ -1672,47 +1673,63 @@ mod tests {
...
+                assert_eq!(found, SNAPSHOT_FORMAT_VERSION + 1);
+                assert_eq!(expected, SNAPSHOT_FORMAT_VERSION);
+            }
+            Err(other) => panic!("expected a format-version refusal, got {other}"),
+            Ok(_) => panic!("expected a foreign format version to refuse boot"),
+        }
+    }
+
+    #[compio::test]
+    async fn recover_pairs_the_checkpoint_against_the_bytes_on_disk() {
```
numinnex, line 1719: **[nit]** This test writes canonical bytes, which re-encode byte-identically (the byte-stability tests pin that), so the recomputed checksum equals the on-disk one under either checksum source and the test passes even if recovery re-encoded the decoded snapshot. Feeding a noncanonical but decodable file, for example the version encoded as a uint32 marker (accepted by `read_int`, canonicalized on re-encode), restores the discrimination the deleted test had.

core/server/src/bootstrap.rs
```diff
@@ -2323,20 +2323,21 @@ fn restore_metadata_consensus(
...
-                    op,
-                    "recovered journal suffix has a gap; stopping re-pipeline"
-                );
-                break;
-            };
-            let mut entry = PipelineEntry::new(*header);
-            entry.add_ack(topology.self_replica_id);
-            pipeline.push(entry);
-        }
+        consensus.with_pipeline_mut(|pipeline| {
```
numinnex, line 2363: **[nit]** This block pushes recovered entries without arming the Prepare timeout, and `init` no longer arms it. The branch is unreachable today (clustered restarts cede primaryship, solo forces the watermark to the restored op), but if it reactivates the recovered suffix never retransmits. Needs a `sync_prepare_timeout` here, which only works once the reset-vs-start issue in that helper is fixed.

core/consensus/src/impls.rs
```diff
@@ -310,6 +320,42 @@ impl RequestEntry {
...
+
+/// Outcome of [`VsrConsensus::rollback_pipelined_prepare`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PrepareRollback {
+    /// Sequencer, parent chain, and pipeline entry all restored. The op is free
+    /// again and the next request reuses it.
+    Unwound,
+    /// This replica never pre-advanced for the prepare, so there is nothing to
+    /// undo. A backup advances only AFTER its append succeeds.
+    NotPreAdvanced,
```
numinnex, line 365: **[nit]** Both labels overclaim. `NotPreAdvanced` is returned for any non-primary, including a demoted ex-primary that did pre-advance (the view-change reset erased it, so the name is operationally true but literally false). `Overtaken` also covers `sequence < header.op` after a state-transfer rewind, where the sibling story in this doc and in the caller's log does not apply. Worth documenting both directions or branching the log text.

core/server/src/responses.rs
```diff
@@ -1303,35 +1303,55 @@ pub(crate) fn build_deny_reply(
...
+/// Build a metadata reply carrying `payload` behind a success result section.
+///
+/// Every metadata reply is result-framed, and the SDK strips that section off
+/// unconditionally (`split_metadata_result`) before decoding what follows. A
+/// payload written without the leading zero count therefore has its first four
+/// bytes eaten as a result count, and the decode fails or, worse, succeeds on the
+/// shifted remainder. That is not hypothetical: the raw-PAT reply shipped once
+/// without the prefix and broke SDK decoding. This is the only way to emit a
+/// success reply with a body, so the prefix cannot be forgotten again.
+fn build_result_framed_reply(
```
numinnex, line 1320: **[nit]** Two overstatements in the doc: the SDK strips the result section only for `is_result_framed()` operations (plus a non-empty Register), not every metadata reply, and `build_reply_from_bytes` still emits success read replies with bodies, so this is not the only path. As written it could lead someone to result-frame a read reply and break decoding in the other direction. Minor: the payload could encode straight into `out[RESULT_COUNT_LEN..]` instead of a temporary `BytesMut`, though this path is cold.

core/consensus/src/impls.rs
```diff
@@ -265,10 +265,20 @@ impl PipelineEntry {
...
+    ///
+    /// Read against `clock_realtime_micros` at promotion for the queue wait:
+    /// what age-based shedding would filter on, and the queueing half of
+    /// end-to-end commit latency.
+    ///
+    /// Deliberately the plain clock read rather than
+    /// [`VsrConsensus::next_monotonic_timestamp`]: this must not consume the
+    /// prepare-stamping monotonic sequence, or parking a request would perturb
+    /// the timestamps replicated to every backup.
+    pub received_at: u64,
```
numinnex, line 281: **[nit]** The documented reader pattern `clock_realtime_micros() - received_at` on `u64` underflows if the realtime clock steps back between park and promotion (the old `i64` kept the sign). No reader exists yet; worth stating the `saturating_sub` requirement here before one lands.

core/server_common/src/consensus_message.rs
```diff
@@ -555,6 +573,76 @@ pub enum MessageBag {
...
 
 impl MessageBag {
+    /// `(operation, group)`: everything the shard router needs to pick a
+    /// target, read off the already-typed header without consuming the bag.
+    ///
+    /// `group` is a plain field on every consensus header rather than a
+    /// [`ConsensusHeader`] method, which is why this is a match and not a trait
+    /// call. `RepairPrepare` reads through its wrapped prepare.
+    #[must_use]
+    pub fn routing(&self) -> (Operation, u64) {
```
numinnex, line 598: **[nit]** `header()` runs a checked bytemuck cast per call and each arm here reads it twice; the Request and Prepare arms in `on_message` then re-derive the same tuple, so a hot frame pays four casts where one `let header = message.header();` per arm suffices. Same pattern in `body()`, which re-reads the header for `size` right after the STM macro's scalar read.

core/metadata/src/stm/snapshot.rs
```diff
@@ -15,12 +15,52 @@
...
+/// renders it as `major.minor.patch`.
+///
+/// Provenance, never a gate. The format version says whether the bytes are
+/// readable; this says which build produced them, which matters most for a snapshot
+/// that arrived over state transfer from a machine whose build this node otherwise
+/// has no record of.
+///
+/// A build constant, identical on every replica, so two replicas holding identical
+/// state still serialize identically (see [`MetadataSnapshot`]).
+pub const SNAPSHOT_RELEASE_FORMAT: u32 = IGGY_PROTOCOL_VERSION;
```
numinnex: **[nit]** The name says format but the content is the packed release semver, and it sits next to a real format version; the doc has to argue against the confusion the name creates. `writer_release` or similar would remove the ambiguity.

core/partitions/src/iggy_partitions.rs
```diff
@@ -301,6 +301,22 @@ where
...
+    ///
+    /// Lets a test prove that an [`Self::insert`] actually REALLOCATED rather
+    /// than landing in spare capacity. The distinction is the whole point of the
+    /// realloc half of PR #3557: `swap_remove` invalidates one slot's reference,
+    /// while a growing `push` moves every element and invalidates all of them. A
+    /// test that only checked "insert happened" would pass on a push into spare
+    /// capacity, which moves nothing and proves nothing.
+    #[cfg(any(test, feature = "simulator"))]
+    #[must_use]
+    pub fn buffer_addr(&self) -> usize {
```
numinnex, line 314: **[nit]** The safety comment argues read-only-ness, but the actual soundness precondition is that no pump-side `&mut` is live when this forms a `&Vec` through the `UnsafeCell`; a shared reborrow aliasing a live `&mut` is UB regardless of whether it only reads. Current callers uphold it; worth stating directly.

core/consensus/src/impls.rs
```diff
@@ -3428,8 +3595,13 @@ impl<B: MessageBus, P: Pipeline<Entry = PipelineEntry>> VsrConsensus<B, P> {
...
-    /// Currently only `PrepareOk` messages are routed here (via `send_or_loopback`).
-    // TODO: Route SVC/DVC self-messages through loopback once VsrAction dispatch is implemented.
+    /// Only `PrepareOk` reaches here (via `send_or_loopback`), and deliberately:
+    /// it is a message to a peer that happens to be this replica, so a one-drain
+    /// delay costs nothing. A replica's own SVC/DVC is not that shape -- it is the
+    /// local decision to change view, recorded synchronously with the view and
+    /// status writes in [`Self::enter_view_change`]. Routing it here instead would
+    /// leave a window where the replica has entered a view change without counting
+    /// itself, which on a solo group is the entire quorum.
     pub(crate) fn push_loopback(&self, message: Message<GenericHeader>) {
```
numinnex, line 3664: **[nit]** Pre-existing, same class this PR removes on the router hop: `send_or_loopback` erases the self-addressed PrepareOk to generic and the drain re-types it with a full frame-checksum verify per self-ack, once per op per plane on the primary. Typing the loopback queue as `VecDeque<Message<PrepareOkHeader>>` would close it.

core/consensus/src/impls.rs
krishvishal: Done.

core/consensus/src/impls.rs

core/metadata/src/impls/metadata.rs

core/consensus/src/impls.rs

core/consensus/src/impls.rs
krishvishal: Done.

core/metadata/src/impls/recovery.rs
krishvishal: Done.

core/server/src/bootstrap.rs
krishvishal: Done.

core/consensus/src/impls.rs
krishvishal: Done.

core/server/src/responses.rs
krishvishal: Both doc overstatements fixed.

core/consensus/src/impls.rs
krishvishal: Done, as an API rather than a documented requirement, so the reader cannot get it wrong: `RequestEntry::queue_wait_micros(now)` does the `saturating_sub`. The field doc now points at it and says not to subtract directly.

core/metadata/src/stm/snapshot.rs
krishvishal: Done. `release_format` -> `writer_release`, `SNAPSHOT_RELEASE_FORMAT` -> `SNAPSHOT_WRITER_RELEASE`. This is TB's pairing too: `SuperBlockHeader` carries `version: u16` next to a plainly named `release`.

Free to do: msgpack encodes the struct positionally, so serde field names never reach the wire. No shape change, no bump. The doc no longer has to argue against the name.

core/partitions/src/iggy_partitions.rs
krishvishal: Done. The comment now states the aliasing precondition directly: read-only-ness is not what makes it sound, since a shared reborrow aliasing a live `&mut` is UB whether or not it reads. What makes it sound is that no `&mut` from `get_mut_by_ns` / `namespace_map_mut` is live across the call, those being pump-only and these callers running on the pump.

core/server_common/src/consensus_message.rs
krishvishal: Partly done, leaving this open for the rest.

Done: one `header()` per arm bound to a local, across all 15 arms of `MessageBag::routing()` and both the Request and Prepare arms in the shard dispatch (which now hoist rather than re-deriving the tuple).

Not done: `body()`, and the underlying cause. Fixing `body()` cleanly means either reintroducing the manual slicing it was created to remove or threading `size` through, both worse than one cast.

numinnex (approved):

spetz (approved):

PR #3872, opened 2026-08-12
feat(server)!: auth-gate cluster metadata, forward register to primary

hubcio:
Unauthenticated clients could read the full cluster roster through
the binary GetClusterMetadata pre-auth carve-out, leaking private
network topology. Plain gating would strand any client that dialed
a backup: Register commits only on the primary, and without the
pre-auth read there is no leader discovery left, so login replays
transient denies until the read budget dies.

Gate the read on every roster shape (PING stays the only pre-auth
code) and let the dialed backup complete the login instead:
credentials are verified locally against replicated user state,
only the verified identity (client id, user id) crosses the
replica interconnect as a sealed ForwardRegister frame, the
primary commits the Register, and the backup binds the session
and replies on its own connection. No credentials travel on the
wire, no reply routing lands on the primary, and the HTTP
forward's fail-closed semantics (an unreplicated PAT is denied at
the backup) carry over; a forward that never left the backup maps
to the retryable class, not a client-side 503. An unbound reader
is denied with a plain Unauthenticated Reply on the status
channel, never a session-terminal Eviction, so a foreign SDK's
stray probe cannot tear down the connection its login is about to
use. The Rust and Node SDKs drop their pre-login leader probe;
the authenticated post-login redirect is the sole leader
settlement, which Node gains here with a short poll through an
election. The C++ e2e expectation of the pre-auth read inverts to
match.

Kill-primary testing of the forward exposed a pre-existing
view-start bug: the metadata parked-view arm dispatched
RebuildPipeline behind the superblock persist, so a login admitted
during that fsync minted an op into the still-empty pipeline and
the deferred rebuild panicked the pump ("sequence must be
sequential"), leaving a deaf primary holding its quorum slot while
the survivor spun elections forever. Dispatch the local actions
before the persist, as every sibling site already does. The drain
path now surfaces the pump's panic verdict instead of claiming a
clean exit, the harness fails any test whose server wrote a panic
to stderr, a simulator test pins the interleaving
deterministically, and forward failures no longer masquerade as a
view-change cancel.

### DIFF
--- a/core/binary_protocol/src/consensus/command.rs
+++ b/core/binary_protocol/src/consensus/command.rs
@@ -70,6 +70,13 @@ pub enum Command2 {
     StateTransferTarget = 23,
     RequestStateChunk = 24,
     StateChunk = 25,
+
+    // Register forwarding: a client that dialed a backup authenticates
+    // there, and only the consensus proposal travels to the primary. The
+    // backup forwards the verified identity and parks the login on the
+    // matching `ForwardRegisterResult`.
+    ForwardRegister = 26,
+    ForwardRegisterResult = 27,
 }
 
 // SAFETY: Command2 is #[repr(u8)] with no padding bytes.
@@ -80,7 +87,7 @@ unsafe impl CheckedBitPattern for Command2 {
     type Bits = u8;
 
     fn is_valid_bit_pattern(bits: &u8) -> bool {
-        *bits <= Self::StateChunk as u8
+        *bits <= Self::ForwardRegisterResult as u8
     }
 }
 
@@ -101,8 +108,8 @@ mod tests {
 
     #[test]
     fn replica_auth_commands_are_valid_bit_patterns() {
-        // Locks the is_valid_bit_pattern bump: 14..=25 parse, 26 still rejects.
-        for command in 14u8..=25 {
+        // Locks the is_valid_bit_pattern bump: 14..=27 parse, 28 still rejects.
+        for command in 14u8..=27 {
             let mut buf: AVec<u8, ConstAlign<16>> = AVec::new(16);
             buf.resize(256, 0);
             buf[60] = command;
@@ -110,7 +117,7 @@ mod tests {
         }
         let mut buf: AVec<u8, ConstAlign<16>> = AVec::new(16);
         buf.resize(256, 0);
-        buf[60] = 26;
+        buf[60] = 28;
         assert!(bytemuck::checked::try_from_bytes::<GenericHeader>(&buf).is_err());
     }
 }
--- a/core/binary_protocol/src/consensus/command.rs
+++ b/core/binary_protocol/src/consensus/command.rs
@@ -77,6 +77,12 @@ pub enum Command2 {
     // matching `ForwardRegisterResult`.
     ForwardRegister = 26,
     ForwardRegisterResult = 27,
+
+    // Logout forwarding: a session bound on a backup asks the primary to
+    // commit its replicated teardown, then the backup answers the client on
+    // the connection it owns.
+    ForwardLogout = 28,
+    ForwardLogoutResult = 29,
 }
 
 // SAFETY: Command2 is #[repr(u8)] with no padding bytes.
@@ -87,7 +93,7 @@ unsafe impl CheckedBitPattern for Command2 {
     type Bits = u8;
 
     fn is_valid_bit_pattern(bits: &u8) -> bool {
-        *bits <= Self::ForwardRegisterResult as u8
+        *bits <= Self::ForwardLogoutResult as u8
     }
 }
 
@@ -108,8 +114,8 @@ mod tests {
 
     #[test]
     fn replica_auth_commands_are_valid_bit_patterns() {
-        // Locks the is_valid_bit_pattern bump: 14..=27 parse, 28 still rejects.
-        for command in 14u8..=27 {
+        // Locks the is_valid_bit_pattern bump: 14..=29 parse, 30 still rejects.
+        for command in 14u8..=29 {
             let mut buf: AVec<u8, ConstAlign<16>> = AVec::new(16);
             buf.resize(256, 0);
             buf[60] = command;
@@ -117,7 +123,7 @@ mod tests {
         }
         let mut buf: AVec<u8, ConstAlign<16>> = AVec::new(16);
         buf.resize(256, 0);
-        buf[60] = 28;
+        buf[60] = 30;
         assert!(bytemuck::checked::try_from_bytes::<GenericHeader>(&buf).is_err());
     }
 }
--- a/core/binary_protocol/src/consensus/header.rs
+++ b/core/binary_protocol/src/consensus/header.rs
@@ -32,6 +32,12 @@
 //! mixed pair is dropped, so no view change reaches a quorum. Nothing enforces this,
 //! because there is nothing left to enforce it with; this note is the declaration.
 //!
+//! The forwarding commands at discriminants 26 through 29 are same-release
+//! replica messages. A build whose command bit-pattern predates them drops the
+//! frames as unparsable and never sends a result, so backup-dialed session
+//! operations wait for their forward timeout in a mixed fleet. These commands are
+//! another reason the release cannot be rolled node by node.
+//!
 //! `Prepare`, `Request`, `Reply`, and `Eviction` are unaffected. Prepares keep
 //! `checksum` as their view-independent identity, and the three client-facing
 //! headers are sealed on neither side, so SDKs are untouched.
@@ -2429,7 +2435,13 @@ impl ConsensusHeader for ForwardRegisterResultHeader {
                 found: self.command,
             });
         }
-        validate_forward_register_frame(self.size, self.client, self.nonce, &self.reserved)
+        validate_forward_register_frame(self.size, self.client, self.nonce, &self.reserved)?;
+        if self.outcome != ForwardRegisterOutcome::Ok && (self.epoch != 0 || self.watermark != 0) {
+            return Err(ConsensusError::InvalidField(
+                "forward register result bind must be zero on failure".to_string(),
+            ));
+        }
+        Ok(())
     }
 }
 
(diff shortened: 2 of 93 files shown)


core/server/src/dispatch.rs
```diff
@@ -2386,15 +2413,267 @@ fn polling_strategy_from_wire(
...
+///   answers 401 until replication catches up rather than relaying an
+///   unverified bearer.
+/// - `ClientIdOwnedByAnotherUser` stays a decision of the caught-up primary
+///   and round-trips as a terminal refusal.
+///
+/// The session binds here before this node applies the commit locally. That
+/// is the window a primary-side login already has against every other node's
+/// apply lag, not a new one.
+#[allow(clippy::future_not_send)]
+async fn submit_register_local_or_forward<B, MJ, S, SB>(
```
numinnex, line 2531: Register gained a forward twin here, but Logout did not: on a backup, `submit_logout_on_owner` still calls `submit_logout_in_process` directly and the primary gate answers `NotPrimary` (core/metadata/src/impls/metadata.rs:2107-2109). An explicit logout at a backup then loops on `TransientNotAccepted` replays, and a disconnect logout is dropped with a warn (dispatch.rs:3026-3035).

Since the SDK redirect mints a fresh client id per re-login (core/sdk/src/session.rs:141-145), every backup-dialed auto-login, which is this PR's flagship flow, orphans one replicated client-table entry until capacity eviction. `evict_oldest` picks by oldest commit regardless of liveness, so the orphans can push idle live sessions into eviction, which erases their watermark and re-opens the at-most-once caveat documented at client_table.rs:862-864.

Consider forwarding Logout the same way as Register; the epoch guard (metadata.rs:2097) already makes a forwarded logout safe against stale replay.

core/server/src/auth.rs
```diff
@@ -356,6 +359,30 @@ async fn send_login_transient_reply<B, MJ, S, SB>(
...
+/// pins the replay to this connection and its client id, where a register
+/// that did commit rebinds its own client-table entry; re-issuing under a
+/// freshly minted id would instead orphan that entry until capacity eviction
+/// reclaims it.
+const fn transient_login_code(error: &LoginRegisterError) -> IggyError {
+    match error {
+        LoginRegisterError::Transient(MetadataSubmitError::ForwardTimedOut) => {
+            IggyError::TransientNotCommitted
+        }
+        _ => IggyError::TransientNotAccepted,
```
numinnex, line 382: `Transient(InProgress)` and `Transient(Canceled)` ride this `_` arm, but the contract above ("nothing was committed: the register never entered a pipeline") does not hold for either, since both can follow pipeline admission.

The sequence is routine with the forward path: a forward times out while the register stays parked on the primary with no deadline, the client replays, the primary dedup answers `InProgress`, and this arm un-pins the client id that `TransientNotCommitted` pinned one replay earlier. A conforming client may then fail over under a fresh identity while the parked register later commits, orphaning the entry. The HTTP path classifies exactly these two as unknown-outcome (core/server/src/http/state.rs:346-352).

Mapping `InProgress` and `Canceled` to `TransientNotCommitted` here would make the doc true as written. Latent for the Rust SDK today, which excludes login codes from failover (core/sdk/src/tcp/tcp_client.rs:652-654), but wrong for any conforming client.

core/binary_protocol/src/consensus/command.rs
```diff
@@ -70,6 +70,13 @@ pub enum Command2 {
     StateTransferTarget = 23,
     RequestStateChunk = 24,
     StateChunk = 25,
+
+    // Register forwarding: a client that dialed a backup authenticates
+    // there, and only the consensus proposal travels to the primary. The
+    // backup forwards the verified identity and parks the login on the
+    // matching `ForwardRegisterResult`.
+    ForwardRegister = 26,
+    ForwardRegisterResult = 27,
```
numinnex, line 79: Rolling-upgrade gap: an already-deployed replica whose `is_valid_bit_pattern` caps at 25 drops these two frames as unparsable, and this pair has no retransmit (core/shard/src/router.rs:263-271), so backup-dialed logins fail through 5s forward timeouts for the whole mixed-fleet window while the cluster otherwise looks healthy. The wire-compatibility note at the top of header.rs covers only the seal transition, not this pair.

At minimum the release notes should state that replicas must be upgraded together for backup-dialed logins to keep working; the version fence the router.rs TODO asks for would close it properly.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -162,6 +164,10 @@ impl BinaryTransport for TcpClient {
             return Err(error);
         }
 
+        if is_unauthenticated_metadata_probe(code, &error) {
```
numinnex: This guard sits below the `!reconnection.enabled` early return above, so a pre-login `get_cluster_metadata` with reconnection disabled reports `Disconnected` while the connection is healthy, instead of the `Unauthenticated` fail-fast that leader_aware.rs:30-36 promises uniformly for every transport. Hoisting this check above the reconnection check fixes it. Same ordering in quic_client.rs and websocket_client.rs.

core/server/src/dispatch.rs
```diff
@@ -2386,15 +2413,267 @@ fn polling_strategy_from_wire(
...
 }
 
+/// Answer a backup's forwarded `Register` from the node it named primary.
+///
+/// Proposes in process, never through [`submit_register_local_or_forward`]:
+/// that is what bounds a forward at one hop. A node that has since lost
+/// primaryship answers `NotPrimary`, and the origin's client replays against
+/// whichever node it names next.
+#[allow(clippy::future_not_send)]
+async fn answer_forwarded_register<B, MJ, S, SB>(
```
numinnex, line 2438: The primary proposes the forwarded register with no credential-freshness check: only the client-id ownership gate and the dedup run in `submit_register_in_process` (core/metadata/src/impls/metadata.rs:1849-1970). A PAT revoked (or a user deactivated) at the primary therefore still authenticates at a backup whose replicated users table lags, and the commit proceeds.

The window is bounded by apply lag (a backup partitioned from the primary cannot complete the forward at all), and the HTTP forward already has the same semantics, but this PR extends it to the binary transports. Worth documenting the window explicitly, or re-verifying user status at register preflight on the caught-up primary.

core/shard/src/lib.rs
```diff
@@ -184,6 +184,23 @@ pub enum MetadataSubmit {
         /// client a retry storm of full password verifications.
```
numinnex, line 187: Related to the storm this comment names: `verify_login_credentials` runs Argon2id synchronously on the shard reactor before consensus admission, and re-runs on every transient replay. This PR makes the bounce paths routine, since clients now dial backups by design and a mid-election login bounces `NotPrimary`.

Pre-existing ordering, so a follow-up is fine: bounce on `!is_normal()` before verifying (no credential bytes are touched, so no timing oracle, and a healthy backup must still verify for the forward path), or move the verify off the reactor thread.

foreign/node/src/client/client.socket.ts
```diff
@@ -385,38 +403,88 @@ export class CommandResponseStream extends EventEmitter {
     });
   }
 
+  // `GetClusterMetadata` is deliberately absent: the server auth-gates it,
+  // so the client authenticates before reading the topology. A login dialed
```
numinnex, line 408: Compat note for the release: the previous Node client awaited `_ensureVsrLeader` inside `sendCommand(LOGIN)` with no catch, and its metadata read rode the unauthenticated carve-out. Against a server carrying this PR the read is denied and the rejection propagates, so every login from the old Node SDK fails outright. The other old SDKs degrade to a warn plus a dead round trip. Release notes should state: upgrade the Node SDK before the server.

foreign/node/src/client/client.socket.ts
```diff
@@ -385,38 +403,88 @@ export class CommandResponseStream extends EventEmitter {
...
+   * Only the leader accepts replicated commands, and the roster read is
+   * auth-gated, so the topology cannot be inspected before a login binds a
+   * session. The redirect drops that session along with the socket, so the
+   * login is replayed on the leader and its answer supersedes the one from the
+   * node the client dialed. One hop is enough: the leader admits logins
+   * locally.
+   *
+   * @returns The leader's login response, or undefined when the client stays
+   */
+  private async _settleOnLeader(
```
numinnex, line 428: Single hop, and `settlingLeader` suppresses a re-settle on the replayed login: if leadership moves between the roster read and the redirect, the client stays on a backup, and `_processVsr` has no data-path leader redirect, so later writes burn their full budgets before failing. Narrow window and a reconnect heals it; noting for a follow-up since the Rust side budgets 3 hops for the same settlement.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClient.java
```diff
@@ -651,10 +651,13 @@ CompletableFuture<IdentityInfo> loginOnLeader(Supplier<CompletableFuture<Identit
...
-     * the leader. Register is sent only after this bounded process settles.
+     * One discovery hop. When the roster names a healthy leader elsewhere,
+     * reconnect to it and re-check from the new node, since mid-election
+     * metadata can point at a node that is itself not the leader. Register is
+     * sent only after this bounded process settles. Reading the roster needs a
+     * bound session, so a connection that has none fails the fetch locally and
+     * stays where it is; the hop redirects for a relogin or a transient
+     * failover recheck, which run on a session.
      */
     private CompletableFuture<Void> redirectToLeader(LeaderRedirectionState redirectionState) {
```
numinnex, line 693: With the client-side auth gate (AsyncTcpConnection.java:632-634) the first-login `redirectToLeader` hop is now a guaranteed no-op, and no post-login settlement replaces it, so a client that dialed a backup stays there until the first transient failover replays auth. Placement heals lazily rather than never, but this diverges from the Rust and Node clients, which settle on the leader right after login. Consider running the hop again after the first successful login.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -2234,13 +2234,250 @@ impl ConsensusHeader for StateChunkHeader {
...
+    /// Echo of [`ForwardRegisterHeader::nonce`]; with `client`, the backup's
+    /// routing key.
+    pub nonce: u128,
+    /// Echo of [`ForwardRegisterHeader::client`]. Half of the routing key, not a
+    /// diagnostic: the backup drops an answer whose client disagrees with the
+    /// login it parked under `nonce`.
+    pub client: u128,
+    /// The committed register's op number, which fences the session. Zero
+    /// unless `outcome` is [`ForwardRegisterOutcome::Ok`].
+    pub epoch: u64,
```
numinnex, line 2361: The docs pin `epoch` and `watermark` to zero unless `outcome` is `Ok`, but `validate` never enforces it, unlike the strict-zero `reserved` bytes one field over, which the shared validator rejects as a builder bug. The receiver only reads these fields on the `Ok` arm so nothing breaks today; either enforce the claim in the result header's `validate` or drop the "Zero unless" wording.

core/server/src/bootstrap.rs
```diff
@@ -1449,25 +1453,54 @@ async fn shard_main(
...
-        .is_err()
-    {
+    let Ok(join_result) = compio::time::timeout(drain_budget, pump_handle).await else {
         warn!(
             shard = shard_id,
             "message pump did not drain within the shutdown budget; \
              committed journal tail may not have flushed"
         );
-    }
+        return Ok(());
```
numinnex: The timeout arm returns `Ok(())`, so shard_main still logs "server shard exited cleanly" right after the warning that the committed journal tail may not have flushed. Pre-existing shape, and this PR fixed the panic path, but now that the panic verdict is surfaced, a stuck-but-alive pump reporting a clean exit is the odd one out. A distinct non-clean log line or an `Err` variant for the timeout would close it.

core/server/src/dispatch/authz.rs
```diff
@@ -131,14 +131,14 @@ where
...
-/// frame: empty body + nonzero `status`. The nonzero status is the whole
+/// Reply to a request rejected before it reached its plane with the request's
+/// own frame: empty body + nonzero `status`. The nonzero status is the whole
 /// point: the SDK peeks it and surfaces the typed error, whereas a status-0
 /// frame reads as a committed ack for work that never happened. Silence is no
 /// better, the connection decodes replies in lockstep and would wedge on every
 /// later request.
 #[allow(clippy::future_not_send)]
-pub(super) async fn send_partition_deny_reply<B, MJ, S, SB>(
+pub(super) async fn send_deny_reply<B, MJ, S, SB>(
```
numinnex, line 141: `send_deny_reply` stamps the live `current_metadata_commit` into the deny, and with this PR that is the standard answer to any unauthenticated read, so pre-auth callers get a commit-height and write-rate oracle on the same PR that closes the roster leak. No SDK reads the commit field on denies (checked Rust, Node, C#, Go, Java), so zeroing it for unbound denies is compat-safe.

core/sdk/src/websocket/websocket_client.rs
```diff
@@ -149,6 +151,10 @@ impl BinaryTransport for WebSocketClient {
             return Err(error);
```
numinnex: Pre-existing asymmetry that this hunk makes more visible: the `AutoLogin::Disabled` arm above returns without the `!is_login_register_code(code)` exception that tcp and quic carry, so a manual-login WebSocket client cannot reconnect-replay a transient register. Same on master; noting it since this PR touches the surrounding block in all three transports.

spetz (approved):

numinnex (approved):

PR #3873, opened 2026-08-12
feat(connectors): add OpenSearch sink connector

### LINKED ISSUE #3504: feat(connector): Add Opensearch sink connector
### Description

Add a Opensearch sink connector that consumes Iggy messages and indexes them into a Opensearch index through the official Rust SDK.

### Motivation
Users should be able to route Iggy stream data into Opensearch for full-text search and retrieval without writing a custom connector.

### Scope
- Add the sink connector crate and sample runtime config.
- Document sink configuration and payload behavior.
- Add Docker-backed integration coverage for indexing JSON messages.

### Affected area / component

Connectors

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Contribution

### Good first issue

mattp5657:

Closes #3504

## Rationale

Iggy connectors ship sinks for several external systems but not for OpenSearch, a widely used search/analytics backend. This adds one, modeled on the existing `elasticsearch_sink` shape but closing a retry gap that sink still has (see Known trade-offs).

## What changed?

Adds `core/connectors/sinks/opensearch_sink/`, following the sink lifecycle end to end:

- **`open()`:** validates config (URL shape, credential pairing, `document_id_field` constraints), then `retry_on_open`-wraps a cluster health check, an index-exists check, and, when `create_index_if_not_exists` (default `true`), index creation with an optional custom mapping. Capped at `max_open_retries` with exponential backoff and jitter (shared `retry` helpers).
- **`consume()`:** batches incoming messages into `batch_size` chunks, builds an `iggy_*`-enriched document per message (hashed or field-derived `_id`), and hands each chunk to `index_chunk`.
- **`index_chunk()`:** POSTs a `_bulk` request and loops up to `max_retries` with the same backoff. `_bulk` answers `200` even when individual documents fail, so each response is parsed per item rather than trusting the top-level status: permanent failures (4xx, e.g. a mapping conflict) are recorded immediately, while transient ones (429/5xx) shrink the pending set to just the rejected documents and get resent, so a partial rejection under load doesn't re-index or lose the rest of the chunk. Counts from earlier attempts are merged into the outcome so a later failure doesn't erase already-indexed documents from the tally.
- **`close()`:** drops the client, no special teardown.

**Integration tests** (`core/integration/tests/connectors/opensearch/`) run against a real container (`testcontainers-modules`, reused across tests via `ReuseDirective::Always`, per-test-unique index names for isolation), covering the happy path plus a static mapping conflict, a missing index with index-creation disabled, and confirming a failing chunk doesn't block chunks queued behind it.

**Credentials:** HTTP Basic auth only (`username`/`password`, both-or-neither validated at config time); `password` is a `SecretString`, never logged or serialized. AWS SigV4 (AWS-managed OpenSearch / Serverless) is not supported.

**Known trade-offs**, deliberately out of scope here, verified against current `master`:
- `elasticsearch_sink` has the same gap this PR fixes for OpenSearch: `bulk_index_documents` (`elasticsearch_sink/src/lib.rs:205-219`) tallies per-item `_bulk` failures into `errors_count` but never retries the transient subset (e.g. 429 `es_rejected_execution_exception`). Worth a follow-up issue rather than folding into this PR.
- The connectors runtime discards a sink's `consume()` return value: `core/connectors/runtime/src/sink.rs:740-748` invokes the FFI `consume` callback as a bare statement, never binding its `i32` result, so `process_messages` always returns `Ok`. Combined with offsets auto-committing at poll time (`sink.rs:522`), a plugin-level failure never reaches connector status, `last_error`, or `/stats`, and the batch is never redelivered. Pre-existing, repo-wide, affects every sink.
- `meilisearch_sink` (`lib.rs:451-454`) and `elasticsearch_sink` (`lib.rs:319-328`) both silently drop `iggy_headers`/`_iggy_headers`: `BTreeMap<HeaderKey, HeaderValue>` can't serialize as a JSON object (`serde_json` requires string keys), and both sinks swallow that error via `if let Ok(...)` instead of surfacing it. `core/common` even ships a `serialize_headers` workaround for this exact case that neither sink uses.
- Single-node transport: like `elasticsearch_source`, the client is built on `SingleNodeConnectionPool` (`lib.rs:208`) with no cluster sniffing or multi-node failover. A dead configured node fails every request rather than routing around it.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. **Tools:** Claude
2. **Use** Used for scaffolding and iteration.
3. **Verification:** Full local compilation, extensive integration testing against a live container.
4. Reviewed and can explain every line if asked.


mattp5657:
Getting some errors with a 503 returned when they passed locally, for example for Typo. Will try running again later: 

```shell
  Connecting to github.com (github.com)|140.82.112.4|:443... connected.
  HTTP request sent, awaiting response... 503 Service Unavailable
  2026-08-12 19:30:09 ERROR 503: Service Unavailable.
  ```
  
  Update: These seem to have resolved.

File: core/connectors/sinks/opensearch_sink/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,2691 @@
...
+    positions
+        .iter()
+        .filter_map(|&position| pending.get(position).copied())
+        .collect()
+}
+
+fn parse_bulk_response(response: &Value) -> BulkAttempt {
+    let Some(items) = response.get("items").and_then(Value::as_array) else {
+        return BulkAttempt::default();
+    };
```
slbotbm: A successful HTTP response with no `items` array is converted to an empty `BulkAttempt`, and malformed entries are skipped later in the loop. `index_chunk()` can therefore return a clean outcome even though none of the submitted documents were accounted for, causing the batch to be treated as successfully handled. Make the parser return a `Result`, require exactly one valid item result per pending document, and treat a missing, short, or malformed `items` array as a retryable response-parsing failure.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,2691 @@
...
+    pub timeout: Option<String>,
+    pub refresh: Option<Refresh>,
+    pub max_retries: Option<u32>,
+    pub retry_delay: Option<String>,
+    pub max_retry_delay: Option<String>,
+    pub max_open_retries: Option<u32>,
+    pub verbose_logging: Option<bool>,
+}
+
+#[derive(Debug)]
```
slbotbm: Deriving `Debug` for the sink leaks the Basic-auth password after `open()`: `OpenSearch` derives `Debug`, its `Transport` includes `credentials`, and `opensearch` 2.4.0 derives `Debug` for `Credentials::Basic(String, String)` without redaction. The existing test formats only an unopened sink, where `client` is still `None`, so it misses this path. Replace the derive with a custom `Debug` implementation that omits or redacts `client`, and add a regression test that formats a sink after installing an authenticated client.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,2691 @@
...
+        Value::Object(object) => object,
+        other => Map::from_iter([("value".to_string(), other)]),
+    }
+}
+
+fn document_from_raw(bytes: Vec<u8>) -> Map<String, Value> {
+    // simd_json parses destructively, so the base64 fallback needs its own copy.
+    let mut parse_buffer = bytes.clone();
+    match simd_json::to_owned_value(&mut parse_buffer) {
+        Ok(value) => document_from_json(owned_value_to_serde_json(&value)),
```
slbotbm: Every raw payload is cloned in the consume hot path so that `simd_json` can mutate one copy while preserving the other for the base64 fallback. For large batches this temporarily doubles the payload memory and copies every byte. Parse the owned bytes non-destructively, for example with `serde_json::from_slice`, so the same buffer can be base64-encoded on parse failure without cloning it.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,2691 @@
...
+        operation: &str,
+        retries: u32,
+        max_retries: u32,
+        failure: &str,
+    ) {
+        let delay = jitter(exponential_backoff(
+            self.config.retry_delay,
+            retries,
+            self.config.max_retry_delay,
+        ));
```
slbotbm: `exponential_backoff()` computes `base * 2^attempt`, but the first retry passes `retries == 1`, so the configured `retry_delay` is never used and the first wait is already twice the base. Jitter is also applied after the helper caps at `max_retry_delay`, allowing the actual sleep to exceed that advertised maximum by up to 20%. Pass `retries - 1` as the attempt index and clamp the jittered result to `max_retry_delay`.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,2691 @@
...
+        return Err(Error::Connection(
+            "Invalid OpenSearch URL: host cannot be empty".to_string(),
+        ));
+    }
+
+    let with_scheme = if trimmed.starts_with("http://") || trimmed.starts_with("https://") {
+        trimmed.to_string()
+    } else {
+        format!("http://{trimmed}")
+    };
```
slbotbm: Scheme detection is case-sensitive, so a valid URL such as `HTTPS://host:9200` is rewritten as `http://HTTPS://host:9200` and targets the wrong host; unsupported explicit schemes are similarly disguised instead of rejected. Detect an explicit scheme case-insensitively, accept only HTTP or HTTPS, and prepend `http://` only when no scheme was supplied.

core/integration/tests/connectors/fixtures/opensearch/container.rs
slbotbm, line 1: Starting a fixed-name `ReuseDirective::Always` container is a check-then-create operation with no cross-process lock, so concurrent nextest invocations on a cold Docker daemon can race and all but one fail with `409 Conflict: name is already in use`. Use the bounded create-or-attach retry now present in the Elasticsearch fixture so losers retry until they attach to the winning container.

slbotbm (changes requested):

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,2903 @@
...
+
+    // The top-level flag lets a clean batch skip the per-item scan entirely.
+    if !response
+        .get("errors")
+        .and_then(Value::as_bool)
+        .unwrap_or(true)
+    {
+        return Ok(BulkAttempt {
+            indexed: items.len(),
+            ..BulkAttempt::default()
```
slbotbm: This fast path still accepts malformed item entries without validating them. For example, `{"errors":false,"items":[{}]}` passes the length check and is reported as one indexed document, even though the response contains no recognizable `index` result or successful status. That leaves the malformed-response fix incomplete, and the runtime will commit the offset for a document that was never actually accounted for. Please validate that every item contains the expected operation result and a 2xx status before returning success, and add an `errors: false` regression case with a malformed item.

core/connectors/sinks/opensearch_sink/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_opensearch_sink"
+version = "0.5.0-edge.1"
```
slbotbm: update to "0.5.0-edge.4"

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,2903 @@
...
+        f.debug_struct("OpenSearchSink")
+            .field("id", &self.id)
+            .field("config", &self.config)
+            .field("client", &self.client.is_some())
+            .field("invocations_count", &self.invocations_count)
+            .field("documents_indexed", &self.documents_indexed)
+            .field("errors_count", &self.errors_count)
+            .finish()
+    }
+}
```
slbotbm, line 120: This hides the authenticated client and relies on `SecretString` for the dedicated password field, but it still prints `self.config.url` verbatim through `self.config`. Since embedded URL credentials are rejected only later in `open()`, formatting a newly constructed sink from `https://admin:hunter2@host` exposes `hunter2`. Let's redact or sanitize the URL in the resolved config's `Debug` output.

core/connectors/sinks/opensearch_sink/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+async-trait = { workspace = true }
+base64 = { workspace = true }
+bytes = { workspace = true }
+iggy_common = { workspace = true }
+iggy_connector_sdk = { workspace = true }
+opensearch = { workspace = true }
+secrecy = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+simd-json = { workspace = true }
```
slbotbm: `simd-json` is used only by test fixtures inside the `#[cfg(test)]` module. It should be under `[dev-dependencies]`.

slbotbm (changes requested):

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3085 @@
...
+
+    if items.len() != expected {
+        return Err(BulkResponseError::ItemCountMismatch {
+            expected,
+            actual: items.len(),
+        });
+    }
+
+    // `errors: false` is only trustworthy if every item has a 2xx result.
+    if !response
```
mattp5657, line 984: > This fast path still accepts malformed item entries without validating them. For example, {"errors":false,"items":[{}]} passes the length check and is reported as one indexed document, even though the response contains no recognizable index result or successful status. That leaves the malformed-response fix incomplete, and the runtime will commit the offset for a document that was never actually accounted for. Please validate that every item contains the expected operation result and a 2xx status before returning success, and add an errors: false regression case with a malformed item.

core/connectors/sinks/opensearch_sink/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_opensearch_sink"
+version = "0.5.0-edge.4"
```
mattp5657, line 20: > update to "0.5.0-edge.4"

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3085 @@
...
+    invocations_count: AtomicU64,
+    documents_indexed: AtomicU64,
+    errors_count: AtomicU64,
+}
+
+// `OpenSearch` derives `Debug` down through its `Transport`, and
+// `opensearch::auth::Credentials::Basic` derives `Debug` on its raw
+// `(String, String)` without redaction, so a derived `Debug` on this struct
+// would print the Basic-auth password in plaintext once `client` is set.
+impl std::fmt::Debug for OpenSearchSink {
```
mattp5657, line 109: > This hides the authenticated client and relies on SecretString for the dedicated password field, but it still prints self.config.url verbatim through self.config. Since embedded URL credentials are rejected only later in open(), formatting a newly constructed sink from https://admin:hunter2@host exposes hunter2. Let's redact or sanitize the URL in the resolved config's Debug output.

core/connectors/sinks/opensearch_sink/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+opensearch = { workspace = true }
+secrecy = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
+url = { workspace = true }
+
+[dev-dependencies]
+simd-json = { workspace = true }
```
mattp5657, line 50: > simd-json is used only by test fixtures inside the #[cfg(test)] module. It should be under [dev-dependencies].

mattp5657:
@slbotbm  Hey, sorry but I had an agent accidentally push some code while working through a push-verify failure. 

To revert it, I force-pushed, which made it hard to track your commented changes. So I quoted them where the changes were, to hopefully make them easier to trace. 

Next time will revert but wasn't thinking. Sorry!

mattp5657:
Looks like one failure on CICD, looks to be a network problem with docker registry. Will try again in a little.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+
+    async fn index_chunk(
+        &self,
+        client: &OpenSearch,
+        documents: &[PreparedDocument],
+    ) -> Result<BulkOutcome, Error> {
+        // Only the documents OpenSearch has not yet accepted. A per-item
+        // transient rejection (429 under load, most commonly) shrinks this to
+        // just those documents rather than resending the whole chunk.
+        let mut pending: Vec<&PreparedDocument> = documents.iter().collect();
```
slbotbm, line 531: `index_chunk` ignores `Retry-After` on retryable bulk responses. Capture and parse the header with `retry::parse_retry_after` before consuming the response body.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+        })
+    }
+
+    fn document_id_from_field(
+        &self,
+        document: &Map<String, Value>,
+    ) -> Result<Option<String>, Error> {
+        let Some(field) = self.config.document_id_field.as_deref() else {
+            return Ok(None);
+        };
```
slbotbm, line 451: `document_id_from_field` accepts whitespace-only document IDs. Reject `id.trim().is_empty()` without otherwise trimming valid IDs.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+/// Requires exactly one item per pending document. A missing, short, or
+/// malformed `items` array leaves the true per-item outcome unknown, so the
+/// whole response is rejected rather than silently under-accounted: a chunk
+/// that OpenSearch answered with fewer or unparsable items must not be
+/// treated as cleanly handled just because nothing came back marked failed.
+fn parse_bulk_response(
+    response: &Value,
+    expected: usize,
+) -> Result<BulkAttempt, BulkResponseError> {
+    let items = response
```
slbotbm, line 970: `parse_bulk_response` ignores per-item `error` objects on its `errors: false` fast path. Treat a 2xx item carrying `error` as an untrusted or failed result rather than a success.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+                    if let Some(error) = outcome.into_error(&self.config.index) {
+                        first_error = first_error.take().or(Some(error));
+                    }
+                }
+                Err(error) => first_error = first_error.take().or(Some(error)),
+            }
+            debug!(
+                "OpenSearch sink with ID: {} indexed {}/{} documents",
+                self.id, attempted, total
+            );
```
slbotbm, line 510: `index_documents` logs cumulative attempted documents under the label `indexed`. Log the actual indexed count or rename the label to `attempted`.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+    value.expose_secret().trim().is_empty()
+}
+
+fn normalize_url(raw: &str) -> Result<String, Error> {
+    let trimmed = raw.trim();
+    if trimmed.is_empty() {
+        return Err(Error::Connection(
+            "Invalid OpenSearch URL: host cannot be empty".to_string(),
+        ));
+    }
```
slbotbm, line 1226: `normalize_url`  emits query/fragment and base-path warnings without the connector ID. Pass an ID or emit the warnings from the caller.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+
+impl From<OpenSearchSinkConfig> for ResolvedOpenSearchSinkConfig {
+    fn from(config: OpenSearchSinkConfig) -> Self {
+        let mut retry_delay = parse_duration(config.retry_delay.as_deref(), DEFAULT_RETRY_DELAY);
+        let mut max_retry_delay =
+            parse_duration(config.max_retry_delay.as_deref(), DEFAULT_MAX_RETRY_DELAY);
+        if retry_delay > max_retry_delay {
+            warn!(
+                "OpenSearch sink retry_delay ({:?}) exceeds max_retry_delay ({:?}). Swapping values.",
+                retry_delay, max_retry_delay
```
slbotbm, line 175: This logs a retry-delay swap without the connector ID. Perform the warning from an ID-aware construction path.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+        })
+        .collect();
+    Value::Object(map)
+}
+
+fn upsert_metadata_field(document: &mut Map<String, Value>, field: &str, value: Value) {
+    if document.insert(field.to_string(), value).is_some() {
+        debug!("Overwriting payload field '{field}' with OpenSearch connector provenance");
+    }
+}
```
slbotbm, line 1135: This logs provenance-field overwrites without the connector ID. Pass the ID through or remove the debug event.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+    /// itself rather than `retry_delay * 2`. Jitter can push the raw backoff
+    /// up to 20% above `max_retry_delay`, so the jittered result is clamped
+    /// back down to it.
+    async fn sleep_before_retry(
+        &self,
+        operation: &str,
+        retries: u32,
+        max_retries: u32,
+        failure: &str,
+    ) {
```
slbotbm, line 662: This emits retry warnings without `self.id`.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+        let status = response.status_code();
+        if status.is_success() {
+            return Ok(());
+        }
+
+        // A 403 only means cluster:monitor/health is missing, not that the cluster is down.
+        if status == StatusCode::FORBIDDEN {
+            warn!(
+                "OpenSearch health check returned 403: the configured user lacks the cluster-scoped cluster:monitor/health privilege. Treating the cluster as reachable; grant that privilege to restore the check."
+            );
```
slbotbm, line 320: Open-time logs here, `:333`, `:372`, `:385`, and `:393`, plus successful consume logs at `:781` and `:786`, lack the connector ID.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+            }),
+            None => Ok(indexed),
+        }
+    }
+
+    async fn index_chunk(
+        &self,
+        client: &OpenSearch,
+        documents: &[PreparedDocument],
+    ) -> Result<BulkOutcome, Error> {
```
slbotbm, line 527: This parses bulk responses into a full `serde_json::Value`. A narrow deserialization type with a dynamic action key would reduce response-side allocation.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+}
+
+// `derive(Debug)` would print `url` verbatim; `open()` isn't what redacts it.
+impl std::fmt::Debug for ResolvedOpenSearchSinkConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ResolvedOpenSearchSinkConfig")
+            .field("url", &redact_url_credentials(&self.url))
+            .field("index", &self.index)
+            .field("username", &self.username)
+            .field("password", &self.password)
```
slbotbm, line 148: `ResolvedOpenSearchSinkConfig::fmt` omits `document_id_field`.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+    pub index: String,
+    pub username: Option<String>,
+    pub password: Option<SecretString>,
+    pub document_id_field: Option<String>,
+    pub create_index_if_not_exists: Option<bool>,
+    pub index_mapping: Option<Value>,
+    pub include_metadata: Option<bool>,
+    pub batch_size: Option<usize>,
+    pub timeout: Option<String>,
+    pub refresh: Option<Refresh>,
```
slbotbm, line 88: This exposes `opensearch::params::Refresh`. A connector-owned serde enum would isolate the configuration surface.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+        Ok(PreparedDocument {
+            id,
+            document: Value::Object(document),
+        })
+    }
+
+    fn document_id_from_field(
+        &self,
+        document: &Map<String, Value>,
+    ) -> Result<Option<String>, Error> {
```
slbotbm, line 448: `document_id_from_field` silently falls back when the configured field is absent. Emit a bounded warning, not one warning per record.

core/connectors/sinks/opensearch_sink/README.md
slbotbm, line 1: This does not document the zero-batch clamp implemented at `src/lib.rs:193`.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+
+/// Serializes the `_bulk` NDJSON payload (alternating `index` action line and
+/// document line per document) into one `Bytes` buffer. Takes references so a
+/// retry can serialize just the rejected subset without cloning documents.
+fn build_bulk_body(index: &str, documents: &[&PreparedDocument]) -> Result<Bytes, Error> {
+    let mut buffer = BytesMut::new();
+    for document in documents {
+        serde_json::to_writer(
+            (&mut buffer).writer(),
+            &json!({ "index": { "_index": index, "_id": document.id } }),
```
slbotbm, line 905: `build_bulk_body` uses `BytesMut::new()` without a conservative initial capacity. Any estimate should be bounded and justified rather than assumed to match all document sizes.

slbotbm: It also allocates an intermediate JSON value for each bulk action. Replace it with serializer-backed borrowed structs; do not concatenate unescaped JSON text.

core/connectors/sinks/opensearch_sink/src/lib.rs
```diff
@@ -0,0 +1,3084 @@
...
+/// (`GENERATED_ID_PREFIX` + 64 hex chars) regardless of input length, while
+/// staying deterministic: same stream, topic, partition, offset, and message
+/// ID always produce the same ID, preserving the upsert-based idempotency
+/// this function exists for.
+fn generated_document_id(
+    topic_metadata: &TopicMetadata,
+    messages_metadata: &MessagesMetadata,
+    offset: u64,
+    message_id: u128,
+) -> Result<String, Error> {
```
slbotbm, line 1149: `generated_document_id` can preserve the exact serialized hash input while returning `String` directly instead of `Result<String, Error>`.

slbotbm (changes requested):

slbotbm:
Could you also resolve the conflict?

PR #3877, opened 2026-08-13
feat(python): let poll_messages take a consumer

### LINKED ISSUE #3876: Python SDK: add a consumer argument to poll_messages
### Description

`poll_messages` hardcodes `Consumer::default()` (foreign/python/src/client.rs:979), so the caller cannot name the consumer. That id resolves to numeric 0, so every Python process polling the same topic/partition shares one server-side offset slot and two independent consumers silently split the stream instead of each receiving it. It also makes consumer groups unreachable on this path, since `ConsumerKind::ConsumerGroup` cannot be selected at all.

### Affected area / component

Python SDK

### Proposed solution

Add a `Consumer` type to the binding, modelled as a complex enum like the other sum types there (`PollingStrategy`, `AutoCommit`), with the id accepting `str | int`:

```rust
pub enum Consumer {
    Single { id: PyIdentifier },
    Group { id: PyIdentifier },
}
```

Make `consumer` a required argument on `poll_messages`, and make `partition_id` optional in the same change: under the vsr feature a consumer-group poll only resolves the member's assignment when no partition is sent, and an omitted partition reads partition 0 for a regular consumer.

```python
await client.poll_messages(
    stream="s", topic="t", partition_id=0,
    polling_strategy=PollingStrategy.Next(), count=10, auto_commit=True,
    consumer=Consumer.Single("my-app"),
)

await client.poll_messages(
    stream="s", topic="t",
    polling_strategy=PollingStrategy.Next(), count=10, auto_commit=True,
    consumer=Consumer.Group("my-group"),
)
```

Note: Requiring `consumer` breaks existing callers.

### Alternatives considered

Keep `consumer` optional and default it to `Consumer::new(Identifier::numeric(1))`, matching the serde default the HTTP API already applies. That stays backwards compatible, but it silently moves an existing `next()` user's stored offset from slot 0 to slot 1 on upgrade, and Python would still be the only SDK picking a consumer for the caller.

To align with other SDKs, which all require a consumer, I lean toward the breaking option: it fails at the call site instead of quietly relocating an offset.

### Contribution

### Good first issue

ethanlin01x:

Closes #3876

## Rationale

The caller could not choose a consumer, so independent Python consumers collided on one server-side offset slot.

## What changed?

`poll_messages` built `Consumer::default()` internally, which resolves to numeric id 0, so every Python process polling the same topic/partition shared one offset slot and consumer groups were unreachable.

It now takes a required `Consumer` (`Consumer.Single(id)` / `Consumer.Group(id)`), and `partition_id` becomes optional so a group poll can resolve the member's assignment. Breaking change: omitting `consumer` raises `TypeError`.

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage

1. Which tools? Claude code
2. Scope of usage? Analysis and implementation
3. How did you verify the generated code works correctly? Local run


foreign/python/tests/test_message_operations.py
```diff
@@ -1217,3 +1246,118 @@ async def test_poll_messages_with_auto_commit_false_does_not_advance_next(
...
+                polling_strategy=PollingStrategy.Next(),
+                count=10,
+                auto_commit=True,
+            )
+            assert [
+                message.payload().decode("utf-8") for message in polled_messages
+            ] == test_messages
+
+    @pytest.mark.asyncio
+    async def test_poll_messages_with_shared_consumer_splits_the_partition(
```
slbotbm: `test_poll_messages_with_shared_consumer_splits_the_partition` -> `test_poll_messages_with_shared_consumer_shares_the_partition`

foreign/python/tests/test_message_operations.py
slbotbm, line 1: There are some more tests that I would like you to add:

- Create a topic with several partitions, put a distinguishable message in each partition, join one client to the group, and repeatedly call `poll_messages()` without a `partition_id`. The client should poll all of its assigned partitions in round-robin order (assert this behaviour). 
- Call `poll_messages()` with `Consumer.Single(...)` and no `partition_id` after sending messages to partition `0`. The call should read partition `0`.
- Create a group member with a known assignment and call `poll_messages()` with both `Consumer.Group(...)` and an explicitly owned `partition_id`. The poll should return messages from that partition.
- Add small tests for the Python-to-Rust conversion boundary (some of these could also be folded into existing tests):
     - Omitting the required `consumer` argument should raise `TypeError`.
     - Passing a value that is not a `Consumer` should raise `TypeError`.
     - An invalid string identifier should raise `ValueError`.
     - A negative integer or an integer larger than `u32` should be rejected.
     - Valid numeric IDs should be accepted for both `Consumer.Single(...)` and `Consumer.Group(...)`.
- If a client polls with `Consumer.Group(group_name)` without first joining that group, it has no assigned partitions and is not allowed to consume for the group. (should error)

slbotbm (dismissed):

foreign/python/tests/test_message_operations.py
ethanlin01x: Addressed [4337625](https://github.com/apache/iggy/pull/3877/commits/4337625f379d95d078fe362dad74d7e77444d517)

foreign/python/tests/test_message_operations.py
ethanlin01x: Addressed each point with its own commit.

ethanlin01x: One note on the invalid string identifier: `Consumer` just holds a `PyIdentifier`, and the conversion only happens inside `poll_messages`. So `Consumer.Single("")` builds fine and you don't get `ValueError: Invalid identifier` until the poll, which is where the test catches it.

foreign/python/src/client.rs
```diff
@@ -1036,20 +1036,24 @@ impl IggyClient {
     }
 
     /// Polls for messages from the specified topic and partition.
```
hubcio: first line still says "from the specified topic and partition" but `partition_id` is optional now - reword so it matches the new signature.

foreign/python/src/consumer.rs
```diff
@@ -215,6 +215,29 @@ impl IggyConsumer {
...
+    /// A member of the consumer group, sharing the group's offset.
+    Group { id: PyIdentifier },
+}
+
+impl TryFrom<&Consumer> for RustConsumer {
+    type Error = PyErr;
+
+    fn try_from(value: &Consumer) -> PyResult<RustConsumer> {
+        Ok(match value {
+            Consumer::Single { id } => RustConsumer::new(Identifier::try_from(id.clone())?),
```
hubcio: `id.clone()` runs on every poll - a `TryFrom<&PyIdentifier> for Identifier` impl (or taking the consumer by value) would drop it. tiny thing.

foreign/python/tests/test_message_operations.py
```diff
@@ -1217,3 +1246,351 @@ async def test_poll_messages_with_auto_commit_false_does_not_advance_next(
...
+                polling_strategy=PollingStrategy.Next(),
+                count=10,
+                auto_commit=True,
+            )
+            polled.extend(
+                (message.partition_id(), message.payload().decode("utf-8"))
+                for message in polled_messages
+            )
+
+        assert polled == [
```
hubcio: this pins the exact 0,1,2 rotation order, which couples the test to where the server's round-robin starts rather than to the consumer contract. asserting the set of (partition, payload) pairs would be less brittle, unless the order is meant to be guaranteed.

foreign/python/apache_iggy.pyi
```diff
@@ -358,6 +388,105 @@ class ConsumerGroupMember:
         Gets the collection of partitions the consumer group member is consuming.
```
hubcio: the `GlobalPermissions`/`UserHeaders` relocation and the new `__setitem__` stub look like stubgen drift that predates this change - fine to carry, just confirming this is pure regenerated output, not hand-edited.

hubcio (changes requested):

foreign/python/src/client.rs
ethanlin01x: cd83510ab6e97768756cc4ef1ae2dba1f0a2b1c9

foreign/python/src/consumer.rs
ethanlin01x: 88d84c8693dc9cb3d20ca85bad7a420d2515b739

spetz (approved):

hubcio (approved):

hubcio (approved):

PR #3880, opened 2026-08-13
feat(common)!: extensible key-value options for topics, streams, users

numinnex:
This PR reworks current Streams/Topics/Users metadata storage by replacing the hard coded values with an extensible key value pair storage. Additionally it moves some of the partition related settings such as `messages_required_to_save` or `segment_size` from the static configuration (config.toml) to topic property.


core/configs/src/configs_impl/file_provider.rs
```diff
@@ -75,6 +118,7 @@ impl<P: Provider + Clone> ConfigProvider for FileConfigProvider<P> {
         // If the config file exists, merge it into the configuration
         if file_exists(&self.file_path) {
             info!("Found configuration file at path: '{}'.", self.file_path);
+            reject_relocated_keys(&self.file_path)?;
```
hubcio: this guard only covers the toml file, and only when one exists - it sits inside `if file_exists(...)` and the env provider merges below it unchecked. so `IGGY_SYSTEM_PARTITION_ENFORCE_FSYNC=true` in a pure-env container (the usual k8s shape) warns and boots with fsync off - exactly the failure the doc above says this prevents. and since the partition runtime knobs are create-only, existing topics can never get fsync back after an upgrade. run the relocated-key check over env names too, outside the `file_exists` branch - the harness already has this exact check (`validate_env_var_names`), the server just never runs it. also `[system.message_deduplication]` was deleted but isn't in this list, so `enabled = true` slips through the same way. worth a test too - this fn has none.

foreign/go/binary_serialization/binary_response_deserializer.go
```diff
@@ -57,8 +57,10 @@ func DeserializeStream(payload []byte) (*iggcon.StreamDetails, error) {
 	if err != nil {
 		return nil, err
 	}
-	topics := make([]iggcon.Topic, 0)
-	for pos < len(payload) {
+	// Count-driven: a topic element carries variable-length options blocks,
+	// so "consume until the buffer ends" no longer delimits it.
+	topics := make([]iggcon.Topic, 0, stream.TopicsCount)
```
hubcio: preallocating from an unvalidated wire u32 - `TopicsCount` at max is a ~400 GB reservation and a go OOM isn't recoverable. `DeserializeTopics` caps the same hint further down; do the same here. (the rust decoders got the same fix via `bounded_capacity`.)

foreign/go/internal/command/update_user.go
```diff
@@ -50,7 +54,8 @@ func (u *UpdateUser) MarshalBinary() ([]byte, error) {
 		length += 2
 	}
 
-	bytes := make([]byte, length+1)
+	optionsBytes := iggcon.GetHeadersBytes(u.Options)
+	bytes := make([]byte, length+1+len(optionsBytes))
```
hubcio: `length` already counts both flag bytes when username and status are set, so the `+1` leaves a trailing `0x00` after the options block - and the server now decodes options to end-of-payload and rejects kind 0, so every `UpdateUser` with both fields set fails. the over-allocation is old but the new wire format makes it fatal; drop the `+1` and size the buffer exactly. (the both-nil case is off by one the other way and panics, but that predates this PR.)

core/common/src/wire_conversions.rs
```diff
@@ -799,6 +813,144 @@ pub(crate) fn user_headers_from_validated_slice(
...
+/// block carries no per-entry provenance. Admission calls this once for the
+/// client block (`explicit == true`) and fills defaults separately.
+///
+/// # Errors
+///
+/// Returns `IggyError::InvalidHeaderKey` / `InvalidHeaderValue` when a kind
+/// code has no domain meaning or a fixed-size kind carries a mismatched
+/// payload. The wire layer accepts unknown kinds for forwarding; interpreting
+/// them into the domain requires known semantics.
+pub fn resource_options_from_wire(
```
hubcio, line 883: `user_headers_from_validated_slice` errors on any value kind outside the known range, but the wire layer deliberately forwards unknown value kinds so mixed-version clusters can round-trip them - that's the stated contract in `options.rs`. this fn sits on the apply path, the sdk list-decode path, and the server's own http reads, so the first new `HeaderKind` code means: old replicas reject a commit new ones accept (permanent divergence), `get_topics()` fails the whole list over one entry, and the topic GET errors server-side. skipping entries with unknown kinds here closes all three at once. the comment at the apply site promises leniency the code doesn't deliver. (side note: the doc here says `InvalidHeaderKey`/`InvalidHeaderValue` but the actual error is `InvalidHeaderKind`.)

core/metadata/src/stm/stream.rs
```diff
@@ -1659,23 +1685,41 @@ impl StateHandler for CreateTopicWithAssignmentsRequest {
...
+        // refused. Refusing would make the verdict depend on the build, and a
+        // replica that predates a key would then be missing a topic its peers
+        // committed.
+        let explicit = TopicCreateOptions::parse_committed(&self.request.options);
+        let derived = TopicCreateOptions::parse_committed(&self.derived_options);
+        let (Ok(explicit_map), Ok(derived_map)) = (
+            resource_options_from_wire(&self.request.options, true),
+            resource_options_from_wire(&self.derived_options, false),
+        ) else {
+            return ApplyReply::err(CreateTopicResult::InvalidOptionValue);
```
hubcio: when this decode fails, the revision bump and the stats-registry insert above have already happened - reconcilers get woken and the registry keeps an orphaned entry for a topic that never got created. deterministic across replicas so nothing diverges, but the "past validation" comment above is no longer true. move the bump and insert below the decode.

core/metadata/src/stm/stream.rs
```diff
@@ -1659,23 +1685,41 @@ impl StateHandler for CreateTopicWithAssignmentsRequest {
...
         };
+        // Explicit wins on collision. `partitions_count` cannot appear here at
+        // all: it is a fixed field of the command, not an option key.
+        let mut options = derived_map;
```
hubcio: `extend` lets a client-sent literal 0 overwrite the derived entry, while the typed field below normalizes 0 to `None` and correctly resolves to the default. net effect: the resolved default is dropped from the persisted map entirely, so the map stops being the single source of truth its doc claims, and one `GetTopic` response shows the resolved value in the fixed field but 0 in the options block. reachable from the shipped cli: `iggy topic create --set max_topic_size=server_default` lands the sentinel via the raw map. after the merge, overwrite these keys from the resolved typed values.

core/server/src/dispatch.rs
```diff
@@ -1138,6 +1176,31 @@ async fn handle_client_request<B, MJ, S, SB>(
...
                 validate_partitions_change_count(delete_partitions.partitions_count)
             }),
+        // Only the updatable subset: the create-time knobs are pushed to
+        // partitions when the topic is built and nothing re-pushes them, so
+        // accepting one here would store a value no partition ever sees.
+        Operation::UpdateTopic => UpdateTopicRequest::decode_from(request_body(&request))
+            .map_err(|_| IggyError::InvalidCommand)
+            .and_then(|update_topic| {
+                validate_option_keys(&update_topic.options, UPDATABLE_TOPIC_OPTION_KEYS)?;
+                TopicCreateOptions::parse(&update_topic.options).map(|_| ())
```
hubcio: update still has no bounds check for `max_topic_size` - create runs `validate_topic_bounds` (twice, here and in `prepare_request`), but there's no `Operation::UpdateTopic` arm in `prepare_request` and no bounds call here, so an update can set a cap below one segment size and the stored map reports it as the topic's setting.

core/common/src/types/options/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                    parsed.max_topic_size = (size != MaxTopicSize::ServerDefault).then_some(size);
+                }
+                topic_option_keys::SEGMENT_SIZE => {
+                    let size = parse_byte_size(entry, key)?;
+                    parsed.segment_size = (size != 0).then_some(IggyByteSize::from(size));
+                }
+                topic_option_keys::ENFORCE_FSYNC => {
+                    parsed.enforce_fsync = Some(parse_bool(entry, key)?);
+                }
+                topic_option_keys::MESSAGES_REQUIRED_TO_SAVE => {
```
hubcio, line 806: only zero is rejected for the two flush thresholds - no upper bound anywhere, and `validate_topic_bounds` doesn't cover them. maxed out, the flush gate never trips and committed messages sit in the journal (which doesn't survive a crash) up to a full segment per partition. these used to be operator-only config; now anyone with create_topic can set them. add ceilings.

core/server/src/responses.rs
```diff
@@ -943,13 +978,122 @@ where
...
+                    description: "Flush the journal once it holds this many bytes, or a \
+                                  byte-size string; whichever threshold trips first flushes"
+                        .to_string(),
+                },
+                OptionDescriptor {
+                    key: WireName::new(topic_option_keys::PREALLOCATE_SEGMENTS)
+                        .map_err(|_| IggyError::InvalidFormat)?,
+                    kind: HeaderKind::Bool.as_code(),
+                    default_value: Bytes::copy_from_slice(&[u8::from(preallocate)]),
+                    description: "Reserve each segment's bytes up front where the filesystem \
```
hubcio: this description is the operator's only discovery surface for the option and it doesn't mention the cost: the fallocate runs inline on the shard thread (every rotation, plus a serial re-run per owned partition at boot), and nothing bounds `segment_size * partitions_count`, so one create can reserve partitions x 1 GiB. worth a sentence here and a cap at admission. the inline call itself is right - the old spawn_blocking panicked the shard.

core/binary_protocol/src/responses/topics/get_topics.rs
```diff
@@ -16,40 +16,57 @@
...
+        buf.put_u32_le(self.topics.len() as u32);
         for topic in &self.topics {
             topic.encode(buf);
         }
     }
 }
 
+/// Smallest a topic header can encode as: the fixed ids, timestamps, sizes and
+/// counts, plus a name length and two length-prefixed option blocks.
+const MIN_TOPIC_HEADER_SIZE: usize = 45;
```
hubcio: 45 contradicts the comment above - the fixed fields alone are 49 bytes, and with `name_len >= 1` plus two 4-byte option-block prefixes the real minimum is 59 (`TopicHeader::FIXED_SIZE + 1 + 8`). undersizing just loosens the cap, but go (`topicMinimumSize`) and java (`MIN_TOPIC_BYTES`) both say 58 from the same zero-length-name assumption - `WireName` rejects empty names, so all three want 59.

core/binary_protocol/src/responses/topics/get_topics.rs
```diff
@@ -16,40 +16,57 @@
...
+/// `GetTopics` response: count-prefixed topic headers.
 ///
 /// Wire format:
 /// ```text
-/// [TopicHeader]*
+/// [topics_count:u32_le][TopicHeader]*
 /// ```
 ///
-/// Empty payload means zero topics.
+/// The count prefix exists because a topic element carries variable-length
```
hubcio: the stated reason doesn't hold - `get_streams`/`get_users` keep run-to-end loops over elements that also carry variable-length options blocks, and they do detect short reads (the element decoder errors). the prefix is justified for pre-sizing; say that instead. also worth a rule somewhere: three list responses now ship with two framings and every sdk special-cases topics.

core/binary_protocol/src/responses/system/describe_options.rs
```diff
@@ -0,0 +1,202 @@
...
+        buf.put_u32_le(self.entries.len() as u32);
+        for entry in &self.entries {
+            entry.encode(buf);
+        }
+    }
+}
+
+/// Smallest a descriptor can encode as: a name, a kind byte, a
+/// length-prefixed default and a length-prefixed description.
+const MIN_DESCRIPTOR_SIZE: usize = 4 + 1 + 1 + 4 + 4;
```
hubcio: counts `WireName` as 4 bytes but its minimum is 2 (`1 + len`, len >= 1), so the floor is 11, not 14. overcounting only tightens the cap so it's safe, just wrong.

core/binary_protocol/src/responses/system/describe_options.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+impl WireDecode for OptionDescriptor {
+    fn decode(buf: &[u8]) -> Result<(Self, usize), WireError> {
+        let (key, mut pos) = WireName::decode(buf)?;
+        let kind = read_u8(buf, pos)?;
+        pos += 1;
+        let default_len = read_u32_le(buf, pos)? as usize;
+        pos += 4;
+        let default_end = pos + default_len;
```
hubcio: `pos + default_len` unchecked (same at line 79) - on a 32-bit target a wire u32 wraps, the `buf.len()` guard passes and the slice panics. `decode_options_prefixed` guards the identical pattern with `checked_add` and a comment explaining why. no 32-bit target ships today, so consistency rather than a live bug - but mirror the guard.

core/binary_protocol/src/requests/topics/create_topic_with_assignments.rs
```diff
@@ -72,9 +88,27 @@ impl WireDecode for CreateTopicWithAssignmentsRequest {
         }
 
         let request = CreateTopicRequest::decode_from(&buf[request_start..request_end])?;
-        let partitions_count = read_u32_le(buf, request_end)? as usize;
-        let mut offset = request_end + 4;
-        let mut partitions = Vec::with_capacity(partitions_count);
+
+        let derived_size = read_u32_le(buf, request_end)? as usize;
+        let derived_start = request_end + 4;
+        let derived_end = derived_start + derived_size;
```
hubcio: unchecked `pos + size` (also `request_end` at line 81) - the same 32-bit wrap `decode_options_prefixed` guards with `checked_add`, and this decode runs on journal replay, so a wrap would be a replica panic rather than a client error. the derived-block read here is that helper hand-rolled minus its guard - call it instead.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesDeserializer.java
```diff
@@ -134,11 +141,12 @@ public static Topic readTopic(ByteBuf response) {
         var messageExpiry = readU64AsBigInteger(response);
         var compressionAlgorithmCode = response.readByte();
         var maxTopicSize = readU64AsBigInteger(response);
-        var replicationFactor = response.readByte();
         var size = readU64AsBigInteger(response);
         var messagesCount = readU64AsBigInteger(response);
         var nameLength = response.readByte();
         var name = response.readCharSequence(nameLength, StandardCharsets.UTF_8).toString();
+        skipOptionsBlock(response, "topic explicit options");
```
hubcio: java skips the options blocks everywhere (also 96, 149, 523) and the records carry no options field, so the java client can't read segment_size, enforce_fsync or anything else this PR adds - and the tcp client can only set the three legacy keys. fine as a phased rollout, but file an issue so it isn't silent; today java is blind to the feature in both directions.

foreign/go/contracts/compression_algorithm.go
```diff
@@ -23,3 +23,12 @@ const (
 	CompressionAlgorithmNone CompressionAlgorithm = 1
 	CompressionAlgorithmGzip CompressionAlgorithm = 2
 )
+
+func (c CompressionAlgorithm) String() string {
+	switch c {
+	case CompressionAlgorithmGzip:
+		return "gzip"
+	default:
```
hubcio, line 57: `String()` returns "none" for every non-gzip value including invalid codes, and the gates in `topic.go` are `!= "none"` - so an out-of-range algorithm is silently dropped client-side and the topic gets the server default. before, the raw byte went on the wire and the server rejected it. switch on the enum and error on unknown codes.

core/common/src/http/topics/create_topic.rs
```diff
@@ -47,10 +47,12 @@ pub struct CreateTopic {
...
     /// Can't be lower than segment size in the config.
     pub max_topic_size: MaxTopicSize,
-    /// Replication factor for the topic.
-    pub replication_factor: Option<u8>,
     /// Unique topic name, max length is 255 characters.
     pub name: String,
+    /// Additional topic options as string key-values, parsed by the server
+    /// with the same rules as its config file. Unknown keys are rejected.
+    #[serde(default)]
+    pub options: BTreeMap<String, String>,
```
hubcio, line 59: options go in as a plain string map but come back as `{"value":{"kind":"uint64","value":"<base64>"},"explicit":true}` - an http caller has to base64+LE-decode to read a knob, and a GET body can't be fed back into a POST, so the re-send-only-options round trip documented on the response side is unimplementable over http. render values in their config-string form on the way out.

core/common/src/types/options/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+        }
+    }
+}
+
+/// One entry of a resource's option catalog, as served by `DescribeOptions`.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct OptionSpec {
+    /// Option key accepted by the create command.
+    pub key: String,
+    /// Canonical kind the server persists the value under. A `String` value
```
hubcio: two contract problems. "canonical kind the server persists the value under" is false for explicit keys - values are stored and echoed with the client's kind byte verbatim, so every cli `--set` lands as a string while the catalog advertises uint64. and `default_value` is documented as "this node's current resolved default" but every value served is a compile-time constant seeded at bootstrap, identical on every node and build. either canonicalize kinds at admission or document that kind is client-chosen, and fix the per-node wording.

core/common/src/types/options/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+/// entry instead of another wire change, and until then every key is rejected
+/// by name rather than stored and ignored.
+pub const UPDATABLE_STREAM_OPTION_KEYS: &[&str] = &[];
+
+/// Keys an `UpdateUser` options block may carry. Empty for the same reason as
+/// [`UPDATABLE_STREAM_OPTION_KEYS`].
+pub const UPDATABLE_USER_OPTION_KEYS: &[&str] = &[];
+
+/// Longest key echoed back in an option error. A rejected key is attacker-sized
+/// by definition, and the error rides the wire back to the client.
```
hubcio: "rides the wire back to the client" only holds for http - binary transports send just the error code and `from_repr` on the client rebuilds the variant with an empty string, so tcp/quic/ws users see "Unsupported option key: " with no key. since key rejection is the discovery mechanism for anyone not calling `describe_options`, carrying the key in the deny reply (or at least scoping this comment to http) matters.

core/cli/src/args/topic.rs
```diff
@@ -117,16 +117,21 @@ pub(crate) struct TopicCreateArgs {
...
     /// Message expiry time in human-readable format like "unlimited" or "15days 2min 2s"
     ///
     /// "server_default" or skipping parameter makes CLI to use server default (from current server config) expiry time
     #[arg(default_value = "server_default", value_parser = clap::value_parser!(IggyExpiry), verbatim_doc_comment)]
     pub(crate) message_expiry: Vec<IggyExpiry>,
+    /// Additional topic option as key=value, repeatable
+    ///
+    /// Values are sent as strings and parsed server-side through each option's
+    /// own FromStr (e.g. --set segment_size=128MiB). The server rejects keys it
+    /// does not support; GET /options/topic lists the ones it accepts.
```
hubcio: this help text points binary-transport cli users at an http endpoint. the sdk has `describe_options` but no cli subcommand calls it - an `iggy options topic` command would make this text honest. the same string is pinned in the cli help snapshot test, so both need the edit.

core/binary_protocol/src/codec.rs
```diff
@@ -263,6 +263,21 @@ pub fn capped_capacity(count: usize, remaining: usize, min_item_size: usize) ->
...
+/// wire.
+///
+/// A count read off the wire is chosen by the peer, and `Vec::with_capacity`
+/// reserves for it before a single element has been bounds-checked. A hostile
+/// or corrupt frame claiming `u32::MAX` elements therefore aborts the process
+/// on allocation failure instead of returning a decode error. The bytes left
+/// to decode are the honest ceiling on how many elements can exist.
+#[must_use]
+#[inline]
+pub fn bounded_capacity(count: usize, remaining: usize, min_element_size: usize) -> usize {
```
hubcio, line 269: this duplicates `capped_capacity` above - same expression, differing only when `min_item_size == 0`, where the old one returns the raw wire count unbounded (and a test pins that). one helper should survive, with the `.max(1)` body - the zero branch is the exact hole this fn's doc warns about. fold them and invert the zero-size test to pin the bounded behavior.

core/metadata/src/impls/metadata.rs
```diff
@@ -758,6 +760,22 @@ pub struct IggyMetadata<C, J, S, M, SB = PingPongSuperblock> {
     /// set from server config at bootstrap ([`Self::set_default_message_expiry`]).
     /// Defaults to never-expire, matching the shipped server config.
     default_message_expiry: Cell<u64>,
+    /// Resolved byte value an absent `segment_size` option takes at
+    /// admission. Same resolution discipline as the two defaults above; set
+    /// from `system.segment.size` at bootstrap
+    /// ([`Self::set_default_segment_size`]). Defaults to 1 GiB, matching the
+    /// shipped server config.
+    default_segment_size: Cell<u64>,
```
hubcio: all eight of these cells (and their setters/getters) only ever transport compile-time constants now - bootstrap seeds each with the same `iggy_common::DEFAULT_*` the `Cell::new` already holds, and the segment ceiling collapses to a constant too since config validation refuses boot unless `transfer_artifact_bytes_max - max_message_size >= SEGMENT_MAX_SIZE_BYTES`. reading the consts at the use sites drops ~110 lines. the docs here still say "set from server config at bootstrap", which stopped being true when those config keys were removed.

core/server/src/bootstrap.rs
```diff
@@ -1127,8 +1129,28 @@ async fn shard_main(
...
-    metadata.set_default_message_expiry(u64::from(config.system.topic.message_expiry));
+    metadata.set_default_max_topic_size(iggy_common::DEFAULT_MAX_TOPIC_SIZE);
+    metadata.set_default_message_expiry(iggy_common::DEFAULT_MESSAGE_EXPIRY);
+    metadata.set_default_segment_size(iggy_common::DEFAULT_SEGMENT_SIZE);
+    // Per-topic segment ceiling: the smaller of the global segment maximum
+    // and the state-transfer artifact budget minus one bus frame (a segment
+    // may close one whole batch past its cap; an artifact ceiling below that
+    // refuses a legal segment and livelocks the partition's rejoin).
+    metadata.set_max_topic_segment_size(
+        configs::validators::SEGMENT_MAX_SIZE_BYTES.min(
```
hubcio: this `.min()` is dead - config validation refuses boot unless `transfer_artifact_bytes_max >= SEGMENT_MAX_SIZE_BYTES + max_message_size`, so the subtraction is always >= the cap and this always yields `SEGMENT_MAX_SIZE_BYTES`. pass the constant and keep a comment pointing at the validator, since this is the only runtime expression of that coupling.

foreign/node/src/wire/options.utils.ts
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @param p - Buffer containing `[options_len:u32_le][options TLV]`
+ * @param pos - Starting position of the length prefix
+ * @returns Bytes consumed (prefix included) and deserialized options
+ */
+export const deserializePrefixedOptions = (
+  p: Buffer, pos = 0
+): OptionsDeserialized => {
+  const length = p.readUInt32LE(pos);
+  const options = deserializeOptions(p, pos + 4, pos + 4 + length);
```
hubcio: no check that `pos + 4 + length` fits the buffer - `subarray` clamps silently, so a truncated known-kind value throws inside `deserializeHeaderValue`, gets swallowed by the forward-compat catch, and surfaces as a raw buffer instead of the typed value, with `bytesRead` over-reporting. go and rust both hard-check the prefix; c# does now too.

core/metadata/src/stm/snapshot.rs
```diff
@@ -562,6 +573,60 @@ mod tests {
         );
     }
 
+    #[test]
```
hubcio, line 576: pins topic and stream shapes, but `UserSnapshot` also gained `options` in this PR and isn't pinned - add it so the next user-snapshot edit trips the test the same way.

core/common/src/types/options/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+        }
+        parsed
+    }
+}
+
+/// What a parse does with an entry this build cannot interpret: a key outside
+/// [`TOPIC_OPTION_KEYS`], or a catalog key whose value kind or payload does
+/// not parse.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum UnknownEntry {
```
hubcio: `UnknownEntry` plus the `absorb` wrapper exist only to turn `absorb_strict`'s err into ok - the call sites already spell the difference (`?` vs `let _ =`), so calling `absorb_strict` directly deletes the enum and ~30 lines with identical behavior. the "assigns only after its parse succeeds" comment should move onto `absorb_strict` - it's the invariant that makes the skip path safe.

core/common/src/wire_conversions.rs
```diff
@@ -799,6 +813,144 @@ pub(crate) fn user_headers_from_validated_slice(
...
+        buf.put_u8(key.kind().as_code());
+        #[allow(clippy::cast_possible_truncation)]
+        buf.put_u32_le(key.as_bytes().len() as u32);
+        buf.put_slice(key.as_bytes());
+        buf.put_u8(option.value.kind().as_code());
+        #[allow(clippy::cast_possible_truncation)]
+        buf.put_u32_le(option.value.as_bytes().len() as u32);
+        buf.put_slice(option.value.as_bytes());
+    }
+    // Structural validity holds by construction: entries come from valid
```
hubcio, line 988: this holds because every producer went through `validate_options`, not because of anything `HeaderKey` guarantees - and `from_validated` skips the check that would catch a violation, which would produce a block the receiving peer refuses. a `debug_assert!(key.kind() == HeaderKind::String)` in the loop would make it self-enforcing.

foreign/csharp/Iggy_SDK/Utils/CommandCodes.cs
```diff
@@ -23,6 +23,7 @@ internal static class CommandCodes
     internal const int GET_STATS_CODE = 10;
     internal const int GET_SNAPSHOT_CODE = 11;
     internal const int GET_CLUSTER_METADATA_CODE = 12;
+    internal const int DESCRIBE_OPTIONS_CODE = 13;
```
hubcio, line 26: dead constant - c# has no describe_options call (same for `DescribeOptions` in node's `command.code.ts`). wire up the call or drop the constants until it exists.

core/cli/src/commands/binary_topics/update_topic.rs
```diff
@@ -21,39 +21,47 @@ use async_trait::async_trait;
...
     update_topic: UpdateTopic,
+    compression_algorithm: CompressionAlgorithm,
     message_expiry: IggyExpiry,
     max_topic_size: MaxTopicSize,
-    replication_factor: u8,
+    options: TopicUpdateOptions,
 }
 
 impl UpdateTopicCmd {
+    #[allow(clippy::too_many_arguments)]
```
hubcio: the arg count is 6, under clippy's threshold of 7 - this allow is dead, remove it.

hubcio (changes requested):
reviewed the whole diff at this head. the recent refinements already closed several things i had flagged (the unbounded `with_capacity` in the rust decoders, `parse_bool`/`parse_compression` length checks, the c# `MapOptions` bounds, update-topic moving to patch-semantics options, gofmt/ruff) - nice.

what i'd still treat as blocking:
- the relocated-config-key guard not covering env vars (comment on `file_provider.rs`) - silent fsync loss on the containerized path
- the go `UpdateUser` trailing-byte break and the go `DeserializeStream` unbounded prealloc

bigger-picture items with no single line to anchor on:
- sdk parity: node exposes all 8 option keys typed, python has typed + a raw map, rust both, go/c#/java can only set the 3 legacy keys, java can't read options at all, c++ has neither raw map nor the new keys. one feature, five shapes - needs a parity plan or per-sdk issues before this is the story for 7 sdks.
- no bdd scenario touches options and nothing asserts cross-sdk byte equality for the new block; the per-sdk unit tests added are self-consistency only.
- no cli/mcp surface for the feature (`--set` exists only on `iggy topic create`; mcp create/update expose no options), and no example demonstrates the discovery or extension mechanism.
- shipped `config.toml` sets `transfer_artifact_bytes_max` exactly at the new boot floor, so raising `max_message_size` alone now refuses boot - worth a doc note.

the design itself holds up: hot paths untouched, the snapshot version bump plus nested arity pins handle the topic-snapshot relayout properly, derived options resolved once on the primary is the right call, and the deduplicator removal is clean dead-code deletion.

hubcio (dismissed):

piotr-laserdata (approved):

spetz (approved):

mmodzelewski (approved):

PR #3886, opened 2026-08-15
feat(connectors): Add Iggy source connector

### LINKED ISSUE #3869: feat(connectors): Iggy source connector implementation
Parent issue: #3764

Implement a source connector module that consumes messages from an upstream Iggy (cluster) into the connectors runtime. This issue serves as the overall implementation tracker. Please refer to the original proposal for detailed design.

### M1: Basic functionality
- Messages flow from upstream Iggy (source) through the connector to downstream Iggy
- Configurable stream and topic selection
- Progress persistence with resumable offset tracking

### M2: Fine-grained configuration
- TLS support
- Additional authentication methods (username/password, token)
- Flow control: finer-grained concurrency control and batching

### M3: Enhancement
- Compression and pro-active retry
- Partition-aware offset tracking

jiengup:

Closes #3869 
Relates to #3764 

## Rationale

Cross-cluster topic replication had no first-class path: the connectors subsystem's sinks and sources target external systems (Postgres, Elasticsearch, HTTP, etc.), and the runtime itself connects to a single Iggy cluster. Cross-cluster sync (disaster recovery, data migration, federated setups) therefore required external tooling. This PR adds `iggy_source`: a source connector that replicates a topic from an **upstream** Iggy cluster into the cluster the connectors runtime is connected to.

## What changed?

Previously there was no way to sync two Iggy clusters through connectors. Now `core/connectors/sources/iggy_source` connects to the upstream cluster and discovers all partitions in `open()`, polls each partition with an explicit `poll_messages(offset)` call per cycle, and passes payloads and user headers through as `Schema::Raw` while preserving upstream message IDs for downstream deduplication. Sync progress (per-partition confirmed offsets, total synced, error count) is persisted in connector state, and restarts resume from `saved_offset + 1` without replaying history.

### Design tradeoffs

- **Low-level `poll_messages` API instead of the high-level `IggyConsumer`**: crash recovery requires per-partition starting offsets, but `IggyConsumer` accepts only one global `polling_strategy` at construction time, which cannot express per-partition offsets. The low-level API allows passing `PollingStrategy::offset(saved + 1)` per partition explicitly.
- **No consumer group**: server-side group offset storage is untrusted (consume ack precedes downstream produce, so a crash loses messages). For single-instance replication the group provides only fencing, no real benefit, while introducing join/rebalance failure modes. `JoinConsumerGroupResponse` is empty (no member id returned), so assignment-driven multi-instance scaling is not expressible through the low-level API. Revisit once the server exposes member ids.
- **Direct offset jump vs replay-and-filter**: recovery uses `offset(saved + 1)` directly rather than `first()` plus in-memory filtering, keeping restart cost O(batch) instead of O(topic history).
- **`Schema::Raw` byte passthrough**: no decode/encode; payloads, user headers, and message IDs are preserved verbatim, avoiding re-serialization cost and format corruption.
- **`InvalidOffset` reset**: when upstream retention expires a saved offset, the partition is warned and reset to `initial_offset` (`earliest` / `latest` / numeric) without blocking other partitions.
- **Exponential backoff**: connection-level failures reuse the SDK's `exponential_backoff` + `jitter` (starting at `retry_interval`, capped at `max_retry_interval`); the failure counter is an `AtomicU64` and resets on any successful cycle.
- **Missing upstream stream/topic**: warn and auto-create (topics with 1 partition).

### Persisted state design

```rust
#[derive(Debug, Serialize, Deserialize)]
struct State {
    offsets: HashMap<u32, u64>,  // partition_id -> offset of the last message confirmed written downstream
    messages_synced: u64,        // cumulative synced message count
    errors_count: u64,           // cumulative errors (connection failures, conversion failures, offset resets)
}
```

- **Encoding and durability**: serialized with MessagePack (`rmp_serde`) via `ConnectorState::serialize/deserialize`; the runtime persists it with the existing `FileStateProvider` to `{state_path}/source_{key}.state`, inheriting its atomic-rename + fsync protocol and `0o600` permissions. **The state save path is untouched.**
- **Confirmation semantics (the key invariant)**: the connector advances offsets in state only for messages handed to the runtime, and the runtime saves state only after a **successful downstream send**. The on-disk state is therefore always the "confirmed delivered downstream" watermark and the single authoritative source for crash recovery.
- **Bounded size**: state is one `u64` per partition plus two counters, O(partition count), so rewriting the whole file every batch is cheap.
- **Failure tolerance**: deserialization failures (corruption, version mismatch) log a warning and start from a fresh state (non-fatal). `initial_offset` applies only to partitions with no saved entry (first run or newly added partitions); existing entries always win.
- **Error-count persistence**: `errors_count` increments and offset resets ride the same state channel, and state is returned even on empty-message cycles (e.g., connection-failure cycles), so error accounting and offset resets survive restarts.

### Crash recovery analysis

| Failure point | Behavior |
|---------------|----------|
| Connector process crash (upstream and downstream healthy) | The runtime persists state only after a successful downstream send; offsets not yet persisted are re-polled → at-least-once, with the preserved upstream message ID enabling downstream dedup |
| Upstream cluster outage | Poll errors → `errors_count` incremented, offsets not advanced, exponential backoff with jitter; the SDK client reconnects automatically |
| Downstream cluster outage | Runtime `producer.send` fails → state not saved; the connector's in-memory offsets have advanced, so that batch is dropped for the lifetime of the process, this can be fixed when #3855 is merged; a connector restart replays from the stale state file → at-least-once |
| Upstream retention expires data | `InvalidOffset` → partition reset to `initial_offset`; expired messages are unrecoverable (inherent to offset-based replication) |

### Sync semantics

- **At-least-once**: state records only offsets confirmed written downstream; recovery always resumes from the confirmation point + 1, so no message is lost, and duplicates within the crash window are deduplicated downstream via the preserved upstream message IDs.
- The state mutex is acquired exactly twice per poll cycle (read offsets / write offsets) and never held across I/O; empty polls do not count as errors.

## Local Execution

- Passed: `cargo fmt --all`, `cargo sort --no-format --workspace`, `cargo clippy -p iggy_connector_iggy_source --all-targets --all-features -- -D warnings`, `cargo test -p iggy_connector_iggy_source` (11 unit tests: state restore, serialization round-trip, config defaults, initial_offset parsing, next_strategy jump, connection-string redaction), taplo, markdownlint, license-headers, cargo machete
- Passed: `cargo test -p integration -- connectors::iggy_source`
- Pre-commit hooks ran / not ran: not ran

### End-to-end verification (two real server-ng instances + runtime)

- Upstream/downstream `iggy-server` (TCP 8090/8091) + `iggy-connectors`; a CLI producer continuously emitted messages with headers
- All 542 messages synced, message counts identical on both clusters; sampled payloads, user headers (`producer:string`, `seq:uint64`), and message IDs are byte-identical
- Crash recovery: killed the connector, produced 3 more messages, restarted → `Restored state ... Offsets: {0: 4}, messages synced: 5` → only the 3 missing messages were synced, the first 5 with zero duplicates
- The topic-size difference between the two clusters was traced to the server's 256-byte per-save metadata blocks (CLI single-message sends vs runtime batched sends), not to any data difference

## AI Usage

1. Deepseek V4 pro (opencode)
2. Architecture and trade-offs were decided by the human; implementation was AI-generated with human review
3. E2E test was verified by both the agent and human

### DIFF
--- a/core/connectors/sources/iggy_source/README.md
+++ b/core/connectors/sources/iggy_source/README.md
@@ -1,0 +1,119 @@
+# Iggy Source Connector
+
+The Iggy source connector replicates a topic from an upstream Apache Iggy
+cluster into a stream and topic managed by the connectors runtime. It polls
+every upstream partition and persists the last acknowledged offset for each
+partition so replication can resume after a restart.
+
+## Features
+
+- Polls every partition of one upstream topic.
+- Preserves payload bytes and non-zero message IDs.
+- Preserves user headers by default, with an option to omit them.
+- Stores per-partition offsets in the connector runtime's state storage.
+- Retries polling failures with exponential backoff and jitter.
+- Creates the configured upstream stream or topic when it does not exist.
+
+## Configuration
+
+```toml
+type = "source"
+key = "iggy"
+enabled = true
+version = 0
+name = "Iggy source"
+path = "target/release/libiggy_connector_iggy_source"
+verbose = false
+benchmark = false
+
+[[streams]]
+stream = "downstream_stream"
+topic = "downstream_topic"
+schema = "raw"
+batch_length = 100
+linger_time = "5ms"
+
+[plugin_config]
+connection_string = "iggy+tcp://iggy:iggy@127.0.0.1:8090"
+upstream_stream = "upstream_stream"
+upstream_topic = "upstream_topic"
+poll_interval = "1s"
+batch_size = 100
+initial_offset = "earliest"
+include_user_headers = true
+retry_interval = "1s"
+max_retry_interval = "60s"
+verbose_logging = false
+```
+
+Use `schema = "raw"` on the downstream stream to replicate payload bytes
+without decoding or re-encoding them. The runtime-level `batch_length` controls
+downstream producer batching, while `plugin_config.batch_size` controls how many
+messages are requested from each upstream partition in one poll cycle.
+
+### Plugin Fields
+
+| Field | Required | Default | Description |
+| --- | --- | --- | --- |
+| `connection_string` | yes | none | Connection string for the upstream Iggy cluster. |
+| `upstream_stream` | yes | none | Name of the stream to replicate from. |
+| `upstream_topic` | yes | none | Name of the topic to replicate from. |
+| `poll_interval` | no | `2s` | Delay before each upstream poll cycle. |
+| `batch_size` | no | `100` | Maximum messages requested from each upstream partition per poll cycle. |
+| `initial_offset` | no | `earliest` | Starting position for a partition without a saved offset. Accepts `earliest`, `latest`, or an absolute numeric offset. |
+| `include_user_headers` | no | `true` | Copy user headers to downstream messages. |
+| `retry_interval` | no | `1s` | Base delay used for exponential backoff after a failed poll cycle. |
+| `max_retry_interval` | no | `60s` | Maximum delay between poll retries. |
+| `verbose_logging` | no | `false` | Log per-cycle connector details at info level instead of debug level. |
+
+The `connection_string` can use any transport supported by the Rust client,
+including TCP, QUIC, HTTP, and WebSocket. It may contain credentials, so keep
+the connector configuration private or supply the value through the
+`IGGY_CONNECTORS_SOURCE_IGGY_PLUGIN_CONFIG_CONNECTION_STRING` environment
+variable.
+
+## Offset and Delivery Semantics
+
+For each partition, the connector resumes at one offset after the last saved
+offset. `initial_offset` is used only when that partition has no saved offset:
+
+- `earliest` starts with the oldest available message.
+- `latest` requests up to `batch_size` of the most recent messages.
+- A numeric value starts at that absolute offset.
+
+The connector stages new offsets while polling. It commits them only after the
+runtime sends the complete downstream batch and saves the connector state. If
+the batch is rejected, the previous offsets remain committed and the messages
+are eligible for replay. Consumers should therefore tolerate duplicate
+delivery after failures or crashes.
+
+> When `initial_offset = "latest"` and no offset has been committed yet, each
+> retry recalculates the current tail of the upstream partition. If a first
+> batch is rejected while new messages arrive, messages from that failed batch
+> can be skipped. Use `earliest` or a numeric offset when replay continuity is
+> required.
+
+If retention makes a saved offset invalid, the connector clears that
+partition's saved position and applies `initial_offset` again.
+
+## Message Mapping
+
+| Upstream field | Downstream behavior |
+| --- | --- |
+| Payload | Copied without modification when the downstream schema is `raw`. |
+| Message ID | Copied when the upstream ID is non-zero. |
+| User headers | Copied when `include_user_headers = true`. |
+| Offset and partition | Reassigned by the downstream Iggy topic. |
+| Timestamp and checksum | Generated for the downstream message. |
+
+## Operational Notes
+
+- The connector discovers upstream partitions when it opens. Restart it after
+  adding partitions to the upstream topic.
+- If the upstream stream is missing, the connector creates it. If the topic is
+  missing, it creates the topic with one partition and no compression.
+- Invalid `initial_offset` values produce a warning and fall back to
+  `earliest`.
+- If user headers cannot be decoded, the connector rejects the complete polled
+  batch so its offsets are not advanced. Set `include_user_headers = false` to
+  replicate payloads without parsing upstream headers.
--- a/core/connectors/sources/iggy_source/README.md
+++ b/core/connectors/sources/iggy_source/README.md
@@ -11,7 +11,7 @@ partition so replication can resume after a restart.
 - Preserves payload bytes and non-zero message IDs.
 - Preserves user headers by default, with an option to omit them.
 - Stores per-partition offsets in the connector runtime's state storage.
-- Retries polling failures with exponential backoff and jitter.
+- Retries fully failed polling cycles with exponential backoff and jitter.
 - Creates the configured upstream stream or topic when it does not exist.
 
 ## Configuration
@@ -41,6 +41,7 @@ poll_interval = "1s"
 batch_size = 100
 initial_offset = "earliest"
 include_user_headers = true
+malformed_message_policy = "block"
 retry_interval = "1s"
 max_retry_interval = "60s"
 verbose_logging = false
@@ -62,7 +63,8 @@ messages are requested from each upstream partition in one poll cycle.
 | `batch_size` | no | `100` | Maximum messages requested from each upstream partition per poll cycle. |
 | `initial_offset` | no | `earliest` | Starting position for a partition without a saved offset. Accepts `earliest`, `latest`, or an absolute numeric offset. |
 | `include_user_headers` | no | `true` | Copy user headers to downstream messages. |
-| `retry_interval` | no | `1s` | Base delay used for exponential backoff after a failed poll cycle. |
+| `malformed_message_policy` | no | `block` | Handling for messages with unparsable user headers: `block` retries from the failed offset; `drop_headers` forwards the payload without those headers. |
+| `retry_interval` | no | `1s` | Base delay used for exponential backoff when every partition poll in a cycle fails. |
 | `max_retry_interval` | no | `60s` | Maximum delay between poll retries. |
 | `verbose_logging` | no | `false` | Log per-cycle connector details at info level instead of debug level. |
 
@@ -114,6 +116,14 @@ partition's saved position and applies `initial_offset` again.
   missing, it creates the topic with one partition and no compression.
 - Invalid `initial_offset` values produce a warning and fall back to
   `earliest`.
-- If user headers cannot be decoded, the connector rejects the complete polled
-  batch so its offsets are not advanced. Set `include_user_headers = false` to
-  replicate payloads without parsing upstream headers.
+- If user headers cannot be decoded, `malformed_message_policy = "block"`
+  forwards the valid prefix before the malformed message, leaves that
+  partition at the failed offset, and continues polling other partitions at
+  the normal `poll_interval`. Record conversion errors do not activate the
+  connector-wide polling backoff.
+- `malformed_message_policy = "drop_headers"` forwards a malformed message's
+  payload without its user headers and advances the partition offset. Setting
+  `include_user_headers = false` bypasses header parsing for every message.
+- A polling error on one partition is logged and counted without skipping the
+  remaining partitions. Connector-wide backoff is activated only when no
+  partition poll succeeds during the cycle.
--- a/core/connectors/sources/iggy_source/README.md
+++ b/core/connectors/sources/iggy_source/README.md
@@ -60,7 +60,7 @@ messages are requested from each upstream partition in one poll cycle.
 | `upstream_stream` | yes | none | Name of the stream to replicate from. |
 | `upstream_topic` | yes | none | Name of the topic to replicate from. |
 | `poll_interval` | no | `2s` | Delay before each upstream poll cycle. |
-| `batch_size` | no | `100` | Maximum messages requested from each upstream partition per poll cycle. |
+| `batch_size` | no | `100` | Maximum messages requested from each upstream partition per poll cycle. Must be between `1` and `10,000`. |
 | `initial_offset` | no | `earliest` | Starting position for a partition without a saved offset. Accepts `earliest`, `latest`, or an absolute numeric offset. |
 | `include_user_headers` | no | `true` | Copy user headers to downstream messages. |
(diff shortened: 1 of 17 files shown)


core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,806 @@
...
+        let stream_id = self
+            .stream_id
+            .as_ref()
+            .ok_or_else(|| Error::InitError("Upstream stream not initialized".to_string()))?;
+        let topic_id = self
+            .topic_id
+            .as_ref()
+            .ok_or_else(|| Error::InitError("Upstream topic not initialized".to_string()))?;
+
+        let failures = self.consecutive_failures.load(Ordering::Relaxed);
```
mattp5657: `consecutive_failures` is incremented *after* a failure and then passed directly as the 0-based `attempt` argument to `exponential_backoff(base, attempt, max)` on the next cycle. The first retry therefore computes `base × 2^1` instead of `base × 2^0`, doubling every step of the backoff schedule versus what `retry_interval` configures.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,806 @@
...
+        if failures > 0 {
+            let delay = jitter(exponential_backoff(
```
mattp5657, line 379: `exponential_backoff` clamps to `max_retry_interval` internally, but `jitter()` (±20%) is applied afterward with no trailing `.min(max_retry_interval)`. A fully backed-off retry can sleep up to ~20% past the configured ceiling.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,806 @@
...
+                )
+                .await;
+
+            match polled {
+                Ok(polled) => {
+                    any_success = true;
+                    if polled.messages.is_empty() {
+                        continue;
+                    }
+                    let last_offset = polled
```
mattp5657: `last_offset` is computed from the last *polled* message regardless of whether `build_produced_message()` succeeded for that or any earlier message in the batch. A conversion failure (e.g. malformed user headers) is logged and counted in an error counter, but the offset still commits past it once acked.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,806 @@
...
+                                     ID: {}: {e}",
+                                    message.header.offset, self.id
+                                );
+                                errors_in_cycle += 1;
+                            }
+                        }
+                    }
+                    candidate_state.offsets.insert(partition_id, last_offset);
+                }
+                Err(IggyError::InvalidOffset(offset)) => {
```
mattp5657: On `InvalidOffset`, the saved offset is dropped and the partition falls back to the configured `initial_offset`, without setting `connection_failure` so this path never engages the backoff/circuit-breaker logic at all and can re-trigger every `poll_interval` indefinitely.

mattp5657 (commented):
Lets also add:

1. README for the specific connector.
2. Some testing around multiple partitions, tests with bad data, and testing around `InvalidOffset` recovery.

core/connectors/sources/iggy_source/src/lib.rs
jiengup: This behavior aligns with the code in `core/connectors/sdk/src/retry.rs`. Should we change it also?

core/connectors/sources/iggy_source/src/lib.rs
jiengup: Same as comment 1.

core/connectors/sources/iggy_source/src/lib.rs
mattp5657: I added an issue(#4084)

mattp5657: This stemmed from a review on my OpenSearch Connector. So I think you're right we should fix the interface instead of correcting it through codebase.

jiengup: In the latest commit, I keep this implementation and it can be fixed automatically when #4084 is solved.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,908 @@
...
+                Ok(polled) => {
+                    if polled.messages.is_empty() {
+                        continue;
+                    }
+                    let partition_messages = build_produced_messages(
+                        &polled.messages,
+                        self.include_user_headers,
+                        partition_id,
+                        self.id,
+                    )?;
```
mattp5657: A malformed/unconvertible message on any partition aborts the entire `poll()` cycle via `?` and discards messages already staged for earlier partitions, skips every partition after it, and bypasses both the error-count and backoff accounting (which live after the early return), so the connector busy-loops on the same message forever with no operator-visible signal.

Can we also maybe add this to the integration tests, how it continues after malformed records?

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,908 @@
...
+                        &mut candidate_state,
+                        partition_id,
+                        poll_error,
+                    ) {
+                        error!(
+                            "Failed to poll partition {partition_id} for {CONNECTOR_NAME} \
+                             connector ID: {}: {poll_error}",
+                            self.id
+                        );
+                        break;
```
mattp5657: A non-recoverable, non-`InvalidOffset` poll error `break`s the per-partition loop instead of `continue`-ing, permanently starving every partition ordered after the failing  iteration always restarts from the front next cycle, so a persistent per-partition error (revoked ACL, corrupt segment) blocks the same later partitions indefinitely.

Think another good example to add to integration tests.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,908 @@
...
+            .unwrap_or_else(|| {
+                warn!(
+                    "Invalid initial offset {:?} for {CONNECTOR_NAME} connector ID: {id}, \
+                     defaulting to earliest",
+                    config.initial_offset
+                );
+                InitialOffset::Earliest
+            });
+
+        let batch_size = config.batch_size.unwrap_or(DEFAULT_BATCH_SIZE);
```
mattp5657, line 190: `batch_size` has no validation, unlike `initial_offset` right above it. Any `u32` is accepted as-is.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,908 @@
...
+                self.id
+            );
+            sleep(delay).await;
+        }
+
+        let consumer = Consumer::default();
+
+        let mut candidate_state = self.state.lock().await.clone();
+
+        let mut messages = Vec::with_capacity(self.partitions.len() * self.batch_size as usize);
```
mattp5657: A large-but-valid value triggers an allocation failure, which aborts the process. Since every connector plugin is `dlopen`ed into one shared process, this takes down every other running connector too, not just this one.

mattp5657 (commented):
2 more things as well:

1. This connector is missing from the release/version-bump tooling. Absent from `scripts/bump-version.sh`, the `connector_plugins` default in `_build_rust_artifacts.yml`, and the connector list in `edge-release.yml`. It'll pass CI but won't ship in edge or tagged release artifacts, and a future version bump will silently skip it.

2. No test combines a multi-partition topology with a poll-time error. Might be valuable to have a test exercising multiple partition failures like this together in one integration test.

core/connectors/sources/iggy_source/src/lib.rs
jiengup: The latest commit determines the object to be blocked (the entire polling or a specific partition) based on different error types.

core/connectors/sources/iggy_source/src/lib.rs
jiengup: A new configuration option has been added in the latest commit to determine whether to block the partition or discard the message header when message construction fails.

jiengup:
@mattp5657 
It's my first contribution to the connector module; sorry for the missing care in the release/version-bump configuration. It's fixed in the latest commit.

More tests have been added to cover the error branch of this connector implementation.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,1072 @@
...
+            .get_topic(&stream_id, &topic_id)
+            .await
+            .map_err(|e| Error::InitError(format!("Failed to fetch upstream topic details: {e}")))?
+            .ok_or_else(|| {
+                Error::InitError(format!(
+                    "Upstream topic '{}/{}' not found after creation",
+                    self.config.upstream_stream, self.config.upstream_topic
+                ))
+            })?;
+        self.partitions = topic
```
hubcio, line 341: critical: if the upstream topic is re-created, saved offsets point past the new frontier, so every poll returns an empty batch with no error and that history is never copied. store the topic `created_at` and clear the offsets on change.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,1072 @@
...
+    include_user_headers: bool,
+    malformed_message_policy: MalformedMessagePolicy,
+) -> PartitionBuildOutcome {
+    let mut produced_messages = Vec::with_capacity(messages.len());
+    let mut last_produced_offset = None;
+    let mut errors = Vec::new();
+
+    for message in messages {
+        let headers = if include_user_headers {
+            match message.user_headers_map() {
```
hubcio, line 594: warning: header bytes that fail to parse make `user_headers_map` return `Ok(None)`, so the message ships without headers and the offset advances. a present `user_headers` with a `None` map is that case, so treat it as a build error and apply `malformed_message_policy`.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,1072 @@
...
+                    }
+
+                    messages.extend(partition_messages);
+                    if let Some(last_produced_offset) = last_produced_offset {
+                        candidate_state
+                            .offsets
+                            .insert(partition_id, last_produced_offset);
+                    }
+                }
+                Err(IggyError::InvalidOffset(offset)) => {
```
hubcio, line 453: warning: this arm can never fire - the server only raises `InvalidOffset` from the consumer-offset store path, and the connector keeps its offsets in its own state file. delete the arm and the README:98 reset claim, or make the reset real.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,1072 @@
...
+
+    async fn ensure_stream_and_topic(
+        &self,
+        client: &IggyClient,
+        stream_id: &Identifier,
+        topic_id: &Identifier,
+    ) -> Result<(), Error> {
+        match client.get_stream(stream_id).await {
+            Ok(Some(_)) => {}
+            Ok(None) => {
```
hubcio, line 231: warning: if the name is already taken (code 1012 for stream, 2013 for topic) `open()` fails and nothing retries, so a lost race leaves the connector down until an operator restarts it. treat already-exists from create as success.

core/connectors/sources/iggy_source/src/lib.rs
```diff
@@ -0,0 +1,1072 @@
...
+                self.id,
+                messages.len(),
+                self.partitions.len(),
+                total_synced,
+                cycle_counts.upstream_poll_errors,
+                cycle_counts.conversion_errors
+            );
+        }
+
+        Ok(ProducedMessages {
```
hubcio, line 513: warning: the poll returns state on every cycle, so an idle connector writes and syncs the same bytes on each poll interval. return `state: None` when nothing changed, like `postgres_source` does.

core/connectors/sources/iggy_source/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+- `earliest` starts with the oldest available message.
+- `latest` requests up to `batch_size` of the most recent messages.
+- A numeric value starts at that absolute offset.
+
+The connector stages new offsets while polling. It commits them only after the
+runtime sends the complete downstream batch and saves the connector state. If
+the batch is rejected, the previous offsets remain committed and the messages
+are eligible for replay. Consumers should therefore tolerate duplicate
+delivery after failures or crashes.
```
hubcio, line 90: warning: this replay only happens while the source is running - five consecutive send failures stop it and the runtime never respawns it, so an outage longer than that needs a manual restart.

hubcio (changes requested):
one note on the description, since it sits outside the diff: the crash-recovery table's downstream-outage row. a rejected batch is not dropped for the process lifetime - the next poll replays it from the last committed offset. what actually ends the polling is the sdk stopping the source after five consecutive nacks, and the runtime does not respawn it, so a long outage needs a manual restart. #3855 already landed.

jiengup:
@hubcio Fixing. Will be in the next commit

PR #3888, opened 2026-08-15
fix(python): read consumer metadata without the consumer lock

ethanlin01x:

Relates to #3776 (Found while reviewing)

## Rationale

The synchronous getters on the Python `IggyConsumer` took the consumer mutex with `blocking_lock()` while holding the GIL, and `consume_messages` holds that mutex for the whole consumption run. Reading an attribute during consumption hung the interpreter; reading one from a callback panicked inside the Tokio runtime. Neither is recoverable from Python.

```python
consume = consumer.consume_messages(handle, shutdown_event)
print(consumer.name())   # never returns
```

## What changed?

None of those getters need exclusive access. 

`name`, `stream` and `topic` never change after construction, so the Python wrapper snapshots them. The partition id and offsets do change, but already live behind `Arc`s in the Rust SDK, which now hands them out as `IggyConsumerState` via `IggyConsumer::state()`. The wrapper keeps a clone and reads them as atomic loads, never taking the lock.

Additive on the Rust side. On the Python side `stream()` and `topic()` now return the identifier directly instead of a `PyResult`.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? Claude
2. Scope of usage? help implement and write PR description
3. How did you verify the generated code works correctly?  Ran the Python test suite against a real server


slbotbm (dismissed):

hubcio (approved):

core/sdk/src/clients/consumer.rs
```diff
@@ -486,8 +526,8 @@ impl IggyConsumer {
         let consumer = self.consumer.clone();
         let stream_id = self.stream_id.clone();
         let topic_id = self.topic_id.clone();
-        let last_consumed_offsets = self.last_consumed_offsets.clone();
-        let last_stored_offsets = self.last_stored_offsets.clone();
+        let last_consumed_offsets = self.state.last_consumed_offsets.clone();
```
hubcio: the loop below holds a dashmap shard read guard across `store_consumer_offset(...).await`, and the poll future can `insert` a new partition into the same map (line 776). that insert blocks synchronously inside an async task for a whole round trip - stalled worker normally, real deadlock with a single worker (`#[tokio::test]` defaults to current_thread). collect `(partition_id, offset)` into a vec first, then await.

core/sdk/src/clients/consumer.rs
```diff
@@ -91,6 +91,47 @@ pub enum AutoCommitAfter {
     ConsumingEveryNthMessage(u32),
 }
 
+/// A cheap, cloneable view of the state shared with an [`IggyConsumer`].
+///
+/// Consuming borrows the consumer as `&mut` for the whole run, so reading its getters
+/// concurrently means sharing it behind a lock and then waiting on that lock. This view
+/// carries the same shared state and needs neither.
+#[derive(Clone)]
```
hubcio, line 104: public through the prelude with no `Debug` - `#[derive(Clone, Debug)]` compiles. minor, `IggyConsumer` below has none either.

core/sdk/src/clients/consumer.rs
```diff
@@ -91,6 +91,47 @@ pub enum AutoCommitAfter {
...
+    fn new() -> Self {
+        Self {
+            current_partition_id: Arc::new(AtomicU32::new(0)),
+            last_consumed_offsets: Arc::new(DashMap::new()),
+            last_stored_offsets: Arc::new(DashMap::new()),
+        }
+    }
+
+    /// Returns the current partition ID of the consumer.
+    pub fn partition_id(&self) -> u32 {
```
hubcio, line 153: worth saying these aren't a combined snapshot - two independent loads, so the partition id can be stale by the time you read an offset for it.

foreign/python/src/consumer.rs
```diff
@@ -44,10 +44,16 @@ use crate::receive_message::ReceiveMessage;
 
 /// A Python class representing the Iggy consumer.
 /// It provides asynchronous functionality through the contained runtime.
+// `inner` stays locked for the whole duration of a consumption run, so the synchronous
```
hubcio: `store_offset` and `delete_offset` still lock `inner`, so awaiting either from inside a `consume_messages` callback deadlocks - the run holds the guard while awaiting the callback. with `AutoCommit.Disabled()` that leaves no manual commit path from a callback at all. both take `&self` on the rust side, so the mutex isn't needed - same trick as the getters here.

foreign/python/src/consumer.rs
```diff
@@ -56,39 +62,33 @@ impl IggyConsumer {
     /// Get the last consumed offset or `None` if no offset has been consumed yet.
```
hubcio: `None` means the partition isn't tracked yet, not that nothing was consumed - the poll future seeds 0 on first sight of a partition.

foreign/python/src/consumer.rs
```diff
@@ -56,39 +62,33 @@ impl IggyConsumer {
...
     #[gen_stub(override_return_type(type_repr = "builtins.int | None"))]
     fn get_last_consumed_offset(&self, partition_id: u32) -> Option<u64> {
-        self.inner
-            .blocking_lock()
-            .get_last_consumed_offset(partition_id)
+        self.state.get_last_consumed_offset(partition_id)
     }
 
     /// Get the last stored offset or `None` if no offset has been stored yet.
```
hubcio: same - after any poll this returns 0 even with `AutoCommit.Disabled()` and nothing stored server-side, which the new test at line 921 pins. don't change the sentinel though, `store_consumer_offset` reads 0 as "nothing stored".

foreign/python/src/consumer.rs
```diff
@@ -56,39 +62,33 @@ impl IggyConsumer {
...
     #[gen_stub(override_return_type(type_repr = "builtins.int | None"))]
     fn get_last_stored_offset(&self, partition_id: u32) -> Option<u64> {
-        self.inner
-            .blocking_lock()
-            .get_last_stored_offset(partition_id)
+        self.state.get_last_stored_offset(partition_id)
     }
 
     /// Gets the name of the consumer group.
     fn name(&self) -> String {
```
hubcio: return `&str` and skip the clone - the generated stub comes out identical.

foreign/python/src/consumer.rs
```diff
@@ -56,39 +62,33 @@ impl IggyConsumer {
...
+        self.name.clone()
     }
 
     /// Gets the current partition id or `0` if no messages have been polled yet.
     fn partition_id(&self) -> u32 {
-        self.inner.blocking_lock().partition_id()
+        self.state.partition_id()
     }
 
     /// Gets the name of the stream this consumer group is configured for.
```
hubcio: also line 89. these return `str | int`, not a name - a numeric-looking name becomes a numeric id, so `consumer_group("g", "42", "t")` gives back the int 42. say identifier.

foreign/python/tests/test_consumer_group.py
```diff
@@ -856,6 +857,74 @@ async def test_consumer_group_metadata(self, iggy_client: IggyClient, unique_nam
...
+
+        consume = consumer.consume_messages(take, shutdown_event)
+        try:
+            await asyncio.wait_for(consuming.wait(), timeout=10)
+
+            # A getter that blocks holds the GIL, so neither pytest-timeout nor asyncio
+            # can fire. The faulthandler watchdog needs no GIL and aborts instead.
+            # A regression hangs forever, so the timeout only has to outlast normal GIL
+            # contention -- it is generous because tripping it kills the whole run.
+            faulthandler.dump_traceback_later(5, exit=True)
```
hubcio: `exit=True` takes the whole session down, and there's no alternative - a gil-pinned deadlock can't be recovered in process. pytest has it built in though: `faulthandler_timeout` + `faulthandler_exit_on_timeout` in pyproject gives the same thing suite-wide and drops these four lines.

hubcio (changes requested):
fix is good, some polishing is needed tho.

three that don't fit on a diff line:

- `foreign/python/src/consumer.rs:410-411`: the `lock_owned()` guard is shadowed, not dropped, so it spans the whole `consume()` including the callback await. harmless today, but `self.task_locals.lock().await.clone()` drops it.
- `foreign/python/src/consumer.rs:202`: `tokio::join!` waits on the shutdown task even when consume already exited, and that task waits forever on `Event.wait()`. the new test is only safe because it sets `shutdown_event` in `finally`.
- nothing regenerates and diffs `apache_iggy.pyi` in ci - php has that gate, python doesn't. no impact here, separate ticket.

slbotbm:
@hubcio do we want to check the pyi file in ci? i ask because this check will require a complete recompilation of the sdk, and our ci runs are very heavy as it is.

maybe we could instead redesign the ci for ffi sdks instead?

hubcio:
@slbotbm like we discussed on discord, please propose something via issues or discussions.

core/sdk/src/clients/consumer.rs
ethanlin01x: Fixed in 1ae5a4cc1 — collected into a Vec before awaiting.

core/sdk/src/clients/consumer.rs
ethanlin01x: Added a manual Debug impl in 984494952 instead of derive, so the client lock and maps don't end up in the output.

core/sdk/src/clients/consumer.rs
ethanlin01x: Doc updated in 984494952 to say these are independent loads, not a snapshot.

foreign/python/src/consumer.rs
ethanlin01x: Both moved onto IggyConsumerState in b0d3c20c0, no lock taken from callbacks anymore.

foreign/python/src/consumer.rs
ethanlin01x: Docs and stubs clarified in 984494952

foreign/python/src/consumer.rs
ethanlin01x: docs clarified in 984494952

foreign/python/src/consumer.rs
ethanlin01x: Done in 984494952

foreign/python/src/consumer.rs
ethanlin01x: Reworded to identifier in 984494952

foreign/python/tests/test_consumer_group.py
ethanlin01x: Moved to pyproject in c5aa65031: faulthandler_timeout + faulthandler_exit_on_timeout.

ethanlin01x:
@hubcio Thanks for the review! Also fixed a clippy error while at it. 
For the apache_iggy.pyi CI gate, I'll dig into it and open a separate issue for discussion.

hubcio (approved):

spetz (approved):

PR #3891, opened 2026-08-16
refactor(sdk): introduce NonZeroIggyDuration for retries and heartbeats

ethanlin01x:

Relates to #3776  (Found while reviewing)

## Rationale

Zero-valued durations reach loops that spin or panic. A zero retry interval panics tokio::time::interval, and a zero heartbeat interval turns the heartbeat task into a continuous ping loop. IggyDuration::from_str maps 0, none, disabled and unlimited to the same zero, so a user asking to turn something off arrives there.

A duration that paces a loop is a different thing from a duration that measures a delay, so it gets its own type. NonZeroIggyDuration rejects the zero where the value is built, which means no call site has to remember the rule and every binding that funnels through the Rust SDK gets the guarantee.

## What changed?

New type `NonZeroIggyDuration` (`core/common/src/utils/duration.rs`): `TryFrom<IggyDuration>`, `TryFrom<u64>`, `TryFrom<Duration>`, `FromStr`, `Display` and serde, all rejecting zero and the three aliases that parse as zero. Serialization emits whole microseconds, so a duration shorter than that is rejected as well. The error is a typed `NonZeroDurationError`.

Fields that now hold it:

- `heartbeat_interval` on the TCP, QUIC, WebSocket and HTTP client configs and connection string options, plus `BinaryTransport::get_heartbeat_interval` and `SystemClient::heartbeat_interval`. The field is new on `HttpClientConfig`, where the connection string option was parsed and thrown away.
- `reconnection.interval` on all three transports
- `IggyConsumer` `init_retry_interval` and `polling_retry_interval`
- `ProducerCore` `send_retries_interval`
- `AutoCommit::Interval`, `AutoCommit::IntervalOrWhen` and `AutoCommit::IntervalOrAfter`

Zero stays legal where it means something, and this is now pinned by tests: `reestablish_after` (zero reconnects immediately), `poll_interval` (zero means no throttling) and the background producer `linger_time`.

The Python and PHP bindings follow the same types.

## Breaking changes

- The config fields above change type. Construction becomes `NonZeroIggyDuration::from_str("5s")?` instead of `IggyDuration::from_str("5s")?`.
- A connection string with `heartbeat_interval=none|disabled|unlimited|0` or `reconnection_interval=0` now fails with `InvalidConnectionString` instead of silently spinning. This includes the HTTP connection string, where the option was parsed and discarded.
- A zero `AutoCommit` interval is now rejected. The three interval variants take `NonZeroIggyDuration`, so Rust rules it out at compile time and the Python and PHP bindings raise at the call. Zero stored the offsets in a busy loop; committing on every poll is spelled `AutoCommit::When` or `AutoCommit::After`.
- Python: `TcpReconnectionConfig(interval=timedelta(0))` is now rejected whatever the retry policy is.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? Claude
2. Scope of usage? help implement and review this PR
3. How did you verify the generated code works correctly? Unit tests pin each rejected value and each still-legal zero


haubur:
The problem is real for consumers:
- panic on a 0 IggyDuration: https://github.com/apache/iggy/blob/ec456d2370de25f6585dc245077a487d9cdf7730/core/sdk/src/clients/consumer.rs#L328
- and (not) sleeping with 0 on the background store offset at interval task loop. : https://github.com/apache/iggy/blob/ec456d2370de25f6585dc245077a487d9cdf7730/core/sdk/src/clients/consumer.rs#L496

producers:
- panic on a 0 IggyDuration: https://github.com/apache/iggy/blob/ec456d2370de25f6585dc245077a487d9cdf7730/core/sdk/src/clients/producer.rs#L265

and connection lifecycle:
- continuous heartbeat spins: https://github.com/apache/iggy/blob/f42f5abc855f5d6f9906373e9a5b74a8bcb96d9d/core/sdk/src/clients/client.rs#L298

Clarification:
- The problem occurs if **manually** setting a IggyDuration equal to zero.
- **No default path** actually sets an IggyDuration with zero. 
- panics occur for IggyDuration in `retry_interval`s not for `heartbeat_interval`s 

Solution:
- Having IggyDuration falling to 0 for "unlimited", "disabled" and "none" is a little misleading.
- Since it's a common type there might be unobserved behavior in other places should we change it now.
- To avoid panics, **I would prefer not checking in methods calls as done in the PR but enforcing the proposed guards on a type level e.g. by introducing a NonZeroIggyDuration.**

ethanlin01x:
@haubur 

Thanks for the review.

I agree that using `NonZeroIggyDuration` is cleaner than checking in each method. My concern is that some of these are cross-field rules, so I am not sure a type can enforce them. For example, `interval = 0` is still valid when `max_retries` is bounded. Changing the field type would also be a breaking change for the SDK API.

So I would prefer to keep the current approach. What do you think?

haubur:
@ethanlin01x fair points.
- A `NonZeroIggyDuration` assumes that 0 intervals are actually **undesirable,** which I would argue makes sense. I think its reasonable to enforce waiting a bit before retrying (instead of rapid fire). You can still set the duration to some very low time period if you want that, but that's intended behavior than not implicit as it is now.
- It would be a breaking API change, true. If thats something we want at this point is a question to the maintainers I guess, because there is a release coming up this week AFAIK.

spetz:
@haubur breaking changes are totally fine at this stage, as we're very close to the next release, with lots of breaking changes anyway :)
And yes, I do agree, there should be at least like a 1 ms wait before the retries (while on the other hand, no wait at all for data producer/consumer running as a busy loop is totally fine for specific use cases).

haubur:
Thanks @spetz! So @ethanlin01x lets have a NonZeroIggyDuration for retries and heartbeats, while keeping the busy loop for AutoCommit on an interval = 0. Would you want to rewrite? Otherwise, I can also take this.

ethanlin01x:
@haubur Thanks, I just rewrote the whole PR to introduce `NonZeroIggyDuration`

core/common/src/utils/non_zero_duration.rs
```diff
@@ -0,0 +1,255 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
haubur: This does not need to be an additional module. Put it in core/common/src/utils/duration.rs.

core/common/src/utils/non_zero_duration.rs
```diff
@@ -0,0 +1,255 @@
...
+use std::{
+    error::Error,
+    fmt::{Display, Formatter},
+    str::FromStr,
+    time::Duration,
+};
+
+/// A duration that is guaranteed to be greater than zero.
+///
+/// Intervals that pace a loop - heartbeats, reconnection and retry delays - turn into
```
haubur: This block can go. It was the reason for introducing it, but it is not limited to that context and can be confusing.

core/common/src/utils/non_zero_duration.rs
```diff
@@ -0,0 +1,255 @@
...
+/// The reason a value could not become a `NonZeroIggyDuration`.
+#[derive(Debug, Clone, PartialEq)]
+pub enum NonZeroDurationError {
+    /// The value parsed or converted to zero.
+    Zero,
+    /// The text is not a duration `humantime` understands.
+    InvalidFormat(humantime::DurationError),
+}
+
+impl NonZeroIggyDuration {
```
haubur: Align the API of this type with `IggyDuration` what is missing:
- new()
- as_secs_f64()
- ads_diff()

core/common/src/utils/non_zero_duration.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+}
+
+impl From<humantime::DurationError> for NonZeroDurationError {
+    fn from(error: humantime::DurationError) -> Self {
+        NonZeroDurationError::InvalidFormat(error)
+    }
+}
+
+impl TryFrom<IggyDuration> for NonZeroIggyDuration {
```
haubur: Also implement
- From<std::time::Duration>
- From<humantime::Duration>
and
- Add
to match the `IggyDuration` API

haubur (changes requested):
Thanks! I add some minor comments, basically just location and making sure that the IggyDuration and NonZeroIggyDuration API will align.

core/common/src/utils/non_zero_duration.rs
ethanlin01x: Done in 9f07b53.

core/common/src/utils/non_zero_duration.rs
ethanlin01x: Removed in 9f07b53

core/common/src/utils/non_zero_duration.rs
ethanlin01x: Added in 71f3484

core/common/src/utils/non_zero_duration.rs
ethanlin01x: Added in 71f3484

haubur (approved):

ethanlin01x:
@haubur thanks, all four comments are addressed in 9f07b53 and 71f3484.

core/common/src/utils/duration.rs
```diff
@@ -226,6 +227,206 @@ impl Visitor<'_> for IggyDurationVisitor {
...
+impl From<humantime::DurationError> for NonZeroDurationError {
+    fn from(error: humantime::DurationError) -> Self {
+        NonZeroDurationError::InvalidFormat(error)
+    }
+}
+
+impl TryFrom<IggyDuration> for NonZeroIggyDuration {
+    type Error = NonZeroDurationError;
+
+    fn try_from(duration: IggyDuration) -> Result<Self, Self::Error> {
```
spetz, line 350: This accepts sub-microsecond durations such as `1ns`, but serialization emits whole microseconds and therefore produces `0`. Please reject values below one microsecond and add a serde round-trip test for `1ns`.

core/sdk/src/http/http_client.rs
```diff
@@ -279,7 +279,7 @@ impl HttpClient {
         Ok(Self {
             api_url,
             client,
-            heartbeat_interval: IggyDuration::from_str("5s").unwrap(),
+            heartbeat_interval: NonZeroIggyDuration::from_str("5s").unwrap(),
```
spetz: `heartbeat_interval` is parsed from the HTTP connection string, but `HttpClientConfig` does not carry it, so this still replaces every valid custom value with five seconds. Please add the interval to `HttpClientConfig` and propagate `connection_string.options().heartbeat_interval()` through the conversion.

core/sdk/src/clients/consumer.rs
```diff
@@ -1239,3 +1241,45 @@ impl Drop for IggyConsumer {
...
+            Identifier::numeric(1).unwrap(),
+            Identifier::numeric(1).unwrap(),
+            None,
+            None,
+            None,
+        )
+    }
+
+    #[tokio::test]
+    async fn should_accept_auto_commit_modes_with_a_zero_interval() {
```
spetz: This test explicitly preserves a busy background loop. A zero interval repeatedly executes `sleep(Duration::ZERO)`, which conflicts with the purpose of this PR. Please reject zero for interval-based auto-commit and use `When` or `After` for immediate commits.

foreign/php/src/client.rs
```diff
@@ -343,9 +343,8 @@ impl IggyClient {
             builder = builder.auto_commit(auto_commit.into());
         }
         builder = match poll_interval_micros {
-            Some(micros) => {
-                builder.poll_interval(non_zero_duration_micros("poll_interval_micros", micros)?)
-            }
+            Some(micros) => builder
+                .poll_interval(non_zero_duration_micros("poll_interval_micros", micros)?.get()),
```
spetz: This rejects a zero polling interval even though the Rust and Python APIs allow it. Please use `IggyDuration::from(micros)` here and keep the nonzero conversion only for retry intervals.

spetz (changes requested):
Looks solid overall, just a few minor improvements left :)

core/common/src/utils/duration.rs
ethanlin01x: `TryFrom<IggyDuration>` now rejects anything below one microsecond with a new `NonZeroDurationError::SubMicrosecond`, and every constructor goes through it. Added tests for `1ns` and for a `1us` serde round trip. Fixed in 3580913e7

core/sdk/src/http/http_client.rs
ethanlin01x: `HttpClientConfig` now has a `heartbeat_interval` field (default 5s) that is filled from the connection string, and `HttpClient::create` reads it instead of hardcoding 5s. Also added a builder setter and a test with `heartbeat_interval=10s`. Fixed in 7c67c9ad1.

foreign/php/src/client.rs
ethanlin01x: Fixed in b5a79bcc0

core/sdk/src/clients/consumer.rs
ethanlin01x: The three interval variants of `AutoCommit` now take `NonZeroIggyDuration`, so a zero interval does not compile any more. Fixed in 62a7fd7e1.

core/common/src/utils/duration.rs
```diff
@@ -226,6 +227,221 @@ impl Visitor<'_> for IggyDurationVisitor {
...
+        if duration.is_zero() {
+            return Err(NonZeroDurationError::Zero);
+        }
+
+        // Serialization emits whole microseconds, so a shorter duration would come back as zero.
+        if duration.as_micros() == 0 {
```
spetz: `IggyDuration::as_micros()` already truncates the underlying `u128` value to `u64`. A duration of `2^64 + 1` microseconds passes this check but serializes as `1`, while exactly `2^64` is incorrectly classified as `SubMicrosecond`. Please compare the underlying `Duration` against `1us` here and make serialization reject values above `u64::MAX` microseconds, with boundary tests.

spetz (changes requested):

core/common/src/utils/duration.rs
ethanlin01x: Fixed in [46f9500](https://github.com/apache/iggy/pull/3891/commits/46f9500f9b16e043bcebbf1eba039d08fa493314)

spetz (approved):

numinnex (approved):

PR #3894, opened 2026-08-16
feat(csharp): add heartbeat mechanism

### LINKED ISSUE #3889: csharp: idle clients cannot survive server-side stale-session eviction
The .NET SDK has no mechanism to keep an idle session alive, and no way to recover once the server drops it.

## The client never sends traffic on its own

`PingAsync` is declared on `IIggySystem` and implemented in `Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs` and `Iggy_SDK/IggyClient/Implementations/HttpMessageStream.cs`, but it is only ever called by application code. Nothing inside the SDK schedules it. The only periodic timer in the SDK is the batching flush in `Iggy_SDK/Publishers/BackgroundMessageProcessor.cs`, and that fires work only when there are messages queued to publish. An application that connects, logs in, and then waits on an idle consumer or an idle publisher puts nothing on the wire.

## Recovery is disabled by default

`Iggy_SDK/Configuration/ReconnectionSettings.cs` sets `Enabled` to `false`, and `Iggy_SDK/Configuration/IggyClientConfigurator.cs` takes that default as is. With reconnection disabled, `SendWithResponseAsync` in `TcpMessageStream.cs` logs "Reconnection is disabled", moves the connection to the disconnected state, and rethrows.

`Iggy_SDK/Configuration/AutoLoginSettings.cs` also defaults `Enabled` to `false`. It is only turned on for builder-owned clients that were given credentials, so a client constructed directly comes back unauthenticated even if a reconnect does happen.

## Why this matters

Once the server enables heartbeat-based verification and reaps sessions it has not heard from, an idle .NET client has no way to notice or repair the loss. The eviction is silent from the client's point of view, the next call fails, and the application is left to rebuild the client and log in again by hand.

Other SDKs already ping on a timer. The Rust client starts a heartbeat task inside `IggyClient::connect` (`core/sdk/src/clients/client.rs`) with a five second default, the Java client schedules one in `AsyncTcpConnection`, the Go client runs a heartbeat loop started from `client/iggy_client.go` with a five second default, the Python client exposes `heartbeat_interval` with a five second default, and the C++ SDK inherits the Rust behaviour through its bridge.

## Requests

- Add an idle ping task to the .NET client, driven by a configurable interval that is enabled by default and sits comfortably below the server's stale-session threshold.
- Reconsider the reconnection defaults. A client that neither pings nor reconnects has no path back from a dropped session, which is a surprising default for a long-lived connection.

lukaszzborek:
- TCP client heartbeat: PeriodicTimer loop pings server every HeartbeatInterval (default 5s, new option on IggyClientConfigurator). Ping goes through normal request path, so reconnect + auto login repair dead session. Skips ping when connecting or when reconnect/auto-login disabled and connection dead. Started under connect gate, no duplicate loops.
- AutoLoginSettings: init-only, PersonalAccessToken support, public For(user, pwd) / ForPersonalAccessToken(token) factories.
- Consumer/Publisher builders: heartbeat + auto-login wiring, warning when reconnect enabled without auto login.
- Enable reconnection by default


hubcio:
@lukaszzborek can you change default heartbeats to enabled in config.toml for server?

spetz (approved):

numinnex (approved):

PR #3897, opened 2026-08-16
fix(server): restore get_user self-read without read_users permission

hubcio:
The legacy server skipped the read_users check when a user fetched
its own account. The server-ng promotion (#3856) gated GET_USER
unconditionally on the binary path, and the HTTP route kept the
bypass only for the "me" alias, so a user without read_users could
no longer fetch itself by id or name. Reported under #3733.

Resolve the target against the committed users STM in the dispatch
gate and the HTTP route; a target resolving to the caller passes
without a permissioner rule, any other target still requires
read_users. A resolution miss keeps the notfound-before-permission
ordering, matching legacy and the other identifier-scoped reads.

### DIFF
--- a/core/integration/tests/server/scenarios/permissions_scenario.rs
+++ b/core/integration/tests/server/scenarios/permissions_scenario.rs
@@ -235,6 +235,24 @@ async fn test_no_permissions(harness: &TestHarness, root_client: &IggyClient) {
         "snapshot must be rejected as unauthorized without read_servers, got {snapshot_result:?}"
     );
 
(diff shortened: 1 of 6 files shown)


spetz (approved):

slbotbm (approved):

amlel-el-mahrouss:
Just reviewed the changes per my message on the Discord, looks good to me.
**Note**: For future revisions of permissions_scenario.rs, we need a less error-prone approach when expanding the file.

PR #3913, opened 2026-08-18
chore(sdk): IggyConsumer docs

### LINKED ISSUE #3912: chore(Rust SDK): IggyConsumer docs
### Description

In an effort to add and improve the Rust SDK docs (first issue: https://github.com/apache/iggy/issues/3306 and first PR here: https://github.com/apache/iggy/pull/3809) a follow up for the IggyConsumer. 

### Affected area / component

Rust SDK

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Contribution

### Good first issue

haubur:
Closes #3912
Relates #3306

## Local Execution
- Pre-commit hooks ran

## AI Usage
None


core/sdk/src/clients/consumer.rs
```diff
@@ -100,6 +100,316 @@ pub enum AutoCommitAfter {
...
+/// | [`AutoCommitWhen::ConsumingEveryNthMessage`] | when the offset of a message handed over divides by `n` |
+/// | [`AutoCommitWhen::ConsumingAllMessages`] | when the buffer of the current batch runs empty |
+/// | [`AutoCommitAfter`] variants | as their [`AutoCommitWhen`] counterparts, but once the handler returned, and only under [`IggyConsumerMessageExt::consume_messages`] |
+///
+/// [`AutoCommit::IntervalOrWhen`] and [`AutoCommit::IntervalOrAfter`] combine an interval with a
+/// message trigger. The default is [`AutoCommit::IntervalOrWhen`] with one second and
+/// [`AutoCommitWhen::PollingMessages`].
+/// Important implications of these defaults:
+/// - [`AutoCommitWhen::PollingMessages`] marks a batch as consumed while it is being delivered,
+///   before your code has seen any of it. If a crash must not skip messages, commit after handling
```
hubcio: this advice doesn't hold - `ConsumingEachMessage` commits at handover, inside `poll_next` before your handler runs (the store fires at 1442-1447 / 1557-1567 and lands concurrently with the handler). a crash mid-handling can still skip the message. crash-safe options are `AutoCommit::Disabled` + `store_offset()` after handling, or the `AutoCommitAfter` variants under `consume_messages`.

core/sdk/src/clients/consumer.rs
```diff
@@ -100,6 +100,316 @@ pub enum AutoCommitAfter {
...
+/// - [`AutoCommitWhen::ConsumingEveryNthMessage`] tests the offset of a message, not a counter of
+///   messages this process handled, so it commits at every `n`-th offset of the partition.
+///
+/// ## Guarantees
+///
+/// - **Each message is handed over once per consumer.** Messages whose offset is not greater than
+///   the reading position of their partition are dropped before they reach the stream. Re-reading
+///   a partition, or letting a message come back because your handler failed, needs
+///   [`allow_replay()`], which turns that filter off.
+/// - **Delivery is at-least-once.** A crash between handling a message and committing its offset
```
hubcio: "at-least-once" isn't true for the default config - `PollingMessages` commits the whole batch as part of the poll request, before your code sees any of it, so a crash mid-batch skips the rest. same for `ConsumingEachMessage` (see above). at-least-once only holds when committing after handling - worth qualifying, since this is the guarantees section.

core/sdk/src/clients/consumer.rs
```diff
@@ -100,6 +100,316 @@ pub enum AutoCommitAfter {
...
+///         break;
+///     }
+/// }
+///
+/// consumer.shutdown().await?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// Committing by hand, so that a message the handler could not process comes back:
```
hubcio: the example doesn't deliver this promise - if message N fails but N+1 succeeds, `store_offset(N+1)` marks N consumed (it's a cumulative watermark), so N never comes back, in-process or after restart. it only works if you stop on failure. either break on error or call out the cumulative-commit caveat.

core/sdk/src/clients/consumer.rs
```diff
@@ -297,16 +639,79 @@ impl IggyConsumer {
...
+    /// - The offset store task is spawned in any case. It can be configured with [`AutoCommitWhen::ConsumingEachMessage`],
+    ///   [`AutoCommitWhen::ConsumingEveryNthMessage`], [`AutoCommitWhen::ConsumingAllMessages`] and
+    ///   their [`AutoCommitAfter`] counterparts. Under [`AutoCommit::Disabled`] nothing is
+    ///   ever sent and the task stays idle.
+    ///
+    /// A variant such as [`AutoCommit::IntervalOrWhen`] runs both together. The message count
+    /// trigger stores as messages are consumed, the interval stores what the trigger has not
+    /// covered yet. There is no double-work, since an offset that is not ahead of the one last stored
+    /// for that partition is skipped instead of sent.
+    /// Unless `allow_replay` is enabled an offset that is not past the last stored offset on the server
```
hubcio: `allow_replay` doesn't affect the auto-commit paths at all - both background tasks hardcode it to `false` (lines 840 and 940). it only applies to manual `store_offset()` and the shutdown flush. also the skip check reads the in-memory `last_stored_offsets`, not the server.

core/sdk/src/clients/consumer.rs
```diff
@@ -1128,36 +1588,66 @@ impl Stream for IggyConsumer {
...
 impl IggyConsumer {
+    /// Shuts the consumer down.
+    ///
+    /// Specifically, run shutdown and await before dropping the consumer to
+    /// - finish storing the offsets that are currently in-flight.
+    ///   There are two background tasks that can have commits in flight. The interval-based one
+    ///   (only spawned for [`AutoCommit`] variants that carry an interval) and the one driven by
+    ///   [`AutoCommitWhen`]/[`AutoCommitAfter`] (always spawned). The consumer waits for
+    ///   `offset_drain_timeout` on each in turn before forcing it to abort.
+    ///   Any offset that is not stored until then will be lost.
```
hubcio: not quite - the final flush below (1668-1695) commits every partition where consumed > stored, and everything the aborted tasks had queued is covered by that. offsets are only lost when the flush request itself fails (errors are swallowed at 1683).

core/sdk/src/clients/consumer.rs
```diff
@@ -1128,36 +1588,66 @@ impl Stream for IggyConsumer {
...
         if self.shutdown.swap(true, ORDERING) {
             return Ok(());
         }
 
         info!("Shutting down consumer: {}...", self.consumer_name);
 
-        // Drain the background commit tasks while still a group member,
-        // before leaving below — otherwise a store they send afterward hits
-        // a group we've already left.
+        // Wake the task responsible for storing the offsets (spawned in store_offset_in_background())
```
hubcio: typo: the fn is `store_offsets_in_background()`. also the old comment here explained why the drain happens before leaving the group (a store sent after the leave would hit a group we already left) - that ordering constraint, and the abort-so-no-stale-store rationale below, are worth keeping.

core/sdk/src/clients/consumer.rs
```diff
@@ -1128,36 +1588,66 @@ impl Stream for IggyConsumer {
...
-            // Still running past the bound: abort it rather than leaving it
-            // detached, so it can't send a stale store after we leave below.
             task.abort();
             warn!(
                 "Timed out waiting for the background offset-commit task to stop for consumer: {}, aborted",
                 self.consumer_name
             );
         }
 
+        // Drop the sending end of the store offset task to end the `recv_async()` loop in `send_store_offset().
```
hubcio: the `recv_async()` loop lives in the task spawned in `init()`, not in `send_store_offset()` - that's the sender side. unclosed backtick too.

core/sdk/src/clients/consumer.rs
```diff
@@ -404,6 +821,9 @@ impl IggyConsumer {
         let (store_offset_sender, store_offset_receiver) = flume::unbounded();
         self.store_offset_sender = store_offset_sender;
 
+        // The IggyClients `poll_next` implementation sends store offset requests down to this receiver.
```
hubcio: `poll_next` is `IggyConsumer`'s `Stream` impl, not `IggyClient`'s. `consume_messages` also sends into this channel.

core/sdk/src/clients/consumer.rs
```diff
@@ -974,13 +1412,24 @@ impl ReceivedMessage {
     }
 }
 
+/// Yields messages one at a time.
+///
+/// Tries the buffer first, before a new batch is fetched from the server and stored in the buffer.
+///
+/// The stream never yields `None`. So a `while let Some(..)` loop over it runs
+/// until the loop body breaks out. Errors are yielded as items and do not end the stream, polling
+/// again retries. See the [type documentation](IggyConsumer#polling) for the details of polling.
```
hubcio: there's no `#polling` anchor - the section is "How messages are read", so `#how-messages-are-read`. rustdoc doesn't check fragments, so this stays green but lands at the top of the page.

core/sdk/src/clients/consumer.rs
```diff
@@ -229,26 +539,48 @@ impl IggyConsumer {
...
 
-    /// Stores the consumer offset on the server either for the current partition or the provided partition ID.
+    /// Stores an offset on the server, marking every message up to and including it as consumed.
+    ///
+    /// This is the manual counterpart to [`AutoCommit`] and is meant for
+    /// [`AutoCommit::Disabled`].
+    ///
+    /// Pass `None` as `partition_id` to use the partition of the most recent batch polled.
+    ///
+    /// An offset that is not ahead of the last one this consumer stored for that partition is
```
hubcio, line 799: small carve-out: the skip check is `offset <= stored_offset && offset >= 1` (line 876), so offset 0 is always sent. and the server allows an explicit store to rewind the watermark, so `store_offset(0)` actually rewinds even without `allow_replay`.

core/sdk/src/clients/consumer.rs
```diff
@@ -229,26 +539,48 @@ impl IggyConsumer {
...
-    /// Returns the stream ID of the consumer.
+    /// Returns the identifier of the stream this consumer reads from.
     pub fn stream(&self) -> &Identifier {
         &self.stream_id
     }
 
-    /// Returns the current partition ID of the consumer.
+    /// Returns the partition the last message came from.
+    ///
+    /// This is `0` until the first message has been read, because a partition is only known once
```
hubcio: it's set on any poll answer, even an empty one (stored at 1492 before the empty check), so it can be nonzero with zero messages read. the all-duplicates filter path also resets it to 0 via `PolledMessages::empty()`. "until the server has answered the first poll" would be closer.

core/sdk/src/clients/consumer.rs
```diff
@@ -1229,6 +1724,10 @@ impl IggyConsumer {
     }
 }
 
+/// Stops the background tasks, nothing more.
+///
+/// Dropping cannot await, so it neither commits pending offsets nor leaves the consumer group.
```
hubcio: two edges off here: the diagnostics listener task isn't stopped by drop (it never checks the shutdown flag, only exits on the `Shutdown` event or channel close), and the store-offset task may still commit queued offsets after drop since the channel drains before closing.

core/sdk/src/clients/consumer.rs
```diff
@@ -100,6 +100,316 @@ pub enum AutoCommitAfter {
...
+/// There are two positions (offsets) tracked in two different places:
+/// - The **reading position** is held by the consumer, one per partition, and is the offset of the
+///   last message handed over
+///   ([`get_last_consumed_offset()`](Self::get_last_consumed_offset)). It dies with the process.
+/// - The **stored offset** lives on the server under the consumer name, or the group name for a
+///   group. This offset survives restarts. Writing it is called *storing* or *committing* an offset.
+///
+/// Committing matters because [`PollingStrategy::next()`] resumes from the stored offset. A
+/// consumer that never commits keeps starting over from the same place.
+/// Note, *committing* is a request of its own, not a side effect of reading and therefore controllable.
```
hubcio: under the default `PollingMessages` the commit is a flag on the poll request itself (line 1141), not a request of its own - the table row below says exactly that. maybe scope this sentence to the other modes.

core/sdk/src/clients/consumer.rs
```diff
@@ -297,16 +639,79 @@ impl IggyConsumer {
             .await
     }
 
-    /// Retrieves the last stored offset (on the server) for the specified partition ID.
-    /// To get the current partition ID use `partition_id()`
+    /// Returns the offset this consumer last stored on the server for the given partition, or
+    /// `None` if it has not stored one yet.
+    ///
+    /// The value is this consumer's own record of what it committed, kept in memory rather than
```
hubcio, line 850: under the default auto-commit-on-poll this lags the server by up to one batch - the local mirror gets the pre-batch consumed offset (1173-1175) while the server committed the batch end.

core/sdk/src/clients/consumer.rs
```diff
@@ -100,6 +100,316 @@ pub enum AutoCommitAfter {
...
+/// The switches have inverse setters as well, such as [`without_poll_interval()`],
+/// [`without_encryptor()`], [`do_not_auto_join_consumer_group()`] and
+/// [`do_not_create_consumer_group_if_not_exists()`].
+///
+/// # Encryption
+///
+/// When the [`IggyClient`] was created with an encryptor, payloads and user headers are decrypted
+/// before a message is yielded, which only works if the producer encrypted them with a matching
+/// key. This is guaranteed if you spawned both, the [`IggyProducer`] and the [`IggyConsumer`] from the same [`IggyClient`].
+/// A message that cannot be decrypted is yielded as an error and the rest of its batch is
```
hubcio: messages before the failing one are dropped too - decryption runs over the whole batch before anything is handed over, so the entire batch goes. and since no offset advanced, the next poll re-fetches the same batch and fails again, so it loops.

core/sdk/src/clients/consumer.rs
```diff
@@ -229,26 +539,48 @@ impl IggyConsumer {
...
+    /// Stores an offset on the server, marking every message up to and including it as consumed.
+    ///
+    /// This is the manual counterpart to [`AutoCommit`] and is meant for
+    /// [`AutoCommit::Disabled`].
+    ///
+    /// Pass `None` as `partition_id` to use the partition of the most recent batch polled.
+    ///
+    /// An offset that is not ahead of the last one this consumer stored for that partition is
+    /// skipped and `Ok(())` is returned without a request.
+    /// If you to re-read messages again, e.g. want to move an offset backwards configure the consumer
```
hubcio: this sentence got scrambled ("If you to re-read messages again... backwards configure"). similar small ones: 658 "the consumers", 671 "a background tasks", 742 "init_retires", 805 "If a the ... an time interval", 915 "Wait the task until", 1482 "A previous used", 1666 "if it the", 371-372 missing word before "Hence", 385 "a `IggyConsumer`", 360 "both, the".

core/sdk/src/clients/consumer.rs
```diff
@@ -332,7 +737,11 @@ impl IggyConsumer {
             let mut stream_exists = client.get_stream(&stream_id).await?.is_some();
             let mut topic_exists = client.get_topic(&stream_id, &topic_id).await?.is_some();
 
+            // Absent streams or topics are not necessarily permanent failures.
+            // It may happen that get_stream/ get_topic races the initial setup of the stream/ topic.
+            // Retry for init_retires times, while waiting interval between retries.
             loop {
+                // immediate happy path
```
hubcio: "immediate happy path" and a few others (1539, 1707) just restate the next line - they can go.

core/sdk/src/clients/consumer.rs
```diff
@@ -1207,9 +1702,9 @@ impl IggyConsumer {
             );
 
             let client = self.client.read().await;
-            // Cleared either way: this consumer is torn down regardless of
-            // whether the broker confirmed the leave.
+            // Update consumer state to not being part of a consumer group.
```
hubcio: the old comment explained why this is cleared before the leave result is known (the consumer is torn down either way, even if the broker never confirms) - the replacement lost that.

hubcio (changes requested):
most behavioral claims here actually check out against the code, which is rare for a docs PR.

the three crash-safety spots (the auto-commit advice, the guarantees section, the manual-commit example) state the opposite of what the code does, and those are exactly the lines people will build on. the rest is wording cleanup.

really good work @haubur

core/sdk/src/clients/consumer.rs
haubur: Is this something we want to add here, this seems rather specific, as it depends on one specific scenario/configuration.

core/sdk/src/clients/consumer.rs
haubur: Thats already what the docs are saying (batch is discarded and error yielded). The user can handle that, so it is not necessarily a loop.

core/sdk/src/clients/consumer.rs
haubur: The task is anyways aborted before the consumer leaves the group. Waking it is faster than waiting for the timeout later. Also it suggested that this only holds for group members, which it does not.

core/sdk/src/clients/consumer.rs
haubur: A previous comment of yours said it can go anyways.

core/sdk/src/clients/consumer.rs
haubur: Very nice! An important detail that would have otherwise been missed. Thanks!

core/sdk/src/clients/consumer.rs
haubur: True, thanks!

core/sdk/src/clients/consumer.rs
haubur: Indeed, missed that it is hard coded for AutoCommit.

core/sdk/src/clients/consumer.rs
haubur: Yes, this sentence needs to go. The final flush is already documented correctly.

core/sdk/src/clients/consumer.rs
hubcio: fair, order hold today but nothing enforces it. leave it as is

core/sdk/src/clients/consumer.rs
hubcio: mayb something like:
"under auto-commit-on-poll (the default) this can trail the server by up to one batch."

core/sdk/src/clients/consumer.rs
hubcio: "the rest of its batch" implies the earlier messages got delivered. they didn't, decryption runs over the whole batch before anything is handed over. just change "rest" to "whole".

core/sdk/src/clients/consumer.rs
hubcio: i think were talkin about different comment, 1707 ("Let the server know...") was the one that can go. here i just want the old why back: the flag is cleared even if the leave fails, since the consumer is torn down either way.

core/sdk/src/clients/consumer.rs
haubur: True, that makes sense as it is precise.

core/sdk/src/clients/consumer.rs
hubcio: those are suggestions of my LLM:

```
576
/// To re-read messages, e.g. to move an offset backwards, configure the consumer

658
/// - the consumer's `stream_id` and `topic_id` exist on the server.

671
/// Calling init spawns a background task that listens for lifecycle changes ([`DiagnosticEvent`]s) of the

742
// Retry for init_retries times, waiting interval between retries.

805
// If the configuration defines a time interval at which the offset should be stored,

915
// Wait until either the interval has passed or

1482
// A previously used (and therefore invalid) future was dropped, thus create a fresh one.

1666
// Store the latest offset for each partition, if the consumed offset

371-372
/// ([`StreamExt::next`]) and [`shutdown()`](Self::shutdown) take exclusive access (`&mut self`).
/// Hence, one task owns and drives a given consumer end to end.

385
/// Note that just dropping an `IggyConsumer` loses everything that is currently in-flight.
("Note, that" also wrong, drop comma.)

360
/// key. This is guaranteed if you spawned both the [`IggyProducer`] and the [`IggyConsumer`] from the same [`IggyClient`].
```

hubcio (dismissed):

core/sdk/src/clients/consumer.rs
```diff
@@ -100,6 +100,322 @@ pub enum AutoCommitAfter {
...
+///
+/// ## Guarantees
+///
+/// - **Each message is handed over once per consumer.** Messages whose offset is not greater than
+///   the reading position of their partition are dropped before they reach the stream. Re-reading
+///   a partition, or letting a message come back because your handler failed, needs
+///   [`allow_replay()`], which turns that filter off.
+/// - **Delivering at-least-once.** If you cannot tolerate missing any messages, use [`AutoCommit::Disabled`]
+///   and store the offset using [`Self::store_offset()`] after handling a message.
+///   Alternatively, you can use [`IggyConsumerMessageExt::consumer_messages()`] with [`AutoCommitAfter`].
```
ethanlin01x:
```suggestion
///   Alternatively, you can use [`IggyConsumerMessageExt::consume_messages`] with [`AutoCommitAfter`].
```

ethanlin01x: This path does not give at-least-once delivery. In `consumer_ext/consumer_message_ext.rs:91` a handler error is only logged, and `send_store_offset()` runs right after it either way, so a failed message still gets committed. I would drop this sentence and keep only the `AutoCommit::Disabled` advice above.

core/sdk/src/clients/consumer.rs
```diff
@@ -332,6 +745,9 @@ impl IggyConsumer {
             let mut stream_exists = client.get_stream(&stream_id).await?.is_some();
             let mut topic_exists = client.get_topic(&stream_id, &topic_id).await?.is_some();
 
+            // Absent streams or topics are not necessarily permanent failures.
+            // It may happen that get_stream/ get_topic races the initial setup of the stream/ topic.
+            // Retry for init_retires times, while waiting interval between retries.
```
ethanlin01x: init_retires ->  init_retries

spetz (approved):

hubcio (approved):

PR #3917, opened 2026-08-18
feat(node): support connection string syntax for nodejs client creation

T1B0:

this PR add node sdk support for connection string syntax on client creation 
accept common url like syntax `new Client("iggy://iggy:iggy@127.0.0.1:8090");`

note `node:url` module was not used since it would have introduced small differencez with rust and other sdk parsing, so connection string is manually parsed to match others sdk behaviors.

### DIFF
--- a/foreign/node/README.md
+++ b/foreign/node/README.md
@@ -67,6 +67,8 @@ new session.
 
 The client pings every `heartbeatInterval` milliseconds, 5000 by default, which
 keeps an idle session alive when the server's `[heartbeat]` eviction is enabled.
+`heartbeatInterval` also accepts a duration expression such as `"10s"` or
+`"1h 30m"`, like the Rust SDK.
 The server evicts a connection silent for 36 s, which is 1.2 x its 30 s
 heartbeat interval. Raising the client interval past that window, or setting it
 to 0 to disable client heartbeats, exposes an idle consumer-group member to
@@ -104,6 +106,43 @@ const client = new Client({
 
 const stats = await client.system.getStats();
 ```
+
+### Connection strings
+
+Every client constructor (except `SimpleClient` see note) also accepts a
+connection string instead of a config object:
+
+```ts
+import { Client } from "apache-iggy";
+
+const client = new Client("iggy://iggy:iggy@127.0.0.1:8090");
+const stats = await client.system.getStats();
+```
+
+Supported schemes are `iggy://` (TCP, default) and `iggy+tcp://`. Credentials
+are `username:password` or a single personal access token. Options mirror the
+other SDKs: `tls`, `tls_domain`, `tls_ca_file`, `reconnection_retries`,
+`reconnection_interval`, `heartbeat_interval` and `nodelay`. `reestablish_after`
+is accepted for format compatibility but has no Node equivalent.
+
+note: `SimpleClient` does not accept a connection string: it wraps an existing
+`RawClient` instance rather than building one from configuration. Pass the
+connection string to `Client`, `SingleClient` or `getRawClient` and hand the
+resulting raw client to `SimpleClient` if needed.
+
+### option limits
+
+| option | limit |
+| --- | --- |
+| `reconnection_retries` | integer up to `4294967295` (u32 max); larger values are rejected like Rust's u32 overflow, and `unlimited` maps to this ceiling. Defaults to unlimited |
+| `heartbeat_interval` | duration up to `2147483647ms` (Node's largest timer delay); `0` disables heartbeats |
+| `reconnection_interval` | positive duration (`ms`, `s`, `m`, `h`) up to `2147483647ms` (Node's largest timer delay); zero spellings are rejected. Defaults to `1s` |
+| port in the authority | decimal up to `65535` |
+
+Durations accept the same expressions as the Rust SDK, for example `500ms`,
+`10s`, `1h 30m`, `5d`, `2w`, `1y`; matching is case-insensitive and
+`0`, `unlimited`, `disabled` and `none` map to zero. Unit-less numbers such as
+`5` are rejected.
 
 ## use sources
 
--- a/foreign/node/src/client/client.config.ts
+++ b/foreign/node/src/client/client.config.ts
@@ -15,7 +15,9 @@
 // specific language governing permissions and limitations
 // under the License.
 
-import type { ClientConfig } from './client.type.js';
+import { MAX_U32 } from '../constant.js';
+import type { ClientConfig, ClientConfigOrString } from './client.type.js';
+import { parseConnectionString } from './client.connection-string.js';
 
 export const DEFAULT_MAX_RESPONSE_FRAME_SIZE = 64 * 1024 * 1024;
 
@@ -29,8 +31,11 @@ export const DEFAULT_HEARTBEAT_INTERVAL = 5 * 1000;
 export const MAX_HEARTBEAT_INTERVAL = 2_147_483_647;
 
 export const normalizeClientConfig = (
-  config: ClientConfig
+  config: ClientConfigOrString
 ): ClientConfig => {
+  if (typeof config === 'string')
+    config = parseConnectionString(config);
+
   const maxResponseFrameSize =
     config.maxResponseFrameSize ?? DEFAULT_MAX_RESPONSE_FRAME_SIZE;
   if (!Number.isSafeInteger(maxResponseFrameSize) ||
@@ -56,6 +61,30 @@ export const normalizeClientConfig = (
     throw new TypeError(
       `heartbeatInterval must be a safe integer of milliseconds between 0 and ${MAX_HEARTBEAT_INTERVAL} (0 disables heartbeats)`
     );
+
+  // Unlike the heartbeat, 0 is not a disable here: an immediate retry delay
+  // turns reconnection into a hot loop. The ceiling guards the same
+  // setInterval clamp, which would turn a long backoff into a 1 ms spin.
+  // A disabled reconnect never schedules a retry, so the interval check
+  // applies only when enabled; callers commonly pass a zero interval with
+  // enabled: false. maxRetries stays bounded either way, matching the
+  // connection-string path.
+  if (config.reconnect !== undefined) {
+    const { enabled, interval, maxRetries } = config.reconnect;
+    if (!Number.isSafeInteger(maxRetries) ||
+        maxRetries < 0 ||
+        maxRetries > MAX_U32)
+      throw new TypeError(
+        `reconnect.maxRetries must be a non-negative integer of at most ${MAX_U32}`
+      );
+    if (enabled &&
+        (!Number.isSafeInteger(interval) ||
+          interval < 1 ||
+          interval > MAX_HEARTBEAT_INTERVAL))
+      throw new TypeError(
+        `reconnect.interval must be a safe integer of milliseconds between 1 and ${MAX_HEARTBEAT_INTERVAL}`
+      );
+  }
 
   return {
     ...config,
--- a/foreign/node/src/client/client.connection-string.ts
+++ b/foreign/node/src/client/client.connection-string.ts
@@ -1,0 +1,226 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { MAX_U32 } from '../constant.js';
+import {
+  nanosecondsToMilliseconds,
+  parseIggyDurationNanoseconds
+} from '../duration.utils.js';
+import type { ClientConfig, ReconnectOption } from './client.type.js';
+
+const DEFAULT_PROTOCOL = 'iggy';
+const SCHEME_PREFIX = 'iggy+';
+const SUPPORTED_PROTOCOLS = ['tcp'] as const;
+
+/** Reconnection defaults carried by every connection string.
+ *
+ * Mirrors TcpConnectionStringOptions: retries default to unlimited and the
+ * interval to 1s. Deliberately not DefaultReconnectOption, whose 12/5s pair
+ * only applies to object configs.
+ */
+const CONNECTION_STRING_RECONNECT: ReconnectOption = {
+  enabled: true,
+  interval: 1000,
+  maxRetries: MAX_U32
+};
+
+/** Parses a duration with the exact grammar of the Rust SDK's
+ * `IggyDuration::from_str`, returning whole milliseconds.
+ */
+export const parseDuration = (value: string): number => {
+  try {
+    return nanosecondsToMilliseconds(parseIggyDurationNanoseconds(value));
+  } catch (error) {
+    throw new TypeError(`invalid duration "${value}"`, { cause: error });
+  }
+};
+
+/**
+ * Parses an Iggy connection string into a client configuration.
+ *
+ * Supports `iggy://` and `iggy+tcp://`; the Node SDK implements TCP/TLS only.
+ * Credentials are either `username:password` or a single personal access
+ * token before the `@`. TLS is enabled with `tls=true`.
+ */
+export const parseConnectionString = (connectionString: string): ClientConfig => {
+  if (typeof connectionString !== 'string' || connectionString.length === 0)
+    throw new TypeError('connection string must be a non-empty string');
+
+  const protocolParts = connectionString.split('://');
+  if (protocolParts.length !== 2)
+    throw new TypeError('invalid connection string');
+
+  const scheme = protocolParts[0];
+  const protocol = scheme === DEFAULT_PROTOCOL
+    ? 'tcp'
+    : scheme.startsWith(SCHEME_PREFIX)
+      ? scheme.slice(SCHEME_PREFIX.length)
+      : undefined;
+  if (protocol === undefined)
+    throw new TypeError('invalid connection string');
+  if (!SUPPORTED_PROTOCOLS.includes(protocol as (typeof SUPPORTED_PROTOCOLS)[number]))
+    throw new TypeError(
+      `unsupported transport "${protocol}", Node SDK supports tcp only`
+    );
+
+  const parts = protocolParts[1].split('@');
+  if (parts.length !== 2)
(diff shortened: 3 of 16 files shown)


foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,212 @@
...
+      serverAddress.startsWith(':'))
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
+
+  const port = serverAddress.slice(serverAddress.lastIndexOf(':') + 1);
+  if (port.length === 0 || !/^\d+$/.test(port) || Number(port) > 65535)
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
+
+  const host = serverAddress.slice(0, serverAddress.lastIndexOf(':'));
+  if (host.length === 0)
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
```
slbotbm: IPv6 connection strings need brackets in the authority, but `net.createConnection` must receive the bare address. This currently produces `host: '[::1]'`, which Node attempts to resolve as a hostname and fails with `ENOTFOUND`.

I would recommend stripping the surrounding brackets when parsing an IPv6 host, so `[::1]:8090` becomes `{ host: '::1', port: 8090 }`. Additionally, the current IPv6 test should exercise an actual connection or at least assert the unbracketed host.

slbotbm (changes requested):

foreign/node/src/client/client.connection-string.ts
T1B0: nice catch ! just pushed a fixed version, added an error & a test case :+1:

slbotbm (dismissed):

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+ * Credentials are either `username:password` or a single personal access
+ * token before the `@`. TLS is enabled with `tls=true`.
+ */
+export const parseConnectionString = (connectionString: string): ClientConfig => {
+  if (typeof connectionString !== 'string' || connectionString.length === 0)
+    throw new TypeError('connection string must be a non-empty string');
+
+  const protocolParts = connectionString.split('://');
+  if (protocolParts.length !== 2)
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
```
hubcio: every parse failure echoes the whole connection string, which holds the password or PAT in plaintext. same at 63, 72, 77, 82, 86, 92, 96, 100, 104, 148, and worded differently at 186, 200, 212. rust returns an opaque error here for that reason - drop the interpolation, or echo scheme + host:port only.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+  const password = credentials[1] ?? '';
+  if (!tokenCredentials && (username.length === 0 || password.length === 0))
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
+
+  const serverAndOptions = parts[1].split('?');
+  if (serverAndOptions.length > 2)
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
+
+  const serverAddress = serverAndOptions[0];
+  if (serverAddress.length === 0 ||
```
hubcio: `lastIndexOf(':')` guesses wrong on multi-colon authorities - `2001:db8::1` parses as `{host:"2001:db8:", port:1}`, and `host:8090:9090` gives port 9090 where rust reads 8090. one regex replaces the whole block: `/^(?:\[([^\]]+)\]|([^:\[\]]+)):(\d+)$/` with `host = m[1] ?? m[2]`, keeping the `> 65535` check.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+        parsed.tls = parseBoolean(name, value, connectionString);
+        break;
+      case 'nodelay':
+        parsed.noDelay = parseBoolean(name, value, connectionString);
+        break;
+      case 'tls_domain':
+        parsed.servername = value;
+        break;
+      case 'tls_ca_file':
+        parsed.ca = readFileSync(value);
```
hubcio: reads the file during parsing even when the transport ends up plain TCP, and a bad path throws a raw ENOENT `Error` instead of the `TypeError` the test asserts. rust stores the path and reads it at connect.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+      throw new TypeError(`invalid connection string: "${connectionString}"`);
+    const [name, value] = optionParts;
+    switch (name) {
+      case 'tls':
+        parsed.tls = parseBoolean(name, value, connectionString);
+        break;
+      case 'nodelay':
+        parsed.noDelay = parseBoolean(name, value, connectionString);
+        break;
+      case 'tls_domain':
```
hubcio, line 168: `tls=true` without `tls_domain` sends no SNI - node won't derive `servername` from `host`, rust falls back to it. fine against iggy's own server, breaks behind an SNI-routing terminator. default it to `host` when `net.isIP(host) === 0`.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+const DURATION_UNITS = {
+  ms: 1,
+  s: 1000,
+  m: 60 * 1000,
+  h: 60 * 60 * 1000
+} as const;
+
+/** Parses a duration such as "500ms" or "5s" into milliseconds. */
+export const parseDuration = (value: string): number => {
+  const match = /^(\d+(?:\.\d+)?)(ms|s|m|h)$/.exec(value);
```
hubcio: strict subset of `IggyDuration::from_str` - `0`, `unlimited`, `disabled`, `none`, `1d`, `5sec`, `1h 30m` all work in rust and throw here. `heartbeat_interval=0` is the documented disable and only `0ms` gets through. also `1.005s` gives 1004.9999999999999, which `normalizeClientConfig` then rejects - `Math.round` the product.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+      case 'tls_domain':
+        parsed.servername = value;
+        break;
+      case 'tls_ca_file':
+        parsed.ca = readFileSync(value);
+        break;
+      case 'reconnection_retries':
+        parsed.reconnect = {
+          enabled: true,
+          interval: parsed.reconnect?.interval ?? 5000,
```
hubcio: defaults don't match rust: `retries=3` leaves interval 5000 (rust 1s), `interval=1s` leaves maxRetries 12 (rust unlimited), and no query emits no `reconnect` key at all (rust applies unlimited/1s). see `core/common/src/types/configuration/tcp_config/tcp_connection_string_options.rs:67-68`.

watch the order - the 12-retry default is the only thing bounding a bad interval today, so switch to unlimited only after line 176 is bounded. and don't reuse `DefaultReconnectOption`, that's 12/5s for object configs.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+          maxRetries: value === 'unlimited'
+            ? Number.MAX_SAFE_INTEGER
+            : parseNumber(name, value, connectionString)
+        };
+        break;
+      case 'reconnection_interval':
+        parsed.reconnect = {
+          enabled: true,
+          maxRetries: parsed.reconnect?.maxRetries ?? 12,
+          interval: parseDuration(value)
```
hubcio: `reconnect.interval` is never validated. `1000h` overflows node's 32-bit timer and clamps to 1 ms, so the longest backoff gives the fastest loop; `0ms` does the same. `normalizeClientConfig` already guards `heartbeatInterval` for this - `reconnect` needs a floor and ceiling too.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+        };
+        break;
+      case 'reestablish_after':
```
hubcio, line 201: value is never checked, so `?reestablish_after=garbage` parses fine. rust validates it as a duration - `parseDuration(value)` and discard.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+  const rawHost = serverAddress.slice(0, serverAddress.lastIndexOf(':'));
+  if (rawHost.length === 0)
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
+  // net.connect resolves a literal IPv6 address only without its brackets.
+  const bracketed = /^\[(.+)\]$/.exec(rawHost);
+  if (rawHost.startsWith('[') && !bracketed)
+    throw new TypeError(`invalid connection string: "${connectionString}"`);
+  const host = bracketed ? bracketed[1] : rawHost;
+
+  const options: ParsedConnectionOptions = serverAndOptions[1]
```
hubcio: truthiness means a trailing `?` parses while `?&` throws. rust rejects both. `serverAndOptions.length === 2`.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,216 @@
...
+      case 'tls_ca_file':
+        parsed.ca = readFileSync(value);
+        break;
+      case 'reconnection_retries':
+        parsed.reconnect = {
+          enabled: true,
+          interval: parsed.reconnect?.interval ?? 5000,
+          maxRetries: value === 'unlimited'
+            ? Number.MAX_SAFE_INTEGER
+            : parseNumber(name, value, connectionString)
```
hubcio: no upper bound - `reconnection_retries=99999999999999` is accepted where rust errors on u32 overflow. second undocumented spelling of `unlimited`.

foreign/node/src/client/index.ts
```diff
@@ -17,6 +17,7 @@
 
 export { Client, SimpleClient, SingleClient } from './client.js'
 export * from './client.config.js';
+export * from './client.connection-string.js';
```
hubcio: `export *` makes `parseDuration` public api. it only needs the module export for its own test - `export { parseConnectionString } from './client.connection-string.js';`

foreign/node/README.md
```diff
@@ -105,6 +105,24 @@ const client = new Client({
 const stats = await client.system.getStats();
 ```
 
+### Connection strings
+
+Every client constructor also accepts a connection string instead of a config
```
hubcio: `SimpleClient` takes a `RawClient`, so not every constructor. worth also listing the reconnect and heartbeat defaults a connection string implies, and that ipv6 literals need brackets.

foreign/node/src/e2e/tcp.connection-string.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+import { after, describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { Client } from '../client/client.js';
+import { getIggyAddress } from '../tcp.sm.utils.js';
+
+const dummyOpt = 'nodelay=true' +
```
hubcio, line 25: `reconnection_retries` is set twice - last wins, so `=1` is dead and this actually runs `unlimited`. only `ping()` is asserted so no option is checked, and `destroy()` on line 39 isn't awaited.

hubcio (changes requested):
a few notes from comparing against the rust parser. one i'd want fixed before merge - the credential echo in errors.

these landed outside the diff:

`foreign/node/src/stream/consumer-stream.ts:81` and `:113` - `singleConsumerStream` and `groupConsumerStream` still take `ClientConfig`, so passing a connection string fails to compile (TS2345). they call the widened `getClient`, and since `getClient` isn't exported they're the only place users can reach the widening.

`foreign/node/src/client/client.connection.ts:86-90` - reconnect timer is never unref'd and the handle isn't kept, so `unlimited` means the process can't exit and `destroy()` can't cancel a wait already in flight. the heartbeat timer next door unrefs itself with a comment saying why. pre-existing, but this PR makes `unlimited` a one-word option and ships it in the new e2e.

`foreign/node/src/client/client.config.ts:61` - error says `heartbeatInterval` in raw ms at someone who typed `heartbeat_interval=1000h`.

nothing tests the tls branch - the pre-merge action runs `tls.system.e2e.ts`, which builds from an object config, so the parser never gets hit.

foreign/node/src/client/client.connection-string.ts
T1B0: fixed :+1:

foreign/node/src/client/client.connection-string.ts
T1B0: fixed :+1:

foreign/node/README.md
T1B0: fixed README.md

foreign/node/src/e2e/tcp.connection-string.e2e.ts
T1B0: fixed :+1:

foreign/node/src/client/client.connection-string.ts
T1B0: matched rust default :+1:

foreign/node/src/client/client.connection.ts
```diff
@@ -84,8 +103,10 @@ const DefaultReconnectOption: ReconnectOption = {
  * @returns Promise resolving after the delay
  */
 function waitForReconnect(timer = 1000): Promise<void> {
-  return new Promise((resolve) => {
-    setTimeout(resolve, timer);
+  return new Promise<void>((resolve) => {
+    const timeout = setTimeout(resolve, timer);
+    (timeout as NodeJS.Timeout).unref();
+    resolve();
```
hubcio: `resolve()` runs right away inside the executor, so this never waits - the loop redials back to back and the unref'd timer later fires into an already settled promise. counted at a local tcp listener that drops every accept: 111258 dials in 2 s with `interval: 1000`, and 2 dials with this line removed. connection strings now default to unlimited retries, so every string client storms for the whole outage.

simplest fix is `await setTimeout(interval, undefined, { ref: false })` from `node:timers/promises`. keep it unref'd - a queued command fails fast on `disconnected`, so the loop is background work like the heartbeat. also worth an elapsed-time assert on the maxRetries test in `client.connection.test.ts` (interval 10 x 3 retries has to take at least 30 ms), nothing checks timing today which is why the suite still passes.

foreign/node/src/client/client.connection-string.ts
```diff
@@ -0,0 +1,238 @@
...
+    .replaceAll('nsec', 'ns')
+    .replaceAll('usec', 'us')
+    .replaceAll('msec', 'ms');
+  if (expression === '0' || expression === 'unlimited' ||
+      expression === 'disabled' || expression === 'none')
+    return 0;
+  // The library alone would accept a bare number as milliseconds.
+  if (/^-?\d*\.?\d+$/.test(expression))
+    throw new TypeError(`invalid duration "${value}"`);
+  const parsed = parseDurationLib(expression);
```
hubcio: `parse-duration` scans for `number unit` tokens and ignores whatever is left, so it takes inputs humantime rejects: `10s5` -> 10005, `1h30` -> 90 min, `1e3` -> 1000, `5s junk` -> 5000, `1,000ms` -> 1000, `1.5.5s` -> 502, and `1h -30m` -> 90 min with the sign dropped. the README promises the rust grammar, so a typo parses here and errors there.

a strict tokenizer is about 20 lines and drops the dep: sticky `/(\d+(?:\.\d+)?)\s*([a-z]+)\s*/y` over humantime's unit table (`ns nsec us usec ms msec s sec second(s) m min minute(s) h hr hour(s) d day(s) w week(s) y year(s)`), throw unless the whole string is consumed, keep the four zero sentinels. if the dep stays, `^2.1.8` in `package.json` breaks the exact pin every other dependency uses.

foreign/node/src/client/client.config.ts
```diff
@@ -48,15 +55,32 @@ export const normalizeClientConfig = (
...
       heartbeatInterval < 0 ||
       heartbeatInterval > MAX_HEARTBEAT_INTERVAL)
     throw new TypeError(
       `heartbeatInterval must be a safe integer of milliseconds between 0 and ${MAX_HEARTBEAT_INTERVAL} (0 disables heartbeats)`
     );
 
+  // Unlike the heartbeat, 0 is not a disable here: an immediate retry delay
+  // turns reconnection into a hot loop. The ceiling guards the same
+  // setInterval clamp, which would turn a long backoff into a 1 ms spin.
+  if (config.reconnect !== undefined) {
```
hubcio, line 72: this runs even when `enabled` is false, so `{ enabled: false, interval: 0, maxRetries: 0 }` - the shape the tests in this repo use and master accepted - now throws from `new Client()`. skip the interval check when reconnect is disabled. while here, `maxRetries` still passes `-1`, `NaN` and `1.5` untouched on the object path even though the string path bounds it.

foreign/node/src/client/client.type.ts
```diff
@@ -160,12 +168,14 @@ export type ClientConfig = {
...
+   * Interval for sending heartbeat pings, either in milliseconds or as a
+   * duration expression such as "5s" or "1h 30m" (same grammar as the Rust
+   * SDK's IggyDuration). Normalizes to an integer between 0 and Node's timer
+   * ceiling. Defaults to 5000. Set to 0 (or "unlimited"/"disabled"/"none")
+   * to disable client heartbeats; any other unusable value is rejected
+   * rather than silently disabling them.
    */
-  heartbeatInterval?: number,
+  heartbeatInterval?: number | string,
   /** Maximum accepted response frame size in bytes */
```
hubcio, line 192: `number | string` here is a public API change beyond the connection-string scope, and it is what forces the cast in `client.socket.ts` `_init()`. intended? if not, keep the string form on the parser side only.

foreign/node/README.md
```diff
@@ -105,6 +107,43 @@ const client = new Client({
...
```
hubcio: "every client constructor" contradicts the `SimpleClient` paragraph below - "every constructor except `SimpleClient`". the `reconnection_interval` row also lists `ms, s, m, h` only, while the paragraph under the table says `d`, `w` and `y` work too.

foreign/node/src/stream/consumer-stream.ts
```diff
@@ -90,7 +91,7 @@ export const singleConsumerStream = (config: ClientConfig) => async (
     endOnLastOffset = false
   }: SingleConsumerStreamRequest
 ): Promise<Readable> => {
-  const c = await getClient(config);
+  const c = await getClient(normalizeClientConfig(config));
```
hubcio: also at line 125.

`getClient` already normalizes inside `CommandResponseStream`, so this extra `normalizeClientConfig` and its import can go.

hubcio (changes requested):
round 2 - everything from the first pass landed, thanks. one new blocker though: the timer change in `waitForReconnect` settles the promise right away, so the reconnect loop no longer waits at all (numbers inline). two smaller things worth landing with it: the `parse-duration` grammar is looser than the rust one the README promises, and the new `reconnect.interval` check rejects configs that disable reconnect. the rest is nits.

hubcio (approved):

spetz (approved):

PR #3922, opened 2026-08-19
test(java): add Pinot connector E2E coverage

### LINKED ISSUE #2598: Add e2e tests for Apache Pinot connector
Currently the Apache Pinot connector has no proper Java integration tests. There is an existing bash script (`external-processors/iggy-connector-pinot/integration-test.sh`) that performs manual integration testing, but this should be replaced with proper Java tests that can run in CI.

  ## What needs testing

  - End-to-end ingestion flow: send messages to Iggy → verify data lands in Pinot realtime table
  - JSON payload parsing and schema mapping
  - Bulk message ingestion

  ## Implementation notes

  - Use Testcontainers for Pinot (controller, broker, server) and Iggy
  - Use JUnit 6 as the test engine
  - Use AssertJ for assertions
  - Consider using `@Testcontainers` annotation for lifecycle management
  - Schema and table configs can be reused from `deployment/` directory
  - Remove `integration-test.sh` once proper tests are in place

  ## Files to reference

  - `foreign/java/external-processors/iggy-connector-pinot/integration-test.sh` - current manual test flow (to be replaced)
  - `foreign/java/external-processors/iggy-connector-pinot/deployment/schema.json` - Pinot schema
  - `foreign/java/external-processors/iggy-connector-pinot/deployment/table.json` - Pinot table config
  - `foreign/java/external-processors/iggy-connector-pinot/src/main/java/` - connector implementation

  ## Acceptance criteria

  - [ ] At least 2-3 tests covering basic ingestion and JSON parsing
  - [ ] Tests use Testcontainers (no manual Docker setup required)
  - [ ] Tests run in CI via Gradle
  - [ ] Test failures are informative with clear error messages
  - [ ] Remove `integration-test.sh` after Java tests provide equivalent coverage

goutamadwant:

Closes #2598

## Rationale

The Pinot connector needs repeatable, Gradle-integrated end-to-end coverage for its documented ingestion workflow.

## What changed?

Before, Pinot connector E2E coverage lived in a manual shell script with fixed ports, fixed sleeps, and a mutable Pinot image. The workflow was not part of Gradle and could report table creation success even when Pinot rejected the connector.

The shell workflow is now a JUnit and Testcontainers suite that verifies JSON field mapping and 10-message batch ingestion. Local runs use an always-pulled VSR-capable Iggy image, while CI uses the branch-built external server through Testcontainers host-port forwarding and an in-memory Pinot table configuration.

Gradle tracks the deployment configuration and external-server settings as test inputs, and Java CI publishes the Pinot XML results in its test summary. The shaded plugin keeps its relocated Netty dependency, while bounded polling and service-log diagnostics make failures actionable.

## Local Execution

- Passed: `./gradlew :iggy-connector-pinot:check --no-daemon --no-build-cache --rerun-tasks`
- Passed: `./gradlew check -x test --no-daemon --no-build-cache --rerun-tasks`
- Passed against a branch-built server: `USE_EXTERNAL_SERVER=true ./gradlew test --no-daemon --no-build-cache --rerun-tasks`
- Pre-commit hooks ran on the changed files and passed.

## AI Usage

Codex was used to understand the existing repo and codebase.


foreign/java/external-processors/iggy-connector-pinot/src/test/java/org/apache/iggy/connector/pinot/IggyPinotIntegrationTest.java
```diff
@@ -0,0 +1,418 @@
...
+import java.util.function.Predicate;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+@Testcontainers
+class IggyPinotIntegrationTest {
+
+    // The Java SDK speaks VSR, so use the same VSR-capable image as its integration tests.
+    private static final DockerImageName IGGY_IMAGE = DockerImageName.parse("apache/iggy:edge");
```
mmodzelewski, line 67: Could we add a `USE_EXTERNAL_SERVER` switch here, mirroring `BaseIntegrationTest`? In CI we want this suite to run against the iggy-server built from the current branch (the same way the SDK tests do), so the E2E actually validates the code under review. The `apache/iggy:edge` container should stay only as the developer-convenience path for local runs.

One thing to solve for the external mode: the Pinot containers resolve the server via the `iggy` network alias and `deployment/table.json` hardcodes `stream.iggy.host: "iggy"` / port `8090`. With an external server on the host, the containers need a route to it (e.g. `host.docker.internal` / `host-gateway` extra host mapped to the `iggy` alias, or templating the host/port into the table config before POSTing it).

foreign/java/external-processors/iggy-connector-pinot/src/test/java/org/apache/iggy/connector/pinot/IggyPinotIntegrationTest.java
```diff
@@ -0,0 +1,418 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.function.Predicate;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+@Testcontainers
```
mmodzelewski: `@Testcontainers` is decorative here: there are no `@Container` fields, so the extension manages nothing - lifecycle is entirely manual in `@BeforeAll`/`@AfterAll`. Either remove the annotation or add `@Container` annotations to let the extension manage the containers. Note that the `@Container` route will be harder to keep once the external-server env switch is introduced, since some containers would then start conditionally, which doesn't fit the annotation-driven lifecycle well.

foreign/java/external-processors/iggy-connector-pinot/src/test/java/org/apache/iggy/connector/pinot/IggyPinotIntegrationTest.java
```diff
@@ -0,0 +1,418 @@
...
+            try {
+                HttpResponse<String> response = post(
+                        pinotBroker,
+                        PINOT_BROKER_PORT,
+                        "/query/sql",
+                        OBJECT_MAPPER.createObjectNode().put("sql", sql).toString());
+                lastResponse = "HTTP " + response.statusCode() + ": " + response.body();
+                if (response.statusCode() >= 200 && response.statusCode() < 300) {
+                    JsonNode json = OBJECT_MAPPER.readTree(response.body());
+                    if (hasExceptionCode(json, 150)) {
```
mmodzelewski: Nit: `150` is Pinot's SQL parsing error code. `pinot-spi` is already on the test classpath, so `QueryErrorCode.SQL_PARSING.getId()` (or a named constant) would document the intent.

foreign/java/external-processors/iggy-connector-pinot/build.gradle.kts
```diff
@@ -43,20 +44,37 @@ dependencies {
...
+tasks.register<Sync>("assemblePlugin") {
+    from(tasks.named("shadowJar"))
     into(layout.buildDirectory.dir("plugin"))
 }
 
 tasks.named("jar") {
     finalizedBy("assemblePlugin")
 }
 
+tasks.named<Test>("test") {
```
mmodzelewski, line 67: The `test` task doesn't declare `deployment/` as an input, so editing `schema.json` / `table.json` and rerunning `test` can report UP-TO-DATE with stale results. Suggest adding:

```kotlin
inputs.dir(layout.projectDirectory.dir("deployment"))
```

foreign/java/external-processors/iggy-connector-pinot/build.gradle.kts
```diff
@@ -43,20 +44,37 @@ dependencies {
...
     testRuntimeOnly(libs.slf4j.simple)
 }
 
-// Assemble connector plugin with all dependencies for Docker deployment
-tasks.register<Copy>("assemblePlugin") {
-    from(tasks.named("jar"))
-    from(configurations.runtimeClasspath)
+tasks.shadowJar {
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+    filesMatching("META-INF/services/**") {
```
mmodzelewski: Nit: the `filesMatching("META-INF/services/**") { duplicatesStrategy = INCLUDE }` block is dead config - `mergeServiceFiles()` installs a transformer that takes service descriptors out of the normal copy path entirely, so the top-level `EXCLUDE` never sees them.

mmodzelewski (changes requested):

mmodzelewski:
@goutamadwant Thanks for the contribution! This is a solid replacement for the shell workflow.

`.github/actions/java-gradle/pre-merge/action.yml` copies test results and feeds the Test Summary only from `java-sdk` and the Flink connector. The new Pinot E2E results won't appear in the summary (a failure still fails the build, but without the report). Worth adding `external-processors/iggy-connector-pinot/build/test-results` to the copy step and the summary paths.

goutamadwant:
@mmodzelewski addressed comments in latest push. 
Let me know if this looks good or any suggestions/comments. thanks!

mmodzelewski (approved):
thanks @goutamadwant, looks good, and sorry for the delay :)

mmodzelewski (changes requested):
Actually, one more thing, there's now a new validation that requires setting `IGGY_NODE_ADVERTISED_ADDRESS` when the bind port is `0.0.0.0`. That variable would be required in the test container setup. Please check it locally and update accordingly.

mmodzelewski (approved):
I'll address the container config separately as the SDK need some adjustments anyway

slbotbm (approved):

hubcio (approved):

PR #3923, opened 2026-08-19
fix(server)!: reject a wildcard bind with no advertised address

### LINKED ISSUE #3890: Server publishes its wildcard bind address as a client-facing address
## Summary

A server whose TCP listener binds a wildcard reports that wildcard as its own client-facing address in `GetClusterMetadata`. The Go SDK collects roster addresses into its reconnect candidate ring, so `0.0.0.0:8090` becomes a dial target.

## Reproduction

The shipped BDD compose file already reproduces it. In `bdd/docker-compose.server.yml`.

https://github.com/apache/iggy/blob/7c0fd68421aa610897fdbc7331f1cb22848fb386/bdd/docker-compose.server.yml#L32

https://github.com/apache/iggy/blob/7c0fd68421aa610897fdbc7331f1cb22848fb386/bdd/docker-compose.server.yml#L67

Confirmed directly by varying only the bind address on one binary:

| Bind | Client dialed | Reported IP |
| ---- | ------------- | ----------- |
| `127.0.0.1:8090` | `127.0.0.1:8090` | `127.0.0.1` (usable) |
| `0.0.0.0:8190` | `127.0.0.1:8190` | `0.0.0.0` (not usable) |

With the cluster disabled, `ClusterRoster::cluster_metadata` reports `self_ip`, which is assigned the bind address verbatim:

```rust
// core/server/src/bootstrap.rs:1719
self_ip: topology.client_listen_addr.ip().to_string(),
```

## What the Go SDK does with it

1. Every sign-in runs leader settlement: `settleOnLeader` (`foreign/go/client/tcp/tcp_session_management.go:69`).
2. `clusterAddresses` (`foreign/go/internal/util/leader_aware.go:126`) collects an address for every healthy node, unconditionally.
3. Those land in `knownServerAddresses` (`foreign/go/client/tcp/tcp_session_management.go:212`).
4. `connectionCandidates` (`foreign/go/client/tcp/tcp_core.go:998`) appends them to the configured addresses, and `Connect` rotates through the result on every retry (`tcp_core.go:924`).

Result for the reproduction above: `["iggy-server:8090", "0.0.0.0:8090"]`.

## Root cause

The advertised-address mechanism already exists and is complete for roster nodes: `ClusterNodeConfig.advertised_address` plus the client-CIDR-scoped `advertised_addresses` selectors (`core/configs/src/server_config/cluster.rs:369-377`), with strict boot validation (`cluster.rs:978-1042`) and cross-node endpoint conflict detection (`cluster.rs:1357`).

`self_metadata()` (`cluster_meta.rs:134`) does not participate in any of it. It is a parallel path that substitutes the bind address, manufacturing an addressthe server does not actually know.

## Proposed fix

Publish only dialable address

A concrete bind IP is an interface the operator named, so it is reported as-is and that path is unchanged. An unspecified one carries no client-facing information, so the node publishes no address at all.

Nothing is invented to fill the gap. The self node is the one node whose address a client never needs told: it is already connected to it, and with a single-node roster there is nowhere to fail over to. Empty is already a value the wire carries, via `ClusterRoster::disabled()` (`cluster_meta.rs:80`), so no protocol change is involved. 

### Affected area / component

_No response_

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

### Contribution

### Good first issue

chengxilo:

  Closes #3890

This PR block #3650, see #3890 for detail.
  ## Rationale

  A wildcard bind says which interfaces a node accepts on, not where a client reaches it, so publishing it as an address hands clients a target they cannot dial.

  ## What changed?

`node.advertised_address` supplies it, and the server now refuses to start when server is configured with wildcard bind while leaves it unset.

 ### **Breaking:** 
deployments binding `0.0.0.0` without a roster must declare an address. The Helm chart can derives it from the Service DNS name and the shipped  compose files name their service; anything else needs `IGGY_NODE_ADVERTISED_ADDRESS`. A `cluster.nodes` ip must now be a literal IP,  and no declared address may be the unspecified one.

### Some detail regarding new behavior:
<details>
<summary>When would it boot?</summary>

## Standalone (`cluster.enabled = false`)

  | `tcp.address` | `node.advertised_address` | Boot | Metadata publishes |
  | --- | --- | --- | --- |
  | `127.0.0.1:8090` | unset | ✅ | `127.0.0.1` (derived from the bind) |
  | `127.0.0.1:8090` | `broker.example.com` | ✅ | `broker.example.com` (declared wins) |
  | `0.0.0.0:8090` | unset | ❌ **rejected** | — |
  | `0.0.0.0:8090` | `broker.example.com` | ✅ | `broker.example.com` |
  | any | `0.0.0.0` / `::` | ❌ **rejected** | — |
  | any | `broker:8090` (carries a port) | ❌ **rejected** | — |

  ## Cluster (`cluster.enabled = true`)

  `tcp.address` only picks the bind interface here — **ports come from the roster** — so a wildcard is perfectly
  normal.

  | Roster field | Value | Boot |
  | --- | --- | --- |
  | `nodes[].ip` | `172.28.0.101` | ✅ |
  | `nodes[].ip` | `0.0.0.0` / `::` | ❌ **rejected** |
  | `nodes[].ip` | `iggy-server` (any hostname) | ❌ **rejected**, points at `advertised_address` |
  | `advertised_address` | unset | ✅ → publishes `ip` |
  | `advertised_address` | `broker.example.com` | ✅ → publishes it |
  | `advertised_address` | `0.0.0.0` / `::` | ❌ **rejected** |
  | selector `address` | `0.0.0.0` / `::` | ❌ **rejected**, error names the CIDR |
  | `node.advertised_address` | set | ✅ but **ignored**, warns at startup |

  Which address a client is told (`advertised_for`):

  ```
  selector matching the client's source IP (longest prefix) → advertised_address → ip
  ```

  ## Both modes

  | `tcp.address` | Boot |
  | --- | --- |
  | `:8090` (empty host) | ❌ **rejected** — Rust's `SocketAddr` has no such spelling |
  | `localhost:8090` (hostname) | ❌ **rejected** — a literal IP is required |
  | anything else that does not parse | ❌ **rejected**, one message naming the fix |

  ## Edge cases

  | Situation | Behavior |
  | --- | --- |
  | `cluster.enabled = false` with `[[cluster.nodes]]` left behind | Roster is **neither resolved nor validated** —
  boots even if an ip is a hostname |
  | A selector whose CIDR does not parse | ❌ rejected (previously dropped in silence) |

  That last row is a side fix: a malformed selector used to be swallowed, surfacing only as "clients on one network
  get the catch-all address" with nothing to debug.
</details>

  ## Local Execution

  - Passed
  - Pre-commit hooks ran

  ## AI Usage

  1. Claude Code (Opus)
  2. Diagnosis and implementation
  3. Reviewed line by line with my best effort. BUT I am not very familiar with the server side code, so I am not sure if these changes introduce any side-effect that I didn't notice. According to the changed part, I think it looks ok.
  4. Yes


core/configs/src/server_config/server.rs
```diff
@@ -111,6 +112,7 @@ pub struct ServerConfig {
     pub consumer_group: ConsumerGroupConfig,
     pub data_maintenance: DataMaintenanceConfig,
     #[serde(default)]
+    pub node: NodeConfig,
```
mmodzelewski, line 115: This insertion lands between the `#[serde(default)]` attribute and `personal_access_token`, so the attribute now applies to `node` and `personal_access_token` has lost its default. A bare deserialize of a config without a `[personal_access_token]` section now fails with "missing field". It's masked in the server binary because the embedded default config.toml is always merged first (`file_provider.rs:162`), but it's a silent contract change on a public struct. Suggest keeping `#[serde(default)]` on both fields.

core/server/src/bootstrap.rs
```diff
@@ -3030,6 +3040,12 @@ async fn start_tcp_runtime(
     // reactor, so it binds independently. Shard-0 gating comes from the sole
     // caller of this function.
     if let Some(http_addr) = topology.http_listen_addr {
+        // One host for all four transports, resolved from the client-facing
+        // TCP bind so both listeners publish the same node address.
+        let self_advertised = self_advertised_address(
```
mmodzelewski: This changes the source of the HTTP metadata self address: before this PR, http.rs published the HTTP listener's own bound address (`self_ip: bound_addr.ip().to_string()`); now it derives from the TCP bind. With the shipped default `tcp.address = 127.0.0.1:8090` and HTTP bound to a concrete external interface, the published address silently flips from the external IP to `127.0.0.1` - no boot refusal, no warning. Consider warning on (or validating) the interface mismatch between the TCP and HTTP binds.

core/configs/src/server_config/cluster.rs
```diff
@@ -629,6 +647,10 @@ pub enum AdvertisedAddress {
 }
 
 impl AdvertisedAddress {
+    pub fn is_unspecified(&self) -> bool {
```
mmodzelewski: `Ipv6Addr::is_unspecified` is true only for `::`, so the v4-mapped `::ffff:0.0.0.0` slips through this check and through every gate built on it (node.advertised_address, roster ip, selector addresses, and the new tcp.address validation, where `[::ffff:0.0.0.0]:8090` binds the v4 wildcard on a dual-stack Linux host - exactly the bug class this PR is fixing). `ip.to_canonical().is_unspecified()` closes it; `canonical_ip_net` above already canonicalizes mapped forms for selector CIDRs. A doc comment on this fn would also match its documented siblings.

core/configs/src/server_config/cluster.rs
```diff
@@ -473,49 +474,80 @@ pub struct AdvertisedAddressSelector {
...
-                Some((canonical_ip_net(network.trunc()), address))
-            })
-            .collect();
-        let catch_all = match config.advertised_address.as_deref() {
-            Some(advertised_address) => advertised_address.parse().ok(),
-            None => config.ip.parse().ok(),
-        };
-        let replica_ip = config.ip.parse().ok();
-        Self {
+impl TryFrom<ClusterNodeConfig> for ResolvedClusterNode {
```
mmodzelewski, line 493: `TryFrom` checks only that the sources parse; it doesn't reject unspecified addresses the way the validator does for roster ip, catch-all, and selectors. The "always dialable" invariant therefore holds only because validation runs before conversion - a direct `TryFrom` caller can still build a node advertising `0.0.0.0`. Mirroring the unspecified rejection here (on the canonical form) would make the type self-defending.

core/server/src/cluster_meta.rs
```diff
@@ -43,6 +44,34 @@ const SELF_NODE_NAME: &str = "iggy-node";
 /// single-node label.
 const SINGLE_NODE_CLUSTER_NAME: &str = "single-node";
 
+pub fn self_advertised_address(declared: Option<&str>, bind: IpAddr) -> String {
```
mmodzelewski, line 58: The declared address is published verbatim while the roster path normalizes through `AdvertisedAddress`'s Display (lowercased hostnames, canonical IPv6 - pinned by the tests in this file). A declared `Broker.Example.COM` or `[2001:db8::1]` reaches the wire as typed, so the same address can publish in two spellings depending on the path. `NodeConfig::validate` already guarantees the parse, so parse-then-Display here is infallible.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -89,6 +89,14 @@ spec:
                   name: {{ include "iggy.fullname" . }}-root-credentials
                   key: password
   {{- end }}{{- end}}
+  {{- $declaredInEnv := false }}
+  {{- range .Values.server.env }}
+    {{- if eq .name "IGGY_NODE_ADVERTISED_ADDRESS" }}{{- $declaredInEnv = true }}{{- end }}
+  {{- end }}
+  {{- if not $declaredInEnv }}
+            - name: IGGY_NODE_ADVERTISED_ADDRESS
```
mmodzelewski, line 105: When `server.env` already carries `IGGY_NODE_ADVERTISED_ADDRESS`, `server.advertisedAddress` is silently ignored - this branch never runs and the value is read nowhere else. Worth a precedence note on `server.advertisedAddress` in values.yaml or the chart README.

mmodzelewski (changes requested):
The core logic is sound, there are a few things worth fixing before merge, though. Besides the comments in-line, here are the ones that don't have a direct reference in the changed code:

- docker-compose.yml (root) — builds the Dockerfile (wildcard binds at Dockerfile:45-48), has no environment: block, and cluster is off by default, so the README-documented docker compose up quickstart now refuses boot. Needs IGGY_NODE_ADVERTISED_ADDRESS=localhost like the eight other compose files patched in this PR.
- foreign/java/java-sdk/src/test/java/org/apache/iggy/client/BaseIntegrationTest.java:92 — the testcontainer sets IGGY_TCP_ADDRESS=0.0.0.0:8090 with no advertised var; it will refuse boot once apache/iggy:edge ships this change. The C# fixture got the fix (VsrCluster.cs:294); the Java one was missed. Dormant today only because CI takes the USE_EXTERNAL_SERVER path.
- core/server/src/args.rs:57 — the --help example IGGY_TCP_ADDRESS=0.0.0.0:8090 is now boot-refusing when copy-pasted standalone; the READMEs were updated but the help text was missed.

core/configs/src/server_config/server.rs
chengxilo: resolved: )

core/configs/src/server_config/cluster.rs
chengxilo: Done

core/server/src/cluster_meta.rs
chengxilo: done

helm/charts/iggy/templates/deployment.yaml
chengxilo: documented, and it will fail if both are provided, should be much better.

core/configs/src/server_config/cluster.rs
chengxilo: After the new change,  it would call `AdvertisedAddress::is_unspecified` (`ip.to_canonical().is_unspecified()`) to check. Should be ok now.

core/server/src/bootstrap.rs
chengxilo: Now it will check and warn HTTP/QUIC/Websocket, TCP is not included, as that's where it was derived from.

core/configs/src/server_config/validators.rs
```diff
@@ -424,6 +429,38 @@ fn reject_unsupported(config: &ServerConfig) -> Result<(), ConfigurationError> {
...
+                "{COMPONENT} - tcp.address '{}' is not an address and port: {error}. The host \
+                     is required and must be a literal IP, so ':PORT' and 'hostname:PORT' are \
+                     both rejected; use 127.0.0.1:PORT for loopback or 0.0.0.0:PORT to accept on \
+                     every interface.",
+                self.tcp.address
+            );
+            ConfigurationError::InvalidConfigurationValue
+        })?;
+
+        if self.cluster.enabled || self.node.advertised_address.is_some() {
```
hubcio, line 442: this reads `tcp.address` even when `tcp.enabled = false`, and so does the derived self address in bootstrap. `IGGY_TCP_ENABLED=false` with the image default `0.0.0.0:8090` now refuses to boot over a listener that isn't running, and an http-only server with the shipped `127.0.0.1:8090` boots but publishes `127.0.0.1` with no warning (the warn loop skips wildcard listeners). derive from the first enabled listener instead, gate the refusal on that one, and warn when another listener binds a wildcard while the derived address is loopback.

core/configs/src/server_config/cluster.rs
```diff
@@ -990,10 +1047,23 @@ impl Validatable<ConfigurationError> for ClusterConfig {
...
-            if node.ip.trim().is_empty() {
+            // The roster ip is dialed verbatim for replica traffic and is
+            // never resolved, so no hostname can work here whatever its
+            // shape.
+            let node_ip = node.ip.parse::<IpAddr>().map_err(|error| {
                 eprintln!(
-                    "Invalid cluster configuration: IP cannot be empty for node '{}'",
-                    node.name
+                    "Invalid cluster configuration: IP '{}' for node '{}' is not a literal IP \
+                     address: {error}; set node.advertised_address for the name clients dial",
```
hubcio: this branch only runs with `cluster.enabled = true`, where `node.advertised_address` is ignored (bootstrap warns exactly that), so the hint sends people to a dead knob. point at `cluster.nodes[*].advertised_address`.

core/integration/tests/server/cluster_metadata_advertised.rs
```diff
@@ -0,0 +1,81 @@
...
+//! `node.advertised_address` on a cluster-disabled server: what a client is
+//! told about the one node in the roster.
+//!
+//! Without a roster the server has only its own bind address to reason from,
+//! and a bind address answers which interfaces it accepts on - never where a
+//! client reaches it. Behind any NAT (published container ports, a Service, a
+//! load balancer) the two are different addresses, and this setting is the
+//! only way to state the second one. The bind-derived fallback and the empty
+//! answer for a wildcard bind are pinned at unit level (`cluster_meta.rs`)
+//! and across the SDKs in `bdd/`; what needs a real server is that a declared
```
hubcio: there's no "empty answer for a wildcard bind" anywhere - a wildcard bind is refused at boot now, and `bdd/` doesn't assert on metadata addresses. drop that sentence. also the `!metadata.name.is_empty()` assert at line 66 can't fail (the name is a const); check for `"single-node"` instead.

core/configs/src/server_config/node.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+impl Validatable<ConfigurationError> for NodeConfig {
+    fn validate(&self) -> Result<(), ConfigurationError> {
+        let Some(address) = self.advertised_address.as_deref() else {
+            return Ok(());
+        };
+
+        let parsed = address.parse::<AdvertisedAddress>().map_err(|error| {
+            eprintln!("{COMPONENT} - node.advertised_address '{address}': {error}");
```
hubcio, line 47: for `localhost:8090` this prints "ports are configured in cluster.nodes.ports", which is wrong here - the single-node port comes from `tcp.address`. generalize the `PortNotAllowed` text so it fits both.

core/server/config.toml
```diff
@@ -757,7 +768,7 @@ skip_shard_zero_for_clients = false
 #
 #   [[cluster.nodes]]
 #   name = "iggy-node-1"
-#   ip = "10.0.1.5"                        # replica plane + last-resort fallback
+#   ip = "10.0.1.5"                        # replica plane, literal IP only
```
hubcio, line 801: since the roster ip is now literal-ip only, the `ca_file` note at lines 693-696 ("use hostnames in the roster if the certificates only have DNS SANs") can't be followed anymore: that ip string is also the tls server name. reword it to say peer certs need IP SANs.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -89,6 +89,17 @@ spec:
                   name: {{ include "iggy.fullname" . }}-root-credentials
                   key: password
   {{- end }}{{- end}}
+  {{- $declaredInEnv := false }}
+  {{- range .Values.server.env }}
+    {{- if eq .name "IGGY_NODE_ADVERTISED_ADDRESS" }}{{- $declaredInEnv = true }}{{- end }}
```
hubcio: this only checks the name, so `{name: IGGY_NODE_ADVERTISED_ADDRESS, value: ""}` in `server.env` suppresses the chart default and renders an empty value, which the server treats as unset - the pod then hits the refusal this is meant to prevent. require a non-empty `.value` here.

web/docker-compose.yml
```diff
@@ -24,6 +24,7 @@ services:
       IGGY_HTTP_ADDRESS: 0.0.0.0:3000
       IGGY_QUIC_ADDRESS: 0.0.0.0:8080
       IGGY_TCP_ADDRESS: 0.0.0.0:8090
+      IGGY_NODE_ADVERTISED_ADDRESS: iggy-server
```
hubcio: ports 8090/8080/8092 are published to the host, so clients on the host get `iggy-server` as an address they can't resolve (the go sdk keeps a single node's advertised address as a reconnect target). the root compose uses `localhost`; same here, the ui itself never reads this.

core/server/src/args.rs
```diff
@@ -54,7 +54,7 @@ ENVIRONMENT VARIABLES:
 
     Common examples:
         IGGY_SYSTEM_PATH=/data/iggy                    # Data directory
-        IGGY_TCP_ADDRESS=0.0.0.0:8090                  # TCP listener address
+        IGGY_TCP_ADDRESS=127.0.0.1:8090                # TCP listener address
```
hubcio, line 57: add `IGGY_NODE_ADVERTISED_ADDRESS=localhost` to the examples - it's the one variable the docker image now requires.

hubcio (changes requested):
two things outside the diff:

- `.devcontainer/devcontainer.json` still binds every listener to `0.0.0.0` with no `IGGY_NODE_ADVERTISED_ADDRESS`, so `cargo run --bin iggy-server` inside the devcontainer refuses to start. add `"IGGY_NODE_ADVERTISED_ADDRESS": "localhost"` there (ports are forwarded to localhost).
- keeping the refusal for the bare image is the right call (a baked `localhost` would give go clients in other containers a self-dial reconnect candidate). the website quick start only passes `-e IGGY_TCP_ADDRESS=0.0.0.0:8090` though, so it needs the extra `-e` in lock-step

optional, non-blocking:

- `ClusterConfig::validate` now repeats every per-node check that `TryFrom<ClusterNodeConfig>` already does, with near-identical messages that have started to drift (1056 has the hint, 499 doesn't). calling `ResolvedClusterNode::try_from(node.clone())?` in the loop and building the conflict pool from the resolved node drops about 70 lines, and removes the `node.ip.parse::<AdvertisedAddress>().ok()` at 1150 that can't be `None` anymore.
- parse-then-`is_unspecified()` appears 5 times; rejecting the wildcard inside `FromStr for AdvertisedAddress` (after `to_canonical()`) removes the method and all five branches, and stops `::ffff:10.0.0.1` and `10.0.0.1` being treated as two different addresses.
- helm README says the server "refuses to start without it", but the chart's default image tag `0.7.0` predates the setting and just logs an unknown-variable warning.
- `[node]` in config.toml sits between `[http.tls]` and `[tcp]`, reads like part of the http block.
- cluster_meta.rs:50 says a bracketed `[2001:db8::1]` gets bracketed twice by clients - true for the go and C# sdks, not the rust one, which just does `format!("{}:{}")`.

core/configs/src/server_config/node.rs
chengxilo: It doesn't always comes from tcp.address too, I'll make the message more generic.

web/docker-compose.yml
chengxilo: fixed

core/server/src/args.rs
chengxilo: fixed

core/configs/src/server_config/validators.rs
chengxilo: done

core/configs/src/server_config/cluster.rs
chengxilo: done

core/integration/tests/server/cluster_metadata_advertised.rs
chengxilo: done

chengxilo:
>cluster_meta.rs:50 says a bracketed [2001:db8::1] gets bracketed twice by clients - true for the go and C# sdks, not the rust one, which just does format!("{}:{}").
It appears that only Go have the problem : (, so I only mention Go in the comment.. (it's because ipv6 with [] is consider as a incorrect input by standard library)

I'll update the iggy-website btw.

numinnex (dismissed):

hubcio (dismissed):

mmodzelewski (dismissed):

hubcio (approved):

spetz (approved):

numinnex (approved):

PR #3931, opened 2026-08-20
fix(shard): tell an unknown consensus operation from a corrupt header

numinnex:
Fixes `given_an_unknown_operation_when_a_consensus_frame_decodes_should_surface_an_upgrade_fence_signal` test.


core/binary_protocol/src/consensus/header.rs
```diff
@@ -971,6 +983,7 @@ impl Default for PrepareHeader {
 }
 
 impl ConsensusHeader for PrepareHeader {
+    const OPERATION_OFFSET: Option<usize> = Some(core::mem::offset_of!(Self, operation));
```
hubcio: `RepairPrepareHeader` doesn't get an `OPERATION_OFFSET` - it's `repr(transparent)` over `PrepareHeader`, so it carries the same `operation` field. a repair-served prepare with a newer-release op falls back to `InvalidBitPattern` and gets counted as `unparsable`, on the exact path the router comment reasons about. `Some(core::mem::offset_of!(PrepareHeader, operation))` fixes it.

core/server_common/src/consensus_message.rs
```diff
@@ -774,6 +774,25 @@ impl MessageBag {
     }
 }
 
+/// Why `H`'s header bytes failed bytemuck's checked cast.
+///
+/// An operation discriminant this build does not define means the sender runs a
+/// newer release; the frame is wire-valid and the node needs upgrading, which is
+/// a different operator action from the corrupted-header case. bytemuck reports
+/// both as one error, so the operation byte is probed here to separate them.
+fn classify_failed_cast<H>(bytes: &[u8]) -> ConsensusError
```
hubcio: this runs before `verify_frame`, so `UnsupportedOperation` is claimed from an unverified byte. for a sealed header (`PrepareOkHeader` carries `operation` and is sealed) a corrupted frame whose op byte flips into 166..=255 produces an error-level "upgrade this node" plus the wrong counter - the inverse of the misdiagnosis this PR fixes. when `H::FRAME_SEALED`, recompute `frame_checksum_bytes` and compare against the stored u128 at `bytes[..16]` before claiming version skew. unsealed carriers can't be checked, worth a comment saying so.

core/shard/src/router.rs
```diff
@@ -62,17 +62,33 @@ where
...
+                // fences the sending peer, so the log and the counter are the
+                // whole signal an operator gets.
+                self.metrics.record_frame_drop(
+                    frame_drop_variant::CONSENSUS,
+                    frame_drop_reason::UNSUPPORTED_OPERATION,
+                );
+                tracing::error!(
+                    shard = self.id,
+                    operation = format_args!("{operation:#04x}"),
+                    build_release = iggy_binary_protocol::IGGY_PROTOCOL_VERSION,
```
hubcio: `IGGY_PROTOCOL_VERSION` is the packed-semver u32, so this logs an opaque int. log `env!("CARGO_PKG_VERSION")` or unpack major.minor so operators can read it.

core/shard/src/metrics.rs
```diff
@@ -134,7 +142,7 @@ pub mod frame_drop_reason {
 // site actually produces it, so the unreachable corners of the 7 x 7 cross
```
hubcio: comment says 7 x 7, table is now 7 x 9.

core/binary_protocol/src/consensus/header.rs
```diff
@@ -511,6 +521,7 @@ impl ConsensusHeader for RoutedRequestHeader {
 }
 
 impl ConsensusHeader for RequestHeader {
+    const OPERATION_OFFSET: Option<usize> = Some(core::mem::offset_of!(Self, operation));
```
hubcio: `RequestHeader` gets the offset too, so a client SDK sending a garbage op byte now decodes to `UnsupportedOperation`, whose message blames "the sender runs a release that added a consensus operation". does any client-facing surface relay that text? if so the wording misleads for the client case.

core/binary_protocol/src/consensus/header.rs
hubcio, line 993: `RepairPrepareHeader` is `repr(transparent)` over `PrepareHeader`, so it carries the same `operation` field but never sets `OPERATION_OFFSET`. a repaired prepare with a newer-release op gets counted as `unparsable` instead of hitting the fence. `Some(core::mem::offset_of!(PrepareHeader, operation))` fixes it.

core/server_common/src/consensus_message.rs
hubcio, line 784: this runs before `verify_frame`, so "sender runs a newer release" is claimed from an unverified byte. `PrepareOkHeader` is sealed and carries `operation`, so a corrupt frame whose op byte lands in 163..=255 or an interior gap draws the error-level "upgrade this node" plus the wrong counter - and a PrepareOk echoes an op this primary minted itself, so corruption is the likelier cause there. when `H::FRAME_SEALED`, recompute `frame_checksum_bytes` and compare the stored u128 at `bytes[..16]` before claiming skew.

core/shard/src/router.rs
hubcio: `IGGY_PROTOCOL_VERSION` is the packed u32, so this prints `build_release=11264`. `%ProtocolVersion(IGGY_PROTOCOL_VERSION)` renders `0.11.0`.

core/shard/src/metrics.rs
hubcio: comment says 7 x 7, table is now 7 x 9.

core/binary_protocol/src/consensus/header.rs
hubcio, line 530: `RequestHeader` sets the offset too, so a client sending a garbage op byte lands this error in the server warn log (`handle_client_request` drops the frame without a reply). only operators see it, but "cannot journal or ack the frame" fits the replica plane, not a client request.

core/shard/src/router.rs
```diff
@@ -804,25 +820,19 @@ mod tests {
...
+    /// no progress (the op is never journaled, never acked, and every later
+    /// prepare dies on the gap). `dispatch` splits its drop arms on this
+    /// distinction; the accounting half is pinned in the server crate by
+    /// `given_an_unknown_operation_when_dispatched_should_account_an_upgrade_fence_drop`.
     #[test]
-    // TODO(hubcio): fix this test
-    #[ignore = "unknown operation collapses into InvalidBitPattern; no upgrade fence"]
     fn given_an_unknown_operation_when_a_consensus_frame_decodes_should_surface_an_upgrade_fence_signal()
      {
         // Far past every defined Operation discriminant (the highest is 165).
```
hubcio: highest discriminant is 162 (`DeleteConsumerOffset`), not 165.

hubcio (changes requested):

hubcio (dismissed):

core/server_common/src/consensus_message.rs
```diff
@@ -774,6 +774,60 @@ impl MessageBag {
...
+/// likelier cause anyway. Checking the seal first keeps that claim honest.
+///
+/// An unsealed header has nothing to check against here (its `checksum` field
+/// carries an identity, not a frame seal), so the byte is taken at face value,
+/// which is the pre-existing contract for those frames.
+fn operation_byte_is_trustworthy<H>(bytes: &[u8]) -> bool
+where
+    H: ConsensusHeader,
+{
+    if !H::FRAME_SEALED {
```
spetz: This cannot distinguish an unsupported operation from a corrupted operation byte on unsealed headers. Please verify the Prepare identity checksum where available, or describe this as possible version skew rather than asserting that the sender is newer.

core/shard/src/router.rs
```diff
@@ -62,17 +62,35 @@ where
...
+                    frame_drop_variant::CONSENSUS,
+                    frame_drop_reason::UNSUPPORTED_OPERATION,
+                );
+                tracing::error!(
+                    shard = self.id,
+                    operation = format_args!("{operation:#04x}"),
+                    build_release = %iggy_binary_protocol::ProtocolVersion(
+                        iggy_binary_protocol::IGGY_PROTOCOL_VERSION
+                    ),
+                    "consensus frame carries an operation this build does not know; the sender \
```
spetz: `UnsupportedOperation` can also come from a routed request, which is dropped before journaling and does not gap-stop a consensus group. Please make this consequence frame-specific or use neutral wording here.

core/server/src/partition_reconciler.rs
```diff
@@ -2395,6 +2395,65 @@ mod tests {
...
+            let frame_size =
+                u32::try_from(iggy_binary_protocol::HEADER_SIZE).expect("header size fits in u32");
+            frame[size_offset..size_offset + 4].copy_from_slice(&frame_size.to_le_bytes());
+            frame[std::mem::offset_of!(PrepareHeader, command)] = Command::Prepare as u8;
+            frame[std::mem::offset_of!(PrepareHeader, operation)] = OPERATION_FROM_A_NEWER_RELEASE;
+            let header: &[u8; iggy_binary_protocol::HEADER_SIZE] = frame
+                [..iggy_binary_protocol::HEADER_SIZE]
+                .try_into()
+                .expect("frame spans a full header");
+            let checksum = iggy_binary_protocol::frame_checksum_bytes(header);
```
spetz: `PrepareHeader` is not frame-sealed. Its checksum is `identity_checksum`, not `frame_checksum_bytes`, so this fixture is not valid as described. Please use `CHECKSUM_UNSEALED` or construct the correct Prepare identity checksum. The router fixture has the same issue.

core/shard/src/router.rs
```diff
@@ -62,17 +62,35 @@ where
...
+                     runs a newer release. This node cannot journal or ack it, so its consensus \
+                     group stops making progress until this node is upgraded"
+                );
+                return;
+            }
```
spetz, line 99: Please add the complementary metric test: dispatch a corrupt sealed frame and assert `unparsable` increments while `unsupported_operation` remains zero.

spetz (changes requested):

core/server_common/src/consensus_message.rs
numinnex: Done — the Prepare family now verifies the identity checksum (recomputed at byte level via a new `prepare_identity_checksum_bytes`, since the unknown discriminant blocks the typed view; equality with `PrepareHeader::identity_checksum` is pinned by a unit test). A client Request has neither checksum, so its byte stays face-value and the error text now claims only "a newer release likely added it".

core/shard/src/router.rs
numinnex: Done — the log now branches on the command: Prepare/RepairPrepare/PrepareOk keep the group-stall consequence, everything else (routed requests included) says the frame is dropped before journaling with no reply. The command is included as a log field.

core/server/src/partition_reconciler.rs
numinnex: Done — both fixtures now stamp the prepare identity checksum (computed over the unknown operation byte, the way a real sender would), which is also what the classifier now verifies. The frame-seal stamping and its comment are gone.

core/shard/src/router.rs
numinnex: Added: given_a_corrupt_operation_byte_when_dispatched_should_account_an_unparsable_drop — identity stamped over operation 0, byte flipped after stamping; asserts unparsable increments and unsupported_operation stays flat.

spetz (approved):

hubcio (approved):

PR #3933, opened 2026-08-20
fix(sdk): respect max_buffer_size when merging batches

### LINKED ISSUE #3932: bug(Rust SDK): on merge of ShardMessages the buffer budget is freed before messages are send
### Bug description

**Situation**
The ShardMessageWithPermit has inner Vec<IggyMessage> for which size the _permit_bytes are taken.
Merging the batches in appends IggyMessages to the Vec without touching the _permit_bytes.
The loop drops on iteration and thereby frees the permit.
https://github.com/apache/iggy/blob/3db955eae215eb815dc2d3825c5b06e9787b77df/core/sdk/src/clients/producer_sharding.rs#L227-L235

**Consequence**
The actual bytes in flight can exceed max_buffer_size and the BackpressureMode does not kick.

### Affected area / component

Rust SDK

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

### Contribution

### Good first issue

haubur:
Closes #3932 

## Rationale
Respect the `max_buffer_size` such that the BackpressureMode can kick in.

## What changed?
Merge both inner IggyMessages and the OwnedSemaphorePermits (bytes_permit), freeing only when written, not after merge.

## Local Execution
- Passed
- Pre-commit hooks ran

## AI Usage
None


core/sdk/src/clients/producer_sharding.rs
```diff
@@ -110,16 +110,23 @@ impl Sizeable for ShardMessage {
...
-            _bytes_permit: Some(permit_bytes),
+            bytes_permit: permit_bytes,
         }
     }
+
+    /// Takes over `other`'s messages together with its byte permit, so the buffered bytes stay
+    /// charged against the `max_buffer_size` budget until the merged batch has been written.
+    fn merge(&mut self, other: Self) {
+        self.inner.messages.extend(other.inner.messages);
+        self.bytes_permit.merge(other.bytes_permit);
```
numinnex: `OwnedSemaphorePermit::merge` sums `u32` permit counts, and after this change the sum is bounded only by `max_buffer_size` — which nothing validates.

`IggyByteSize::from(u64)` accepts any value and `BackgroundConfig` has no validator, so a `max_buffer_size` above 4 GiB is constructible. With `batch_size(0)` + `batch_length(0)` and a long linger, one shard buffer accumulates adjacent same-destination batches until the merged sum passes `u32::MAX`. Note this needs only normal-sized batches — each individual charge is lossless, e.g. 4096 x 1 MiB against a 6 GiB budget; it is the sum that wraps.

Reproduced on tokio 1.53.1:

- release: wraps silently and **permanently destroys permits** (6 GiB budget leaves 1.7 GiB available after one merge + drop), after which the default `BackpressureMode::Block` blocks `send()` forever with no error and no log.
- debug: panics `attempt to add with overflow` inside the shard task, where it is swallowed — the `JoinHandle` at line 135 is never awaited, so no test in this file can observe it.

Worth separating from the pre-existing issues nearby: this is a new failure mode, since nothing summed permits before this change. On master the same configuration merely under-enforces the budget and the producer keeps running.

Suggested fix is one line at construction — reject `max_buffer_size > u32::MAX` in `BackgroundConfig` / `ProducerDispatcher::new`. That also closes the related truncation at `producer_dispatcher.rs:135`, where `IggyByteSize::as_bytes_u32()` charges `size mod 2^32` (exactly 0 at whole multiples of 4 GiB, so such a batch bypasses the budget entirely).

core/sdk/src/clients/producer_sharding.rs
```diff
@@ -110,16 +110,23 @@ impl Sizeable for ShardMessage {
...
     pub fn new(msg: ShardMessage, permit_bytes: OwnedSemaphorePermit) -> Self {
         Self {
             inner: msg,
-            _bytes_permit: Some(permit_bytes),
+            bytes_permit: permit_bytes,
         }
     }
+
+    /// Takes over `other`'s messages together with its byte permit, so the buffered bytes stay
+    /// charged against the `max_buffer_size` budget until the merged batch has been written.
```
numinnex: This doc states the new contract correctly, but the public documentation of `max_buffer_size` no longer matches it.

`core/sdk/src/clients/producer_config.rs:108-109` still says:

> Upper bound for the **bytes held in memory** across *all* shards.

After this change permits are held through `core.send_internal` until the write returns, so the budget covers buffered **plus in-flight** bytes, and it now couples to `max_in_flight`. That is the user-visible semantic change this PR makes, and it is currently undocumented.

The practical consequence is that `Block`-mode producers can block where they previously did not, most visibly on the linger-only shape (`batch_length(0)` + `batch_size(0)`) that this repo's own integration tests use at `core/integration/tests/sdk/producer/background.rs:135-136,177-178,216-217`, which pins close to the whole budget for a full write RTT. Default config pins under 3% (32 MiB budget against a 1 MiB `batch_size`), so there is no out-of-the-box cliff.

Suggest amending the `max_buffer_size` doc to say the budget spans buffered and in-flight bytes until the write completes, and calling the behavior change out in the release notes.

core/sdk/src/clients/producer_sharding.rs
haubur: @numinnex thanks for the review! Thats actually an interesting one, since tokios `Semaphore` is usize, but the `OwnedSemaphorePermit` is u32 and merge operates and returns `OwnedSemaphorePermit`s. So as you say, we acquire from a usize and then merge on a u32, which then can silently overflow.
Seems odd to me, I will also ask tokio whether this is intended.

haubur: @numinnex okay, so literally 8h ago someone ran into the same issue:
https://github.com/tokio-rs/tokio/issues/8395
It seems like they gonna change it to usize. I would wait for that, since it solves our problem?

spetz (dismissed):

hubcio (dismissed):

spetz (dismissed):

hubcio (dismissed):

spetz (approved):

mmodzelewski (approved):

PR #3934, opened 2026-08-20
fix(consensus): keep a client's dedup fence across capacity eviction

numinnex:


numinnex:
Merge after #3931

core/integration/tests/cluster/client_table_adversarial.rs
```diff
@@ -77,25 +77,21 @@ const REPLY_WAIT: Duration = Duration::from_secs(5);
...
-/// means the state machine ran the operation a second time. At-most-once
-/// holds only for clients the table happened not to evict.
-// TODO(hubcio): fix this test
-#[ignore = "capacity eviction erases a live client's dedup watermark; replay re-executes"]
+/// A committed duplicate-name rejection is the proof of re-execution: a dedup
+/// hit replays the cached success bytes, so any committed rejection means the
+/// state machine ran the operation a second time.
 #[iggy_harness(cluster_nodes = 1, server(metadata.clients_table_max = "2"))]
-async fn given_a_low_client_table_cap_when_connects_churn_should_erase_a_live_dedup_watermark(
+async fn given_a_low_client_table_cap_when_connects_churn_should_keep_a_live_dedup_watermark(
```
spetz, line 94: This test currently fails with the original duplicate-name rejection, proving request 1 was re-executed. The PR should not unignore this spec until it passes end to end.

core/consensus/src/client_table.rs
```diff
@@ -830,6 +907,13 @@ impl ClientTable {
     ///
     /// [`Operation::Register`]: iggy_binary_protocol::Operation
     pub fn remove_client(&mut self, client_id: u128) -> bool {
```
spetz: This removes every fence sharing `client_id`, including fences belonging to another user. It also erases the evicted client’s fence when disconnect cleanup commits Logout before resume, which is why the integration test fails. Fence removal needs the owning `user_id` and must distinguish explicit session termination from reconnect cleanup.

core/consensus/src/client_table.rs
```diff
@@ -407,6 +435,17 @@ pub struct ClientTable {
     slots: Vec<Option<ClientEntry>>,
     /// `client_id` -> slot index. Rebuilt on decode.
     index: HashMap<u128, usize>,
+    /// Fences of clients capacity eviction reclaimed, oldest at the front.
+    ///
+    /// Bounded by the slot count. A fence is the entry's header fields plus, at
+    /// most, the watermark request's own reply, so it costs a fraction of the
+    /// entry it replaces. Trimmed oldest-first.
+    ///
+    /// Replica-local best-effort, NOT replicated state: the bound is
```
spetz, line 513: The watermark is safety state, not a volatile reply cache. Dropping these fences on checkpoint or state transfer restores the original re-execution bug after restart or failover and lets replicas derive different watermarks from the same register commit. Please include fences in snapshot and state-transfer encoding.

core/consensus/src/client_table.rs
```diff
@@ -876,13 +965,67 @@ impl ClientTable {
...
+            .retain(|fence| fence.client_id != entry.client_id || fence.user_id != entry.user_id);
+        self.evicted_fences.push_back(EvictedFence {
+            client_id: entry.client_id,
+            epoch: entry.epoch,
+            user_id: entry.user_id,
+            watermark: entry.watermark,
+            watermark_checksum: entry.watermark_checksum,
+            latest: entry.find_cached(entry.watermark).cloned(),
+        });
+        while self.evicted_fences.len() > self.slots.len() {
```
spetz: Silently trimming the oldest fence after `slots.len()` active-client evictions only postpones the at-most-once failure. Please define and test the retention guarantee, or use durable bounded tombstones with explicit expiry rather than silently restoring watermark zero.

core/consensus/src/client_table.rs
```diff
@@ -876,13 +965,67 @@ impl ClientTable {
         let (slot_idx, _) = evictee?;
         let entry = self.slots[slot_idx].take().expect("evictee must exist");
         self.index.remove(&entry.client_id);
+        // Reclaim the replies, keep the fence: the evicted client's own resume
+        // must not read as a first-time register, or the retry of a committed
+        // request re-executes.
+        self.remember_fence(&entry);
```
spetz, line 1144: This snapshots the watermark at eviction time. If an already-prepared request commits afterward, `commit_reply` finds no entry and cannot advance this fence. A later resume restores the older watermark and may re-execute that committed request. Please add this ordering test and update the retained fence when the late commit lands.

spetz (approved):

spetz (dismissed):
Previous comments

core/integration/tests/cluster/client_table_adversarial.rs
numinnex: You were right, and the cause is the comment below: the resume's new connection triggers disconnect cleanup of the old one, whose Logout commits after the eviction and — under the previous code — erased the fence by client_id. Whether the register or that Logout committed first decided the outcome, which is why it passed here and failed for you. With cleanup now distinguished (see below) the spec passes 15/15 consecutive runs locally against a freshly built server.

core/consensus/src/client_table.rs
numinnex: Done. remove_client now takes (client_id, user_id, SessionEnd). SessionEnd is derived at apply from the Logout's request id — the server's disconnect Logout already carried a u64::MAX sentinel as a local const in dispatch.rs; it is now consensus::DISCONNECT_LOGOUT_REQUEST_ID and the apply path reads it. Explicit removes the fence for the owner only; the owner is resolved by the primary at prepare time (ClientTable::user_id_for_session: live entry, else fence matching the Logout's session epoch) and stamped into the prepare, since a Logout is ungated and otherwise carries user_id 0. An unattributed Logout leaves fences alone rather than guess. DisconnectCleanup never removes a fence and, when the entry is still live, converts it into one — a dropped transport is exactly the case the resume contract exists for.

core/consensus/src/client_table.rs
numinnex: Done — fences are now persisted in both places. ClientTableSnapshot gained a trailing #[serde(default)] fences field; SNAPSHOT_FORMAT_VERSION is 4 with MIN_READABLE_SNAPSHOT_FORMAT_VERSION = 3, so an existing checkpoint still boots (test: decode_reads_a_version_3_snapshot_as_carrying_no_fences) while anything else is refused by version as before. The state-transfer artifact gained a fence section under magic ICT3; ICT2 is still decoded as fence-less so a node on this build can join a cluster whose primary predates it, and the reverse direction fails on magic. Round-trip tests for both, plus the nested field-count pin.

core/consensus/src/client_table.rs
numinnex: Done. The guarantee is now explicit and tested: ClientTable::fence_retention() = slots.len(), i.e. a reclaimed session's watermark survives a full turnover of the table's capacity (evictions and disconnect cleanups combined). a_fence_survives_exactly_the_retention_bound pins the boundary. A fence dropped past the bound is logged at warn with client_id/user_id/watermark, so an undersized table shows up in the logs rather than as a silent re-execution; and with fences now in the checkpoint and the transfer artifact they are durable bounded tombstones rather than volatile ones.

core/consensus/src/client_table.rs
numinnex: Done. commit_reply now takes the prepare's user_id, and the no-entry path folds the late commit into the session's fence (advance_fence): a request above the fence's watermark advances it and stores its reply (new CommitReply::AdvancedFence), a request below is skipped as a regression, mirroring the live entry's rules. Tests: a_commit_landing_after_eviction_advances_the_fence (evict at watermark 1, late-commit request 2, resume — request 2 replays the late commit's own bytes and 3 is New) and a_late_commit_below_the_fence_watermark_is_skipped.

hubcio (approved):

spetz (approved):

PR #3935, opened 2026-08-20
fix(bdd): resolve server address and credentials only from env

### LINKED ISSUE #3883: fix(bdd): every SDK suite silently falls back to a hardcoded address/credentials when its env var is missing
## Summary
Every BDD step-definition file resolves the server address and root credentials with an "env var or hardcoded default" helper. If the env variable is absent, misspelled, or dropped from a compose overlay, the suite does not fail. Instead, it quietly connects to the default value. Test should be loud when it is broken, we don't need safety or robust here; a fallback turns a configuration bug into either a misleading connection error or, worse, a green run against the wrong server.
 
The fallbacks also disagree with each other across SDKs, so the *same* missing variable produces eight different behaviours.

Examples:
https://github.com/apache/iggy/blob/2a8a5c96925f8b96a7e8718448079a63ce57980c/bdd/rust/tests/helpers/cluster.rs#L23-L35

https://github.com/apache/iggy/blob/2a8a5c96925f8b96a7e8718448079a63ce57980c/bdd/go/tests/basic_messaging.go#L53-L61

https://github.com/apache/iggy/blob/2a8a5c96925f8b96a7e8718448079a63ce57980c/foreign/php/tests/bootstrap.php#L87-L95

## Proposed Change

only use the enviroment varaible, if it's not available, just panic.
For example:
```rust
pub fn resolve_server_address(role: &str, port: u16) -> String {
     match (role.to_lowercase().as_str(), port) {
         ("leader", 8091) => required_env("IGGY_TCP_ADDRESS_LEADER"),
         ("follower", 8092) => required_env("IGGY_TCP_ADDRESS_FOLLOWER"),
         ("single", 8090) | (_, 8090) => required_env("IGGY_TCP_ADDRESS"),
         _ => panic!("no address mapping for role '{role}' on port {port}"),
    }
}

fn required_env(name: &str) -> String {
    match env::var(name) {
        Ok(value) if !value.is_empty() => value,
        _ => panic!("{name} must be set; run the suite via scripts/run-bdd-tests.sh"),
    }
}
```

----

Can be a `good first issue`

chengxilo:

  Closes #3883

  ## Rationale

  A fallback turns a configuration bug into either a misleading connection error
  or, worse, a green run against the wrong server. Tests should be loud when they
  are broken.

  ## What changed?

  Each suite now reads those values through a required-env helper that aborts when
  a variable is unset or empty, and the image no longer carries defaults. The PHP
  path moves onto the key names the other seven SDKs already use
  (`IGGY_TCP_ADDRESS`, `IGGY_ROOT_USERNAME`, `IGGY_ROOT_PASSWORD`); the PHP SDK
  suite keeps its own defaults, so only the names change there. Two CI steps run
  `bdd/go` outside `run-bdd-tests.sh` and were relying on the removed defaults,
  one of them with no environment block at all; both now pass the variables
  explicitly.

  ## Local Execution

  - Passed
  - Pre-commit hooks ran

  ## AI Usage

  1. Claude (Claude Code).
  2. Implementation
  3. I reviewed the code line by line. There was some changes to PHP unit test by AI but I reverted them. I am not planning to make the unit test be affected by this change since they don't have `./scripts/run-bdd-tests.sh` to handle the chores. It's kinda annoying to export some environment variable everytime you want to run the unit tests.
  4. Yes


foreign/node/src/bdd/README.md
slbotbm, line 1: `foreign/node/README.md:124-151` still tells users that `IGGY_TCP_ADDRESS` is the only BDD configuration variable and shows `npm run test:bdd` without any environment variables. Update this README to direct users to `scripts/run-bdd-tests.sh node`, matching `foreign/node/src/bdd/README.md`.

slbotbm (changes requested):

foreign/node/src/bdd/README.md
chengxilo: tks! I updated the README.md, it should be ok now.

slbotbm (dismissed):

mmodzelewski (approved):

numinnex (approved):

PR #3942, opened 2026-08-21
feat(simulator): drive the workload under network and replica faults

krishvishal:
Faults:

* Network parameters drawn from the seed (`swarm`), or taken as a named profile. Per-command delivery counters report what a run reached.
* `FaultInjector` crashes and restarts replicas behind stability windows, so a crash lasts long enough to need repair.
* Clients resend on timeout, rotating target. Unblocks the rest: with one in-flight slot per client, a dropped frame stranded it for the run, so primaries had to be spared and view changes were unreachable.
* `SimJournal` honours a slot count, drains on checkpoint, advances its snapshot watermark. `RangeEvicted` and state transfer were unreachable while it answered a constant 0.
* Restart carries a partition's log across the rebuild. Without it, a rebuilt partition reports `commit_offset` 0, tripping the monotonicity invariant on a harness artefact.

Oracles:

* State checker folds every replica's committed metadata into one chain, asserting agreement where they overlap in both directions of “same op iff same prepare.” Reports ops seen on more than one replica, so a green run cannot mean “compared nothing.”
* Dispatch shell: pre-commit denials, result-framed transients, PAT client wire shape, logout forwarding, eviction re-login.
* `settle_to_stable_view` before the leader-relative asserts. A failed drain reports what is outstanding and what each live replica believes.

Two real bugs found:

* Consumer-offset ops drew `consumer_kind` as `u8::from(bool)`, so half of every request carried kind 0. `WireConsumer` rejects it, the partition plane drops the frame with no reply, and the client wedges.
* Metadata backup gated prepare admission on its sequencer, not its journal head. `on_start_view` adopts the announced head before bodies arrive, so the backup refused the prepare that would heal its log. The op never reaches quorum and the client is never answered. This affected any rejoining replica.


core/metadata/src/impls/metadata.rs
```diff
@@ -1170,21 +1175,48 @@ where
...
+        // replica already holds was re-acked and returned above. What reaches HERE
+        // is the next op or a gap, and not every gap is fillable: metadata repair
+        // covers only `commit_min + 1 ..= commit_max`, so an interior hole below
+        // the head and a forward gap above `commit_max` both sit outside it.
         let is_backup = consensus.is_follower();
         if is_backup {
-            if header.op != current_op + 1 {
+            let handle = journal.handle();
+            let journal_head = handle.last_op().unwrap_or(0).max(handle.snapshot_op());
+            if header.op != journal_head + 1 {
```
numinnex, line 1212: This drop arms no repair, and the PR's own fault config wedges behind it. Repro: `workload-fuzz --plane metadata --crash-prob 0.01 --restart-prob 0.05 --ticks 30000 --seed 1` (also seeds 2 and 5) fails `did not drain`.

Shape: a restarted backup adopts StartView at head N and arms repair up to N; live prepares N+1.. arrive during the repair and are dropped here. The other backup crashes, so `commit_max` is pinned (the primary needs this ack) and `on_commit` never sees `CommitOutcome::Advanced` to re-arm. The stalled-repair retry at `core/shard/src/lib.rs:5077` has no else branch for `from_op > to_op`, so the finished session is never cleared and the `is_none()` gate at `:5379` blocks a new one. The primary retransmits only `commit_max + 1`, which lands here forever. 2/3 live, zero progress.

Pre-existing (the old sequencer gate drops the same op), so not a defect of this PR, but the advertised config is red on roughly a third of seeds. Worth an issue and a mention in the PR body. Fix looks small: clear the session when `from_op > to_op`, and level-trigger the arm in `tick_metadata` when `commit_min < commit_max` with no session.

core/simulator/src/workload/oracle.rs
```diff
@@ -117,29 +139,227 @@ pub fn drive_to_quiesce(sim: &mut Simulator, workload: &mut Workload, max_ticks:
...
+            // reads as a healthy Normal primary, silently dropping every request.
+            let _ = write!(
+                report,
+                " | metadata status={:?} view={} log_view={} commit={}..{} barrier={} \
+                 transferring={} primary={}",
+                consensus.status(),
+                consensus.view(),
+                consensus.log_view(),
+                consensus.commit_min(),
+                consensus.commit_max(),
```
numinnex, line 189: Consider printing the metadata repair session here too (armed or not, `from_op..to_op`). On the seed-1 wedge this shows `commit=1278..1300` for the stuck backup, and the session state is what points at the shard retry path rather than the gap drop.

core/simulator/src/lib.rs
```diff
@@ -2078,909 +2015,1228 @@ mod tests {
...
-    #[test]
-    fn multi_client_interleaves_commits() {
+    fn workload_hash_for_seed(seed: u64) -> u64 {
+        workload_hash(seed, 1).0
+    }
+
+    /// Reply-trace and executor-schedule hashes for a full workload run at
+    /// `shards_per_replica` shards. Shared by the single-shard locked
+    /// baseline and the multi-shard replay tests.
+    fn workload_hash(seed: u64, shards_per_replica: u16) -> (u64, u64) {
```
numinnex, line 2045: `workload_hash` never touches `FaultInjector`, `run_with_faults` or `resubmit_due`, so the new fault path has no same-seed replay test; an injector drawing from `rand::random` would pass `cargo test -p simulator`.

Deleted without a replacement: both `*_solo_replica_restarts_*` tests (the only asserts that restart replay rebuilds the STM and client table), `prepare_timestamps_replay_with_seed`, `multi_shard_replay_is_deterministic` (the schedule hash this fn returns is now discarded), `shell_on_consensus_schedule_matches_shell_off`, and `crash_injection_spares_primary_and_keeps_quorum` (`spare_primary` / `min_survivors` are now unasserted). `SimSuperblock::set_yield_writes` is orphaned by the same deletion.

Suggest restoring the solo-restart pair and adding one test that runs `run_with_faults` twice on a seed and compares reply hash, `schedule_hash()`, crashes and restarts.

core/simulator/src/lib.rs
```diff
@@ -854,7 +1145,11 @@ impl Simulator {
             .collect();
         materialised.sort_unstable_by_key(IggyNamespace::inner);
         for namespace in materialised {
-            materialise_partition(&self.replicas[idx], namespace);
+            materialise_partition(
```
numinnex, line 1147: Ordering on restart: `new_shard` replays the WAL before this call seeds the namespace, while a fresh boot seeds first (`:519`). `CreateStream::apply` takes `vacant_key()` at apply time (`core/metadata/src/stm/stream.rs:1607`), so a restarted replica ends up with `wl-*` streams at slabs 0-1 and the fillers above, the reverse of its peers. Same committed log, different slab ids.

Nothing observes it today (the state checker compares headers, the entity oracle filters by name), but a `DeleteStream` hitting the impersonating slab would surface as a drain failure blamed on the cluster. Seeding before the replay loop in `new_shard`, or journaling the seed, closes it.

numinnex (approved):

spetz (approved):

PR #3944, opened 2026-08-21
feat(sdk): fail over to a surviving node when the current one dies

numinnex:

PR #3945, opened 2026-08-21
fix(consensus): keep a late retry's reply instead of refusing it

numinnex:


numinnex:
Merge after #3944

core/integration/tests/cluster/client_table_adversarial.rs
```diff
@@ -165,8 +163,8 @@ async fn given_replays_past_the_reply_ring_when_a_request_id_falls_off_should_no
     let aged_payload = create_stream_payload("adv-l-aged");
     let committed = commit_request(&mut stream, CLIENT_A, session, 1, &aged_payload).await;
 
-    // One more commit than the ring holds, so request 1's reply is evicted
-    // even though the register reply seeded a slot of its own.
+    // More commits than the unconditional floor holds, so passing depends on
+    // the byte-budgeted retention rather than on the floor alone.
     let later_requests = REPLY_RING_CAPACITY as u64 + 1;
```
spetz, line 177: This test no longer exercises the behavior described by its name. Request 1 does not fall out of the cache. The test only proves that these replies fit within the new retention budget. `check_request_below_watermark_past_retention_is_already_applied` still confirms that a sufficiently late retry receives terminal `RequestAlreadyApplied`. Please either restore a test where the reply actually ages out and fix that outcome, or explicitly reframe this PR as a bounded cache-depth mitigation rather than a fix for replies that have fallen out.

core/consensus/src/client_table.rs
```diff
@@ -1679,6 +1710,19 @@ impl ClientEntry {
...
     }
 
+    /// The newest replies a state transfer carries, oldest first.
+    ///
+    /// Capped at [`REPLY_RING_CAPACITY`] rather than shipping whatever
+    /// retention holds locally: an artifact a recovering node has to fetch is
+    /// worth keeping small, and the deeper history rebuilds itself from the
+    /// receiver's own commits. Dropping it costs a late retrier its bytes on
+    /// the transferred node, which the watermark still refuses to re-execute.
+    fn transferable_replies(&self) -> impl Iterator<Item = &CachedReply> {
```
spetz, line 1749: Capping state transfer at `REPLY_RING_CAPACITY` discards the deeper history introduced by this PR. A request that remains replayable on the serving replica can become `AlreadyApplied` immediately after transfer, and the receiving replica can later become primary. Checkpoint recovery has the same problem because `to_snapshot` stores only `entry.latest()`. Please preserve the advertised retention consistently across recovery and state transfer, or explicitly define and test that the guarantee is limited to uninterrupted in-memory operation.

core/consensus/src/client_table.rs
```diff
@@ -1556,15 +1588,14 @@ impl ClientTable {
...
             // The artifact checksum only proves the bytes survived transit; it
-            // says nothing about the peer that computed them. An over-capacity
-            // ring is admitted forever after (`push_latest` evicts only on
-            // equality) and eventually wraps `encode`'s `u8` length, making
-            // this table permanently un-transferable onward.
-            if usize::from(ring_len) > REPLY_RING_CAPACITY {
+            // says nothing about the peer that computed them. A ring longer
+            // than this crate retains eventually wraps `encode`'s `u8` length,
+            // making this table permanently un-transferable onward.
+            if usize::from(ring_len) > REPLY_RING_MAX_CAPACITY {
```
spetz: The decoder now accepts up to 64 replies, but this version's encoder emits at most `REPLY_RING_CAPACITY`, which is five. It also installs the received ring without applying the byte-retention limit. This allows peer input to create an untrimmed state that the local encoder cannot produce. Please keep the wire validation bound aligned with `transferable_replies()`, or validate both the count and byte budget if larger transferred rings are intended.

core/integration/tests/cluster/client_table_adversarial.rs
spetz, line 20: This module documentation is stale. The test is no longer ignored or expected to fail, and entries may retain more than `REPLY_RING_CAPACITY` replies. Please update this description to match the new behavior and its bounded retention semantics.

spetz (changes requested):

core/consensus/src/client_table.rs
```diff
@@ -1687,14 +1731,29 @@ impl ClientEntry {
...
+    /// [`REPLY_RING_CAPACITY`] and either retention bound is exceeded.
+    ///
+    /// The total is summed here rather than denormalized onto the entry: the
+    /// ring is capped at [`REPLY_RING_MAX_CAPACITY`], so this is a bounded walk
+    /// of buffer lengths with no header casts, and it leaves no running total
+    /// for the sites that write the ring to drift out of sync with.
+    fn trim_ring(&mut self) {
+        let mut bytes: usize = self.ring.iter().map(CachedReply::byte_len).sum();
+        while self.ring.len() > REPLY_RING_CAPACITY
+            && (self.ring.len() > REPLY_RING_MAX_CAPACITY || bytes > REPLY_RING_RETENTION_BYTES)
```
hubcio: this disjunct can never fire. `ReplyHeader` is 256 bytes and `try_into_typed` rejects any backing shorter than the header, so `bytes > REPLY_RING_RETENTION_BYTES` already trips at 33 replies - `ring.len() > REPLY_RING_MAX_CAPACITY` at 64 is unreachable, and the tiny-reply flood its doc describes can't exist. the real cap is 8192 / 256 = 32.

drop the disjunct here, then either delete the constant or define it as `REPLY_RING_RETENTION_BYTES / size_of::<ReplyHeader>()` and keep it only for the wire check. its doc at line 166 is stale either way - `encode` writes `transferable_replies().count()` now, so the u8 limit constrains `REPLY_RING_CAPACITY`, not this one. `decode_rejects_a_ring_longer_than_capacity` at line 3215 needs a rename to match whichever bound survives.

core/consensus/src/client_table.rs
```diff
@@ -135,8 +140,36 @@ pub const CLIENTS_TABLE_SLOT_MAX: usize = 1 << 16;
...
+/// can only be answered "already applied, reply gone", which tells the caller
+/// its operation succeeded but hands back no result. Sizing the extra
+/// retention in bytes rather than in replies spends the same memory where it
+/// buys the most history.
+///
+/// The added cost is bounded by this budget per occupied slot, and it is less
+/// than the [`REPLY_RING_CAPACITY`] floor already costs for any reply above
+/// roughly 1.6 KiB, so a table full of large replies is dominated by the floor
+/// exactly as it was before.
+pub const REPLY_RING_RETENTION_BYTES: usize = 8 * 1024;
```
hubcio: worth spelling out what this actually costs. at the default `clients_table_max` of 8192 with header-only replies (256 bytes, the common metadata reply), per-slot retention goes from 5 x 256 = 1.25 KiB to 8 KiB, so a full table goes from 10 MiB to 64 MiB and live `Frozen` buffers from 41k to 262k. at the 65536 ceiling it's 512 MiB. the doc reads cost-neutral ("spends the same memory where it buys the most history") but it's 6.4x on the common case - say the multiplied figure.

also, where does 8 KiB come from? nothing ties 32 replies of depth to an observed retry depth or to how wide clients actually pipeline.

hubcio (changes requested):
one more that lands on a line outside the diff, so leaving it here:

`encode_decode_roundtrip_preserves_dedup_state` (core/consensus/src/client_table.rs:3044) builds a 3-deep ring, so the new `transferable_replies()` cap is exercised by nothing - shipping 0, 5, or the whole ring would all pass. commit more than `REPLY_RING_CAPACITY` replies, encode/decode, then assert exactly the newest 5 replay and the 6th-newest draws `AlreadyApplied`. that's also the concrete test for the transfer guarantee question on line 1720.

spetz (approved):

hubcio (approved):

PR #3948, opened 2026-08-21
feat(common): add deferred poll timeout support

### LINKED ISSUE #3470: feat: add deferred polling timeout to PollMessages

arunsingh:

Closes #3470

## Rationale

`PollMessages` needs a backward-compatible way for clients to request a bounded wait when no messages are available. Immediate polling remains unchanged for existing clients and for a zero timeout.

## What changed?

Empty `PollMessages` requests can now carry a trailing `wait_timeout_us` field; legacy requests decode it as zero, while partial trailing fields are rejected. The common, SDK, consumer-group, simulator, and HTTP-zero paths carry the value through the current tree.

This is a clean rebuild after closed PR #3605. The active `core/server` path currently rejects non-zero waits with typed `FeatureUnavailable` until a safe waiter/notifier implementation is designed. No `core/server-ng`, legacy server, benchmark markdown, or runbook files are included.

## Local Execution

- `cargo fmt --all` — passed
- `cargo clippy --all-targets --all-features -- -D warnings` — passed
- `cargo test -p iggy_binary_protocol poll_messages` — 15 passed
- `cargo test -p iggy_common wait_timeout` — 2 passed
- `cargo check -p iggy -p server -p simulator` — passed
- `cargo check -p integration --tests` — passed
- `cargo machete` — passed
- `cargo sort --workspace` — passed; no changes needed
- `typos` — passed
- `./scripts/ci/license-headers.sh --check` — passed with HawkEye 7
- `git diff --check` — passed
- Pre-commit hooks — not run; `prek` is not installed in this environment

`cargo build --workspace` was attempted but could not link in this x86_64-targeted macOS environment because the installed Homebrew `xz` and `hwloc` libraries are arm64. Full local `cargo test` is not claimed. Fresh GitHub CI run `32720521140` passed all 74 jobs after the rebase onto current `master`.

## AI Usage

AI was used for code review, and test/CI orchestration. The changed code was reviewed and verified with the local checks above and GitHub CI; I can explain every line of the code if asked.

### DIFF
--- a/core/binary_protocol/src/requests/messages/poll_messages.rs
+++ b/core/binary_protocol/src/requests/messages/poll_messages.rs
@@ -17,7 +17,7 @@
 
 use crate::WireError;
 use crate::WireIdentifier;
-use crate::codec::{WireDecode, WireEncode, read_u8, read_u32_le};
+use crate::codec::{WireDecode, WireEncode, read_u8, read_u32_le, read_u64_le};
 use crate::primitives::consumer::WireConsumer;
 use crate::primitives::polling_strategy::WirePollingStrategy;
 use bytes::{BufMut, BytesMut};
@@ -27,7 +27,7 @@ use bytes::{BufMut, BytesMut};
 /// Wire format:
 /// ```text
 /// [consumer][stream_id][topic_id][partition_flag:1][partition_id:4 LE]
-/// [strategy:9][count:4 LE][auto_commit:1]
+/// [strategy:9][count:4 LE][auto_commit:1][wait_timeout_us:8 LE]
 /// ```
 ///
 /// `partition_id` encoding: a u8 flag (1=Some, 0=None) followed by 4 bytes
@@ -41,6 +41,7 @@ pub struct PollMessagesRequest {
     pub strategy: WirePollingStrategy,
     pub count: u32,
     pub auto_commit: bool,
+    pub wait_timeout_us: u64,
 }
 
 const PARTITION_FLAG_SIZE: usize = 1;
@@ -48,6 +49,7 @@ const PARTITION_VALUE_SIZE: usize = 4;
 const STRATEGY_SIZE: usize = 9;
 const COUNT_SIZE: usize = 4;
 const AUTO_COMMIT_SIZE: usize = 1;
+const WAIT_TIMEOUT_SIZE: usize = 8;
 
 impl WireEncode for PollMessagesRequest {
     fn encoded_size(&self) -> usize {
@@ -59,6 +61,7 @@ impl WireEncode for PollMessagesRequest {
             + STRATEGY_SIZE
             + COUNT_SIZE
             + AUTO_COMMIT_SIZE
+            + WAIT_TIMEOUT_SIZE
     }
 
     fn encode(&self, buf: &mut BytesMut) {
@@ -75,6 +78,7 @@ impl WireEncode for PollMessagesRequest {
         self.strategy.encode(buf);
         buf.put_u32_le(self.count);
         buf.put_u8(u8::from(self.auto_commit));
+        buf.put_u64_le(self.wait_timeout_us);
     }
 }
 
@@ -104,6 +108,13 @@ impl WireDecode for PollMessagesRequest {
         pos += 4;
         let auto_commit = read_u8(buf, pos)? != 0;
         pos += 1;
+        let wait_timeout_us = if buf.len() == pos {
+            0
+        } else {
+            let wait_timeout_us = read_u64_le(buf, pos)?;
+            pos += WAIT_TIMEOUT_SIZE;
+            wait_timeout_us
+        };
 
         Ok((
             Self {
@@ -114,6 +125,7 @@ impl WireDecode for PollMessagesRequest {
                 strategy,
                 count,
                 auto_commit,
+                wait_timeout_us,
             },
             pos,
         ))
@@ -124,9 +136,8 @@ impl WireDecode for PollMessagesRequest {
 mod tests {
     use super::*;
 
-    #[test]
-    fn roundtrip_with_partition() {
-        let req = PollMessagesRequest {
+    fn request(wait_timeout_us: u64) -> PollMessagesRequest {
+        PollMessagesRequest {
             consumer: WireConsumer::consumer(WireIdentifier::numeric(1)),
             stream_id: WireIdentifier::numeric(10),
             topic_id: WireIdentifier::numeric(20),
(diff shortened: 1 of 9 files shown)


arunsingh:
Thanks, for reviewing earlier old PR #3605. I reopened this cleanly against current master and moved the implementation to the active core/server path. This is trailing PR to #3605, I kept the markdown/runbook artifact out of the new PR.

core/common/src/traits/message_client.rs
```diff
@@ -41,6 +42,38 @@ pub trait MessageClient {
         auto_commit: bool,
     ) -> Result<PolledMessages, IggyError>;
 
+    /// Poll messages and wait up to `wait_timeout` when no messages are
+    /// immediately available. A zero timeout preserves immediate polling.
+    /// Transports without deferred-poll support return `FeatureUnavailable`.
+    #[allow(clippy::too_many_arguments)]
+    async fn poll_messages_with_timeout(
```
hubcio, line 49: any non-zero timeout fails on every transport, so this is a public method with no working path. it also freezes the shape (8 positional args, `Duration`) before a server exists to validate it - hold the trait change until the wait lands.

core/common/src/traits/binary_impls/messages.rs
```diff
@@ -205,6 +225,7 @@ async fn poll_group_messages<B: BinaryClient>(
             strategy: polling_strategy_to_wire(strategy),
             count,
             auto_commit,
+            wait_timeout_us,
```
hubcio, line 228: group polls pick one assigned partition round-robin, so a wait here parks the consumer on that partition while the others may already have data. the group case needs a design answer before the field goes on the wire.

core/server/src/dispatch.rs
```diff
@@ -2115,6 +2115,32 @@ async fn evict_stale_client<B, MJ, S, SB>(
     }
 }
 
+/// Reject a non-zero wait timeout until active-server deferred waits are implemented.
+async fn reject_deferred_poll<B, MJ, S, SB>(
```
hubcio, line 2119: this is the whole server side: log and deny. the issue asks for the wait itself - port the #3605 waiter to this server instead of stubbing.

core/server/src/dispatch.rs
```diff
@@ -2491,6 +2521,9 @@ where
     S: 'static,
     SB: SuperblockStore + 'static,
 {
+    if wire.wait_timeout_us != 0 {
```
hubcio, line 2524: dead check - `handle_poll_messages` already returned at line 2175 and the http wire hardcodes 0. drop it.

core/integration/tests/server/poll_semantics_vsr.rs
```diff
@@ -88,6 +88,49 @@ async fn given_missing_partition_when_polling_should_reject_partition_not_found(
     assert_eq!(valid.messages.len(), 0, "empty topic polls empty");
 }
 
+#[iggy_harness(test_client_transport = [Tcp])]
+async fn given_non_zero_wait_timeout_when_polling_should_reject_feature_unavailable(
```
hubcio, line 92: this pins the stub, not a behavior anyone wants. it goes away the day the wait lands - drop it, or make it the real wait test.

core/binary_protocol/src/requests/messages/poll_messages.rs
```diff
@@ -104,6 +108,13 @@ impl WireDecode for PollMessagesRequest {
         pos += 4;
         let auto_commit = read_u8(buf, pos)? != 0;
         pos += 1;
+        let wait_timeout_us = if buf.len() == pos {
```
hubcio, line 111: presence is sniffed from the remaining length, so the next optional field has to repeat the trick. a flags byte or a version would age better.

core/common/src/traits/binary_impls/messages.rs
```diff
@@ -37,11 +37,27 @@ use iggy_binary_protocol::requests::messages::{
...
 
+struct PollGroupOptions<'a> {
+    consumer: &'a Consumer,
+    strategy: &'a PollingStrategy,
+    count: u32,
+    auto_commit: bool,
+    wait_timeout_us: u64,
+}
+
+fn duration_to_wait_timeout_us(wait_timeout: Duration) -> Result<u64, IggyError> {
```
hubcio, line 54: sub-microsecond timeouts truncate to 0 and silently become an immediate poll. round up or reject.

hubcio (changes requested):
i dont see a point in merging this PR because there is no working path: every non-zero timeout is rejected by the server and by the SDK default impl, so this merges a public API and a wire field for a feature that does not exist. either put the waiter in this PR or cut it down to the protocol hunk and drop the trait method and the rejection test. `Closes #3470` has to go either way - merging this would close the issue with nothing implemented. `IggyConsumer` still sleeps on `poll_interval`, so the high level consumer needs another pass even after the server side lands.

hubcio:
closing this PR due to inactivity of author, feel free to reopen new one with addressed comments.

PR #3951, opened 2026-08-22
fix(connectors): propagate sink plugin consume failure status

### LINKED ISSUE #3950: connectors: runtime discards the sink consume FFI status, so failed batches are counted as processed
## The shape

`iggy_sink_consume` is declared `-> i32` and the SDK returns non-zero on failure (`-1` when the batch cannot be deserialized, `1` when the sink's `consume()` errors — `sdk/src/sink.rs`). The runtime discards that status: `process_messages` calls `(consume)(...)` without binding the result and returns `Ok(SinkBatchTiming { .. })` unconditionally (`runtime/src/sink.rs`).

Since the consumer commits offsets at poll time (`AutoCommitWhen::PollingMessages`), a sink that cannot write still increments `messages_processed`, keeps `status: running` with no `last_error`, and the batch is permanently lost.

The same file checks the same channel for `iggy_sink_open` (`if result != 0 → Err`); only the per-batch call is unchecked.

## Reproduced end-to-end

postgres_source → server → postgres_sink on master, with a simulated sink DB failure: a 6-row source produces a 5-row sink. The plugin logs the insert error, the runtime still logs `Consumed 1 messages`, `/sinks` reports `running`, and the lost message is never redelivered.

## Fix

Bind the status; on non-zero, return `RuntimeError::SinkConsumeFailed { plugin_id, status }` — the caller's existing `Err` path already logs the failure and skips the processed counter. I have this ready with a regression test (stub `extern "C"` callbacks; the test fails on master with `expected SinkConsumeFailed, got success`).

---

**Affected area**: connectors runtime (`core/connectors/runtime/src/sink.rs`)
**Version**: master @ `cc269ef70`
**Logs** (from the E2E repro):
```
ERROR connector_target="iggy_connector_postgres_sink" Failed to insert batch: Cannot store data: ...
INFO  iggy_connectors::sink: Consumed 1 messages in 3.9ms for sink connector with ID: 2
```

## Scope

Propagating the status is fail-fast only — it surfaces the failure but does not redeliver the batch, since offsets are already committed at poll time. The full fix is the offset-commit/DLQ design discussed in #2927.

minleejae:
Binds the `iggy_sink_consume` FFI status and returns `RuntimeError::SinkConsumeFailed { plugin_id, status }` on non-zero, matching the existing `iggy_sink_open` check in the same file. The caller's existing `Err` path handles logging and skips the processed counter, so a failed batch is no longer counted as processed.

Includes a regression test with stub `extern "C"` callbacks for both paths — the failure-path test fails on master with `expected SinkConsumeFailed, got success` and passes with this change.

Closes #3950

### DIFF
--- a/core/connectors/runtime/src/sink.rs
+++ b/core/connectors/runtime/src/sink.rs
@@ -737,7 +737,7 @@ async fn process_messages(
     })?;
 
     let ffi_start = Instant::now();
-    (consume)(
+    let status = (consume)(
         plugin_id,
         topic_meta.as_ptr(),
         topic_meta.len(),
@@ -747,6 +747,18 @@ async fn process_messages(
         messages.len(),
     );
     let ffi_elapsed = ffi_start.elapsed();
+
+    // The status code is the plugin's only channel for reporting a failed write:
+    // the SDK returns non-zero when the sink's consume() errors or the batch cannot
+    // be deserialized. Ignoring it would count the batch as processed and advance
+    // consumer offsets over messages the sink never stored — the same silent-loss
+    // class that the iggy_sink_open status check prevents at startup.
+    if status != 0 {
+        error!(
+            "Sink plugin consume failed with status: {status} for sink connector with ID: {plugin_id}"
+        );
+        return Err(RuntimeError::SinkConsumeFailed { plugin_id, status });
+    }
 
     Ok(SinkBatchTiming {
         processed_count,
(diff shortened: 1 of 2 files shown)


core/connectors/runtime/src/sink.rs
```diff
@@ -748,6 +748,18 @@ async fn process_messages(
     );
     let ffi_elapsed = ffi_start.elapsed();
 
+    // The status code is the plugin's only channel for reporting a failed write:
+    // the SDK returns non-zero when the sink's consume() errors or the batch cannot
+    // be deserialized. Ignoring it would count the batch as processed and advance
+    // consumer offsets over messages the sink never stored — the same silent-loss
+    // class that the iggy_sink_open status check prevents at startup.
```
kriti-sc: while I agree that the plugin should exit when plugin returns non-zero status, simply exitting the plugin does not solve the problem highlighted in the issue. 

I think we should fix the comment and update the issue description to reflect this. 

Also see discussion: https://github.com/apache/iggy/issues/2927

kriti-sc (changes requested):

core/connectors/runtime/src/sink.rs
minleejae: You're right — reworded the comment and updated the issue description: this change is fail-fast (surfaces the failure, stops counting failed batches as processed), not redelivery, since offsets are already committed at poll time. Happy to adjust this PR to whatever design lands in #2927.

hubcio:
closing in favour of #3954, which does the same status bind at `sink.rs:740` plus the offset-commit half you scoped out of here - so a rejected batch isn't just reported, it's redelivered.

nothing wrong with this patch; #3950's repro is what got the runtime fix prioritized. flagged on #3954 that your process_messages unit tests should carry over - it only has integration coverage of that path.

#3950 stays open until #3954 merges. thanks for the repro. @kriti-sc

PR #3953, opened 2026-08-22
fix(sdk): error callback should await shard flush on producer shutdown

### LINKED ISSUE #3947: bug(Rust SDK): producer shutdown() error_callback is closed immediately, misses potential flush failures
### Bug description

**Situation**
On shutdown() the IggyProducer calls the shutdown() method of the Dispatcher which broadcasts
a termination signal to it's workers: https://github.com/apache/iggy/blob/cc269ef70ef2d78af5102f5560b5b03c9dd81401/core/sdk/src/clients/producer_dispatcher.rs#L187

this immediately breaks the loop of the callback task: https://github.com/apache/iggy/blob/cc269ef70ef2d78af5102f5560b5b03c9dd81401/core/sdk/src/clients/producer_dispatcher.rs#L53-L74

The shutdown() of the Shards, however, is graceful and everything in buffer still gets flushed.
If the flush fails, messages that were supposed to be handled by a ErrorCallback are not, because nothing listens anymore on that channel and the error remains silent:
https://github.com/apache/iggy/blob/cc269ef70ef2d78af5102f5560b5b03c9dd81401/core/sdk/src/clients/producer_sharding.rs#L268

**Problem**
Messages that were not send and come back at ErrorCtx.messages cannot be handled and are permanently lost.
On default, which logs the ErrorCtx the log does not trigger.

### Affected area / component

Rust SDK

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

### Contribution

### Good first issue

haubur:
Closes #3947 

## Rationale
On shutdown() the IggyProducer calls the shutdown() method of the Dispatcher which broadcasts
a termination signal to it's workers. This immediately breaks the loop of the error callback task, which then
does not receive any errors from the final flush anymore.

## What changed?
The error callback task does not receive the stop signal anymore. The channel is still open,
and shards can put errors in the queue. On join in shutdown() (after shard tasks are joined) the
error callback is awaited for as long as the queue is emptied.

## Local Execution
- Passed 
- Pre-commit hooks ran

## AI Usage
Test proposed by Claude code.


core/sdk/src/clients/producer_dispatcher.rs
```diff
@@ -49,29 +49,17 @@ impl ProducerDispatcher {
         let err_callback = config.error_callback.clone();
         let (stop_tx, _) = broadcast::channel::<()>(1);
 
-        let mut stop_rx = stop_tx.subscribe();
+        // A task that receives errors from shards when writing messages failed.
```
hubcio: restates what the spawn does, drop it.

core/sdk/src/clients/producer_dispatcher.rs
```diff
@@ -192,6 +180,8 @@ impl ProducerDispatcher {
             }
         }
 
+        // After shards are closed await the error callback task,
+        // that might drain queued errors from the final flush.
         if let Err(e) = self._join_handle.await {
```
hubcio: `_join_handle` is awaited here, the underscore prefix says unused.

core/sdk/src/clients/producer_dispatcher.rs
```diff
@@ -432,4 +422,52 @@ mod tests {
...
+            )
+            .await
+            .unwrap();
+
+        // trigger the final flush
+        dispatcher.shutdown().await;
+
+        // Passes, if the error callback is hit once after final
+        // flush from shutdown.
+        assert_eq!(error_called.load(Ordering::SeqCst), 1);
```
hubcio, line 470: the stub drops `ctx`, so this proves the callback ran but not that the failed batch came back with it - which is the point of #3947. record `ctx.messages.len()` in the stub and assert it is 1.

core/sdk/src/clients/producer_dispatcher.rs
```diff
@@ -432,4 +422,52 @@ mod tests {
         assert_eq!(sharding_called.load(Ordering::SeqCst), 1);
         assert_eq!(error_called.load(Ordering::SeqCst), 1);
     }
+
+    #[tokio::test]
+    async fn test_shutdown_reports_errors_from_final_flush() {
+        let mut mock = MockProducerCoreBackend::new();
+        // mock a failed send to be caught by the error callback
+        mock.expect_send_internal()
+            .times(1)
```
hubcio: `.times(1)` is not enforced here - the last `Arc` owner of the mock is the shard task, so a violated expectation panics there and `shutdown()` swallows it as "shard panicked". the `error_called` assert carries the test, just don't rely on `times`.

core/sdk/src/clients/producer_dispatcher.rs
```diff
@@ -432,4 +422,52 @@ mod tests {
         assert_eq!(sharding_called.load(Ordering::SeqCst), 1);
         assert_eq!(error_called.load(Ordering::SeqCst), 1);
     }
+
+    #[tokio::test]
+    async fn test_shutdown_reports_errors_from_final_flush() {
+        let mut mock = MockProducerCoreBackend::new();
+        // mock a failed send to be caught by the error callback
```
hubcio: also at 448, 466, 469. these restate the code, drop them (and the stray "block/ ").

hubcio (changes requested):

core/sdk/src/clients/producer_dispatcher.rs
haubur: True. Didnt touch it, but removed the prefix.

core/sdk/src/clients/producer_dispatcher.rs
haubur: makes sense, thanks

core/sdk/src/clients/producer_dispatcher.rs
haubur: This also applies to `test_custom_sharding_and_error_callback` then. Changed accordingly.

hubcio (approved):

mmodzelewski (approved):

PR #3955, opened 2026-08-22
test(connectors): add Apache Fluss sink

adamrecsko:
testing workflow


adamrecsko:
I've sent this by accident, I wanted to create this on my fork's master to see if the full build flow pass

PR #3958, opened 2026-08-24
feat(sdk): consume the request counter for partition operations

numinnex:
Our partition operations did not use the `request` counter which meant that partition ops did not participate in the deduplication process. This PR makes the partition operations bump the request counter for all of the SDKs


numinnex:
Merge after #3945

core/sdk/src/vsr.rs
```diff
@@ -96,26 +96,22 @@ pub(crate) fn encode_request_header(
...
     };
-    // Stamped only for ops the server's `ClientTable` dedups. Partition ops are
-    // at-least-once with no reply cache to poison, and theirs are the large payloads,
-    // already covered client-side by `batch_checksum` over the same bytes.
-    // NonReplicated ops bypass dedup too.
-    let request_checksum = if operation.is_partition() || operation == Operation::NonReplicated {
+    // Every replicated op stamps: the stamp is what lets a dedup layer tell a
+    // genuine retry from a request id reused for different bytes.
+    // NonReplicated ops bypass dedup, so the hash pass buys nothing there.
+    let request_checksum = if operation == Operation::NonReplicated {
```
hubcio: dropping `is_partition()` costs an extra full hash on both sides per send - the sdk already hashes the batch at `send_messages.rs:177`, and the server rehashes it at `dispatch.rs:1009` now that the stamp is nonzero. nothing reads it for partition ops, so split it out until partition dedup lands.

core/sdk/src/vsr.rs
```diff
@@ -96,26 +96,22 @@ pub(crate) fn encode_request_header(
                     session.session().unwrap_or(0),
                 )
             } else if operation.is_partition() {
```
hubcio: this arm is identical to the `else` below it. it's also a live `clippy::if_same_then_else` - delete just the comment and `-D warnings` fails, so the comment is what holds the lint open. collapse both arms.

core/sdk/src/vsr.rs
```diff
@@ -96,26 +96,22 @@ pub(crate) fn encode_request_header(
...
                 (operation, session.next_request_id(), session_id)
             }
         }
     };
-    // Stamped only for ops the server's `ClientTable` dedups. Partition ops are
-    // at-least-once with no reply cache to poison, and theirs are the large payloads,
-    // already covered client-side by `batch_checksum` over the same bytes.
-    // NonReplicated ops bypass dedup too.
-    let request_checksum = if operation.is_partition() || operation == Operation::NonReplicated {
+    // Every replicated op stamps: the stamp is what lets a dedup layer tell a
```
hubcio: only the rust sdk stamps this. go, node, c# and java write zero for every op and each documents that as a protocol contract, so say the rust sdk stamps rather than claiming a global rule.

core/sdk/src/vsr.rs
```diff
@@ -487,26 +485,56 @@ mod tests {
     }
 
     #[test]
-    fn request_checksum_is_stamped_only_for_deduped_operations() {
-        // The stamp exists to stop a reused `request` number returning the wrong
-        // cached reply, so it is worth its hashing pass only where `ClientTable`
-        // dedups. Partition payloads are the large ones and carry `batch_checksum`
-        // over the same bytes already; hashing them again is pure cost.
+    fn request_checksum_is_stamped_for_replicated_operations() {
```
hubcio: this hashes the same `payload` the encoder was handed, so it can't catch a slicing mistake between what gets hashed and what gets framed. encode, slice at `HEADER_SIZE`, and assert against the frame bytes instead.

core/sdk/src/vsr.rs
```diff
@@ -487,26 +485,56 @@ mod tests {
...
+            decode_request_header(&partition).request_checksum,
             u128::from(calculate_checksum(&payload)),
         );
 
         let ping = encode_contiguous_request(&mut session, PING_CODE, &Bytes::new()).unwrap();
         assert_eq!(decode_request_header(&ping).request_checksum, 0);
     }
 
+    #[test]
+    fn partition_request_consumes_the_request_counter() {
```
hubcio, line 532: this restates `next_request_id()`. the claim that matters is server-side - n partition sends, then a metadata op that still dedups - and nothing covers it.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrRequestEncoder.java
```diff
@@ -77,10 +77,12 @@ public ByteBuf encode(ByteBufAllocator alloc, int commandCode, ByteBuf payload)
                 requestId = session.nextCorrelationId();
                 sessionId = session.sessionOrZero();
             } else if (VsrOperation.isPartition(operation)) {
```
hubcio: same as the `else` below it - both do `boundSession()` then `nextRequestId()`. go, node and c# all collapsed this branch, so drop it here too.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/ConsensusSession.java
```diff
@@ -71,7 +71,7 @@ synchronized void bind(long sessionEpoch) {
         this.session = sessionEpoch;
     }
 
-    /** Replicated metadata ops consume the monotonic VSR dedup counter. */
+    /** Replicated ops (metadata and partition) consume the monotonic VSR dedup counter. */
     synchronized long nextRequestId() {
```
hubcio, line 93: bare `requestCounter++` on a signed long with no exhaustion guard - it goes negative at 2^63. the other four sdks all guard, and partition ops now burn ids per batch instead of per admin call.

foreign/node/src/wire/vsr/index.ts
```diff
@@ -20,11 +20,7 @@ import type { CommandResponse } from '../../client/client.type.js';
 import { COMMAND_CODE } from '../command.code.js';
 import { responseError } from '../error.utils.js';
 import { HEADER_SIZE, encodeRequestHeader } from './header.js';
-import {
-  Operation,
-  isPartition,
-  operationForCode,
-} from './operation.js';
+import { Operation, operationForCode } from './operation.js';
```
hubcio, line 23: `isPartition` has no production caller left in node or c# now, and go's only backs the parity test. delete them with their tests, or say in the pr body they're the seam partition dedup reattaches to.

hubcio (changes requested):
the id half looks right and lands uniformly in all five sdks. the checksum half doesn't belong in the same pr. inline comments cover what's inside the diff - the rest sits outside it:

`core/server/src/dispatch.rs:1594-1604` - the partition path overwrites `new_header.client` with `transport_client_id` and clamps `request` to `max(1)`. so the client half of a (client, request) dedup key is already discarded before the partition plane sees it, and `transport_client_id` changes on the very reconnect a retry follows. minting distinct ids buys a future dedup layer nothing until that rewrite changes too. the comment sitting right there - "a zero request id (the SDK does not number data-plane ops) is normalized" - is falsified by this pr and should be updated in it.

`core/sdk/src/tcp/tcp_client.rs:1266` - the new hash runs inside the `consensus_session` std mutex. the comment at :121-124 picks std over tokio precisely because that section is "pure CPU"; it's now a full-batch hash blocking a tokio worker. `quic_client.rs:630` is the same shape. hash the payload before taking the lock.

`foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/ConsensusSession.java:56` - `beginRegister()` resets `requestCounter` to 1 but leaves `correlationCounter` alone. partition ops now draw from the counter that resets, so a re-login with an in-flight send can re-mint a `RequestKey` already in `pendingRequests`; `registerRequest` throws and `writeVsrFrame` turns that into `closeChannel`. that window used to be metadata-only.

`core/simulator/src/client.rs:121-144` - still models the pre-pr client: partition ids from a separate counter at `1 << 63`, metadata ids contiguous, checksum zero. the one tool that could find this interaction can't generate the pattern any more. its doc at :64 and :124-125 also still claims a gap-free requirement and a `RequestGap`, which `client_table.rs:525` says does not exist.

stale doc sites: `core/binary_protocol/src/consensus/header.rs:290` says the wire sends zero, `client_table.rs:581` and `:2825` repeat it, and node `header.ts:116`, go `header.go:119`, csharp `VsrHeader.cs:24`, java `VsrHeaders.java:29` all state the server does not read the field. six places, all false. fix header.rs first since the others quote it.

`foreign/go/internal/vsr/envelope_test.go:164` - the unbound-partition test checks only the error; its metadata twin at :156 also pins that the rejection burns no id. that invariant is live now that partition ops draw ids.

`core/server/src/dispatch.rs:1015` - stamping makes the checksum-mismatch deny path reachable for `SendMessages` for the first time. it's terminal, no retry, and nothing tests it.

and nothing anywhere sends n batches then retries a metadata op to prove the gapped sequence still dedups end to end. worth one case in `core/integration/tests/sdk/`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrRequestEncoder.java
```diff
@@ -76,12 +76,11 @@ public ByteBuf encode(ByteBufAllocator alloc, int commandCode, ByteBuf payload)
...
-                // table dedup, so use the independent correlation sequence.
-                sessionId = session.boundSession();
-                requestId = session.nextCorrelationId();
             } else {
+                // Partition ops consume the dedup counter too, even though no
+                // partition-plane dedup exists yet: dedup needs each send to
+                // carry a distinct number, and the metadata watermark
+                // tolerates the gaps.
                 sessionId = session.boundSession();
                 requestId = session.nextRequestId();
```
spetz, line 85: Partition sends now use `requestCounter`, but `beginRegister()` resets that counter to 1 while earlier sends may still be present in `pendingRequests`. A delayed pre-login `SendMessages(1)` and the first post-login `SendMessages(1)` produce the same `(operation, request)` key. `registerRequest()` then throws and closes the channel. Please drain or fail pending requests before rearming, serialize re-login behind them, or preserve a correlation generation. Add a regression test with an in-flight send across re-login.

core/consensus/src/client_table.rs
```diff
@@ -2681,6 +2681,36 @@ mod tests {
         assert_eq!(table.get_watermark(1), Some(9));
     }
 
+    // The shape a client that spends request ids off the metadata plane
+    // produces: partition-plane ids never reach this table, so the next
+    // metadata request arrives with a gap under it. It executes, moves the
+    // watermark to itself, and its retry still replays the original reply --
+    // gaps cost the skipped ids and nothing else.
+    #[test]
+    fn check_request_dedups_a_metadata_request_that_arrives_after_a_gap() {
```
spetz, line 2692: This new test manually inserts requests 1 and 6, so it only combines the existing gap-above-watermark and duplicate-at-watermark unit tests. It does not verify the behavior introduced here. Please add an integration test that sends several partition batches and then retries a metadata operation with the same request frame, proving that the gapped metadata sequence still deduplicates end to end.

spetz (changes requested):
Some extra findings

### Simulator

`core/simulator/src/client.rs` still models partition operations with a separate counter starting at `1 << 63`, and its test asserts that partition operations do not consume the shared counter. Please update the simulator to use one counter for all replicated operations so it can generate the same sequence as the SDKs after this PR.

### Server comment and dependency

`core/server/src/dispatch.rs:1602` still says the SDK does not number data-plane operations, which becomes false after this PR. The compatibility normalization can remain, but the comment should distinguish current SDKs from older or internal callers. The stable VSR client-ID rewrite is handled by PR #3959, so that dependency should also be made explicit.

### Go rejection invariant

`foreign/go/internal/vsr/envelope_test.go` verifies that an unbound metadata request does not consume an ID, but the equivalent partition test only checks the error. Since partition operations now consume the same counter, please also assert that `CurrentRequestID()` remains 1 after the rejection.

### Existing checksum documentation

Several comments still claim request checksums are always zero or are not read by the server, including `header.rs`, `client_table.rs`, and the Go, Node, C#, and Java header mirrors. That was already stale before this final diff, so it does not need to block this PR, but it should be cleaned up separately.

spetz:
LGTM after deep review.

spetz (dismissed):

hubcio (dismissed):

spetz (approved):

hubcio (approved):

PR #3959, opened 2026-08-24
feat(server): dedup partition writes with per-group client table slices

numinnex:
Implement deduplication for partition plane operations.


numinnex:
Merge after #3958

core/partitions/src/iggy_partition.rs
```diff
@@ -2136,8 +2225,9 @@ where
                 // request room -> buffer; both full -> drop+warn (client retries
                 // via read-timeout).
                 if consensus.pipeline_is_full() {
-                    let push_result =
-                        consensus.push_queued_request(consensus::RequestEntry::new(message));
+                    let push_result = consensus.push_queued_request(
+                        consensus::RequestEntry::with_sender(message, reply.take()),
```
hubcio, line 2252: the sender parked here never comes back out - `drain_request_queue_into_prepares` at 2316 projects the message and calls plain `pipeline_message`, so a write queued behind a full pipeline commits but never answers. take it with `take_reply_sender()` and use `pipeline_message_with_sender`, like the metadata plane does.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1142,118 @@ impl ClientTable {
...
+            self.evict_oldest();
+        }
+        let Some(slot_idx) = self.first_free_slot() else {
+            // Only reachable at a zero cap, which config validation rejects.
+            return;
+        };
+        self.index.insert(client_id, slot_idx);
+        self.slots[slot_idx] = Some(ClientEntry {
+            epoch: 0,
+            user_id: 0,
```
hubcio: nothing checks `user_id` here and nothing clears the entry on logout. a writer can log in on a free id, commit once at `request = u64::MAX`, log out, and the next holder of that id gets every write absorbed with status 0 - stamp the acting user and reset on mismatch.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1142,118 @@ impl ClientTable {
...
+    /// cluster-wide, so it doubles as the marker for a request whose id carries
+    /// no dedup meaning (an unbound caller, or a transport that numbers
+    /// requests without the one-in-flight-per-group discipline the watermark
+    /// assumes).
+    ///
+    /// # Panics
+    /// If called on a table that fences epochs -- that plane must go through
+    /// [`Self::check_request`], which enforces the fence.
+    #[must_use]
+    pub fn is_duplicate(&self, client_id: u128, request: u64) -> bool {
```
hubcio: keyed on client id only, no session fence. HTTP ids are sequential and reseeded from live slots only, so a new session can inherit a logged-out id and have its first writes absorbed by the old watermark - fold the register session in and reset on a higher one.

core/server/src/http/submit.rs
```diff
@@ -406,6 +406,7 @@ pub(in crate::http) async fn partition_write_replicated(
         session.session,
         session.client_id,
         Some(session.user_id),
+        PartitionReplyMode::Awaited,
```
hubcio: two concurrent writes on one session can reach the partition out of id order (routable-wait sleep, or one parked in `park_if_unmaterialised`), and the lower id is then absorbed as a duplicate with a 201. cap awaited writes at one per session, or stamp the id under a gate held until `try_send`.

core/partitions/src/iggy_partition.rs
```diff
@@ -2054,6 +2102,45 @@ where
...
+            // earned, not the typed 404 the existence check would raise now
+            // that the offset is gone.
+            //
+            // A replay racing its own in-flight original is absorbed here: the
+            // slice only knows committed ops, so it cannot yet see the copy
+            // still in the pipeline. Keyed on the exact `(client, request)` --
+            // matching any request from the client would serialize its
+            // pipeline depth to one in-flight write per group.
+            if !is_auto_commit_client(client_id) {
+                if consensus.pipeline_has_message_from_client_request(client_id, request) {
```
hubcio, line 2130: this assumes a client's writes arrive in id order, but the java async SDK pipelines and replays a denied frame under the same request id - a replay landing after a later write committed is absorbed. needs a reorder-tolerant window below the watermark, or lockstep on the SDK side.

core/server/src/dispatch.rs
```diff
@@ -1590,17 +1605,72 @@ pub async fn dispatch_partition_request<B, MJ, S, SB>(
...
+        // Current SDKs number partition ops; a caller that does not sends
+        // every op under request 1 and only its first is executed.
         new_header.request = new_header.request.max(1);
     });
-    shard.dispatch(request.into_generic());
+    if reply_mode == PartitionReplyMode::FireAndForget {
+        shard.dispatch(request.into_generic());
+        return;
+    }
+    relay_partition_reply(
```
hubcio: awaiting the commit here inside the per-client drain loop means one partition write in flight per connection, and every queued poll or metadata op waits behind it. spawn only the reply wait - keep the `try_send` on the loop or writes reorder.

core/server/src/http/submit.rs
```diff
@@ -406,6 +406,7 @@ pub(in crate::http) async fn partition_write_replicated(
...
     )
     .await;
     let outcome = compio::time::timeout(PARTITION_WRITE_REPLY_TIMEOUT, receiver).await;
```
hubcio, line 417: this 10 s timer only starts after the 30 s `partition_submit` wait inside `dispatch_partition_request`, so a handler can hold its admission slot for 40 s, not 10. race the two waits or bound the dispatch await by `PARTITION_WRITE_REPLY_TIMEOUT`.

core/partitions/src/iggy_partition.rs
```diff
@@ -3346,13 +3464,19 @@ where
...
+                // consensus id and carries no home-shard routing. The awaiting
+                // shard owns the socket. A dropped receiver is ignored -- the
+                // client recovers on its own read-timeout.
+                if let Some(sender) = entry.take_reply_sender() {
+                    let _ = sender.send(reply);
+                } else if let Err(error) = self
                     .consensus
                     .message_bus()
-                    .send_to_client(prepare_header.client, reply_buffers)
+                    .send_to_client(prepare_header.client, reply.into_generic().into_frozen())
```
hubcio, line 3532: `prepare_header.client` is now the unroutable vsr id, and `rebuild_pipeline_entries` creates senderless entries on every view change, so this fires per re-committed op and logs an error instead of delivering. keep the transport id on the prepare, or drop this to debug - do not assert.

core/partitions/src/state_transfer.rs
```diff
@@ -50,15 +50,17 @@ use std::path::{Path, PathBuf};
 use std::rc::Rc;
 use std::sync::atomic::Ordering;
 
-/// Framing marker for the consumer-offsets wire artifact, "ICO1".
-pub(crate) const CONSUMER_OFFSETS_MAGIC: [u8; 4] = *b"ICO1";
+/// Framing marker for the consumer-offsets wire artifact, "ICO2". Bumped with
+/// the version when the dedup section was appended; the magic moves too so a
+/// v1 artifact fails on the cheaper check.
+pub(crate) const CONSUMER_OFFSETS_MAGIC: [u8; 4] = *b"ICO2";
```
hubcio, line 58: no v1 decode path, so in a rolling upgrade a rejoining replica behind the repair floor re-pulls forever against an un-upgraded peer. decode v1 with an empty dedup section, or call it out in the release notes.

core/configs/src/server_config/partition.rs
```diff
@@ -114,6 +114,14 @@ pub const DEFAULT_EVICTED_RING_BYTES_MAX: u64 = 16 * 1024 * 1024;
 pub const MAX_EVICTED_RING_BYTES: u64 = 256 * 1024 * 1024;
 
 /// Capacity tunables for the per-partition consensus plane.
+/// Shipped default for [`PartitionConfig::dedup_clients_max`]; pinned against
+/// the runtime constant by a bootstrap assert.
+pub const DEFAULT_PARTITION_DEDUP_CLIENTS_MAX: usize = 4096;
+
+/// Ceiling for [`PartitionConfig::dedup_clients_max`]. A per-group budget, so
+/// the ceiling bounds worst-case memory at `partitions * this * ~40 bytes`.
```
hubcio: `ClientEntry` is 96 bytes before the index map, so the real per-entry cost is about 4x this - the `~384 KiB` comment in `client_table.rs` already assumes 96. fix the number or slim the slice entry.

core/configs/src/server_config/partition.rs
```diff
@@ -114,6 +114,14 @@ pub const DEFAULT_EVICTED_RING_BYTES_MAX: u64 = 16 * 1024 * 1024;
 pub const MAX_EVICTED_RING_BYTES: u64 = 256 * 1024 * 1024;
 
 /// Capacity tunables for the per-partition consensus plane.
```
hubcio: this line was `PartitionConfig`'s doc - the consts inserted below it leave the struct undocumented and give `DEFAULT_PARTITION_DEDUP_CLIENTS_MAX` two unrelated sentences. move it back above the `#[derive]`.

core/configs/src/server_config/partition.rs
```diff
@@ -114,6 +114,14 @@ pub const DEFAULT_EVICTED_RING_BYTES_MAX: u64 = 16 * 1024 * 1024;
...
+/// Shipped default for [`PartitionConfig::dedup_clients_max`]; pinned against
+/// the runtime constant by a bootstrap assert.
+pub const DEFAULT_PARTITION_DEDUP_CLIENTS_MAX: usize = 4096;
+
+/// Ceiling for [`PartitionConfig::dedup_clients_max`]. A per-group budget, so
+/// the ceiling bounds worst-case memory at `partitions * this * ~40 bytes`.
+pub const MAX_PARTITION_DEDUP_CLIENTS: usize = 1 << 16;
```
hubcio: `MAX_PARTITION_DEDUP_CLIENTS`, `DEFAULT_PARTITION_DEDUP_CLIENTS_MAX` and `PARTITION_DEDUP_CLIENTS_MAX` in `consensus` are three near-identical names for two things. `_DEFAULT` and `_CEILING` would read unambiguously.

core/consensus/src/client_table.rs
```diff
@@ -1310,8 +1511,18 @@ impl ClientTable {
...
     }
 
-    fn first_free_slot(&self) -> Option<usize> {
-        self.slots.iter().position(Option::is_none)
+    /// Lowest free slot, growing the array when slots are allocated lazily.
+    /// Assignment is identical to the preallocated case: both hand out the
+    /// lowest free index, so eviction order and the wire encoding do not
+    /// depend on the mode.
+    fn first_free_slot(&mut self) -> Option<usize> {
+        if let Some(index) = self.slots.iter().position(Option::is_none) {
```
hubcio: below cap the slice has no holes, so this scan is a guaranteed full miss before every push - quadratic fill on the commit path and in `install_watermarks`. scan only when `index.len() < slots.len()`, and keep the `clients_max` bound on the push.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1142,118 @@ impl ClientTable {
...
+                entry.watermark = request;
+                entry.latest_commit = commit_op;
+            }
+            return;
+        }
+
+        if self.index.len() >= self.clients_max {
+            self.evict_oldest();
+        }
+        let Some(slot_idx) = self.first_free_slot() else {
```
hubcio: `evict_oldest` returns the freed slot but it is discarded and `first_free_slot` rescans - two full walks per new client at cap. do what `commit_register` does: `freed.or_else(|| self.first_free_slot())`.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1142,118 @@ impl ClientTable {
...
+            latest_commit: commit_op,
+        });
+    }
+
+    /// Replace every entry, as a state-transfer install does.
+    ///
+    /// # Panics
+    /// If called on a table whose mode caches replies (those install through
+    /// the snapshot / wire codecs, which carry the rings).
+    pub fn install_watermarks(&mut self, entries: impl IntoIterator<Item = (u128, u64, u64)>) {
```
hubcio: each entry past cap costs two full scans on the pump, and the survivors are whatever came last in wire order, not the newest. sort by `latest_commit`, truncate to `clients_max`, then push straight in - keep the decoder ceiling, caps differ per node.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1142,118 @@ impl ClientTable {
...
+            user_id: 0,
+            watermark: request,
+            watermark_checksum: 0,
+            ring: VecDeque::new(),
+            client_id,
+            latest_commit: commit_op,
+        });
+    }
+
+    /// Replace every entry, as a state-transfer install does.
```
hubcio: `install_watermarks(std::iter::empty())` is the clear in two places - say so here so readers do not hunt for a `clear`.

core/partitions/src/iggy_partition.rs
```diff
@@ -86,16 +86,21 @@ use tokio::sync::Mutex as TokioMutex;
...
-// Duplicate suppression is a consensus-layer concern: the VSR client table
-// dedups by request id (at-most-once), so the data plane needs no message-id set.
 pub struct IggyPartition<B = IggyMessageBus, SB = PingPongSuperblock>
 where
     B: MessageBus,
 {
     consensus: VsrConsensus<B>,
+    /// This group's slice of the VSR client table, run in
+    /// [`ClientTableMode::PARTITION_SLICE`]: per-client request watermarks
+    /// folded in at commit, so every replica derives the same slice from the
```
hubcio: not true after a restart - the slice is memory-only and boot lifts the commit frontier without re-applying, so a restarted replica comes back empty while its peers keep theirs. say replica-local, rebuilt only by commits after boot or state transfer.

core/integration/tests/cluster/partition_dedup.rs
```diff
@@ -0,0 +1,660 @@
...
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Spec tests for partition-plane request dedup (IGGY-274).
+//!
+//! Each partition consensus group keeps a slice of the VSR client table:
+//! per-client request watermarks folded in at commit. A replay of an
+//! already-committed `(client, request)` is answered with the empty success its
```
hubcio, line 22: nothing here covers a write promoted from the request queue while holding a reply sender - every test is lockstep on one connection. add one with 33+ concurrent connections to one namespace and assert every reply arrives.

core/integration/tests/cluster/partition_dedup.rs
```diff
@@ -0,0 +1,660 @@
...
+    stream.write_all(bytemuck::bytes_of(header)).await.unwrap();
+    if !body.is_empty() {
+        stream.write_all(body).await.unwrap();
+    }
+
+    let mut reply_header = [0u8; HEADER_SIZE];
+    timeout(REPLY_WAIT, stream.read_exact(&mut reply_header))
+        .await
+        .expect("reply header timed out")
+        .expect("reply header read failed");
```
hubcio: `exchange` panics on a server-closed socket, so `send_reconnecting` only survives an explicit eviction frame - phase 3 stops nodes around this connection, so EOF here is a flake. return a disconnected sentinel on io error and treat it like `EVICTED`.

core/shard/src/lib.rs
```diff
@@ -1872,6 +1889,69 @@ where
...
+            request,
+            reply: reply_tx,
+        });
+        let sender = self.senders.get(target as usize)?;
+        if let Err(error) = sender.try_send(frame) {
+            self.metrics.record_frame_drop(
+                crate::metrics::frame_drop_variant::PARTITION,
+                crate::coordinator::classify_try_send_err(&error),
+            );
+            tracing::warn!(
```
hubcio, line 1957: a refused `try_send` provably never reached the plane, yet it returns the same `None` as an unknown outcome so the client eats a full read timeout. return a typed `Refused` and deny with `TransientNotAccepted`.

core/shard/src/lib.rs
```diff
@@ -1872,6 +1889,69 @@ where
...
+            Some(Err(_)) => {
+                tracing::debug!(
+                    shard = self.id,
+                    target,
+                    "partition_submit: reply sender dropped before commit"
+                );
+                None
+            }
+            None => {
+                tracing::warn!(
```
hubcio: neither silent exit bumps `frame_drops_total`, and this warns per request so one stalled group floods the log. bump the partition drop counter with distinct reasons and gate the warn like `first_shed` does.

core/partitions/src/iggy_partition.rs
```diff
@@ -2136,8 +2225,9 @@ where
                 // request room -> buffer; both full -> drop+warn (client retries
                 // via read-timeout).
                 if consensus.pipeline_is_full() {
-                    let push_result =
-                        consensus.push_queued_request(consensus::RequestEntry::new(message));
+                    let push_result = consensus.push_queued_request(
+                        consensus::RequestEntry::with_sender(message, reply.take()),
+                    );
                     if push_result.is_err() {
```
hubcio: the entry comes back in the `Err` with the sender still inside and is dropped - the client waits out a timeout for a frame that never entered the pipeline. take the sender back and answer `TransientNotAccepted`.

core/partitions/src/iggy_partitions.rs
```diff
@@ -586,7 +591,20 @@ where
             );
             return;
```
hubcio, line 614: also at lines 552 and 578. these early returns drop `reply` unanswered, so an encrypt failure costs the client a read timeout instead of an error - answer on `reply` before each `return`.

core/simulator/src/lib.rs
```diff
@@ -2617,22 +2501,29 @@ mod tests {
...
         let client = SimClient::new(CLIENT_ID);
         sim.shell_login(&client);
-        for payload in [
+        let lagging_shard = Rc::clone(&sim.replicas[1].shards[0]);
+        let mut successful_replies = 0usize;
+        let mut parked = 0usize;
+        // One send in flight at a time. The partition dedup slice keys on a
+        // per-client watermark, so if the network reordered two concurrent sends
+        // the lower request id would arrive after the higher one committed and
+        // read as already applied. Every real transport is lockstep per
```
hubcio: this rewrite deletes the only test of two concurrent same-client partition writes, and the comment explains exactly the bug it hides. keep the concurrent form asserting both payloads commit, and let it fail until ordering is enforced.

core/shard/src/lib.rs
```diff
@@ -3413,6 +3568,103 @@ where
...
+                    .on_request_with_reply(request, Some(sender))
+                    .await;
+                // Await OFF the pump: the commit that fires this receiver needs
+                // the pump to keep draining acks, so blocking here would
+                // deadlock the very reply being waited on. The task holds only
+                // owned channel halves, never a partitions borrow.
+                let Some(reply) = parked_reply else { return };
+                // Through the bus, not the runtime directly: the simulator
+                // supplies its own executor and virtual clock.
+                self.bus.spawn(async move {
```
hubcio, line 3716: each acked write now allocates a channel, a oneshot, a spawned task and a timer where the old path was fire-and-forget. fine for now - if it shows in profiles, carry the sender in `PipelineEntry` and drop the oneshot and spawn.

core/shard/src/lib.rs
```diff
@@ -3413,6 +3568,103 @@ where
...
+                let (sender, receiver) = consensus::oneshot_channel();
+                self.plane
+                    .partitions()
+                    .on_request_with_reply(request, Some(sender))
+                    .await;
+                // Await OFF the pump: the commit that fires this receiver needs
+                // the pump to keep draining acks, so blocking here would
+                // deadlock the very reply being waited on. The task holds only
+                // owned channel halves, never a partitions borrow.
+                let Some(reply) = parked_reply else { return };
```
hubcio: `Deliver` always holds the reply, since `park_if_unmaterialised` only takes it on `Parked` - this guard silently strands a request if that ever breaks. bind the sender before the match and pass `&mut Some(sender)`.

core/partitions/src/iggy_partition.rs
```diff
@@ -3571,13 +3695,51 @@ where
     /// body, op=0), logging a WARN under `send_fail_label` if the reply send
     /// fails. Callers deny on the primary, before the op enters the pipeline,
     /// so nothing replicates.
+    /// Answer an absorbed duplicate with the reply its original earned. Same
```
hubcio: the four doc lines above belong to `send_partition_deny_or_log`, not this fn - move them back above line 3728.

core/partitions/src/iggy_partition.rs
```diff
@@ -3571,13 +3695,51 @@ where
...
+    /// delivery split as [`Self::send_partition_deny_or_log`]: the submit's
+    /// channel when one is waiting, the bus otherwise.
+    async fn answer_duplicate_or_log(
+        consensus: &VsrConsensus<B>,
+        header: &RoutedRequestHeader,
+        reply: Message<ReplyHeader>,
+        waiter: Option<consensus::Sender<Message<ReplyHeader>>>,
+    ) {
+        if let Some(waiter) = waiter {
```
hubcio: body is line-for-line `send_partition_deny_or_log` except the diag label. one `deliver_reply_or_log(consensus, header, reply, waiter, label)` and have the deny path build its reply first - about 25 lines gone.

core/partitions/src/iggy_partition.rs
```diff
@@ -1951,16 +1987,27 @@ where
...
+    /// A replay of a committed `(client, request)` is absorbed by this group's
+    /// dedup slice; anything above the watermark projects into a prepare.
+    /// Session lifecycle + eviction live on the metadata plane.
     ///
     /// # Panics
     /// Panics if called when this partition's consensus instance is not the
     /// primary, is not in normal status, or is currently syncing.
     #[allow(clippy::future_not_send, clippy::too_many_lines)]
-    pub async fn on_request(&mut self, message: Message<RoutedRequestHeader>) {
+    /// `reply` is the in-process channel a `PartitionSubmit` carried in. When
```
hubcio, line 1998: this `reply` paragraph sits after the `#[allow]`, so rustdoc renders it under `# Panics`. move it above.

core/partitions/src/iggy_partition.rs
```diff
@@ -555,6 +564,25 @@ where
...
+        &self.dedup
+    }
+
+    /// Mutable slice, for the commit path and state-transfer install.
+    pub(crate) const fn dedup_mut(&mut self) -> &mut ClientTable {
+        &mut self.dedup
+    }
+
+    /// Size the dedup slice to `[partition] dedup_clients_max`. Boot-time:
+    /// shrinking a live slice evicts its oldest-committed entries, which costs
```
hubcio: `set_capacity` asserts the table is empty and replaces it - nothing evicts. say boot-only, panics on a populated slice.

core/partitions/src/iggy_partition.rs
```diff
@@ -555,6 +564,25 @@ where
         &self.consensus
     }
 
+    /// This group's dedup slice. Read at admission to classify a request,
+    /// written only from the commit path.
+    #[must_use]
+    pub const fn dedup(&self) -> &ClientTable {
```
hubcio: only caller is `state_transfer.rs` in this crate - `pub(crate)`, like `dedup_mut`.

core/partitions/src/iggy_partition.rs
```diff
@@ -555,6 +564,25 @@ where
...
+    /// Mutable slice, for the commit path and state-transfer install.
+    pub(crate) const fn dedup_mut(&mut self) -> &mut ClientTable {
+        &mut self.dedup
+    }
+
+    /// Size the dedup slice to `[partition] dedup_clients_max`. Boot-time:
+    /// shrinking a live slice evicts its oldest-committed entries, which costs
+    /// dedup coverage for those clients rather than correctness.
+    pub fn set_dedup_clients_max(&mut self, clients_max: usize) {
+        self.dedup.set_capacity(clients_max.max(1));
```
hubcio: config validation already rejects 0, so this `.max(1)` only hides a value that cannot arrive. pass `clients_max` through.

core/partitions/src/iggy_partition.rs
```diff
@@ -2054,6 +2102,45 @@ where
...
             };
 
+            // Dedup BEFORE the admission checks below: a replay of an
+            // already-committed delete must answer the success its original
+            // earned, not the typed 404 the existence check would raise now
+            // that the offset is gone.
+            //
+            // A replay racing its own in-flight original is absorbed here: the
+            // slice only knows committed ops, so it cannot yet see the copy
+            // still in the pipeline. Keyed on the exact `(client, request)` --
```
hubcio: the pipeline-depth rationale only holds for HTTP - tcp clients are already serialized per connection by `drain_client_requests`. say which plane this protects.

core/partitions/src/iggy_partition.rs
```diff
@@ -2054,6 +2102,45 @@ where
...
+                        reply.take(),
+                    )
+                    .await;
+                    return;
+                }
+                if self.dedup.is_duplicate(client_id, request) {
+                    let committed = build_reply_from_request(
+                        &self.consensus,
+                        message.header(),
+                        committed_reply_body(message.header().operation),
```
hubcio, line 2153: `committed_reply_body` is empty for `SendMessages`, but the original reply carried offset confirmations. either keep the committed body on the entry or document that a deduped produce returns none.

core/partitions/src/state_transfer.rs
```diff
@@ -295,22 +300,29 @@ pub(crate) struct ConsumerOffsetsWire {
     pub consumers: Vec<(u32, u64)>,
     /// `(consumer group id, offset)`, ascending by id.
     pub groups: Vec<(u32, u64)>,
+    /// This group's dedup slice: `(client, watermark, latest_commit)`,
+    /// ascending by client. Carried so a replica rejoining behind the repair
+    /// floor can absorb a replay of what the group already committed instead
+    /// of re-executing it.
+    pub dedup: Vec<(u128, u64, u64)>,
```
hubcio: `watermark` and `latest_commit` are both `u64` and get destructured positionally into `commit_request` - easy to swap silently. a small `DedupWatermark { client, watermark, latest_commit }` struct here and in `client_table` removes that.

core/partitions/src/state_transfer.rs
```diff
@@ -377,9 +398,42 @@ impl ConsumerOffsetsWire {
...
+        if count as usize * DEDUP_ENTRY_LEN > cursor.remaining().len() {
+            return Err(ConsumerOffsetsWireError::Truncated);
+        }
+        let mut entries = Vec::with_capacity(count as usize);
+        let mut previous: Option<u128> = None;
+        for _ in 0..count {
+            let client = cursor.u128()?;
+            let watermark = cursor.u64()?;
+            let latest_commit = cursor.u64()?;
+            if previous.is_some_and(|previous| client <= previous) {
```
hubcio, line 461: a `client == 0` entry passes here and is then silently dropped by `commit_request`. it is a peer artifact, so fail closed and reject it like `NonAscendingClient`.

core/partitions/src/state_transfer.rs
```diff
@@ -584,6 +650,69 @@ mod tests {
         );
     }
 
+    #[test]
+    fn given_dedup_entries_when_encoded_should_round_trip() {
```
hubcio: `table()` already carries dedup entries, so this asserts exactly what `given_offset_table_when_encoded_should_round_trip` asserts. delete it.

core/server/src/dispatch.rs
```diff
@@ -1590,17 +1605,72 @@ pub async fn dispatch_partition_request<B, MJ, S, SB>(
...
-        // (at-least-once, no `ClientTable` dedup), so the bound VSR
-        // session merely satisfies validation. Current SDKs do number
-        // partition ops, but older and internal callers may still send
-        // zero, so a zero id is normalized to the compatibility value 1.
+        // The VSR consensus id, exactly as metadata ops are stamped: it is the
+        // dedup key every replica keys its slice by, and unlike the transport
+        // id it stays valid across nodes. Replies therefore cannot be routed
+        // by this field -- they ride the submit's channel back to this shard,
+        // which owns the socket.
+        new_header.client = vsr_client_id;
```
hubcio: the transmute never stamps `user_id`, so the client-supplied value rides into the replicated log - the metadata plane overwrites it from the session for exactly that reason. set `new_header.user_id = acting_user_id.unwrap_or(0)` here, the dedup fix has to trust it.

core/server/src/dispatch.rs
```diff
@@ -1590,17 +1605,72 @@ pub async fn dispatch_partition_request<B, MJ, S, SB>(
...
+        // Header validation requires `session > 0` for non-register ops. The
+        // slices mint no epoch of their own, so the bound VSR session only
+        // satisfies validation here.
         new_header.session = bound_session;
+        // Header validation requires `request > 0` for non-register ops, and
+        // the routed header cannot carry the exempt client id (validation
+        // refuses client 0), so an unnumbered data-plane op normalizes to 1.
+        // Current SDKs number partition ops; a caller that does not sends
+        // every op under request 1 and only its first is executed.
         new_header.request = new_header.request.max(1);
```
hubcio: every caller already guarantees `request > 0` and `transmute_header` re-validates below, so this never fires and the paragraph documents write loss as intended behaviour. drop both.

core/server/src/dispatch.rs
```diff
@@ -1446,13 +1447,26 @@ async fn handle_get_me<B, MJ, S, SB>(
...
 
+/// Whether a partition write waits for its committed reply.
+///
+/// `Awaited` attaches an in-process channel to the request, so the committed
+/// reply comes back to this shard instead of the bus and is written to the
+/// caller's socket here. `FireAndForget` attaches none: the reply takes the bus
+/// path and is shed when nothing is listening, which is what `?ack=none` asks
+/// for.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PartitionReplyMode {
```
hubcio: for HTTP the vsr id is the transport id, so the commit-path `send_to_client` already fires its slot - `Awaited` just adds a second hop and a second timer. let HTTP use `FireAndForget` and assert that id invariant, then this enum collapses.

core/consensus/src/client_table.rs
```diff
@@ -587,18 +637,41 @@ const fn checksums_conflict(stored: u128, received: u128) -> bool {
...
+            index,
+            clients_max: max_clients,
+            mode,
             evicted_fences: VecDeque::new(),
         }
     }
 
+    /// The mechanisms this instance runs.
+    #[must_use]
+    pub const fn mode(&self) -> ClientTableMode {
```
hubcio: no callers anywhere - delete.

core/consensus/src/client_table.rs
```diff
@@ -508,6 +508,45 @@ pub enum CommitReply {
...
 
+/// Which of the table's mechanisms an instance runs.
+///
+/// The metadata plane needs all of them. A partition group's slice needs only
+/// the watermark: it has no register to mint an epoch from, no result section
+/// worth caching, and one table per group rather than per node, so the
+/// preallocated slot array would reserve ~384 KiB per partition before a single
+/// client connects.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ClientTableMode {
```
hubcio: three pub bools admit eight states and only two exist, with runtime asserts catching the rest one bool at a time. `enum ClientTableMode { Metadata, PartitionSlice }` plus `const fn` predicates - same line count, no invalid states.

core/consensus/src/impls.rs
```diff
@@ -313,6 +321,22 @@ impl RequestEntry {
         (entry, receiver)
     }
 
+    /// Queued request carrying a sender the caller already owns, for a submit
+    /// that parked before reaching a prepare slot. Mirrors
+    /// [`Self::with_subscriber`], except the receiver half lives with the
+    /// caller rather than being minted here.
+    #[must_use]
+    pub const fn with_sender(
```
hubcio, line 320: `new` and `with_subscriber` are this fn with `None` and with a fresh channel - route both through here.

core/consensus/src/client_table.rs
```diff
@@ -587,18 +637,41 @@ const fn checksums_conflict(stored: u128, received: u128) -> bool {
 }
 
 impl ClientTable {
+    /// Client id that opts out of dedup entirely. Zero is reserved cluster-wide
+    /// and every mutating entry point refuses it.
+    pub const EXEMPT_CLIENT: u128 = 0;
```
hubcio: unreachable from every ingress (wire refuses 0, HTTP mint skips 0, auto-commit is guarded at the call sites), so this const and the `is_duplicate` early return are dead. keep the `commit_request` return until the decoder rejects 0, or a peer artifact panics the next admission.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1142,118 @@ impl ClientTable {
...
+    /// no dedup meaning (an unbound caller, or a transport that numbers
+    /// requests without the one-in-flight-per-group discipline the watermark
+    /// assumes).
+    ///
+    /// # Panics
+    /// If called on a table that fences epochs -- that plane must go through
+    /// [`Self::check_request`], which enforces the fence.
+    #[must_use]
+    pub fn is_duplicate(&self, client_id: u128, request: u64) -> bool {
+        assert!(
```
hubcio: also at lines 1190 and 1232. mode is fixed at construction, so a hard `assert!` in a library crate buys nothing over `debug_assert!`.

hubcio (changes requested):
items on lines outside the diff:

- no shipped SDK preserves its client id across a reconnect and HTTP mints a fresh request id per call, so nothing reaches the dedup today - the PR body should say so and link the SDK follow-up.
- `purge` in `iggy_partition.rs` clears offsets, caches and consumer offsets but not the dedup slice, while `converge_to_empty_after_failed_install` clears it for the same reason. clear it in `purge` too.
- the `decode_section` byte-guard comment in `state_transfer.rs` is stale now that the dedup section follows: an over-stated group count fails as `NonAscendingId` or `TrailingBytes` instead of `Truncated`. fix the comment or subtract the trailing sections' minimum.

core/consensus/src/client_table.rs
numinnex: Left as is: this contradicts the rebind contract the slice inherits from the metadata plane. `commit_register` keeps the watermark when an existing key re-registers (only the epoch and user move), and the transferred-slice spec in this PR replays after a re-register and must be absorbed. Resetting on a higher session would re-execute exactly the reconnect replay the dedup exists for.

The cross-user half of the hazard (a re-minted HTTP id landing on another user's watermark) is closed in this revision: the partition prepare now carries the acting `user_id`, the slice keys on it, and a commit by a different user resets the entry. The same-user re-mint of a logged-out id needs the slice to observe Register/Logout, which is the clients-table follow-up; a session compare cannot do it without breaking rebind.

core/partitions/src/iggy_partition.rs
numinnex: Took the window route. Each slice entry now carries a 128-bit committed-id window under the watermark (`consensus::COMMITTED_WINDOW_BITS`): a commit above the watermark shifts it by the gap, a commit below sets its bit, and `is_duplicate` treats an unmarked id inside the window as new. A transiently refused write replayed after its successors committed therefore executes instead of being absorbed.

Sized above the default per-group in-flight ceiling (32 prepares + 64 queued); an id older than the window still reads as committed, which is the rule that applied everywhere before. The window rides the transfer wire as `DedupWatermark::committed_window`, and the concurrent-send simulator test is restored on top of it.

core/simulator/src/lib.rs
numinnex: Restored the concurrent form: both sends are submitted before stepping and both payloads plus both parks are asserted, as before. It passes because the slice now keeps a committed-id window below the watermark, so the lower id arriving after the higher one committed is admitted rather than absorbed. The test comment states that dependency.

core/shard/src/lib.rs
numinnex: Agreed, leaving it. `PipelineEntry` already carries the sender; the oneshot plus spawn exists only because the cross-shard submit channel is a `Send` mpsc while the pipeline sender is the single-threaded oneshot. Collapsing them means one channel type for both paths, worth doing once it shows in a profile.

core/consensus/src/client_table.rs
```diff
@@ -508,6 +515,82 @@ pub enum CommitReply {
...
+///
+/// A client that pipelines writes can see one of them refused transiently and
+/// replay it after later ids have committed, so "at or below the watermark"
+/// alone would absorb that replay as a duplicate and lose the write. The window
+/// records which ids under the watermark actually committed; an unmarked one
+/// inside it executes. Sized above the default in-flight ceiling per group
+/// (`PIPELINE_PREPARE_QUEUE_MAX + PIPELINE_REQUEST_QUEUE_MAX` = 96): a client
+/// with more than this many writes outstanding on one partition can still have
+/// a replay absorbed once its id ages out of the window.
+pub const COMMITTED_WINDOW_BITS: u64 = 128;
```
hubcio, line 599: sized against the per-group in-flight ceiling of 96, but the window lives in the client's request-id space and `ConsensusSession` runs one counter across every partition, stream and metadata op. over 8 partitions that leaves ~16 usable ids each, so a replay past the window is absorbed with status 0 and the write is lost.

core/consensus/src/client_table.rs
```diff
@@ -1059,6 +1180,176 @@ impl ClientTable {
...
+        };
+        self.index.insert(client_id, slot_idx);
+        self.slots[slot_idx] = Some(ClientEntry::watermark_only(
+            client_id, user_id, request, 1, commit_op,
+        ));
+    }
+
+    /// Replace every entry, as a state-transfer install does. An empty iterator
+    /// is the clear: there is no separate `clear`, and both callers that need
+    /// one (a failed install converging to empty, a purge) come through here.
```
hubcio: purge isn't a caller - `IggyPartition::purge` never touches `self.dedup`. either wire it up or drop it from this list.

core/configs/src/server_config/partition.rs
```diff
@@ -113,6 +113,15 @@ pub const DEFAULT_EVICTED_RING_BYTES_MAX: u64 = 16 * 1024 * 1024;
 /// trips first evicts; this byte ceiling is the second typo guard.
 pub const MAX_EVICTED_RING_BYTES: u64 = 256 * 1024 * 1024;
 
+/// Shipped default for [`PartitionConfig::dedup_clients_max`]; pinned against
+/// the runtime constant by a bootstrap assert.
+pub const PARTITION_DEDUP_CLIENTS_DEFAULT: usize = 4096;
+
+/// Ceiling for [`PartitionConfig::dedup_clients_max`]. A per-group budget, so
+/// the ceiling bounds worst-case memory at roughly `partitions * this * 130
+/// bytes`: a 96-byte slot entry plus its index-map slot.
```
hubcio: `ClientEntry` is 112 bytes now that it carries the u128 window, not 96, so the ceiling works out closer to 146 bytes an entry. update both numbers.

core/partitions/src/state_transfer.rs
```diff
@@ -377,9 +423,56 @@ impl ConsumerOffsetsWire {
...
+                user_id,
+                watermark,
+                latest_commit,
+                committed_window,
+            });
+        }
+        Ok(entries)
+    }
+
     fn decode_section(
```
hubcio, line 476: the comment at line 494 says the groups call sees exactly `12 * count` bytes remaining, but the dedup section now follows groups so there's always more. the guard still holds, the reasoning is stale.

hubcio (changes requested):
round 2 only fixed the doctest link, so most of round 1 still stands. two findings have no line in the diff to hang on:

`ShardZeroCoordinator::seed_client_sequence` in `core/shard/src/coordinator.rs` only ever raises the counter, and it seeds from live metadata ids at boot - a logged-out id is already gone from the table by then. so after a restart the same user can be handed the same client id while a peer primary still holds that id's slice watermark, and the first writes come back 201 without being written. minting ids that never repeat (boot nonce or view in the upper bits) closes it without touching the slice.

the PR body is still one line. worth saying which SDKs this reaches and what a client sees when a write is refused.

separate from this PR: the `scripts/ci/storage-compat.sh` fix in 85aa01e4c is ok.

hubcio (approved):

spetz (approved):

PR #3964, opened 2026-08-25
fix(sdk): stream should terminate after consumer shutdown

### LINKED ISSUE #3910: bug(Rust SDK): IggyConsumer::next() after shutdown() hangs or silently drops offsets
### Bug description

**Description**
When calling shutdown() on the IggyConsumer polling is still possible with next() but causes unintended behavior for:
1.  standalone consumer next() still consumes messages from buffer or server, while sending offsets is prevented due to dropping the channel in shutdown() 
2. a consumer group member after shutdown can still call next but hangs indefinitely: https://github.com/apache/iggy/blob/6043d194000cff5f5498a91dc2259e34762f7d1c/core/sdk/src/clients/consumer.rs#L695

Calling init() again is idempotent, so a user cannot recover from this state.

**Options**
- make poll_next() check shutdown() and return 
 - `None` to respect the Stream trait
 - `IggyError` that can be handled by the user (preferred)

### Affected area / component

Rust SDK

### Deployment

None

### Versions

_No response_

### Hardware / environment

_No response_

### Sample code

_No response_

### Logs

_No response_

### Iggy server config

_No response_

### Reproduction

_No response_

### Contribution

### Good first issue

haubur:
Closes #3910

## Rationale
After `consumer.shutdown()` calling consumer.next() was still possible. Causing
- for a _standalone consumer_ to poll fresh from the server, while storing offsets was unreachable due to dropping the channel on shutdown()
- for a consumer group to hang indefinitely while trying to build a new poll future

## What changed?
When writing the issue I said that the preferred way would be to return an IggyError to allow user handling.
However, calling next() on a shut down consumer now returns: `Poll::Ready(None)` indicating a terminated stream as intended by the futures crate: https://docs.rs/futures/latest/futures/prelude/trait.Stream.html#:~:text=Poll%3A%3AReady%28None%29%20means%20that%20the%20stream%20has%20terminated%2C%20and%20poll%5Fnext%20should%20not%20be%20invoked%20again

## Local Execution
- Passed
- Pre-commit hooks ran


core/sdk/src/clients/consumer.rs
```diff
@@ -1239,3 +1243,97 @@ impl Drop for IggyConsumer {
...
+        ]);
+        let mut context = Context::from_waker(Waker::noop());
+
+        assert!(matches!(
+            Pin::new(&mut consumer).poll_next(&mut context),
+            Poll::Ready(Some(Ok(_)))
+        ));
+
+        consumer.shutdown().await.unwrap();
+
```
ethanlin01x, line 1387: the standalone test asserts remaining buffered messages are preserved (`buffered_messages.len() == 1`) but this one doesn't — worth aligning for consistency.

ethanlin01x (approved):
LGTM

spetz (approved):

hubcio (approved):

PR #3970, opened 2026-08-26
feat(partitions): persist log and index concurrently, harden recovery

hubcio:
## Concurrent log and index fsync

Every journal flush issued two serialized fdatasyncs, log then index, so an fsync-gated topic paid two device round trips per ack. Both files now persist under one `futures::future::join`: 13.4k to 26.1k msg/s and p50 5.83 to 3.01 ms on ext4 with `enforce_fsync` and `messages_required_to_save = 1`, 8 producers into one partition. Cursors advance only after both saves succeed, so a failed half keeps its slot and the callers that survive rewrite the same positions rather than appending a duplicate.

## Recovery had to change first

The old ordering was a barrier only under `enforce_fsync = true`. With the shipped default the page cache already wrote the two inodes back in any order, and recovery's own pass C truncates the log before the index, so an index holding entries the log does not back was already ordinary. Boot refused it as `IndexLogDivergence`, which on a single-replica node permanently tombstoned a partition whose log was healthy.

An index entry is derived from the log, so it is never evidence about the log. Recovery now drops such an index whole and rebuilds it from a byte-0 walk rather than stepping back to the highest entry the log still proves. Stepping back is unsound: that anchor can sit above interior damage, and the residue probe only scans forward from where a walk stopped, so a hole under it goes unread while `end_offset` still advertises the offsets over it. The byte-0 walk reads the damage and refuses with every byte preserved, and checksums every batch it accepts, since unordered files mean a durable index entry no longer implies a durable chunk.

Under `enforce_fsync = true` the step-back depth is measured first, because there it is evidence about the log. Flushes are serialized and each fdatasyncs the whole log file, so an entry above entry N proves the log was synced through chunk N, and only the chunk in flight at the crash can strand one. A deeper gap refuses as the new `FsyncedLogLoss` rather than rebuilding, which would let the partition re-mint offsets a client was already promised. The probe measuring that depth is capped, since past the cap the verdict is refusal either way.

That allowance is **not** a promise that nothing acked is lost. The partition journal is in memory and never replayed at boot, and `enforce_fsync` decides whether a flush fdatasyncs, never whether one happens, so at the default `messages_required_to_save = 1024` up to 1023 acked messages sit unflushed. Only `messages_required_to_save = 1` gates every ack on its own fdatasync. The refusal is unaffected: an acked-but-unflushed message writes neither log bytes nor an index entry, so it can neither manufacture a step-back nor mask one.

Called out rather than fixed: when the log does back the last index entry, the walk starts there and `[0, last.position)` is not re-read at boot, where at-rest damage surfaces on the read path via `validate_checksum` instead. Closing it would mean re-hashing every segment on every boot, since no clean-shutdown marker exists on disk.

## A persist failure now stops the server, not one shard

A local commit failing for an op the cluster had already committed panicked on purpose: the replica is divergent, its `commit_min` never advanced, and the next advance would assert. That panic was meant to end the process and never did. It runs on the shard's message pump, and `compio::runtime::spawn` catches the unwind, so the process lived on with one pump dead. That pump drives every partition on its shard, so all of them stopped committing, ticking and replying while the listeners and other shards kept answering. Clients got no error either: frames to the dead shard were dropped once its inbox filled, so each request blocked until the client's own read timeout.

The partition now records the fault instead of panicking, which also fences it, so `on_ack`, `commit_journal` and the tick become no-ops. The tick reports it to the pump, which breaks, skips the queued tail drain (those frames re-enter the commit path that just failed) and still runs its final flush. The spawn site flips the shared shutdown flag so sibling shards stop gracefully, and the verdict leaves as `ShardFatal`, so the process exits non-zero.

Quarantining only the faulting partition was the alternative. A runtime persist failure is a device signal rather than one bad chain, and a quarantined partition cannot resume without a restart anyway.

## Also in this PR

The active segment now fills the same per-segment read-state slot that sealed segments use. It was populated for sealed segments only, so a consumer following the head paid one `openat` per poll, each an io_uring operation prone to an io-wq punt. It is demand-filled, dropped when the segment seals, and outside the sealed LRU budget so unrelated traffic cannot evict it.

Three smaller fixes: latency extremes are backfilled only when all three of min, max and standard deviation are absent, since backfilling individually let a report mix a raw value with two derived ones. The segment writer save methods are narrowed to the crate. And both index scans yield on a stride, since each can run a megabyte-scale index with no disk read to key a yield on.

## Tests and verification

The integration test SIGKILLs a node after an eager flush with its log cut at the last index entry's position, the only depth a crash can produce under `enforce_fsync`, then asserts the rebuild marker in the boot log, that every rebuilt entry opens inside the truncated log, the surviving prefix served, nothing fenced, and the node caught up. Unit tests cover the refusal, the probe cap, and the commit fault fencing instead of panicking.

fmt, sort and clippy clean with all features and targets. partitions 134, shard 45, server 326, and the `cluster::` suite against a server binary rebuilt from the branch head, since the integration crate does not rebuild it.

### DIFF
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -3091,12 +3091,20 @@ where
             };
 
             // Persist BEFORE eviction so a write failure leaves the rest of the
-            // committed prefix resident for retry. The persist is idempotent on
-            // failure: it advances its write cursors only once both the batch
-            // and the index are durable, so the retry overwrites the same
-            // positions instead of appending a duplicate. Chunks already durable
-            // are evicted before the error propagates, so the retry cannot
-            // re-read them (and re-write them past a rotation).
+            // committed prefix resident instead of dropping it. Nothing retries
+            // the persist: on the commit path a failure panics
+            // `handle_committed_entries`, the shutdown flush warns and moves to
+            // the next namespace, and the transfer offer turns it into
+            // `FlushFailed`. Recovery is a restart, which reopens each writer at
+            // the on-disk length boot recovery validated.
+            //
+            // The ordering therefore buys a non-corrupting failure, not an
+            // in-process one. Write cursors advance only once both the batch and the
+            // index are durable, so where the error does not panic the partition,
+            // its next flush rewrites the same positions instead of appending a
+            // duplicate. Chunks already durable are evicted before the error
+            // propagates, so it cannot re-read them (and re-write them past a
+            // rotation).
             if let Err(error) = self
                 .persist_frozen_batches_to_disk(frozen_batches, index_bytes, batch_count)
                 .await
@@ -3106,9 +3114,9 @@ where
             }
             // Insert the flushed sparse-index entry into the in-mem cache only now
             // that the batch + index are durable. Inserting in the build loop (before
-            // persist) re-inserts a duplicate on a persist-failure retry, which
-            // re-reads the same prefix. The active segment has not rotated yet, so
-            // this targets the segment that received the batches.
+            // persist) re-inserts a duplicate on the next flush after a persist
+            // failure, which re-reads the same prefix. The active segment has not
+            // rotated yet, so this targets the segment that received the batches.
             if let Some(index) = flush_index {
                 self.log.ensure_indexes();
                 let indexes = self.log.active_indexes_mut().expect("indexes must exist");
@@ -3723,6 +3731,22 @@ where
         )
         .await;
 
+        // The match below collapses to the log's error (the durable record, the
+        // index being derived from it). The halves write different files and can
+        // fail for unrelated reasons, so name the index failure here instead of
+        // letting the log's error stand for both.
+        if let (Err(log_error), Err(index_error)) = (&log_result, &index_result) {
+            warn!(
+                target: "iggy.partitions.diag",
+                plane = "partitions",
+                namespace_raw = self.namespace().inner(),
+                batch_count,
+                %log_error,
+                %index_error,
+                "failed to persist frozen batches: log and index both failed"
+            );
+        }
+
         let (saved, saved_index_bytes) = match (log_result, index_result) {
             (Ok(saved), Ok(saved_index_bytes)) => (saved, saved_index_bytes),
             (Err(error), _) | (Ok(_), Err(error)) => {
@@ -3739,12 +3763,13 @@ where
         };
 
         // Advance both cursors only here, back to back with no await between.
-        // They are the writers' next-write positions: a half whose save failed
-        // must keep its cursor so the retry overwrites the same slot instead of
-        // appending a duplicate index entry or leaving a hole in the segment.
-        // Moving them together keeps the pair describing one durable prefix,
-        // which is what a re-opened writer asserts against the length of the
-        // file it finds.
+        // They are the writers' next-write positions, so a half whose save
+        // failed must keep its cursor: the pair then still describes one durable
+        // prefix, which is what a writer re-opened by boot recovery asserts
+        // against the length of the file it finds
+        // (`SegmentSizeMismatchAtOpen`). In-process it also keeps a later flush
+        // rewriting the same slot rather than appending a duplicate index entry
+        // or leaving a hole in the segment.
         messages_writer.advance(saved.as_bytes_u64());
         index_writer.advance(saved_index_bytes);
 
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -1643,20 +1643,22 @@ where
         let (start_segment, start_position) = self.disk_poll_start(&query);
         // Cap resident sealed read handles: touch this poll's start segment so
         // the LRU keeps the hot set and drops the least-recently-used fd +
-        // index (a no-op for the active segment, whose handle never caches).
+        // index (a no-op for the active segment, whose slot is bounded by
+        // rotation instead).
         self.log.touch_sealed_read_state(start_segment);
         // Snapshot only the segments the disk walk visits (`start_segment..`),
-        // so `start_position` applies to the first snapshotted segment. A sealed
-        // segment carries its shared read-state handle (fd + sparse index) so
-        // the off-borrow read reuses (or fills) it; the active segment opens
-        // fresh and resolves from its resident index.
+        // so `start_position` applies to the first snapshotted segment. Every
+        // segment carries its shared read-state handle so the off-borrow read
+        // reuses (or fills) the cached fd; only a sealed one also resolves its
+        // start byte from the shared sparse index.
         let segments = self.log.segments()[start_segment..]
             .iter()
             .zip(self.log.sealed_read_state()[start_segment..].iter())
             .map(|(segment, read_state)| DiskSegment {
                 start_offset: segment.start_offset,
                 persisted: segment.size.as_bytes_u64(),
-                read_state: segment.sealed.then(|| Rc::clone(read_state)),
+                read_state: Rc::clone(read_state),
+                sealed: segment.sealed,
             })
             .collect();
         let disk = DiskReadPlan {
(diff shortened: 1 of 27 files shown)


core/server/src/segment_recovery.rs
```diff
@@ -1001,154 +1035,162 @@ async fn recover_segment_bounds(
...
+                // tail of the INDEX, never proof that the log lost data.
+                // That shape is ordinary: the two files persist concurrently
+                // under every config, so they have no write ordering between
+                // them, and pass C truncates the log before the index, so a
+                // crash between those two mutations leaves exactly this.
+                // Refusing it handed a benign torn tail a permanent tombstone
+                // on a single-replica node. Step down
+                // to the highest entry the log still proves and floor the
+                // index there, so pass C drops the entries nothing backs.
+                let anchor = find_provable_index_anchor(
```
numinnex: The step-back can anchor **above** interior damage, and nothing ever walks `[0, anchor.position)`. `refuse_if_survivor_past_damage` only scans forward from `walk.position` and returns `Ok` immediately when `damage_position >= messages_size` (:1757), so damage below the anchor is invisible to it.

Reachable without hand-crafting an index. One entry per flushed chunk is what the writer emits; at-rest bit rot inside chunk 3, plus the same crash tearing the tail so the last entry cannot prove. The step-back lands on chunk 4, which verifies, and the walk runs from there to EOF and returns `Ok`. The rot at chunk 3 is never read. Result: a checksum-failing batch inside the advertised offset range, `stats.increment_messages_count` at :339 counting offsets that cannot be served, and `current_offset` written into the advance-only superblock. master refused these bytes.

Suggest the floored path fall through to `recover_by_walking_log` (walk from byte 0) instead of a second anchored walk. That reaches the probe with `damage_position < messages_size`, finds the survivor, and refuses with `InteriorDamage` with the bytes preserved.

Related: the justification at :1080-1085 ("an entry the log cannot back is a torn tail of the INDEX, never proof that the log lost data") is false in exactly this case, since the unprovable tail entry *is* evidence the log lost an interior page.

core/server/src/segment_recovery.rs
```diff
@@ -1001,154 +1035,162 @@ async fn recover_segment_bounds(
...
-                        }));
-                    }
-                    return Err(
-                        identity.refusal(PartitionRecoveryRefusal::OffsetDiscontinuity {
-                            start_offset,
-                            expected_offset,
-                            found_offset: header.base_offset,
-                            position,
-                        }),
+            let mut walk = walk_chain_from_anchor(
```
numinnex: The walk still starts at `last.position`, so `[0, last.position)` is accepted on index evidence alone, which is the evidence this PR itself declares insufficient (`server_error.rs:288-294`, and :1370-1377 here).

On master that was safe for a specific reason: `save_frozen_batches` fsyncs the whole file (`messages_writer.rs:129-131`), not a range, so under the old serial order "index entry N is durable" implied "log fdatasync N completed" implied "log bytes [0, N) are all durable". The `futures::future::join` in `iggy_partition.rs` deletes that implication. So under `enforce_fsync = true` this is newly unsound rather than pre-existing, and the premise now applies to every entry rather than only the last.

`stats.increment_messages_count(end_offset - start_offset + 1)` at :339 then reports a count over a span nothing read.

Suggest checksum-walking from byte 0 whenever the index was floored or the segment was not cleanly sealed. At minimum, stop deriving `messages_count` from an unwalked span.

core/partitions/src/iggy_partition.rs
```diff
@@ -3090,9 +3094,9 @@ where
 
             // Persist BEFORE eviction so a write failure leaves the rest of the
             // committed prefix resident for retry. The persist is idempotent on
-            // failure: a batch write that lands but whose index save then fails
-            // rewinds the segment write cursor, so the retry overwrites those
-            // bytes instead of appending a duplicate. Chunks already durable
+            // failure: it advances its write cursors only once both the batch
+            // and the index are durable, so the retry overwrites the same
+            // positions instead of appending a duplicate. Chunks already durable
```
numinnex: There is no retry. The only production consumer of a persist `Err` is `commit_messages_inner` -> `commit_partition_entry` (sets `failed_commit`, returns false) -> `handle_committed_entries`, which panics at :3246. Nothing re-invokes the persist against the un-advanced cursors. The shutdown path (`flush_committed_messages`) warns and moves to the next namespace, and `state_transfer` re-arms the transfer rather than the persist.

That panic also does not do what its own comment promises. The shard message pump runs inside `compio::runtime::spawn` (`bootstrap.rs:1274`), which this repo documents as silently swallowing spawned-task panics (`core/message_bus/tests/installer_panic_cleanup.rs:20-27`, and the two production scopeguards built around it at `installer/replica.rs:424-430` and `installer/tcp.rs:112-117`). There is no `panic = "abort"` and no custom hook. So an ENOSPC or EIO unwinds one shard's pump and the process stays up: the listener keeps accepting, other shards keep running, and every partition on that shard silently stops committing, ticking and replying. `ShardPumpDied` is only raised by `drain_shard_pump`, which runs at shutdown.

The cursor discipline itself is worth keeping: it is what makes the shutdown-flush warn-and-continue path non-corrupting, and boot recovery is the real retry since `SegmentStorage::new` reopens at the recovered lengths. Suggest dropping the retry framing here and at :3743-3749 and documenting boot recovery as the sole recovery, then fixing the wedge separately.

core/server/src/segment_recovery.rs
```diff
@@ -995,148 +1064,133 @@ async fn recover_segment_bounds(
...
+                    stream_id = identity.stream_id,
+                    topic_id = identity.topic_id,
+                    partition_id = identity.partition_id,
+                    start_offset,
+                    messages_size,
+                    entry_count,
+                    last_entry_offset = last.offset,
+                    last_entry_position = last.position,
+                    "the log does not back its last sparse index entry; \
+                     discarding the index and rebuilding it from a byte-0 \
```
hubcio: the step-back is gone - an index whose last entry the log can't back is dropped whole and rebuilt from a byte-0 walk, so the probe reaches the damage and refuses. the old justification went with it.

core/server/src/segment_recovery.rs
```diff
@@ -995,148 +1064,133 @@ async fn recover_segment_bounds(
...
-                    let verifies = scanner
-                        .slice_at(position, header.total_size())
-                        .map_err(|source| scan_read_failure(identity, messages_path, &source))?
-                        .is_some_and(|batch| decode_batch_slice(batch).is_ok());
-                    if !verifies {
-                        break;
-                    }
-                    if foreign_partition {
-                        return Err(identity.refusal(PartitionRecoveryRefusal::ForeignBatch {
+            let walk = walk_chain_from_anchor(
```
hubcio: flushes serialize and cursors advance only after both halves succeed, so entry N+1 proves flush N's log fdatasync finished - the join strands only the last entry, not every entry. `MAX_FSYNCED_INDEX_STEP_BACK_ENTRIES = 1` encodes that, and deeper gaps refuse.

core/partitions/src/iggy_partition.rs
```diff
@@ -3089,12 +3110,21 @@ where
             };
 
             // Persist BEFORE eviction so a write failure leaves the rest of the
-            // committed prefix resident for retry. The persist is idempotent on
-            // failure: a batch write that lands but whose index save then fails
-            // rewinds the segment write cursor, so the retry overwrites those
-            // bytes instead of appending a duplicate. Chunks already durable
-            // are evicted before the error propagates, so the retry cannot
-            // re-read them (and re-write them past a rotation).
+            // committed prefix resident instead of dropping it. Nothing retries
```
hubcio: retry framing is dropped, and the wedge is fixed - the partition fences itself instead of panicking, the pump breaks and flips `shutdown_flag`, and the process exits non-zero as `ServerError::ShardFatal`.

piotr-laserdata (approved):

spetz (approved):

numinnex (approved):

PR #3973, opened 2026-08-26
feat(connectors): add RabbitMQ sink

amr8t:
Replaces #3811 (could not be reopened after rebasing onto master). Rebased to current master and addressed all review comments.

Relates to #3747

## Summary

Adds the RabbitMQ sink connector via the lapin client. Source will be a separate PR.

## Review feedback addressed

- `amqp_url` stored as `secrecy::SecretString`, redacted via `iggy_common::serde_secret` (never logged or serialized verbatim)
- Exchange declaration durability exposed as `durable_exchange` config (default `true`)
- Publisher confirms matched explicitly: only `Ack(None)` counts as success; `Ack(Some(_))`/`Nack(_)` (unroutable mandatory publish) fail the batch permanently
- AMQP delivery mode exposed as `delivery_mode` config (default `persistent`)
- User-supplied Iggy headers forwarded as AMQP headers (`LongString` for strings, `ByteArray` for binary) so headers exchanges can route on them
- `iggy_offset` encoded as full `i64` instead of narrowing to `u32`
- Immediate `basic_publish` errors routed through the same retry flow; the retry loop resumes at the first unconfirmed message instead of republishing confirmed ones
- JSON payloads serialized via `Payload::try_to_bytes()` (no deep clone)
- README documents every config field with type, default, and behavior
- Integration tests cover durable exchanges, unroutable routing keys, and headers-exchange routing; unit tests cover offset > u32::MAX, header encoding, delivery mode, and retry classification

## AI usage

Which tools? opencode
Scope of usage? investigation, code suggestions, implementation of review feedback
How did you verify the generated code works correctly? Read through, compiled, ran unit tests, ran clippy/fmt/sort
Can you explain every line of the code if asked? Yes


core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+                .connection
+                .close(200, "OK")
+                .await
+                .map_err(|e| Error::Connection(e.to_string()))?;
+        }
+        Ok(())
+    }
+}
+
+fn is_publish_retryable(error: &Error) -> bool {
```
numinnex: These substrings don't match what lapin actually emits. A dropped channel renders as `invalid channel state: Closed` (not `channel closed`), heartbeat loss as `no heartbeat received from server for too long`, AMQP uses `RESOURCE_LOCKED` with an underscore, and `io::ErrorKind::TimedOut` prints `timed out`, not `timeout`.

So a channel-level exception is classified permanent and returns at L239 before `reconnect()` is ever called. Since `state` is only written on success (L350), the dead `Channel` is then re-read by every subsequent batch and the sink never self-heals.

Suggest matching on `lapin::Error` variants instead, and clearing `state` on any publish failure — the classifier fix alone still wedges on an error the list doesn't cover.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+                {
+                    Ok(confirm) => confirm,
+                    Err(e) => {
+                        last_error = Some(Error::CannotStoreData(e.to_string()));
+                        break;
+                    }
+                };
+                match confirm.await {
+                    Ok(Confirmation::Ack(None)) => confirmed += 1,
+                    Ok(Confirmation::Ack(Some(_))) | Ok(Confirmation::Nack(_)) => {
```
numinnex: `Nack` probably shouldn't share this arm with `Ack(Some(_))`. `Ack(Some(_))` is the mandatory-return (unroutable) case, whereas a `Nack` means the broker refused responsibility (internal error, disk alarm) — the one confirm outcome RabbitMQ documents as safe to re-publish.

As written it becomes `InvalidRecordValue`, which `is_publish_retryable` rejects, so the batch fails permanently. Splitting the arms with `Nack` retryable would also fix the message, which currently misreports the cause.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+                    )
+                    .await
+                {
+                    Ok(confirm) => confirm,
+                    Err(e) => {
+                        last_error = Some(Error::CannotStoreData(e.to_string()));
+                        break;
+                    }
+                };
+                match confirm.await {
```
numinnex: Awaiting the confirm inside the per-message loop serializes a broker round-trip per message, so throughput is capped at 1/RTT regardless of payload or batch size (~2k msg/s on a LAN, well under that cross-AZ). With `batch_length = 100` here and 1000 as the runtime default, that's 100–1000 sequential RTTs per batch, and it also stops lapin from coalescing frames.

Worth flagging that the obvious fix isn't safe as-is: lapin pairs `Basic.Return` to a confirm FIFO with no delivery tag, so with several publishes in flight an unroutable message can be attributed to the wrong one. Pipelining would need `mandatory` off on that path, or a per-message `message_id` to correlate on.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+            .compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire)
+            .is_err()
+        {
+            tokio::time::sleep(self.retry_delay).await;
+            return Ok(());
+        }
+
+        warn!("Reconnecting RabbitMQ sink ID: {}", self.id);
+        let result = async {
+            let conn = Connection::connect(
```
numinnex: No timeout on `Connection::connect` here, nor on `basic_publish` / `confirm.await` in the publish loop. RabbitMQ blocks publishers on a disk or memory alarm while still answering heartbeats, so the confirm can hang indefinitely, and a blackholed SYN here waits out the kernel TCP timeout while holding the `reconnecting` CAS.

Since the FFI entry point is `block_on`, that parks a connectors-runtime worker thread. `surrealdb_sink` and `clickhouse_sink` both expose a timeout config for this.

core/connectors/sinks/rabbitmq_sink/README.md
```diff
@@ -0,0 +1,38 @@
...
+| `include_metadata` | bool | `true` | Add `iggy_stream`, `iggy_topic`, `iggy_partition_id`, `iggy_offset` message headers. User-supplied headers are always preserved, regardless of this flag. |
+| `verbose_logging` | bool | `false` | Log each published batch at `info` level instead of `debug`. |
+| `max_retries` | u32 | `3` | Maximum transient publish retries before failing the batch. |
+| `retry_delay_secs` | u64 | `1` | Base retry delay in seconds. |
+| `max_retry_delay_secs` | u64 | `5` | Upper bound for exponential backoff. |
+
+User headers on consumed Iggy messages are forwarded as AMQP headers: string values become AMQP `LongString`, raw binary values become `ByteArray`. This allows routing through a `headers` exchange on original user headers.
+
+Publishes are confirmed via `ConfirmSelect`. With `mandatory = true`, a message with a routing key that matches no binding is returned by RabbitMQ and the batch fails with a permanent error (delivery is at-least-once: if the connection drops mid-batch, the sink resumes from the first unconfirmed message, so a broker-side outcome may be unknowable and could be delivered more than once).
```
numinnex: The at-least-once claim doesn't hold as shipped. The runtime commits offsets at poll time (`AutoCommitWhen::PollingMessages`) and discards the plugin's FFI return code, so a permanently-failed batch is dropped and still counted as processed.

The real guarantee is at-most-once across batches, with at-least-once only within one. `doris_sink/README.md` has the house wording for this.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Makes sense. Switched to matching lapin::Error variants

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Split Nack  from Ack.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Added a timeout_secs config

core/connectors/sinks/rabbitmq_sink/README.md
amr8t: Fixed!

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: This is tricky. I've pipelined publishes and confirm them in order, removing the 1/RTT cap using the message_id stamp. This is same as the suggestion.  A failure after the publish phase can re-publish the message twice. Lapin's Return-to-confirm pairing has no delivery tag, hence using the message_id.

amr8t:
Thanks for the comments @numinnex ! Addressed, pls take a look

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+                        last_retryable = true;
+                        break;
+                    }
+                };
+                in_flight.push((confirm, message.offset));
+            }
+
+            // Confirms resolve in publish order over already-overlapped RTTs; returns are
+            // matched by message_id (lapin's Return-to-confirm FIFO has no delivery tag), so a
+            // post-publish failure re-publishes the already-delivered tail (at-least-once in batch).
```
numinnex, line 248: The comment claims returns are "matched by message_id", but nothing matches on it. `returned_id` at L253 is only `{:?}`-formatted into the error string, and `confirmed` is never corrected.

lapin staples a return to whichever confirm resolves next: `complete_pending` pops `returned_messages.get_waiting_message()` from a FIFO with no delivery tag (`lapin-2.5.5/src/acknowledgement.rs:115-121`), and for `Basic.Ack{multiple:true}` — RabbitMQ's normal mode under load — `complete_pending_before` collects tags into a `HashSet<DeliveryTag>` before iterating (`acknowledgement.rs:155-168`, reached via `channel.rs:1144-1147`). So the offset reported at L254 is nondeterministic and usually not the message that was actually returned; the genuinely unroutable one resolves `Ack(None)` and is counted into `confirmed`.

This is the misattribution hazard flagged in the earlier confirm-pipelining thread, and that thread was resolved on the premise that message_id matching had been implemented. It has not been. Either parse `returned.delivery.properties.message_id()` back to an offset and fail only that message, or drop the claim from the comment and document the offset as unattributable.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+            for (confirm, offset) in in_flight {
+                match timeout(self.timeout, confirm).await {
+                    Ok(Ok(Confirmation::Ack(None))) => confirmed += 1,
```
numinnex, line 251: `confirmed += 1` here can count a message the broker refused.

A `break` on any earlier arm drops the remaining `PublisherConfirm`s, but dropping one does not remove its broadcaster from `Acknowledgements::Inner::pending` (`acknowledgement.rs:99-111` inserts; only `drop_pending`/`drop_all` remove). When that orphaned tag's ack finally arrives, `complete_pending` (`acknowledgement.rs:113-121`) unconditionally pops `get_waiting_message()` and staples a *later* batch's `Basic.Return` onto the dead confirm, where `register_dropped_confirm` parks it unseen (`returned_messages.rs:116-130`). The live confirm for that later batch then resolves `Ack(None)` and lands on this line.

Result: an unroutable message is counted in `confirmed` and `messages_published`, `consume()` returns `Ok(())`, and the runtime has already committed the offset (`AutoCommitWhen::PollingMessages`, `core/connectors/runtime/src/sink.rs:522`) while discarding the FFI return code. Silent loss reported as a successful delivery, with no log and no metric.

The window is not narrow. `delivery_mode = 2` (the default) defers the ack until RabbitMQ's batched msg-store fsync (~100-200 ms), while the shipped `poll_interval = "5ms"` re-enters `consume()` two orders of magnitude sooner, so the next batch is publishing on the same channel while the previous batch's orphaned tags are still unacked. The two preconditions are also correlated: the `Ack(Some)` break that creates the orphans is the same unroutable path that produces the returns.

Fixed by never reusing a channel that has unresolved confirms — see the comment on the confirm loop below.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+                        last_error =
+                            Some(Error::CannotStoreData("message nack'd by RabbitMQ".into()));
+                        last_retryable = true;
+                        break;
+                    }
+                    Ok(Ok(Confirmation::NotRequested)) => {
+                        last_error = Some(Error::CannotStoreData(
+                            "publisher confirms not enabled".into(),
+                        ));
+                        last_retryable = false;
```
numinnex, line 270: Every `break` in this loop abandons the unread `in_flight` tail, and the two permanent arms (`Ack(Some)` here, `NotRequested` at L266) return at L296-302 without `clear_state()`, so the same channel is reused on the next poll.

`PublisherConfirm::drop` hands each abandoned confirm to `register_dropped_confirm` (`lapin-2.5.5/src/publisher_confirm.rs:81-88` -> `returned_messages.rs:116-130`), which parks the promise in `Inner::dropped_confirms` or, if a return already resolved it, the whole `BasicReturnMessage` — including the full `Delivery.data` payload — in `Inner::messages`. Both are cleared only by `Inner::drain()`, whose sole caller is `Channel::wait_for_confirms` (`channel.rs:334-341`). This sink never calls it.

With a misconfigured routing key and `mandatory: true`, a `batch_length = 100` batch retains ~99 entries with their payloads, every poll, for the process lifetime. Unbounded RSS growth with no diagnostic. The same orphans are what enable the counted-as-delivered case flagged at L251.

Suggest `clear_state().await` on every non-`Ok` exit from this loop (safer than `wait_for_confirms()`, which can itself block on a stalled broker), so a channel with unresolved confirms is never reused.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+                    Some(state) => state.channel.clone(),
+                    None => {
+                        drop(guard);
+                        self.reconnect().await?;
+                        self.state
+                            .lock()
+                            .await
+                            .as_ref()
+                            .map(|s| s.channel.clone())
+                            .ok_or_else(|| Error::Connection("RabbitMQ not connected".into()))?
```
numinnex, line 196: This can drop a whole batch without consuming a single retry attempt.

`reconnect()` is a try-lock: a caller that loses the CAS at L372-378 sleeps `retry_delay` and returns `Ok(())` having connected nothing. Control then arrives here, `state` is still `None`, and the `?` propagates `Error::Connection("RabbitMQ not connected")` straight out of `publish_batch_with_retry` with `attempts == 0` and `publish_errors` never incremented — the retry loop, the backoff and the error accounting are all bypassed.

The runtime discards the FFI return code and has already committed the offset at poll time (`core/connectors/runtime/src/sink.rs:522` and `:745-753`), so the batch is unrecoverable and invisible.

Concurrency is real whenever a connector has 2+ (stream, topic) pairs: `setup_sink_consumers` builds one consumer per pair (`runtime/src/sink.rs:519-546`) and `spawn_consume_tasks` spawns one task each against the same plugin instance (`:266-299`). The window is widest exactly when it matters — the CAS winner can hold the flag for up to `timeout_secs` (30s default, and unbounded given the untimed channel RPCs noted at L395) while the loser gives up after 1s.

The loser should wait on the winner's outcome and re-read state, or return a retryable error. It should not report success for work it did not do.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+                let body = message.payload.try_to_bytes()?;
+                let mut props = BasicProperties::default()
+                    .with_delivery_mode(self.delivery_mode)
+                    .with_message_id(ShortString::from(message.offset.to_string()));
+                let headers = self.build_headers(topic_metadata, messages_metadata, message);
+                if !headers.inner().is_empty() {
+                    props = props.with_headers(headers);
+                }
+
+                let confirm = match timeout(
```
numinnex, line 215: Wrapping `basic_publish` in `timeout()` does not make it abortable, so a publish that "times out" here has usually already been delivered.

`before_basic_publish` registers the pending confirm and burns a delivery tag before anything is sent (`lapin-2.5.5/src/channel.rs:503-509`), and `send_method_frame_with_body` pushes every frame into `Frames` and wakes the io_loop synchronously on first poll, only then awaiting the write promise (`channel.rs:393-408` -> `frames.rs:149-170`). tokio's `Timeout` polls the inner future before checking the deadline, so that prologue always runs.

The timeout arm at L237-241 treats this as not-sent: it breaks without `clear_state()` and the retry republishes `messages[confirmed..]`, so the tail is delivered twice and each cancelled future leaves another orphaned confirm on the live channel (see L252). A publish timeout is a sent-unknown outcome, not a not-sent one.

Related: `timeout_secs` is accepted unvalidated at L151, so `timeout_secs = 0` makes every publish time out while every message still lands.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+        {
+            tokio::time::sleep(self.retry_delay).await;
+            return Ok(());
+        }
+
+        warn!("Reconnecting RabbitMQ sink ID: {}", self.id);
+        let result = async {
+            let conn = timeout(
+                self.timeout,
+                Connection::connect(
```
numinnex, line 386: This timeout does not bound the connect syscall, which is the case raised in the earlier timeouts thread ("a blackholed SYN here waits out the kernel TCP timeout").

lapin runs the connect under `executor.spawn_blocking` (`lapin-2.5.5/src/connection.rs:300-337`), and `amq-protocol-tcp-7.2.3/src/lib.rs:43-48` calls a bare `TcpStream::connect` unless the URI carries `?connection_timeout=<ms>`. Since `ConnectionProperties::default()` leaves the executor unset, that lands on async-global-executor's blocking pool. When the outer `timeout` fires it cancels the await only; the blocking closure keeps running to the kernel SYN timeout (~127s), leaking one wedged pool thread per attempt, in a pool the connectors runtime cannot see.

Minimal fix is to append `?connection_timeout={timeout_ms}` to the AMQP URI (or supply a custom connector). Same applies to the `open()` path at L433.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
```diff
@@ -0,0 +1,746 @@
...
+                .create_channel()
+                .await
+                .map_err(|e| Error::Connection(e.to_string()))?;
+            channel
+                .confirm_select(ConfirmSelectOptions::default())
+                .await
+                .map_err(|e| Error::Connection(e.to_string()))?;
+            let exchange_kind = self.exchange_kind()?;
+            channel
+                .exchange_declare(
```
numinnex, line 404: `create_channel`, `confirm_select` and `exchange_declare` are awaited with no timeout, in both `reconnect()` and `open()` (L442-455). Only `Connection::connect` is wrapped.

A broker that completes the TCP handshake and answers heartbeats while stalling channel RPCs — the usual shape of a disk or memory alarm, or a half-open connection behind a load balancer — hangs here, bounded only by the negotiated heartbeat (~2x60s on RabbitMQ defaults) and unbounded if the server negotiates `heartbeat=0`.

This compounds the `reconnect()` try-lock: the CAS winner is parked inside this block with `reconnecting == true` and no deadline, so every concurrent consume task takes the loser path and drops its batch (see the comment at L196).

Wrapping the whole open/reconnect body in a single `timeout(self.timeout, async { .. })` covers these three and the await half of the connect.

numinnex (changes requested):

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Fixed, dropping message and failing early

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Fixed, doing a clear_state()

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Fixed, clearing state before each break arm.

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: CAS is now a reconnect mutex, losers wait on the winner and re-read state, and acquisition failures count as retry attempts with backoff

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: both timeout arms now clear_state()

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Added ?connection_timeout={timeout_ms} to the AMQP URI (

core/connectors/sinks/rabbitmq_sink/src/lib.rs
amr8t: Wrapping open/reconnected in a timeout

numinnex (approved):

numinnex:
LGTM, Thanks for the contribution.

hubcio (approved):

spetz (approved):

PR #3975, opened 2026-08-27
fix(partitions): reserve offsets before confirming them

krishvishal:
## The defect

A solo node ACKs sends from its in-memory journal. The client receives a concrete base offset before the threshold-gated flush writes it to a segment. A SIGKILL in this window loses the only record that the offset was issued.

```text
LIFE 1  threshold = 4 messages
        offsets 0 1 2 3 | 4 5 6 7 8 9    all ACKed
        disk:   [0..3]  | RAM only         lost on SIGKILL

BOOT    counter := highest segment offset + 1 = 4

LIFE 2  next send ACKed at offset 4         already issued for another message
```

Two messages now share offset 4. Consumers positioned above the reissued range also miss the new messages.

`offset_frontier` already existed in the superblock, but stable-view traffic never persisted it because its write gate only ran on view changes.

## The fix

Add `offset_reserved`, a monotonic ceiling on offsets that may have been issued. Before an offset can escape, the append fence reserves its block in the superblock. On boot, minting starts at this ceiling.

```text
send -> mint -> reserve_offsets_through -> journal -+-> commit -> ACK
                   superblock write                 +-> poll tier
                                                    +-> prepare to peers
```

The reservation covers both primary mints and backup re-stamps. It requires one write per block, not per batch. With the default 64 Ki lease at 100,000 messages per second, this is about three fsyncs per second. A crash wastes at most one block of the `u64` offset space.

The lease is configurable through `[partition] offset_reservation_lease`. A failed reservation rejects the append.

### Why a second field?

```text
segments  [0 ..... 3]
journal            [4 ..... 9]   ACKed, RAM only
                              ^10                              ^65537
                   offset_frontier                     offset_reserved

                   offsets known to exist              offsets that may
                                                       have been issued
```

Combining the fields would make valid transfer offers inside the reserved block look like rewinds. The rewind guard must instead compare against stored data: sized segments plus the resident journal. The append counter may be one lease block ahead after recovery.

### Segment re-anchoring

A gap inside a segment is not recoverable. `recover_segment_bounds` expects contiguous offsets and truncates everything after a gap, allowing another crash to reissue confirmed offsets.

```text
BEFORE                              AFTER BOOT RE-ANCHORING

00000000000000000000.log [0..3]     00000000000000000000.log [0..3] SEALED
  append 65537 inside it             00000000000000065537.log [empty]
  next boot truncates suffix         gap lies on the segment boundary
```

An empty tail that falsely claims a range is removed. A sized tail is sealed, and a new segment is created at the frontier. Graceful shutdown collapses the reservation to the frontier, so only crashes spend offsets.

This applies only to solo groups. A backup rejects prepares whose `base_offset` does not continue its counter.

### Upgrading and rolling back

The superblock record grows from 66 to 74 bytes. Upgrading is transparent: a 66-byte record decodes with the reservation seeded from the frontier it carries. Rolling back is not, since a build that predates the field rejects a 74-byte record, so a downgrade needs the data directory wiped on every node.

Anyone running a build from an EARLIER push of this branch must also wipe their data directory before running this one. Those builds planted re-anchor gaps without the `.anchor` record the chain guard now requires, so this build reads such a chain as a lost segment and the solo arm tombstones the partition. `master` plants no gaps and edge images come only from `master`, so nothing deployed is affected.

### DIFF
--- a/core/configs/src/server_config/partition.rs
+++ b/core/configs/src/server_config/partition.rs
@@ -44,6 +44,7 @@ use crate::common::validators::SEGMENT_MAX_SIZE_BYTES;
 use configs::ConfigEnv;
 use iggy_common::{IggyByteSize, Validatable};
 use serde::{Deserialize, Serialize};
+use std::num::NonZeroU32;
 
 /// Mirrors `consensus::PIPELINE_PREPARE_QUEUE_MAX`.
 pub const DEFAULT_PARTITION_PREPARE_QUEUE_DEPTH: usize = 32;
@@ -97,6 +98,22 @@ pub const DEFAULT_TRANSFER_SERVED_CACHE_BYTES_MAX: u64 =
 /// count.
 pub const MAX_TRANSFER_BYTES: u64 = 64 * 1024 * 1024 * 1024;
 
+/// Upper bound on `offset_reservation_lease`: a typo guard, so a slipped digit
+/// cannot reach the arithmetic in the append fence.
+pub const MAX_OFFSET_RESERVATION_LEASE: u32 = 16 * 1024 * 1024;
+
+/// Mirrors `partitions::DEFAULT_OFFSET_RESERVATION_LEASE`; pinned against drift
+/// by `default_offset_reservation_lease_matches_partitions_constant` in the
+/// server crate, which can see both.
+pub const DEFAULT_OFFSET_RESERVATION_LEASE: u32 = 64 * 1024;
+
+/// Serde fallback for a `[partition]` table that omits
+/// `offset_reservation_lease`.
+fn default_offset_reservation_lease() -> NonZeroU32 {
+    NonZeroU32::new(DEFAULT_OFFSET_RESERVATION_LEASE)
+        .expect("DEFAULT_OFFSET_RESERVATION_LEASE is a nonzero literal")
+}
+
 /// Mirrors `partitions::EVICTED_RING_CAPACITY`.
 pub const DEFAULT_EVICTED_RING_CAPACITY: usize = 4096;
 
@@ -143,6 +160,27 @@ pub struct PartitionConfig {
     /// worst case scales with partition count: size it to the producers a
     /// single partition actually sees, not the node's client total.
     pub dedup_clients_max: usize,
+
+    /// Offsets claimed in the superblock ahead of the mint counter before an
+    /// append, so a crash-restarted replica resumes above what it confirmed.
+    /// One superblock write per block: lowering it raises the fsync rate,
+    /// raising it wastes at most one block per crash. Must be <=
+    /// [`MAX_OFFSET_RESERVATION_LEASE`].
+    ///
+    /// SINGLE-REPLICA groups only: a replicated group acks once a quorum has
+    /// journaled the batch, claims nothing, and ignores this.
+    ///
+    /// `NonZeroU32` rather than a `u32` with a floor check: a zero lease claims
+    /// nothing and would write the superblock before every append, and the type
+    /// is what stops the partition-side setter from having to silently coerce it
+    /// to one, a coercion that hid wiring errors the validator could not see.
+    ///
+    /// The serde fallback is for the providers that do NOT merge the embedded
+    /// defaults: the file provider does, so a partial `[partition]` table only
+    /// fails through direct deserialization or an alternate provider.
+    #[serde(default = "default_offset_reservation_lease")]
+    #[config_env(leaf)]
+    pub offset_reservation_lease: NonZeroU32,
 
     /// Entries the evicted ring retains per multi-replica partition for
     /// journal repair after a peer rejoins. Larger widens the window a
@@ -203,6 +241,14 @@ impl Validatable<ConfigurationError> for PartitionConfig {
                 "{COMPONENT} partition.dedup_clients_max ({}) must be > 0 and <= \
                  {PARTITION_DEDUP_CLIENTS_CEILING}",
                 self.dedup_clients_max
+            );
+            return Err(ConfigurationError::InvalidConfigurationValue);
+        }
+        if self.offset_reservation_lease.get() > MAX_OFFSET_RESERVATION_LEASE {
+            eprintln!(
+                "{COMPONENT} partition.offset_reservation_lease ({}) exceeds the maximum \
+                 ({MAX_OFFSET_RESERVATION_LEASE})",
+                self.offset_reservation_lease
             );
             return Err(ConfigurationError::InvalidConfigurationValue);
         }
@@ -327,6 +373,69 @@ mod tests {
     fn accepts_prepare_queue_depth_at_ceiling() {
         let config = PartitionConfig {
             prepare_queue_depth: MAX_PARTITION_PREPARE_QUEUE_DEPTH,
+            ..PartitionConfig::default()
+        };
+        assert!(config.validate().is_ok());
+    }
+
+    fn lease(value: u32) -> NonZeroU32 {
+        NonZeroU32::new(value).expect("a nonzero test lease")
+    }
+
+    /// A `[partition]` table as an alternate provider hands it over: every other
+    /// field present, the lease optional.
+    fn partial_table(lease: Option<u32>) -> String {
+        let entry = lease.map_or_else(String::new, |lease| {
+            format!(r#""offset_reservation_lease": {lease},"#)
+        });
+        format!(
+            r#"{{"prepare_queue_depth": 32, {entry}
+               "dedup_clients_max": 4096,
+               "evicted_ring_capacity": 4096,
+               "evicted_ring_bytes_max": "16 MiB",
+               "transfer_served_cache_bytes_max": "64 MiB",
+               "transfer_artifact_bytes_max": "64 MiB"}}"#
+        )
+    }
+
+    /// The floor is the type's, so a zero cannot be constructed to validate --
+    /// it is refused at deserialization instead.
+    #[test]
+    fn rejects_zero_offset_reservation_lease_at_deserialization() {
+        let error = serde_json::from_str::<PartitionConfig>(&partial_table(Some(0)))
+            .expect_err("a zero lease reserves nothing and must not deserialize");
+        assert!(
+            error.to_string().contains("nonzero"),
+            "the refusal must name the constraint, got {error}"
+        );
+    }
+
+    /// The providers that do not merge the embedded defaults hand over a partial
+    /// table, which must still deserialize.
+    #[test]
+    fn given_a_table_without_the_lease_when_deserialized_should_fall_back_to_the_default() {
+        let config = serde_json::from_str::<PartitionConfig>(&partial_table(None))
+            .expect("a table omitting the lease must deserialize");
+        assert_eq!(
+            config.offset_reservation_lease.get(),
+            DEFAULT_OFFSET_RESERVATION_LEASE
+        );
+        assert!(config.validate().is_ok());
+    }
+
+    #[test]
+    fn rejects_offset_reservation_lease_above_ceiling() {
+        let config = PartitionConfig {
+            offset_reservation_lease: lease(MAX_OFFSET_RESERVATION_LEASE + 1),
+            ..PartitionConfig::default()
+        };
+        assert!(config.validate().is_err());
+    }
+
+    #[test]
+    fn accepts_offset_reservation_lease_at_ceiling() {
+        let config = PartitionConfig {
+            offset_reservation_lease: lease(MAX_OFFSET_RESERVATION_LEASE),
             ..PartitionConfig::default()
         };
         assert!(config.validate().is_ok());
--- a/core/consensus/src/vsr_state.rs
+++ b/core/consensus/src/vsr_state.rs
@@ -31,20 +31,38 @@ use std::fmt;
 /// Number of bytes [`VsrState::to_bytes`] produces: `cluster`(16) +
 /// `replica_id`(1) + `replica_count`(1) + `view`(4) + `log_view`(4) +
 /// `commit_max`(8) + `checkpoint_op`(8) + `checkpoint_checksum`(16) +
-/// `offset_frontier`(8).
-pub const ENCODED_LEN: usize = 66;
-
-/// The layout before `offset_frontier` was appended.
+/// `offset_frontier`(8) + `offset_reserved`(8).
 ///
-/// [`VsrState::try_from`] still accepts records of this length and zero-fills
-/// the new field. Without it every superblock already on disk -- the metadata
-/// plane writes one on every view change and checkpoint, single-node included --
-/// would decode as [`VsrStateError::WrongLength`] and refuse boot as a
-/// durability violation. A version bump instead of this would not help on its
-/// own: `classify` compares the version for exact equality, so a v2 build turns
-/// every v1 record into `Unreadable`, which is the same refusal wearing a
-/// different name.
-pub const ENCODED_LEN_WITHOUT_FRONTIER: usize = 58;
+/// Growing this is one-way: a record of this length is [`VsrStateError::WrongLength`]
+/// to every build that predates the field, so a ROLLBACK needs the data directory
+/// wiped even though the upgrade does not. Stated for operators beside
+/// `partition.offset_reservation_lease` in `config.toml`.
+pub const ENCODED_LEN: usize = 74;
+
+/// The layout before `offset_reserved` was appended.
+///
+/// [`VsrState::try_from`] accepts records of this length. Without it every
+/// superblock already on disk decodes as [`VsrStateError::WrongLength`], which
+/// the metadata plane treats as a durability violation and refuses the whole
+/// node's boot on.
+///
+/// "Already on disk" is not a clustering concern. `PingPongSuperblock::open`
+/// runs unconditionally in metadata recovery, so EVERY server writes one of
+/// these into `metadata/superblock.a`, clustered or not, on every view change
+/// and checkpoint. Every release from `server-0.9.0-edge.2` (the first that
+/// carries this module at all) through `edge.6` wrote exactly 66 bytes,
+/// single-node deployments that never enabled clustering included.
+///
+/// The 58-byte layout that preceded it is deliberately NOT accepted: it left
+/// trunk before any release carried it -- `server-0.8.2-edge.1` has no
+/// `vsr_state.rs`, and `edge.2` already wrote 66 -- so a record that short is
+/// corruption, not history.
+///
+/// A version bump instead of this would not help on its own: `classify` compares
+/// the version for exact equality, so a v2 build turns every v1 record into
+/// `Unreadable`, which is the same refusal wearing a different name -- and it
+/// would make this tolerance unreachable, refusing even records that decode.
+pub const ENCODED_LEN_WITHOUT_RESERVATION: usize = 66;
 
 /// The durable consensus state of one replica for one consensus group.
 ///
@@ -97,6 +115,20 @@ pub struct VsrState {
     ///
     /// Always `0` on the metadata plane, which mints no message offsets.
     pub offset_frontier: u64,
+    /// PARTITION plane: a monotone CEILING on the offsets this replica may
+    /// already have minted, claimed ahead of the counter in blocks so an append
+    /// pays one superblock write per block instead of one per batch.
+    ///
+    /// Never folded into [`Self::offset_frontier`]. The frontier is a claim
+    /// about DATA, which state transfer's rewind guard refuses to destroy; a
+    /// reservation names no bytes, only "an offset up to here may have reached
+    /// a client". Comparing an offer against it refuses every legitimate offer
+    /// below the lease headroom, and the replica cycles transfer -> refusal ->
+    /// backoff forever. Boot seeds the mint counter from both; the rewind guard
+    /// reads the frontier alone.
+    ///
+    /// Always `0` on the metadata plane, which mints no message offsets.
+    pub offset_reserved: u64,
 }
 
 impl VsrState {
@@ -113,6 +145,7 @@ impl VsrState {
         out[34..42].copy_from_slice(&self.checkpoint_op.to_le_bytes());
         out[42..58].copy_from_slice(&self.checkpoint_checksum.to_le_bytes());
         out[58..66].copy_from_slice(&self.offset_frontier.to_le_bytes());
+        out[66..74].copy_from_slice(&self.offset_reserved.to_le_bytes());
         out
     }
 }
@@ -121,16 +154,27 @@ impl TryFrom<&[u8]> for VsrState {
     type Error = VsrStateError;
 
     fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
-        // Length-tolerant: a pre-`offset_frontier` record is padded out and the
-        // new field reads as 0, which is exactly "no recorded frontier" (the
-        // read sites filter it). One length check up front then puts every
-        // field slice below in bounds by construction, so the `try_into`s
-        // cannot fail.
+        // Length-tolerant for ONE legacy layout, the pre-`offset_reserved` record
+        // every tagged release wrote. The one length check then puts every field
+        // slice below in bounds by construction, so the `try_into`s cannot fail.
+        //
+        // `offset_reserved` is filled from `offset_frontier`, not zeroed. The two
+        // agree on a record written before the reservation existed: the frontier
+        // is what that build's data proved, and the write side clamps the
+        // reservation up to it anyway, so this is the same value a first write
+        // under this build would record. A 0 would instead claim "nothing
+        // reserved" for offsets the frontier says exist.
+        //
+        // It cannot recover what the old build never wrote down -- offsets acked
+        // out of RAM above the frontier are gone with the process either way --
+        // but refusing the record recovers nothing and costs the node its boot.
         let mut padded = [0u8; ENCODED_LEN];
         match bytes.len() {
             ENCODED_LEN => padded.copy_from_slice(bytes),
-            ENCODED_LEN_WITHOUT_FRONTIER => {
-                padded[..ENCODED_LEN_WITHOUT_FRONTIER].copy_from_slice(bytes);
+            ENCODED_LEN_WITHOUT_RESERVATION => {
+                padded[..ENCODED_LEN_WITHOUT_RESERVATION].copy_from_slice(bytes);
+                padded[ENCODED_LEN_WITHOUT_RESERVATION..ENCODED_LEN]
+                    .copy_from_slice(&bytes[58..66]);
             }
             actual => {
                 return Err(VsrStateError::WrongLength {
@@ -150,6 +194,7 @@ impl TryFrom<&[u8]> for VsrState {
             checkpoint_op: u64::from_le_bytes(field(bytes, 34)),
             checkpoint_checksum: u128::from_le_bytes(field(bytes, 42)),
             offset_frontier: u64::from_le_bytes(field(bytes, 58)),
+            offset_reserved: u64::from_le_bytes(field(bytes, 66)),
         };
         // A record violating `log_view <= view` decodes into a replica that looks
         // healthy locally while `DoViewChangeHeader::validate` makes every peer drop
@@ -190,8 +235,8 @@ impl fmt::Display for VsrStateError {
             Self::WrongLength { expected, actual } => {
                 write!(
                     f,
-                    "VsrState needs {expected} bytes (or {ENCODED_LEN_WITHOUT_FRONTIER}, \
-                     the layout before the offset frontier), got {actual}"
+                    "VsrState needs {expected} bytes (or {ENCODED_LEN_WITHOUT_RESERVATION}, \
+                     the layout before the offset reservation), got {actual}"
                 )
             }
             Self::LogViewAheadOfView { view, log_view } => write!(
@@ -224,7 +269,8 @@ mod tests {
             commit_max: 6,
             checkpoint_op: 7,
             checkpoint_checksum: 8,
-            offset_frontier: 0,
+            offset_frontier: 9,
+            offset_reserved: 10,
         };
         let bytes = state.to_bytes();
         assert_eq!(bytes.len(), ENCODED_LEN);
@@ -236,17 +282,20 @@ mod tests {
         assert_eq!(bytes[26], 6, "commit_max low byte");
         assert_eq!(bytes[34], 7, "checkpoint_op low byte");
         assert_eq!(bytes[42], 8, "checkpoint_checksum low byte");
+        assert_eq!(bytes[58], 9, "offset_frontier low byte");
+        assert_eq!(bytes[66], 10, "offset_reserved low byte");
 
         assert_eq!(VsrState::try_from(&bytes[..]).unwrap(), state);
         assert!(VsrState::try_from(&bytes[..ENCODED_LEN - 1]).is_err());
     }
 
-    /// A superblock written before `offset_frontier` existed must still decode:
-    /// the metadata plane writes one on every view change, so an exact-length
-    /// decode turns an in-place upgrade into a boot refusal on every deployment
-    /// that ever ran.
+    /// A superblock written before `offset_reserved` existed must still decode:
+    /// every release from `server-0.9.0-edge.2` to `edge.6` wrote that layout on
+    /// both planes -- the metadata one unconditionally, so single-node
+    /// deployments that never enabled clustering have one -- and the metadata
+    /// plane refuses the whole node's boot on a record it cannot decode.
     #[test]
-    fn given_pre_frontier_record_when_decoded_should_accept_and_zero_fill() {
+    fn given_pre_reservation_record_when_decoded_should_fill_from_the_frontier() {
         let full = VsrState {
             cluster: 3,
             replica_id: 1,
@@ -257,23 +306,62 @@ mod tests {
             checkpoint_op: 7,
             checkpoint_checksum: 5,
             offset_frontier: 77,
+            offset_reserved: 88,
         }
         .to_bytes();
 
-        let legacy = &full[..ENCODED_LEN_WITHOUT_FRONTIER];
-        let decoded = VsrState::try_from(legacy).expect("a pre-frontier record must decode");
-        assert_eq!(decoded.offset_frontier, 0, "the new field zero-fills");
+        let legacy = &full[..ENCODED_LEN_WITHOUT_RESERVATION];
+        let decoded = VsrState::try_from(legacy).expect("a pre-reservation record must decode");
+        assert_eq!(decoded.offset_frontier, 77);
+        assert_eq!(
+            decoded.offset_reserved, 77,
+            "the reservation fills from the frontier, not from zero: a 0 would claim \
+             nothing was reserved for offsets the frontier says exist"
+        );
         assert_eq!(decoded.view, 9);
         assert_eq!(decoded.log_view, 8);
         assert_eq!(decoded.commit_max, 41);
         assert_eq!(decoded.checkpoint_op, 7);
         assert_eq!(decoded.checkpoint_checksum, 5);
 
-        // Anything that is neither layout is still refused.
-        assert!(matches!(
-            VsrState::try_from(&full[..40]),
-            Err(VsrStateError::WrongLength { .. })
-        ));
+        // Anything that is neither layout is still refused, the 58-byte
+        // pre-frontier layout included: `server-0.8.2-edge.1` has no
+        // `vsr_state.rs` and `edge.2` already wrote 66, so no release ever put a
+        // 58-byte record on a disk and one that short is corruption, not
+        // history.
+        for short in [40, 58] {
+            assert!(
+                matches!(
+                    VsrState::try_from(&full[..short]),
+                    Err(VsrStateError::WrongLength { .. })
+                ),
+                "a {short}-byte record must not decode"
+            );
+        }
+    }
+
+    /// A zero frontier is the shape a record written before either offset field
+    /// existed decodes into, and the fill must not turn that into a claim.
+    #[test]
+    fn given_pre_reservation_record_with_no_frontier_when_decoded_should_reserve_nothing() {
+        let full = VsrState {
+            cluster: 3,
+            replica_id: 1,
+            replica_count: 3,
+            view: 2,
+            log_view: 2,
+            commit_max: 0,
+            checkpoint_op: 0,
+            checkpoint_checksum: 0,
+            offset_frontier: 0,
+            offset_reserved: 0,
+        }
+        .to_bytes();
+
+        let decoded = VsrState::try_from(&full[..ENCODED_LEN_WITHOUT_RESERVATION])
+            .expect("a pre-reservation record must decode");
+        assert_eq!(decoded.offset_frontier, 0);
+        assert_eq!(decoded.offset_reserved, 0);
     }
 
     #[test]
@@ -294,9 +382,11 @@ mod tests {
             // Distinct and nonzero: with 0 here a transposed write over the
             // trailing field would still satisfy every assertion below.
             offset_frontier: 9,
+            offset_reserved: 11,
         }
         .to_bytes();
         assert_eq!(bytes[58], 9, "offset_frontier must occupy bytes 58..66");
+        assert_eq!(bytes[66], 11, "offset_reserved must occupy bytes 66..74");
         bytes[22] = 5; // log_view = 5, view stays 4
 
         assert_eq!(
--- a/core/integration/tests/cluster/crash_offset_reuse.rs
+++ b/core/integration/tests/cluster/crash_offset_reuse.rs
@@ -15,17 +15,17 @@
 // specific language governing permissions and limitations
 // under the License.
 
-//! RED SPEC, expected to FAIL: offset identity across a crash.
+//! Offset identity across a crash.
 //!
 //! `SendMessagesResponse::confirmations` hands clients concrete base offsets,
-//! which makes offset reuse client-visible: a client that recorded offset N
-//! for its message must never see the server confirm a DIFFERENT message at
-//! N later. A solo node acks below the flush thresholds from RAM only and
-//! persists no offset watermark, so after a SIGKILL it restarts the partition
-//! log at the last flushed position and re-mints offsets it already
-//! confirmed. Passes only once a durable watermark (or durable journal tail)
-//! keeps post-restart offsets above everything ever acked.
+//! which makes offset reuse client-visible: a client that recorded offset N for
+//! its message must never see the server confirm a DIFFERENT message at N
+//! later. A solo node acks below the flush thresholds from RAM only, so nothing
+//! in the segments says those offsets were ever handed out. What keeps them
+//! from being re-minted is the offset RESERVATION in the partition superblock,
+//! claimed by the append fence before any of them exist and read back by boot.
 
+use std::path::Path;
 use std::time::Duration;
 
 use iggy::prelude::*;
@@ -97,13 +97,8 @@ async fn wait_until_serving(harness: &TestHarness, budget: Duration) -> IggyClie
     }
 }
 
-// TODO(hubcio): fix this test
-#[ignore = "confirmed offsets re-minted after a crash; no durable offset watermark"]
-#[iggy_harness(cluster_nodes = 1)]
-async fn given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_should_not_remint_offsets(
-    harness: &mut TestHarness,
-) {
-    let client = harness.tcp_root_client().await.unwrap();
+/// Create the stream and its single-partition topic.
+async fn create_topic(client: &IggyClient, messages_required_to_save: Option<u32>) {
     client
         .create_stream(STREAM_NAME)
         .await
@@ -115,20 +110,83 @@ async fn given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_
             &TopicCreateOptions {
                 partitions_count: Some(1),
                 message_expiry: Some(IggyExpiry::NeverExpire),
+                messages_required_to_save,
                 ..TopicCreateOptions::default()
             },
         )
         .await
         .expect("create topic");
+}
+
+/// Base offsets of every segment file under `root`, from the file names, which
+/// are the on-disk claim about where each range begins.
+///
+/// Reading them is the only way to assert the shape the re-anchor produces. A
+/// black-box offset assertion passes either way on the boot that WRITES the
+/// wrong shape; the cost only lands on the boot that reads it back, where the
+/// walk refuses and the solo arm tombstones the partition.
+fn segment_base_offsets(root: &Path) -> Vec<u64> {
+    let mut offsets = Vec::new();
+    let mut stack = vec![root.to_path_buf()];
+    while let Some(dir) = stack.pop() {
+        let Ok(entries) = std::fs::read_dir(&dir) else {
+            continue;
+        };
+        for entry in entries.flatten() {
+            let path = entry.path();
+            if path.is_dir() {
+                stack.push(path);
+            } else if path.extension().is_some_and(|extension| extension == "log")
+                && let Some(stem) = path.file_stem().and_then(|stem| stem.to_str())
+                && let Ok(offset) = stem.parse::<u64>()
+            {
+                offsets.push(offset);
+            }
+        }
+    }
+    offsets.sort_unstable();
+    offsets
+}
+
+/// Poll until the only node has the replicated consumer offset on disk. Panics
+/// at the deadline.
+async fn wait_for_stored_offset_on_disk(harness: &TestHarness, expected: u64, budget: Duration) {
+    let data_path = harness.node(0).data_path();
+    let deadline = tokio::time::Instant::now() + budget;
+    loop {
+        if integration::harness::disk::read_replicated_consumer_offset(&data_path) == Some(expected)
+        {
+            return;
+        }
+        assert!(
+            tokio::time::Instant::now() < deadline,
+            "the node did not persist consumer offset {expected} within {budget:?} \
+             (found {:?})",
+            integration::harness::disk::read_replicated_consumer_offset(&data_path),
+        );
+        sleep(POLL_INTERVAL).await;
+    }
+}
+
+/// Kill the node, bring it back, and return a client onto the restarted one.
+async fn crash_and_recover(harness: &mut TestHarness) -> IggyClient {
+    harness.kill_node(0).expect("SIGKILL the only node");
+    harness.restart_node(0).expect("restart it");
+    wait_until_serving(harness, SERVE_TIMEOUT).await
+}
+
+#[iggy_harness(cluster_nodes = 1)]
+async fn given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_should_not_remint_offsets(
+    harness: &mut TestHarness,
+) {
+    let client = harness.tcp_root_client().await.unwrap();
+    create_topic(&client, None).await;
 
     let acked = produce_acked(&client, "pre-crash", PRE_CRASH_SENDS).await;
     let highest_confirmed = *acked.last().expect("confirmed sends");
     drop(client);
 
-    harness.kill_node(0).expect("SIGKILL the only node");
-    harness.restart_node(0).expect("restart it");
-
-    let client = wait_until_serving(harness, SERVE_TIMEOUT).await;
+    let client = crash_and_recover(harness).await;
     let post_crash_offset = produce_acked(&client, "post-crash", 1).await[0];
 
     assert!(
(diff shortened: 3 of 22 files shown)


core/partitions/src/iggy_partition.rs
```diff
@@ -910,6 +1032,25 @@ where
...
+    /// The reservation is there for the crash case, where they do not. Left
+    /// standing it would make every ordinary restart resume a lease block
+    /// higher and hole the offset space for nothing.
+    ///
+    /// Callers must have flushed FIRST, and must not call this when the flush
+    /// failed: the claim it makes is precisely that the flush succeeded.
+    #[allow(clippy::future_not_send)]
+    #[must_use = "the bool is the durability verdict; a failed collapse leaves a gap"]
+    pub async fn collapse_offset_reservation(&self) -> bool {
+        let frontier = self.offset_frontier();
```
numinnex: `offset_frontier()` is the append counter, so a boot mint floor that was never consumed is erased here.

Reproduced on this branch: crash, restart, clean stop, restart, produce — offset 4 was ACKed before the SIGKILL and the first send after the clean restart is confirmed at offset 0 again. The graceful stop is what destroys the protection; a second SIGKILL would have preserved it, so the runbook-correct incident response is the one action that defeats the fix. Nothing logs the moment.

Suggest:

```rust
let frontier = self
    .offset_frontier()
    .max(self.armed_mint_floor())
    .max(self.durable_offset_frontier.get());
```

The `armed_mint_floor` term covers the case above. The `durable_offset_frontier` term is a separate fix: `reset_offset_frontier_at` is the non-monotone writer, so in the failed-install window this currently regresses the recorded frontier.

core/partitions/src/iggy_partition.rs
```diff
@@ -933,6 +1074,80 @@ where
...
+    /// SOLO GROUPS ONLY. A backup validates an incoming prepare with
+    /// `base_offset == dirty_offset + 1`
+    /// ([`Self::append_received_send_messages_to_journal`]), so a primary
+    /// minting from a floor its peers do not share would have every one of them
+    /// refuse the batch; relaxing that check deserves its own evidence. A
+    /// replicated group is also less exposed -- an ack means a quorum journaled
+    /// the batch, so the hole there is a FULL-cluster crash.
+    fn mint_floor(&self) -> u64 {
+        let floor = self.armed_mint_floor();
+        self.mint_floor_pending.set(false);
```
numinnex: The flag is spent at mint time, but every failure path after the mint discards the mint itself: the `count == 0` return, the `checked_add`, `stamp_prepare_for_persistence`'s `?`, the reservation fence, and the journal append all return before `dirty_offset` is stored.

The next append then mints `dirty_offset + 1` — the pre-crash value — while `reserve_offsets_through` short-circuits on `durable_offset_reserved > end_offset` and writes nothing. So one transient superblock error on the first append after a crash-restart re-opens the exact defect this PR closes, silently.

Suggest disarming where `dirty_offset` is stored instead. That point is past every early return, and it also covers the `?` in `stamp_prepare_for_persistence`, which sits after the mint and is not covered today.

core/partitions/src/iggy_partition.rs
```diff
@@ -2666,6 +2895,19 @@ where
...
+        // Past this line the offsets are in the journal, hence committable,
+        // pollable, confirmable and forwardable, and no later gate can take
+        // them back. See [`Self::reserve_offsets_through`].
+        //
+        // LOCK ORDER: the caller holds `write_lock` and this takes
+        // `superblock_lock` under it. The install path takes them in the
+        // reverse order, safely only because `reset_offset_frontier_at` drops
+        // `superblock_lock` before `try_install` takes `write_lock`. Never hold
+        // `superblock_lock` across a `write_lock` acquire.
+        if !self.reserve_offsets_through(last_dirty_offset).await {
```
numinnex, line 3703: This two-fsync `atomic_replace` runs inline in the shard's sequential frame pump, and the consensus tick is a `select_biased!` arm on that same pump — its own comment bounds tick delay to one main frame body's longest `.await`. A slow superblock write therefore withholds heartbeats for every group on the shard, not only this partition, and peers can elect around the node. It also sits ahead of `send_prepare_ok` on the backup path, so it is on the primary's commit-latency path too.

The LOCK ORDER note above is correct about lock ordering but understates the blast radius.

Worth saying that the throughput cost is negligible and the arithmetic in the description checks out: roughly 3 fsyncs/s at 100k msg/s with the 64Ki lease, about 15 ns/message amortized. The concern is only where the stall lands. Extending the reservation off the critical path at a watermark, keeping this fence as the backstop, would preserve the guarantee without putting the write inside the pump.

core/partitions/src/iggy_partition.rs
```diff
@@ -4033,6 +4275,121 @@ where
...
+                %error,
+                "boot re-anchor could not fsync the partition dir after unlinking"
+            );
+        }
+        // Only a SIZED tail: an empty one either just went, or is already named
+        // at the frontier and can take the appends as it is.
+        let needs_plant = self.log.segments().last().is_some_and(|segment| {
+            segment.size.as_bytes_u64() > 0 && segment.end_offset + 1 < frontier
+        });
+        if !needs_plant {
```
numinnex: When the unlink loop empties the chain, `last()` is `None`, `needs_plant` is false, and this returns without planting. `ensure_initial_segment` then plants at `offset_frontier()` and knows nothing about `armed_mint_floor()`, so the first mint lands at the reservation inside a segment named for the frontier — the hole-inside-a-segment shape the re-anchor exists to prevent.

It is tolerated while the index survives, so it is invisible in the common case. With a torn index (`enforce_fsync = false` is the shipped default) the index-less walk hits `base_offset != expected_offset` and raises `OffsetDiscontinuity`, tombstoning the partition.

This is the shape `given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_should_not_remint_offsets` produces; its assertion only checks the offset moved forward, so it passes over the segment shape. `build_partition_fresh` has the same gap — it arms the floor via `set_superblock` but never re-anchors before `ensure_initial_segment`.

Suggest planting here when the chain is empty and `frontier > 0`, or passing the floor into `ensure_initial_segment`.

core/server/src/segment_recovery.rs
```diff
@@ -507,7 +513,18 @@ fn ensure_contiguous_chain(
...
+        // A gap whose far side sits inside `offset_reserved` is the boot
+        // re-anchor's, not damage: `reanchor_to_offset_frontier` seals the
+        // recovered tail and plants the next segment at a frontier the
+        // superblock already claimed, so the skipped offsets were minted (or
+        // may have been) and no file was ever meant to hold them. Refusing it
+        // would tombstone the partition on the boot AFTER the one that fixed
+        // the re-mint. The ceiling is what bounds the leniency: a stray file
+        // splicing a gap ABOVE anything this replica claimed still refuses.
+        if previous.end_offset.checked_add(1) != Some(next.start_offset)
+            && next.start_offset > offset_reserved
```
numinnex: `write_superblock_inner` clamps `offset_reserved` up to `offset_frontier`, and `advanced_frontier` floors the frontier at `held_offset_frontier()`. `offset_reserved` therefore sits at or above the end of every legitimate segment, which makes `next.start_offset > offset_reserved` false for any real gap — so `Hole` can no longer fire for a partition that has ever appended. A genuinely lost middle segment now recovers silently as a holed log instead of refusing loudly, and the admitted arm logs nothing.

Narrowing the predicate is harder than it first looks. The gaps accumulate, one per crash. Probed on this branch, segment base offsets across successive crash cycles:

```
[0]
[0, 65537]
[0, 65537, 131074]
[0, 65537, 131074, 196611]
```

So an "at most one gap" rule would tombstone a healthy partition by the second crash, and `given_a_crash_restarted_node_when_it_flushes_and_crashes_again_should_still_not_remint_offsets` already reaches the two-gap state. Pinning `next.start_offset == offset_reserved` is unstable for a different reason: the first append after the plant extends the reservation past it, so the same pair fails on the following boot.

A single monotone scalar may not be able to separate re-anchor gaps from damage across repeated cycles. Recording the anchor points durably, or making the planted segment self-describing (carrying its predecessor's end offset), would give the guard something stable to check against.

core/partitions/src/iggy_partition.rs
```diff
@@ -975,6 +1097,25 @@ where
...
+    /// The reservation is there for the crash case, where they do not. Left
+    /// standing it would make every ordinary restart resume a lease block
+    /// higher and hole the offset space for nothing.
+    ///
+    /// Callers must have flushed FIRST, and must not call this when the flush
+    /// failed: the claim it makes is precisely that the flush succeeded.
+    #[allow(clippy::future_not_send)]
+    #[must_use = "the bool is the durability verdict; a failed collapse leaves a gap"]
+    pub async fn collapse_offset_reservation(&self) -> bool {
+        let frontier = self.offset_frontier();
```
hubcio: after a crash-restart with no sends the counter is still below the armed floor, so a clean stop writes `(0,0)` or a reservation under the planted segment - next boot re-mints offset 0 or refuses the chain. collapse to `offset_frontier().max(armed_mint_floor())`.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
...
+    /// a segment from its FILENAME with a running `expected_offset`, so the next
+    /// boot truncates the post-hole suffix and the counter falls back below what
+    /// was already confirmed, undoing the fix on the second crash. On a segment
+    /// BOUNDARY every reader copes -- absolute offsets in the index,
+    /// `disk_poll_start` walking on into later segments, a contiguity guard that
+    /// only looks within one file.
+    ///
+    /// So an empty tail is unlinked (its name claims a range it does not hold)
+    /// and a sized tail, the only copy of its messages, is sealed with a fresh
+    /// segment planted at the frontier. An empty chain is left to the caller's
```
hubcio: `ensure_initial_segment` plants at `offset_frontier()`, still 0 here, while the first mint lands at the floor - the hole inside a segment this doc calls unsurvivable; the boot after the first flush refuses it. plant the empty chain at `frontier` here too.

core/server/src/partition_helpers.rs
```diff
@@ -653,6 +653,7 @@ pub async fn build_partition_fresh(
         config.partition.evicted_ring_capacity,
         config.partition.evicted_ring_bytes_max.as_bytes_u64(),
     );
+    partition.set_offset_reservation_lease(config.partition.offset_reservation_lease);
```
hubcio, line 1267: same hole on this path: `set_superblock` arms the floor from the surviving `offset_reserved`, `ensure_initial_segment` plants at the frontier, and the re-anchor never runs here. the first mint lands a lease block into a segment named below it, no crash needed.

core/partitions/src/iggy_partition.rs
```diff
@@ -1812,14 +2037,18 @@ where
...
                 .load(Ordering::Relaxed)
                 .checked_add(1)
                 .ok_or(IggyError::CannotAppendMessage)?
         } else {
             0
         };
+        // Only here: this is the only path that mints. A backup re-stamps what
+        // the primary sends (`append_received_send_messages_to_journal`) and
+        // must follow it exactly, so raising ITS counter would fork the group.
+        let dirty_offset = next.max(self.mint_floor());
```
hubcio: `mint_floor()` clears the flag before the fence or the journal append can fail, so the retry mints below the reservation and the fence fast path waves it through - a confirmed offset re-minted. clear the flag next to `dirty_offset.store` instead.

core/server/src/segment_recovery.rs
```diff
@@ -567,7 +575,18 @@ fn ensure_contiguous_chain(
...
```
hubcio: this also admits overlaps and missing middle segments, since every real start sits below the reservation - after a clean stop, the whole chain. only accept a forward gap on the last pair, and pass 0 for replicated groups.

core/partitions/src/iggy_partition.rs
```diff
@@ -2768,6 +2997,19 @@ where
...
+        // Past this line the offsets are in the journal, hence committable,
+        // pollable, confirmable and forwardable, and no later gate can take
+        // them back. See [`Self::reserve_offsets_through`].
+        //
+        // LOCK ORDER: the caller holds `write_lock` and this takes
+        // `superblock_lock` under it. The install path takes them in the
+        // reverse order, safely only because `reset_offset_frontier_at` drops
+        // `superblock_lock` before `try_install` takes `write_lock`. Never hold
+        // `superblock_lock` across a `write_lock` acquire.
+        if !self.reserve_offsets_through(last_dirty_offset).await {
```
hubcio, line 3703: on a solo primary a failed or backed-off write here makes `on_replicate` drop the op after the sequencer took it, so nothing commits again until restart and every produce times out. reserve in `on_request` before the pipeline push and reply transient.

core/partitions/src/iggy_partition.rs
```diff
@@ -873,6 +965,34 @@ where
...
+            .log
+            .segments()
+            .iter()
+            .filter(|segment| segment.size.as_bytes_u64() > 0)
+            .map(|segment| segment.end_offset.saturating_add(1))
+            .max()
+            .unwrap_or(0);
+        let journal = self.log.journal().info;
+        let resident = if journal.messages_count > 0 {
+            journal.current_offset.saturating_add(1)
```
hubcio: `current_offset` is the dirty tail, so the fence and view-change persists record uncommitted offsets as the durable frontier. after a view change truncates that tail, a restart seeds the counter above the group and this replica refuses every prepare until state transfer. read the committed offset.

core/partitions/src/state_transfer.rs
```diff
@@ -1994,7 +1994,13 @@ where
...
             || (self.applied_purge_generation < committed_purge_generation
                 && offsets_wire.next_offset == 0);
-        let local_next_offset = self.offset_frontier();
+        // What this replica HOLDS, not the append counter: the two diverge once
+        // the counter is seeded from an offset reservation, which stands a lease
+        // block above the last message anywhere and names none of them. Against
+        // that, every legitimate offer inside the block reads as a rewind, and
+        // the replica that restarted with a reservation is exactly the one
+        // needing the transfer, so it would cycle refusal -> backoff forever.
+        let local_next_offset = self.held_offset_frontier();
```
hubcio: an empty chain installed at frontier N has `held_offset_frontier() == 0`, so the guard is skipped and a stale offer below N rewinds the counter. max it with `installed_frontier`.

core/partitions/src/state_transfer.rs
```diff
@@ -1994,7 +1994,13 @@ where
         let purge_advances = offsets_wire.purge_generation > committed_purge_generation
             || (self.applied_purge_generation < committed_purge_generation
                 && offsets_wire.next_offset == 0);
-        let local_next_offset = self.offset_frontier();
+        // What this replica HOLDS, not the append counter: the two diverge once
+        // the counter is seeded from an offset reservation, which stands a lease
```
hubcio: same claim at iggy_partition.rs:734. the counter is never seeded from the reservation - boot restores the frontier only and the floor bites at the first mint - so this reasoning does not hold.

core/consensus/src/vsr_state.rs
```diff
@@ -121,11 +136,15 @@ impl TryFrom<&[u8]> for VsrState {
...
-        // cannot fail.
+        // Length-tolerant for ONE legacy layout: a pre-`offset_frontier` record
+        // pads out and its trailing fields read as 0, which the read sites
+        // filter. The one length check then puts every field slice below in
+        // bounds by construction, so the `try_into`s cannot fail.
+        //
+        // `offset_reserved` gets no tolerance of its own: a silent 0 reads as
+        // "nothing reserved" and re-mints confirmed offsets, the exact defect
+        // the field closes, so a frontier-but-no-reservation record is refused.
+        // Pre-production, so an older data directory is wiped, not migrated.
```
hubcio: server-0.9.0-edge.2 through edge.6 wrote 66-byte records on both planes, so this refuses boot node-wide after an upgrade, with an error naming a 58-byte layout no build ever wrote. accept 66 with zero-fill and drop the dead 58 arm.

core/consensus/src/vsr_state.rs
```diff
@@ -31,8 +31,8 @@ use std::fmt;
 /// Number of bytes [`VsrState::to_bytes`] produces: `cluster`(16) +
 /// `replica_id`(1) + `replica_count`(1) + `view`(4) + `log_view`(4) +
 /// `commit_max`(8) + `checkpoint_op`(8) + `checkpoint_checksum`(16) +
-/// `offset_frontier`(8).
-pub const ENCODED_LEN: usize = 66;
+/// `offset_frontier`(8) + `offset_reserved`(8).
+pub const ENCODED_LEN: usize = 74;
```
hubcio, line 40: a 74-byte record is `WrongLength` on the previous builds too, so rollback needs a wipe as well - worth saying, or bump the record version.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1139,80 @@ where
...
+    /// re-stamp take. Claiming through `end_offset + 1 + lease` rather than from
+    /// the live counter needs no special case for an oversized batch.
+    ///
+    /// `false` only when a write was attempted and failed, and the caller must
+    /// refuse the append. Fail-closed: the send is rejected with nothing
+    /// externalised, or a backup withholds its `PrepareOk` and the group elects
+    /// around it, exactly as a failed view persist does.
+    #[allow(clippy::future_not_send)]
+    #[must_use = "the bool is the fence verdict; dropping it lets the append escape unreserved"]
+    pub async fn reserve_offsets_through(&self, end_offset: u64) -> bool {
```
hubcio, line 1558: nothing on a replicated group benefits from the reservation (the floor and the re-anchor are solo-only), yet every replica pays this write per block and the chain guard reads it to loosen itself. return early when `replica_count() > 1`.

core/server/config.toml
```diff
@@ -973,6 +973,14 @@ clients_table_max = 8192
 # u128 bitset, and this depth bounds that suffix.
 prepare_queue_depth = 32
 
+# How many offsets a partition claims in its superblock ahead of the mint
+# counter before it will append, so a crash-restarted replica resumes above
+# every offset it confirmed to a client instead of re-minting it for a different
+# message. One superblock write (two fsyncs) per block: lowering it raises the
+# fsync rate on the write path, raising it wastes at most one block of the u64
```
hubcio, line 1013: same in core/configs/src/server_config/partition.rs:133. this protection only applies to single-replica groups; say so here.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
...
+                plane = "partitions",
+                namespace_raw = namespace.inner(),
+                start_offset = segment.start_offset,
+                offset_frontier = frontier,
+                "unlinked an empty segment named below the restored offset frontier"
+            );
+            retired += 1;
+        }
+        // Durable before anything is planted beside them: a crash in between
+        // would boot the stale name back into the chain. No stats decrement to
```
hubcio: boot does count the recovered chain - `load_persisted_segments` increments per segment, empty tails included - so each retire here leaves `segments_count` one too high on the wire. decrement per retire, like retention does.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
...
+                match compio::fs::remove_file(&path).await {
+                    Ok(()) => {}
+                    Err(error) if error.kind() == std::io::ErrorKind::NotFound => {}
+                    Err(error) => {
+                        warn!(
+                            target: "iggy.partitions.diag",
+                            plane = "partitions",
+                            namespace_raw = namespace.inner(),
+                            path = %path,
+                            %error,
```
hubcio, line 5236: planting after a failed unlink leaves `[tail][stale empty][R]` on disk, which the next boot refuses as an empty non-tail segment. return the error instead (it aborts the node's boot, not a retry).

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1139,80 @@ where
...
+        if self.superblock_write_is_backed_off() {
+            return false;
+        }
+        let _superblock_guard = self.superblock_lock.acquire().await;
+        // A batch queued behind another append's write finds the block already
+        // extended.
+        if self.durable_offset_reserved.get() > end_offset {
+            return true;
+        }
+        let claim = end_offset
```
hubcio: two fsyncs awaited inline under `write_lock` inside the shard pump, so every partition on the core stalls once per block - the first fsync on the default append path. claim the next block off the append path, about half a block early.

core/shard/src/lib.rs
```diff
@@ -6762,6 +6762,20 @@ where
...
                 partition.fence_flush_failure();
+                // The collapse below claims the segments account for every
+                // confirmed offset, which a failed flush is exactly the case
+                // against, so leave the reservation standing.
+                continue;
+            }
+            // The segments now prove where the offset space ends, so the
+            // reservation has nothing left to witness. Without the collapse
+            // every clean stop would leave a lease-block-wide hole.
+            if !partition.collapse_offset_reservation().await {
```
hubcio: one superblock write per active partition, serially, on every graceful stop. fan these out with `join_all` - the lock and the store are per partition.

core/integration/tests/cluster/crash_offset_reuse.rs
```diff
@@ -97,8 +97,64 @@ async fn wait_until_serving(harness: &TestHarness, budget: Duration) -> IggyClie
...
+}
+
+/// Kill the node, bring it back, and return a client onto the restarted one.
+async fn crash_and_recover(harness: &mut TestHarness) -> IggyClient {
+    harness.kill_node(0).expect("SIGKILL the only node");
+    harness.restart_node(0).expect("restart it");
+    wait_until_serving(harness, SERVE_TIMEOUT).await
+}
+
 #[iggy_harness(cluster_nodes = 1)]
```
hubcio, line 178: this re-inlines `create_topic` and `crash_and_recover` from above.

core/integration/tests/cluster/crash_offset_reuse.rs
```diff
@@ -97,8 +97,64 @@ async fn wait_until_serving(harness: &TestHarness, budget: Duration) -> IggyClie
...
 async fn given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_should_not_remint_offsets(
```
hubcio, line 179: this builds the empty-chain shape but restarts only once, so it never sees the boot that refuses it. add a send, a clean restart and a third boot, and assert `segment_base_offsets`.

core/integration/tests/cluster/crash_offset_reuse.rs
```diff
@@ -141,3 +197,91 @@ async fn given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_
...
     );
 }
+
+/// The fix has to survive its own side effect: the hole the reservation leaves
+/// between the recovered segments and the new append point truncates everything
+/// past it on the next boot if it lands INSIDE a segment.
+///
+/// So the SECOND crash is the one that matters, and only if the run between the
+/// two reaches disk, which is what the flush threshold is for.
+#[iggy_harness(cluster_nodes = 1)]
```
hubcio, line 471: none of these stops the node cleanly between crashes, and all three are single-node - the shutdown collapse and the replicated fence path stay untested. add a crash, restart, clean stop, restart, send case and a 3-node one.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
         Ok(())
     }
 
+    /// Re-anchor the append point after boot re-seeded the offset counter above
+    /// what the recovered segment chain holds.
+    ///
+    /// A hole INSIDE a segment is not survivable: `recover_segment_bounds` walks
+    /// a segment from its FILENAME with a running `expected_offset`, so the next
+    /// boot truncates the post-hole suffix and the counter falls back below what
+    /// was already confirmed, undoing the fix on the second crash. On a segment
```
hubcio: also at bootstrap.rs:2808 and crash_offset_reuse.rs:202, 252. neither recovery walk truncates a hole any more - both refuse with `OffsetDiscontinuity` and the solo arm tombstones, which is worse than this says.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
...
+                namespace_raw = namespace.inner(),
+                partition_dir,
+                %error,
+                "boot re-anchor could not fsync the partition dir after unlinking"
+            );
+        }
+        // Only a SIZED tail: an empty one either just went, or is already named
+        // at the frontier and can take the appends as it is.
+        let needs_plant = self.log.segments().last().is_some_and(|segment| {
+            segment.size.as_bytes_u64() > 0 && segment.end_offset + 1 < frontier
```
hubcio: `saturating_add(1)`, like the sibling at 984.

core/configs/src/server_config/partition.rs
```diff
@@ -177,6 +188,16 @@ impl Validatable<ConfigurationError> for PartitionConfig {
             );
             return Err(ConfigurationError::InvalidConfigurationValue);
         }
+        if self.offset_reservation_lease == 0
```
hubcio: no zero or above-ceiling test for this knob; the `prepare_queue_depth` and `evicted_ring` ones have both.

core/partitions/src/log.rs
```diff
@@ -314,6 +314,28 @@ where
         Some((segment, storage))
     }
 
+    /// Retire the NEWEST segment, the mirror of [`Self::retire_front`].
+    ///
+    /// Boot re-anchor only, and only for an EMPTY tail named below the re-seeded
+    /// counter, which would otherwise claim a range it does not hold. Nothing
+    /// else may take from the back: a sized tail is the only copy of its
+    /// messages.
+    pub fn retire_back(&mut self) -> Option<(Segment, SegmentStorage)> {
```
hubcio, line 324: the `segments_mut` doc at 194 lists `add_persisted_segment` and `retire_front` as the only length mutators; add this one.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1139,80 @@ where
...
+    /// is client-visible long before the threshold-gated flush names it in a
+    /// segment, and a crash in between hands a second message an offset a client
+    /// already holds. Fencing here rather than at commit puts it upstream of
+    /// every way an offset escapes -- the reply, the poll tier, the peers a
+    /// prepare reaches -- on the one path both a primary's mint and a backup's
+    /// re-stamp take. Claiming through `end_offset + 1 + lease` rather than from
+    /// the live counter needs no special case for an oversized batch.
+    ///
+    /// `false` only when a write was attempted and failed, and the caller must
+    /// refuse the append. Fail-closed: the send is rejected with nothing
```
hubcio: the backoff arm returns false without attempting a write, and on the primary the append error is swallowed in `on_replicate`, not rejected. reword once the fence moves.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1139,80 @@ where
...
+    /// The append fence: make sure the durable record already permits every
+    /// offset up to and including `end_offset` before the caller lets them
+    /// exist.
+    ///
+    /// `SendMessagesResponse` hands clients concrete base offsets and the poll
+    /// path serves committed messages out of the resident journal, so an offset
+    /// is client-visible long before the threshold-gated flush names it in a
+    /// segment, and a crash in between hands a second message an offset a client
+    /// already holds. Fencing here rather than at commit puts it upstream of
+    /// every way an offset escapes -- the reply, the poll tier, the peers a
```
hubcio: `append_repaired_send_messages` journals without this fence, so 'every way an offset escapes' overstates it; qualify the claim.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
...
+        }
+        let sealed_index = self.log.segments().len() - 1;
+        let sealed_end = self.log.active_segment().end_offset;
+        self.log.active_segment_mut().sealed = true;
+        let sealed_storage = &mut self.log.storages_mut()[sealed_index];
+        let _ = sealed_storage.shutdown();
+        self.log.messages_writers_mut()[sealed_index] = None;
+        self.log.index_writers_mut()[sealed_index] = None;
+        self.log.indexes_mut()[sealed_index] = None;
+        self.install_empty_segment(config, frontier).await?;
```
hubcio: `purge` fsyncs the dir after its plant and says why; this one does not. a line on why the floor makes it unnecessary would save the next reader the question.

core/partitions/src/iggy_partition.rs
```diff
@@ -752,14 +837,21 @@ where
...
+    /// because four paths leave a replica whose counter would otherwise
+    /// restart below where the group already is (a transfer install of an
+    /// all-GC'd origin, a crash inside the install's swap window, the
+    /// fence-and-rebuild path, which needs no crash at all, and a crash while
+    /// acked messages were still resident in the journal). Restarting the
     /// counter is not a lag -- replicas re-stamp `base_offset` from it and
     /// recompute `batch_checksum` over the result, so the next replicated
     /// prepare would persist different bytes here than on every peer, silently.
     ///
+    /// The RESERVATION is deliberately not folded in here: a backup mints
```
hubcio: seed the solo counter from `max(offset_frontier, offset_reserved)` here (gated on `replica_count() == 1`): `offset_frontier()` is then right for the re-anchor, `ensure_initial_segment` and the collapse at once, and the mint-floor state machine goes away. keep `held_offset_frontier` bytes-based so a later 1 -> N topology change does not inherit an inflated frontier.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,6 +4431,121 @@ where
...
+        let needs_plant = self.log.segments().last().is_some_and(|segment| {
+            segment.size.as_bytes_u64() > 0 && segment.end_offset + 1 < frontier
+        });
+        if !needs_plant {
+            return Ok(());
+        }
+        let sealed_index = self.log.segments().len() - 1;
+        let sealed_end = self.log.active_segment().end_offset;
+        self.log.active_segment_mut().sealed = true;
+        let sealed_storage = &mut self.log.storages_mut()[sealed_index];
```
hubcio: this seal-and-plant is `rotate_segment` minus `reset_read_state`. a `rotate_segment_at(config, start)` shared by both keeps one seal path and its create-then-teardown order.

core/partitions/src/iggy_partition.rs
```diff
@@ -1103,6 +1318,16 @@ where
             .unwrap_or(config.messages_required_to_save)
     }
 
+    /// Install the offset-reservation block size resolved from this node's
+    /// `PartitionsConfig`.
+    ///
+    /// Carried on the partition because the fence runs inside `on_request` /
+    /// `on_replicate`, which take no config. Floored at 1: a zero block reserves
+    /// nothing and would write the superblock before every append.
+    pub const fn set_offset_reservation_lease(&mut self, lease: u32) {
```
hubcio: `u64::from(lease)` and drop `const` - the cast only exists to dodge `cast_lossless`. keep the zero floor, unit tests and the simulator set this directly.

core/server/src/segment_recovery.rs
```diff
@@ -3018,6 +3051,52 @@ mod tests {
         assert_eq!(bytes_of(&next_index_path), next_index);
     }
 
+    #[compio::test]
```
hubcio, line 3176: this pins the over-broad rule (any hole below the reservation); once the guard only admits the last pair's forward gap, this test needs the tighter shape, plus an overlap case and a mid-chain gap case.

hubcio (changes requested):
the crash shapes reproduce on this branch with curl against a solo node: a crash below the flush threshold then a clean restart re-mints offset 0 or tombstones the partition, and a crash, one send, and a clean restart tombstones on the next boot. inline comments have the details.

outside the diff: `init_partition` in core/shard/src/lib.rs:3636 arms the floor on every simulator restart with no re-anchor, so the simulator diverges from the server here. the poll ceiling at core/partitions/src/iggy_partition.rs:1788-1806 serves an uncommitted offset 0 before the first commit - pre-existing, but it is how the spent-floor re-mint reaches consumers. the primary apply error at iggy_partition.rs:2676-2688 has no `rollback_pipelined_prepare` on the partitions plane - pre-existing, this fence is the new trigger. the doc on `persist_offset_frontier_at` at iggy_partition.rs:1122 still describes the install use it lost.

core/server/src/bootstrap.rs
```diff
@@ -2601,7 +2618,6 @@ async fn load_partition(
         .max()
         .filter(|&start| sized_end.is_none() && start > 0);
     let current_offset = sized_end.or_else(|| empty_frontier.map(|start| start - 1));
```
hubcio: an empty chain's name used to be a committed frontier; the re-anchor and `ensure_initial_segment` now plant it at `mint_frontier()`, which is a reservation. two crashes below the flush threshold then store committed offset 65537 for a partition holding nothing, and `store_consumer_offset` admits the whole hole. bound the promotion by `recovered_state.offset_frontier`, which a real install writes at the group frontier and a re-anchor plant leaves far below.

core/partitions/src/iggy_partition.rs
```diff
@@ -752,36 +829,78 @@ where
...
             namespace_raw = self.consensus().group(),
             offset_frontier = frontier,
-            "restored partition offset frontier from its superblock"
+            offset_reserved = reserved,
+            append_point,
+            "restored partition offset counters from its superblock"
         );
-        self.offset.store(recovered_end, Ordering::Release);
-        self.dirty_offset.store(recovered_end, Ordering::Relaxed);
         self.should_increment_offset = true;
```
hubcio: `should_increment_offset` now means the append counter is live, but `offset_frontier()` still reads it as the committed counter naming data, so it returns 1 for a partition holding nothing. a separate bit for the committed seed stops that; it does not help the boot path above, which rebuilds from the file name.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,41 +4613,245 @@ where
...
+        // No directory means an in-memory partition, whose chain no boot reads.
+        let Some(partition_dir) = self.partition_dir() else {
+            return Ok(());
+        };
+        let anchor = crate::segment_anchor::SegmentAnchor {
+            sealed_start,
+            sealed_end,
+        };
+        if let Err(error) =
+            crate::segment_anchor::write_anchor(&partition_dir, frontier, anchor).await
```
hubcio: anchors have no lifecycle - every unlink site omits `.anchor`, including the quarantine and converge sweeps that filter a directory listing to `.log`/`.index`/`.staging`. purge resets the offset space to 0 and leaves the anchor there, where two of `covers`'s three checks match for free and only the far-side offset has to coincide. add `ANCHOR_EXTENSION` to those sets.

core/partitions/src/segment_anchor.rs
```diff
@@ -0,0 +1,205 @@
...
+    crate::state_transfer::fsync_dir(partition_dir).await
+}
+
+/// Read the anchor beside the segment starting at `start_offset`.
+///
+/// `None` when the file is absent or does not decode; both mean the same thing
+/// to the guard, so the caller needs no distinction.
+pub async fn read_anchor(partition_dir: &str, start_offset: u64) -> Option<SegmentAnchor> {
+    let path = anchor_path(partition_dir, start_offset);
+    let bytes = compio::fs::read(&path).await.ok()?;
```
hubcio: `.ok()?` turns an `EACCES` or `EIO` into no gap intended, so a healthy planted chain is refused as `Hole` - a solo group goes dark for the life of the process and every boot while the error lasts, a replicated one re-transfers the whole partition over one bad read. `file_len` fail-stops on exactly this class; do the same and treat `NotFound` separately.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1208,158 @@ where
...
+        }
+        // A frontier: offsets strictly below it are permitted, so covering
+        // `end_offset` needs a record strictly above it.
+        if self.durable_offset_reserved.get() > end_offset {
+            return true;
+        }
+        let Some(superblock) = self.superblock.as_ref().map(Rc::clone) else {
+            return true;
+        };
+        if self.superblock_write_is_backed_off() {
```
hubcio: one failed superblock write by any writer on this partition - a purge frontier reset, the tick's own extension - opens a 20 ms window, up to 1 s after repeats, where an already-admitted send whose mint crosses the lease boundary sets `fatal` and takes the node down. rare at the default lease, routine at a small one. give it the grace `superblock_wedged` grants.

core/shard/src/lib.rs
```diff
@@ -6554,6 +6573,26 @@ where
                 }
                 continue;
             }
+            // Same bound the metadata plane fail-stops on, applied per group. A
+            // partition whose superblock keeps refusing withholds every
+            // view-scoped send AND refuses every append, so it serves nothing
+            // while the process still reports healthy.
+            let superblock_failures = partition.superblock_write_failures();
+            if superblock_wedged(
```
hubcio, line 6913: one unwritable partition dir now exits the node and crashloops on restart, while `iggy_partition.rs:692` still says only that group is fenced and the rest keeps serving. `refuses every append` is also wrong above one replica - `reserve_offsets_through` returns true there - and the timeout's operator docs still scope it to the metadata superblock.

core/partitions/src/iggy_partition.rs
```diff
@@ -975,6 +1145,46 @@ where
...
+    #[must_use = "the bool is the durability verdict; a failed collapse leaves a gap"]
+    pub async fn collapse_offset_reservation(&self) -> bool {
+        let append_point = self.mint_frontier();
+        if self.durable_offset_reserved.get() <= append_point {
+            return true;
+        }
+        let Some(superblock) = self.superblock.as_ref().map(Rc::clone) else {
+            return true;
+        };
+        if self.superblock_write_is_backed_off() {
```
hubcio: a superblock error just before a clean stop leaves the reservation standing, so the next boot seeds the append counter a lease block above the data and holes the offset space for nothing - what the doc at :1152 warns about. `record_frontier_before_quarantine` skips the backoff gate for exactly this reason; do the same here.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,41 +4613,245 @@ where
...
-    /// be the one that erases the data proving the old frontier. The caller
-    /// RETRIES; it must not fence, since the chain is still whole and the live
-    /// counter still names the pre-purge space.
+    /// [`IggyError`] when the fresh segment cannot be created, leaving the
+    /// partition without a serviceable chain.
     #[allow(clippy::future_not_send)]
-    async fn record_purge_frontier_reset(&mut self, generation: u64) -> Result<(), PurgeError> {
-        if self.reset_offset_frontier_at(0).await {
-            self.purge_deferred = false;
+    pub async fn reanchor_to_offset_frontier(
```
hubcio, line 5193: a data dir written by this branch's earlier push has no anchor, so this build boots it as `Hole` and the solo arm tombstones it. master plants no gaps and edge builds come only from master, so nothing deployed regresses - but anyone who ran the earlier push needs a line in the PR description telling them to wipe.

core/partitions/src/iggy_partition.rs
```diff
@@ -4189,41 +4613,245 @@ where
...
+            );
+            // Boot DOES count the recovered chain -- `load_persisted_segments`
+            // increments per segment before it looks at the size, so empty tails
+            // are in the total -- and retention pairs its own retire with a
+            // decrement. Without this the count stays one high on the wire for
+            // the life of the process.
+            self.stats.decrement_segments_count(1);
+            retired += 1;
+        }
+        // Durable before anything is planted beside them: a crash in between
```
hubcio, line 5260: the comment promises durability before anything is planted, but the fsync error is only logged. the sized-tail arm is saved by `write_anchor`'s own directory fsync; the emptied-chain arm goes through `install_empty_segment`, which has neither - so the promise holds by accident.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1208,158 @@ where
...
+            || self.superblock.is_none()
+            || !self.should_increment_offset
+        {
+            return false;
+        }
+        let headroom = self
+            .durable_offset_reserved
+            .get()
+            .saturating_sub(self.mint_frontier());
+        headroom < self.offset_reservation_lease / 2
```
hubcio: `lease / 2` is 0 when the lease is 1, which validation allows, so the extension never fires and every append pays the inline claim - two fsyncs on every multi-message batch. use `(lease / 2).max(1)`, or floor the config at 2.

core/partitions/src/iggy_partition.rs
```diff
@@ -1812,14 +2184,18 @@ where
...
                 .load(Ordering::Relaxed)
                 .checked_add(1)
                 .ok_or(IggyError::CannotAppendMessage)?
         } else {
             0
         };
+        // Only here: this is the only path that mints. A backup re-stamps what
+        // the primary sends (`append_received_send_messages_to_journal`) and
+        // must follow it exactly, so raising ITS counter would fork the group.
+        let dirty_offset = next;
```
hubcio: dead rebinding left over from `next.max(self.mint_floor())`. rename the binding above back to `dirty_offset` and drop this line.

core/partitions/src/iggy_partition.rs
```diff
@@ -3846,98 +4319,49 @@ where
...
-                .map_err(|_| IggyError::CannotCreateSegmentIndexFile(index_path.clone()))?,
-        );
+    /// Seal the active segment and plant a fresh empty one at `start_offset`.
+    ///
+    /// Shared by the size-driven roll, which plants at `end_offset + 1`, and the
+    /// boot re-anchor, which plants at the append point the reservation moved the
+    /// counter to. One seal path, and one order: the new segment's files are
+    /// created BEFORE the sealed segment's writers are torn down, so a failed
+    /// create leaves the chain serviceable.
+    async fn rotate_segment_at(
```
hubcio, line 4860: nothing here ties `start_offset` to the sealed tail or to an anchor having been written. both callers are correct today, but the obligation lives only in prose at :4754 - a `debug_assert!` would stop a third caller inheriting it silently.

core/partitions/src/segment_anchor.rs
```diff
@@ -0,0 +1,205 @@
...
+///
+/// Any I/O failure. The caller must NOT plant the segment: a planted segment
+/// whose anchor is missing reads as damage on the next boot.
+pub async fn write_anchor(
+    partition_dir: &str,
+    start_offset: u64,
+    anchor: SegmentAnchor,
+) -> io::Result<()> {
+    let path = anchor_path(partition_dir, start_offset);
+    let mut file = compio::fs::File::create(&path).await?;
```
hubcio: `File::create` truncates in place with no second slot, unlike the superblock in this same directory (tmp, `sync_all`, rename, dir fsync). unreachable today because the boot sweep drops any anchor whose `.log` is missing, and a live `.log` cannot be re-selected by `sealed_end + 1 < frontier`.

core/partitions/src/iggy_partition.rs
```diff
@@ -677,22 +692,78 @@ where
...
+    /// lease block above them after a reservation-seeded boot and names none of
+    /// them.
+    fn advanced_frontier(&self, claim: u64) -> u64 {
+        claim
+            .max(self.held_offset_frontier())
+            .max(self.durable_offset_frontier.get())
+    }
+
+    /// Record an incoming state-transfer frontier, advancing the frontier but
+    /// SETTING the reservation to it.
```
hubcio: not what happens - `write_superblock_inner` clamps the reservation to `max(offset_frontier)`, so it lands on `advanced_frontier(frontier)`, above the offer whenever an earlier over-claiming write left the durable frontier higher. say to the frontier this write records.

core/consensus/src/vsr_state.rs
```diff
@@ -31,20 +31,26 @@ use std::fmt;
 /// Number of bytes [`VsrState::to_bytes`] produces: `cluster`(16) +
 /// `replica_id`(1) + `replica_count`(1) + `view`(4) + `log_view`(4) +
 /// `commit_max`(8) + `checkpoint_op`(8) + `checkpoint_checksum`(16) +
-/// `offset_frontier`(8).
-pub const ENCODED_LEN: usize = 66;
+/// `offset_frontier`(8) + `offset_reserved`(8).
+///
+/// Growing this is one-way: a record of this length is [`VsrStateError::WrongLength`]
```
hubcio, line 36: a rollback needing the data directory wiped is an operator instruction living only in rustdoc on a const. put it where someone planning a downgrade would look.

core/consensus/src/vsr_state.rs
```diff
@@ -121,16 +142,26 @@ impl TryFrom<&[u8]> for VsrState {
...
+        // out of RAM above the frontier are gone with the process either way --
+        // but refusing the record recovers nothing and costs the node its boot.
         let mut padded = [0u8; ENCODED_LEN];
         match bytes.len() {
             ENCODED_LEN => padded.copy_from_slice(bytes),
-            ENCODED_LEN_WITHOUT_FRONTIER => {
-                padded[..ENCODED_LEN_WITHOUT_FRONTIER].copy_from_slice(bytes);
+            ENCODED_LEN_WITHOUT_RESERVATION => {
+                padded[..ENCODED_LEN_WITHOUT_RESERVATION].copy_from_slice(bytes);
+                padded[66..74].copy_from_slice(&bytes[58..66]);
```
hubcio: the `66` here is a literal two lines after `ENCODED_LEN_WITHOUT_RESERVATION` names it. `padded[ENCODED_LEN_WITHOUT_RESERVATION..ENCODED_LEN]` is the whole fix - the `58` has no constant and numeric offsets are this file's convention.

core/partitions/src/state_transfer.rs
```diff
@@ -2044,7 +2051,7 @@ where
             self.reset_offset_frontier_at(offsets_wire.next_offset)
                 .await
         } else {
-            self.persist_offset_frontier_at(offsets_wire.next_offset)
+            self.install_offset_frontier_at(offsets_wire.next_offset)
```
hubcio, line 2315: does this reset buy anything today? the fence never runs above one replica, so on every group that can receive an offer the reservation already equals the frontier and `advanced_frontier(offer)` is never below it - the write is identical to the advancing form's.

core/partitions/src/iggy_partition.rs
```diff
@@ -954,7 +1124,7 @@ where
         let _superblock_guard = self.superblock_lock.acquire().await;
         match intended {
             Some(frontier) => {
-                self.write_superblock_inner(superblock.as_ref(), frontier)
+                self.write_superblock_inner(superblock.as_ref(), frontier, frontier)
```
hubcio: this writes `(f, f)` with no max against `durable_offset_reserved`, which contradicts the monotone ceiling `vsr_state.rs:106` claims. inert today - the only `Some` caller is the replicated `ConvergeFailed` arm, where the fence never ran and `reserved == frontier`. carry the max anyway?

core/configs/src/server_config/partition.rs
```diff
@@ -177,6 +191,16 @@ impl Validatable<ConfigurationError> for PartitionConfig {
...
```
hubcio: the validation landed but `mod tests` has no zero or above-ceiling case for this knob, while `prepare_queue_depth` and `evicted_ring_capacity` each keep theirs.

core/integration/tests/cluster/crash_offset_reuse.rs
```diff
@@ -115,20 +110,63 @@ async fn given_confirmed_sends_below_flush_threshold_when_a_solo_node_is_killed_
...
 
+/// Kill the node, bring it back, and return a client onto the restarted one.
+async fn crash_and_recover(harness: &mut TestHarness) -> IggyClient {
     harness.kill_node(0).expect("SIGKILL the only node");
     harness.restart_node(0).expect("restart it");
+    wait_until_serving(harness, SERVE_TIMEOUT).await
+}
 
-    let client = wait_until_serving(harness, SERVE_TIMEOUT).await;
+#[iggy_harness(cluster_nodes = 1)]
```
hubcio, line 178: no test combines a flush with a graceful stop - `:220` stops cleanly but takes no traffic between lives, and `:299` flushes then re-enters through SIGKILL. `ensure_initial_segment` also never runs with a non-zero frontier: the only superblock fixture hardcodes `offset_reserved: 0`.

hubcio (changes requested):

core/partitions/src/iggy_partition.rs
```diff
@@ -2132,6 +2517,40 @@ where
...
+                // retry instead of costing the process its life.
+                //
+                // `TransientNotAccepted`, per its contract: nothing was admitted,
+                // so the client may re-issue anywhere without double-apply risk.
+                // It does make the SDK recheck the leader and walk the roster,
+                // which finds no better node when the fault is this one's disk --
+                // wasteful, but the weaker code would claim an unknown outcome
+                // for a request that provably has none.
+                if message.header().operation == Operation::SendMessages
+                    && let Some(ceiling) = self.request_mint_ceiling(&message)
```
hubcio: this never fires on an ordinary send - the convert above passes `ChecksumMode::Skip`, which zeroes `batch_checksum`, so `request_mint_ceiling`'s verifying decode returns `None`. only the encrypt re-entry reaches the fence, so with encryption off every fence failure lands at the mint instead, where `on_replicate` takes the node down.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1208,158 @@ where
...
+    /// covers every batch in flight unless a run of them crosses a block
+    /// boundary.
+    ///
+    /// `None` when the body is not one canonical batch, which
+    /// `convert_request_message` has already rejected by the time this runs.
+    fn request_mint_ceiling(&self, message: &Message<RoutedRequestHeader>) -> Option<u64> {
+        let body = message
+            .as_slice()
+            .get(std::mem::size_of::<RoutedRequestHeader>()..message.header().size as usize)?;
+        let count = decode_batch_slice(body).ok()?.message_count();
```
hubcio: `decode_batch_slice` verifies, so this hashes every message body again just to read `message_count`, a `u32` sitting in the batch header - a second full pass per send on top of the one `admit_wire_request` already did. use `BatchHeader::decode(body)`, and check `replica_count()` before any of it.

hubcio (changes requested):

core/server/src/bootstrap.rs
krishvishal: Fixed. `empty_frontier` now takes `min(recovered_state.offset_frontier)`, so an uncorroborated plant no longer seeds the committed counter.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. Split into `OffsetSpace { append_live, committed_seeded }`; `offset_frontier()` reads the committed bit, which `restore_offset_frontier` only sets when `frontier > 0`.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. Anchors unlink with their segment at all three retire sites, and `ANCHOR_SUFFIX` joined the quarantine and converge sweeps.

core/partitions/src/segment_anchor.rs
krishvishal: Fixed. `read_anchor` returns `io::Result<Option<_>>`, `NotFound` is `Ok(None)`, and boot fail-stops on anything else like `file_len`.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. `reserve_offsets_through` bypasses the backoff gate; `superblock_wedged` stays the gate that decides a run of failures is terminal.

core/shard/src/lib.rs
krishvishal: Fixed. Comment now says the bound exits the node, scopes "refuses every append" to solo, and the timeout's operator text covers both planes.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. `collapse_offset_reservation` bypasses the backoff gate, same rationale as `record_frontier_before_quarantine`.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. PR description gained a wipe instruction for earlier-push data dirs; the upgrade paragraph above it was backwards and is corrected too.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. The re-anchor's directory fsync failure returns `CannotSyncFile` instead of warning, so the promise no longer rests on `write_anchor`.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. `(lease / 2).max(1)`, so the knob's documented range stays as it is.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. Binding renamed back to `dirty_offset`, rebinding dropped.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. `debug_assert!(start_offset > sealed_end)`, plus a debug-only read-back asserting the anchor `covers` the sealed bounds when the gap is wider.

core/partitions/src/segment_anchor.rs
krishvishal: Fixed. Temp, fsync, rename, fsync the dir, like the superblock; the temp uses `STAGING_SUFFIX` so every existing sweep already unlinks it.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. Doc now says the reservation lands on the frontier this write records, and names the clamp that puts it there.

core/consensus/src/vsr_state.rs
krishvishal: Fixed. Moved to a `DOWNGRADE:` note beside `offset_reservation_lease` in `config.toml` and into the PR description; the rustdoc points at it.

core/consensus/src/vsr_state.rs
krishvishal: Fixed. `padded[ENCODED_LEN_WITHOUT_RESERVATION..ENCODED_LEN]`; `bytes[58..66]` left numeric per the file.

core/partitions/src/state_transfer.rs
krishvishal: No, it buys nothing today, exactly as you describe. Kept and documented: the branch is keyed on `purge_advances`, where the reset is the right shape rather than a coincidence.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. The `Some` arm writes `frontier.max(durable_offset_reserved)`, so the monotone ceiling holds here too.

core/configs/src/server_config/partition.rs
krishvishal: Fixed. Zero, above-ceiling and at-ceiling cases added, mirroring the other two knobs.

core/integration/tests/cluster/crash_offset_reuse.rs
krishvishal: Fixed. Added a flush-then-graceful-stop cluster case here, and `ensure_initial_segment` over a non-zero reservation in `partition_helpers`.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed, and this was the one that mattered. `request_mint_ceiling` reads the batch header instead of the verifying decode, so the fence fires on ordinary sends again; covered by a test that pins the premise.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. `BatchHeader::decode` for `message_count`, and an early `replica_count() > 1` return before the body is touched.

core/configs/src/server_config/partition.rs
hubcio: still no zero, above-ceiling or at-ceiling case for this knob in `mod tests` - the reply says they were added, but ccf5adcde does not touch this file.

core/partitions/src/iggy_partition.rs
```diff
@@ -998,6 +1262,169 @@ where
...
+    /// shapes the chain around it never runs there either. A replicated group
+    /// paying a superblock write per block would buy nothing -- and an ack there
+    /// already means a quorum journaled the batch, so re-minting needs a
+    /// FULL-cluster crash.
+    ///
+    /// `false` when the write was attempted and failed. Fail-closed: the send is
+    /// rejected with nothing externalised, exactly as a failed view persist
+    /// withholds its sends.
+    ///
+    /// BYPASSES the retry backoff, like `record_frontier_before_quarantine` and
```
hubcio, line 1532: `superblock_wedged` assumes one failure per backoff step (`2m` / 1s cap = 120), and now every refused send adds one - a few producers retrying through a short disk fault exit the node in seconds. count a failure only outside the backoff window, or make the wedge wall-clock.

core/partitions/src/iggy_partition.rs
```diff
@@ -1564,7 +2001,7 @@ where
         // commit). Also used below as the poll's high-water bound: this function
         // is fully synchronous, so the single load cannot drift mid-plan.
         let commit_offset = self.offsets().commit_offset;
-        if !self.should_increment_offset || args.count == 0 {
+        if !self.offset_space.append_live || args.count == 0 {
```
hubcio: `append_live` is the wrong bit here - the comment above says this reads the committed frontier, and on a reservation-only boot it is live with `offset` at 0 naming nothing. gate on `committed_seeded`.

hubcio (changes requested):
round 3 against ccf5adcde (head b10db896c): 21 of the 22 threads check out, the lease knob tests are the one reply that is not in the push, and the backoff bypass in `reserve_offsets_through` needs one more change - inline. out of the diff: `should_increment_offset` is still named in the `build_partition_fresh` doc (partition_helpers.rs:528) and the shard restore comment (shard/lib.rs:3677), and the install unlink at state_transfer.rs:2172 still drops only `.log` and `.index` - unreachable for an anchor today since anchors are solo-only, so hygiene.

core/configs/src/server_config/partition.rs
krishvishal: You are right, and my reply was wrong: the file was written but never made it into ccf5adcde. Re-added, three cases, mirroring the other two knobs. Not pushed yet, so it will show up in the next push rather than on this head.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed, and good catch: the bypass turned the wedge count into a function of producer retry rate. `write_superblock_inner` now samples the backoff window before the write and, on a failure inside it, neither counts nor extends. So the count is one per backoff step again, which is what `superblock_wedged` reads as elapsed time, and a bypassing writer can no longer both starve the gated writers and pull the exit forward. Lands in the next push.

core/partitions/src/iggy_partition.rs
krishvishal: Fixed. Gated on `committed_seeded`, which is the bit the comment above it describes. Lands in the next push.

krishvishal:
The three out-of-diff items from the review body are fixed too:

- `partition_helpers.rs:528` and `shard/lib.rs:3677` no longer name `should_increment_offset`; they say `OffsetSpace` / `OffsetSpace::append_live`.
- `state_transfer.rs:2172` chains `anchor_cleanup_path` after `.log` and `.index`, with a comment recording that it is unreachable today because anchors are solo-only.

All six round-3 items are in the working tree, not in a push yet, so this head still shows none of them. Verification after the six: `fmt`, `sort --no-format`, `clippy --all-features --all-targets -D warnings` clean; 1005 unit tests across partitions, configs, server, shard and consensus; `cluster::crash_offset_reuse` 6/6 on a rebuilt server.

numinnex (dismissed):

numinnex (dismissed):

hubcio (dismissed):

hubcio (approved):

mmodzelewski (approved):

PR #3976, opened 2026-08-27
fix(journal): repair a torn WAL tail instead of panicking the shard

numinnex:
`FileStorage::truncate` is the boot-time repair for a torn metadata WAL tail.
It used compio's `set_len`, which submits `IORING_OP_FTRUNCATE`. That opcode
landed in kernel 6.9. Below it the driver probes the opcode as unsupported and
falls back to `push_blocking`, but shard proactors are built with
`thread_pool_limit(0)`, so the fallback panics with `the thread pool is needed
but no worker thread is running`. The panic fires inside dispatch, outside
`catch_unwind_io`.

Net effect: a crash that tore the last WAL append made the next boot kill the
shard instead of repairing it, and the repair is idempotent, so the node stayed
down across restarts.

Two conditions have to coincide, so this is not a routine path:

kernel < 6.9          IORING_OP_FTRUNCATE absent, compio falls back
  AND
torn WAL tail         crash mid-append, so boot calls truncate_or_fail

The affected range is 5.19 through 6.8, not everything below 6.9. Ring setup
already requires IORING_SETUP_COOP_TASKRUN and IORING_SETUP_TASKRUN_FLAG,
which need 5.19, so RHEL 9 (5.14) and stock Ubuntu 22.04 (5.15) never start the
server at all and were never exposed. What this actually broke is Debian 12 and
AL2023 (6.1), and Ubuntu 24.04 and 22.04-HWE (6.8). macOS aarch64 is exempt
because create_shard_executor keeps a blocking pool there by design.

The fix

Truncate synchronously through std::fs on the stored path, which needs
neither the opcode nor the blocking pool. This mirrors what segment recovery
already does in truncate_to. The sync_all moves inside truncate, so the
repair is durable on its own and the caller no longer pairs it with a separate
fsync. sync_all rather than sync_data because the file length is metadata,
and without it a power cut right after the repair re-presents the torn tail.

truncate is no longer async. An async fn that never awaits trips
clippy::unused_async, and the journal crate denies clippy::pedantic.
Blocking the shard thread costs nothing here: the sole caller is boot-time
repair, before the shard serves traffic.


spetz (dismissed):

mmodzelewski (dismissed):

helm/charts/iggy/README.md.gotmpl
```diff
@@ -33,8 +33,16 @@ under the License.
...
+1. **Required io_uring setup features** - The node must provide both flags:
+
+   * `IORING_SETUP_COOP_TASKRUN`
+   * `IORING_SETUP_TASKRUN_FLAG`
+
+   Mainline Linux provides both starting in 5.19. Vendor kernels may backport them to older kernel versions.
+
+   The server fails during startup if either is unavailable. Node capabilities matter, not the container image.
+2. **IPC_LOCK capability** - For locking memory required by io_uring
+3. **Unconfined seccomp profile** - To allow io_uring syscalls
```
hubcio: iggy requires kernel 6.11, please check compio io uring syscalls

hubcio (changes requested):

spetz (dismissed):

core/message_bus/src/client_listener/mod.rs
```diff
@@ -108,6 +108,11 @@ pub mod wss;
 /// `TIME_WAIT`); `SO_REUSEPORT` is intentionally not set: only shard 0
 /// binds the client listeners (see each caller).
 ///
+/// On Linux, compio submits this path's bind and listen as
+/// `IORING_OP_BIND` and `IORING_OP_LISTEN`. Mainline Linux added both in
+/// 6.11. Shard 0 has no blocking fallback pool, so an older kernel without
```
hubcio: the 6.11 floor is only this helper - tls, wss, replica and quic bind sync through socket2 (`bind_reusable_tcp_listener`). route this one the same way, keeping `set_tcp_nodelay(true)` on the listener since plain ws and http set none per connection - floor back at 5.19, and the 6.11 text in diagnostics and the helm readme should not land.

please verify it :)

core/server_common/src/diagnostics.rs
```diff
@@ -300,9 +315,11 @@ fn report_io_uring_environment() {
 
         if let Some((major, minor)) = parse_kernel_version(&release) {
             kernel_version = Some((major, minor));
-            if (major, minor) < (MIN_KERNEL_MAJOR, MIN_KERNEL_MINOR) {
+            if (major, minor) < (MAINLINE_MIN_KERNEL_MAJOR, MAINLINE_MIN_KERNEL_MINOR) {
```
hubcio: this used to give a definite answer for kernels below 5.19, the only cause of EINVAL here. now everything below 6.11 gets a hedge about vendor backports, and the bind/listen text lands on a path where those opcodes were never submitted.

core/journal/src/file_storage.rs
```diff
@@ -57,18 +58,26 @@ impl FileStorage {
...
-    // synchronous `std::fs` truncate here (boot-time path) or gate on a probe.
-    pub async fn truncate(&self, len: u64) -> io::Result<()> {
-        let file = unsafe { &*self.file.get() };
-        file.set_len(len).await?;
+    /// Returns an I/O error if the file cannot be opened, truncated, or synced.
+    pub fn truncate(&self, len: u64) -> io::Result<()> {
+        let file = fs::OpenOptions::new().write(true).open(&self.path)?;
+        file.set_len(len)?;
+        file.sync_all()?;
         self.write_offset.set(len);
```
hubcio, line 84: set `write_offset` right after `set_len`, before `sync_all` - a failed fsync now leaves it at the old length while the file is already shorter.

core/journal/src/file_storage.rs
```diff
@@ -57,18 +58,26 @@ impl FileStorage {
...
-    // TODO(hubcio): compio `set_len` submits IORING_OP_FTRUNCATE, which kernels
-    // below 6.9 do not support; the driver then falls back to its blocking
-    // pool, and shard proactors run with `thread_pool_limit(0)`, so the torn
-    // WAL repair panics the shard on such kernels instead of repairing. Use a
-    // synchronous `std::fs` truncate here (boot-time path) or gate on a probe.
-    pub async fn truncate(&self, len: u64) -> io::Result<()> {
-        let file = unsafe { &*self.file.get() };
-        file.set_len(len).await?;
+    /// Returns an I/O error if the file cannot be opened, truncated, or synced.
+    pub fn truncate(&self, len: u64) -> io::Result<()> {
```
hubcio: `pub(crate)` - the only caller is in-crate, and this now blocks the shard thread, so keep it off the public surface.

core/journal/src/file_storage.rs
```diff
@@ -57,18 +58,26 @@ impl FileStorage {
...
+        let file = fs::OpenOptions::new().write(true).open(&self.path)?;
```
hubcio: `fs::File::from(file.as_fd().try_clone_to_owned()?)` truncates the same open file without going back through the path - `compio::fs::File` is `AsFd`. keeps the fd and `path` from ever disagreeing.

core/journal/src/file_storage.rs
```diff
@@ -178,3 +187,55 @@ impl Storage for FileStorage {
...
+mod tests {
+    use super::FileStorage;
+    use server_common::executor::create_shard_executor;
+    use tempfile::tempdir;
+
+    /// Torn-tail repair runs on a shard executor, which builds its proactor
+    /// with `thread_pool_limit(0)`. Any truncate that reaches compio's
+    /// blocking pool panics that shard rather than repairing the WAL, which
+    /// is what mainline Linux did before 6.9 while `set_len` was an `io_uring`
+    /// submission. Driving the repair through a real shard executor is the only
```
hubcio: this can't fail on any kernel with `IORING_OP_FTRUNCATE` (6.9+), and ci has it - a revert to compio `set_len` would still pass. say what it does pin: a sync signature and no pool on a modern kernel.

core/journal/src/file_storage.rs
```diff
@@ -178,3 +187,55 @@ impl Storage for FileStorage {
...
+            assert_eq!(std::fs::metadata(&path).unwrap().len(), 64);
+        });
+    }
+
+    /// The repair has to survive the crash it is repairing from: a reopen
+    /// that still saw the torn tail would walk and truncate it again on every
+    /// boot, and a reopen that saw a longer file would resurrect the bytes
+    /// recovery just proved dead.
+    #[test]
+    fn given_a_truncated_file_when_reopened_should_see_the_shortened_length() {
```
hubcio: this passes with `sync_all` deleted - a same-process reopen reads the page cache, so it can't see durability. it also repeats `truncate_or_fail_durably_repairs_torn_tail`, drop it.

hubcio (changes requested):
the body says this fixes debian 12 / al2023 (6.1) and ubuntu 24.04 (6.8). it doesn't on those kernels: recovery runs before the client listeners bind, and plain tcp/ws/http bind through `IORING_OP_BIND`/`LISTEN` (6.11) with the same no-pool panic, so the crash just moves from `recover` to listener startup. only tls/wss/quic setups with http off gain today. rewrite the body before squash, and drop the `sync_all` over `sync_data` line - fdatasync already flushes a size change.

spetz (dismissed):

spetz (approved):

hubcio (approved):

PR #3980, opened 2026-08-27
ci(python): fold the build task into test and check the stub

### LINKED ISSUE #3977: CI: merge the Python SDK's three tasks into one job
### Description

`sdk-python` declares `tasks: ["lint", "test", "build"]`, and each task becomes its own job on its own runner. The `build` task compiles the wheel and throws it away: its `dist/` never reaches `Upload test artifacts`, which only runs under `test`. So the crate is compiled twice per push and nothing is shared between the jobs.

At the same time `apache_iggy.pyi` has no freshness gate. The file header says it is generated by pyo3_stub_gen, the README says "nothing in CI checks stub freshness", and the stub has already drifted from the source: `describe_options` sits in a different position with its return type written as `builtins.list[OptionSpec]`, and two docstrings were edited by hand. Checking this needs a built crate, so today there is nowhere cheap to put the check.

### Affected area / component

CI / build / tooling

### Proposed solution

Drop the `build` task and move `maturin build -o dist` into `test`:

```yaml
tasks: ["lint", "test"]
```

- `lint` = ruff + cargo fmt + clippy + pyrefly, unchanged
- `test` = wheel + stub check + pytest

Keeping `lint` separate preserves the per-task status in the checks list, and `fail-fast: true` on the SDK matrices already cancels `test` when `lint` fails, so nothing needs coordinating.

Then add the stub gate to `test`, right after the wheel is built:

```bash
cargo run --bin stub_gen
uv run --no-sync ruff format apache_iggy.pyi
uv run --no-sync ruff check --fix apache_iggy.pyi
git diff --exit-code -- apache_iggy.pyi
```

The ruff steps are required, not cosmetic: the tracked stub is stub_gen output *after* ruff, since `.pre-commit-config.yaml` runs `ruff-check --fix` and `ruff-format` over `foreign/python/**/*.pyi`. Format has to run before check, because the raw output leaves whitespace on blank lines, which `check` reports but only `format` fixes.

`stub_gen` costs a full rebuild in a lint-only job, because clippy leaves nothing behind, but about 10s after `maturin build` since only pyo3 and above rebuild when the extension-module feature drops out.

### Alternatives considered

_No response_

### Contribution

### Good first issue

ethanlin01x:

Closes #3977

## Rationale

Each task in `components.yml` becomes its own job, so `sdk-python` compiled the same crate twice per push. Folding `build` into `test` also gives `apache_iggy.pyi` somewhere cheap to be checked, which it has never had.

## What changed?

1. `build` is folded into `test`. `maturin build -o dist` runs there, ahead of the coverage build so pytest still gets the instrumented module. `lint` stays its own job.
2. `test` now checks that `apache_iggy.pyi` is in sync: regenerate, normalise with ruff, diff. The stub had already drifted, so the first commit is the regeneration.
3. `stub_gen` is anchored on `CARGO_MANIFEST_DIR`. Running it from a subdirectory used to strip the license header off the tracked stub, which the README carried as a known bug. That TODO is gone.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? Claude
2. Scope of usage? help implement and review this PR
3. How did you verify the generated code works correctly? Unit tests pin each rejected value and each still-legal zero


jiengup:
LGTM

ethanlin01x:
I think the Go e2e failure is a flake, unrelated to this PR.

slbotbm (approved):

hubcio (approved):

numinnex (approved):

PR #3987, opened 2026-08-28
fix(cluster): recover writes after a fast partition primary rejoin

spetz:
A fast partition primary rejoin could leave the group unable to
commit new writes. A backup could adopt suffix headers without
obtaining their bodies because repair stopped at commit_max. Clients
then handled the typed non-admission response by checking metadata
leadership, even though metadata and partition groups can elect
different primaries.

Extend suffix repair to the available frontier and clear completed
sessions after the later commit catches up. Reject writes on stale
backups before local offset validation, so callers receive the typed
response required for safe rerouting.

Add one-pass roster routing for HTTP and the Rust, Node, Go, Java,
and C# clients. Coordinate concurrent Rust reconnects and prevent
ambiguous replicated writes from being replayed under a new session.

### DIFF
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -1974,6 +1974,30 @@ where
                 message
             };
 
+            // State-dependent admission belongs to the primary. A backup can
+            // lag the committed offset table or message frontier and would
+            // otherwise turn a routing artifact into a terminal 404 or 400.
+            // Reject it first with the only response that proves the request
+            // was never admitted, so the caller may safely retry elsewhere.
+            if consensus.is_follower() || !consensus.is_normal() || consensus.is_transferring() {
+                emit_partition_diag(
+                    tracing::Level::WARN,
+                    &PartitionDiagEvent::new(
+                        ReplicaLogContext::from_consensus(consensus, PlaneKind::Partitions),
+                        "rejecting client request on non-primary partition replica",
+                    )
+                    .with_operation(message.header().operation),
+                );
+                Self::send_partition_deny_or_log(
+                    consensus,
+                    message.header(),
+                    IggyError::TransientNotAccepted.as_code(),
+                    "non-primary transient reply send failed",
+                )
+                .await;
+                return;
+            }
+
             // Parse once for both the delete-existence check and AckLevel dispatch.
             let consumer_offset = match message.header().operation {
                 Operation::StoreConsumerOffset | Operation::DeleteConsumerOffset => {
@@ -2062,32 +2086,6 @@ where
                     .await;
                     return;
                 }
-            }
-
-            // A client op landing on a non-primary (or mid-view-change)
-            // replica is a routing artifact -- e.g. the roster still points
-            // here while this group's primaryship moved after a restart.
-            // Answer the typed transient instead of asserting: the SDK
-            // replays and its leader recheck re-routes, whereas a panic
-            // kills the shard and a silent drop wedges the client until its
-            // read timeout.
-            if consensus.is_follower() || !consensus.is_normal() || consensus.is_transferring() {
-                emit_partition_diag(
-                    tracing::Level::WARN,
-                    &PartitionDiagEvent::new(
-                        ReplicaLogContext::from_consensus(consensus, PlaneKind::Partitions),
-                        "rejecting client request on non-primary partition replica",
-                    )
-                    .with_operation(message.header().operation),
-                );
-                Self::send_partition_deny_or_log(
-                    consensus,
-                    message.header(),
-                    IggyError::TransientNotAccepted.as_code(),
-                    "non-primary transient reply send failed",
-                )
-                .await;
-                return;
             }
 
             // NoAck -> fast path. Quorum -> VSR pipeline.
@@ -5476,13 +5474,20 @@ mod tests {
     type SentFrames = Rc<RefCell<Vec<(u128, Frozen<MESSAGE_ALIGN>)>>>;
 
     fn recording_partition() -> (IggyPartition<RecordingBus>, SentFrames) {
+        recording_partition_at(0, 1)
+    }
+
+    fn recording_partition_at(
+        replica: u8,
+        replica_count: u8,
+    ) -> (IggyPartition<RecordingBus>, SentFrames) {
         let namespace = IggyNamespace::new(1, 1, 0);
         let bus = RecordingBus::default();
         let sent_to_clients = bus.sent_to_clients.clone();
         let consensus = VsrConsensus::new(
             TEST_CLUSTER,
-            0,
-            1,
+            replica,
+            replica_count,
             namespace.inner(),
             bus,
             LocalPipeline::new(),
@@ -5582,6 +5587,32 @@ mod tests {
             partition.consensus().pipeline_len(),
             1,
             "existing offset delete must replicate"
+        );
+    }
+
+    #[compio::test]
+    async fn on_request_delete_on_stale_backup_replies_transient_before_not_found() {
+        let (mut partition, sent_to_clients) = recording_partition_at(1, 3);
+        let client_id = 42;
+
+        partition
+            .on_request(delete_offset_request(client_id, 7, 5))
+            .await;
+
+        let sent = sent_to_clients.borrow();
+        assert_eq!(sent.len(), 1, "exactly one routing denial");
+        let (reply_client, frame) = &sent[0];
+        assert_eq!(*reply_client, client_id);
+        let header = bytemuck::checked::try_from_bytes::<ReplyHeader>(
+            &frame.as_slice()[..std::mem::size_of::<ReplyHeader>()],
+        )
+        .expect("deny frame starts with a valid reply header");
+        assert_eq!(header.status, IggyError::TransientNotAccepted.as_code());
+        assert_eq!(header.op, 0, "the backup admitted nothing");
+        assert_eq!(
+            partition.consensus().pipeline_len(),
+            0,
+            "a backup must not replicate the delete"
         );
     }
 
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -4463,10 +4463,29 @@ where
     /// commit walk runs at `RepairDone`, after the floor is known.
     pub async fn apply_repaired_prepare(&mut self, message: Message<PrepareHeader>) {
         let header = *message.header();
-        let Some(session) = &self.repair else {
+        let Some(session) = self.repair else {
             return;
         };
-        if header.op <= self.consensus().commit_min() || header.op > session.to_op {
+        let consensus = self.consensus();
+        if !consensus.is_normal() || consensus.view() != session.view {
+            self.repair = None;
+            return;
+        }
+        if header.op <= consensus.commit_min() || header.op > session.fetch_to_op {
+            return;
+        }
+        let canonical_checksum = consensus
+            .with_pending_view_log(|pending| {
+                pending
+                    .headers
+                    .iter()
+                    .find(|expected| expected.op == header.op)
+                    .map(|expected| expected.checksum)
+            })
+            .flatten();
+        if canonical_checksum.is_some_and(|expected| expected != header.checksum)
+            || (header.op > session.commit_to_op && canonical_checksum.is_none())
+        {
             return;
         }
         // Any in-window frame proves the stream is alive; only silence
@@ -4480,7 +4499,9 @@ where
         let applied = if header.operation == Operation::SendMessages {
             match self.append_repaired_send_messages(message).await {
                 Ok(base_offset) => {
-                    if let (Some(base_offset), Some(session)) = (base_offset, self.repair.as_mut())
+                    if header.op <= session.commit_to_op
+                        && let (Some(base_offset), Some(session)) =
+                            (base_offset, self.repair.as_mut())
                     {
                         session.first_batch_offset = Some(
                             session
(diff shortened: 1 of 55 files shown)


core/sdk/src/tcp/tcp_client.rs
```diff
@@ -248,11 +263,24 @@ impl BinaryTransport for TcpClient {
...
         // replaying, so that replay is the protocol rather than a retry.
-        let replay_after_reconnect = replay_is_safe(code, &error);
+        let replay_after_reconnect = replay_after_session_reset_is_safe(code, &error);
 
+        let skip_auto_login = is_login_register_code(code);
+        let nested_connect = skip_auto_login && self.connect_coordinator.is_active();
+        let _routing_guard = if nested_connect {
+            None
+        } else {
+            Some(self.routing_lock.lock().await)
```
numinnex, line 271: `_routing_guard` is a named binding, so it lives to the end of the function, and both replay sites below (`:280` and `:316`) call `self.send_raw(...)` while it is still held. `send_raw` re-acquires the same lock at `:1265` when the server answers `TransientNotAccepted`. Same task, non-reentrant `tokio::Mutex`, no timeout on `lock()`.

Reachable on this PR's own scenario: `SEND_MESSAGES` fails with `NotConnected`, `replay_after_session_reset_is_safe` returns true, the reconnect lands on the restarted node which is now a partition backup, the backup answers `TransientNotAccepted`, and the second acquire never returns. No error, no log, and the request timeout does not apply. The guard is never released, so every later refused request on the same client blocks behind it.

QUIC and WebSocket are immune: their guard is scoped to the walk block, and their `send_raw` never takes it.

Dropping the guard before the replay, or lifting the walk out of `send_raw` the way QUIC/WS do, would close this.

core/sdk/src/tcp/tcp_client.rs
```diff
@@ -248,11 +263,24 @@ impl BinaryTransport for TcpClient {
         // Login and register are the exception: the server stays deliberately
         // silent on a transient register failure and relies on the client
         // replaying, so that replay is the protocol rather than a retry.
-        let replay_after_reconnect = replay_is_safe(code, &error);
+        let replay_after_reconnect = replay_after_session_reset_is_safe(code, &error);
 
+        let skip_auto_login = is_login_register_code(code);
+        let nested_connect = skip_auto_login && self.connect_coordinator.is_active();
```
numinnex: `nested_connect` is decided from a global flag with no task identity, so it cannot distinguish the owner's own sign-in from an unrelated caller.

Any task calling the public `login_user()` while another task owns the connect takes this branch: it skips `routing_lock`, skips the coordinator, runs `disconnect_transport()` at `:282` unconditionally (tearing down the owner's in-flight stream), then calls `connect_inner()` at `:298` concurrently with the owner's. Two connect sequences race `stream`, `state` and `current_server_address`, which is what `ConnectCoordinator` was added in this PR to prevent.

Same shape at `quic_client.rs:251` and `websocket_client.rs:241`.

An explicit ownership token threaded down the connect chain would also fix the `settle_off_leader_once` leak: waiters in `ConnectCoordinator::run` return the owner's result without running the closure, so `connect_inner`'s `swap(false)` never runs and the flag leaks into a later unrelated reconnect.

core/shard/src/lib.rs
```diff
@@ -7433,16 +7456,33 @@ where
...
+        let missing_suffix = consensus.commit_max().checked_add(1).is_some_and(|first| {
+            (first..=head).any(|op| partition.log.journal().inner.header_by_op(op).is_none())
+        });
+        if !commit_lag && !missing_suffix {
             return;
         }
         let nonce = iggy_common::random_id::get_uuid();
         let from_op = consensus.commit_min() + 1;
-        let to_op = consensus.commit_max();
+        let to_op = if missing_suffix {
```
numinnex: Widening `to_op` to the sequencer head means repair now carries ops above `commit_max`, but the partition ingest arm is fence-free precisely because that could not happen before.

`apply_repaired_prepare` (`core/partitions/src/iggy_partition.rs:4464`) gates only on `commit_min < op <= session.to_op` plus `verify_prepare_integrity`, which is frame self-consistency. No view check, no merged-log comparison, no peer check, and `RepairPrepareHeader` carries no nonce so the ingest could not filter by session even if it wanted to. The metadata arm has all of these (`repair_op_in_scope` plus the checksum `disagrees` test); the comment at `:4589-4591` states the gap outright.

The session also survives a view change (`maybe_request_partition_repair` no-ops while `repair.is_some()`), so a late or duplicated `RepairPrepare` from the previous window can refill a truncated op with a non-canonical entry. The primary's retransmit is then rejected by `journaled_prepare_matches_retransmit` and never acked, and once `commit_max` passes that op the divergent bytes are flushed to segment storage. `reconcile_partition_view_divergence` runs only at StartView adoption, never at repair ingest.

Two further consequences of the same line. `repaired_window_is_complete` can never turn true against a peer with an empty memory-only journal, so the `FloorRefused` to state-transfer escape stops firing exactly when a fast rejoin needs it. And `to_op = head` is view-scoped rather than monotone, so after a view change discards the suffix `commit_min >= session.to_op` is unreachable and the new clear at `:6521` never fires.

Bounding the completeness and clear verdict at `commit_max` while widening only the fetch would address all three.

core/shard/src/lib.rs
```diff
@@ -7433,16 +7456,33 @@ where
...
+        // suffix HEADERS above `commit_max` whose bodies it may never have
+        // received: its ack for them is withheld until the body is journaled,
+        // and the primary's retransmit is dropped by the backup gap check
+        // because adoption already advanced the sequencer to the head. With a
+        // commit-bounded window nothing ever delivers those bodies, the
+        // primary cannot gather quorum for the suffix, and the group wedges
+        // one op below its head with the client write never confirmed.
+        let head = consensus.sequencer().current_sequence();
+        let commit_lag = consensus.commit_min() < consensus.commit_max();
+        let missing_suffix = consensus.commit_max().checked_add(1).is_some_and(|first| {
```
numinnex: `header_by_op` is an unindexed `headers.iter().find` (`core/partitions/src/journal.rs:760`), so this is O(window x resident headers), synchronous, on the shard pump. `journal.rs:768-775` documents this exact op-by-op shape as what can "turn one rejoin into an election storm", which is why `repaired_window_shape` exists.

The cheap `commit_min() >= commit_max()` early-out that used to guard this function now sits below the scan at `:7476`, so a caught-up backup that is merely pipelining pays the full walk where it previously returned on three boolean loads. Reached from `on_commit` / `CommitOutcome::Advanced`, so roughly once per `commit_broadcast_interval` per partition group per backup.

`!partition.log.journal().inner.repaired_window_shape(commit_max, head).complete` (`journal.rs:779`) is a single pass and is semantically identical, empty window included.

core/sdk/src/quic/quic_client.rs
```diff
@@ -129,7 +152,60 @@ impl BinaryTransport for QuicClient {
...
+                        break;
+                    };
+                    self.settle_on_endpoint(next).await?;
+                } else {
+                    let target = self.current_server_address.lock().await.clone();
+                    if let Some(walk) = roster_walk.as_mut() {
+                        walk.record_attempt(&target);
+                    }
+                }
+                self.connect().await?;
```
numinnex: `self.connect().await?` propagates `CannotEstablishConnection` out of `send_raw_with_response`, so the first undialable roster hop aborts the whole walk. `settle_on_endpoint(next).await?` a few lines above has the same escape.

TCP handles it instead: `tcp_client.rs:1310` swallows `Err(IggyError::CannotEstablishConnection)`, pulls the next `RosterWalk::next`, and returns `TransientNotAccepted` only once the walk is exhausted.

This is guaranteed to bite in the scenario the PR targets, since a stopped or restarting primary is exactly a roster entry that will not dial. QUIC and WebSocket then never reach the partition primary, and the caller sees a connect error instead of the typed refusal.

Related: this loop has no overall deadline, where TCP bounds its walk by `overall_deadline`. Each hop mints a fresh `RESPONSE_READ_TIMEOUT` plus a full connect retry ladder.

core/sdk/src/websocket/websocket_client.rs
```diff
@@ -122,7 +145,60 @@ impl BinaryTransport for WebSocketClient {
...
+                        break;
+                    };
+                    self.settle_on_endpoint(next).await?;
+                } else {
+                    let target = self.current_server_address.lock().await.clone();
+                    if let Some(walk) = roster_walk.as_mut() {
+                        walk.record_attempt(&target);
+                    }
+                }
+                self.connect().await?;
```
numinnex: Same as `quic_client.rs:200`. `self.connect().await?`, and `settle_on_endpoint(next).await?` above it, abort the roster walk on the first undialable hop, where TCP skips and continues at `tcp_client.rs:1310`. A stopped primary in the roster is this PR's own scenario, so the walk never reaches the partition primary.

Also unbounded in wall time: no caller deadline is threaded through the loop.

core/sdk/src/quic/quic_client.rs
```diff
@@ -129,7 +147,136 @@ impl BinaryTransport for QuicClient {
...
+                    needs_settle = true;
+                }
+                result = match tokio::time::timeout_at(
+                    overall_deadline,
+                    self.send_raw(code, payload.clone()),
+                )
+                .await
+                {
+                    Ok(result) => result,
+                    Err(_) => Err(IggyError::TransientNotAccepted),
```
numinnex: This maps an elapsed outer timer to `TransientNotAccepted`, which is the one code the PR defines as proof the request never entered a pipeline and is therefore safe to re-issue anywhere. At this point the frame has been written and the reply is unread, so the outcome is unknown.

`overall_deadline` is pinned at fn entry (`:150`) while `send_raw` mints a fresh `RESPONSE_READ_TIMEOUT` per call (`:921`). On the ordinary refusal path the inner 2s window returns first, so this is benign. On a **stalled reply** the inner budget is a fresh 30s while the outer has `30 - elapsed`, so from hop 2 onward the outer binds and fires exactly when the request is in flight, possibly admitted, replicated and committed.

The fabricated code is then consumed by the SDK's own walk loop, not just the app: it satisfies `while matches!(result, Err(TransientNotAccepted))` at `:178`, so the walk hops on, `connect_off_leader` mints a new VSR client id, and `:271` re-issues the same payload under a session the server's dedup fence cannot match. The app-boundary contract breach via error 58 is second-order.

Pre-fix this was a bare `send_raw(...).await` that surfaced `Disconnected` on expiry, which the comment at `:915-919` documents as deliberate ("later commits would commit it twice"). The outer timeout added for the walk-deadline fix silently overrides it.

`TransientNotCommitted` or `Disconnected` is the correct mapping for an elapsed outer timer, and either also falsifies the loop condition and stops the hop chain, which is the right answer for an unknown outcome. Same shape at `:180`. The sibling `Err(_)` arms at `:168`, `:200`, `:232`, `:258` are fine and should be left alone — nothing is on the wire there.

core/sdk/src/websocket/websocket_client.rs
```diff
@@ -122,7 +140,136 @@ impl BinaryTransport for WebSocketClient {
...
+                    needs_settle = true;
+                }
+                result = match tokio::time::timeout_at(
+                    overall_deadline,
+                    self.send_raw(code, payload.clone()),
+                )
+                .await
+                {
+                    Ok(result) => result,
+                    Err(_) => Err(IggyError::TransientNotAccepted),
```
numinnex: Same defect as `quic_client.rs:275` (and `:173` here): an elapsed outer timer is mapped to `TransientNotAccepted`, the code that licenses re-issue anywhere, while the request is on the wire with the reply unread. It feeds the walk condition at `:179`, so the SDK itself re-issues the payload on the next hop under a new session the dedup fence cannot match.

WebSocket is the worse of the two: `send_raw`'s body is a `tokio::spawn`, so cancelling the outer timeout drops the `JoinHandle` and **detaches** the task, which runs the exchange to completion. The write is guaranteed delivered and only the reply is discarded, so the replay double-applies rather than merely risking it.

Return `TransientNotCommitted` or `Disconnected` on an elapsed outer timer. The `Err(_)` arms wrapping awaits that issue no request are correct as they stand.

core/partitions/src/iggy_partition.rs
```diff
@@ -4616,7 +4648,14 @@ where
         // that reached the requested frontier closes the session; anything
         // less keeps it armed and the stall retry re-requests the remains
         // (`commit_min + 1..`), converging over rounds.
-        let done = commit_min >= session.to_op;
+        let fetch_complete = session.fetch_to_op <= session.commit_to_op
```
numinnex, line 4660: `fetch_complete` is proved from resident journal headers, but `commit_journal` at `:4643` has just evicted them, so a repair that fully succeeds reports itself incomplete.

`commit_journal` -> `committed_prefix(commit_max)` bounds the eviction by the **live** `commit_max`, not the session's snapshotted `commit_to_op`. Delivering the suffix bodies is what lets the group commit past `commit_to_op`, so the flush evicts into `(commit_to_op, fetch_to_op]`, and `repaired_window_shape` reads only `headers` and never the evicted ring (`journal.rs:770-772`). `complete` is false, so `fetch_complete` is false and `done` is false at `:4658` even though `commit_min >= commit_to_op`.

The asymmetry is what makes it self-defeating: if the group does *not* commit past `commit_to_op` the window sits above `commit_max`, survives eviction, and the session closes correctly. It is the successful commit that destroys the evidence.

Nothing recovers it on a healthy group. The tick twin at `core/shard/src/lib.rs:6534-6541` evaluates the identical expression and is blocked identically. The stall retry goes silent because `from_op = commit_min + 1` then exceeds `fetch_to_op` and trips the `from_op <= to_op` guard at `shard:6574`, so the wedge is also invisible in the logs. `partition.repair.is_some()` (`shard:7479`) then blocks every later re-arm for that partition, and every remaining clear path is view-change-gated, so it pins until the next election.

The new integration config (`messages_required_to_save: Some(1)`, `enforce_fsync: true`) guarantees the flush that triggers this.

Suggest deciding completion from a monotone fact the flush cannot erase: record suffix arrivals on the session as they are ingested (a high-water op or counter), or use `commit_min >= fetch_to_op`, rather than re-deriving presence from a vector `commit_journal` just truncated.

numinnex (approved):

hubcio (approved):

Issue #2835, opened 2026-02-27
feat(python): Add QUIC, HTTP, and WebSocket Transport Protocol Support to Python SDK


saie-ch:
## Summary
  The Python SDK currently only exposes TCP transport via the default `IggyClient()` constructor. While
  `IggyClient.from_connection_string()` theoretically supports all transport protocols (TCP, QUIC, HTTP, WebSocket)

  ## Current State

  ### Implementation Status
  - **TCP**: Fully implemented and documented

  - **QUIC**:  Theoretically supported via connection string, but untested
    - No constructor method
    - Connection string format: `iggy+quic://user:pass@host:port` (undocumented)
    - No examples or tests

  - **HTTP**: Theoretically supported via connection string, but untested
    - No constructor method
    - Connection string format: `iggy+http://user:pass@host:port` (undocumented)
    - No examples or tests

  - **WebSocket**: Theoretically supported via connection string, but untested
    - No constructor method
    - Connection string format: `iggy+ws://user:pass@host:port` (undocumented)
    - No examples or tests

  ### Integration Tests
  - No QUIC tests in `foreign/python/tests/`
  - No HTTP tests in `foreign/python/tests/`
  -  No WebSocket tests in `foreign/python/tests/`
  -  Only TCP tests exist: `foreign/python/tests/test_iggy_sdk.py`

  ### Examples
  -  No QUIC examples in `examples/python/`
  -  No HTTP examples in `examples/python/`
  -  No WebSocket examples in `examples/python/`
  - Only TCP examples exist: `examples/python/basic/` and `examples/python/getting-started/`

  ## Requirements

  ### 1. API Enhancement
  Add explicit constructor methods or builder pattern for each transport protocol:

  2. Integration Tests

  Add test coverage for each transport protocol in foreign/python/tests/:

  3. Examples

  Add transport-specific examples in examples/python/:

  Each example should demonstrate:
  - Connection establishment using the specific transport
  - Authentication
  - Basic producer/consumer operations
  - Error handling
  - Proper cleanup

  **Reference**

  Rust SDK (Underlying Implementation)

  - Client builder: core/sdk/src/clients/client_builder.rs
    - TCP: with_tcp() method (line 99)
    - QUIC: with_quic() method
    - HTTP: with_http() method
    - WebSocket: with_websocket() method
  - Transport enum: core/common/src/types/configuration/transport.rs
  - Connection string parser: core/common/src/types/configuration/auth_config/connection_string.rs

[...]

PR #3992, opened 2026-08-29
feat(python): add HttpConfig transport configuration

saie-ch:

Relates to #2835.

**Rationale**

  Python's HTTP transport was only reachable through the untested from_connection_string() path, with no config object like TCP got in #3776.

 Added `HttpConfig`, accepted by `IggyClient(...)` alongside a plain address and
  `TcpConfig`, mirroring the TcpConfig pattern. An earlier revision exposed this
  as an `IggyClient.http(...)` classmethod; that was dropped in favour of widening
  the existing constructor union, per review.

  **Local Execution**

  - `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `ruff`, `pyrefly`,
    `markdownlint` all clean
  - `apache_iggy.pyi` regenerated via `stub_gen` and in sync
  - Full Python suite passes against a local server (524 passed; `test_tls.py`
    needs Docker)
  - Both getting-started examples run, including with the HTTP config swap applied
  - Pre-commit hooks ran

  **AI Usage**

  Claude Sonnet 5 and Claude Opus 5


foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,145 @@
...
+        """Test that each configured field is readable back unchanged."""
+        config = HttpConfig(
+            api_url="http://127.0.0.1:3001",
+            retries=5,
+            jwt="a-token",
+        )
+
+        assert config.api_url == "http://127.0.0.1:3001"
+        assert config.retries == 5
+        assert config.has_jwt is True
```
ethanlin01x, line 63: Also assert `heartbeat_interval`

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,145 @@
...
+class TestHttpConfig:
+    """Test the transport configuration."""
+
+    def test_defaults_match_the_rust_sdk(self):
+        """Test that an unconfigured transport matches the Rust SDK defaults."""
+        config = HttpConfig()
+
+        assert config.api_url == "http://127.0.0.1:3000"
+        assert config.retries == 3
+        assert config.has_jwt is False
```
ethanlin01x, line 49: Also assert `heartbeat_interval`

examples/python/http/producer.py
```diff
@@ -0,0 +1,166 @@
...
+    except ValueError as error:
+        logger.error(f"Invalid client configuration: {error}")
+        return
+    logger.info(f"Connecting to {args.api_url}")
+
+    client = IggyClient.http(config)
+    logger.info("Connecting to IggyClient")
+    await client.connect()
+    logger.info("Connected.")
+    # HTTP is a stateless transport: log in explicitly rather than relying
```
ethanlin01x: Nit: "stateless" is loose here - `HttpClient` does hold a JWT and swaps it on `/users/refresh-token`. It explains the missing reconnection policy fine, but not the missing auto-login. The second half of the sentence already gives the real reason, so dropping the first clause would be enough.  Same in `consumer.py` and the README.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,145 @@
...
+        """Test that a client can be built from a config object."""
+        assert IggyClient.http(HttpConfig(api_url="http://127.0.0.1:3000")) is not None
+
+    def test_accepts_nothing(self):
+        """Test that the default configuration is used when no argument is given."""
+        assert IggyClient.http() is not None
+
+
+@pytest.mark.integration
+class TestHttpConfigAgainstServer:
```
ethanlin01x, line 139: Could you add one HTTP round trip here - login, create stream/topic, send, poll, compare the payload? That is the part the PR description says was untested.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Nothing downstream reads zero as "disabled"; it heartbeats in a
+        continuous loop for as long as the client lives.
+        """
+        with pytest.raises(ValueError, match="zero"):
+            HttpConfig(heartbeat_interval=timedelta(0))
+
+
+@pytest.mark.unit
+class TestClientConstruction:
```
ethanlin01x: nit: `test_client_config.py` already has a class with this name. Please rename it to `TestHttpClientConstruction`.

ethanlin01x (commented):
I ran this locally and it works well. I have a few suggestions, mostly around tests and docs.

saie-ch:
@ethanlin01x Could you please check now.

slbotbm (dismissed):
As was requested in the websocket pr: fold the http config example into existing examples as comments, and change the exposed API signature to `IggyClient( ... | HttpConfig | ...)`.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+    }
+}
+
+#[gen_stub_pymethods]
+#[pymethods]
+impl HttpConfig {
+    /// Constructs an HTTP configuration.
+    ///
+    /// Args:
+    ///     api_url: Base URL of the Iggy HTTP API. Defaults to `http://127.0.0.1:3000`.
```
hubcio, line 435: warning: `validate_api_url` rejects any path, query, fragment or credentials, so `https://gw.example.com/iggy` fails with a reasonless `ValueError`. say the accepted shape is `scheme://host[:port]` only, and add a path case to `test_invalid_api_url_is_rejected`.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+    ///     retries: Number of retries to perform on transient errors. Defaults to 3.
```
hubcio, line 436: warning: defaults to 3 and reqwest-retry replays POST, so a 504 on produce appends the batch twice - the server mints a fresh request id per request, so dedup misses it. document that HTTP sends are at-least-once and that `retries=0` restores TCP semantics.

foreign/python/README.md
```diff
@@ -168,6 +168,11 @@ async def main():
 asyncio.run(main())
 ```
 
+`IggyClient(...)` also accepts an `HttpConfig` for the HTTP transport. HTTP is a
```
hubcio, line 171: warning: consumer groups don't work over HTTP - `consumer_group(...)` defaults to auto-join, and `join_consumer_group` returns `Feature is unavailable`, so `init()` throws. list that here.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
     }
 }
 
+/// Configuration for the HTTP transport, accepted by `IggyClient(...)`.
```
hubcio, line 412: warning: a group poll over HTTP returns `[]` forever with no error - the server can't register an HTTP client, so it fences and answers 200 with an empty list. say HTTP is single-consumer only.

examples/python/getting-started/producer.py
```diff
@@ -113,6 +113,13 @@ def build_config(args: ArgNamespace) -> TcpConfig:
         tls_enabled=args.tls,
         tls_ca_file=args.tls_ca_file or None,
     )
+    # IggyClient(...) also accepts an HttpConfig for the HTTP transport. HTTP
```
hubcio, line 116: warning: this sits after `return TcpConfig(...)`, so uncommenting it doesn't switch transport - the example silently keeps running over TCP. move it above the return and widen the annotation to `TcpConfig | HttpConfig`.

same block at `consumer.py:117`.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+        if let Some(api_url) = api_url {
+            builder = builder.with_api_url(api_url);
+        }
+        if let Some(retries) = retries {
+            let retries = u32::try_from(retries).map_err(|_| {
+                PyValueError::new_err(format!("'retries' must be between 0 and {}", u32::MAX))
+            })?;
+            builder = builder.with_retries(retries);
+        }
+        if let Some(jwt) = jwt {
```
hubcio, line 465: warning: `jwt=""` is accepted, so `has_jwt` says True while `is_authenticated()` is false and every call fails `Unauthenticated`. whitespace is worse - it passes the local check and 401s server-side. reject empty and whitespace here.

foreign/python/README.md
```diff
@@ -168,6 +168,11 @@ async def main():
...
+stateless per-request transport, so there is no reconnection policy to configure.
+There is also no `AutoLogin`: call `login_user(...)` explicitly after connecting.
+See `examples/python/getting-started/producer.py` for a config swap example.
```
hubcio, line 174: nit: that block is commented-out code sitting after a `return`, so it isn't a usable example. inline the four lines here instead.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+    /// Constructs an HTTP configuration.
+    ///
+    /// Args:
+    ///     api_url: Base URL of the Iggy HTTP API. Defaults to `http://127.0.0.1:3000`.
+    ///     retries: Number of retries to perform on transient errors. Defaults to 3.
+    ///     jwt: JWT token for A2A (Agent-to-Agent) authentication. Defaults to `None`.
+    ///     heartbeat_interval: Interval of heartbeats sent by the client. Defaults to 5 seconds.
+    ///
+    /// Raises:
+    ///     ValueError: If `api_url` is not a valid URL, if `retries` is outside the
```
hubcio, line 441: nit: `retries` is `Option<i64>`, so `retries=2**64` raises `OverflowError` from pyo3 before this check runs. name `OverflowError` here, or take a `&Bound<'_, PyAny>` and range-check it.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+#[gen_stub_pymethods]
+#[pymethods]
+impl HttpConfig {
+    /// Constructs an HTTP configuration.
+    ///
+    /// Args:
+    ///     api_url: Base URL of the Iggy HTTP API. Defaults to `http://127.0.0.1:3000`.
+    ///     retries: Number of retries to perform on transient errors. Defaults to 3.
+    ///     jwt: JWT token for A2A (Agent-to-Agent) authentication. Defaults to `None`.
+    ///     heartbeat_interval: Interval of heartbeats sent by the client. Defaults to 5 seconds.
```
hubcio, line 438: nit: copied from `TcpConfig`, but over HTTP the ping is a bare `GET /ping` against a handler that registers nothing - it keeps no session alive. say it's a liveness probe only.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,190 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""
+Tests for the HTTP client configuration surface.
+
+`HttpConfig` mirrors the Rust SDK's `HttpClientConfig` the same way
+`TcpConfig`/`QuicConfig` do, so most of these assert that a value set from
```
hubcio, line 22: nit: no `QuicConfig` is bound - `lib.rs` registers only `AutoLogin`, `TcpConfig`, `TcpReconnectionConfig` and `HttpConfig`. drop it here; line 24's "unlike TCP/QUIC" is fine, that's true of the transports.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,190 @@
...
+    def test_repr_hides_the_jwt(self):
+        """Test that the JWT does not leak through repr but still parses as Python."""
+        config = HttpConfig(jwt="a-secret-token")
+
+        printed = repr(config)
+
+        assert "a-secret-token" not in printed
+        ast.parse(printed)
+
+    def test_repr_shows_every_field_as_python(self):
```
hubcio, line 81: nit: this skips `heartbeat_interval`, the one field whose repr is format-fragile, and `ast.parse` won't catch a regression there. add `heartbeat_interval=timedelta(seconds=15)` to the config below, then assert it - asserting alone goes red against the 5s default.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,190 @@
...
+        config = HttpConfig(
+            api_url="http://127.0.0.1:3001",
+            retries=5,
+            jwt="a-token",
+            heartbeat_interval=timedelta(seconds=15),
+        )
+
+        assert config.api_url == "http://127.0.0.1:3001"
+        assert config.retries == 5
+        assert config.has_jwt is True
```
hubcio, line 63: nit: `has_jwt` only proves the flag, nothing checks the token reaches `access_token`. `/users/login` is public, so stdlib `urllib` can mint one out of band and assert an authenticated call works.

core/sdk/src/prelude.rs
```diff
@@ -41,6 +41,7 @@ pub use crate::clients::producer_builder::IggyProducerBuilder;
 pub use crate::clients::producer_config::{BackgroundConfig, DirectConfig};
 pub use crate::clients::producer_sharding::{BalancedSharding, OrderedSharding, Sharding};
 pub use crate::consumer_ext::IggyConsumerMessageExt;
+pub use crate::http::http_client::HttpClient;
```
hubcio, line 44: nit: the prelude now carries clients for http, tcp and websocket but not `QuicClient`, though `QuicClientConfig` is already here. add it, or drop this line and import `HttpClient` from its module in the binding.

.github/actions/python-maturin/pre-merge/action.yml
```diff
@@ -171,6 +171,7 @@ runs:
         # overwrite the coverage-instrumented .so with a non-instrumented one
         IGGY_SERVER_HOST=127.0.0.1 \
         IGGY_SERVER_TCP_PORT=8090 \
+        IGGY_SERVER_HTTP_PORT=3000 \
```
hubcio, line 174: nit: host and both ports duplicate `server-start`'s defaults, so changing one desyncs silently. if you wire the outputs, split them first - `http_address` is `host:port`, and `int()` on that raises in `utils.py`.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        with pytest.raises(ValueError, match="zero"):
+            HttpConfig(heartbeat_interval=timedelta(0))
+
+
+@pytest.mark.unit
+class TestHttpClientConstruction:
+    """Test that `IggyClient(...)` accepts an `HttpConfig`."""
+
+    def test_accepts_a_config(self):
```
hubcio, line 129: nit: nothing here pins the transport - both assert only `is not None`, and the integration test would pass with a TCP client too, since CI has 8090 listening. `IggyClient(HttpConfig(api_url="http://127.0.0.1:1"))` then `ping()` must raise.

foreign/python/README.md
```diff
@@ -168,6 +168,11 @@ async def main():
 asyncio.run(main())
 ```
 
+`IggyClient(...)` also accepts an `HttpConfig` for the HTTP transport. HTTP is a
+stateless per-request transport, so there is no reconnection policy to configure.
+There is also no `AutoLogin`: call `login_user(...)` explicitly after connecting.
```
hubcio, line 173: nit: over HTTP `connect()` is a no-op that never dials, so "after connecting" misleads. the SDK's own wording is "call `login_user()` instead of `connect()`".

foreign/python/src/client.rs
```diff
@@ -86,45 +86,73 @@ fn resolve_topic_params(
...
+        conn: Option<PyClientConfig>,
     ) -> PyResult<Self> {
-        let config = match conn {
-            Some(PyClientConfig::Config(config)) => config.client_config(),
-            Some(PyClientConfig::ServerAddress(server_address)) => Arc::new(
-                TcpClientConfigBuilder::new()
-                    .with_server_address(server_address)
-                    .build()
-                    .map_err(|e| {
+        match conn {
```
hubcio, line 113: simplification: four arms each end in the same `Ok(Self { inner: Arc::new(RustIggyClient::new(...)) })`, and the `PyErr::new` closure is written five times. let the match yield a `ClientWrapper` and construct once after it.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+        if let Some(retries) = retries {
+            let retries = u32::try_from(retries).map_err(|_| {
+                PyValueError::new_err(format!("'retries' must be between 0 and {}", u32::MAX))
+            })?;
+            builder = builder.with_retries(retries);
+        }
+        if let Some(jwt) = jwt {
+            builder = builder.with_jwt(jwt);
+        }
+        let mut inner = builder
```
hubcio, line 468: simplification: `HttpClientConfigBuilder::with_heartbeat_interval` already takes the `NonZeroIggyDuration` that `reject_zero` returns, so this post-build patch and the `mut` aren't needed. `TcpConfig` patches only because its builder has no setter.

foreign/python/src/client.rs
```diff
@@ -86,45 +86,73 @@ fn resolve_topic_params(
...
-    ///         raises `ValueError` when it is constructed, before it ever reaches
-    ///         this call. Neither exception is a subclass of the other.
+    ///     conn: A `host:port` address, a `TcpConfig`, or an `HttpConfig`. Defaults
+    ///         to `127.0.0.1:8090` over TCP with auto-login disabled. A malformed
+    ///         address is reported differently depending on the form: the string
+    ///         form raises `RuntimeError` here, while `TcpConfig`/`HttpConfig`
+    ///         raise `ValueError` when they are constructed, before either ever
+    ///         reaches this call. Neither exception is a subclass of the other.
     ///
     /// Raises:
```
hubcio, line 101: simplification: the `HttpConfig` half can't fire - `build()` already validated `api_url` and there's no setter, so `create` re-checks the same string. keep the string-address clause, drop the justification.

foreign/python/tests/utils.py
```diff
@@ -58,6 +62,26 @@ def get_server_config() -> tuple[str, int]:
...
+    """
+    Get TCP server configuration from environment variables or defaults.
+
+    Returns:
+        tuple: (host, port) for the Iggy server
+    """
+    return get_transport_config("IGGY_SERVER_TCP_PORT", 8090)
+
+
+def get_http_server_config() -> tuple[str, int]:
```
hubcio, line 75: simplification: a one-line delegation with a six-line docstring and two call sites. give `get_server_config` default args instead, or call `get_transport_config` directly at both sites.

foreign/python/src/config.rs
```diff
@@ -408,16 +409,127 @@ impl TcpConfig {
...
+        heartbeat_interval: Option<Py<PyDelta>>,
+    ) -> PyResult<Self> {
+        // The builder starts from `HttpClientConfig::default()`, and its `build()`
+        // trims and validates the API URL whether or not one was set here.
+        let mut builder = HttpClientConfigBuilder::new();
+        if let Some(api_url) = api_url {
+            builder = builder.with_api_url(api_url);
+        }
+        if let Some(retries) = retries {
+            let retries = u32::try_from(retries).map_err(|_| {
```
hubcio, line 460: simplification: same conversion and error template as `TcpReconnectionConfig::new` above. a private `u32_arg(value, parameter)` in this file covers both, and gives the `OverflowError` fix one place to live.

foreign/python/tests/test_http_config.py
```diff
@@ -0,0 +1,190 @@
...
+        """Test that a client can be built from a config object."""
+        assert IggyClient(HttpConfig(api_url="http://127.0.0.1:3000")) is not None
+
+    def test_accepts_the_default_config(self):
```
hubcio, line 133: simplification: this is the same test as the one above - `HttpConfig()` defaults `api_url` to that exact literal, and both hit the same union arm. keep one, once the transport assertion is in.

hubcio (changes requested):
my main worry is that HTTP now reads as a first-class transport without saying what it cannot do, and two of the gaps are silent.

A few findings land on code this PR doesn't touch, so I couldn't attach them inline:

- `core/sdk/src/http/http_client.rs:98,117,137,157,173,192,213` - the `access_token` read guard is held across the whole `.send().await` in all seven verbs, including every retry backoff. `IggyRwLock` is write-preferring, so one `login_user` queues behind in-flight requests and then blocks new readers. There's no request timeout either (`:272` is a bare `reqwest::Client::new()`), so a hung server pins the guard with no bound. `refresh_access_token` at `:312-320` already does the clone-then-drop thing.
- `foreign/python/README.md:137` - lead-in still says `IggyClient` takes "either a server address or a `TcpConfig`", 34 lines above the new paragraph.
- `.github/actions/python-maturin/pre-merge/action.yml:167` - `${{ steps.iggy.outputs.address }}` isn't an output `server-start` declares, so that line logs an empty address. `tcp_address` is the one you want.
- `foreign/python/src/client.rs:1117` - `consumer_group()` is the one HTTP-hostile entry point with no HTTP note; join and leave both carry one.
- `core/common/src/types/configuration/http_config/http_client_config.rs:47-55` - separate from this PR, but worth an issue: the HTTP connection string requires `user:pass@` and then drops it, so `from_connection_string("iggy+http://iggy:iggy@...")` gives an unauthenticated client. tcp, quic and websocket all carry `auto_login` through; http is the only one that doesn't. Same line hardcodes `http://`, so `HttpConfig` is the only Python route to an https endpoint.
- `core/sdk/src/clients/client.rs:358` - the doctest connection string has no `@`, so it can't parse. It's `no_run`, so nothing catches it.

foreign/python/src/config.rs
saie-ch: I went to add a test for this and couldn't reproduce the empty result. Same
  server, same stream/topic, same call, only the transport differs:

      group EXISTS (2 messages produced, partition 0)
        HTTP group poll -> 2 message(s)
        TCP  group poll -> 0 message(s)

  TCP behaves exactly as you describe. HTTP returns the full batch.

  As far as I can tell the fence never runs over HTTP: `Consumer::kind` is
  `#[serde(skip)]`, so the kind never reaches the server and the poll takes the
  plain-consumer path, served as an ordinary consumer named after the group.

  I've kept "single-consumer only" but reworded the reason. Am I reading this
  right?

foreign/python/README.md
saie-ch: HttpClient::connect() is a no-op, but IggyClient::connect() isn't—the latter starts the heartbeat configured by heartbeat_interval.

So the README now instructs users to call connect() followed by login_user(...), rather than using one instead of the other.

Is this approach fine for you?

hubcio (approved):

mmodzelewski (approved):

examples/python/http/producer.py
hubcio: dropped it. the whole `examples/python/http/` dir is gone too, folded into getting-started as a comment. wording now just names the two real gaps, no AutoLogin and no reconnection policy.

foreign/python/src/config.rs
hubcio: you're right, and i was wrong. kind is `#[serde(skip)]`, so the poll gets served as a plain consumer named after the group. the `[]` i described is TCP.

but your repro passed an explicit partition_id, and `consumer_group(...)` can't get there. it pins partition_id to None, and the default `poll_messages_with_strategy_for` rejects a partition-less group poll client-side with `Feature is unavailable`. so that path always fails. what you measured needs a direct poll with a partition. docs cover both now.

foreign/python/README.md
hubcio: yep, and my wording was wrong. `HttpClient::connect` is a bare `Ok(())` but `IggyClient::connect` starts the heartbeat after it, so connect() then login_user() is right and "instead of" wasn't. the class docstring says the same now so they can't drift.

hubcio (approved):

spetz (approved):

PR #3995, opened 2026-08-30
feat(cpp): let poll and offset calls take a typed consumer

ethanlin01x:

Relates to #3876

## Rationale

The C++ SDK already let the caller name a consumer, but as a kind string paired with an identifier. Nothing checked that pair until it reached Rust, and the "let the server pick the partition" sentinel had no name at all.

## What changed?

`poll_messages` and the three consumer-offset calls took `consumer_kind: String`, `consumer_id: Identifier`. A typo such as "`consumer_grp`" compiled fine and failed at runtime, and `ConsumerKind::ConsumerGroup` was reachable only by spelling the kind correctly.

They now take a single `Consumer` carrying a `ConsumerKind` enum, so the kind is checked where the call is written. `iggy::Consumer::Single(id)` and `iggy::Consumer::Group(id) `build it from either a name or a number, mirroring how `TopicOption` builds its entries. `iggy::kAnyPartitionId` gives the `u32::MAX `partition sentinel a name and documents what each call does with it: a group poll reads one of the partitions assigned to the polling member, a regular poll and get_consumer_offset read partition 0, and the two offset writes reject it.

The bridge already used the name Consumer for the placeholder wrapping the high-level consumer, so that one is renamed `IggyConsumer`, matching the Rust SDK.

### Breaking change: 
the four calls replace their `consumer_kind` and `consumer_id` arguments with a single consumer.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? Claude code
2. Scope of usage? Analysis and implementation
3. How did you verify the generated code works correctly? Local run

### DIFF
--- a/bdd/cpp/features/step_definitions/messaging_steps.cpp
+++ b/bdd/cpp/features/step_definitions/messaging_steps.cpp
@@ -150,7 +150,7 @@ WHEN("^I poll messages from stream ([0-9]+), topic ([0-9]+), partition ([0-9]+) 
     const auto polled           = context->client->poll_messages(
         bdd::make_numeric_identifier(static_cast<std::uint32_t>(stream_id)),
         bdd::make_numeric_identifier(static_cast<std::uint32_t>(topic_id)), static_cast<std::uint32_t>(partition_id),
-        "consumer", bdd::make_numeric_identifier(1), std::string(polling_strategy.PollingStrategyKind()),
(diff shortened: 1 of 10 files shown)


slbotbm:
@ethanlin01x could you convert this branch to draft? I have code in #4022 that will simplify some of the code you're writing here, and so I'd like to review this code after merging that.

slbotbm:
@ethanlin01x the pr was merged, so could you rebase and resolve the conflicts please?

PR #3996, opened 2026-08-30
feat(connectors): add DynamoDB sink connector

### LINKED ISSUE #3731: Add Amazon DynamoDB sink connector
### Description

Add a native Rust sink connector for Amazon DynamoDB, allowing Iggy stream messages to be written directly into DynamoDB tables.

This is tracked in the connector ecosystem roadmap: #2753

### Affected area / component

Connectors

### Proposed solution

**Crate:** [`aws-sdk-dynamodb`](https://crates.io/crates/aws-sdk-dynamodb) 
(official AWS SDK, async/Tokio compatible)

**Key design decisions:**

1. **Batching** — `batch_write_item` has a hard limit of 25 items per request. The `consume()` handler will split incoming message batches into chunks of ≤25 before sending.

2. **Idempotency** — The connector will support a configurable `partition_key_field` (and optional `sort_key_field`) mapped from message payload. `batch_write_item` uses `PutRequest` which overwrites 
existing items on key conflict, making repeated delivery idempotent as long as the primary key is deterministic.

3. **UnprocessedItems handling** — `batch_write_item` can return partial failures via `UnprocessedItems` without throwing an error. The connector will retry unprocessed items with exponential backoff rather than silently dropping them.

4. **Throughput errors** — `ProvisionedThroughputExceededException` will be handled with retry logic compatible with both Provisioned and On-Demand capacity modes.

5. **Item size limit** — DynamoDB enforces a 400KB per-item limit. Oversized messages will be logged and skipped rather than failing the entire batch.

### Alternatives considered

_No response_

### Contribution

### Good first issue

ethanlin01x:

Closes #3731 

## Rationale

Iggy can write to several databases, but not to Amazon DynamoDB, so users have to write their own consumer for it.

## What changed?

Stream messages could not reach a DynamoDB table without custom code. This adds a sink connector that maps each message to a DynamoDB item and writes it with `BatchWriteItem`, splitting batches at the 25 item limit and retrying throttled requests and unprocessed items with exponential backoff. The item key is built from the stream, topic, partition and message offset when the payload does not carry the configured key field, so a redelivered message overwrites the same item. Records that DynamoDB would reject, such as items over 400 KB or an invalid key type, are skipped with a warning instead of failing the whole batch.

The sink uses the official `aws-sdk-dynamodb`, so the repo now carries the AWS SDK next to the `rust-s3` stack that the S3 sink uses. DynamoDB has no comparable third-party client.

The commits are split so that each one is a working increment: base sink, keys, skipped records, retries, key schema check, then tests and docs.

## Local Execution

- Passed
- Pre-commit hooks ran 

## AI Usage
1. Which tools? Claude Code
2. Scope of usage? Implementation of connector
3. How did you verify the generated code works correctly? Verified with the unit tests, the container-backed integration tests, and a manual end-to-end run against DynamoDB Local


core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+        }
+
+        item.entry(self.partition_key_field.clone())
+            .or_insert_with(|| {
+                AttributeValue::S(build_message_key(
+                    topic_metadata,
+                    messages_metadata,
+                    message.id,
+                ))
+            });
```
slbotbm, line 563: `DynamoDbSink::build_item` derives the default `iggy_id` from `build_message_key`, which uses `message.id`. `IggyMessage::new` defaults an omitted message ID to `0`, so ordinary messages built without `.id(...)` receive the same default DynamoDB key within a stream partition. `DynamoDbSink::deduplicate_items` then keeps only the newest item in a consume call, and later batches overwrite the same DynamoDB record. The integration helper always sets an ID, so it does not cover the default. Use the stable, unique message offset in the generated key, or require and validate a non-default message ID.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+fn build_message_key(
+    topic_metadata: &TopicMetadata,
+    messages_metadata: &MessagesMetadata,
+    message_id: u128,
+) -> String {
+    format!(
+        "{}:{}:{}:{message_id}",
+        topic_metadata.stream, topic_metadata.topic, messages_metadata.partition_id
+    )
+}
```
slbotbm, line 690: `build_message_key` serializes stream, topic, partition, and message ID with an unescaped `:` separator. Stream and topic names permit `:`, so `(stream = "a:b", topic = "c")` and `(stream = "a", topic = "b:c")` generate the same key for the same partition and message ID. A connector configured for both topics can overwrite unrelated items. Use length-delimited or binary encoding, or a collision-resistant digest, and add collision tests.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+
+    /// `AttributeValue` is not hashable, so the key attributes are rendered
+    /// into a string that only lives for the deduplication pass.
+    fn key_signature(&self, item: &HashMap<String, AttributeValue>) -> String {
+        let mut signature = key_attribute_signature(item.get(&self.partition_key_field));
+        if let Some(sort_key_field) = &self.sort_key_field {
+            signature.push('|');
+            signature.push_str(&key_attribute_signature(item.get(sort_key_field)));
+        }
+        signature
```
slbotbm, line 410: `DynamoDbSink::deduplicate_items` has an independent collision bug for composite String keys. `key_signature` joins unescaped `S:`/`N:` values with `|`, so `{pk: S("a"), sk: S("b|S:c")}` collides with `{pk: S("a|S:b"), sk: S("c")}`. The second item silently replaces the first before the request is sent. Use a typed, length-delimited key representation rather than a display string.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+                    self.config.table,
+                    describe_sdk_error(&error)
+                ))
+            })?;
+        self.validate_key_schema(
+            description
+                .table
+                .and_then(|table| table.key_schema)
+                .unwrap_or_default(),
+        )?;
```
slbotbm, line 178: `DynamoDbSink::open` passes only `TableDescription::key_schema` to `DynamoDbSink::validate_key_schema`. `validate_key_attribute` consequently accepts any non-empty String, Number, or Binary value and never enforces the 2,048-byte partition-key or 1,024-byte sort-key limits. DynamoDB requires an exact key type: a table with an `N`/`B` partition key still opens, although the connector's generated partition key is always `S`; a missing sort key is always generated as `N`. A mismatched or overlong key causes DynamoDB to reject the whole `BatchWriteItem` request. Retain the table's attribute definitions, validate each key's exact type and byte length before batching, and reject configurations incompatible with generated key values.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+        AttributeValue::N(number) => number.len(),
+        AttributeValue::B(blob) => blob.as_ref().len(),
+        AttributeValue::Bool(_) | AttributeValue::Null(_) => 1,
+        AttributeValue::Ss(values) => values.iter().map(String::len).sum(),
+        AttributeValue::Ns(values) => values.iter().map(String::len).sum(),
+        AttributeValue::Bs(values) => values.iter().map(|blob| blob.as_ref().len()).sum(),
+        AttributeValue::L(values) => values.iter().map(attribute_value_size).sum(),
+        AttributeValue::M(values) => estimate_item_size(values),
+        _ => 0,
+    }
```
slbotbm, line 735: `estimate_item_size` and `attribute_value_size` are not DynamoDB item-size calculations. They omit the 100-byte base item overhead and Map/List overhead, and size Numbers by their string representation rather than DynamoDB's number-sizing rules. Nested items can therefore pass the `max_item_size` check but make DynamoDB reject the entire request, contrary to the README's promise that oversized records are skipped. Implement an exact or deliberately conservative size check and add nested boundary tests. DynamoDB documents the relevant overhead in https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CapacityUnitCalculations.html .

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+
+    async fn backoff(&self, attempt: u32) {
+        let delay = jitter(exponential_backoff(
+            self.retry_delay,
+            attempt - 1,
+            self.max_retry_delay,
+        ));
+        tokio::time::sleep(delay).await;
+    }
```
slbotbm, line 508: `DynamoDbSink::backoff` caps the exponential delay and then applies `jitter`. Since `jitter` adds up to 20%, `max_retry_delay = "5s"` can sleep for six seconds even though the README calls it an upper bound. Clamp after jitter, as the Doris sink does, or describe it as the pre-jitter cap.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+    matches!(
+        code,
+        "ProvisionedThroughputExceededException"
+            | "RequestLimitExceeded"
+            | "ThrottlingException"
+            | "InternalServerError"
+            | "ServiceUnavailable"
+            | "TransactionInProgressException"
+    )
+}
```
slbotbm, line 762: `is_transient_code` treats `ReplicatedWriteConflictException` as permanent. DynamoDB marks this multi-Region strongly consistent global-table conflict retryable, but the connector returns a permanent error rather than applying its configured backoff. Add the error code and a regression test.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+    async fn build_client(&self) -> Result<Client, Error> {
+        if self.config.access_key_id.is_some() != self.config.secret_access_key.is_some() {
+            return Err(Error::InvalidConfigValue(
+                "Partially configured credentials. You must provide both access_key_id \
+                 and secret_access_key, or omit both."
+                    .to_owned(),
+            ));
+        }
+
+        let mut loader = aws_config::defaults(BehaviorVersion::latest());
```
slbotbm, line 223: `DynamoDbSink::build_client` only checks whether `access_key_id` and `secret_access_key` are paired. A configured `session_token` without that pair is silently ignored and the default AWS credential chain is used. Reject that invalid configuration so an operator cannot believe STS credentials are active when they are not.

core/connectors/sinks/dynamodb_sink/src/lib.rs
```diff
@@ -0,0 +1,1348 @@
...
+        if let Some(region) = &self.config.region {
+            loader = loader.region(Region::new(region.clone()));
+        }
+        if let Some(endpoint) = &self.config.endpoint {
+            info!("Using custom DynamoDB endpoint: {endpoint}");
+            loader = loader.endpoint_url(endpoint);
+        }
```
slbotbm, line 230: `build_client` logs custom endpoint URLs verbatim. An arbitrary URL may contain user information or query credentials. Redact the URL at the log site, consistent with connector secret-handling guidance.

core/connectors/sinks/dynamodb_sink/README.md
```diff
@@ -0,0 +1,95 @@
...
+connector injects a key built from the stream, topic, partition, and message ID.
+When `sort_key_field` is configured and missing, the message offset is injected.
+A payload value always wins over the injected one, so a message that carries the
+key field with an empty value, or with a value that is neither a string, a
+number, nor binary, is skipped rather than falling back to the injected key,
+because DynamoDB would reject the whole batch.
+
+The key fields are checked against the table on startup. A `partition_key_field`
+or `sort_key_field` that does not match the table key schema fails the connector
+while it opens, instead of on the first write.
```
slbotbm, line 80: This section says key fields are checked against the table on startup, but the implementation checks only names and HASH/RANGE roles. It should document exact `S`/`N`/`B` compatibility, the generated partition key's `S` requirement, and the generated sort key's `N` requirement. Its statement that a payload key always wins also conflicts with `DynamoDbSink::build_item`, where enabled Iggy metadata overwrites a colliding payload field before generated keys are filled.

core/connectors/sinks/dynamodb_sink/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+JSON objects are written attribute by attribute, so a message field becomes a
+DynamoDB attribute of the matching type. JSON arrays and scalars are nested
+under a `payload` attribute, because a DynamoDB item must be a map. Text
+payloads go into `payload` as a string. Raw payloads are parsed as JSON when
+possible, otherwise they are stored as binary. Protobuf, FlatBuffer, and Avro
+payloads are not supported and are skipped with a warning.
+
+Metadata attributes are written after the payload, so they overwrite payload
+fields of the same name.
```
slbotbm, line 62: `DynamoDbSink::build_item` never consumes `ConsumedMessage::headers`. The README should explicitly state that user headers are not persisted, or the connector should provide a documented representation for them.

slbotbm (changes requested):
In addition to the comments, you have not added the dynamodb connector to the CI, so while the plugin compiles and the unit tests run, ` integration::connectors::dynamodb` do not run if changes in dynamodb connector are detected.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 96e5757b9. The key now uses the message offset, which is unique in a partition and stays the same on redelivery.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 96e5757b9. Each name now carries its byte length in front, so a : inside a name cannot build another topic's key.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 61f81ebe0. Deduplication now compares a typed key instead of a joined string.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in b77e1c8ef. The size now counts the 100 byte item overhead, the map and list overhead, and sizes numbers by the DynamoDB rules. Every unknown rounds up, so the check is conservative rather than exact.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 5fa507db6. I took the first option: the clamp happens after the jitter, like the Doris sink, so `max_retry_delay` is a real upper bound.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 2ffa46219

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in e3add3d77

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 84caa3a2f

core/connectors/sinks/dynamodb_sink/README.md
ethanlin01x: Fixed in bef6bb677.

core/connectors/sinks/dynamodb_sink/README.md
ethanlin01x: Fixed in bef6bb677. I took the first option: the README now says message headers are not written, and an item holds only the payload and the enabled iggy_* attributes.

core/connectors/sinks/dynamodb_sink/src/lib.rs
ethanlin01x: Fixed in 5f3ff7f16. The attribute definitions are kept now, and every item is checked against the declared `S`/`N`/`B` type and the key size limits before it is batched. A message that fails is skipped on its own, not with the whole request.

I did one part differently: a wrong key *name* fails the connector at open, but a wrong key *type* only warns. The generated key is used only when the payload does not carry the field, so a table with an `N` partition key and a payload carrying a numeric key is a valid setup, and failing at open would block it. I can make it a hard error if you prefer.

ethanlin01x:
@slbotbm 
Thanks for your review, I've learned a lot!
You're right that `integration::connectors::dynamodb` doesn't run when only the sink changes. However, this issue isn't specific to this connector, so I've opened a PR to address it across the codebase: https://github.com/apache/iggy/pull/4077

PR #3998, opened 2026-08-30
test(partitions): gate the persist-retry test on Linux

hubcio:
The retry test reaches for /dev/full, which only exists on
Linux, but its #[cfg(target_os = "linux")] gate and doc
comment landed on the preceding test instead. macOS builds
of the test target failed with E0425 on DEV_FULL.

Pre-merge builds macOS without --tests, so the break only
surfaced in the post-merge macOS cache warm, which does
compile tests.


djanand:
Looks good. Facing the same on ci-pre hooks when merging upstream/master to forked repo.

spetz (approved):

mmodzelewski (approved):

PR #3999, opened 2026-08-30
ci(repo): upgrade Cargo-Rail planning to v0.29

loadingalias:

N/A

## Rationale

I'm the maintainer of Cargo-Rail. This PR originally targeted Cargo-Rail v0.25.0 and `cargo-rail-action` v8.2.0.

Testing the integration against Iggy exposed - a lot of small issues - in both projects. Those issues have since been fixed, and this PR now uses Cargo-Rail v0.29.0 with `cargo-rail-action` v10.0.1.

Iggy's pre-merge workflow currently consumes a retired Cargo-Rail plan format, while post-merge image selection maintains a separate affected-change policy. These paths can drift from Cargo's dep graph and from each other.

Planning remains an optimization. If the Action, planner, validated reader, or matrix selection fails, Iggy retains its existing full-workspace and all-image behavior.

## What Changed?

The Cargo-Rail config was regenerated and reduced to the policy required for Iggy's planning integration, and only planning integration.

Pre-merge now uses the v10 validated selector reader instead of parsing planner JSON in workflow shell. Test selection uses typed Cargo arguments, while integration tests include their required server, CLI, benchmark, connectors, MCP, and connector `cdylib` runtime artifacts.

Post-merge image selection now comes from a five-variant Cargo-Rail catalog. This removes the duplicate publish gate and the legacy `edge-affected-images.sh` policy.

The Action is pinned to the exact v10.0.1 commit:

`loadingalias/cargo-rail-action@4a8df181e4f340a164190cf092c9f9239b20edab`

Releases:

- https://github.com/loadingalias/cargo-rail/releases/tag/v0.29.0
- https://github.com/loadingalias/cargo-rail-action/releases/tag/v10.0.1

The ASF allowlist request was reviewed and merged:

- https://github.com/apache/infrastructure-actions/pull/1298

Cargo-Rail Action v10.0.1 and the exact SHA used here are now present on the ASF allowlist.

The Cargo-Rail plan contract remains v9, and the root Action's planner interface is unchanged.

## Migration and fallback behavior

This PR initially compares against a base containing Cargo-Rail v0.25's retired `[change-detection]` configuration. Cargo-Rail v0.29.0 rejects that historical key, so the first planning attempt deliberately restores full-workspace testing and all-image refreshes.

This is a safe, one-time migration fallback. Once the new config is present on the base branch, normal scoped planning applies.

The workflow also falls back conservatively if:

- planning fails
- plan validation fails
- a selector fails
- runtime-artifact scope is incomplete
- the image matrix is invalid
- the checkout changes after plan capture

## Local Execution

No Rust source or Cargo manifests are changed by this PR.

Validation performed locally:

- released Cargo-Rail v0.29.0 and Action v10.0.1 tag identities
- ASF verifier against the exact Action SHA, exit code 0
- expected historical-config failure and full-scope fallback
- successful post-migration planning
- plan contract v9
- saved-plan verification against the matching checkout
- checkout-drift rejection
- skipped, package-scoped, workspace-scoped, and `--all` paths
- typed test, target, runtime-artifact, and image selectors
- connector SDK regression:
- `cargo.test` remains package-scoped
- 24 test packages are selected
- `integration` is selected
- 26 runtime-artifact packages are selected
- only the `rust-connectors` image is selected
- repository Taplo and ShellCheck checks
- targeted Actionlint validation
- YAML and JSON validation
- trailing whitespace and trailing newline checks
- binary-artifact check
- `git diff --check`

Unrelated changes can produce an empty edge-image selection. Rust work remains conservative when portable compiler, build-script, or proc-macro evidence is unavailable.

Local environment limitations:

- `cargo-sort` was unavailable (this is something cargo-rail does natively!)
- repository-wide clippy was blocked by a missing system `hwloc` library.
- the full external-service integration suite was not run locally.

## AI Usage

1. **Tool:** Codex/Sol xhigh
2. **Scope:** Review and validation of Iggy's Cargo-Rail planning integration, fallback behavior, workflow configuration, selector contracts, and external Action trust boundary. Also, I asked the model to remove the 'cache', 'unify', and 'surface' options from the PR to keep it cleaner. It's a bad habit and I shouldn't have added it. I'd be happy to open those PRs if ever you guys are interested.
3. **Verification:** I reviewed the complete diff and reran the applicable planner, Action, workflow, configuration, and repository checks listed above.
4. **Can you explain every line if asked?** Yes. I am the maintainer of Cargo-Rail.


loadingalias:
Aside from this PR, I evaluated Cargo-Rail v0.25 against Iggy in an isolated Linux build.

`unify` produced one concrete follow up: promote `server`, `iggy-gateway-kafka`, and `iggy_connector_doris_sink` into `workspace.dependencies`, then convert 33 existing declarations across 10 manifests to workspace inheritance. It preserved the optional `mimalloc` and `strum` feature edges and found no unused dependencies, dead features, duplicate version opportunities, version mismatches, or transitive pins.

If the dev team are using cargo-udeps, cargo-msrv, cargo-workspace-hack, etc. - you can free the resources on those dev machines now. Cargo-Rail `unify` will cover it.

I would keep it sparse for now:

```toml
[unify]
exact_pin_handling = "preserve"
```

That protects Iggy's exact compio pin while retaining Cargo-Rail's conservative defaults. Because ordinary analysis currently trips Iggy's `-D unused-crate-dependencies` policy, I would treat this as a reviewed cleanup PR rather than enable unify in CI yet.

I would like to slowly get the other workflows integrated, but I can't recommend them until I have more testing/benching time.

- Cache needs an isolated benchmark against Iggy's current Swatinem/rust-cache setup. However, Cargo-Rail caching is restoring more locally and remotely that sccache in current testing. I would almost bet a 50% reduction in build times across the entire workspace.

- Surface in v0.25 stops while acquiring facts for the `tokio::task_local!` declaration in `leader_aware.rs`; I've patched in v0.25.1, but haven't cut the release yet. No intergration for Iggy under 'surface' just yet.

- I think the reviewed `.changes` files look useful as a standalone release intent/notes pilot, too. I have a few fixes to make under the full release workflow; Iggy exposed shortcomings in the current. Cargo-Rail cannot currently preserve Iggy's four tag formats or broader Apache and cross-language release process. 

If you wanted me to submit the PRs, I think I'd prefer one small dep-inheritance PR followed by the .changes pilot. I will make sure cache gets a measured experiment in the next few days, Surface gets patched, and full release automation stays out for now.

hubcio:
hi,
as a apache project we cannot use actions that are not approved in https://github.com/apache/infrastructure-actions due to supply attack risk.

hubcio:
btw, unify sounds nice. can we use it without action, via something fast, like `cargo binstall`? (+pin the version)

and thanks for contribution!

hubcio:
have you already joined our discord?

loadingalias:
Hey @hubcio.

Thanks for the quick review. I genuinely didn't realize that the ASF projects had the guard in place. It makes sense.

So, I'm going to prepare and submit the cargo-rail-action to the ASF for review. I've got some cleanup and patches to make w/ respect to the compiler fight I'm in, but when it's green - I will submit. I don't see it taking longer than today, or maybe tomorrow. 

I'm really glad that I ran cargo-rail across Iggy. I've found a handful of simple bugs and one significant issue w/ the compiler work. These will get patched in v0.26.0.

With respect to 'unify' - yes... it's independent of the GHA. A simple `cargo binstall cargo-rail` does the trick. You can run `cargo rail unify --check` to see what's what. `unify --check` doesn't modify manifests; it exits 1 when it finds a pending dep-coherence change.

For this PR, I can either revise it to avoid the external action for now or return with the exact approved SHA after ASF review... whatever is less disruptive for you guys. TBH, I would prefer to return post ASF review. Just give me a day or two? 

Discord - no, I find it to be a time sink and I just don't have that luxury. Cargo-Rail is a side quest because my own storage engine & query engine work needed it. I say this to say... I'm just too busy. Haha. I will join, though. It's easier to communicate there, for sure.

hubcio:
no worries. you can take your time. we're almost done with cluster-related code and it's even more convenient to put non-production code further away in time, due to incoming release.

thanks for contribution!

hubcio:
moving this to draft to not trigger stale-detection bot in 7 days. ping us when done @loadingalias

loadingalias:
Hey, @hubcio. I just wanted to drop a courtesy message. This is very nearly ready. I've found and solved a lot. I should have worked in smaller chunks, but the legacy cruft and junk had piled up.

Anyway, another day or two - maximum. I don't want to leave this sitting here stale. It's in progress and really important to me.

Thanks for giving me a little time. Sorry, man.

hubcio:
@loadingalias any estimates? 😅

loadingalias:
@hubcio - tomorrow, man. It's like nearly 2AM here. I can't get a safe release cut tonight. It's just too risky. I've solved a lot, though. I'll cut the new releases for Cargo-Rail and Cargo-Rail-Action tomorrow; submit the GHA to ASF right after that.

I'm fighting Cargo's architecture at this point, but I've found a clean path forward and spent the last few days ironing it out.

I'm sorry for the delay; I genuinely appreciate the patience/check-in.

loadingalias:
Cutting the releases now!

.config/rail.toml
```diff
@@ -15,14 +15,41 @@
...
-[change-detection]
-# Changes to these paths trigger a full workspace rebuild/retest
-infrastructure = [
-    "Cargo.lock",
-    "rust-toolchain.toml",
-    ".cargo/**",
-    ".github/**",
+[plan.work.runtime-artifacts]
+scope = "cargo"
+cargo_prerequisites = [
```
hubcio, line 20: critical: a `gateways/**` change builds no `iggy-server`, so `bridge_iggy_integration_tests.rs` panics - or worse, silently runs a stale master-tip binary from the restored cache. add a `cargo_prerequisites` entry for `iggy-gateway-kafka` requiring the `iggy-server` bin.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -325,33 +260,30 @@ runs:
...
         bins_duration=$((bins_end - bins_start))
         echo "::notice::Binaries and libraries built in ${bins_duration}s ($(date -ud @${bins_duration} +'%M:%S'))"
 
         compile_start=$(date +%s)
-        if [[ -n "$PACKAGE_FLAGS" ]]; then
-          cargo test --locked --no-run $PACKAGE_FLAGS
-        else
-          cargo test --locked --no-run
-        fi
+        cargo test --locked --no-run "${TEST_CARGO_ARGS[@]}" "${TEST_TARGET_ARGS[@]}"
```
hubcio, line 269: warning: `--test <name>` cancels everything else in the run. `cargo test -p a -p b --test only_in_a` runs one target, skips `b` and all unit tests, exit 0 - so editing `core/server/tests/sdk_e2e.rs` plus a `src/` file drops `integration` and `simulator`. drop `TEST_TARGET_ARGS`.

.config/rail.toml
```diff
@@ -15,14 +15,41 @@
...
-# See: https://github.com/loadingalias/cargo-rail
-
-[change-detection]
-# Changes to these paths trigger a full workspace rebuild/retest
-infrastructure = [
-    "Cargo.lock",
-    "rust-toolchain.toml",
-    ".cargo/**",
-    ".github/**",
+[plan.work.runtime-artifacts]
```
hubcio, line 18: warning: dropping `[change-detection] infrastructure` means a `.github/**`, `.config/**`, `Cargo.lock` or `rust-toolchain.toml` change now skips the rust unit and integration suites in `test-1/2/3`. clippy, miri and the bdd fleet still cover it, but declare these paths full-scope anyway.

.config/rail.toml
```diff
@@ -15,14 +15,41 @@
...
+cargo_prerequisites = [
+    { source_work = "cargo.test", when = [
+        { package = "integration" },
+    ], require = [
+        { package = "server", target = { name = "iggy-server", kind = "bin" } },
+        { package = "iggy-cli", target = { name = "iggy", kind = "bin" } },
+        { package = "iggy-bench", target = { name = "iggy-bench", kind = "bin" } },
+        { package = "iggy-connectors", target = { name = "iggy-connectors", kind = "bin" } },
+        { package = "iggy-mcp", target = { name = "iggy-mcp", kind = "bin" } },
+        { package = "iggy_connector_clickhouse_sink", target = { name = "iggy_connector_clickhouse_sink", kind = "cdylib" } },
```
hubcio, line 29: warning: the 21 cdylib entries used to come from `cargo metadata`, now they are hand-written with nothing checking them. a new connector silently misses the build until its suite fails to load a missing `.so`.

.github/workflows/post-merge.yml
```diff
@@ -60,28 +60,56 @@ jobs:
...
+          if [[ "$PLAN_STATUS" == "success" ]] &&
+            IMAGE_MATRIX=$(cargo-rail-action plan matrix "$PLAN_FILE" edge-images); then
+            echo "::notice::Using validated Cargo-Rail edge image plan"
+          else
+            echo "::warning::Cargo-Rail plan unavailable or invalid; refreshing all edge images"
+          fi
+
+          if [[ "$IMAGE_MATRIX" == "all" ]]; then
+            DOCKER_COMPONENTS="$ALL_DOCKER_COMPONENTS"
+          elif DOCKER_COMPONENTS=$(jq -er --argjson configured "$COMPONENTS_JSON" '
```
hubcio, line 95: warning: this only checks that every selected component exists in `publish.yml`, never that every dockerhub component has a catalog row - so a sixth image never gets an `:edge` refresh. the deleted script was fail-open here via `has("gate")`.

.github/config/edge-image-variants.json
```diff
@@ -0,0 +1,117 @@
...
+        ".github/actions/utils/docker-buildx/**",
+        ".github/actions/utils/docker-login/**",
+        ".github/config/publish.yml",
+        ".github/workflows/post-merge.yml",
+        ".github/workflows/publish.yml",
+        "LICENSE",
+        "NOTICE",
+        "about.toml",
+        "about.hbs",
+        "core/server/Dockerfile",
```
hubcio, line 23: warning: these four `dockerfile:` paths are copied from `publish.yml` and nothing compares the two. move a dockerfile and the catalog keeps the old path, so edits at the new one select no row and the image quietly stops refreshing.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -272,44 +180,71 @@ runs:
...
+          TEST_PACKAGES=()
+          RUNTIME_PACKAGES=()
+          echo "::warning::Cargo-Rail plan unavailable or invalid; running the full workspace"
         fi
-        if [[ -f /tmp/plugin-suites.txt ]]; then
-          PLUGIN_SUITES=$(cat /tmp/plugin-suites.txt)
+
+        if [[ "$PLAN_USABLE" == true && "$TEST_SCOPE" == "skipped" ]]; then
+          echo "::notice::Cargo-Rail reports no affected Rust tests"
+          exit 0
```
hubcio, line 226: nit: this exit skips `cargo llvm-cov report`, so the codecov step logs an error for a missing `codecov.json`. harmless given `carryforward: true` and `fail_ci_if_error: false`, but it is noise on every skipped run.

.github/workflows/post-merge.yml
```diff
@@ -60,28 +60,56 @@ jobs:
...
         run: |
-          chmod +x scripts/extract-version.sh
+          set -euo pipefail
+
+          COMPONENTS_JSON=$(yq -o=json -I=0 '.components' .github/config/publish.yml)
+          ALL_DOCKER_COMPONENTS=$(jq -r \
+            '[to_entries[] | select(.value.registry == "dockerhub") | .key] | join(",")' \
+            <<< "$COMPONENTS_JSON")
+
+          IMAGE_MATRIX=all
```
hubcio, line 85: nit: line 87 overwrites this `all` default with the selector's stdout even when it exits non-zero. fail-open still holds because the empty value fails both gates, but reset `IMAGE_MATRIX=all` in the `else` branch.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -272,44 +180,71 @@ runs:
...
+
+        if [[ "$PLAN_STATUS" == "success" ]] &&
+          TEST_SCOPE=$(cargo-rail-action plan cargo-scope "$PLAN_FILE" cargo.test) &&
+          RUNTIME_SCOPE=$(cargo-rail-action plan cargo-scope "$PLAN_FILE" runtime-artifacts) &&
+          cargo-rail-action plan cargo-args "$PLAN_FILE" cargo.test > "$PLAN_OUTPUT_DIR/test-args" &&
+          cargo-rail-action plan target-args "$PLAN_FILE" cargo.test > "$PLAN_OUTPUT_DIR/test-target-args" &&
+          cargo-rail-action plan package-names "$PLAN_FILE" cargo.test > "$PLAN_OUTPUT_DIR/test-packages" &&
+          cargo-rail-action plan package-names "$PLAN_FILE" runtime-artifacts > "$PLAN_OUTPUT_DIR/runtime-packages"; then
+          mapfile -d '' -t TEST_CARGO_ARGS < "$PLAN_OUTPUT_DIR/test-args"
+          mapfile -d '' -t TEST_TARGET_ARGS < "$PLAN_OUTPUT_DIR/test-target-args"
```
hubcio, line 210: nit: `mapfile -d ''` needs NUL-separated selector output, and the comment above names the validation but not the separator. worth a word in that comment.

.github/workflows/post-merge.yml
```diff
@@ -60,28 +60,56 @@ jobs:
...
+            echo "::warning::Cargo-Rail plan unavailable or invalid; refreshing all edge images"
+          fi
+
+          if [[ "$IMAGE_MATRIX" == "all" ]]; then
+            DOCKER_COMPONENTS="$ALL_DOCKER_COMPONENTS"
+          elif DOCKER_COMPONENTS=$(jq -er --argjson configured "$COMPONENTS_JSON" '
+            [.include[].component] as $selected
+            | if ($selected | all(. as $component | $configured[$component].registry == "dockerhub"))
+              then $selected | join(",")
+              else error("Cargo-Rail selected an unknown Docker component")
```
hubcio, line 99: nit: this error passes a constant string, so the log names neither the rejected component nor the matrix. interpolate `$component` into it.

.github/config/publish.yml
```diff
@@ -55,13 +55,6 @@ components:
     platforms: ["linux/amd64", "linux/arm64"]
```
hubcio, line 55: nit: nothing here points at `.github/config/edge-image-variants.json`, which decides whether this image gets an `:edge` refresh. the removed `gate:` block carried that link - add a comment.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -272,44 +180,71 @@ runs:
...
+          mapfile -d '' -t TEST_PACKAGES < "$PLAN_OUTPUT_DIR/test-packages"
+          mapfile -d '' -t RUNTIME_PACKAGES < "$PLAN_OUTPUT_DIR/runtime-packages"
+          PLAN_USABLE=true
+        else
+          TEST_SCOPE=workspace
+          RUNTIME_SCOPE=workspace
+          TEST_CARGO_ARGS=()
```
hubcio, line 217: simplification: these four array resets are dead - only the `mapfile` calls in the success branch ever fill them. keep the two scope resets, they do matter.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -272,44 +180,71 @@ runs:
...
-        if [[ -f /tmp/test-packages.txt ]]; then
-          TEST_PACKAGE_FLAGS=$(cat /tmp/test-packages.txt)
-        fi
-        if [[ -f /tmp/total-crates.txt ]]; then
-          TOTAL_CRATES=$(cat /tmp/total-crates.txt)
+        # Read only validated selectors through the v10 runtime. Any planner or
+        # selector failure restores the previous full-workspace behavior.
+        PLAN_OUTPUT_DIR="${RUNNER_TEMP:?RUNNER_TEMP is required}/cargo-rail-selectors-${TASK}"
+        mkdir -p "$PLAN_OUTPUT_DIR"
+        PLAN_USABLE=false
```
hubcio, line 194: simplification: `PLAN_USABLE` is redundant - the `else` branch resets `TEST_SCOPE=workspace`, so testing `TEST_SCOPE` alone says the same thing, which lines 279 and 343 already do.

also at lines 224, 231, 244, 371.

.github/actions/rust/pre-merge/action.yml
```diff
@@ -272,44 +180,71 @@ runs:
...
-          # Each connector suite contributes its own `package(integration)`, so
-          # subtract them to keep the crate count a crate count.
-          TEST_CRATE_COUNT=$(( $(echo "$NEXTEST_FILTER" | grep -o 'package(' | wc -l) - PLUGIN_SUITES ))
-          BUILD_CRATE_COUNT=$(echo "$PACKAGE_FLAGS" | grep -o '\-p ' | wc -l)
-          echo "::notice::DAG-scoped: testing ${TEST_CRATE_COUNT} crates + ${PLUGIN_SUITES} connector suites, building ${BUILD_CRATE_COUNT} (cargo check/clippy cover full workspace separately)"
+        BUILD_ARGS=()
+        BUILD_SCOPE=workspace
+        if [[ "$PLAN_USABLE" == true && "$TEST_SCOPE" == "packages" &&
+          ( "$RUNTIME_SCOPE" == "packages" || "$RUNTIME_SCOPE" == "skipped" ) ]]; then
+          declare -A SEEN_BUILD_PACKAGES=()
```
hubcio, line 233: simplification: three structures track one package set here. collapse to one `mapfile` with an `awk 'NF && !seen[$0]++'` dedupe - keep the dedupe, `iggy-cli` lands in both lists and line 370 reads the count.

.github/workflows/post-merge.yml
```diff
@@ -60,28 +60,56 @@ jobs:
...
+              end
+          ' <<< "$IMAGE_MATRIX"); then
+            :
```
hubcio, line 102: simplification: the `elif` body is a bare `:`. negate it instead - `elif ! DOCKER_COMPONENTS=$(jq -er ...)` - and keep the warning body.

hubcio (changes requested):
Two findings sit outside this diff, so they are here instead of inline.

`scripts/ci/coverage-baseline-affected.sh:77` - a `.config/**` change runs no rust test, before or after merge. line 59 covers `.github`, `scripts` and `codecov.yml`, line 77 lists `core gateways Cargo.toml Cargo.lock rust-toolchain.toml .cargo`. neither names `.config`, so an edit to `rail.toml` itself is untested.

`.github/workflows/publish.yml:244` - a third hard-coded copy of the dockerhub component list. a sixth image needs three edits and two of them fail quietly. pre-existing, but it raises the cost of the two catalog findings above.

One thing that does work: the reverse `cargo_prerequisites` edge. server-only and plugin-only PRs do still select `integration`, because `prerequisite_source_roots` pulls the `when` roots back into the test scope. The gap is only for packages not named in `require`, which is why `gateways/**` is the one blocker here.

PR #4000, opened 2026-08-31
feat(python): add WebSocketConfig transport configuration

saie-ch:

Relates to #2835.

**Rationale**

  Adds WebSocket transport support to the Python SDK, following the same <Transport>Config pattern already used for TCP.

  **What changed?**

  Adds WebSocketConfig, WebSocketReconnectionConfig, and WebSocketFramingConfig, plus IggyClient.websocket(config). Mirrors TcpConfig's fields and validation.

  **Local Execution**

  - Passed
  - Pre-commit hooks ran

  **AI Usage**
Claude Sonnet 5 (Claude Code)


examples/python/websocket/consumer.py
slbotbm, line 1: No need for examples like these. You can add comments to the existing examples on how to configure different transports instead.

examples/python/websocket/producer.py
slbotbm, line 1: Not needed.

foreign/python/src/config.rs
```diff
@@ -408,10 +411,465 @@ impl TcpConfig {
...
+            inner.read_buffer_size = Some(usize_param(read_buffer_size, "read_buffer_size")?);
+        }
+        if let Some(write_buffer_size) = write_buffer_size {
+            inner.write_buffer_size = Some(usize_param(write_buffer_size, "write_buffer_size")?);
+        }
+        if let Some(max_write_buffer_size) = max_write_buffer_size {
+            inner.max_write_buffer_size =
+                Some(usize_param(max_write_buffer_size, "max_write_buffer_size")?);
+        }
+        if let Some(max_message_size) = max_message_size {
```
slbotbm, line 594: Passing `None` for `max_message_size` or `max_frame_size` cannot disable the limit as documented. An omitted argument and an explicit Python `None` both arrive as `Option<i64>::None`, the constructor starts from the Rust defaults of 64 MiB and 16 MiB, and these assignments run only for `Some`. The Rust `to_tungstenite_config()` conversion also leaves tungstenite's default in place when its field is `None`. A caller that requests no limit will therefore still have large messages or frames rejected. Let's distinguish an omitted argument from an explicit `None` and make the Rust conversion call the tungstenite setters with `None`.

slbotbm (dismissed):
Other than the comments below, the API exposed in this PR does not match the established pattern, which is `IggyClient(server_address: str | TcpConfig | ...)`. Let's conform to that pattern instead of exposing `.websocket` method.

examples/python/websocket/consumer.py
hubcio: Removed. `examples/python/websocket/` is gone from the branch. `examples/python/getting-started/{consumer,producer}.py` now carry a commented `WebSocketConfig` alternative next to the existing QUIC and HTTP ones, so switching transport there is a one-line uncomment.

examples/python/websocket/producer.py
hubcio: Removed along with the consumer example, see the reply on that thread.

foreign/python/src/config.rs
hubcio: Fixed on both ends, and checked against the built extension rather than by reading.

The pyo3 signature now carries `max_message_size=64 << 20` and `max_frame_size=16 << 20`, so an omitted argument arrives as `Some(default)` and only an explicit `None` arrives as `None`. `to_tungstenite_config()` calls both setters unconditionally and passes the `Option` straight through.

Measured on the built module:

```
F()                                            -> 67108864, 16777216
F(max_message_size=None, max_frame_size=None)  -> None, None
F(max_message_size=None)                       -> None, 16777216
```

The last hop is covered by `none_size_limits_should_reach_tungstenite_as_no_limit` in `websocket_client_config.rs`, and `websocket_client.rs:637` and `:689` are the two connect paths that call `to_tungstenite_config()`. `defaults_should_match_the_sdk` in `config.rs` pins the two signature literals to the SDK, so a tungstenite bump cannot desync them silently.

numinnex (approved):

hubcio (approved):

PR #4004, opened 2026-08-31
docs(repo): add repo-wide team-review agent skill

hubcio:
This PR adds a new skill used to review changes.
Contributors are encouraged to use it.

It is placed under .claude/skills/ next to the connector skills.
The report lands in the agent session scratchpad.

Skills carry no disable-model-invocation flag, so the
description states the skill is user-invoked only: one run
spawns roughly ten subagents.

### DIFF
--- a/.claude/skills/team-review/SKILL.md
+++ b/.claude/skills/team-review/SKILL.md
@@ -15,7 +15,7 @@ You = **moderator**. You never open the diff or a source file: you route paths, 
 
 > You think big brain. You speak caveman. Separate things.
 >
-> **Thinking, unchanged.** Read the diff, then every changed file in full from the local checkout, then whatever call sites you need. Trace call chains. Verify invariants. Prove findings, don't guess. Cite exact `file:line`.
+> **Thinking, unchanged.** Read the diff, then every changed file in full from the local checkout, then whatever call sites you need. Trace call chains. Verify invariants. Prove findings, don't guess. Cite exact `file:line`. Running tests or builds needs a stated justification: reading and tracing settles most claims, and parallel cargo runs block on one target-dir lock.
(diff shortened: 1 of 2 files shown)


.claude/skills/team-review/SKILL.md
```diff
@@ -0,0 +1,113 @@
+---
+name: team-review
+description: Adversarial 4-expert review (storage, perf, distsys, ecosystem) of a PR, branch, or ref range, with clean-room validation of every finding. Experts work alone, no peer debate. Invoke only when the user explicitly asks for a team review: it spawns ~10 subagents and is expensive.
+---
```
lukaszzborek, line 6: add `disable-model-invocation: true`
this skill should be run only via explicit user command. it can use a lot of tokens 😅
https://code.claude.com/docs/en/skills#frontmatter-reference

ryankert01: if added `disable-model-invocation: true`, which excludes the skill's description from claude's context at all, we should prabably not include it in `Agents.md` or need to rephrase what's in `AGENTS.md` to `after tests pass, suggest running /team-review` etc.

hubcio: done in https://github.com/apache/iggy/commit/3ff7671b81a4b7e2cc3b6667e513007b744f7f2b - flag set, and the description dropped the "invoke only when asked" prose since the flag enforces it. kept the AGENTS.md entry so contributors can find it, but reworded to user-invoked only, never replicate its steps, suggest /team-review <target> once a non-trivial change passes verification, and dropped it from the top load note.

numinnex (approved):

spetz (approved):

PR #4006, opened 2026-08-31
fix(shard): drive partition repair and the commit walk from the tick

numinnex:
Resident committed ops re-drive commit_journal directly, undebounced but capped at sixteen walks per tick, since each walk reaches a segment flush and a node-wide rejoin makes every group on the shard owe one in the same tick. The cap cannot starve a group: the sweep resumes at the first one it turned away, so the owed set drains round-robin.


numinnex:
Merge after #4005

core/shard/src/lib.rs
```diff
@@ -7041,6 +7061,92 @@ where
...
+                // walking them, so this is the steady state for the whole
+                // duration of a rejoin and would be one line per group per tick.
+                tracing::debug!(
+                    shard = self.id,
+                    namespace_raw = namespace.inner(),
+                    commit_min = consensus.commit_min(),
+                    commit_max = consensus.commit_max(),
+                    "partition commit walk parked over resident committed ops; resuming"
+                );
+                partition.commit_journal(config).await;
```
spetz: `tick_partitions` samples `partition.fatal()` before this call, but `commit_journal()` can set the partition fatal during the tick-owned walk. The function then returns the stale fatal aggregate, allowing the router to resume inbox processing after a local durability failure. Please re-read `partition.fatal()` immediately after `commit_journal()` and propagate it before transfer work. Please also add an injected journal-failure test proving this tick returns `FatalCommit`.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7061,92 @@ where
...
+            };
+
+            // Capped like the repair arm, and for the same reason: a node-wide
+            // rejoin leaves every group on the shard walk-stalled in the same
+            // tick, and each walk reaches a segment flush. Undebounced, though
+            // -- the predicate guarantees the walk finds at least the next op
+            // (it and `collect_committable_from_journal` read the same journal
+            // index), so it cannot spin, and a skipped group is re-evaluated
+            // unchanged next tick.
+            if walk_stalled && walks < PARTITION_WALKS_PER_TICK_MAX {
```
spetz: The commit-walk cap is reset on every tick, while namespaces are visited in stable `BTreeMap` order. If the first 16 namespaces remain eligible, they consume the entire budget on every pass and later namespaces can retain resident committed operations indefinitely. This also conflicts with the stated design that local owed commit work is uncapped. Please remove the commit-walk cap or add persistent round-robin scheduling, with a sustained test covering more than 16 eligible partitions.

core/shard/src/lib.rs
```diff
@@ -6890,9 +6892,27 @@ where
...
+        // `PARTITION_WALKS_PER_TICK_MAX`.
+        let mut walks = 0usize;
 
         let mut fatal: Option<FatalCommit> = None;
         for namespace in namespace_scratch.drain(..) {
+            {
+                // Ahead of the fence check and every `continue` below: the
+                // count is the only record those prepares existed, and a
+                // partition that fences here never ticks again.
+                let Some(partition) = partitions.get_mut_by_ns(&namespace) else {
```
spetz: The gap-drop metric can lose events during partition removal. The counter is buffered on the partition, but `get_mut_by_ns` stops returning it once the namespace is tombstoned, and `ConfirmRemove` then drops the accumulator. Please either increment the shard-owned metric synchronously when the prepare is dropped or explicitly drain the counter during tombstoning or removal. A drop-then-delete test would cover this path.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7061,92 @@ where
...
+                        tracing::debug!(
+                            shard = self.id,
+                            namespace_raw = namespace.inner(),
+                            commit_min,
+                            commit_max,
+                            peer,
+                            recovery = ?probe.recovery,
+                            "partition gap-stopped past the debounce; arming repair from the primary"
+                        );
+                        self.maybe_request_partition_repair(partition, peer).await;
```
spetz: A successful repair arm leaves `gap_ticks` saturated. If that repair completes before the next sweep and another gap appears, the new gap inherits the old count and bypasses the intended debounce. Please reset `gap_ticks` when an arm is accepted, while retaining saturation when the per-tick arm cap merely defers the repair. Please add a back-to-back short-repair test

core/shard/src/lib.rs
```diff
@@ -7041,6 +7061,92 @@ where
...
+            let walk_stalled = {
+                let Some(partition) = partitions.get_mut_by_ns(&namespace) else {
+                    continue;
+                };
+                let probe = partition_gap_probe(partition);
+                let walk_stalled = partition_is_walk_stalled(&probe);
+                if drive_partition_gap_debounce(
+                    &probe,
+                    &mut partition.gap_ticks,
+                    repair_retry_ticks,
```
spetz: `repair_retry_interval` now also affects initial tick-driven gap detection and is subject to a 50-tick minimum. The canonical configuration documentation currently describes only stalled repair-stream retries. Please document the new behavior and effective minimum so operators can predict repair latency.

spetz (approved):

spetz (changes requested):
As commented before

core/partitions/src/iggy_partition.rs
```diff
@@ -3159,7 +3213,7 @@ where
             return;
         }
 
-        let journal_holds_op = self.log.journal().inner.header_by_op(header.op).is_some();
+        let journal_holds_op = self.log.journal().inner.holds_op(header.op);
```
hubcio, line 3246: warning: `apply_repaired_prepare` at line 5839 still uses `header_by_op(..).is_some()`, the scan this line replaces, on the repair ingest path where a miss walks the whole vec. switch it to `holds_op`.

also at `core/simulator/src/lib.rs:5329`.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7094,120 @@ where
...
+                    // whole duration of a rejoin and would be one line per group
+                    // per tick.
+                    tracing::debug!(
+                        shard = self.id,
+                        namespace_raw = namespace.inner(),
+                        commit_min = consensus.commit_min(),
+                        commit_max = consensus.commit_max(),
+                        "partition commit walk parked over resident committed ops; resuming"
+                    );
+                    partition.commit_journal(config).await;
```
hubcio, line 7304: warning: this walk reaches `committed_headers_from`, which probes the linear `header_by_op` once per op - so it is O(window x headers), not the single scan its doc claims. do one pass over `headers` like `repaired_window_shape` does.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7094,120 @@ where
...
+                let walk_stalled = partition_is_walk_stalled(&probe);
+                match drive_partition_gap_debounce(
+                    &probe,
+                    &mut partition.gap_ticks,
+                    repair_retry_ticks,
+                    repair_arms,
+                ) {
+                    GapArm::NotDue => {}
+                    GapArm::Deferred => {
+                        deferred_cursor.get_or_insert_with(|| namespace.inner());
```
hubcio: warning: one cursor, two caps with different budgets - a repair deferral here pins the next sweep's start, so the walk re-spends on the same prefix. the `ceil(groups / cap)` bound at 9495 does not hold. one cursor per cap fixes it.

core/shard/src/lib.rs
```diff
@@ -10519,3 +10926,431 @@ mod superblock_fail_stop_tests {
...
+        (0..GROUPS)
+            .map(|partition| IggyNamespace::new(1, 1, partition))
+            .collect()
+    }
+
+    /// One sweep: rotate to `cursor`, walk the snapshot spending at most
+    /// `PARTITION_WALKS_PER_TICK_MAX` walks, and return the cursor the sweep
+    /// leaves behind. Models `tick_partitions` over a shard where EVERY group is
+    /// eligible on every pass, which is the shape the cap starves.
+    fn sweep(cursor: u64, served: &mut [u32]) -> u64 {
```
hubcio: warning: this models one cap, but production shares a cursor between the 3-arm and 16-walk budgets. so the `most - fewest <= 1` assert below claims a property `tick_partitions` does not have. model both budgets.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7094,120 @@ where
...
+                    GapArm::Deferred => {
+                        deferred_cursor.get_or_insert_with(|| namespace.inner());
+                    }
+                    GapArm::Arm => {
+                        let consensus = partition.consensus();
+                        let self_id = consensus.replica();
+                        let peer = consensus.primary_index(consensus.view());
+                        let commit_min = consensus.commit_min();
+                        let commit_max = consensus.commit_max();
+                        if peer == self_id {
```
hubcio, line 7235: warning: a gap-stopped replica that is itself the view primary has no way out - every arming site needs an inbound frame or a transfer peer, so this warns every debounce forever. once a new op commits in that state, `advance_commit_min` hits its `assert_eq!` and panics the pump. rotate to another live replica.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+/// arm is a `RequestPrepares` plus a repair stream the serving peer walks
+/// synchronously and a node-wide gap (a rejoin, a lossy link) makes every group
+/// on this shard due in the same tick. Over-cap groups stay due, and the next
+/// sweep resumes at the first of them ([`rotate_sweep_to_cursor`]), so they arm
+/// on the following pass rather than queueing behind the same prefix.
+const PARTITION_REPAIR_ARMS_PER_TICK_MAX: usize = 3;
+
+/// Commit walks the partition tick sweep will RUN per pass.
+///
+/// Same correlated-fan-out argument as the repair arm, and the walk is the
```
hubcio, line 9649: warning: this caps walk count, not work - `collect_committable_from_journal` returns the whole `(commit_min, commit_max]` run and `handle_committed_entries` walks all of it on the pump. after a rejoin that is one uninterruptible tick. bound ops or bytes per walk and resume through the cursor.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+/// A RATE cap, not a concurrency one: unlike
+/// `IggyShard::PARTITION_TRANSFERS_INFLIGHT_MAX`, which counts live sessions
+/// and refuses over the count, nothing bounds how many partition repair
+/// sessions are live at once. This only spreads the opening cost, because one
+/// arm is a `RequestPrepares` plus a repair stream the serving peer walks
+/// synchronously and a node-wide gap (a rejoin, a lossy link) makes every group
+/// on this shard due in the same tick. Over-cap groups stay due, and the next
+/// sweep resumes at the first of them ([`rotate_sweep_to_cursor`]), so they arm
+/// on the following pass rather than queueing behind the same prefix.
+const PARTITION_REPAIR_ARMS_PER_TICK_MAX: usize = 3;
```
hubcio, line 9628: warning: rate cap only - nothing bounds live repair sessions, so a node-wide rejoin puts every group's stream in flight within `groups / 3` ticks and the serving peer walks each on its pump. count live sessions like `PARTITION_TRANSFERS_INFLIGHT_MAX` does.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+
+/// Commit walks the partition tick sweep will RUN per pass.
+///
+/// Same correlated-fan-out argument as the repair arm, and the walk is the
+/// costlier half: `commit_journal` reaches `commit_messages`, which flushes a
+/// segment and fsyncs under `enforce_fsync`. Run serially over every group on
+/// the shard in one tick body, a node-wide rejoin exceeds the view-change
```
hubcio: warning: the superblock pre-pass analogy does not hold - it runs its 16 concurrently under `join_all` and drains everything in one body, while these 16 walks are serial and deferred. size the cap from a real flush latency instead.

core/partitions/src/iggy_partition.rs
```diff
@@ -3843,6 +3898,10 @@ where
     }
 
     async fn commit_messages(&mut self, config: &PartitionsConfig) -> Result<(), IggyError> {
+        #[cfg(any(test, feature = "simulator"))]
```
hubcio: warning: this puts a fault-injection branch in the durability path, and feature unification means it ships - the release build at `_build_rust_artifacts.yml:152` has no package flag, so `partitions/simulator` is on. move it behind a feature only `core/simulator`'s dev-dependencies request.

core/configs/src/server_config/cluster.rs
```diff
@@ -264,6 +264,16 @@ pub struct ClusterConfig {
     /// partition repair loops. Sizes the retry threshold in consensus ticks.
     /// Zero (and the `0` / `disabled` / `unlimited` sentinels, which all parse
     /// to zero) is rejected at boot.
+    ///
+    /// It also sizes the partition sweep's gap debounce: how long a backup must
```
hubcio: warning: this knob now gates how long a replication hole stays open, not just retry pacing, and `shell.rs:204` clamps at `u32::MAX`. someone who raises it to quiet repair chatter widens the gap window by the same amount - give the debounce its own setting.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+/// Whether this replica holds committed ops it cannot walk to, because the op
+/// one past its commit frontier is missing from its journal.
+///
+/// The journal-hole half is not redundant: a follower advances `commit_max`
+/// from every prepare header in `replicate_preflight`, so `commit_min <
+/// commit_max` is transiently true on every healthy pipelined tick and a bare
+/// lag test would arm repair against ordinary produce.
+const fn partition_is_gap_stopped(probe: &GapProbe) -> bool {
+    probe.normal
+        && !probe.transferring
```
hubcio: warning: `commit_min < commit_max` means the detector misses the missing-suffix shape, where headers were adopted but bodies never arrived and both marks sit below `head`. that class still hangs on the single `on_start_view` edge. add a `repaired_window_shape(commit_max, head)` term.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+        debounce_ticks
+    };
+    *gap_ticks = gap_ticks.saturating_add(1);
+    if *gap_ticks < debounce_ticks {
+        return GapArm::NotDue;
+    }
+    if arms_this_tick >= PARTITION_REPAIR_ARMS_PER_TICK_MAX {
+        return GapArm::Deferred;
+    }
+    *gap_ticks = 0;
```
hubcio: warning: only sweep-armed repairs clear `gap_ticks` - the four edge-triggered sites never touch it, so a short edge-armed repair leaves the count saturated and the next real gap arms with no debounce. reset it in `maybe_request_partition_repair`, which already holds `&mut IggyPartition`.

core/server/src/partition_reconciler.rs
```diff
@@ -109,9 +109,21 @@
...
+//! post-transfer tail), whose edges a produce stream can starve. A repair range
+//! the primary has already evicted escalates to partition state transfer. The park policy
+//! above still shrinks the exposure to a genuinely exhausted byte budget and a
+//! namespace this shard cannot serve; the driver bounds how long either costs,
+//! and `partition_prepare_gap_drops_total` counts what reached the gap check.
+//!
 //! # Known gaps
 //!
-//! Recorded here because both were previously carried as a TODO on the
+//! Recorded here because they were previously carried as a TODO on the
```
hubcio, line 126: warning: the known-gaps entry just below still names the commit heartbeat as the recovery path for a shed prepare, which the paragraph you added above says is starvable. delete it or restate it as the residual debounce exposure.

core/server/config.toml
```diff
@@ -656,6 +656,19 @@ view_probe_attempts_max = 5
...
+#
+# Also how long a gap-stopped partition backup waits before OPENING a repair
+# session: the shard's tick sweep arms repair once a group has held committed
+# ops it cannot walk to for this long, so shortening the interval makes a
+# replication hole recover sooner and lengthening it delays that recovery, on
+# top of the stalled-stream retry above.
+#
+# The second use is floored at 50 consensus ticks (500ms at the 10ms tick), so
+# values under that still arm no sooner: one tick of lag is ordinary
+# pipelining, and repair against it would fire on healthy traffic. Repair
```
hubcio: warning: this latency bound ignores the 3-arms-per-tick cap - on the node-wide rejoin the cap exists for, the Nth due group waits `ceil(N/3)` ticks past the stated max. mention the cap here or give a real bound.

core/simulator/src/lib.rs
```diff
@@ -5106,6 +5106,837 @@ mod repair_frontier_tests {
...
+        // distinction a naive detector would miss.
+        //
+        // On a two-replica group the backup keeps pace and the tick's own
+        // walk-stalled backstop drains what little lag a step leaves, so this
+        // usually samples none and the check costs nothing. It is the guard for
+        // when that stops being true, not the proof: the predicate itself is
+        // pinned exhaustively by `gap_detector_tests`, and what this run adds is
+        // that the DRIVER sends no repair over a live, loaded group.
+        assert_eq!(
+            observed.samples,
```
hubcio: warning: nothing makes the backup lag, so `samples` is usually 0 and this assert is vacuous - the repair count below then passes for an unrelated reason. add `assert!(observed.samples > 0)` or drive the lag on purpose.

core/partitions/src/journal.rs
```diff
@@ -762,6 +762,20 @@ where
...
     }
 
+    /// Whether `op` is resident, without reading its header.
+    ///
+    /// Equivalent to `header_by_op(op).is_some()` and answers in O(log n)
+    /// instead of scanning: `append` writes `headers` and
+    /// `op_to_storage_offset` together and every clear site
+    /// ([`Self::commit`], [`Self::evict_prefix`], the restore path) clears
+    /// both, so the two populations cannot diverge. Callers that only need
+    /// presence must use this - a miss is the common case on the residency
```
hubcio: nit: the doc says `append` writes both maps together, but `append_with_meta` pushes the header before the fallible write and inserts the offset after. `append_with_meta`'s own note at 625 already states the precondition - point at it from here.

core/shard/src/metrics.rs
```diff
@@ -447,6 +452,34 @@ impl ShardMetrics {
...
+    /// family means the bus or the router shed a frame, and the simulator
+    /// asserts it stays at zero on runs with no injected loss. A gap drop is a
+    /// protocol-ordering drop that any real loss produces, and the sweep repairs
+    /// it, so folding the two would turn a routing-fault alert into noise.
+    ///
+    /// Shard-scoped, with no namespace label: a server runs hundreds of groups
+    /// per shard, so labelling by namespace is unbounded cardinality, and every
+    /// other partition counter in this file is shard-scoped for the same
+    /// reason. The per-group detail is in the arm's log line. The metadata
+    /// plane's own gap drop is NOT counted here - it has its own recovery path.
```
hubcio: nit: 'it has its own recovery path' oversells it - the metadata plane's repair is edge-triggered and starvable in the same way, and it has no gap-drop counter at all. say that instead.

core/partitions/src/iggy_partition.rs
```diff
@@ -1710,6 +1728,42 @@ where
         self.write_superblock_advancing(superblock, 0, claim).await
     }
 
+    /// Record one prepare destroyed by the backup gap check. Drained by the
+    /// shard sweep into `partition_prepare_gap_drops_total`; the counter is the
```
hubcio: nit: 'the only production signal that the driver has work to do' is too strong - the counter records the prepares that arrived after the hole, so a gap opened by the last prepare of a burst leaves it at zero. `metrics.rs:88` words it right.

core/partitions/src/iggy_partition.rs
```diff
@@ -3255,11 +3309,12 @@ where
                     tracing::Level::WARN,
                     &PartitionDiagEvent::new(
```
hubcio, line 3344: nit: this event carries `header.op` but not `current_op`, so a forward gap and a retransmit of an adopted suffix op look identical in the log and share one counter. add the sequencer to the event.

core/partitions/src/iggy_partition.rs
```diff
@@ -3255,11 +3309,12 @@ where
...
                         self.diag_ctx(),
-                        "dropping out-of-order prepare (gap)",
+                        "dropping out-of-order partition prepare (gap)",
```
hubcio: nit: the test already separates the planes with `PARTITION_PLANE_FIELD`, which the metadata line never emits, so the rename buys nothing and breaks any operator alert on the old string. keep the string, keep the comment.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7094,120 @@ where
...
+                    }
+                }
+                walk_stalled
+            };
+
+            // Capped like the repair arm, and for the same reason: a node-wide
+            // rejoin leaves every group on the shard walk-stalled in the same
+            // tick, and each walk reaches a segment flush. Undebounced, though
+            // -- the predicate guarantees the walk finds at least the next op
+            // (it and `collect_committable_from_journal` read the same journal
```
hubcio: nit: not the same index - `partition_is_walk_stalled` reads `holds_op` off the `op_to_storage_offset` map, `committed_headers_from` scans the `headers` vec. the anti-spin argument rests on both being written and cleared together, so say that.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+/// without a floor of its own an operator shortening the retry interval would
+/// also arm repair against a single reordered prepare.
+///
+/// Public because it bounds what that operator knob can do: gap recovery starts
+/// after `max(repair_retry_interval, this)`, which the `[cluster]`
+/// documentation states.
+pub const PARTITION_GAP_DEBOUNCE_TICKS_MIN: u32 = 50;
+
+/// What already owns a partition's recovery, if anything.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
hubcio: simplification: `RecoveryOwner`'s three non-`Nobody` variants are unobservable - the predicate only asks `matches!(.., Nobody)` and the log field at 7151 sits where the probe is always `Nobody`. a `recovery_owned: bool` drops ~40 lines including the test loops.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7094,120 @@ where
...
+                            tracing::warn!(
+                                shard = self.id,
+                                namespace_raw = namespace.inner(),
+                                commit_min,
+                                commit_max,
+                                "partition primary is gap-stopped below its own commit frontier; \
+                                 repair has no peer to request from"
+                            );
+                        } else {
+                            tracing::debug!(
```
hubcio: simplification: `maybe_request_partition_repair` runs next and always logs the same fields at info, since the `Arm` probe implies every guard in it. drop this debug line.

core/shard/src/lib.rs
```diff
@@ -6890,9 +6916,36 @@ where
...
+        // cursor that stood still would re-skip the same tail forever.
+        let mut deferred_cursor: Option<u64> = None;
 
         let mut fatal: Option<FatalCommit> = None;
         for namespace in namespace_scratch.drain(..) {
+            {
+                // Ahead of the fence check and every `continue` below: the
+                // count is the only record those prepares existed, and a
+                // partition that fences here never ticks again.
+                let Some(partition) = partitions.get_mut_by_ns(&namespace) else {
```
hubcio: simplification: `get_mut_by_ns` here and `get_by_ns` eight lines down look up the same group twice per tick. make `prepare_gap_drops` a `Cell<u64>` like `superblock_write_failures`, or just reborrow this `&mut` for the reads below.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7094,120 @@ where
...
+                ) {
+                    GapArm::NotDue => {}
+                    GapArm::Deferred => {
+                        deferred_cursor.get_or_insert_with(|| namespace.inner());
+                    }
+                    GapArm::Arm => {
+                        let consensus = partition.consensus();
+                        let self_id = consensus.replica();
+                        let peer = consensus.primary_index(consensus.view());
+                        let commit_min = consensus.commit_min();
```
hubcio: simplification: `probe` already carries both, sampled a few lines up with no await between. use `probe.commit_min` and `probe.commit_max` so the log cannot drift from what armed.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9469,230 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+fn rotate_sweep_to_cursor(namespaces: &mut [IggyNamespace], cursor: u64) {
+    if cursor == 0 {
+        return;
+    }
+    debug_assert!(
+        namespaces.is_sorted(),
+        "the sweep snapshot must be in namespace order for the cursor to resume in it",
+    );
+    let resume_at = namespaces.partition_point(|namespace| namespace.inner() < cursor);
+    if resume_at < namespaces.len() {
```
hubcio: simplification: `rotate_left` only panics on `mid > len`, and `mid == len` early-returns as a no-op. drop the `if` and call it unconditionally.

core/shard/src/lib.rs
```diff
@@ -2413,7 +2424,17 @@ where
...
                     self.metrics.record_partition_removed();
                     confirmed_remove = true;
-                    if removed.is_none() {
+                    if let Some(mut partition) = removed {
+                        // Tail of the gap-drop count. The tick sweep drains it
+                        // per pass, but `get_mut_by_ns` stops answering the
+                        // moment the reconciler tombstones the namespace, so
+                        // whatever the last pass before the tombstone left
+                        // would go to the floor with the partition value.
+                        let gap_drops = partition.take_prepare_gap_drops();
```
hubcio: simplification: `partition_prepare_gap_drops_total` is a plain counter, so `inc_by(0)` is a no-op that creates no series. drop the guard here - keep the one in the sweep, which runs per group per tick.

core/shard/src/lib.rs
```diff
@@ -1486,6 +1486,15 @@ where
...
     repair_retry_ticks: Cell<u32>,
 
+    /// Namespace the next partition sweep starts from: the first group a
+    /// per-tick cap turned away last pass, or 0 to start at the front.
+    ///
+    /// The sweep visits namespaces in `BTreeMap` order, so without a carried
+    /// cursor the leading groups would spend the whole per-tick budget on every
+    /// pass and the tail would never be reached. See
+    /// [`rotate_sweep_to_cursor`].
+    partition_sweep_cursor: Cell<u64>,
```
hubcio: simplification: `0` doubles as 'nobody deferred' and as the legal namespace `(0, 0, 0)`. `Cell<Option<IggyNamespace>>` makes that unrepresentable and drops the sentinel branch in `rotate_sweep_to_cursor`.

core/partitions/src/iggy_partition.rs
```diff
@@ -1710,6 +1728,42 @@ where
         self.write_superblock_advancing(superblock, 0, claim).await
     }
 
+    /// Record one prepare destroyed by the backup gap check. Drained by the
+    /// shard sweep into `partition_prepare_gap_drops_total`; the counter is the
+    /// only production signal that the level-triggered repair driver has work
+    /// to do, since the drop itself answers nobody.
+    const fn note_prepare_gap_drop(&mut self) {
```
hubcio: simplification: one-line private helper with one caller. inline the `saturating_add` at 3317 and fold the metric name and the driver tie into the field doc at 162.

core/simulator/src/lib.rs
```diff
@@ -5106,6 +5106,837 @@ mod repair_frontier_tests {
...
+    fn given_a_backup_that_dropped_a_committed_prepare_when_heartbeat_advances_are_starved_should_repair_in_normal_status()
+     {
+        // Statics, not captures: the link hooks are bare `fn` pointers. Declared
+        // inside the test because the sibling tests in this binary run in
+        // parallel and would otherwise share them.
+        static GAP_NS: AtomicU64 = AtomicU64::new(0);
+        static WITHHELD_OP: AtomicU64 = AtomicU64::new(0);
+
+        /// Chain link 1 -> 2: swallow the first partition prepare, once.
+        fn withhold_one_prepare(packet: &Packet) -> bool {
```
hubcio: simplification: `withhold_one_prepare` is byte-identical at four sites here. a `macro_rules!` emitting the per-test statics and hook saves ~35 lines, though the tests then name items the macro hides.

hubcio (changes requested):

spetz (dismissed):

core/shard/src/lib.rs
```diff
@@ -9286,6 +9539,303 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+/// rejoin puts every group's stream in flight within `groups / arms` ticks, and
+/// each one is a window the SERVING peer walks on its own pump, so the cost
+/// lands on a node that has nothing wrong with it. Sized at twice
+/// [`IggyShard::PARTITION_TRANSFERS_INFLIGHT_MAX`]: a repair streams journal
+/// entries the peer already holds resident, where a transfer reads and hashes
+/// whole segments, so more of them fit in the same serving budget.
+///
+/// Over-cap groups stay gap-stopped with their debounce satisfied, so they arm
+/// as sessions complete.
+const PARTITION_REPAIRS_INFLIGHT_MAX: usize = 8;
```
hubcio, line 9645: warning: this binds the sweep only - the four edge-triggered sites call `maybe_request_partition_repair` with no in-flight check, so a node-wide view change drives `on_start_view` past the cap for every group. move the refusal inside the arm fn, like `arm_partition_transfer` does.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7115,156 @@ where
...
+                        // view log, which a settled primary has none of), so a
+                        // peer holding those ops holds them identically.
+                        //
+                        // The pick is positional, not liveness-aware: a dead
+                        // choice leaves the session re-requesting on the stall
+                        // timer, which is where a repair abandon budget (what
+                        // `burn_transfer_attempt` gives transfers) would rotate
+                        // it. Still strictly better than the warn this replaced,
+                        // which recovered nothing at all.
+                        let peer = if primary == self_id {
```
hubcio, line 7230: warning: a partition repair session has no abandon budget, so if this successor is down the stall timer re-sends to it forever while `repair.is_some()` fences both the sweep and every edge site. that wedges harder than the warn it replaced - rotate on repeated stalls.

core/partitions/src/iggy_partition.rs
```diff
@@ -3488,6 +3552,24 @@ where
 
     #[allow(clippy::future_not_send)]
     pub async fn commit_journal(&mut self, config: &PartitionsConfig) {
+        self.commit_journal_bounded(config, usize::MAX).await;
```
hubcio: warning: `PARTITION_WALK_OPS_MAX` bounds the tick sweep only. `on_commit`, `on_start_view`, the post-transfer tail and the post-repair walk all come through here with `usize::MAX` and still walk a whole rejoin backlog on the pump.

core/partitions/src/journal.rs
```diff
@@ -842,27 +857,60 @@ where
...
+        // The contiguous run can never outlast the resident headers, and
+        // `commit_max` is a cluster frontier this replica may be far below, so
+        // the slot vec is sized off what is actually here.
+        // Both `min` operands are `usize`-derived, so the narrowing cannot
+        // truncate whatever `commit_max` is.
+        #[allow(clippy::cast_possible_truncation)]
+        let span = (commit_max - from_op + 1)
+            .min(headers.len() as u64)
+            .min(limit as u64) as usize;
+        let mut slots: Vec<Option<PrepareHeader>> = vec![None; span];
```
hubcio, line 913: warning: for the `usize::MAX` callers `span` collapses to `headers.len()`, so a gap-stopped replica whose run is one op still allocates and fills a slot per resident header at 256 bytes each. pass a real limit, or size the vec off the run.

core/shard/src/lib.rs
```diff
@@ -7988,16 +8244,7 @@ where
...
-            .unwrap_or(false);
-        let missing_suffix = canonical_suffix
-            && !partition
-                .log
-                .journal()
-                .inner
-                .repaired_window_shape(commit_to_op, head)
-                .complete;
+        let missing_suffix = partition_missing_suffix(partition);
         if !commit_lag && !missing_suffix {
```
hubcio, line 8314: warning: a group that is both walk-stalled and missing-suffix refetches its whole committed prefix - `from_op` stays `commit_min + 1` while `fetch_to_op` jumps to `head`. the doc at 9670 claims these windows are disjoint. gate the widened `fetch_to_op` on `!commit_lag`.

core/shard/src/lib.rs
```diff
@@ -10519,3 +11069,517 @@ mod superblock_fail_stop_tests {
...
+        let probe = gap_stopped();
+        let mut gap_ticks = 0;
+        for _ in 1..DEBOUNCE {
+            drive_partition_gap_debounce(&probe, &mut gap_ticks, DEBOUNCE, true);
+        }
+        assert_eq!(
+            drive_partition_gap_debounce(&probe, &mut gap_ticks, DEBOUNCE, true),
+            GapArm::Arm
+        );
+        gap_ticks = 0;
```
hubcio: warning: this replaced the `assert_eq!(gap_ticks, 0)` that pinned the reset, so the test now hand-feeds the state it used to check. nothing asserts the reset at 8261 any more - drive `maybe_request_partition_repair` and assert it there.

core/shard/src/lib.rs
```diff
@@ -10519,3 +11069,517 @@ mod superblock_fail_stop_tests {
...
+        let fewest = walks.iter().min().copied().unwrap_or_default();
+        assert!(
+            most - fewest <= 1,
+            "walks are not spread evenly: one group got {most}, another {fewest}"
+        );
+    }
+
+    #[test]
+    fn given_a_capped_arm_backlog_when_swept_should_arm_every_gap_stopped_group() {
+        // The arm cap carries no cursor because arming REMOVES a group from the
```
hubcio: warning: the model never applies `PARTITION_REPAIRS_INFLIGHT_MAX` and never completes a session, so it asserts all 14 groups arm in 5 passes when production allows 8 live at once. model the inflight cap and session completion.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9539,303 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
 
+/// Repair sessions the partition tick sweep will OPEN per pass.
+///
+/// A RATE cap, not a concurrency one: unlike
+/// `IggyShard::PARTITION_TRANSFERS_INFLIGHT_MAX`, which counts live sessions
+/// and refuses over the count, nothing bounds how many partition repair
+/// sessions are live at once. This only spreads the opening cost, because one
+/// arm is a `RequestPrepares` plus a repair stream the serving peer walks
+/// synchronously and a node-wide gap (a rejoin, a lossy link) makes every group
+/// on this shard due in the same tick. Over-cap groups stay due, and the next
```
hubcio: nit: both halves are now false - `PARTITION_REPAIRS_INFLIGHT_MAX` does bound live sessions, and the cursor is walk-only after this commit, so over-cap arms queue in namespace order. the sweep comment at 6952 says the opposite.

core/partitions/src/iggy_partition.rs
```diff
@@ -154,6 +154,33 @@ where
...
+    ///
+    /// `Cell` for the same reason as [`Self::prepare_gap_drops`]: the sweep
+    /// drives it from the shared borrow it probes the partition through, so the
+    /// in-flight scan the arm budget needs can run without a `&mut` outstanding.
+    pub gap_ticks: Cell<u32>,
+    /// Prepares the backup gap check destroyed since the shard last drained
+    /// the count, folded into `partition_prepare_gap_drops_total` there.
+    /// Replicated traffic has no client to answer and retransmit skips ops that
+    /// already reached quorum, so nothing else records that the frame existed.
+    /// It counts the prepares that ARRIVED after a hole, not the holes: a gap
```
hubcio: nit: a nonzero value does not prove the driver has work - the branch fires above and below the sequencer, as your own comment at 3310 says, so a duplicate delivery of an already-journaled op bumps it too.

core/shard/src/lib.rs
```diff
@@ -7739,6 +7969,23 @@ where
     /// node-wide view change or rejoin, and capped arms reschedule on the flat
     /// retry interval, so the losers stay phase-locked and the sweep repeats
     /// every interval for the whole rejoin.
+    /// Live journal-repair sessions on this shard, for
+    /// [`PARTITION_REPAIRS_INFLIGHT_MAX`]. Same shape and same caution as
+    /// [`Self::partition_transfers_inflight`]: it borrows every partition, so
+    /// callers count BEFORE forming a `&mut`.
+    fn partition_repairs_inflight(&self) -> usize {
```
hubcio: nit: this fn got spliced under the transfers doc block, so it now reads 'receiving-side transfers currently in flight'. `partition_transfers_inflight` at 7989 has no doc left.

core/partitions/src/journal.rs
```diff
@@ -842,27 +857,60 @@ where
...
         }
-        result
+        let headers = unsafe { &*self.headers.get() };
+        // The contiguous run can never outlast the resident headers, and
+        // `commit_max` is a cluster frontier this replica may be far below, so
+        // the slot vec is sized off what is actually here.
+        // Both `min` operands are `usize`-derived, so the narrowing cannot
+        // truncate whatever `commit_max` is.
+        #[allow(clippy::cast_possible_truncation)]
+        let span = (commit_max - from_op + 1)
```
hubcio: nit: at `commit_max == u64::MAX` the `+ 1` overflows - debug panics, release wraps to a `span` of 0 and the walk stops forever. use `(commit_max - from_op).saturating_add(1)`.

core/partitions/src/journal.rs
```diff
@@ -842,27 +857,60 @@ where
...
+            .min(headers.len() as u64)
+            .min(limit as u64) as usize;
+        let mut slots: Vec<Option<PrepareHeader>> = vec![None; span];
+        for header in headers {
+            if header.op < from_op {
+                continue;
+            }
+            // Bounded by the `span` test right below, which is itself bounded
```
hubcio: nit: the `slot < span` test runs after the cast, so it does not stop truncation. what actually holds this is `usize` being 64-bit - say that, or test `header.op - from_op < span as u64` first.

core/partitions/src/journal.rs
```diff
@@ -762,6 +762,21 @@ where
         headers.iter().find(|header| header.op == op).copied()
     }
 
+    /// Whether `op` is resident, without reading its header.
+    ///
+    /// Equivalent to `header_by_op(op).is_some()` and answers in O(log n)
```
hubcio: nit: 'equivalent to `header_by_op(op).is_some()`' contradicts the note you just added below - across the await in `append_with_meta` the header is present and this answers false. `apply_repaired_prepare` now leans on it for idempotence, which is the direction a false negative hurts.

core/shard/src/lib.rs
```diff
@@ -9286,6 +9539,303 @@ fn repair_serve_ceiling(requested_to_op: u64, commit_max: u64, head: u64) -> u64
...
+        && probe.next_op_resident
+}
+
+/// What the debounce says about one partition on one sweep.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum GapArm {
+    /// Not gap-stopped, or gap-stopped for less than the debounce.
+    NotDue,
+    /// Due, but this sweep's arm budget is spent. The debounce stays satisfied
+    /// and the sweep resumes at this group, so the next pass arms it.
```
hubcio: nit: stale after the cursor change - an arm deferral records no cursor now, so the sweep does not resume at this group. drop the resume clause and keep the debounce half.

core/configs/src/server_config/cluster.rs
```diff
@@ -1002,6 +1028,13 @@ impl Validatable<ConfigurationError> for ClusterConfig {
         // The repair retry interval sizes a tick threshold that has to advance;
         // `0` / `disabled` / `unlimited` all collapse to zero and would wedge
         // every stalled repair stream - reject them.
+        if self.repair_gap_debounce_interval.get_duration().is_zero() {
```
hubcio, line 1031: nit: no test for this branch, unlike `validate_rejects_zero_repair_retry_interval`. worth pinning since `0`, `disabled` and `unlimited` all parse to `IggyDuration::ZERO`.

core/shard/src/lib.rs
```diff
@@ -10519,3 +11069,517 @@ mod superblock_fail_stop_tests {
...
+        partition.is_multiple_of(3)
+    }
+
+    fn namespaces() -> Vec<IggyNamespace> {
+        (0..GROUPS)
+            .map(|partition| IggyNamespace::new(1, 1, partition))
+            .collect()
+    }
+
+    #[derive(Default)]
```
hubcio: nit: `Served::default()` gives empty vecs next to a `new()` that sizes them by `GROUPS`, so calling it panics on the first index. drop the derive.

core/shard/src/lib.rs
```diff
@@ -7041,6 +7115,156 @@ where
...
+                        // choice leaves the session re-requesting on the stall
+                        // timer, which is where a repair abandon budget (what
+                        // `burn_transfer_attempt` gives transfers) would rotate
+                        // it. Still strictly better than the warn this replaced,
+                        // which recovered nothing at all.
+                        let peer = if primary == self_id {
+                            next_transfer_peer(
+                                self_id,
+                                self_id,
+                                self.partition_consensus.replica_count,
```
hubcio: nit: this reaches for the shard-wide count while `consensus` is in scope and 8967 uses `consensus.replica_count()`. equal today, but this is the one site that would pick a peer outside the group if per-group replication ever lands.

hubcio (changes requested):

hubcio (dismissed):

mmodzelewski (dismissed):

spetz (approved):

hubcio (approved):

PR #4008, opened 2026-08-31
fix(shard): drive metadata repair and the commit walk from the tick

numinnex:
A metadata backup that fell behind had no way back on its own, the same starvation the partition plane just had. A follower advances commit_max from every prepare header in replicate_preflight, before the gap check drops the prepare, so the commit heartbeat lands as Accepted rather than Advanced and the arming site inside that branch never fires. Under sustained metadata traffic the gap wedged until an unrelated view change. The same starvation left a backup holding resident committed ops that nothing re-drove, since the follower walk runs at the tail of an accepted prepare and the gap check returns before it.

The metadata tick now evaluates the same two level-triggered checks the partition driver uses, off one probe. A hole below commit_max arms the existing repair against the primary, debounced on the repair retry interval. Resident committed ops re-drive the walk, follower only, because a backup's walk ships no wire replies while a stranded primary is resume_stranded_commits' job. A gap below the serving peer's retention floor needs no special handling: the serve path answers RangeEvicted and the existing conversion arms a state transfer.


numinnex:
Merge after #4006

core/metadata/src/impls/metadata.rs
```diff
@@ -838,12 +853,26 @@ where
...
             transfer_offer_cache: RefCell::new(None),
+            prepare_gap_drops: Cell::new(0),
             applied_frontier: Arc::default(),
         }
     }
 }
 
 impl<C, J, S, M, SB> IggyMetadata<C, J, S, M, SB> {
+    /// Record one prepare destroyed by the backup gap check
+    /// ([`Self::prepare_gap_drops`]).
```
hubcio: critical: this doc link points at the private field `Self::prepare_gap_drops`, so `cargo doc` with `-D warnings` fails and the doctest CI job goes red, which then cancels the test jobs. use plain backticks or inline the method.

also at line 870.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9351,95 @@ where
...
+            //
+            // Follower only. A backup's `commit_journal` ships no wire replies,
+            // while a stranded primary owes its clients answers, which is
+            // `resume_stranded_commits` above. Not gated on `recovery_owned`
+            // (repaired prepares are journaled without being walked, so gating
+            // parks the walk for the whole session); `group_is_walk_stalled`
+            // itself refuses mid-transfer, where a walk past the incoming
+            // `snapshot_seq` would break the install.
+            if group_is_walk_stalled(&probe) && consensus.is_follower() {
+                tracing::info!(
```
hubcio: warning: this fires every tick for the whole repair stream, since repaired prepares are journaled without walking, so it is one info line per tick until the rejoin ends. use `debug!` like the partition arm at line 7325.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9351,95 @@ where
...
+                    // Always repair, never classify the gap up front: a window
+                    // below the peer's retention floor is answered
+                    // `RangeEvicted`, and `on_repair_range_reply` converts that
+                    // to a state transfer. The floor is only ever learned
+                    // through that refusal.
+                    self.maybe_request_metadata_repair(consensus, peer).await;
+                }
+            }
+            // Undebounced and uncapped, like the partition walk arm: the
+            // predicate guarantees the walk finds at least the next op, so it
```
hubcio: warning: this can spin - the probe checks only the header ring, but the walk needs the body from `entry()`, which stays `None` once the journal is poisoned, so nothing moves and repair never arms. if the walk did not move `commit_min`, count the tick toward the repair debounce instead.

core/shard/src/lib.rs
```diff
@@ -5561,6 +5572,18 @@ where
                     consensus.group(),
                 )
                 .await;
```
hubcio, line 5652: warning: this re-sends to `session.peer` forever, with no attempt budget and no rotation, so a dead pick (the gap-stopped-primary arm can choose a down backup) wedges the plane for good. port `burn_repair_attempt` plus `next_transfer_peer` from the partition retry at lines 7141-7164.

core/shard/src/lib.rs
```diff
@@ -5850,6 +5877,11 @@ where
             let nonce = iggy_common::random_id::get_uuid();
```
hubcio, line 5970: warning: the guards above have no `peer == consensus.replica()` check, and the fallback at line 9478 passes `primary_index(view)` unrotated, which is self on a leading replica. the self-send fails but the session stays and fences every arming site, so refuse self here and rotate at 9478.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9351,95 @@ where
...
+            // itself refuses mid-transfer, where a walk past the incoming
+            // `snapshot_seq` would break the install.
+            if group_is_walk_stalled(&probe) && consensus.is_follower() {
+                tracing::info!(
+                    shard = self.id,
+                    commit_min = probe.commit_min,
+                    commit_max = probe.commit_max,
+                    "metadata commit walk parked over resident committed ops; resuming"
+                );
+                metadata.commit_journal().await;
```
hubcio, line 9537: warning: this walk can land between a chunk's last prepare and its `RepairDone`, and `on_repair_range_reply` only pulls the next chunk when its own walk moved `commit_min`, so the session idles a full retry interval per hit. reset `session.idle_ticks` on each accepted repair prepare, like the partition ingest.

core/shard/src/lib.rs
```diff
@@ -5561,6 +5572,18 @@ where
                     consensus.group(),
                 )
                 .await;
+            } else {
```
hubcio, line 5653: nit: this close only runs after the session sat idle for the full retry window, and a standing session zeroes the gap debounce every tick meanwhile. check `commit_min >= to_op` at the top of each tick, like `repair_finished` in the partition sweep.

core/shard/src/lib.rs
```diff
@@ -5834,8 +5857,12 @@ where
     }
 
     /// Start metadata tail journal-repair from `peer` when the commit walk
-    /// gap-stopped below the known frontier. Shared by `StartView` adoption
-    /// and the post-install step of a state transfer.
+    /// gap-stopped below the known frontier.
+    ///
+    /// Every arming site funnels through here -- `StartView` adoption, the
```
hubcio: nit: not every arming site goes through here - `advance_pending_metadata_view` builds its session inline at line 5834 and skips these guards. name the exception.

core/simulator/src/lib.rs
```diff
@@ -5964,6 +5965,509 @@ mod partition_repair_driver_tests {
...
+        let sends = LOAD_TICKS / TICKS_PER_SEND;
+        assert!(
+            committed >= COMMITTED_MIN,
+            "the backup committed only {committed} ops across {sends} sends, so the \
+             driver was never ticked over a loaded group"
+        );
+        // Recorded, not load-bearing: the tail walk in `on_replicate` keeps a
+        // healthy backup caught up at quiescence, so a naive lag detector has
+        // nothing to misread here; `gap_detector_tests` pins the predicate.
+        assert_eq!(
```
hubcio, line 6445: nit: the comment says `longest_lag_run` is recorded and not load-bearing, but this `assert_eq!` fails the test on the first non-zero sample. either own the assert and drop the comment, or log the value instead.

core/shard/src/lib.rs
```diff
@@ -1319,6 +1319,15 @@ where
     /// repair takes over at install. See [`MetadataTransferSession`].
     metadata_transfer: RefCell<Option<MetadataTransferSession>>,
 
+    /// Consecutive ticks the metadata group has been seen gap-stopped
+    /// (committed ops it cannot walk to, because the op one past its commit
+    /// frontier is missing from the WAL). Debounces `tick_metadata`'s
+    /// level-triggered repair arm; the partition twin is
+    /// `IggyPartition::gap_ticks`, a `Cell` there for a borrow reason this one
```
hubcio: nit: this says the partition twin is a `Cell` for a borrow reason this field does not have, but this field is a `Cell` too. say why (`&self` methods) or drop the comparison.

core/metadata/src/impls/metadata.rs
```diff
@@ -776,6 +776,21 @@ pub struct IggyMetadata<C, J, S, M, SB = PingPongSuperblock> {
...
+    /// already reached quorum, so nothing else records that the frame existed.
+    ///
+    /// It counts what the ordering check destroyed, which is neither the holes
+    /// nor only them: a gap opened by the last prepare of a burst leaves it at
+    /// zero, and a duplicate delivery of an op this replica already sequenced
+    /// bumps it without any hole existing. Zero proves nothing, and nonzero is
+    /// a reason to look at the drop log. The partition twin is
+    /// `IggyPartition::prepare_gap_drops`.
+    ///
+    /// `Cell`: `tick_metadata` drains it once per tick off the same shared
```
hubcio: nit: there is no shared borrow here - `MetadataHandle::metadata` hands out a plain reference. it is a `Cell` because every method on this type takes `&self`, so say that or drop the sentence.

core/simulator/src/lib.rs
```diff
@@ -5209,9 +5209,10 @@ mod partition_repair_driver_tests {
         };
     }
 
-    /// The partition-plane prepare a packet carries, if it carries one for
-    /// `group`.
-    fn prepare_for(packet: &Packet, group: u64) -> Option<PrepareHeader> {
+    /// The prepare a packet carries, if it carries one for `group`. Keyed on
+    /// the group, so `metadata_repair_driver_tests` reads the metadata plane's
+    /// prepares with the same helper.
+    pub fn prepare_for(packet: &Packet, group: u64) -> Option<PrepareHeader> {
```
hubcio, line 5219: nit: `pub` inside a private `#[cfg(test)]` module for one sibling - `pub(super)` says the actual reach.

also at lines 5225, 5235, 5246, 5256.

core/shard/src/lib.rs
```diff
@@ -1705,12 +1715,13 @@ where
...
-    /// tests keep the compile-time [`partitions::REPAIR_RETRY_TICKS`] default.
-    /// [`PARTITION_GAP_DEBOUNCE_TICKS_MIN`] still floors whatever is set.
-    pub fn set_partition_gap_debounce_ticks(&self, ticks: u32) {
-        self.partition_gap_debounce_ticks.set(ticks);
+    /// Override the tick gap debounce (consensus ticks) from configuration,
+    /// for both planes' detectors. Called once per shard at bootstrap; the
+    /// simulator and tests keep the compile-time
+    /// [`partitions::REPAIR_RETRY_TICKS`] default.
+    /// [`GAP_DEBOUNCE_TICKS_MIN`] still floors whatever is set.
+    pub fn set_gap_debounce_ticks(&self, ticks: u32) {
```
hubcio: nit: the config key is `repair_gap_debounce_interval` and the sibling is `set_repair_retry_ticks`, so this setter reads odd without the `repair_` prefix. if renaming, move the field and `GAP_DEBOUNCE_TICKS_MIN` with it.

core/shard/src/lib.rs
hubcio, line 9537: nit: `commit_journal` has no per-call cap, unlike `COMMIT_WALK_OPS_MAX` on the partition walk, so one tick arm can read through the whole journal ring. fine while the resident backlog stays small, worth a cap if it can grow.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9351,95 @@ where
...
+            if verdict == GapArm::Arm {
+                let self_id = consensus.replica();
+                let primary = consensus.primary_index(consensus.view());
+                // A gap-stopped primary cannot ask itself, and rotating is the
+                // same call the partition sweep makes for the same reason: no
+                // other site re-drives a settled primary's own hole, and its
+                // window here is its COMMITTED prefix, which every peer holds
+                // identically. Positional, not liveness-aware -- a dead pick
+                // leaves the session re-requesting on the stall timer.
+                let peer = if primary == self_id {
```
hubcio: simplification: this peer pick is a line-for-line copy of the partition one at line 7258. one shared `gap_repair_peer(self_id, replica_count, primary) -> Option<u8>` keeps the two planes from drifting - it is a wash on line count, the value is the single rule.

core/shard/src/lib.rs
```diff
@@ -9212,7 +9240,53 @@ where
...
+        <MJ as JournalHandle>::Target:
+            Journal<Entry = Message<PrepareHeader>, Header = PrepareHeader>,
+    {
+        let commit_min = consensus.commit_min();
+        let commit_max = consensus.commit_max();
+        let normal = consensus.is_normal();
+        let transferring = consensus.is_transferring();
+        // The stage counts as a transfer too: `begin_state_transfer_await` owns
+        // the recovery before a session exists.
+        let recovery_owned = self.metadata_transfer.borrow().is_some()
```
hubcio: simplification: `|| transferring` is redundant - the only reader, `group_is_gap_stopped`, already refuses on `probe.transferring`, and `is_transferring()` covers `AwaitingTarget`. drop the clause and the comment.

core/metadata/src/impls/metadata.rs
```diff
@@ -776,6 +776,21 @@ pub struct IggyMetadata<C, J, S, M, SB = PingPongSuperblock> {
     /// whole snapshot on shard 0's pump, and hands each requester its own
     /// multi-MB copy.
     transfer_offer_cache: RefCell<Option<Rc<StateTransferOffer>>>,
+    /// Prepares the backup gap check destroyed since the shard last drained
```
hubcio: simplification: most of this doc is the `IggyPartition::prepare_gap_drops` text again, and the metrics recorder explains it a third time. one line pointing at the partition field plus what differs here is enough.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9351,95 @@ where
...
+                    self.metadata_gap_ticks.set(0);
+                    tracing::warn!(
+                        shard = self.id,
+                        commit_min = probe.commit_min,
+                        commit_max = probe.commit_max,
+                        "metadata is gap-stopped below its own commit frontier with no peer to \
+                         repair from"
+                    );
+                } else {
+                    tracing::info!(
```
hubcio: simplification: `maybe_request_metadata_repair` logs the same arm right after this, and from here it cannot return without logging. drop this `info!` and put `peer` on the arm fn's line, like the partition tick does.

core/metadata/src/impls/metadata.rs
```diff
@@ -838,12 +853,26 @@ where
...
+            prepare_gap_drops: Cell::new(0),
             applied_frontier: Arc::default(),
         }
     }
 }
 
 impl<C, J, S, M, SB> IggyMetadata<C, J, S, M, SB> {
+    /// Record one prepare destroyed by the backup gap check
+    /// ([`Self::prepare_gap_drops`]).
+    pub fn note_prepare_gap_drop(&self) {
```
hubcio: simplification: one caller, and the partition twin just inlines the `saturating_add`. inline it and drop the method - that also removes one of the two private doc links.

hubcio (changes requested):
three findings sit on lines outside the diff:

- critical: `core/shard/src/lib.rs:4316-4322` - nothing clears `metadata_repair` on StartView adoption, so a session armed against the old primary keeps asking a dead peer forever and the backup never acks live prepares - with 3 replicas that is the commit quorum. clear it in the adopted block, like `partition.repair = None` at line 4432.
- warning: `core/shard/src/lib.rs:5816` - this defers to any live session, so a primary-elect keeps re-requesting from whatever peer a Normal-era arm picked, usually the primary that just died, and the view change parks until the escalation timer. drop a session whose `to_op != pending.op_head` here instead. a plain `!is_normal()` clear would kill the primary-elect session, which lives outside Normal on purpose.
- warning: `core/shard/src/lib.rs:6210-6227` - a refused transfer re-arms repair on the same peer at once, and the peer refuses again while not caught up, so a backup with an evicted gap ping-pongs at wire speed until the primary idles. the sender hardcodes `unavailable(false, ...)` at line 6154. pass the real transient flag and back off on it like the partition twin.

core/simulator/src/lib.rs
numinnex: Tried this and backed it out: `pub(super)` inside a private module resolves to the same reach as `pub(crate)`, so `clippy::redundant_pub_crate` rejects it on all five helpers:

```
error: pub(crate) function inside private module
    --> core/simulator/src/lib.rs:5215:5
     |
5215 |     pub(super) fn prepare_for(packet: &Packet, group: u64) -> Option<PrepareHeader> {
     |     ----------^^^^^^^^^^^^^^^
     |     help: consider using: `pub`
     = note: `-D clippy::redundant-pub-crate` implied by `-D warnings`
```

The module is private and `#[cfg(test)]`, so `pub` and `pub(super)` are the same reach here and the lint wins. Kept `pub` and noted why above the first helper so nobody re-tightens it. Happy to add `#[allow(clippy::redundant_pub_crate)]` instead if you would rather have the narrower spelling.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9455,100 @@ where
...
+            // hands the op to repair.
+            //
+            // Follower only. A backup's `commit_journal` ships no wire replies,
+            // while a stranded primary owes its clients answers, which is
+            // `resume_stranded_commits` above. Not gated on `recovery_owned`
+            // (repaired prepares are journaled without being walked, so gating
+            // parks the walk for the whole session); `group_is_walk_stalled`
+            // itself refuses mid-transfer, where a walk past the incoming
+            // `snapshot_seq` would break the install.
+            if group_is_walk_stalled(&probe) && consensus.is_follower() {
```
hubcio, line 9527: critical: `is_follower()` leaves a primary with no re-driver for a walk that `COMMIT_WALK_OPS_MAX` cut short, and once the inherited suffix reaches quorum `advance_commit_min` trips its sequential assert and the shard panics. drop the gate, like the partition arm at line 7400.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9455,100 @@ where
...
+                    commit_max = probe.commit_max,
+                    "metadata commit walk parked over resident committed ops; resuming"
+                );
+                metadata.commit_journal().await;
+                // A walk that moved nothing found the header and not the body;
+                // see `metadata_gap_probe`. Recorded rather than retried, so the
+                // detector stops calling this a parked walk and arms repair for
+                // the body instead.
+                let walked = consensus.commit_min();
+                self.metadata_walk_stuck_op
```
hubcio, line 9543: warning: repair cannot refill this op: the ingest at line 5008 skips a resident header and `append` refuses the same slot, so the session only stalls and rotates peers. drop the stuck header from the WAL before arming.

core/metadata/src/impls/metadata.rs
```diff
@@ -218,6 +218,20 @@ impl IggySnapshot {
...
 /// unacceptable: silently accepting corruption, or bricking a healthy node.
+/// Committed ops one [`IggyMetadata::commit_journal`] call applies before
+/// returning to the pump.
+///
+/// The twin of `partitions::COMMIT_WALK_OPS_MAX`, and needed for the same
+/// reason: the walk reads a WAL body and applies it per op with no await the
+/// pump can interleave, and the resident `(commit_min, commit_max]` run is the
+/// whole backlog after a repair or a rejoin, not the pipeline depth.
+///
+/// Every caller is re-driven, so a truncated walk resumes rather than losing
```
hubcio, line 229: warning: `resume_stranded_commits` re-drives the pipeline, not this walk, so nothing covers a primary that stopped at the cap. say follower-only here and drop the `is_follower()` gate on the tick arm.

core/shard/src/lib.rs
```diff
@@ -4990,6 +5017,15 @@ where
                 );
                 return;
             }
+            // The stream is alive, so only SILENCE should age it. The tick's own
+            // walk can land between a chunk's last prepare and its `RepairDone`,
+            // and `on_repair_range_reply` pulls the next chunk only when its own
+            // walk moved `commit_min`, so without this the session idles a full
+            // retry interval on every such landing.
+            if let Some(session) = self.metadata_repair.borrow_mut().as_mut() {
```
hubcio, line 5025: warning: a peer serving ops this replica already holds ages the session as silent and burns the budget, since both resets sit below the returns at 5008 and 5011. move them above 5008 like the partition ingest, together with the stuck-op fix or that session never rotates.

core/shard/src/lib.rs
```diff
@@ -5510,6 +5546,26 @@ where
...
     {
+        // Closed at the TOP of the tick, not after an idle window: a session
+        // that outlives its window fences every arming site and holds the gap
+        // debounce at zero (`recovery_owned`), so waiting a full retry interval
+        // to notice costs that interval on every arm behind it. The partition
+        // sweep's `repair_finished` check is the same idea.
+        let finished = self
+            .metadata_repair
+            .borrow()
+            .is_some_and(|session| consensus.commit_min() >= session.to_op);
```
hubcio, line 5557: warning: this closes on commit progress only. the partition `repair_finished` at line 7199 also closes on a view change or on leaving normal, so a session here outlives both - stamp the session with its view and test it here.

core/simulator/src/lib.rs
```diff
@@ -5964,6 +5969,512 @@ mod partition_repair_driver_tests {
     }
 }
 
+#[cfg(test)]
+mod metadata_repair_driver_tests {
```
hubcio, line 5973: warning: nothing tests the new machinery: stall budget and rotation, self-peer guard, top-of-tick close, idle reset on ingest, stuck-op record, walk cap. the four tests here are backup-side and never reach the cap.

core/shard/src/lib.rs
```diff
@@ -5532,6 +5588,39 @@ where
...
+                tracing::warn!(
+                    shard = self.id,
+                    peer,
+                    next_peer,
+                    to_op,
+                    "metadata repair stalled past its retry budget; re-arming from another \
+                     replica"
+                );
+                *self.metadata_repair.borrow_mut() = None;
+                self.note_metadata_repair_progress();
```
hubcio, line 5613: nit: this zeroes the counter the doc at 6529 says must outlive the session, so each new peer gets a fresh budget. either drop the reset or fix the doc - the partition twin at 7280 has the same mismatch.

core/shard/src/lib.rs
```diff
@@ -9806,6 +10104,33 @@ const fn drive_partition_gap_debounce(
...
+///
+/// The primary, except when this replica IS the primary: no site re-drives a
+/// settled primary's own hole, so leaving it to warn wedges the group, and the
+/// next op to commit walks `advance_commit_min` into its sequential assert. Any
+/// replica in `Normal` or `ViewChange` serves `RequestPrepares`, and a
+/// gap-stopped replica's window is its COMMITTED prefix, which every peer that
+/// holds those ops holds identically.
+///
+/// Positional, not liveness-aware. A dead pick is corrected by the stall
+/// budget on either plane, which drops the session and rotates one step further
```
hubcio, line 10118: nit: for more than 3 replicas this does not walk the ring - `next_transfer_peer` prefers the primary, so a backup only alternates primary and primary+1 and never asks the rest. any replica serves `RequestPrepares`, so is the shortcut intended here?

core/shard/src/lib.rs
```diff
@@ -6395,6 +6521,27 @@ where
...
 
+    /// Burn one stalled repair round; `true` once the budget is exhausted and
+    /// the session should be re-armed against a different peer.
+    ///
+    /// The partition twin is `IggyPartition::burn_repair_attempt`, and it lives
+    /// on the shard here for the same reason `metadata_transfer_attempts` does:
+    /// one metadata group per node. It has to outlive the SESSION either way,
+    /// or the rotation that mints a new one would reset the count and re-target
+    /// forever without ever giving up on a peer.
+    fn burn_metadata_repair_attempt(&self) -> bool {
```
hubcio, line 6532: nit: missing `#[must_use]` - the partition twin has it because dropping the bool disables the stall budget.

also at line 6547.

core/metadata/src/impls/metadata.rs
```diff
@@ -218,6 +218,20 @@ impl IggySnapshot {
...
+/// anything: `tick_metadata`'s walk backstop covers a follower and
+/// `resume_stranded_commits` covers the primary, both level-triggered on
+/// `commit_min < commit_max` every tick.
+const COMMIT_WALK_OPS_MAX: usize = 64;
```
hubcio, line 233: simplification: same value as `partitions::COMMIT_WALK_OPS_MAX` with nothing tying them together. move it to `consensus` or `server_common`, which both crates already depend on, or say here that the pairing is convention only.

core/metadata/src/impls/metadata.rs
```diff
@@ -218,6 +218,20 @@ impl IggySnapshot {
 /// accepted (unverified, loudly) while a PRESENT but mismatching one refuses boot. A
 /// bare checksum could not tell those apart, and guessing wrong in either direction is
 /// unacceptable: silently accepting corruption, or bricking a healthy node.
+/// Committed ops one [`IggyMetadata::commit_journal`] call applies before
```
hubcio, line 221: nit: the doc block above belongs to `SNAPSHOT_TRAILER_MAGIC` and now attaches here, leaving the magic undocumented. move this const and its doc above that block or below the magic.

hubcio (changes requested):
three findings sit on lines outside the diff:

- warning: `core/shard/src/lib.rs:4332` - still open from the last round: the adopted block never clears `metadata_repair`, while the partition twin drops `partition.repair` at line 4448. the stall budget now rotates off the dead peer after `REPAIR_MAX_STALL_RETRIES` retries, but the one-line clear is still the fix.
- warning: `core/shard/src/lib.rs:5896` - still open: this defers to any live session, so a primary-elect inherits a session armed against the primary that just died. drop a session whose `to_op != pending.op_head` here.
- warning: `core/shard/src/lib.rs:6247` and `6319` - still open: the serving side hardcodes `unavailable(false, ..)` and the receiver re-arms repair on the same peer at once, so a gap-stopped primary repairing from a backup loops repair, `RangeEvicted`, transfer, `NotCaughtUpPrimary` at network rate. port `transfer_failures` and `transfer_rearm_backoff` from the partition plane.

core/shard/src/lib.rs
spetz: The 64-op cap can leave an inherited committed prefix unapplied on a primary. This follower-only gate prevents resuming it, and `resume_stranded_commits` drains the pipeline rather than that journal prefix. Ensure primary catch-up resumes before the inherited suffix commits

core/shard/src/lib.rs
spetz: Completion only checks commit progress, so a repair session survives StartView adoption and keeps its old peer. Invalidate stale sessions when the view changes while preserving valid primary-elect repair sessions

core/shard/src/lib.rs
spetz: Marking the op stuck triggers repair, but ingestion rejects the replacement because the old header remains resident. Make the unreadable entry replaceable, or explicitly handle an unusable journal. Peer rotation alone cannot recover it.

core/shard/src/lib.rs
```diff
@@ -9277,6 +9455,100 @@ where
...
+                // Debug, not info: a repair stream journals its prepares without
+                // walking them, so this is the steady state for the whole
+                // duration of a rejoin and would be one line per tick.
+                tracing::debug!(
+                    shard = self.id,
+                    commit_min = probe.commit_min,
+                    commit_max = probe.commit_max,
+                    "metadata commit walk parked over resident committed ops; resuming"
+                );
+                metadata.commit_journal().await;
```
spetz, line 9607: Resetting the idle timer does not request the next chunk when this tick has already walked the current chunk before RepairDone arrives. Continue based on completed chunk progress, rather than requiring the RepairDone handler itself to advance `commit_min`

spetz (changes requested):

core/shard/src/lib.rs
spetz: Both progress resets remain below the resident-header return, so valid duplicate repair traffic is counted as silence and burns the retry budget. Account for valid received traffic before that return, together with the unreadable-entry fix

spetz (changes requested):
Also, some related issues:

**`core/shard/src/lib.rs:5896`**
Primary-elect recovery defers to any existing repair session, including one targeting the previous primary and an incompatible window. Replace that stale session before deferring to an active stream.

**`core/shard/src/lib.rs:6319`**
An unavailable transfer immediately restarts repair against the same peer. An evicted gap therefore loops through repair, RangeEvicted, transfer, and refusal at network speed. Preserve refusal state and apply backoff across these transitions.

hubcio (approved):

spetz (approved):

PR #4009, opened 2026-08-31
fix(journal): reject recovery slot collisions

Aias00:
## Why

`metadata.journal_slots` is configurable, but the WAL opening scan rebuilds its
in-memory index using the current value. Reducing the value can map multiple
live, unsnapshotted operations onto one slot. The scan previously kept the
later operation and silently made the earlier operation unreachable for
recovery, even though both entries remained durable in the WAL.

## What changed

- Reject a scan when different operations collide and the existing operation
  is above `snapshot_op`.
- Preserve the existing semantics for duplicate operations and entries already
  covered by the snapshot.
- Leave the WAL unchanged when recovery is refused and include an actionable
  error that tells operators to restore the previous `journal_slots` value.
- Document the fail-fast behavior on the metadata configuration field.

This does not change the wire protocol or WAL format. It prevents unsafe
shrinking from silently losing replay input; it does not add a WAL migration
mechanism for arbitrary slot-count reductions.

## Verification

- `cargo +1.98 fmt --all -- --check`
- `cargo +1.98 clippy -p journal -p configs --all-features --all-targets -- -D warnings`
- `cargo +1.98 test -p journal` (50 passed)
- `cargo +1.98 test -p metadata` (144 passed)
- `cargo +1.98 test -p configs` (242 passed)

`cargo sort` and `taplo` were not available locally. No dependency files were
changed, and the embedded server configuration was covered by the configs test
suite.

### DIFF
--- a/core/configs/src/server_config/metadata.rs
+++ b/core/configs/src/server_config/metadata.rs
@@ -100,7 +100,9 @@ pub struct MetadataConfig {
     /// Size of the metadata WAL's in-memory index, in slots (one
     /// committed-but-unsnapshotted op per slot). Headroom between forced
     /// checkpoints; more slots = rarer checkpoints, more memory, larger
-    /// per-checkpoint WAL rewrites.
+    /// per-checkpoint WAL rewrites. Reducing this for an existing data directory
+    /// is refused at boot when the live WAL suffix cannot fit without collisions;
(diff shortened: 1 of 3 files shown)


spetz (approved):
LGTM

hubcio (approved):

PR #4013, opened 2026-08-31
fix(csharp): set default socket buffer size to OS default

### LINKED ISSUE #4010: [dotnet] Default Send / Recieve TCP buffer is very small, impacting throughput
I tried to set up a C#/.NET equivalent of `iggy-bench`, and found the performance was 100x worse, then found that the default buffer size is fixed to just 4KB:

In `TcpMessageStream.cs` line 1067, in `TryEstablishConnectionAsync`:

```
                socket = new Socket(ServerAddress.AddressFamilyOf(host), SocketType.Stream, ProtocolType.Tcp);
                socket.SendBufferSize = _configuration.SendBufferSize;
                socket.ReceiveBufferSize = _configuration.ReceiveBufferSize;
```
Which pick up from `IggyClientConfigurator.cs` line 49:

```
    /// <summary>
    ///     The size of the receive buffer in bytes. Default is 4096.
    /// </summary>
    public int ReceiveBufferSize { get; set; } = 4096;

    /// <summary>
    ///     The size of the send buffer in bytes. Default is 4096.
    /// </summary>
    public int SendBufferSize { get; set; } = 4096;
```

A 4KB buffer is too small for the kind of throughput needed:

| C# 1-producer, 1000×1000B batches	| Throughput	| p50 latency |
| :---:   | :---: | :---: |
| Default config (4 KB buffers)	| 1.7 MB/s	| 670 ms |
| SendBufferSize/ReceiveBufferSize = int.MaxValue	| 102 MB/s	| 9.8 ms |

The C#/.NET is the only one that sets a fixed socket size, the rest leave it to the OS to auto-size.

richardcocks:
Closes #4010 

## Rationale

The previous default buffer size was too small, affecting throughput. 

## What changed?

A null value for the socket buffer will cause the OS to auto-size the buffer, rather than being a fixed small buffer. Configuration and Builder validation now accepts null, but still validates for positive integers when set. 

## AI Usage

AI was used to quickly scaffold benchmarks, discover and diagnose the issue, as well as quickly validate the fix-shape, but all code in this PR was hand-edited.


foreign/csharp/Iggy_SDK/Configuration/IggyClientConfigurator.cs
```diff
@@ -44,14 +44,14 @@ public sealed class IggyClientConfigurator
     public int MaxResponseFrameSize { get; set; } = 64 * 1024 * 1024;
 
     /// <summary>
-    ///     The size of the receive buffer in bytes. Default is 4096.
+    ///     The size of the receive buffer in bytes. Default is null.
```
hubcio: also at line 52.

`Default is null.` says nothing - say that nothing is set on the socket then, so the system default receive buffer is used and on linux the kernel keeps auto-tuning it (an explicit size turns that off). the linux start value is the middle number of `/proc/sys/net/ipv4/tcp_rmem` (`tcp_wmem` for send), not `rmem_default` - TCP overrides that one.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumerConfig.cs
```diff
@@ -76,14 +76,14 @@ public class IggyConsumerConfig
     public string PersonalAccessToken { get; set; } = string.Empty;
 
     /// <summary>
-    ///     The size of the receive buffer in bytes. Default is 4096.
+    ///     The size of the receive buffer in bytes. Default is null.
```
hubcio: also at line 84, `IggyPublisherConfig.cs` lines 97 and 103, `IggyPublisherBuilder.cs` lines 112 and 113.

every place that documents the null default should say the same thing - right now `IggyPublisherConfig` says `(OS default)` and the rest say nothing.

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs
```diff
@@ -1066,8 +1066,15 @@ private async Task TryEstablishConnectionAsync(bool autoLogin, bool settleOnLead
             try
             {
                 socket = new Socket(ServerAddress.AddressFamilyOf(host), SocketType.Stream, ProtocolType.Tcp);
-                socket.SendBufferSize = _configuration.SendBufferSize;
-                socket.ReceiveBufferSize = _configuration.ReceiveBufferSize;
+                if (_configuration.SendBufferSize is > 0)
```
hubcio: `is > 0` quietly turns 0 and negatives into the OS default, while the builders reject them. check `HasValue` here and add the same `<= 0` check to `IggyClientFactory.Validate`, so a raw configurator fails the same way.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumerBuilder.cs
```diff
@@ -347,14 +347,14 @@ protected virtual void Validate()
                 "AutoCommitMode.Auto with a message encryptor risks silent message loss: the offset is committed before decryption. Use AutoCommitMode.AfterReceive or AutoCommitMode.Disabled.");
         }
 
-        if (Config.ReceiveBufferSize <= 0)
+        if (Config.ReceiveBufferSize is <= 0)
```
hubcio, line 366: nothing covers this - add a test in `IggyConsumerBuilderTests` (and the publisher one) for `receiveBufferSize: 0` throwing and the default null building fine.

foreign/csharp/README.md
```diff
@@ -80,9 +80,9 @@ var client = IggyClientFactory.CreateClient(new IggyClientConfigurator
     BaseAddress = "127.0.0.1:8090",
     Protocol = Protocol.Tcp,
 
-    // Buffer sizes (optional, default: 4096)
-    ReceiveBufferSize = 4096,
-    SendBufferSize = 4096,
+    // Buffer sizes (optional, in bytes, default: null (for OS automatic))
```
hubcio: nested parens read badly - `// Socket buffer sizes in bytes (optional, null = OS default)`.

hubcio (changes requested):
`foreign/csharp/Benchmarks/Program.cs` still pins `int.MaxValue` per OS. worth re-measuring against the new default and dropping the override if it no longer buys anything.

spetz (approved):

hubcio (approved):

spetz:
Thank you for the contribution, I pushed a few more fixes, so that we could merge this before the upcoming release :)

PR #4017, opened 2026-09-01
feat(python): expose partition management

### LINKED ISSUE #4014: Python SDK: expose create_partitions and delete_partitions
### Description

The Rust client exposes `create_partitions` and `delete_partitions` (`core/common/src/traits/partition_client.rs`); the Python SDK exposes neither, so a Python user cannot add or remove partitions on an existing topic without dropping to `send_binary_request`. Part of the parity work tracked in #3893.

  ```rust
  async fn create_partitions(&self, stream_id: &Identifier, topic_id: &Identifier,
                            partitions_count: u32) -> Result<(), IggyError>;
  async fn delete_partitions(&self, stream_id: &Identifier, topic_id: &Identifier,
                            partitions_count: u32) -> Result<(), IggyError>;
  ```

  Note that `delete_partitions` removes the last N partitions: a topic with 5, minus 2, keeps partitions 1-3.

### Affected area / component

Python SDK

### Proposed solution

 - Add both methods to the `#[pymethods]` block in `foreign/python/src/client.rs`, taking `stream_id: PyIdentifier`, `topic_id: PyIdentifier` and `partitions_count: int`. Both return `None` on success and raise on failure, matching `delete_topic` alongside them.
  - Regenerate the stub entries in `foreign/python/apache_iggy.pyi`.
  - Add `foreign/python/tests/test_partition.py`: create a topic, add partitions, assert the count via `get_topic`, delete partitions, assert the count again.

No new wrapper types are needed - delete_topic and purge_topic in the same file take the same identifier arguments and return None, so they are the pattern to follow.

### Alternatives considered

_No response_

### Contribution

### Good first issue

Elioooon:
Closes #4014

## Summary

Expose `create_partitions` and `delete_partitions` on the Python `IggyClient`, delegating to the existing Rust SDK operations and error conversion. Document ID reuse, validation limits, asynchronous materialization, consumer-group rebalancing, and the permission hierarchy. Creating a topic accepts zero partitions; adding or deleting partitions requires a positive count.

The Python tests cover string/numeric IDs, invalid bounds, authorization and topic scoping, deleting all partitions, and retained-versus-deleted message statistics. The simulator regression checks partition, topic, and stream counters after a tombstoned partition with buffered replication gaps is removed.

The server counter rollback is supplied by upstream #4046. This PR retains the simulator assertions and makes no changes to the shard or metadata implementation relative to current master.

## Design and risk boundary

Reuse the Rust SDK's existing async operations instead of introducing a separate Python protocol implementation. The bindings do not change the wire format or server admission rules. Partition materialization and deletion remain asynchronous, so callers must observe the updated topic before assuming the new topology is available. The simulator deliberately injects a replication gap before removal; it is not a substitute for the Python running-server integration suite.

## Validation

Latest review follow-up, after merging master including #4046:

- `cargo fmt --manifest-path foreign/python/Cargo.toml -- --check`
- `cargo clippy -p shard -p metadata -p simulator -p server --all-targets --features hwlocality/vendored --offline -- -D warnings`
- `cargo clippy --manifest-path foreign/python/Cargo.toml --all-targets --all-features --offline -- -D warnings`
- `cargo test -p metadata --offline` (162 passed; rerun outside the restricted sandbox to allow io_uring runtime creation)
- `cargo test -p simulator -p server --features hwlocality/vendored --offline given_buffered_gap_drops_when_the_namespace_is_removed_should_still_count_them -- --nocapture` (1 passed)
- `cargo test -p server --features hwlocality/vendored --offline partitions_count_cap_denies_pre_consensus` (1 passed)
- `cargo test --manifest-path foreign/python/Cargo.toml --lib --offline` (3 passed)
- `cargo run --manifest-path foreign/python/Cargo.toml --bin stub_gen --offline` with Python 3.13, followed by `ruff format foreign/python/apache_iggy.pyi`
- `ruff check foreign/python/apache_iggy.pyi foreign/python/tests/test_partition.py`
- `git diff --check upstream/master`

The Python running-server integration tests were not rerun locally in this follow-up; the previous head's Python CI passed, and the updated head requires a fresh CI run. Local `cargo sort` and Taplo checks are unavailable because those tools are not installed; this follow-up does not edit Cargo/TOML files.

Disclosure: This contribution was developed with assistance from generative AI. I reviewed the implementation and ran the validation listed above locally.


foreign/python/src/client.rs
```diff
@@ -777,6 +777,72 @@ impl IggyClient {
         })
     }
 
+    /// Create partitions for a topic.
+    ///
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     partitions_count: Number of partitions to create.
```
ethanlin01x: Add as `int`. Same fix for `delete_partitions`

justinmclean:
Clean, minimal binding. It follows the existing `purge_topic` shape exactly and does everything #4014 asked for: both methods on the `#[pymethods]` block, regenerated stubs, and a test that round-trips the count through `get_topic`. Two observations, neither blocking.

### Smaller observations

- `foreign/python/src/client.rs:791` and `:824` — the `Raises:` block says `RuntimeError: If an identifier is invalid or the request fails.`, but an invalid identifier never reaches the request. `impl TryFrom<&PyIdentifier> for Identifier` in `foreign/python/src/identifier.rs` maps both the string and the numeric arm to `PyValueError`, and `TryFrom<PyIdentifier>` delegates to it, so `create_partitions("", "topic", 2)` raises `ValueError`. `create_consumer_group`, the next method in this file, documents the same code path correctly:

```text
  /// Raises:
  ///     ValueError: If an identifier is invalid.
  ///     RuntimeError: If the request fails.
```

  Splitting the line the same way and re-running `cargo run --bin stub_gen` would keep `apache_iggy.pyi` in step. To be clear about where this came from: it is a file-wide pattern, not something this PR introduced. `purge_topic`, `delete_topic` (`:731`) and `get_topics` (`:628`) conflate the two the same way, and `get_topic` collapses it to a single "raises RuntimeError on failure" line. Fixing the whole file is its own change; matching `create_consumer_group` in the two new methods is enough here.

- `foreign/python/tests/test_partition.py:33-41` — this matches #4014's test spec exactly, so nothing is missing. One optional strengthening if you feel like it: the count assertions hold whichever partitions the server removed, and #4014's design note is specifically that `delete_partitions` drops the *last* N. `TopicDetails.partitions[].id` is exposed and `test_topic.py::TestGetTopic::test_get_topic_partitions` already reads ids that way, so asserting that the surviving ids are the first N would pin the documented semantics rather than just the arithmetic.

Nothing here blocks the merge.

---

This review was drafted by an AI-assisted tool (Apache Magpie), so it may contain mistakes. If you think one of them is misapplied, please reply on the PR, and a maintainer will weigh in.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,72 @@ impl IggyClient {
...
+        future_into_py(py, async move {
+            inner
+                .create_partitions(&stream_id, &topic_id, partitions_count)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(())
+        })
+    }
+
+    /// Delete partitions from a topic.
```
ethanlin01x: Please document that this deletes the last `partitions_count` partitions and drops their messages.

Elioooon:
@justinmclean Thanks for the careful review. I updated both new methods to document `ValueError` for identifier conversion separately from request `RuntimeError`, regenerated and formatted the stub, and strengthened the integration test to assert partition IDs `[0, 1, 2, 3]` after creation and `[0, 1]` after deleting the last two. Revalidated Rust fmt/check/clippy, Ruff, and the local-server partition integration test (`1 passed`).

ethanlin01x (commented):
The implementation looks correct to me, but it would be better to add more test coverage.`test_partition.py` only covers one happy path with string ids. Could you add:

- numeric stream/topic ids — the only type conversion here is untested (see `test_consumer_group.py:101` for the pattern)
- assert the remaining partition ids after delete, not just the count — `delete_partitions` removes the *last* N
- `partitions_count = 0` is rejected on both `create_partitions` and `delete_partitions` — the binding does no client-side validation, so this behaviour comes from the server and is worth pinning down
- deleting more partitions than the topic has is rejected
- missing stream/topic raises `RuntimeError`

Elioooon:
@ethanlin01x Thanks for the review. Added the requested coverage in `a4aab32`:

- numeric stream/topic IDs exercise both create and delete
- surviving partition IDs are asserted after deletion
- zero-count create and delete are rejected
- deleting more partitions than the topic contains is rejected
- missing stream and missing topic are covered for both operations

Revalidated with:

- `cargo fmt --check`
- `foreign/python/.venv/bin/ruff check foreign/python/tests/test_partition.py`
- `foreign/python/.venv/bin/ruff format --check foreign/python/tests/test_partition.py`
- local `iggy-server --with-default-root-credentials --fresh` plus `foreign/python/.venv/bin/pytest -q foreign/python/tests/test_partition.py` (5 passed)

foreign/python/src/client.rs
Elioooon: Addressed in ec33ea7: `test_partition_management_accepts_numeric_ids` now exercises both `create_partitions` and `delete_partitions` with integer stream and topic IDs.

foreign/python/src/client.rs
Elioooon: Addressed in 8080ce3. The Python docstring now states that deletion removes the last `partitions_count` partitions and all messages stored in them, and the generated `.pyi` stub is synchronized.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
         })
     }
 
+    /// Create partitions for a topic.
+    ///
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     partitions_count: Number of partitions to create.
```
hubcio: warning: `partitions_count` has no documented range. server takes 1..=1000 (`MAX_PARTITIONS_PER_REQUEST`) and reports 0 as "Too many partitions", so spell it out here - the stub is the only doc a python caller gets.

also at line 819.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+):
+    stream_name = unique_name()
+    topic_name = unique_name()
+
+    await iggy_client.create_stream(stream_name)
+    await iggy_client.create_topic(
+        stream=stream_name, name=topic_name, partitions_count=2
+    )
+
+    with pytest.raises(RuntimeError):
```
hubcio: warning: bare `pytest.raises(RuntimeError)` passes on any failure - every server error maps to `RuntimeError` here. add `match=`: "Too many partitions" for zero count, "Invalid partitions count" for over-count, "was not found." for missing stream/topic.

also at lines 84, 100, 117, 119, 121, 123.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
...
+            Ok(())
+        })
+    }
+
+    /// Delete the last partitions from a topic, including all messages stored in them.
+    ///
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     partitions_count: Number of partitions to delete from the end of the topic.
```
hubcio: nit: `partitions_count` drops the "as `int`" suffix the other args in this same docstring carry. adding it pushes the generated stub line past 88 chars, so rewrap rather than extend.

also at line 785.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
...
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the partitions are deleted.
```
hubcio: nit: the awaitable resolves on metadata commit, not after teardown - the unlink runs later in the reconciler and can back off. say "when the deletion is accepted; storage teardown completes asynchronously".

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
...
+        future_into_py(py, async move {
+            inner
+                .create_partitions(&stream_id, &topic_id, partitions_count)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+            Ok(())
+        })
+    }
+
+    /// Delete the last partitions from a topic, including all messages stored in them.
```
hubcio, line 822: nit: neither docstring mentions the consumer group side effect - delete rebalances members off the removed partitions and drops their stats, create leaves the new ones unassigned until the next rebalance. one line on each.

also at line 780.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
...
+    ///
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     partitions_count: Number of partitions to create.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the partitions are created.
+    ///
+    /// Raises:
```
hubcio, line 797: nit: `Raises:` misses `OverflowError`, which pyo3 throws for a negative or oversized count and which is neither `ValueError` nor `RuntimeError`. `test_topic.py:367` already pins that behaviour.

also at line 824.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
...
 
+    /// Create partitions for a topic.
+    ///
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     partitions_count: Number of partitions to create.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the partitions are created.
```
hubcio, line 795: nit: nothing says which ids the new partitions get, and the caller needs them for `send_messages`. they're 0-based and append above the current max, reused after a delete - don't copy `partition_client.rs:26`, it claims 1-based and is wrong.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+    topic = await iggy_client.get_topic(stream.id, topic_name)
+    assert topic is not None
+
+    await iggy_client.create_partitions(stream.id, topic.id, 1)
+    await iggy_client.delete_partitions(stream.id, topic.id, 1)
+
+    topic = await iggy_client.get_topic(stream.id, topic.id)
+    assert topic is not None
+    assert [partition.id for partition in topic.partitions] == [0, 1]
```
hubcio: nit: this asserts `[0, 1]`, the same state as before the create and delete, so two no-ops that cancel out would pass. re-read between them: `created = await iggy_client.get_topic(stream.id, topic.id)`, then assert `created.partitions_count == 3`.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+
+from apache_iggy import IggyClient
+
+
+@pytest.mark.asyncio
+async def test_create_and_delete_partitions(iggy_client: IggyClient, unique_name):
```
hubcio: nit: no test in this file ever sends a message, so delete is never exercised with data present and the new partitions are never shown to be usable. `test_topic.py:1258` is the shape to copy. don't assert topic `size_bytes` after the delete though - it isn't reclaimed.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+    await iggy_client.create_topic(
+        stream=stream_name, name=topic_name, partitions_count=2
+    )
+
+    with pytest.raises(RuntimeError):
+        await iggy_client.delete_partitions(stream_name, topic_name, 3)
+
+
+@pytest.mark.asyncio
+async def test_partition_management_rejects_missing_stream_or_topic(
```
hubcio: nit: the documented `ValueError` for a bad identifier is never exercised here, and there's no delete-all-partitions case. skip an over-cap test though - zero and over-cap return the same code, so it would duplicate the zero-count assertion.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import pytest
+
+from apache_iggy import IggyClient
+
+
+@pytest.mark.asyncio
```
hubcio: nit: module-level test functions - 9 of the 10 other suites group into `Test*` classes. wrap these in `class TestPartitionManagement:`.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+async def test_create_and_delete_partitions(iggy_client: IggyClient, unique_name):
+    stream_name = unique_name()
```
hubcio, line 24: simplification: the same 7-line create_stream + create_topic preamble is copy-pasted in four tests. pull it into a module-level helper like `test_consumer_group.py:37` - there is already a TODO there asking for exactly this.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,124 @@
...
+    topic_name = unique_name()
+    missing_name = unique_name()
+
+    await iggy_client.create_stream(stream_name)
+    await iggy_client.create_topic(
+        stream=stream_name, name=topic_name, partitions_count=2
+    )
+
+    with pytest.raises(RuntimeError):
```
hubcio: simplification: these four raises blocks only differ in which identifier is missing. one `parametrize` over `(method, missing)` collapses them and keeps the test name.

foreign/python/src/client.rs
```diff
@@ -777,6 +777,74 @@ impl IggyClient {
...
+    ) -> PyResult<Bound<'a, PyAny>> {
+        let stream_id = Identifier::try_from(stream_id)?;
+        let topic_id = Identifier::try_from(topic_id)?;
+        let inner = self.inner.clone();
+
+        future_into_py(py, async move {
+            inner
+                .create_partitions(&stream_id, &topic_id, partitions_count)
+                .await
+                .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
```
hubcio: simplification: this makes 38 copies of the same `PyErr::new::<PyRuntimeError, _>(e.to_string())` map_err in this file. a `to_runtime_error(e: impl ToString)` helper, like `to_value_error` in `send_message.rs:93`, cuts each site to `.map_err(to_runtime_error)?`.

also at line 843.

hubcio (changes requested):
small note on the PR description - the validation block says `1 passed`, but the file now has 5 tests, the last four added after that run.

found other problem during review, in server: `delete_partitions` never rolls the deleted partitions' bytes back out of `TopicStats` / `StreamStats` - `core/metadata/src/stm/stream.rs:2232` evicts the child entries only - so `get_topic`, `get_stream` and `/metrics` keep counting them until restart. purge does it right via `zero_out_all`, and `delete_topic` has the same gap one level up - we will fix it (ignore it in your pr)

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. Both docstrings now state that `partitions_count` is an `int` in the range 1..=1000; deletion additionally documents that the count cannot exceed the topic's current partition count. The generated stub is synchronized.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. The tests now match `Too many partitions`, `Invalid partitions count`, and `was not found.` on their respective server error paths, so unrelated `RuntimeError`s no longer satisfy them.

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. Both count parameters are documented as `int`, with the generated stub lines wrapped rather than extended past the formatter width.

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. The return contract now says deletion is accepted at awaitable completion and storage teardown completes asynchronously.

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. Create documents that new partitions remain unassigned until the next consumer-group rebalance; delete documents that groups are rebalanced away from removed partitions.

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. Both methods now document `OverflowError` for negative or oversized Python integers, with parametrized tests for -1 and 2**32.

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. The create docstring now describes consecutive zero-based IDs above the current maximum and reuse after deletion.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. The numeric-ID test now reads immediately after create and asserts count 3 and IDs `[0, 1, 2]` before deleting and asserting the final state.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. The success test sends a message to newly created partition 3 before deleting it. The server-side fix also zeros deleted `PartitionStats` before eviction so their message and byte counters are rolled out of topic and stream totals; the metadata regression test pins this behavior.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. Parametrized tests now exercise invalid identifiers as `ValueError`, and a separate case deletes every partition and asserts an empty partition list. I skipped the redundant over-cap test as requested.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. The module-level tests are now grouped under `TestPartitionManagement`.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. The repeated stream/topic setup is centralized in the module-level `_create_topic` helper.

foreign/python/tests/test_partition.py
Elioooon: Addressed in 67df430. Missing stream/topic coverage is parametrized over both methods and both missing identifier positions.

foreign/python/src/client.rs
Elioooon: Addressed in 67df430. A shared `to_runtime_error` helper is used by both new bindings, so they do not add more copies of the conversion pattern.

Elioooon:
Addressed the requested changes in 67df430.

The server-side root cause was that `remove_partitions_from` evicted deleted `PartitionStats` entries without first rolling their counters out of the parent `TopicStats` and `StreamStats`. It now detaches entries under the registry lock, releases the lock, then calls `zero_out_all()` on each removed counter. Replayed left-right applies remain idempotent because the entries are already absent.

I also tightened the Python contract and tests: numeric IDs now assert the intermediate created state; error tests match specific server messages; invalid identifiers, Python integer overflow, delete-all, and a message-bearing new partition are covered; the API docs include count bounds, partition IDs, consumer-group effects, asynchronous teardown, and exceptions.

Local validation:
- `cargo clippy -p metadata --all-targets -- -D warnings`
- `cargo test -p metadata` (146 passed)
- `cargo clippy --lib -- -D warnings` from `foreign/python`
- `cargo check --lib` from `foreign/python`
- targeted whitespace/newline checks

The Python integration suite and generated-stub consistency are left to the standard CI environment because this host has Python 3.7 shared (below PyO3 minimum) and Python 3.13 static (not linkable into the extension).

Elioooon:
Base branch sync completed without conflicts at `c8d918136` (using GitHub's update-branch flow, no force-push). I re-ran the focused validation on the updated head:

- `cargo clippy -p metadata --all-targets -- -D warnings`
- `cargo test -p metadata` — 150 passed
- `cargo check --lib` and `cargo clippy --lib -- -D warnings` in `foreign/python`
- `python -m py_compile tests/test_partition.py`

All passed locally. The two pull-request workflow runs currently show `action_required`, so GitHub Actions is waiting for maintainer approval rather than reporting a code/test failure.

foreign/python/apache_iggy.pyi
```diff
@@ -1194,6 +1194,56 @@ class IggyClient:
...
+    ) -> collections.abc.Awaitable[None]:
+        r"""
+        Delete the last partitions from a topic, including all messages stored in them.
+        Consumer groups are rebalanced away from the removed partitions.
+
+        Args:
+            stream_id: Stream identifier as `str | int`.
+            topic_id: Topic identifier as `str | int`.
+            partitions_count: Number of partitions to delete as `int` from the end of
+                the topic; must be 1..=1000 and no greater than its current count.
```
slbotbm, line 1236: `1..=1000` is rust lingo, not suitable for python. Write it in a way python programmers will understand.

foreign/python/apache_iggy.pyi
```diff
@@ -1194,6 +1194,56 @@ class IggyClient:
...
+        r"""
+        Create partitions for a topic. New partitions use consecutive, zero-based IDs
+        after the current maximum; IDs removed by deletion can be reused. Existing
+        consumer groups leave them unassigned until their next rebalance.
+
+        Args:
+            stream_id: Stream identifier as `str | int`.
+            topic_id: Topic identifier as `str | int`.
+            partitions_count: Number of partitions to create as `int`; must be
+                1..=1000.
```
slbotbm, line 1212: same for this

foreign/python/src/client.rs
```diff
@@ -57,6 +57,10 @@ pub struct IggyClient {
     inner: Arc<RustIggyClient>,
 }
 
+fn to_runtime_error(error: impl ToString) -> PyErr {
+    PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(error.to_string())
```
slbotbm, line 61: No need for a new function. Inline this at call sites.

foreign/python/apache_iggy.pyi
slbotbm, line 1: In the docs, also mention if there are any permission-related restrictions for these two functions.

foreign/python/apache_iggy.pyi
```diff
@@ -1194,6 +1194,56 @@ class IggyClient:
...
+    def create_partitions(
+        self,
+        stream_id: builtins.str | builtins.int,
+        topic_id: builtins.str | builtins.int,
+        partitions_count: builtins.int,
+    ) -> collections.abc.Awaitable[None]:
+        r"""
+        Create partitions for a topic. New partitions use consecutive, zero-based IDs
+        after the current maximum; IDs removed by deletion can be reused. Existing
+        consumer groups leave them unassigned until their next rebalance.
```
slbotbm, line 1206: > Existing consumer groups leave them unassigned until their next rebalance.

This documents wrong behaviour. A clearer description would be 
 > Creating partitions immediately rebalances existing consumer groups, redistributing all topic partitions among their current members and advancing the group generation.

foreign/python/apache_iggy.pyi
slbotbm: Also add Python integration tests for authorization. For both create_partitions and delete_partitions, verify that a logged in user without manage_topic permission receives Unauthorized and the partition count remains unchanged. Also verify that a user granted topic-level manage_topic can perform both operations. A topic-scoping check, proving permission for topic A does not authorize topic B, would be valuable.

slbotbm (dismissed):
You can also run the team-review skill in .claude folder on your own as well to iron out common bugs.

foreign/python/src/client.rs
hubcio: actually this was my comment, imho it looks looks cleaner this way.

slbotbm: This change only replaces the two instances in this PR, but there are a lot more instances in the file. Let's change every place that calls the error into this function then

core/metadata/src/stm/stream.rs
```diff
@@ -561,12 +561,25 @@ impl StatsRegistry {
...
+                removed.push(entry.stats.clone());
+            }
+            keep
+        });
+        drop(partitions);
+
+        // Roll counters out of the parents after releasing the registry lock.
+        // A second left-right apply finds no entries and is a no-op.
+        for stats in removed {
+            stats.zero_out_all();
```
hubcio, line 581: warning: this zero runs while the owning shard can still touch the partition - a late segment-cleaner decrement wraps the partition counter and double-subtracts from topic and stream totals, a late append leaks into them. move the roll-out to the `ConfirmRemove` arm, where writes are fenced.

core/metadata/src/stm/stream.rs
```diff
@@ -561,12 +561,25 @@ impl StatsRegistry {
     }
```
hubcio, line 561: warning: pre-existing, not blocking - `remove_topic` above has the same leak one level up: `TopicStats` is dropped without rolling it out of `StreamStats`, so `get_stream` and `get_stats` keep a deleted topic's totals until restart. zeroing partitions at `ConfirmRemove` cascades through the topic and covers this too.

core/metadata/src/stm/stream.rs
```diff
@@ -561,12 +561,25 @@ impl StatsRegistry {
...
+        partitions.retain(|(sid, tid, pid), entry| {
+            let keep = !(*sid == stream_id && *tid == topic_id && *pid >= first_removed);
+            if !keep {
+                removed.push(entry.stats.clone());
+            }
+            keep
+        });
+        drop(partitions);
+
+        // Roll counters out of the parents after releasing the registry lock.
```
hubcio, line 578: nit: not true - the second absorb runs after the next publish, and the reconciler's get-or-create can re-insert the key in between, so the second apply does find entries. replay is safe only because `zero_out_all` swaps, say that instead.

core/metadata/src/stm/stream.rs
```diff
@@ -561,12 +561,25 @@ impl StatsRegistry {
...
+        let mut partitions = self
+            .partitions
             .lock()
-            .expect("stats registry mutex poisoned")
-            .retain(|(sid, tid, pid), _| {
-                !(*sid == stream_id && *tid == topic_id && *pid >= first_removed)
-            });
+            .expect("stats registry mutex poisoned");
+        let mut removed = Vec::new();
+        partitions.retain(|(sid, tid, pid), entry| {
```
hubcio, line 569: simplification: if the zero stays here, do it under the guard - `zero_out_all` only touches atomics, so releasing the lock first buys nothing. `partitions.extract_if(|(sid, tid, pid), _| *sid == stream_id && *tid == topic_id && *pid >= first_removed).for_each(|(_, entry)| entry.stats.zero_out_all())` replaces the vec, the clones and the `drop`.

core/metadata/src/stm/stream.rs
```diff
@@ -3069,6 +3082,62 @@ mod tests {
...
+        assert_eq!(partition_stats[0].segments_count_inconsistent(), 1);
+        for stats in &partition_stats[1..] {
+            assert_eq!(stats.messages_count_inconsistent(), 0);
+            assert_eq!(stats.size_bytes_inconsistent(), 0);
+            assert_eq!(stats.segments_count_inconsistent(), 0);
+        }
+        assert_eq!(topic_stats.messages_count_inconsistent(), 1);
+        assert_eq!(topic_stats.size_bytes_inconsistent(), 100);
+        assert_eq!(inner.items[0].stats.messages_count_inconsistent(), 1);
+        assert_eq!(inner.items[0].stats.size_bytes_inconsistent(), 100);
```
hubcio, line 3135: nit: if this test stays, also assert `segments_count_inconsistent()` on topic and stream, and run the delete through both left-right buffers - clone `inner` before the delete and apply to both copies.

foreign/python/src/client.rs
```diff
@@ -777,6 +781,82 @@ impl IggyClient {
         })
     }
 
+    /// Create partitions for a topic. New partitions use consecutive, zero-based IDs
+    /// after the current maximum; IDs removed by deletion can be reused. Existing
+    /// consumer groups leave them unassigned until their next rebalance.
```
hubcio, line 786: warning: this is backwards - `CreatePartitions` rebalances every group in the same commit (`rebalance_consumer_groups` in the STM apply): generation bump, all partitions redistributed round-robin, pending revocations dropped. say that, and regenerate the stub.

also at line 823 - the delete side understates it the same way.

foreign/python/src/client.rs
```diff
@@ -777,6 +781,82 @@ impl IggyClient {
         })
     }
 
+    /// Create partitions for a topic. New partitions use consecutive, zero-based IDs
```
hubcio, line 784: nit: "zero-based" only holds for an empty topic, ids start at max + 1. say they continue from one past the current highest id and that freed ids are reused.

foreign/python/src/client.rs
```diff
@@ -777,6 +781,82 @@ impl IggyClient {
...
+    /// after the current maximum; IDs removed by deletion can be reused. Existing
+    /// consumer groups leave them unassigned until their next rebalance.
+    ///
+    /// Args:
+    ///     stream_id: Stream identifier as `str | int`.
+    ///     topic_id: Topic identifier as `str | int`.
+    ///     partitions_count: Number of partitions to create as `int`; must be
+    ///         1..=1000.
```
hubcio, line 792: nit: the `1..=1000` literal duplicates `MAX_PARTITIONS_COUNT`, and `create_topic` documents no cap at all. either drop the number or add it there too.

also at line 829.

foreign/python/src/client.rs
```diff
@@ -777,6 +781,82 @@ impl IggyClient {
...
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `None` when the partitions are created.
```
hubcio, line 795: nit: create is just as async as delete - the reply comes after the STM apply and the reconciler materializes storage later, sends park until then. say "when the partitions are committed" here, same shape as the delete docstring.

foreign/python/src/client.rs
```diff
@@ -57,6 +57,10 @@ pub struct IggyClient {
     inner: Arc<RustIggyClient>,
 }
 
+fn to_runtime_error(error: impl ToString) -> PyErr {
```
hubcio, line 60: nit: this helper has 2 callers while 38 siblings in this file still inline `PyErr::new::<PyRuntimeError, _>(e.to_string())`. inline these two like the rest, or move everything over in one sweep.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,149 @@
...
+        assert [partition.id for partition in created.partitions] == [0, 1, 2, 3]
+
+        await iggy_client.send_messages(
+            stream_name, topic_name, 3, [SendMessage("partition payload")]
+        )
+        await iggy_client.delete_partitions(stream_name, topic_name, 2)
+        deleted = await iggy_client.get_topic(stream_name, topic_name)
+        assert deleted is not None
+        assert deleted.partitions_count == 2
+        assert [partition.id for partition in deleted.partitions] == [0, 1]
```
hubcio, line 54: warning: nothing end-to-end checks the stats roll-out - the only message goes to partition 3, which gets deleted, so an over-eviction bug would pass too. send to partition 0 as well, then poll `get_topic` until `messages_count == 1` with partition 0 still holding its message, since stats fold asynchronously.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @pytest.mark.asyncio
+    async def test_partition_management_rejects_zero_count(
+        self, iggy_client: IggyClient, unique_name
+    ):
+        stream_name, topic_name = await _create_topic(iggy_client, unique_name)
+
+        with pytest.raises(RuntimeError, match="Too many partitions"):
```
hubcio, line 84: nit: zero hitting "Too many partitions" is a deliberate legacy quirk of `validate_partitions_change_count` - leave a one-line comment so nobody "fixes" the error code and breaks this test.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,149 @@
...
+        assert [partition.id for partition in created.partitions] == [0, 1, 2]
+
+        await iggy_client.delete_partitions(stream.id, topic.id, 1)
+        deleted = await iggy_client.get_topic(stream.id, topic.id)
+        assert deleted is not None
+        assert deleted.partitions_count == 2
+        assert [partition.id for partition in deleted.partitions] == [0, 1]
+
+    @pytest.mark.asyncio
+    async def test_partition_management_rejects_zero_count(
```
hubcio, line 79: nit: the docstrings promise `1..=1000` but only 0 is tested. parametrize over `[0, 1001]`, both hit `TooManyPartitions` before consensus.

foreign/python/tests/test_partition.py
```diff
@@ -0,0 +1,149 @@
...
+            stream_name, topic_name, 3, [SendMessage("partition payload")]
+        )
+        await iggy_client.delete_partitions(stream_name, topic_name, 2)
+        deleted = await iggy_client.get_topic(stream_name, topic_name)
+        assert deleted is not None
+        assert deleted.partitions_count == 2
+        assert [partition.id for partition in deleted.partitions] == [0, 1]
+
+    @pytest.mark.asyncio
+    async def test_partition_management_accepts_numeric_ids(
```
hubcio, line 57: simplification: this repeats the create/delete flow above with int ids. parametrize the plain flow over name vs numeric ids and keep the message-bearing delete as its own test.

hubcio (changes requested):
inline comments cover the diff. two pre-existing items outside it, not blocking:

nit: `core/metadata/src/stm/stream.rs:2204` - the comment "Added partitions are unassigned until the groups rebalance." sits right above the call that rebalances them, and it is where the new docstring came from. reword it.

nit: `foreign/python/src/client.rs:629`, `:732`, `:763` and the matching stub lines - `get_topics`, `delete_topic` and `purge_topic` say RuntimeError for an invalid identifier, but `Identifier::try_from` raises ValueError, as the new methods correctly document. follow-up sweep.

Elioooon:
Addressed the requested changes in `120f5d17a` and merged the latest `upstream/master` without force-pushing.

Key fixes:
- moved stats roll-out from metadata eviction to shard `ConfirmRemove`, after the tombstone/write fence
- added a simulator regression asserting partition, topic, and stream counters all return to zero
- added Python authorization coverage for unauthorised users, topic-level `manage_topic`, and topic scoping
- strengthened message-bearing deletion coverage to retain partition 0 while deleting partition 3
- corrected partition ID, rebalance, async materialization, permission, and count-limit docs; regenerated the stub
- applied the shared RuntimeError conversion consistently across the Python client

Local validation is now reflected in the PR body, including metadata tests, shard/metadata clippy, the targeted simulator test with vendored hwloc, Python Rust checks, stub generation, Ruff, and py_compile. The full Python running-server integration suite is left to CI.

Elioooon:
The newly approved CI run caught a formatting regression in my branch after the upstream merge: replacing error closures with `to_runtime_error` left 11 expressions in `foreign/python/src/client.rs` that Rust 1.98 rustfmt reformats. The Python lint job failed at `cargo fmt`; Ruff passed, and Clippy/pyrefly were not reached. The Python test job was cancelled, so this run does not establish integration-test success.

Fixed in `2969bc30c` with rustfmt only. Local verification:

```text
cargo fmt --manifest-path foreign/python/Cargo.toml -- --check
cargo clippy --manifest-path foreign/python/Cargo.toml --all-targets --all-features --offline -- -D warnings
cargo test --manifest-path foreign/python/Cargo.toml --lib --offline
# 3 passed
```

Formatting and Clippy pass with the pinned Rust 1.98.0 toolchain. No API or behavior changes; I did not rerun the server-backed Python integration suite for this formatting-only follow-up. The new head still needs CI verification.

core/shard/src/lib.rs
```diff
@@ -2530,6 +2530,10 @@ where
                     self.metrics.record_partition_removed();
                     confirmed_remove = true;
                     if let Some(partition) = removed {
+                        // The tombstone fences writes before this point. Rolling the
+                        // detached counters out here prevents late shard activity from
+                        // underflowing or leaking into the topic and stream totals.
+                        partition.stats.zero_out_all();
```
hubcio: warning: #4046 landed on master with this same `zero_out_all()` call a few lines below, so after merge it runs twice. drop this hunk and the roll-out claim in the PR body; the simulator assertions still hold on the master call.

foreign/python/src/client.rs
```diff
@@ -704,7 +681,8 @@ impl IggyClient {
     /// Args:
     ///     stream: Stream identifier as `str | int`.
     ///     name: Topic name as `str`.
-    ///     partitions_count: Number of partitions as `int`.
+    ///     partitions_count: Number of partitions as `int`, between 1 and 1000
```
hubcio: nit: `create_topic` accepts zero partitions - admission only rejects counts above 1000, and the HTTP validator allows 0 too. say "at most 1000" here and regenerate the stub.

hubcio (changes requested):

core/shard/src/lib.rs
Elioooon: Addressed in a23d05d57 after merging master at a71ed7420 (#4046). Removed the duplicate call and its comment; core/shard/src/lib.rs now has no diff from master. The simulator counter assertions remain and pass with the upstream cleanup (1 fault-injection test passed); metadata tests also pass (162). Updated the PR body to attribute the rollback to #4046 and describe only the remaining binding/test changes.

foreign/python/src/client.rs
Elioooon: Addressed in a23d05d57: changed create_topic to “at most 1000” and regenerated apache_iggy.pyi with stub_gen, then applied Ruff formatting. Confirmed the shared admission helper accepts zero and ran partitions_count_cap_denies_pre_consensus (1 passed). The positive-count constraint remains on create_partitions/delete_partitions. Python Clippy, Ruff, and the 3 binding unit tests pass.

hubcio (approved):

slbotbm:
@Elioooon is this pr ready for review?

spetz (approved):
LGTM

PR #4018, opened 2026-09-01
feat(python): expose get_stats on IggyClient

### LINKED ISSUE #4016: Python SDK: expose get_stats
### Description

`get_stats` is the server's headline diagnostic call and is exposed by every other SDK; the Python SDK has no way to reach it. Part of the parity work tracked in #3893.

```rust
async fn get_stats(&self) -> Result<Stats, IggyError>;
```

`Stats` carries scalar counts (streams, topics, partitions, segments, messages, clients, consumer groups), host and version strings, thread count, free and total disk space, and `cache_metrics: HashMap<CacheMetricsKey, CacheMetrics>`.

### Affected area / component

Python SDK

### Proposed solution

- Add `foreign/python/src/stats.rs` wrapping `Stats`, `CacheMetrics` and `CacheMetricsKey`, following `src/user.rs`. `CacheMetricsKey` needs `__hash__` and `__eq__` so it can key a Python dict.
- Expose `cache_metrics` as a `dict[CacheMetricsKey, CacheMetrics]`.
- Add `get_stats` to `src/client.rs` and the stub entries to `foreign/python/apache_iggy.pyi`.
- Add `foreign/python/tests/test_stats.py`: create a stream and topic, send messages, then assert the counts moved and the version string is non-empty.

### Alternatives considered

_No response_

### Contribution

### Good first issue

7487:

Closes #4016

## Rationale

`get_stats` is the server's headline diagnostic call and is exposed by every other SDK; the Python SDK could not reach it.

## What changed?

The Python SDK had no binding for `get_stats`, so server counts, host/version details and cache metrics were unreachable from Python.

A new `foreign/python/src/stats.rs` wraps `Stats`, `CacheMetrics` and `CacheMetricsKey` following the `user.rs` pattern. `CacheMetricsKey` is frozen with `__eq__`/`__hash__`, so `Stats.cache_metrics` converts to `dict[CacheMetricsKey, CacheMetrics]`. `IggyClient.get_stats` returns an awaitable resolving to `Stats`; byte sizes are exposed as integer bytes and times as microseconds, matching the existing getters. Stubs were regenerated with `cargo run --bin stub_gen` (purely additive diff after ruff).

The getter docstrings mirror the server semantics (per-shard CPU delta with a zero first sample, cpuset/cgroup scoping, whole-second run/start time, disk zero fallback, non-atomic snapshot), and `get_stats` documents its authentication and `read_servers`/`manage_servers` requirement.

`tests/test_stats.py` covers the counters moving after a stream/topic/messages are created, a single-node topology (2 streams x 2 topics x 3 partitions, consumer groups, a second client) on the way up and after cleanup, process identity fields staying stable across calls, memory/disk bounds, the empty `cache_metrics` map, key construction and hashing without a server, the auth gate (before connect, before login, after logout) and the `read_servers`/`manage_servers` permission gate. A Rust unit test in `stats.rs` pins the `None` branch of the semver conversion.

## Local Execution

- Passed
- Pre-commit hooks ran

Ran against a locally built `iggy-server` from this branch: `pytest tests/ --ignore=tests/test_tls.py` gives 509 passed (`test_tls` needs a Docker daemon, unavailable locally). `cargo fmt`, `cargo clippy --all-features --all-targets -- -D warnings`, `cargo test` (binding unit test), ruff check/format and pyrefly are clean; the stub is regenerated with `cargo run --bin stub_gen`.

## AI Usage

1. Claude Code (Fable 5).
2. Entire implementation and tests, following the pattern proposed in the issue.
3. Built the extension and ran the new and full Python test suites against a locally built server; regenerated and diffed the stubs; ran clippy/fmt/ruff/pyrefly.
4. Yes.


justinmclean:
Nice addition, follows the existing `user.rs` wrapper pattern and the stub is regenerated. Nothing blocking; a few things worth a look.

- `src/stats.rs:30` — `CacheMetricsKey` has no `#[new]`, so Python can't construct one. The dict ends up iterate-only: you can't ask for stream 1 / topic 2 / partition 0 directly. A constructor would fix that cheaply.
- `src/stats.rs:264` — `cache_metrics` is a `#[getter]`, so every access re-collects the whole map. Iterating the property and indexing it in the same loop is quadratic, which is what the test does at `tests/test_stats.py:66` and `:70`. Building it once in `From<RustStats>` would avoid the surprise.
- `tests/test_stats.py:66` — the loop body is the only thing `test_get_stats_cache_metrics_dict` asserts, and on a server with no cached partitions it never runs. Sending and polling messages first, then asserting the map is non-empty, would make it test the conversion.
- `tests/test_stats.py:70`-`:73` — `stats.cache_metrics[key] is not None` can't fail (a miss raises `KeyError`), and `>= 0` on `u32` / `u64` is always true.
- `tests/test_stats.py:47`-`:49` — exact equality on server-global counters. Fine today since the suite runs serially, but `pytest-xdist` is a declared dev dependency; `>=` avoids the trap.
- `src/stats.rs:255` — `1.2.3 -> 100200300` should be `1002003`. Inherited verbatim from `core/common/src/types/stats/mod.rs:73`, so not yours. Fixing it means the Rust doc comment plus a `stub_gen` regen, ideally in `core/common` too.

---

This review was drafted by an AI-assisted tool (Apache Magpie), so it may contain mistakes. If you think one of them is misapplied, please reply on the PR, and a maintainer will weigh in.

foreign/python/src/stats.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    /// The available memory of the system, in bytes.
+    #[getter]
+    pub fn available_memory(&self) -> u64 {
+        self.inner.available_memory.as_bytes_u64()
+    }
+
+    /// The run time of the server process, in microseconds.
+    #[getter]
+    pub fn run_time(&self) -> u64 {
```
ethanlin01x: `run_time` is an `IggyDuration`, and this SDK already exposes durations as `datetime.timedelta`. Returning raw microseconds here is inconsistent.

foreign/python/src/stats.rs
```diff
@@ -0,0 +1,289 @@
...
+    /// The version of the Iggy server.
+    #[getter]
+    pub fn iggy_server_version(&self) -> String {
+        self.inner.iggy_server_version.clone()
+    }
+
+    /// The numeric semantic version of the Iggy server, or `None` when unknown.
+    /// E.g. 1.2.3 -> 100200300 (major * 1000000 + minor * 1000 + patch).
+    #[getter]
+    #[gen_stub(override_return_type(type_repr = "int | None"))]
```
ethanlin01x: Nit: the other overrides write `builtins.int | None` (e.g. `config.rs:191`). Same meaning here, but this string is copied verbatim into the generated stub, so it stays as the only hand-written bare `int` across every future regen.

foreign/python/apache_iggy.pyi
```diff
@@ -1834,6 +1895,153 @@ class SendMessagesResponse:
         with an offset a client has already recorded.
         """
 
+@typing.final
+class Stats:
```
ethanlin01x, line 1904: `CacheMetrics` and `CacheMetricsKey` both define `__repr__`, but `Stats` does not. Since `get_stats` is a diagnostic call, `print(stats)` showing an object address is not great. A short repr with a few key fields would help.

ethanlin01x (commented):
Three small things, mostly about matching the SDK's existing conventions.

foreign/python/src/stats.rs
7487: Done in 7a28235 — `run_time` now returns `datetime.timedelta` via the shared `iggy_duration_to_py_delta` helper, and the stub is regenerated.

foreign/python/src/stats.rs
7487: Fixed in 7a28235 — the override now reads `builtins.int | None`, and the regenerated stub carries it.

foreign/python/apache_iggy.pyi
7487: Added in 7a28235 — `Stats.__repr__` now prints hostname, server version and the headline counters (streams/topics/partitions/messages/clients).

7487:
Thanks for the thorough pass — all addressed, point by point:

- **`CacheMetricsKey` constructor**: added `#[new]`, so `CacheMetricsKey(stream_id=1, topic_id=2, partition_id=3)` now works, and there is a new test that an independently constructed equal key addresses a dict entry (eq/hash round-trip).
- **`cache_metrics` re-collection**: the map is now converted exactly once in `From<RustStats>` and stored as a `Py<PyDict>`; the getter hands back the same dict on every access, so iterate-and-index is linear. The test asserts `stats.cache_metrics is stats.cache_metrics` to pin that down.
- **Non-empty assertion**: I could not make that work, and traced why — the server currently hardcodes an empty map in the `GetStats` reply (`core/server/src/responses.rs`: `cache_metrics: Vec::new()`), so the map stays empty no matter how much traffic precedes the call. A non-empty assertion would fail unconditionally. I left the per-entry type checks (with a comment explaining they only run once the server starts populating the field) and covered the key/dict semantics in the new server-free test instead.
- **Vacuous assertions** (`is not None`, `>= 0` on unsigned): dropped.
- **Exact equality on server-global counters**: switched to `>=` with a comment naming the pytest-xdist reason.
- **Semver docstring**: you're right — `get_numeric_version` formats `{major}{minor:03}{patch:03}`, so 1.2.3 -> 1002003. Fixed in the binding, in `core/common/src/types/stats/mod.rs`, and the stub is regenerated.

Also merged master to catch the branch up.

foreign/python/src/client.rs
```diff
@@ -157,6 +158,25 @@ impl IggyClient {
         })
     }
 
+    /// Get the statistics and details of the server and its running process.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `Stats`.
+    ///
+    /// Raises:
+    ///     RuntimeError: If the request fails.
```
slbotbm, line 167: The Rust `SystemClient::get_stats` contract requires authentication and the `read_servers` permission. Document both. Test before connect/login, after disconnect, without permission, and with `read_servers`/`manage_servers`.

foreign/python/src/stats.rs
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    /// The total memory of the system, in bytes.
+    #[getter]
+    pub fn total_memory(&self) -> u64 {
+        self.inner.total_memory.as_bytes_u64()
+    }
+
+    /// The available memory of the system, in bytes.
+    #[getter]
```
slbotbm, line 171: Document that CPU is a per-shard delta and the first sample can be zero. Also document that runtime and start time have whole-second precision.

slbotbm: Also, mirror the Rust docs: CPU is cpuset-scoped and memory is cgroup-scoped.

slbotbm: Also, document that these are not an atomic snapshot.

foreign/python/src/stats.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    /// Cache metrics per partition.
+    ///
+    /// Built once when the stats snapshot is created; every access returns the
+    /// same dict.
+    #[getter]
+    #[gen_stub(override_return_type(type_repr = "builtins.dict[CacheMetricsKey, CacheMetrics]"))]
+    pub fn cache_metrics(&self, py: Python<'_>) -> Py<PyDict> {
+        self.cache_metrics.clone_ref(py)
+    }
```
slbotbm, line 295: The server currently returns an empty map. Document that and assert `stats.cache_metrics == {}`; keep the key/hash test separate.

foreign/python/src/stats.rs
```diff
@@ -0,0 +1,328 @@
...
+    #[getter]
+    pub fn free_disk_space(&self) -> u64 {
+        self.inner.free_disk_space.as_bytes_u64()
+    }
+
+    /// The total disk space for the data directory, in bytes.
+    #[getter]
+    pub fn total_disk_space(&self) -> u64 {
+        self.inner.total_disk_space.as_bytes_u64()
+    }
```
slbotbm, line 313: Document that disk values can be zero when the data path or probe is unavailable. Also assert `free_disk_space <= total_disk_space` in tests.

foreign/python/tests/test_stats.py
slbotbm, line 1: Add a single-node topology test with multiple streams/topics/partitions, consumer groups, and a second client. Verify increases in `streams_count`, `topics_count`, `partitions_count`, `segments_count`, `consumer_groups_count`, and `clients_count`; after cleanup use `<=` deltas.

foreign/python/tests/test_stats.py
```diff
@@ -0,0 +1,103 @@
...
+
+        assert stats.iggy_server_version
+        assert stats.hostname
+        assert stats.process_id > 0
+        assert stats.start_time > 0
+        assert stats.total_memory > 0
+        assert stats.total_disk_space > 0
+
+        assert isinstance(stats.run_time, datetime.timedelta)
+        assert stats.run_time >= stats_before.run_time
```
slbotbm, line 65: Also assert that `messages_size_bytes` increases after sending messages.

slbotbm: In addition, ddd checks for `threads_count > 0`, `available_memory <= total_memory`, `free_disk_space <= total_disk_space`, and non-empty `os_name`, `os_version`, and `kernel_version`. A second call should keep static fields stable and runtime nondecreasing.

foreign/python/tests/test_stats.py
```diff
@@ -0,0 +1,103 @@
...
+        stats = await iggy_client.get_stats()
+
+        assert isinstance(stats.cache_metrics, dict)
+        # The getter must not re-collect the map on every access.
+        assert stats.cache_metrics is stats.cache_metrics
+        # The server does not populate cache metrics yet (`GetStats` replies
+        # with an empty map), so entries are only checked when present.
+        for key, metrics in stats.cache_metrics.items():
+            assert isinstance(key, CacheMetricsKey)
+            assert isinstance(metrics, CacheMetrics)
```
slbotbm, line 83: Assert the current empty server response here. Keep key equality and hashing in the separate no-server test.

foreign/python/src/stats.rs
```diff
@@ -0,0 +1,328 @@
...
+        self.inner.iggy_server_version.clone()
+    }
+
+    /// The numeric semantic version of the Iggy server, or `None` when unknown.
+    /// E.g. 1.2.3 -> 1002003 (major * 1000000 + minor * 1000 + patch).
+    #[getter]
+    #[gen_stub(override_return_type(type_repr = "builtins.int | None"))]
+    pub fn iggy_server_semver(&self) -> Option<u32> {
+        self.inner.iggy_server_semver
+    }
```
slbotbm, line 285: Add a binding unit test for the `None` branch when a `RustStats` conversion can be constructed directly.

slbotbm (dismissed):

foreign/python/src/client.rs
7487: Done in b76b445. The docstring now states the authenticated-session and `read_servers`/`manage_servers` requirement (`Permissioner::get_server_info` accepts either flag). Tests: `test_get_stats_requires_connection_and_auth` covers before connect, connected-but-not-logged-in, and after `logout_user`; `test_get_stats_requires_server_permission` is parametrized over `read_servers` and `manage_servers` (denied with neither, granted with each). The SDK has no `disconnect`, so logout stands in for the post-disconnect case.

foreign/python/src/stats.rs
7487: Done in b76b445. `cpu_usage` and `total_cpu_usage` now say they are a delta since the previous `get_stats` served by the same shard, so a shard's first sample is 0 (the thread-local `SYSINFO` in `core/server/src/responses.rs`). `total_cpu_usage` carries the cpuset note and `total_memory`/`available_memory` the cgroup note from `core/common`'s `Stats`; `run_time`/`start_time` say whole-second precision; the `Stats` class doc says the fields are not an atomic snapshot. All of it is in the regenerated stub.

foreign/python/src/stats.rs
7487: Done in b76b445: the getter doc says the server does not populate it yet and always replies with an empty map, `test_get_stats_cache_metrics_dict` asserts `stats.cache_metrics == {}`, and the key construction/hash checks stay in the separate server-free test.

foreign/python/src/stats.rs
7487: Done in b76b445: both disk getters document the 0 fallback (no data path captured or probe error, `STATS_DATA_PATH.get().map_or((0, 0), ..)`), and `test_get_stats` asserts `free_disk_space <= total_disk_space`.

foreign/python/tests/test_stats.py
7487: Added `test_get_stats_reflects_topology` in b76b445: 2 streams x 2 topics x 3 partitions, one consumer group per topic, and a second logged-in client; it asserts `>=` increases for streams, topics, partitions, segments (one per new partition), consumer groups and clients. Cleanup deletes the topics and drops the second client, then asserts `<=` deltas for topics, partitions, segments and consumer groups and a non-increasing `clients_count` (the server reaps a dropped connection asynchronously). The SDK has no `delete_stream`, so `streams_count` is only checked on the way up.

foreign/python/tests/test_stats.py
7487: Done in b76b445: `messages_size_bytes` must grow after the send. Also added `available_memory <= total_memory`, `free_disk_space <= total_disk_space`, non-empty `os_name`/`os_version`/`kernel_version`, and the process identity fields (`process_id`, `start_time`, host/OS/kernel, version, semver) equal across the two calls with `run_time` non-decreasing. `threads_count > 0` is asserted unless the server reports `os_name == "Darwin"`: sysinfo's `Process::tasks()` is `None` on macOS, so `SystemProbe` sends 0 there (a server built from this branch reports 0 on my Mac; Linux CI gets the strict check).

foreign/python/tests/test_stats.py
7487: Done in b76b445: the loop is replaced by `assert stats.cache_metrics == {}`; key equality and hashing stay in `test_cache_metrics_key_is_constructible_and_hashable`.

foreign/python/src/stats.rs
7487: Added in b76b445: `stats::tests::iggy_server_semver_none_survives_conversion` builds a `RustStats` with `iggy_server_semver: None` (and one with `Some(1_002_003)`), converts through `Stats::from` and checks the getter. `cargo test` in `foreign/python` runs it.

7487:
Thanks for the second pass. Everything is addressed inline in b76b445 (docstrings mirror the server semantics, stub regenerated, auth/permission/topology/bounds tests added, Rust unit test for the `None` semver branch), and master is merged again.

Two things worth knowing while reading the tests: the SDK has no `disconnect` or `delete_stream`, so `logout_user` stands in for the post-disconnect case and topology cleanup deletes topics rather than streams; and `threads_count` is 0 when the server runs on macOS (sysinfo cannot enumerate threads there), so that assertion is skipped when the server reports `os_name == "Darwin"`.

Local run against a server built from this branch: `pytest tests/ --ignore=tests/test_tls.py` 509 passed; fmt, clippy (`-D warnings`), `cargo test`, ruff and pyrefly clean.

spetz (approved):

hubcio (approved):

numinnex (approved):

PR #4022, opened 2026-09-01
feat(cpp): add stream, topic, and partition functions to high level client

slbotbm:

Closes #
Relates to #2100 

## Rationale

Adds function to the high level cpp client.

## What changed?

Adds stream, topic, and partition management to the high-level C++ client
(`IggyBlockingClient`):

- Streams: `CreateStream` / `UpdateStream` / `GetStreams` / `GetStream` / `PurgeStream` / `DeleteStream`
- Topics: `CreateTopic` / `UpdateTopic` / `GetTopics` / `GetTopic` / `PurgeTopic` / `DeleteTopic`
- Partitions: `CreatePartitions` / `DeletePartitions`

New supporting types in `iggy.hpp`: `IggyException`, `LoginInfo`, `Identifier`,
`HeaderKind` / `HeaderField` / `HeaderEntry`, `ResourceOptions`, and the
`Topic` / `Partition` / `TopicDetails` / `StreamDetails` / `Stream` entities;
option builders `TopicCreateOptions`, `TopicUpdateOptions`, `StreamUpdateOptions`.
`UpdateStream` now takes `StreamUpdateOptions`; `CreateTopic` / `UpdateTopic`
now take option builders instead of string args. `PollingStrategy` returns a
structured `Kind()` / `Value()` object instead of two strings.

Rust FFI (`lib.rs`, `client.rs`, `type_conversion.rs`) gains
`ffi::TopicCreateOptions` / `ffi::TopicUpdateOptions` structs (with `has_*`
flags), updates `create_topic` / `update_topic` / `update_stream` signatures,
and adds `TryFrom` conversions. New `type_conversions.cpp` adds `FromFfi` /
`ToFfi` helpers for the new types.

Tests: `LowLevelE2E_Stream` -> `E2E_Stream` / `LowLevelE2E_Topic` -> `E2E_Topic` / `LowLevelE2E_Partition` -> `E2E_Partition` suites; `client`,
`consumer_group`, and `message` suites converted to the high-level client; new
unit tests for the option builders; BDD messaging steps switched to
`iggy::ffi::TopicCreateOptions`; 

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? codex and muse spark 1.2
3. How did you verify the generated code works correctly? read through it. tested it.
4. Can you explain every line of the code if asked? yes.


slbotbm:
A note to the reviewers: since this is a pretty big PR, I would suggest ignoring that changes in `foreign/cpp/tests/*` since those are mostly mechanical translations of the tests from the low level API to the high level API, and it is still not yet complete.

amlel-el-mahrouss (commented):
Reviewed the C++ part and it looks correct, additions of `[[nodiscard]]`, is a good evolution.
One thing I'll note: It would be nice in the next PR to write a set of tests for `type_conversions.cpp`.

amlel-el-mahrouss (commented):
One comment I'd make would be to turn the vector into a static const vector variable for the sake of the unit test.

foreign/cpp/src/type_conversion.rs
```diff
@@ -525,6 +528,167 @@ impl From<RustOptionSpec> for ffi::OptionSpec {
...
+                    format!(
+                        "invalid max topic size '{}': {error}",
+                        options.max_topic_size
+                    )
+                })?,
+            )
+        } else {
+            None
+        };
+        let durability = if options.has_durability {
```
hubcio, line 579: warning: typed durability does not win over raw entries, although the docs say it does. `SetDurability(Replicated)` plus a raw `durability=persisted` creates a persisted topic, so erase the matching key from `raw` when `has_durability` or `has_consumer_offset_durability` is set.

foreign/cpp/include/iggy.hpp
```diff
@@ -98,186 +876,613 @@ class SnapshotCompression final : private detail::StringTag<SnapshotCompression>
...
+     *        topic deletion.
+     */
+    static Expiry NeverExpire() { return Expiry("never_expire", std::numeric_limits<std::uint64_t>::max()); }
+
+    /**
+     * @brief Creates a time-based expiry policy.
+     * @param micros Message lifetime in microseconds.
+     * @return Time-based expiry policy.
+     */
+    static Expiry Duration(std::uint64_t micros) { return Expiry("duration", micros); }
```
hubcio: warning: `Expiry::Duration(0)` is sent as 0, which means server default, so a new topic never expires and an update keeps the old expiry. throw for `micros == 0` here and reject `duration` with value 0 in `type_conversion.rs`.

foreign/cpp/include/iggy.hpp
```diff
@@ -366,54 +1571,32 @@ inline rust::Vec<std::uint8_t> to_key_bytes(const std::string_view key) {
...
  *
  * Each factory encodes one key from the server's topic option catalog using
  * that key's required value kind. The server rejects unknown keys and values
  * encoded with a different kind.
  *
- * Use `Client::describe_options("topic")` to retrieve the supported keys,
- * value kinds, and defaults. Binary transport errors do not identify a rejected
- * key, so the catalog is the discovery mechanism for those transports.
+ * For example, pass
+ * `ResourceOptions::Explicit({TopicOption::SegmentSize(1024 * 1024)})` to
```
hubcio: warning: `TopicOption` is dead public API - its factories return `iggy::HeaderEntry`, which no client method accepts, and this example does not compile because `ResourceOptions::Explicit()` takes no arguments. delete lines 1542-1680 together with `TopicOptionTest` and its helpers in `unit_tests.cpp`.

foreign/cpp/tests/unit/unit_tests.cpp
```diff
@@ -187,18 +366,6 @@ TEST(IggyExceptionTest, StoresMessage) {
     EXPECT_EQ(std::string(from_string.what()), message);
 }
```
hubcio, line 339: warning: `fefe92a77` deleted `MovedFromOperationsThrow` without a replacement, so nothing tests the `Handle()` guard that every new method relies on. restore it and cover `CreateStream`, `GetStream`, `CreateTopic` and `CreatePartitions` too.

foreign/cpp/tests/e2e/topic.cpp
```diff
@@ -158,171 +155,175 @@ TEST_F(LowLevelE2E_Topic, CreateTopicWithInvalidNamesThrows) {
...
-        has_header(topic_details.options, key_kind, to_payload("preallocate_segments"), bool_kind, bool_bytes(false)));
-
-    for (const auto &derived_option : topic_details.derived_options) {
-        const std::string derived_key(derived_option.key.value.begin(), derived_option.key.value.end());
-        EXPECT_NE(derived_key, "segment_size") << "segment_size was set explicitly, so it cannot be derived";
-    }
+        client.GetTopic(iggy::Identifier::String(stream_name), iggy::Identifier::String(topic_name));
+
+    const auto &explicit_options = topic_details.Options().Explicit();
+    EXPECT_EQ(explicit_options.count("segment_size"), 1u);
```
hubcio, line 305: warning: the low-level version of this test checked every option value, but this one only checks keys and kinds. restore the value checks for `segment_size`, `messages_required_to_save`, `size_of_messages_required_to_save` and `preallocate_segments`, because they now go through the new conversion code.

foreign/cpp/MODULE.bazel.lock
```diff
@@ -483,109 +483,6 @@
           }
         }
       }
```
hubcio, line 485: nit: this drops the rules_rust `cu_nr` entry that master still has, with no `MODULE.bazel` or bazel version change - intentional? it came in with `fefe92a77`, which also deleted `MovedFromOperationsThrow`, so check what else that rebase dropped.

foreign/cpp/src/client.cpp
```diff
@@ -55,13 +55,251 @@ void IggyBlockingClient::Shutdown() {
...
+
+void IggyBlockingClient::PurgeStream(const Identifier &stream) {
+    RethrowAsIggyException([this, &stream] { Handle()->purge_stream(stream.ToFfi()); });
+}
+
+TopicDetails IggyBlockingClient::CreateTopic(const Identifier &stream,
+                                             std::string name,
+                                             const TopicCreateOptions &options) {
+    return RethrowAsIggyException([this, &stream, &name, &options] {
+        ffi::TopicCreateOptions ffi_options;
```
hubcio: simplification: the raw-entry loop is copied three times (lines 71, 193, 242) and the compression/expiry/size block twice (lines 128, 219). move them into helpers and drop the else branches, since cxx already initializes these fields to zero or empty.

foreign/cpp/src/type_conversion.rs
```diff
@@ -525,6 +528,167 @@ impl From<RustOptionSpec> for ffi::OptionSpec {
     }
 }
 
+impl TryFrom<ffi::TopicCreateOptions> for RustTopicCreateOptions {
+    type Error = String;
+
+    fn try_from(options: ffi::TopicCreateOptions) -> Result<Self, Self::Error> {
+        use std::str::FromStr;
```
hubcio: simplification: the compression/expiry/max-size parsing below is repeated in the update impl (lines 639-681), along with this inline `use std::str::FromStr;`. share one helper per field and move the import to the top of the file.

foreign/cpp/include/iggy.hpp
```diff
@@ -98,186 +876,613 @@ class SnapshotCompression final : private detail::StringTag<SnapshotCompression>
...
+ * that do not yet have a typed setter. When both specify the same option, the
+ * typed setting takes precedence.
+ */
+class TopicCreateOptions final {
+  public:
+    TopicCreateOptions() = default;
+
+    /**
+     * @brief Returns the number of partitions to create.
+     * @return Configured partition count, or `std::nullopt` to use the server
```
hubcio: nit: an unset partition count is not a server default - the client fills in `DEFAULT_PARTITIONS_COUNT`, which is 1. say "defaults to 1".

foreign/cpp/src/type_conversions.cpp
```diff
@@ -0,0 +1,139 @@
...
+    std::optional<std::uint64_t> access_token_expiry;
+    if (login_info.has_access_token) {
+        access_token.emplace(login_info.access_token.c_str(), login_info.access_token.size());
+        access_token_expiry = login_info.access_token_expiry;
+    }
+
+    return LoginInfo(login_info.user_id, std::move(access_token), access_token_expiry);
+}
+
+Identifier Identifier::FromFfi(ffi::Identifier identifier) {
```
hubcio: simplification: `Identifier::FromFfi` has no caller. delete it and its declaration in `iggy.hpp`.

hubcio (changes requested):
the invalid-identifier tests in `partition.cpp` pass only because `Identifier::String` throws before the partition call runs. the throw checks in `topic.cpp`, `stream.cpp` and `partition.cpp` expect `std::exception` where these methods document `IggyException` (the low-level `describe_options` call can stay as is). `UpdateTopicWithAllOptionCombinationsUpdatesInputFields` only checks that nothing throws - read the topic back and compare the values.

foreign/cpp/src/type_conversions.cpp
slbotbm: done

foreign/cpp/include/iggy.hpp

foreign/cpp/src/type_conversion.rs
slbotbm: done

foreign/cpp/src/client.cpp
slbotbm: done

foreign/cpp/MODULE.bazel.lock
slbotbm: looks like it was dropped by mistake. restored.

foreign/cpp/tests/e2e/topic.cpp
slbotbm: done

foreign/cpp/tests/unit/unit_tests.cpp
slbotbm: done

foreign/cpp/include/iggy.hpp

foreign/cpp/include/iggy.hpp

foreign/cpp/src/type_conversion.rs
slbotbm: done

hubcio (approved):

mmodzelewski (approved):

PR #4024, opened 2026-09-01
fix(cluster): serve metadata reads at or above the client's own writes

numinnex:
A client that commits a metadata write and then re-homes its session onto a backup can be served the pre-write state. auth.rs already documents the gap: the session epoch is the register's commit op, and on a backup that forwarded the proposal the local applied commit still lags it. A backup applies committed ops only as its commit walk advances, and nothing tied a read to the op the client's own write committed.

Metadata reads now gate on the connection's committed watermark. One applied-frontier counter per process advances after every metadata apply and is shared with every shard, which for the first time gives a shard without consensus the applied position it had no way to observe. The watermark comes from the commit field replies already carry, seeded at bind from the session epoch. The fast path is a single atomic load with no awaits, so the shared-nothing read path is unchanged; a lagging node parks briefly, then fails the read retryable rather than answering stale.
 
Over HTTP this closes the forwarded-register case, where a healthy backup forwards the register so the bound epoch can exceed the local frontier. It does not close the case where forwarding is active: the follower relays the write, its handler never runs, so the node that later serves the read holds no session and no watermark. Closing that needs the serving primary's commit op to travel back to the reading node, for instance a response header beside the view the forward middleware already relays. That is additive but touches every control-plane write response, so it is left out here and documented at the gate.

The BDD delete-then-get steps now assert "not the stream we deleted" instead of "nothing at this id". The server hands a deleted stream's numeric id straight to the next create, so once scenarios share a server the old assertion cannot hold, and removing the polling loop without this would have left the spec flaky for an unrelated reason.


core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+/// The two exclusions only look like metadata reads: `DescribeOptions` decodes
+/// a static catalog, and `GetClusterMetadata` answers from the configured
+/// roster plus the consensus view. Holding either buys no consistency, and the
+/// roster read is on the SDK's leader-discovery path, where the wait would be
+/// real.
+///
+/// Shared with the HTTP read path, which gates the identical set of command
+/// codes through `build_non_replicated_response`: two lists would drift, and a
+/// code dropped from one plane's list is a silent stale read on that plane.
+pub const fn read_needs_metadata_frontier(code: u32) -> bool {
```
hubcio, line 184: warning: this predicate is only consulted in the `_` arm, so `GET_ME`, `GET_CLIENTS`, `GET_CLIENT` and `SYNC_CONSUMER_GROUP` read the metadata state machine ungated. gate those arms individually - hoisting above the match would park `PING` and `POLL_MESSAGES`, since this is a deny-list.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+        shard.bus.sleep(READ_FRONTIER_POLL).await;
+        if metadata.applied_frontier() >= watermark {
+            return Ok(());
+        }
+    }
+    warn!(
+        frontier = metadata.applied_frontier(),
+        watermark, "metadata read frontier unreached past deadline; failing the read retryable"
+    );
+    Err(IggyError::TransientNotCommitted)
```
hubcio: warning: every SDK replays `TransientNotCommitted` on the same connection and never fails over, so a durably lagging replica burns the client's full 30s read timeout and then errors. use `TransientNotAccepted` - a read never entered a pipeline, and it's the code that drives the roster walk.

core/server/src/dispatch/reads.rs
```diff
@@ -254,6 +339,14 @@ pub(in crate::dispatch) async fn handle_non_replicated_request<B, MJ, S, SB>(
             handle_sync_consumer_group(shard, transport_client_id, &request).await;
         }
         _ => {
+            if read_needs_metadata_frontier(code)
```
hubcio: warning: this parks before authorization and before decode, so an unauthorized read or an unknown command code holds the connection for the full budget before its terminal error. move the gate after `authorize_default_read` - the HTTP side already does this and says why.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
...
 
+/// Poll cadence while a read waits for this node's applied metadata frontier.
+/// The consensus tick, so a node one commit behind resumes on the next commit
+/// broadcast rather than a tick later.
+const READ_FRONTIER_POLL: Duration = Duration::from_millis(10);
+
+/// Polls a held read is given before it fails retryable: 3s at the cadence
+/// above. Long enough to ride out a view change, far below the SDK's 30s
+/// request budget, inside which it replays the same id on the same connection.
+const READ_FRONTIER_MAX_POLLS: u32 = 300;
```
hubcio: warning: 3s can't ride out a view change - detection needs `heartbeat_timeout` (5s) and escalation another `view_change_status_timeout` (5s), and `recovery_barrier_deadline` budgets at least 15s for the same event. derive the deadline from those timeouts, or restate this as ~6 commit broadcast intervals.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+    SB: SuperblockStore + 'static,
+{
+    // Own statement: the borrow has to be released before the poll below, which
+    // awaits on the same task the session manager's mutators run on.
+    let watermark = sessions.borrow().metadata_watermark(transport_client_id);
+    let metadata = shard.plane.metadata();
+    if metadata.applied_frontier() >= watermark {
+        return Ok(());
+    }
+    for _ in 0..READ_FRONTIER_MAX_POLLS {
```
hubcio: warning: this park sits inside the per-socket drain loop, which awaits one frame at a time, so a held metadata read also blocks that connection's queued `SendMessages`, `PollMessages` and `PING`. the doc above says only this connection waits - it doesn't say the data plane waits with it.

core/server/src/http/reads.rs
```diff
@@ -94,7 +97,15 @@ pub(in crate::http) async fn read_local(
     rule: impl FnOnce(&Permissioner, u32) -> Result<(), IggyError>,
 ) -> Result<Bytes, ReadError> {
     await_recovery_barrier(&state.shard).await?;
+    // Ahead of the frontier wait on purpose. `authorize_read` renders the
+    // linearizable follower redirect, which must answer 307 immediately -
+    // parking first would delay a request this node is not going to serve at
+    // all - and an authorization denial is terminal, so holding the connection
+    // for it buys nothing.
     authorize_read(state, identity, consistency, rule)?;
```
hubcio: warning: the scope was resolved off the pre-wait state machine but the read re-resolves after it. an entity created during the wait is served with no permissioner call at all, since a scope miss maps to `Ok(())`. re-resolve the scope after the wait.

core/server/src/http/extractor.rs
```diff
@@ -94,6 +94,14 @@ impl FromRequestParts<HttpState> for Authenticated {
...
 pub struct Identity {
     pub user_id: u32,
+    /// Session-table key of the presenting credential (`jwt:{jti}` /
+    /// `pat:{sha}`), the SAME key [`Authenticated`] resolves its session under.
+    ///
+    /// A read mints no session, so this is the only join back to what this
+    /// credential's writes committed: the read gate looks its metadata
+    /// watermark up by this key. Carried rather than resolved again in the
+    /// handler because [`resolve_credential`] already computed it.
+    pub session_key: String,
```
hubcio: warning: `POST /users/refresh-token` mints a fresh `jti` without carrying the watermark or registering a session, so reads on the refreshed bearer hit an unknown key, get 0, and the gate no-ops. key the watermark per user id, or carry it in `issue_identity`.

core/server/src/http/state.rs
```diff
@@ -229,6 +229,25 @@ impl HttpInner {
...
+    ///
+    /// Deliberately NOT expiry-filtered, unlike [`Self::live_session`]: the
+    /// number is a consistency floor, not a capability, and the request that
+    /// consults it has already re-verified the bearer. Dropping the floor
+    /// because a swept-but-still-present entry aged out would reintroduce the
+    /// stale read for exactly the callers still holding a committed reply.
+    ///
+    /// Confines the shared `RefCell` borrow to this call, so it can never span
+    /// the read gate's `.await`.
+    pub(in crate::http) fn metadata_watermark(&self, key: &str) -> u64 {
```
hubcio: warning: `forget_session` removes the entry outright on eviction, so this returns 0 again while the bearer stays valid. a caller that had written then reads ungated until its next write re-registers - keep a key-scoped floor outside the session entry.

core/server/src/http/submit.rs
```diff
@@ -113,6 +113,16 @@ pub(in crate::http) async fn submit_committed(
...
             submit_gated(&shard, &task_session, operation, max_tokens_per_user, &body).await;
+        // Recorded here rather than after the await below, for the same reason
+        // the submit is detached: a caller that disconnected mid-write still
+        // committed the op, and its next request on this credential must not be
+        // served state older than what committed. Ordered before the wake, so a
+        // read issued the instant the response lands already sees the mark.
+        if let Ok((_, reply, _)) = &result
+            && let Some(commit) = committed_reply_commit(reply)
+        {
+            task_session.record_metadata_watermark(commit);
```
hubcio: warning: on a follower with HTTP forwarding on, control-plane writes are relayed before the handler runs, so this never fires there and the watermark stays frozen at the register epoch. the HTTP read-your-writes guarantee doesn't hold on a forwarding follower.

core/simulator/src/lib.rs
```diff
@@ -5097,3 +5105,370 @@ mod repair_frontier_tests {
...
+    /// answers a metadata read from an unconverged state answers within a
+    /// couple of these; the gate must hold the read past all of them.
+    ///
+    /// Well under the gate's own poll budget, so expiry cannot masquerade as a
+    /// held read.
+    const STALE_WINDOW_STEPS: u32 = 50;
+
+    /// Steps allowed for repair to reach the backup and the held read to answer
+    /// once replication is restored.
+    const CONVERGE_STEPS: u32 = 2_000;
```
hubcio: warning: `CONVERGE_STEPS` is unreachable. one `sim.step()` is 10ms, which is one gate poll, so phase 1 burns 50 of the gate's 300 and the read fails at poll 300 - the status assert trips first and this panic never runs. derive it as `READ_FRONTIER_MAX_POLLS - STALE_WINDOW_STEPS`.

core/server/src/dispatch/submit.rs
```diff
@@ -192,3 +195,27 @@ where
...
     rx.recv().await.ok().flatten()
 }
+
+/// The commit position a COMMITTED metadata reply carries, or `None` when the
+/// frame promises the client nothing.
+///
+/// Three frames arrive on this path and only one is a promise. An eviction is
+/// an `EvictionHeader` whose bytes would cast cleanly as a reply, so the
+/// command is checked first (same guard as `build_raw_pat_reply`). A
+/// pre-consensus rejection stamps the primary's `commit_max`, an op the caller
```
hubcio: nit: the doc says pre-consensus rejections are excluded, but the filter only matches codes 57 and 58, and `ReplyHeader.status` - the documented pre-commit denial channel - is never read. add `status != 0`, stamp `commit: 0` on the terminal rejection arms, and fix the doc.

core/server/src/dispatch/submit.rs
```diff
@@ -192,3 +195,27 @@ where
...
+/// expiry) DID commit and counts.
+///
+/// Shared with the HTTP write path, which grades the same three frames off the
+/// same submit entry point ([`submit_client_request_on_owner`]); one classifier
+/// is what keeps the two planes' watermarks meaning the same thing.
+pub fn committed_reply_commit(reply: &Message<GenericHeader>) -> Option<u64> {
+    if reply.header().command != Command::Reply || transient_code(reply).is_some() {
+        return None;
+    }
+    let header = reply.as_slice().get(..size_of::<ReplyHeader>())?;
```
hubcio: nit: two silent drops here - the slice `?` swallows a short frame and `.ok()` swallows a bad cast, neither logs. `warn!` on both.

core/server/src/dispatch/submit.rs
```diff
@@ -192,3 +195,27 @@ where
...
+/// command is checked first (same guard as `build_raw_pat_reply`). A
+/// pre-consensus rejection stamps the primary's `commit_max`, an op the caller
+/// was never told committed and, on a backup-homed caller, one the read gate
+/// would then wait for. A committed business rejection (duplicate name, bad
+/// expiry) DID commit and counts.
+///
+/// Shared with the HTTP write path, which grades the same three frames off the
+/// same submit entry point ([`submit_client_request_on_owner`]); one classifier
+/// is what keeps the two planes' watermarks meaning the same thing.
+pub fn committed_reply_commit(reply: &Message<GenericHeader>) -> Option<u64> {
```
hubcio, line 224: nit: `committed_reply_commit` has no unit test, and the binary gate's expiry arm has never run in one. a table over committed reply, transient rejection, terminal rejection and eviction would pin it.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
     .await;
 }
 
+/// Poll cadence while a read waits for this node's applied metadata frontier.
+/// The consensus tick, so a node one commit behind resumes on the next commit
+/// broadcast rather than a tick later.
+const READ_FRONTIER_POLL: Duration = Duration::from_millis(10);
```
hubcio: nit: the doc calls this the consensus tick then redefines 10ms by hand; `consensus::TICK_INTERVAL` is public and already reachable from here. the same pair is declared again in `http/reads.rs`, and a third 10ms lives in its recovery barrier.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+where
+    B: ShellBus,
+    MJ: JournalHandle + 'static,
+    MJ::Target: Journal<Entry = Message<PrepareHeader>, Header = PrepareHeader>,
+    S: 'static,
+    SB: SuperblockStore + 'static,
+{
+    // Own statement: the borrow has to be released before the poll below, which
+    // awaits on the same task the session manager's mutators run on.
+    let watermark = sessions.borrow().metadata_watermark(transport_client_id);
```
hubcio: nit: `read_context` already probed this connection a few lines down, so this is a second borrow and hash lookup. it has exactly one caller - widen its tuple and the borrow-scoping comment above goes away too.

core/server/src/http/reads.rs
```diff
@@ -117,6 +128,87 @@ pub(in crate::http) async fn read_local(
...
+    max_polls: u32,
+) -> Result<(), ReadError> {
+    if frontier() >= watermark {
+        return Ok(());
+    }
+    for _ in 0..max_polls {
+        // `compio::time::sleep` like the recovery barrier below: this listener
+        // is pinned to shard 0's compio thread, which has no blocking pool to
+        // hand a wait to. Only this request parks.
+        compio::time::sleep(READ_FRONTIER_POLL).await;
```
hubcio: nit: reads poll uncapped here while writes are capped. compio coalesces these into one driver timeout so the cost is small, but an event-driven wait off `fire_commit_notifier` would be the real fix.

core/server/src/http/reads.rs
```diff
@@ -117,6 +128,87 @@ pub(in crate::http) async fn read_local(
...
+/// (307 to the primary), which [`authorize_read`] decides before this wait and
+/// which this wait never sees. This gate makes an UNQUALIFIED read
+/// read-your-writes for its own credential, at no redirect and no consensus
+/// round trip.
+///
+/// Scope is this node's own view. A credential whose write this node relayed
+/// over HTTP, or that wrote through a different node entirely, left no
+/// watermark here; closing that needs the serving primary's commit op to reach
+/// the reading node, which nothing in the response carries today.
+async fn await_metadata_read_frontier(
```
hubcio, line 201: nit: no end-to-end test of the HTTP gate at all - seed, record, hold, serve. the binary plane got two simulator tests; `hold_for_frontier` is only exercised against a closure.

core/server/src/http/reads.rs
```diff
@@ -290,7 +382,117 @@ pub(in crate::http) fn authorize_data_plane(
...
+            GET_PERSONAL_ACCESS_TOKENS_CODE,
+            GET_STATS_CODE,
+        ] {
+            assert!(
+                read_needs_metadata_frontier(code),
+                "code {code} answers from the metadata STM and must be gated"
+            );
+        }
+        for code in [DESCRIBE_OPTIONS_CODE, GET_CLUSTER_METADATA_CODE] {
+            assert!(
```
hubcio: nit: `/cluster/metadata` is a local route with its own handler and never calls `read_local`, so this pins a code the HTTP plane can't produce. drop it, or say it's asserted for the shared predicate.

core/server/src/http/session.rs
```diff
@@ -269,6 +297,7 @@ mod tests {
             data_gate: Mutex::new(FIRST_REQUEST_ID),
             registry_token: Cell::new(None),
             in_flight_writes: Cell::new(0),
+            metadata_watermark: Cell::new(1),
```
hubcio: nit: this fixture spells the epoch as a literal while `fake_session` below uses `FIXTURE_EPOCH`. it's field-for-field `fake_session("jwt:test", 7, u64::MAX)` - delete it and call the helper.

core/metadata/src/impls/metadata.rs
```diff
@@ -808,11 +825,41 @@ where
...
             transfer_offer_cache: RefCell::new(None),
+            applied_frontier: Arc::new(AtomicU64::new(0)),
         }
     }
 }
 
 impl<C, J, S, M, SB> IggyMetadata<C, J, S, M, SB> {
+    /// Share one process-wide applied frontier with every other shard.
+    ///
+    /// Consumed at construction rather than swapped in later: a shard that
```
hubcio, line 843: nit: `applied_frontier` collides with `InstallOutcome::applied_frontier`, and the two diverge - install advances the cell to `snapshot_seq` but reports `snapshot_seq.max(local_applied)`. rename one.

bdd/java/src/test/java/org/apache/iggy/bdd/BasicMessagingSteps.java
```diff
@@ -142,8 +142,14 @@ public void deleteStreamByNumericId() {
...
     @Then("getting the stream by its numeric ID should return no stream")
     public void getStreamReturnsNoStream() {
+        // The assertion is "not the stream we deleted", not "nothing at this id":
+        // the server hands out the lowest free stream id, so once these scenarios
+        // run concurrently against one server a fresh create can legitimately
+        // occupy the deleted stream's id.
         Optional<StreamDetails> stream = getClient().streams().getStream(context.lastStreamId);
-        assertTrue(stream.isEmpty(), "Deleted stream should not be returned");
+        assertTrue(
+                stream.isEmpty() || !stream.get().name().equals(context.lastStreamName),
```
hubcio, line 153: nit: `!name.equals(lastStreamName)` is vacuously true when `lastStreamName` is null. not reachable in this feature today, but the rust twin fails loud on a missing pre-value - add `assertNotNull`.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +123,83 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+///
+/// Fast path is a single `Acquire` load and no await, which is what keeps an
+/// uncontended read shared-nothing. Otherwise it polls the bus timer (virtual
+/// under the simulator, wall clock in production) from inside the
+/// per-connection drain task, so only this connection waits. Expiry fails
+/// loud and retryable rather than serving state the client already saw
+/// replaced; the log carries both numbers so a frontier that stopped moving
+/// is visible instead of showing up as a hang.
+#[allow(clippy::future_not_send)]
+async fn await_metadata_read_frontier<B, MJ, S, SB>(
```
hubcio, line 275: simplification: this loop is a second copy of `hold_for_frontier` in `http/reads.rs`, differing only in sleep source and error type. make that one generic over the sleep and call it from both - about 17 lines and one drift vector.

core/server/src/dispatch/submit.rs
```diff
@@ -29,9 +29,12 @@ use crate::dispatch::session_ops::{
     submit_register_local_or_forward,
 };
 use crate::dispatch::upgrade_shard_handle;
+use crate::http::reply::transient_code;
```
hubcio: simplification: this closes a cycle - `http/submit.rs` imports `committed_reply_commit` from here, which calls back into `http::reply`. move `transient_code` to `crate::responses`, which both planes already import and which already open-codes `reply_body`.

core/simulator/src/client.rs
```diff
@@ -648,6 +649,48 @@ impl SimClient {
...
+    /// Panics if `name` is not a valid wire name or the request buffer is
+    /// invalid.
+    #[allow(clippy::cast_possible_truncation)]
+    pub fn get_stream(&self, name: &str) -> Message<RoutedRequestHeader> {
+        let body = GetStreamRequest {
+            stream_id: WireIdentifier::named(name).expect("stream name must be valid"),
+        }
+        .to_bytes();
+
+        let header_size = std::mem::size_of::<RoutedRequestHeader>();
```
hubcio: simplification: this block and `poll_messages` above are 21 identical lines apart from code, group and panic text. one `non_replicated_request(code, group, body)` helper - it must not delegate to `header()`, which advances the request counter.

core/server/src/boot/mod.rs
```diff
@@ -580,7 +587,15 @@ async fn shard_main(
         superblock_for_metadata,
         mux_stm,
         Some(PathBuf::from(&config.system.path)),
-    );
+    )
+    .with_applied_frontier(metadata_applied_frontier);
+    // Recovery already replayed the committed WAL prefix into the state
+    // machine, so the frontier resumes where the commit walk will rather than
+    // at zero -- otherwise every read on a rebooted node parks until its
+    // deadline. No-op on peer shards, which share shard 0's cell.
```
hubcio: simplification: this seed is copy-pasted verbatim in `simulator/src/replica.rs`. fold it into one `seed_applied_frontier_from_consensus()` on the `VsrConsensus`-specialized impl, since `C` carries no bound on the generic block.

hubcio (changes requested):
two things I couldn't anchor to the diff:

- `responses.rs:241` is the shared `consumer_group_memberships` read behind `GET_ME`, `GET_CLIENTS` and `GET_CLIENT`, and the HTTP `get_clients` / `get_client` handlers reach it without `read_local` either. Same gap as the binary arms, on both planes.
- `dispatch/submit.rs:217` and `responses.rs:1569` are the same `Command::Reply` guard plus `try_from_bytes::<ReplyHeader>` decode, and both fire on the same message in `dispatch/mod.rs`. One `reply_header()` helper in `responses.rs` covers both.

One thing I looked at and decided against: merging `hold_for_frontier` with `await_recovery_barrier`. They're bounded differently (poll count vs wall-clock deadline), and the merge would delete `barrier_state`'s tests and turn the `max_polls: 8` injection into a flaky wall-clock test. Worth taking only the one-line crumb - drop the local `POLL` const and reuse `READ_FRONTIER_POLL`.

core/server/src/http/submit.rs
```diff
@@ -106,13 +105,32 @@ pub(in crate::http) async fn submit_committed(
...
     compio::runtime::spawn(async move {
         let result =
             submit_gated(&shard, &task_session, operation, max_tokens_per_user, &body).await;
+        // Recorded here rather than after the await below, for the same reason
+        // the submit is detached: a caller that disconnected mid-write still
+        // committed the op, and its next request as this user must not be
+        // served state older than what committed. Ordered before the wake, so a
+        // read issued the instant the response lands already sees the mark.
+        //
+        // This node's own view only: on a follower with HTTP forwarding ON the
```
hubcio: critical: with forwarding on, the follower relays the write to the primary, so this task never runs and no floor lands. unqualified GETs stay local, so `POST /streams` then `GET /streams` here is the stale read this PR exists to stop.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +128,187 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+/// detecting one costs `heartbeat_timeout` and escalating it another
+/// `view_change_status_timeout`, and `recovery_barrier_deadline` budgets at
+/// least 15s for the same event, so a read that waits out an election is a read
+/// the caller should retry elsewhere. Far below the SDK's 30s request budget,
+/// inside which it replays the same id on the same connection.
+///
+/// In ticks rather than a bare duration because that is the unit the thing
+/// being waited for moves in, and the unit the simulator steps in.
+#[allow(clippy::cast_possible_truncation)]
+pub const READ_FRONTIER_BUDGET_TICKS: u32 = 6 * TimeoutManager::COMMIT_MESSAGE_TICKS as u32;
```
hubcio: critical: `COMMIT_MESSAGE_TICKS` is only the compile-time default - `set_commit_message_ticks` overrides it at runtime from `[cluster] commit_broadcast_interval`. at 2s this budget is 1.5 broadcasts, not six, so a merely lagging backup gets spurious 503s. derive it from the configured ticks.

core/server/src/http/reads.rs
```diff
@@ -91,10 +97,26 @@ pub(in crate::http) async fn read_local(
     consistency: Consistency,
     code: u32,
     body: &[u8],
-    rule: impl FnOnce(&Permissioner, u32) -> Result<(), IggyError>,
+    rule: impl Fn(&Permissioner, u32) -> Result<(), IggyError>,
 ) -> Result<Bytes, ReadError> {
     await_recovery_barrier(&state.shard).await?;
```
hubcio: critical: `get_clients`, `get_client`, `get_snapshot`, `poll_messages` and `get_consumer_offset` call `authorize_read` directly and never reach here, so they get neither the barrier nor the frontier gate. their binary twins are gated. route them through `read_local`.

core/metadata/src/applied_frontier.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub fn advance(&self, op: u64) {
+        if self.op.fetch_max(op, Ordering::Release) >= op {
+            return;
+        }
+        let mut waiters = self.waiters.lock().unwrap_or_else(PoisonError::into_inner);
+        waiters.entries.retain(|waiter| {
+            if waiter.target > op {
+                return true;
+            }
+            waiter.waker.wake_by_ref();
```
hubcio: warning: `wake_by_ref` runs inside `retain` with the mutex held, on the commit path. every woken reader then blocks on that same lock in `deregister`. collect the wakers, drop the guard, then wake.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +128,187 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+/// from a state machine below that op contradicts the frame the client is
+/// holding. The lag is real on a node whose commit walk trails the client's
+/// epoch -- a backup that forwarded the client's register binds a committed
+/// session while its own `commit_journal` is still behind it (see
+/// [`crate::dispatch::session_ops`]) -- so the gate is not about peer shards:
+/// every shard of a node reads one shared frontier and gates identically.
+///
+/// Fast path is a single `Acquire` load and no await, which is what keeps an
+/// uncontended read shared-nothing. A park costs this connection more than the
+/// read itself: the per-connection drain loop serves one frame at a time, so
```
hubcio, line 263: warning: the per-client queue in `dispatch/mod.rs` is an uncapped `VecDeque` drained one frame at a time, so a park here lets that client's pipelined frames pile up unbounded for the whole budget. cap it, or cut the binary budget.

core/server/src/dispatch/reads.rs
```diff
@@ -250,7 +459,13 @@ pub(in crate::dispatch) async fn handle_non_replicated_request<B, MJ, S, SB>(
         }
         SYNC_CONSUMER_GROUP_CODE => {
```
hubcio: warning: the poll and consumer-offset arms above skip the gate, but `authorize_partition_read` and `resolve_poll_request` both read the metadata STM. create a topic then poll on a lagging node and you get a silent empty poll or a spurious 403.

core/server/src/responses.rs
```diff
@@ -1571,23 +1610,14 @@ pub fn build_raw_pat_reply(
         bytemuck::checked::try_from_bytes::<ReplyHeader>(&committed.as_slice()[..header_len])
```
hubcio: warning: `[..header_len]` panics on a frame shorter than `ReplyHeader`, where the twin in `dispatch/submit.rs` uses `.get()` and warns. use a checked slice.

core/integration/tests/server/http_read_your_writes.rs
```diff
@@ -0,0 +1,137 @@
...
+//! the session entry down again, so the floor the read waits on has to outlive
+//! the session that established it.
+//!
+//! The seeding runs over TCP rather than the primary's own REST listener so the
+//! only HTTP sessions in play are the follower's: a second long-lived REST
+//! session would be competing for VSR client ids with the fresh register each
+//! round mints, which is a different subject.
+//!
+//! The suite asserts the GUARANTEE, not the mechanism: whether the gate parked
+//! is invisible from outside, and on a fast local cluster the follower often
```
hubcio: warning: the header above concedes the follower usually applies within the tick, so this passes with the gate deleted. it pins the wiring, not the guarantee - the binary plane has a deterministic simulator proof, this one has none.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +128,187 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+///
+/// The two named exclusions only look like metadata reads: `DescribeOptions`
+/// decodes a static catalog, and `GetClusterMetadata` answers from the
+/// configured roster plus the consensus view. Holding either buys no
+/// consistency, and the roster read is on the SDK's leader-discovery path,
+/// where the wait would be real. A code this build does not know is excluded
+/// too: its only outcome is `InvalidCommand`, and parking a terminal error for
+/// the whole budget serves nobody.
+///
+/// A deny-list otherwise, so a read code added later is gated by default: the
```
hubcio: nit: the deny-list only covers codes reaching the `_` arm - every named arm opts in by hand, and three don't. drop the claim or move the gate to one call at the top of the match.

core/metadata/src/applied_frontier.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// Monotone, so a lower value is a no-op and wakes nobody.
+    ///
+    /// Must run AFTER the apply's `publish()` and, on the commit path, in the
+    /// same await-free region as `advance_commit_min`: a reader that sees the
+    /// frontier must be guaranteed to see the op's effects.
+    pub fn advance(&self, op: u64) {
+        if self.op.fetch_max(op, Ordering::Release) >= op {
+            return;
+        }
+        let mut waiters = self.waiters.lock().unwrap_or_else(PoisonError::into_inner);
```
hubcio: nit: takes the mutex on every advancing commit even when nothing is parked, which is the normal case. guard it with a relaxed waiter count.

core/server/src/dispatch/reads.rs
```diff
@@ -122,6 +128,187 @@ async fn handle_get_me<B, MJ, S, SB>(
...
+    }
+    let reached = pin!(frontier.reached(watermark));
+    let budget = pin!(budget);
+    match select(reached, budget).await {
+        Either::Left(((), _)) => Ok(FrontierWait::CaughtUp),
+        Either::Right(((), _)) => {
+            warn!(
+                frontier = frontier.get(),
+                watermark,
+                budget = ?READ_FRONTIER_BUDGET,
```
hubcio: nit: one `warn!` per expired read is a warn per client per budget window on a lagging backup - a flood in exactly the case it's meant to surface. drop to `debug!` plus a counter.

core/simulator/src/lib.rs
```diff
@@ -5097,3 +5105,377 @@ mod repair_frontier_tests {
...
+
+    /// Steps left for repair to reach the backup and the held read to answer
+    /// once replication is restored.
+    ///
+    /// Derived, not chosen: one `sim.step()` advances the virtual clock by one
+    /// consensus tick, the unit the gate's budget is denominated in, so phase 1
+    /// spends `STALE_WINDOW_STEPS` of that budget and what remains is the whole
+    /// window the read can still be answered in. A larger number would just
+    /// spin past an expiry the status assertion below already caught.
+    const CONVERGE_STEPS: u32 = server::METADATA_READ_FRONTIER_BUDGET_TICKS - STALE_WINDOW_STEPS;
```
hubcio: nit: `STALE_WINDOW_STEPS + CONVERGE_STEPS` is exactly the budget, so repair one tick slow flips this to the expiry path and the status assert fails confusingly. leave margin and cover expiry in its own case.

core/server/src/session_manager.rs
```diff
@@ -268,15 +281,47 @@ impl SessionManager {
...
+        bound.state = ConnectionState::Bound {
             user_id,
             client_id,
             session,
         };
+        // The session IS the register's commit op, so it floors every metadata
+        // op this client saw committed before it re-homed here. Without the
+        // seed a re-homed connection reads at zero and the gate admits the
+        // pre-write state its own last write already replaced.
+        bound.metadata_watermark = bound.metadata_watermark.max(session);
```
hubcio, line 304: nit: `metadata_watermark` survives the demotion above and isn't cleared on `login`, so a different user on the same socket inherits the old floor. only over-waits, never serves stale, but worth resetting.

core/server/src/dispatch/submit.rs
```diff
@@ -192,3 +196,143 @@ where
...
+///
+/// Every reply on this path is result-framed (`Operation::is_result_framed`
+/// covers every metadata op; the partition plane grades through
+/// `classify_partition_reply` instead), so a missing result section is a
+/// malformed frame, not a bare payload.
+///
+/// Shared with the HTTP write path, which grades the same frames off the same
+/// submit entry point ([`submit_client_request_on_owner`]); one classifier is
+/// what keeps the two planes' watermarks meaning the same thing.
+pub fn committed_reply_commit(reply: &Message<GenericHeader>) -> Option<u64> {
```
hubcio, line 224: simplification: this and `build_raw_pat_reply` run the same three checks on the same frames and have to stay in step. extract one `committed_reply_header` helper both call.

hubcio (changes requested):

core/server/src/http/submit.rs
```diff
@@ -106,13 +105,32 @@ pub(in crate::http) async fn submit_committed(
...
+        // write is relayed to the primary by the middleware and never reaches
+        // this task, so the follower's floor stays where the register left it
+        // and its read-your-writes guarantee is the register epoch's. Closing
```
hubcio: the last three lines are stale as of this branch. `forward.rs` now copies `iggy-applied-op` off the relayed response and `record_relayed_floor` records it as this caller's floor, so the primary's commit op does come back with the relayed response. a future reader lands here and concludes the forwarding hole is still open.

one caveat that does survive and is worth keeping in the rewrite: `classify_reply` relays a 503 carrying `TransientNotCommitted` instead of retrying it, and `record_relayed_floor` only records on 2xx - so a write that did commit behind that code still leaves no floor. that's the honest version of what this comment is reaching for.

core/server/src/http/reads.rs
```diff
@@ -117,6 +119,108 @@ pub(in crate::http) async fn read_local(
...
+/// (307 to the primary), which [`authorize_read`] decides before this wait and
+/// which this wait never sees. This gate makes an UNQUALIFIED read
+/// read-your-writes for its own user, at no redirect and no consensus round
+/// trip. Per user rather than per credential because a bearer is not stable:
+/// a refreshed access token is a new credential for the same writer (see
+/// [`crate::http::state::MetadataWatermarks`]).
+///
+/// Scope is this node's own view. A user whose write this node relayed over
+/// HTTP, or who wrote through a different node entirely, left no floor here;
+/// closing that needs the serving primary's commit op to reach the reading
```
hubcio: first clause is stale - a write this node relayed over HTTP does leave a floor now, `record_relayed_floor` records the serving primary's applied op off the response header. second clause still holds: a user who wrote through a different node entirely leaves nothing here.

something like:

```
/// Scope is this node's own view. A write this node RELAYED leaves a floor
/// here: the relay records the serving primary's applied op off the response
/// (see `http::forward::record_relayed_floor`). A user who wrote through a
/// different node entirely still leaves none.
```

hubcio (changes requested):
pr description paragraph 3 still says the forwarding-active case is not closed - "that is additive but touches every control-plane write response, so it is left out here and documented at the gate". 456c38a01 closes it in this same pr via the applied-op header. worth updating before merge, since the body lands in the squash commit message.

the two comments below are the same staleness in the code docs.

hubcio (approved):

mmodzelewski (approved):

PR #4025, opened 2026-09-01
perf(server): serve poll replies without copying record bytes

hubcio:
A PolledMessages reply concatenated the poll fragments into one
buffer, walked the batch records into a second, then copied that
into a zeroed reply frame: three copies plus a memset per poll,
about four times the payload in memory traffic before the kernel
copy. Nothing in the format required it, the reply record
encoding is the storage encoding, except at-rest decryption.

Build the reply as one 272-byte head fragment (ReplyHeader plus
partition_id, current_offset, count) followed by the journal or
disk Frozen fragments as they are, deriving count by walking the
batch headers across fragments. The frame travels end to end as
BusMessage = ResponseBacking (SmallVec inline for the single
fragment every other reply is, 32 bytes) through send_to_client,
the shard reply lane, the connection mailboxes and the writers:
TCP flattens the fragments into one writev chunked at IOV_MAX,
QUIC and TLS write per fragment, WS and WSS join once because
tungstenite takes a single payload buffer. The decrypt path and
the HTTP JSON handler keep the flattening builder, which also
pins the wire bytes as the oracle in the new unit tests.
LifecycleFrame stays 160 bytes.

### DIFF
--- a/core/message_bus/src/transports/tcp.rs
+++ b/core/message_bus/src/transports/tcp.rs
@@ -245,6 +245,8 @@ async fn writer_loop(
     let _wake_reader = scopeguard::guard(conn_shutdown, |s| s.trigger());
     let mut batch: Vec<BusMessage> = Vec::with_capacity(max_batch);
     let mut iovecs: Vec<Frozen<MESSAGE_ALIGN>> = Vec::with_capacity(max_batch);
+    // Sized by the first wide batch; most connections never need it.
+    let mut scratch: Vec<Frozen<MESSAGE_ALIGN>> = Vec::new();
     let mut shutdown_fut = Box::pin(shutdown.wait().fuse());
 
     loop {
@@ -279,7 +281,7 @@ async fn writer_loop(
         for frame in batch.drain(..) {
             iovecs.extend(frame.into_fragments());
         }
-        let result = write_iovecs(&mut write_half, &mut iovecs).await;
+        let result = write_iovecs(&mut write_half, &mut iovecs, &mut scratch).await;
 
         if let Err(e) = result {
             error!(
@@ -296,12 +298,14 @@ async fn writer_loop(
 
 /// `write_vectored_all` over `iovecs`, at most `IOV_MAX` entries per call so
 /// a frame fragmented past the syscall limit (a wide poll reply) cannot fail
-/// with `EMSGSIZE` (the socket path is `io_uring` `sendmsg(2)`). Leaves
-/// `iovecs` empty with an allocation ready for the next batch.
+/// with `EMSGSIZE` (the socket path is `io_uring` `sendmsg(2)`). `scratch`
+/// carries each chunk in flight. Leaves both vecs empty with their
+/// allocations intact for the next batch.
 #[allow(clippy::future_not_send)]
 async fn write_iovecs(
     write_half: &mut TcpStream,
     iovecs: &mut Vec<Frozen<MESSAGE_ALIGN>>,
+    scratch: &mut Vec<Frozen<MESSAGE_ALIGN>>,
 ) -> io::Result<()> {
     if iovecs.len() <= IOV_MAX {
         let compio::BufResult(result, mut returned) =
@@ -311,22 +315,21 @@ async fn write_iovecs(
         *iovecs = returned;
         return Ok(());
     }
-    // Chunk through a cursor and one reused scratch vec. `split_off`-style
-    // chunking memmoves the whole tail per syscall, O(n^2) over a batch of n
-    // iovecs on the shard's single-threaded executor, and n is client-chosen
-    // (an unclamped poll count fans out to 1-2 iovecs per batch).
-    let mut source = mem::take(iovecs).into_iter();
-    let mut scratch = Vec::with_capacity(IOV_MAX);
+    // A full-range drain shifts nothing, so the batch moves into `scratch` in
+    // O(n). Draining `IOV_MAX` at a time from the front would memmove the tail
+    // per syscall, O(n^2) over a client-chosen n (an unclamped poll count fans
+    // out to 1-2 iovecs per batch) on the shard's single-threaded executor.
+    let mut source = iovecs.drain(..);
     loop {
         scratch.extend(source.by_ref().take(IOV_MAX));
         if scratch.is_empty() {
-            *iovecs = scratch;
             return Ok(());
         }
-        let compio::BufResult(result, mut returned) = write_half.write_vectored_all(scratch).await;
+        let compio::BufResult(result, mut returned) =
+            write_half.write_vectored_all(mem::take(scratch)).await;
         result?;
         returned.clear();
-        scratch = returned;
+        *scratch = returned;
     }
 }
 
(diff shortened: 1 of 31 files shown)


core/server/src/responses.rs
```diff
@@ -1652,8 +1662,148 @@ where
...
+    if encryptor.is_some() {
+        let body = build_polled_messages_body(partition_id, current_offset, fragments, encryptor)?;
+        let reply = build_reply_from_bytes(request_header, client_id, session, commit, &body);
+        return Ok(reply.into_generic().into_frozen().into());
+    }
+
+    let mut frames = ResponseFragments::with_capacity(fragments.len() + 1);
+    frames.extend(fragments.into_iter().map(Fragment::into_frozen));
+    let count = polled_message_count(&frames)?;
+    let records_len: usize = frames.iter().map(Frozen::len).sum();
```
numinnex, line 1695: Each fragment keeps a refcount on its **whole** source allocation, not just the sliced window: `Frozen::slice` narrows ptr/len while the `ControlBlock` still owns the base buffer. On the disk tier that source is a `DISK_POLL_CHUNK` read buffer (1 MiB, `poll_plan.rs:493`, grown 4^k at `:607`), and `walk_disk_chunk` stops as soon as `matched >= count`.

On master `build_polled_messages_body` took `PollFragments` by value and released the chunk before the reply was enqueued. Here the pin lasts as long as the frame sits in the per-connection mailbox (`peer_queue_capacity = 256`).

So a `count=1` poll off a cold partition ships ~600 B and retains 1 MiB. 100 stalled consumers x 256 queued is ~25.6 GiB held against a few MB of actual reply bytes.

What makes this different from the pre-existing `256 * MAX_MESSAGE_SIZE` ceiling: that bound is client-requested and therefore self-limiting, while this one is server read granularity with no relation to what was asked for. And `peer_queue_capacity` counts frames, so nothing in the system observes the retained bytes -- there is no gauge, no byte knob, and every other retained-payload path in the server has an explicit byte budget (`transfer_served_cache_bytes_max` / `PARTITION_ARTIFACT_LEN_DEFAULT`).

Suggest bounding the pin: copy the fragment out when `fragment.len()` is a small fraction of its source allocation, or read the exact batch extent on the disk tier instead of a 1 MiB chunk.

core/message_bus/src/transports/tcp.rs
```diff
@@ -289,6 +294,33 @@ async fn writer_loop(
...
+        };
+        let compio::BufResult(result, mut returned) = write_half.write_vectored_all(chunk).await;
+        result?;
+        if rest.is_empty() {
+            returned.clear();
+            *iovecs = returned;
+            return Ok(());
+        }
+        chunk = rest;
+    }
```
numinnex: `Vec::split_off(at)` allocates a new `Vec` **and** moves the `len - at` tail elements, so chunking this way costs ~`n^2/2048` element moves for an `n`-iovec batch rather than `n`.

`n` is client-chosen: `wire.count` reaches `MessageLookup::count` unclamped (`dispatch.rs:2519`) and `push_selected_batch_fragments` emits 1-2 fragments per batch. 50k fragments is ~29 MB of memmove; 262k (64 MiB of one-message batches) is ~805 MB, roughly 80 ms. There is no `.await` in the loop and `writer_loop` runs on the shard's single-threaded compio executor, so that blocks every other task on the core, consensus I/O included.

At that shape the new path also moves more bytes than the gather it replaces (~805 MB vs ~192 MB), which inverts the premise of the change.

Note `drain(..k)` from the front has the same problem (it shifts the tail too). `into_iter()` advances a cursor instead, which is O(n) overall:

```rust
// `scratch` is reused across chunks and across batches.
let mut it = mem::take(iovecs).into_iter();
loop {
    scratch.extend(it.by_ref().take(IOV_MAX));
    if scratch.is_empty() {
        return Ok(());
    }
    let compio::BufResult(result, mut returned) =
        write_half.write_vectored_all(mem::take(scratch)).await;
    returned.clear();
    *scratch = returned;
    result?;
}
```

Two smaller things in the same function:

- The doc comment says "Leaves `iovecs` empty, keeping its allocation for the next batch", but the original allocation is dropped at the `chunk = rest` rebind whenever chunking fires -- what gets restored is the last `split_off` remainder, i.e. the reuse is lost on exactly the wide-poll path this was written for.
- "the syscall fails with `EINVAL`" is the wrong errno here. `write_vectored_all` on a `TcpStream` goes `Socket::send_vectored` -> `SendVectored` -> io_uring `opcode::SendMsg` (`compio-driver/src/sys/op/socket/iour.rs:167`), i.e. `sendmsg(2)`, and `__copy_msghdr` returns **EMSGSIZE** for `msg_iovlen > UIO_MAXIOV`. `EINVAL` is correct only for the constant's other consumer, `messages_writer.rs:223`, which goes through `opcode::Writev` / `pwritev(2)`. The same wording is in the new `iobuf.rs` doc; `config.rs:110`'s existing `EMSGSIZE` was right.

core/server/src/responses.rs
hubcio: good catch, fixed in https://github.com/apache/iggy/commit/7df4ee4b45e64567c713256e61a306560166e4ab. matched slices copy out into one compact buffer when they cover under a quarter of the chunk, so a sparse poll cannot pin the 1 MiB read. dense walks stay zero-copy.

core/message_bus/src/transports/tcp.rs
hubcio: fixed in https://github.com/apache/iggy/commit/7df4ee4b45e64567c713256e61a306560166e4ab. chunking now drains an `into_iter` cursor through one reused scratch vec, and the docs say EMSGSIZE for the socket path - EINVAL stays only on the pwritev side in `messages_writer`.

core/partitions/src/poll_plan.rs
```diff
@@ -1032,6 +1034,54 @@ struct ChunkWalk {
...
 }
 
+/// A fragment sliced from a disk chunk keeps the WHOLE chunk allocation
+/// alive until the reply frame is written out, and a reply can sit in a
+/// per-connection mailbox for a while. Copy the matched bytes out when they
+/// cover less than this fraction of the chunk, so a sparse match (a
+/// `count=1` poll off a cold partition) cannot pin ~1 MiB per queued reply;
+/// a dense catch-up read keeps the zero-copy path, where retained bytes stay
+/// within this factor of shipped bytes.
+const SPARSE_CHUNK_PIN_DIVISOR: usize = 4;
```
numinnex: Disk tier verified fixed — the bound composes across chunks and segments, the grow-and-re-read path is a provable no-op (`consumed == 0` means nothing was pushed, so `borrowed == 0`), and `chunk.len()` is the right denominator since `read_chunk_with_retry` never slices the buffer. Byte-for-byte parity survives the in-place rewrite. Two things about the framing, though.

**The resident tier keeps the same pin.** `try_push_resident_entry` (journal.rs, `Fragment::slice(prepare.clone(), ..)`) gets no unpinning. The reason this was dismissed earlier — "the journal owns those bytes anyway" — does not hold: `Journal::evict_prefix` (journal.rs:553) calls `inner.storage.drain()`, which is `std::mem::take(entries)` (journal.rs:282-291), on the routine commit-flush path. The journal releases its refcount and a still-queued reply becomes sole owner of the whole prepare allocation. That window is unbounded on single-replica groups, where the `evicted_ring` backstop is gated off by `repair_retention` (iggy_partition.rs:504-506).

This is introduced here, not pre-existing: on master `build_polled_messages_body` consumed `PollFragments` by value and dropped every refcount before the reply was enqueued, so no fragment outlived `handle_poll_messages` on either tier.

Magnitude is workload-dependent and mostly benign — `batch_length` defaults to 1000 on both producer and consumer, so the default path selects whole batches, `borrowed` is approximately the whole prepare, and the predicate would not copy anyway. The bad case is a small `count` against a large-batch producer, where a partial selection ships ~1.3 KiB and pins the entire batch.

One caveat if you reuse this helper there: `execute_resident()` runs inline on the shard pump (the disk path is detached through `spawn_poll_io`), so a ratio-gated copy could put a ~1 MiB memcpy on the pump. An absolute byte floor (copy when the selected bytes are under ~64 KiB, regardless of ratio) keeps the copy bounded and fires only on the at-most-two partial batches per poll.

**The stated bound is not what the predicate delivers.** "retained bytes stay within this factor of shipped bytes" (and "4x shipped bytes" in the commit message) overstates it three ways: it is disk-tier only; `borrowed` excludes the owned rewritten-header fragments that are also shipped, so the denominator is not shipped bytes; and the copy allocates 4096-aligned, so a chunk contributing 300 B still retains a page. Suggested replacement:

> On the disk tier, the chunk allocations a poll reply keeps alive are bounded by `SPARSE_CHUNK_PIN_DIVISOR` times the record bytes it serves from disk, plus one page per compacted chunk. This is a ratio, not an absolute cap — a grown chunk can still retain tens of MiB — and it does not cover the resident tier or the reply's absolute size.

core/partitions/src/poll_plan.rs
```diff
@@ -1032,6 +1034,54 @@ struct ChunkWalk {
...
+        return;
+    }
+
+    let mut compact = Owned::<4096>::zeroed(borrowed);
+    let mut cursor = 0;
+    for fragment in pushed.iter().filter(|f| f.borrows_from(chunk)) {
+        let bytes = fragment.as_slice();
+        compact.as_mut_slice()[cursor..cursor + bytes.len()].copy_from_slice(bytes);
+        cursor += bytes.len();
+    }
```
numinnex: `zeroed(borrowed)` memsets every byte the loop below then overwrites — the borrowing fragments' lengths sum to exactly `borrowed` by construction, so nothing is left at zero. That doubles the compaction's memory traffic (2x `borrowed` instead of 1x, up to ~256 KiB of dead memset per 1 MiB chunk), inside the function added to bound the copy in the first place.

This file already makes the argument, 260 lines up at `read_chunk_with_retry`:

> `with_capacity` (len == 0, capacity == len) instead of `zeroed`: `read_exact_at` fills the whole capacity in place and advances the length via `SetLen`, so the `zeroed` memset of up to 1MiB per chunk was pure waste - every byte is overwritten by the read.

`Owned::with_capacity` alone will not work here (it gives `len == 0`, so `as_mut_slice()` is empty). The gather idiom this PR already uses is the fix — `ResponseBacking::into_contiguous`, core/server_common/src/consensus_message.rs:122-135:

```rust
let mut compact: AVec<u8, ConstAlign<4096>> = AVec::with_capacity(4096, borrowed);
for fragment in pushed.iter().filter(|f| f.borrows_from(chunk)) {
    compact.extend_from_slice(fragment.as_slice());
}
let compact = Frozen::from(Owned::from(compact));
```

That also drops the `cursor` bookkeeping in the first loop. `From<AVec> for Owned` is at iobuf.rs:31.

core/message_bus/src/transports/tcp.rs
```diff
@@ -289,6 +294,42 @@ async fn writer_loop(
...
+        if scratch.is_empty() {
+            *iovecs = scratch;
+            return Ok(());
+        }
+        let compio::BufResult(result, mut returned) = write_half.write_vectored_all(scratch).await;
+        result?;
+        returned.clear();
+        scratch = returned;
+    }
+}
```
numinnex: Chunking fix verified: O(n), terminates, FIFO-preserving, and the error path leaves the caller sane. `into_iter` advances a cursor instead of shifting a tail, so each element moves exactly once. Boundaries check out too — `len == IOV_MAX` correctly takes the fast path (`sendmsg` rejects only `> UIO_MAXIOV`), and the exact-multiple case terminates on the empty `extend` before issuing a zero-iovec write.

One residual: the two branches do not treat the caller's allocation the same way. The fast path round-trips the caller's own Vec, but the chunked path moves it into `into_iter()` and drops it, handing back a fresh `Vec::with_capacity(IOV_MAX)`. `writer_loop` then re-grows `iovecs` from 1024 to n on every wide batch — geometric, so amortised O(n) and nowhere near the quadratic this replaced, but it is ~1.2 MB of memcpy plus a handful of malloc/free per batch at n=50k. So the doc's "an allocation ready for the next batch" is true but capped at `IOV_MAX`, not at the high-water mark.

Hoisting `scratch` up into `writer_loop` next to `iovecs` fixes both: each buffer then persists at its own high-water mark, the per-wide-batch `Vec::with_capacity(IOV_MAX)` disappears, and the doc claim becomes true for both branches without rewording.

Also worth a test while this is fresh — the rewritten loop has no boundary coverage. `run_writes_frames_fragmented_past_iov_max` produces 1793 fragments, which exercises the chunked branch but misses `len == IOV_MAX`, `IOV_MAX + 1`, and `2 * IOV_MAX`. The exact multiple is the one worth pinning, since that is where a wrong loop would emit a zero-iovec `sendmsg`.

core/partitions/src/poll_plan.rs
hubcio: fixed in 4a1344c3c. the resident walk now runs the same helper per prepare, capped at 64 KiB per copy since it sits inline on the pump. doc carries your bound wording.

core/partitions/src/poll_plan.rs
hubcio: fixed in 065b99cc7, `Owned::extend_from_slice` into an exact-capacity buffer, one pass.

core/message_bus/src/transports/tcp.rs
hubcio: fixed in 78097865e. `scratch` lives in `writer_loop` now and the chunked branch drains the caller's vec in place, so both keep their high-water marks. boundary test covers `IOV_MAX`, `IOV_MAX + 1` and `2 * IOV_MAX`.

numinnex (approved):

spetz (approved):

PR #4027, opened 2026-09-01
refactor(csharp): update integer properties to unsigned types for consistency

lukaszzborek:


foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -706,7 +720,7 @@ headers[new HeaderKey
             }
 
             var valueLength = BinaryPrimitives.ReadInt32LittleEndian(payload[position..(position + 4)]);
-            if (valueLength is <= 0 or > 255)
+            if (valueLength is <= 0 or > 255 || !ValueLengthMatchesKind(valueKind, valueLength))
```
hubcio: one header with a bad width now nulls the whole `UserHeaders` map, which callers read as "no headers". throw `MalformedResponseException` from the getter instead, like `MapHeaders` does at line 641.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -1339,37 +1439,31 @@ internal static ClusterMetadata MapClusterMetadata(ReadOnlySpan<byte> payload)
...
-
-        // Read IP
-        var ipLength = BinaryPrimitives.ReadUInt32LittleEndian(payload[position..(position + 4)]);
-        position += 4;
-
-        var ip = Encoding.UTF8.GetString(payload[position..(position + (int)ipLength)]);
-        position += (int)ipLength;
-
-        // Read transport endpoints (4 ports, each 2 bytes)
         var tcp = BinaryPrimitives.ReadUInt16LittleEndian(payload[position..(position + 2)]);
```
hubcio, line 1391: nothing checks that the 10-byte tail fits, so a truncated node throws `ArgumentOutOfRangeException` instead of `MalformedResponseException` like the other mappers. add `if (position + 10 > payload.Length) throw new MalformedResponseException(...)` first.

foreign/csharp/Iggy_SDK_Tests/MapperTests/BinaryMapper.cs
```diff
@@ -615,4 +673,199 @@ private static byte[] BuildEncryptedFrame(AesMessageEncryptor encryptor, uint of
...
+        Assert.Equal(8091, metadata.Nodes[1].Endpoints.Tcp);
+    }
+
+    [Fact]
+    public void MapClusterMetadata_TruncatedNode_Throws()
+    {
+        var payload = CreateClusterMetadataPayload("cluster", ("node-1", "10.0.0.1", 8090, 1, 1));
+        var truncated = payload.AsSpan(0, payload.Length - 3).ToArray();
+
+        Assert.ThrowsAny<SystemException>(() => Mappers.BinaryMapper.MapClusterMetadata(truncated));
```
hubcio: `MalformedResponseException` is not a `SystemException`, so this only passes because of the slicing exception from the unguarded tail. assert `Throws<MalformedResponseException>` once the tail is checked.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -758,6 +772,71 @@ internal static HeaderKind MapHeaderKind(byte value)
...
+        return (int)count;
+    }
+
+    /// <summary>
+    ///     Length prefixes are u32 on the wire. Reading them signed would let 2^31 and above go negative and
+    ///     surface as a slicing exception instead of <see cref="MalformedResponseException" />.
+    /// </summary>
+    private static int ReadLength(ReadOnlySpan<byte> payload, int position, string field)
+    {
+        var length = BinaryPrimitives.ReadUInt32LittleEndian(payload[position..(position + 4)]);
```
hubcio, line 763: also at line 1415.

this slice is unguarded, so a payload cut inside the prefix throws `ArgumentOutOfRangeException` - the exact thing the doc comment above says this helper prevents. check `position + 4 > payload.Length` first.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -1318,18 +1413,23 @@ private static void EnsureFits(ReadOnlySpan<byte> payload, int position, int len
     internal static ClusterMetadata MapClusterMetadata(ReadOnlySpan<byte> payload)
     {
         var nameLength = BinaryPrimitives.ReadUInt32LittleEndian(payload[..4]);
+        if (nameLength > payload.Length - 4 - 4)
+        {
+            throw new MalformedResponseException(
+                $"Cluster name length {nameLength} exceeds remaining payload of {payload.Length - 4} bytes.");
```
hubcio: the check keeps 8 bytes back but the message says `payload.Length - 4`. `ReadString(payload, ref position, "Cluster name")` plus a 4-byte check before the nodes count read would replace this block and fix the message.

foreign/csharp/Iggy_SDK_Tests/MapperTests/BinaryMapper.cs
```diff
@@ -615,4 +673,199 @@ private static byte[] BuildEncryptedFrame(AesMessageEncryptor encryptor, uint of
...
+        return payload;
+    }
+
+    [Fact]
+    public void MapTopics_OversizedTopicsCount_Throws()
+    {
+        var payload = new byte[4];
+        BinaryPrimitives.WriteUInt32LittleEndian(payload, uint.MaxValue);
+
+        Assert.ThrowsAny<Exception>(() => Mappers.BinaryMapper.MapTopics(payload));
```
hubcio: the path throws `MalformedResponseException` deterministically and the sibling tests at lines 684 and 734 assert it. use `Assert.Throws<MalformedResponseException>` here too.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -444,7 +444,7 @@ private async Task PollMessagesAsync(CancellationToken ct)
                 {
                     _logger.LogDebug("No new messages found, committing offset {Offset} for partition {PartitionId}",
                         lastPolledPartitionOffset, messages.PartitionId);
-                    await StoreOffsetAsync(lastPolledPartitionOffset, (uint)messages.PartitionId, false, ct);
+                    await StoreOffsetAsync(lastPolledPartitionOffset, messages.PartitionId, false, ct);
```
hubcio, line 460: line 453 still does `AddOrUpdate` with a lambda while the rented loop switched to `_lastPolledOffset[id] = currentOffset`. same indexer here.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -1244,16 +1330,29 @@ var partitionId
...
         var partitionsCount = BinaryPrimitives.ReadUInt32LittleEndian(payload[(position + 4)..(position + 8)]);
         var membersCount = BinaryPrimitives.ReadUInt32LittleEndian(payload[(position + 8)..(position + 12)]);
         var nameLength = payload[position + 12];
+        if (position + CONSUMER_GROUP_HEADER_SIZE + nameLength > payload.Length)
+        {
+            throw new MalformedResponseException(
+                $"Malformed consumer group at byte {position}: name length {nameLength} does not fit the response.");
+        }
+
         var name = Encoding.UTF8.GetString(payload[(position + 13)..(position + 13 + nameLength)]);
```
hubcio: `position + 13` is `CONSUMER_GROUP_HEADER_SIZE`. same for `MEMBER_HEADER_SIZE` at line 1323.

foreign/csharp/Iggy_SDK/Kinds/Consumer.cs
```diff
@@ -48,6 +48,20 @@ public static Consumer New(int id)
         };
     }
 
+    /// <summary>
+    ///     Creates a new regular consumer identifier.
+    /// </summary>
+    /// <param name="id">Identifier value</param>
+    /// <returns>Consumer instance</returns>
+    public static Consumer New(uint id)
```
hubcio, line 53: also at line 98.

`New(uint)` and `Group(uint)` have no callers. either use them or drop them.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -758,6 +772,71 @@ internal static HeaderKind MapHeaderKind(byte value)
         };
     }
 
+    /// <summary>
+    ///     The typed accessors on <see cref="HeaderValue" /> slice the raw bytes by kind, so a value whose
+    ///     length disagrees with its kind byte is rejected at parse time instead of throwing in the caller's
+    ///     message handler.
+    /// </summary>
+    private static bool ValueLengthMatchesKind(HeaderKind kind, int length)
```
hubcio, line 718: the rust decoder rejects a key whose length doesn't match its kind too. apply this to `keyKind`/`keyLength` as well - after the null return at line 723 becomes a real error, or it just adds another silent drop.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -611,8 +619,8 @@ internal static Dictionary<HeaderKey, HeaderValue> MapHeaders(ReadOnlySpan<byte>
             var keyKind = MapHeaderKind(payload[position]);
             position++;
 
-            var keyLength = BinaryPrimitives.ReadInt32LittleEndian(payload[position..(position + 4)]);
-            if (keyLength is 0 or > 255)
+            var keyLength = (int)BinaryPrimitives.ReadUInt32LittleEndian(payload[position..(position + 4)]);
+            if (keyLength is <= 0 or > 255)
             {
                 throw new ArgumentException("Key has incorrect size, must be between 1 and 255", nameof(keyLength));
```
hubcio: also at line 638.

`ArgumentException` here and `MalformedResponseException` at line 643 for the same kind of bad wire data. use `MalformedResponseException` for both.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -706,7 +720,7 @@ headers[new HeaderKey
             }
 
             var valueLength = BinaryPrimitives.ReadInt32LittleEndian(payload[position..(position + 4)]);
```
hubcio: also at line 690.

still `ReadInt32LittleEndian` while `MapHeaders` moved to `ReadUInt32LittleEndian` - and this is the copy on the production path. swap both.

foreign/csharp/Iggy_SDK.Tests.Integration/ClusterRedirectionTests.cs
```diff
@@ -135,7 +135,7 @@ public async Task ConnectToFollowerWithPersonalAccessToken_Should_RedirectToLead
         }
 
         authResponse.ShouldNotBeNull();
-        authResponse!.UserId.ShouldBeGreaterThanOrEqualTo(0);
+        authResponse!.UserId.ShouldBeGreaterThanOrEqualTo(0u);
```
hubcio: `>= 0u` on a `uint` always passes, so this asserts nothing now. drop it or assert `UserId.ShouldNotBe(uint.MaxValue)`.

foreign/csharp/Iggy_SDK.Tests.Integration/PersonalAccessTokenTests.cs
```diff
@@ -91,7 +91,7 @@ public async Task LoginWithPersonalAccessToken_Should_Be_Successfully(Protocol p
         var authResponse = await loginClient.LoginWithPersonalAccessTokenAsync(response!.Token);
 
         authResponse.ShouldNotBeNull();
-        authResponse.UserId.ShouldBeGreaterThanOrEqualTo(0);
+        authResponse.UserId.ShouldBeGreaterThanOrEqualTo(0u);
```
hubcio: always true on a `uint`. drop it or assert `UserId.ShouldNotBe(uint.MaxValue)`.

foreign/csharp/Iggy_SDK.Tests.Integration/SystemTests.cs
```diff
@@ -141,7 +141,7 @@ await client.SendMessagesAsync(Identifier.String(streamName),
 
         var response = await client.GetStatsAsync();
         response.ShouldNotBeNull();
-        response.ProcessId.ShouldBeGreaterThanOrEqualTo(0);
+        response.ProcessId.ShouldBeGreaterThanOrEqualTo(0u);
```
hubcio: always true on a `uint`. drop it or assert `ProcessId.ShouldNotBe(0u)`.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -624,12 +632,18 @@ internal static Dictionary<HeaderKey, HeaderValue> MapHeaders(ReadOnlySpan<byte>
...
 
-            var valueLength = BinaryPrimitives.ReadInt32LittleEndian(payload[position..(position + 4)]);
-            if (valueLength is 0 or > 255)
+            var valueLength = (int)BinaryPrimitives.ReadUInt32LittleEndian(payload[position..(position + 4)]);
+            if (valueLength is <= 0 or > 255)
             {
                 throw new ArgumentException("Value has incorrect size, must be between 1 and 255", nameof(valueLength));
             }
 
+            if (!ValueLengthMatchesKind(valueKind, valueLength))
```
hubcio: `MapHeaders` has no production callers, only tests - `TryMapHeaders` is the real path. drop it, or make it `TryMapHeaders(payload) ?? throw new MalformedResponseException(...)` so there is one parser to keep in sync.

foreign/csharp/Iggy_SDK/Kinds/Partitioning.cs
```diff
@@ -72,6 +72,24 @@ public static Partitioning PartitionId(int value)
         };
     }
 
+    /// <summary>
+    ///     Creates a partitioning strategy that use a specific partition id.
+    /// </summary>
+    /// <param name="value">Partition id</param>
+    /// <returns>Partitioning instance</returns>
+    public static Partitioning PartitionId(uint value)
```
hubcio, line 73: the `int` overload above duplicates this body and lets `PartitionId(-1)` send partition 4294967295. make it forward: `ArgumentOutOfRangeException.ThrowIfNegative(value); return PartitionId((uint)value);` - same for `Consumer.New`/`Group` and `Identifier.Numeric`.

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -1309,7 +1404,7 @@ private static void EnsureFits(ReadOnlySpan<byte> payload, int position, int len
     {
```
hubcio: `EnsureFits` has one caller left after this change (line 1377). inline it.

foreign/csharp/Iggy_SDK/Contracts/Tcp/TcpContracts.cs
```diff
@@ -335,7 +335,7 @@ private static int CalculatePermissionsSize(Permissions data)
     {
         var size = 10;
 
-        if (data.Streams is not null)
+        if (data.Streams is { Count: > 0 })
```
hubcio: this walks the same tree as `GetBytesFromPermissions` with hand-kept offsets, only to size a `stackalloc` that grows with the caller's streams x topics. one pass into an `ArrayBufferWriter<byte>` would drop the pre-pass and the stack risk.

foreign/csharp/Iggy_SDK_Tests/Utils/BinaryFactory.cs
```diff
@@ -167,20 +167,21 @@ internal static byte[] CreatePartitionPayload(int id, int segmentsCount, int cur
     internal static byte[] CreateGroupPayload(uint id, uint membersCount, uint partitionsCount, string name,
         List<int>? partitionsOnMember = null)
```
hubcio: `List<int>? partitionsOnMember` and the `WriteInt32` calls below are the last signed bits in this factory now that the mapper reads `uint`. widen them.

hubcio (changes requested):
outside the diff:

- `TcpContracts.cs:282` - `stream.Topics != null` has the same empty-dictionary bug you fixed for `Streams`: an empty `Topics` writes `has_topics = 1` with no entries and the server rejects the request. same `is { Count: > 0 }` here and at line 347.
- `TcpContracts.cs:151` - `UpdatePermissions(userId, null)` sizes the buffer without the has-permissions byte, then writes it at line 167. index out of range on the documented default `permissions = null`.
- `BinaryFactory.cs:155` - `CreatePartitionPayload` has no callers and throws on first use (16-byte buffer, 8-byte writes at offsets 12 and 16). delete it.

foreign/csharp/Iggy_SDK/Kinds/Consumer.cs
lukaszzborek: kept, `int` overloads now forward to `uint` ones

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
lukaszzborek: moved to `MapHeaders` and throw `MalformedResponseException`

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
lukaszzborek: dropped `TryMapHeaders`, `MapHeaders` is the single parser now

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
```diff
@@ -88,7 +88,7 @@ public sealed partial class TcpMessageStream : ISessionGenerationProvider
     ///     <c>RESYNC_REQUIRED_PARTITION_SENTINEL</c> (<c>u32::MAX</c>). The reply header carries no status for an
     ///     empty poll, so the sentinel is the only channel the coordinator has to ask for a re-sync.
     /// </summary>
-    private const int VsrResyncRequiredPartitionSentinel = -1;
+    private const uint VsrResyncRequiredPartitionSentinel = uint.MaxValue;
```
hubcio, line 91: pre-existing, not this PR: a member holding zero partitions fails `HasAssignment`, so every poll re-syncs the group, returns empty, and the consumer re-polls with no delay unless `PollingIntervalMs > 0`. mark that empty reply with the `0xFFFF_FFFE` no-assigned-partition sentinel go and node use, and back off in the consumer when it sees it.

hubcio (changes requested):

spetz (dismissed):
Looks good, here are 2 extra things to improve outisde the scope of this PR, so feel free to either update this PR or we can add these to the next one.

`foreign/csharp/Iggy_SDK/Vsr/ConsumerGroupClientState.cs:46` - an empty cached assignment is treated as missing, causing repeated synchronization and a possible hot loop when `PollingIntervalMs` is `0`. Consider preserving empty assignments and returning the `0xFFFF_FFFE` no-assignment sentinel with backoff.

`foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs:803` - the options-length prefix is read without confirming four bytes remain, so truncated input throws `ArgumentOutOfRangeException` instead of `MalformedResponseException`. Consider reusing `ReadLength`.

numinnex (dismissed):

foreign/csharp/Iggy_SDK/IggyClient/Implementations/TcpMessageStream.Vsr.cs
lukaszzborek: fixed

lukaszzborek:
2 extra comments from @spetz was also done in this PR

foreign/csharp/Iggy_SDK/Mappers/BinaryMapper.cs
```diff
@@ -773,25 +795,15 @@ private static bool TryMapHeaderKind(byte value, out HeaderKind kind)
...
+        var optionsLength = ReadLength(payload, position, "Options block");
+        readBytes = 4 + optionsLength;
 
         var options = new Dictionary<HeaderKey, HeaderValue>();
         var cursor = position + 4;
-        var end = cursor + (int)optionsLength;
+        var end = cursor + optionsLength;
         while (cursor < end)
         {
             var keyKindCode = ReadOptionByte(payload, ref cursor, end, position);
```
hubcio, line 809: `MapOptions` never runs `ValueLengthMatchesKind`, so a 3-byte `Uint64` option parses clean here and then throws out of `BitConverter.ToUInt64` in the caller. rust checks the width on this path too - add the same check `MapHeaders` does.

foreign/csharp/Iggy_SDK/Vsr/ConsumerGroupClientState.cs
```diff
@@ -38,12 +38,15 @@ internal sealed class ConsumerGroupClientState
     /// <summary>Nothing tells this client when another one resizes a topic, so the count expires on its own.</summary>
     private const long PartitionCountTtlMs = 30_000;
 
-    /// <summary>True when a non-empty assignment is cached for the group.</summary>
+    /// <summary>
+    ///     True when an assignment is cached for the group, even one holding zero partitions. Treating an empty
+    ///     assignment as missing would re-sync on every poll of a member that owns nothing.
+    /// </summary>
     internal bool HasAssignment(GroupKey key)
```
hubcio, line 54: a zero-partition member now only picks up a rebalance on the next ping, so pickup latency is whatever `HeartbeatInterval` happens to be - and that knob says nothing about assignments. give `GroupAssignment` its own TTL like go's `assignmentRefreshInterval`.

foreign/csharp/Iggy_SDK/Consumers/IggyConsumer.cs
```diff
@@ -409,6 +416,12 @@ private async Task PollMessagesAsync(CancellationToken ct)
 
             if (messages.Messages.Count == 0)
             {
+                if (messages.PartitionId == PolledMessages.NO_ASSIGNED_PARTITION)
+                {
+                    LogNoPartitionAssignedBackingOff(NoAssignedPartitionBackoffMs);
+                    await Task.Delay(NoAssignedPartitionBackoffMs, ct);
```
hubcio, line 422: pre-existing, but this delay makes it fire on every shutdown: there is no `catch (OperationCanceledException)` rethrow here like `IggyConsumer.Rented.cs:198` has, so cancelling mid-delay gets published as a `Failed to poll messages` error. add the same rethrow.

foreign/csharp/Iggy_SDK/Contracts/PolledMessages.cs
```diff
@@ -22,10 +22,17 @@ namespace Apache.Iggy.Contracts;
 /// </summary>
 public sealed class PolledMessages
 {
+    /// <summary>
+    ///     Partition id an empty group poll reports when the member currently owns no partition of the group,
+    ///     for example mid-rebalance or when the group has more members than partitions. Matches the Go and
+    ///     Node SDKs. Consumers should back off before polling again rather than spin.
+    /// </summary>
+    public const uint NO_ASSIGNED_PARTITION = 0xFFFF_FFFE;
```
hubcio: this is the only `public const` in the SDK and the only public SCREAMING_SNAKE name - everything else in that style is internal. rename to `NoAssignedPartition` before release, it is breaking after.

hubcio (changes requested):
went through all 21 comments from the last round against the code - every one is genuinely fixed, and the unit suite is green locally (1028 tests across net8.0 and net10.0, 0 build warnings). the `MapHeaders` rewrite onto shared `ReadHeaderKind` / `ReadHeaderField` and the `ArrayBufferWriter` permissions pass came out cleaner than what I asked for.

two things live outside the diff so I could not anchor them:

`Iggy_SDK_Tests/VsrTests/SyncConsumerGroupTests.cs` - nothing covers the sentinel this fix is built on. neither the `PollGroupMessagesRentedAsync` return nor the consumer backoff has a test, while go covers both in `tcp_group_polling_test.go:262` and `:279`. the assert that actually guards the `HasAssignment` change is that a second poll issues no new `SYNC_CONSUMER_GROUP`.

`Iggy_SDK/IggyClient/IIggyConsumer.cs:45` and `:57` - `PollMessagesAsync` and `PollMessagesRentedAsync` are public and do not say `PartitionId` can come back as `NO_ASSIGNED_PARTITION` with an empty batch. go states it in the interface contract (`contracts/client.go:128`). anyone using the raw poll instead of `IggyConsumer` would key offsets on partition 4294967294.

foreign/csharp/Iggy_SDK/Vsr/ConsumerGroupClientState.cs
lukaszzborek: added TTL. 
I also found that rust have similar behaviour like this (only resync during heartbeat). It would be good to verify thit

hubcio (approved):

mmodzelewski (approved):

PR #4031, opened 2026-09-01
refactor(server): split bootstrap.rs into the boot/ tree

hubcio:
bootstrap.rs had grown to 5.1k lines holding every boot concern at
once: config and credential loading, listener start-up, partition
recovery, shard thread handles, the metadata handoff and the
cluster topology. None of it could be read or tested in isolation.

Move each concern into its own boot/ module and keep the shard_main
narrative, with its ordering invariants, whole in boot/mod.rs. Fold
three duplications the split exposed. The five-channel stop
fan-out was copied between the listener-failure and normal
shutdown paths; StopSignals owns it now. The rustls crypto provider
was installed at four sites; bootstrap installs it once before any
shard thread exists. The cluster roster carried the config-declared
ports while the HTTP state patched in the bound one, so the two
disagreed under port-0 binds; shard 0 now publishes the bound ports
into the one roster, and current_config.toml gains the bound HTTP
address so a harness can discover an ephemeral HTTP port like the
other transports.

New integration tests pin the iggy-view header on success, denial,
ping and follower relay, and bound-port discovery under port-0
binds.


spetz (dismissed):

core/server/src/boot/listeners.rs
```diff
@@ -0,0 +1,631 @@
...
+) -> Result<(), ServerError> {
+    // HTTP is served over TCP but sits outside the replica_io / manual client
+    // reactor, so it binds on its own: first, so its port is known when the
+    // roster and `current_config.toml` are written, and served last, so no
+    // request is answered before they are. Shard-0 gating comes from the sole
+    // caller of this function.
+    let http = topology
+        .http_listen_addr
+        .map(|addr| http::bind(addr, &config.http, &config.cluster))
+        .transpose()?;
```
numinnex, line 81: Binding here opens the socket, but the serve loop only starts at line 144. Everything in between runs with the HTTP port `listen()`ing and nothing accepting, including the replica peer dial: `connect_one` (`core/message_bus/src/connector.rs:151`) is a bare `TcpStream::connect().await` with no timeout, joined across peers.

Against a peer that DROPs SYN rather than refusing (a security group with no matching ingress rule, a node mid-replacement) that blocks for the kernel SYN-retry budget, ~127s at the Linux default `tcp_syn_retries=6`. The kernel completes the handshake into the backlog, so a TCP-level readiness probe passes and traffic routes to a node with no HTTP serve loop. On master the socket was still unbound at this point and the probe got `ECONNREFUSED`.

The comment above is right that the port has to be known before the roster and `current_config.toml` are written, but that only requires binding before line 141, not before the client binds.

`bind()` is already ordered validation-then-socket (JWT, forward state, CORS, metrics endpoint, then `client_listener::tcp::bind` at `http.rs:145`), so splitting it keeps both properties: leave the validation here, so a bad `[http.cors]` or `[http.jwt]` still fails before any listener accepts, and move only the socket bind down to just above the `publish` at line 140.

hubcio: done in 4bdf28de6 - `http::bind` is now `http::prepare` (validation, still first) plus `PreparedHttp::bind` (socket only, right above the publish), so the port opens after the peer dial returns.

spetz (approved):

mmodzelewski (approved):

PR #4035, opened 2026-09-02
feat(helm): deploy a cluster with one release per node

numinnex:
The chart predates server-ng and only ever described a single node. Cluster
mode had no representation in it at all, and `server.replicaCount: 3` looked
like the way to ask for one: it produced three independent servers behind one
Service, sharing a single PVC subpath with no lock between them, corrupting the
data directory while `helm --wait` reported success.

Cluster mode now has a home. `server.cluster` mirrors the server's
`[[cluster.nodes]]` config one for one and renders into `IGGY_CLUSTER_*`
environment variables plus `--replica-id`, so no config file is mounted. Each
node is its own release sharing the same roster and overriding only
`selfReplicaId`, with `examples/cluster-3-node.yaml` as a starting point.
Roster mistakes fail at render instead of after the pod is scheduled.

Two constraints come from the server rather than from taste, and both are
documented. The consensus listener binds `cluster.nodes[*].ip` verbatim, while
client listeners keep their own bind address and take only the port, so a pod
must own the address its roster entry names: that means `hostNetwork` with each
release pinned to a node. Host ports in turn make a rolling update impossible,
since the replacement pod cannot bind ports the outgoing pod still holds, so the
server switches to `Recreate` whenever `hostNetwork` is set.

`server.replicaCount > 1` and `autoscaling.enabled` are now refused at render
time. This leaves `hpa.yaml` unreachable, since an HPA can only target the
Deployment that is refused; the template is left in place for a follow-up to
remove. The validate scenarios that exercised autoscaling were dropped, which
also drops the `autoscaling/v2beta2` API-selection coverage along with the
feature that needed it, and scenarios asserting each guard now take their place.

Along the way: QUIC was published as TCP and so was unreachable through the
Service, WebSocket had no port at all, Kubernetes service links were injecting
`IGGY_*` variables into the server's own config namespace, and a stray suffix
landed on whichever UI label rendered last. The default image moves off 0.7.0,
which predates server-ng, so the CI smoke path exercises the current server for
the first time.

Trim the last two paragraphs if you want it shorter; the first three carry the substance.


helm/charts/iggy/examples/cluster-3-node.yaml
```diff
@@ -0,0 +1,106 @@
...
+# hostNetwork it must be the node's own IP.
+#
+# Every secret below has to be byte-identical on every node, so they live in one
+# Secret that all three releases reference rather than in this file. Create it
+# once, in the release namespace, before the first install:
+#
+#   kubectl create secret generic iggy-cluster-secrets \
+#     --from-literal=clusterSharedSecret="$(head -c 32 /dev/urandom | base64)" \
+#     --from-literal=jwtEncodingSecret="$(head -c 32 /dev/urandom | base64)" \
+#     --from-literal=jwtDecodingSecret="$(head -c 32 /dev/urandom | base64)" \
```
hubcio: these two `head -c 32 /dev/urandom` calls give the jwt keys different values, so every token from `/users/login` fails with 401 - the server only warns about the mismatch. use one value for both, as the readme does.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -105,6 +145,12 @@ spec:
             - name: IGGY_NODE_ADVERTISED_ADDRESS
               value: {{ .Values.server.advertisedAddress | default (printf "%s.%s.svc.cluster.local" (include "iggy.fullname" .) .Release.Namespace) | quote }}
```
hubcio, line 161: in cluster mode the server ignores `IGGY_NODE_ADVERTISED_ADDRESS` and publishes the roster `ip` or `advertisedAddress`, so `server.advertisedAddress` silently does nothing there. skip this env when `cluster.enabled` and fail the render if `server.advertisedAddress` is set with it.

helm/charts/iggy/templates/service.yaml
```diff
@@ -32,11 +32,21 @@ spec:
...
     - name: tcp
       port: {{ .Values.server.ports.tcp }}
       targetPort: tcp
       protocol: TCP
+    - name: websocket
+      port: {{ .Values.server.ports.websocket }}
+      targetPort: websocket
+      protocol: TCP
+  {{- if .Values.server.cluster.enabled }}
+    - name: tcp-replica
```
hubcio: nothing dials this port through the service - replicas connect straight to the roster `ip:tcpReplica`. drop it, or a `LoadBalancer` service exposes the consensus port with auth off by default.

helm/charts/iggy/examples/cluster-3-node.yaml
```diff
@@ -0,0 +1,106 @@
...
+#     --from-literal=encryptionKey="$(head -c 32 /dev/urandom | base64)"
+#
+# The JWT encoding and decoding secrets are an HMAC key pair, so they hold the
+# same value; generate them once and reuse it if you set them by hand.
+
+ui:
+  enabled: false
+
+server:
```
hubcio, line 52: persistence is off here, so each node runs on `emptyDir` and every pod replacement wipes the replica, which then refetches everything from its peers. set `persistence.enabled: true` with a node-local storage class, since the pod is pinned.

helm/charts/iggy/README.md.gotmpl
```diff
@@ -101,6 +101,142 @@ If Prometheus Operator is installed and you want monitoring, set
...
+Turning `hostNetwork` on for a release that already exists is rejected by the
+API server, because a Deployment cannot move from `RollingUpdate` to `Recreate`
+in place:
+
+```text
+Deployment.apps "iggy-n0" is invalid: spec.strategy.rollingUpdate: Forbidden:
+  may not be specified when strategy `type` is 'Recreate'
+```
+
+Uninstall that release and install it again. Set `server.strategy` explicitly if
```
hubcio: a deployment can switch to `Recreate` in place - this only fails under helm 4 server-side apply, which cannot drop the server-defaulted `rollingUpdate`. `helm upgrade --server-side=false` does the switch fine on kind, so document that instead of uninstall.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -55,16 +77,34 @@ spec:
...
               protocol: TCP
+            - name: websocket
+              containerPort: {{ .Values.server.ports.websocket }}
+              protocol: TCP
             - name: quic
               containerPort: {{ .Values.server.ports.quic }}
+              protocol: UDP
+  {{- if .Values.server.cluster.enabled }}
+            - name: tcp-replica
+              containerPort: {{ .Values.server.ports.tcpReplica }}
```
hubcio: in cluster mode these ports come from `server.ports`, not from this node's roster entry, so a node with other roster ports gets wrong container ports, probes and service targets. take them from the `selfReplicaId` entry instead.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -19,6 +19,16 @@
   {{- if hasKey .Values.server "podSecurityContext" }}
     {{- fail "server.podSecurityContext has been moved to podSecurityContext (root level). Please update your values." }}
   {{- end }}
+  {{- if gt (int .Values.server.replicaCount) 1 }}
+    {{- fail "server.replicaCount > 1 is refused: the replicas share one PVC at one subPath and one advertised address, so they would run as independent servers writing the same data directory, and the server takes no lock against that. Run cluster mode instead: one release per node, each with server.cluster.enabled and its own server.cluster.selfReplicaId. See the Cluster Mode section of the chart README." }}
+  {{- end }}
+  {{- if .Values.autoscaling.enabled }}
+    {{- fail "autoscaling.enabled is refused: horizontally scaling the server Deployment produces the same shared-data-directory corruption as server.replicaCount > 1. Cluster size is a roster decision, not an HPA one." }}
```
hubcio: with this guard `hpa.yaml` can only render for a ui-only install, where it targets a server deployment that does not exist. delete it, the `autoscaling` values and the matching assert in `test-helm.sh`.

helm/charts/iggy/values.yaml
```diff
@@ -100,6 +105,107 @@ server:
...
+  hostNetwork: false
+
+  # -- Deployment update strategy. Empty lets the chart choose: `Recreate` when
+  # `hostNetwork` is set, because a rolling update would wait forever for a
+  # replacement pod that cannot bind host ports the outgoing pod still holds,
+  # and the Kubernetes default otherwise.
+  strategy: {}
+
+  # -- Extra command-line arguments appended to the server entrypoint, e.g.
+  # `["--replica-id", "0"]` when running a cluster node.
```
hubcio: the chart already passes `--replica-id` when `cluster.enabled`, so this example makes clap exit with `cannot be used multiple times`, and outside cluster mode a non-zero id is refused anyway. use a different example flag.

helm/charts/iggy/examples/cluster-3-node.yaml
```diff
@@ -0,0 +1,106 @@
...
+#
+# Replace the `ip` values with the IPs of the nodes you pin each release to.
+# The server binds this address verbatim for replica traffic, so with
+# hostNetwork it must be the node's own IP.
+#
+# Every secret below has to be byte-identical on every node, so they live in one
+# Secret that all three releases reference rather than in this file. Create it
+# once, in the release namespace, before the first install:
+#
+#   kubectl create secret generic iggy-cluster-secrets \
```
hubcio, line 37: root credentials must match on every release as well - each node creates root locally from its own `IGGY_ROOT_*` env, so a different password on one release only works on that node. add them to this secret or say so.

helm/charts/iggy/templates/_helpers.tpl
```diff
@@ -97,3 +97,182 @@ Create the name of the service account to use
...
 {{- end }}
+
+{{/*
+Validate the cluster roster and fail the render with an actionable message.
+Every check here is one the server would otherwise only reject at boot, after
+the pod is already scheduled.
+*/}}
+{{- define "iggy.validateCluster" -}}
+  {{- $cluster := .Values.server.cluster }}
+  {{- if not $cluster.nodes }}
```
hubcio, line 108: a fresh cluster boot refuses to start without both `IGGY_ROOT_USERNAME` and `IGGY_ROOT_PASSWORD`, so `users.root.createSecret: false` with no `existingSecret` renders fine and dies at boot. fail here when neither is set.

helm/charts/iggy/templates/_helpers.tpl
```diff
@@ -97,3 +97,182 @@ Create the name of the service account to use
...
+    {{- fail "server.cluster.enabled is true but server.cluster.nodes is empty. Every node runs the identical roster; see the Cluster Mode section of the chart README." }}
+  {{- end }}
+  {{- $count := len $cluster.nodes }}
+  {{- $seen := dict }}
+  {{- range $cluster.nodes }}
+    {{- if not .name }}
+      {{- fail "every server.cluster.nodes entry needs a name" }}
+    {{- end }}
+    {{- if not .ip }}
```
hubcio, line 124: only emptiness is checked, so a hostname renders and the server dies at boot parsing `hostname:9090`. reject anything outside `[0-9a-fA-F.:]` and the wildcards `0.0.0.0` and `::`, keeping ipv6 allowed.

helm/charts/iggy/templates/_helpers.tpl
```diff
@@ -97,3 +97,182 @@ Create the name of the service account to use
...
+{{/*
+Validate the secret configuration. The server enforces all of this at boot, so
+catching it during render only saves a scheduling round trip, but a cluster
+whose PSK differs between nodes fails as a handshake rejection rather than as
+anything that names the cause.
+*/}}
+{{- define "iggy.validateSecrets" -}}
+  {{- $server := .Values.server }}
+  {{- if $server.encryption.enabled }}
+    {{- if and (not $server.encryption.key) (not $server.encryption.existingSecret.name) }}
```
hubcio, line 255: an inline key is not checked here, so a bad value renders and the server fails with `Invalid encryption key`. add `ne (len (b64dec $key)) 32` - sprig `b64dec` returns an error string on bad input, so length catches both.

helm/charts/iggy/templates/_helpers.tpl
```diff
@@ -97,3 +97,182 @@ Create the name of the service account to use
...
+    {{- end }}
+  {{- end }}
+  {{- if $server.cluster.auth.enabled }}
+    {{- if not $server.cluster.enabled }}
+      {{- fail "server.cluster.auth.enabled is true but server.cluster.enabled is false. Replica authentication only applies to the consensus port, which exists in cluster mode." }}
+    {{- end }}
+    {{- if and (not $server.cluster.auth.sharedSecret) (not $server.cluster.auth.existingSecret.name) }}
+      {{- fail "server.cluster.auth.enabled is true but no shared secret was given. Set server.cluster.auth.sharedSecret, or point server.cluster.auth.existingSecret.name at a Secret holding one. Every node needs the byte-identical value." }}
+    {{- end }}
+    {{- if and $server.cluster.auth.sharedSecret (lt (len $server.cluster.auth.sharedSecret) 32) }}
```
hubcio, line 271: a `previousSharedSecret` equal to `sharedSecret`, or shorter than 32 bytes, renders fine and the server rejects it at boot. mirror both checks here.

scripts/ci/test-helm.sh
```diff
@@ -266,11 +259,57 @@ validate() {
   fi
   grep -q 'name: supersecret' "$HELM_RENDER_DIR/existing-secret.yaml"
 
+  helm template iggy "$CHART_DIR" \
```
hubcio, line 263: this only renders the `existingSecret` path, so the chart-generated secret in `server-secrets.yaml` is never asserted. add a render with an inline `server.encryption.key` and check the secret and its `secretKeyRef`.

helm/charts/iggy/values.yaml
```diff
@@ -32,14 +32,19 @@ server:
     # -- Image pull policy
     pullPolicy: Always
     # -- Server image tag (overrides chart appVersion)
-    tag: "0.7.0"
+    tag: "0.9.0-edge.6"
```
hubcio: this duplicates `appVersion`, which the template already falls back to. if you drop it, `test-helm.sh` needs to read `Chart.yaml` instead, since it extracts this tag.

helm/charts/iggy/Chart.yaml
```diff
@@ -20,8 +20,8 @@ apiVersion: v2
 name: iggy
 description: A Helm chart for Apache Iggy server and web-ui
 type: application
-version: 0.5.0
-appVersion: "0.7.0"
+version: 0.6.0
+appVersion: "0.9.0-edge.6"
```
hubcio, line 24: `0.9.0-edge.6` was built from 996ac04c8, before `node.advertised_address` landed in #3923, so this image logs `IGGY_NODE_ADVERTISED_ADDRESS` as an unknown variable and the advertised-address handling is inert. pin the next edge tag or say so in the readme.

helm/charts/iggy/templates/_helpers.tpl
```diff
@@ -97,3 +97,182 @@ Create the name of the service account to use
...
+    secretKeyRef:
+      name: {{ default $generated $server.encryption.existingSecret.name }}
+      key: {{ ternary $server.encryption.existingSecret.key "encryptionKey" (ne $server.encryption.existingSecret.name "") }}
+  {{- end }}
+  {{- if or $server.jwt.existingSecret.name $server.jwt.encodingSecret }}
+- name: IGGY_HTTP_JWT_ENCODING_SECRET
+  valueFrom:
+    secretKeyRef:
+      name: {{ default $generated $server.jwt.existingSecret.name }}
+      key: {{ ternary $server.jwt.existingSecret.encodingSecretKey "jwtEncodingSecret" (ne $server.jwt.existingSecret.name "") }}
```
hubcio, line 306: with `existingSecret.name` set both refs are required, so a secret holding only one jwt key leaves the pod in `CreateContainerConfigError`, though the server mirrors a single secret itself. make the decoding ref `optional: true`.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -19,6 +19,16 @@
...
+  {{- end }}
+  {{- if .Values.server.cluster.enabled }}
```
hubcio, line 25: cluster mode renders fine with `hostNetwork: false` even though the values and readme call it required, and the pod then dies at boot with `CannotBindToSocket`. fail here unless an explicit opt-out is set.

helm/charts/iggy/templates/server-secrets.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+{{- if and .Values.server.enabled (include "iggy.createsSecret" .) }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "iggy.secretName" . }}
+  labels:
+    {{- include "iggy.labels" . | nindent 4 }}
+type: Opaque
+stringData:
```
hubcio, line 26: changing any of these through `helm upgrade` never restarts the pod, since the pod template only holds `secretKeyRef`s - the old secret stays live until a manual rollout restart. add a checksum annotation over this secret to the pod template.

hubcio (changes requested):
README.md.gotmpl line 377 still says the chart pins `0.7.0` by default.

pre-existing, not from this PR: `ingress.yaml` and `NOTES.txt` use `server.service.port` while the service publishes `server.ports.http`, and several cluster releases with ingress enabled all render `chart-example.local`.

helm/charts/iggy/README.md.gotmpl
```diff
@@ -101,6 +101,153 @@ If Prometheus Operator is installed and you want monitoring, set
...
+replica on `emptyDir` and forces exactly that on every pod replacement, so the
+cluster example enables persistence and leaves `storageClass` for you to point
+at a node-local provisioner.
+
+### What the chart refuses
+
+`server.replicaCount > 1` fails at render time. Scaling the server Deployment
+produces N independent servers behind one Service, all writing the same PVC
+subpath with no lock between them, which corrupts the data directory while
+`helm --wait` still reports success. The chart ships no HorizontalPodAutoscaler
```
hubcio, line 250: `templates/hpa.yaml` is still in the tree and its `.Values.autoscaling.enabled` now dereferences nil, so nothing renders and `helm lint` fails - that is the red validate job. `git rm` the template, nothing else references autoscaling.

helm/charts/iggy/templates/deployment.yaml
```diff
@@ -27,17 +35,31 @@ metadata:
...
   {{- end }}
   selector:
     matchLabels:
       {{- include "iggy.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-  {{- with .Values.podAnnotations }}
+  {{- $secretsChecksum := "" }}
+  {{- if (include "iggy.createsSecret" .) }}
+    {{- $secretsChecksum = (include (print $.Template.BasePath "/server-secrets.yaml") . | sha256sum) }}
```
hubcio, line 53: this hashes `server-secrets.yaml` only, so a root password change through `helm upgrade` does not roll the pod, and root is rebuilt from `IGGY_ROOT_*` on every boot - the old password stays valid until someone restarts it by hand. hash `root-user-credentials.yaml` too.

helm/charts/iggy/templates/_helpers.tpl
```diff
@@ -97,3 +97,234 @@ Create the name of the service account to use
...
+  {{- $count := len $cluster.nodes }}
+  {{- $seen := dict }}
+  {{- range $cluster.nodes }}
+    {{- if not .name }}
+      {{- fail "every server.cluster.nodes entry needs a name" }}
+    {{- end }}
+    {{- if not .ip }}
+      {{- fail (printf "server.cluster.nodes entry %q has no ip. The replica listener binds this address verbatim, so it must be a literal IP the pod owns." .name) }}
+    {{- end }}
+    {{- if not (regexMatch "^[0-9a-fA-F.:]+$" .ip) }}
```
hubcio: a numeric-looking ip (`--set 'server.cluster.nodes[1].ip=2001'` or an unquoted `ip: 2001`) arrives as a number and this aborts with `wrong type for value; expected string; got int64`. use `(toString .ip)` here and in the two `eq` checks below.

helm/charts/iggy/README.md.gotmpl
```diff
@@ -101,6 +101,153 @@ If Prometheus Operator is installed and you want monitoring, set
...
+  --from-literal=username=iggy \
+  --from-literal=password="$(head -c 24 /dev/urandom | base64)" \
+  --from-literal=clusterSharedSecret="$(head -c 32 /dev/urandom | base64)" \
+  --from-literal=jwtEncodingSecret="$JWT" \
+  --from-literal=jwtDecodingSecret="$JWT" \
+  --from-literal=encryptionKey="$(head -c 32 /dev/urandom | base64)"
+```
+
+`examples/cluster-3-node.yaml` points `server.users.root`, `server.encryption`,
+`server.jwt` and `server.cluster.auth` at that one Secret. The last three also
```
hubcio: "the last three" is off - `server.users.root` takes inline `username`/`password` too, and the chart turns them into `<release>-root-credentials`. say all four, then rerun helm-docs.

helm/charts/iggy/README.md.gotmpl
```diff
@@ -101,6 +101,153 @@ If Prometheus Operator is installed and you want monitoring, set
...
+from `RollingUpdate` to `Recreate`. Under Helm 4's server-side apply that is
+rejected, because the patch cannot drop the `rollingUpdate` block the API server
+defaulted in:
+
+```text
+Deployment.apps "iggy-n0" is invalid: spec.strategy.rollingUpdate: Forbidden:
+  may not be specified when strategy `type` is 'Recreate'
+```
+
+Run that one upgrade with `helm upgrade --server-side=false`, which replaces the
```
hubcio: `--server-side` only exists in helm 4, and the prerequisites at the top still say helm 3.2.0+. say "on helm 4" here so a helm 3 reader does not hit an unknown flag.

helm/charts/iggy/values.yaml
```diff
@@ -31,15 +32,20 @@ server:
...
     # -- HTTP API port
     http: 3000
-    # -- QUIC protocol port
+    # -- QUIC protocol port (UDP)
     quic: 8080
     # -- TCP protocol port
     tcp: 8090
+    # -- WebSocket protocol port
+    websocket: 8092
+    # -- Replica-to-replica consensus port. Only published when
```
hubcio: this key never wins - the roster entry's `tcpReplica` always overrides it (`--set server.ports.tcpReplica=1234` renders identically), and the service port it was "published" on is gone. drop it or say the roster owns the port.

hubcio (changes requested):

hubcio (approved):

krishvishal (approved):

PR #4036, opened 2026-09-02
refactor(server): make dispatch routing and failure paths explicit

hubcio:
The TCP dispatch funnel hid its rules in control flow: ordered
if-return probes where the order was the only spec, inline rewrite
chains duplicated across two places, send sites each picking a
failure channel by whichever builder they called, duplicate shard-0
client-request queues, and a read gate that let unknown codes fall
through to a catch-all.

Each rule is now written down once: classify() returns a RequestClass
that the handler matches on, rewrite.rs holds both chains as pure
functions, and dispatch/failure.rs holds the failure channel table
plus the single exit every host-built frame takes. Read paths that
answered a permanent client error with a fail-open shape every SDK
decodes as success now deny typed instead. Dropping the shard
reference cycle exposed a shutdown race on the metadata write handle,
so peers count themselves out and shard 0 fences on that, floored at
one drain budget because the wait is a safety invariant rather than a
latency bound.

### DIFF
--- a/core/integration/tests/server/legacy_login_vsr.rs
+++ b/core/integration/tests/server/legacy_login_vsr.rs
@@ -27,19 +27,15 @@
 //! TCP socket: a header-only non-replicated frame carrying the code in the
 //! reserved command slot.
 
-use iggy_binary_protocol::HEADER_SIZE;
+use iggy_binary_protocol::EvictionReason;
 use iggy_binary_protocol::codes::{LOGIN_USER_CODE, LOGIN_WITH_PERSONAL_ACCESS_TOKEN_CODE};
-use iggy_binary_protocol::consensus::{Command, Operation, RequestHeader};
+use iggy_binary_protocol::consensus::Command;
 use integration::harness::TestHarness;
 use integration::iggy_harness;
-use std::mem::offset_of;
-use std::time::Duration;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::TcpStream;
-use tokio::time::timeout;
 
-// Wire byte pinned to `EvictionReason::MalformedLogin` in consensus::header.
-const EVICTION_REASON_MALFORMED_LOGIN: u8 = 15;
+use crate::server::raw_tcp::{
+    connect, eviction_reason, frame_command, non_replicated_header, read_frame_header, write_frame,
+};
 
 #[iggy_harness]
 async fn given_legacy_login_user_code_when_sent_raw_should_evict_malformed_login(
@@ -60,45 +56,26 @@ async fn given_legacy_pat_login_code_when_sent_raw_should_evict_malformed_login(
 /// eviction. The reject runs before the session gate, so this unbound socket
 /// exercises the same path a bound connection would.
 async fn assert_legacy_login_code_evicted(harness: &TestHarness, code: u32) {
-    let mut header = RequestHeader {
-        command: Command::Request,
-        operation: Operation::NonReplicated,
-        size: u32::try_from(HEADER_SIZE).unwrap(),
-        // NonReplicated leaves session / request unchecked, but the header
-        // validator still requires a nonzero client id.
-        client: 0xC0FFEE,
-        session: 0,
-        request: 0,
-        ..Default::default()
-    };
-    // A non-replicated command code travels in the first 4 reserved bytes.
-    header.reserved[..4].copy_from_slice(&code.to_le_bytes());
+    // NonReplicated leaves session / request unchecked, but the header
+    // validator still requires a nonzero client id.
+    let header = non_replicated_header(0xC0FFEE, 0, 0, code);
 
-    let addr = harness
-        .server()
-        .tcp_addr()
-        .expect("server must expose a TCP address");
-    let mut stream = TcpStream::connect(addr).await.unwrap();
-    stream.write_all(bytemuck::bytes_of(&header)).await.unwrap();
+    let mut stream = connect(harness).await;
+    write_frame(&mut stream, &header, &[]).await;
 
-    // Eviction is header-only: exactly 256 bytes. The timeout makes the
+    // Eviction is header-only: exactly 256 bytes. The bounded read makes the
     // fail-fast contract explicit -- a regression that silently drops the
     // frame trips this instead of hanging until the test wall clock.
-    let mut reply = [0u8; HEADER_SIZE];
-    timeout(Duration::from_secs(5), stream.read_exact(&mut reply))
-        .await
-        .expect("server must answer a legacy login code within 5s, not stall")
-        .expect("reading the eviction frame must succeed");
+    let reply = read_frame_header(&mut stream).await;
 
-    let command_offset = offset_of!(RequestHeader, command);
     assert_eq!(
-        reply[command_offset],
+        frame_command(&reply),
         Command::Eviction as u8,
         "expected an Eviction frame for legacy login code {code}, not a Reply"
     );
     assert_eq!(
-        reply[HEADER_SIZE - 1],
-        EVICTION_REASON_MALFORMED_LOGIN,
+        eviction_reason(&reply),
+        EvictionReason::MalformedLogin as u8,
         "legacy login code {code} must evict with MalformedLogin"
     );
 }
--- a/core/integration/tests/server/raw_tcp.rs
+++ b/core/integration/tests/server/raw_tcp.rs
@@ -1,0 +1,207 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Raw TCP framing for the server suites that hand-craft client frames the
+//! SDK cannot emit: connect to the harness server, write one request frame,
+//! read the header the server answers with, and register root so a frame
+//! can ride a bound session.
+
+use std::mem::offset_of;
+use std::time::Duration;
+
+use iggy::prelude::*;
+use iggy_binary_protocol::codec::{WireDecode, WireEncode};
+use iggy_binary_protocol::consensus::{
+    Command, Operation, ReplyHeader, RequestHeader, read_size_field, result_code,
+    result_section_len,
+};
+use iggy_binary_protocol::requests::users::LoginRegisterRequest;
+use iggy_binary_protocol::responses::users::LoginRegisterResponse;
+use iggy_binary_protocol::{
+    ClientVersionInfo, EvictionHeader, HEADER_SIZE, IGGY_PROTOCOL_VERSION, WireName,
+};
+use integration::harness::TestHarness;
+use secrecy::SecretString;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpStream;
+use tokio::time::{Instant, sleep, timeout};
+
+/// Per-frame reply wait. A server that drops the frame answers nothing at
+/// all, so an unanswered read is a verdict, not a reason to wait longer.
+const REPLY_WAIT: Duration = Duration::from_secs(5);
+
+/// Budget for the register to commit: right after boot the single node may
+/// still be electing itself and answers transient rejections meanwhile.
+const COMMIT_BUDGET: Duration = Duration::from_secs(15);
+
+const RETRY_PAUSE: Duration = Duration::from_millis(100);
+
+pub(crate) async fn connect(harness: &TestHarness) -> TcpStream {
+    let addr = harness
+        .server()
+        .tcp_addr()
+        .expect("server must expose a TCP address");
+    TcpStream::connect(addr).await.unwrap()
+}
+
+pub(crate) fn request_header(
+    operation: Operation,
+    client: u128,
+    session: u64,
+    request: u64,
+    body_len: usize,
+) -> RequestHeader {
+    RequestHeader {
+        command: Command::Request,
+        operation,
+        size: u32::try_from(HEADER_SIZE + body_len).unwrap(),
+        client,
+        session,
+        request,
+        ..Default::default()
+    }
+}
+
+/// A header-only `NonReplicated` frame; the command code travels in the
+/// first 4 reserved bytes.
+pub(crate) fn non_replicated_header(
+    client: u128,
+    session: u64,
+    request: u64,
+    code: u32,
+) -> RequestHeader {
+    let mut header = request_header(Operation::NonReplicated, client, session, request, 0);
+    header.reserved[..4].copy_from_slice(&code.to_le_bytes());
+    header
+}
+
+pub(crate) async fn write_frame(stream: &mut TcpStream, header: &RequestHeader, body: &[u8]) {
+    stream.write_all(bytemuck::bytes_of(header)).await.unwrap();
+    if !body.is_empty() {
+        stream.write_all(body).await.unwrap();
+    }
+}
+
+/// Read the header of the next server frame, within [`REPLY_WAIT`].
+pub(crate) async fn read_frame_header(stream: &mut TcpStream) -> [u8; HEADER_SIZE] {
+    let mut header = [0u8; HEADER_SIZE];
+    timeout(REPLY_WAIT, stream.read_exact(&mut header))
+        .await
+        .expect("server must answer within the reply wait, not stall")
+        .expect("reply header read failed");
+    header
+}
+
+/// Write one frame and read one Reply off the lockstep connection, the body
+/// sized by the reply's size field.
+pub(crate) async fn exchange(
+    stream: &mut TcpStream,
+    header: &RequestHeader,
+    body: &[u8],
+) -> ([u8; HEADER_SIZE], Vec<u8>) {
+    write_frame(stream, header, body).await;
+    let reply_header = read_frame_header(stream).await;
+    let command = frame_command(&reply_header);
+    assert_eq!(
+        command,
+        Command::Reply as u8,
+        "expected a Reply frame, got command byte {command} (an Eviction carries reason {})",
+        eviction_reason(&reply_header)
+    );
+
+    let total_size = read_size_field(&reply_header).expect("reply size field") as usize;
+    // `read_size_field` is a bare 4-byte LE read with no floor (the SDK adds
+    // its own guard, which this helper cannot inherit), so an under-sized
+    // field would either panic on subtract overflow or wrap into a ~1.8e19
+    // allocation and abort - either way hiding the regression under test.
+    assert!(
+        total_size >= HEADER_SIZE,
+        "reply size field {total_size} is below the {HEADER_SIZE}-byte header"
+    );
+    let mut reply_body = vec![0u8; total_size - HEADER_SIZE];
+    timeout(REPLY_WAIT, stream.read_exact(&mut reply_body))
+        .await
+        .expect("reply body timed out")
+        .expect("reply body read failed");
+    (reply_header, reply_body)
+}
+
+pub(crate) fn frame_command(header: &[u8; HEADER_SIZE]) -> u8 {
+    header[offset_of!(RequestHeader, command)]
+}
+
+/// The typed reason byte of an `Eviction` frame, read by field offset rather
+/// than by a hardcoded index: `EvictionReason` is renumberable, so a literal
+/// would keep passing while checking a different reason.
+pub(crate) fn eviction_reason(header: &[u8; HEADER_SIZE]) -> u8 {
+    header[offset_of!(EvictionHeader, reason)]
+}
+
+pub(crate) fn reply_status(reply_header: &[u8; HEADER_SIZE]) -> u32 {
+    let offset = offset_of!(ReplyHeader, status);
+    u32::from_le_bytes(reply_header[offset..offset + 4].try_into().unwrap())
+}
+
+/// Root login/register on the socket as `client`, replayed on transient
+/// rejections until it commits; returns the bound session id. The session
+/// binds to THIS transport connection server-side, so a frame that must ride
+/// it has to reuse the stream.
+pub(crate) async fn register_root(stream: &mut TcpStream, client: u128) -> u64 {
+    let body = LoginRegisterRequest {
+        version_info: ClientVersionInfo {
+            protocol_version: IGGY_PROTOCOL_VERSION,
+            sdk_name: WireName::new("raw-tcp").unwrap(),
+            sdk_version: WireName::new("0.0.1").unwrap(),
+        },
+        username: WireName::new(DEFAULT_ROOT_USERNAME).unwrap(),
+        password: SecretString::from(DEFAULT_ROOT_PASSWORD),
+        client_context: None,
+    }
+    .to_bytes();
+    let header = request_header(Operation::Register, client, 0, 0, body.len());
+    let deadline = Instant::now() + COMMIT_BUDGET;
+    loop {
+        let (reply_header, reply_body) = exchange(stream, &header, &body).await;
+        // A pre-commit deny rides the status word; a committed verdict rides
+        // the result section that leads a Register reply body.
+        let code = match reply_status(&reply_header) {
+            0 => result_code(&reply_body).expect("register reply must carry a result section"),
+            status => status,
+        };
+        if code == 0 {
+            let payload_start = result_section_len(&reply_body).unwrap();
+            let response = LoginRegisterResponse::decode_from(&reply_body[payload_start..])
+                .expect("register payload must decode");
+            assert_ne!(response.session, 0, "server must bind a nonzero session");
+            return response.session;
+        }
+        assert!(
+            is_transient(code),
+            "register rejected with a terminal code {code}"
+        );
+        assert!(
+            Instant::now() < deadline,
+            "register did not commit within {COMMIT_BUDGET:?}"
+        );
+        sleep(RETRY_PAUSE).await;
+    }
+}
+
+fn is_transient(code: u32) -> bool {
+    code == IggyError::TransientNotCommitted.as_code()
+        || code == IggyError::TransientNotAccepted.as_code()
+}
--- a/core/server/src/boot/mod.rs
+++ b/core/server/src/boot/mod.rs
@@ -280,12 +280,20 @@ pub fn bootstrap(
     let config = Arc::new(config);
     // One post-shutdown budget for the whole process: the main thread's
     // shard joins and shard 0's peer wait both count down this instant
-    // instead of each arming a full `shutdown_join_timeout`. The drain
-    // budget floors the peer wait so a spent join budget cannot release the
-    // metadata writer while a peer still reads through it.
+    // instead of each arming a full `shutdown_join_timeout`. The floor keeps a
+    // spent join budget from releasing the metadata writer while a peer still
+    // reads through it, so it covers a peer's whole exit: one poll interval to
+    // observe the shutdown flag, then one drain budget to drain. Flooring at
+    // the drain alone is short by a poll interval, and
+    // `shutdown_join_timeout == shutdown_drain_timeout` is a legal config.
     let shutdown_deadline = Arc::new(ShutdownDeadline::new(
         config.system.sharding.shutdown_join_timeout.get_duration(),
-        config.system.sharding.shutdown_drain_timeout.get_duration(),
+        config
+            .system
+            .sharding
+            .shutdown_drain_timeout
+            .get_duration()
+            .saturating_add(config.system.sharding.shutdown_poll_interval.get_duration()),
     ));
     // One owner table per server process, Arc-cloned into every shard's bus so
     // any shard's bus reads the same atomic slots that the owning
@@ -411,7 +419,9 @@ pub fn bootstrap(
                 // that made it; the rest are peers shard 0's exit wait would
                 // otherwise sit out its whole budget for.
                 let spawned_peers = shard_threads.len().saturating_sub(1);
-                peer_exit.peers_never_spawned(shards_count.saturating_sub(1) - spawned_peers);
+                peer_exit.peers_never_spawned(
+                    shards_count.saturating_sub(1).saturating_sub(spawned_peers),
+                );
                 join_partial_shard_survivors(shard_threads, &shutdown_deadline);
                 return Err(ServerError::ShardSpawnFailed { shard_id, source });
             }
--- a/core/server/src/boot/mod.rs
+++ b/core/server/src/boot/mod.rs
@@ -47,12 +47,12 @@ use crate::boot::listeners::{
     make_replica_delegation_fns, make_shard_zero_client_accept_fns, start_tcp_runtime,
 };
 use crate::boot::recovery::{
-    RecoveredOwnerState, build_shard_for_thread, restore_metadata_consensus,
+    RecoveredOwnerState, ShardBuild, build_shard_for_thread, restore_metadata_consensus,
 };
 use crate::boot::threads::{
-    StopSignals, await_pump_drain, install_panic_hook, join_partial_shard_survivors,
-    resolve_shard_assignments, run_shard_thread, spawn_shutdown_watchdog,
-    validate_sharding_runtime_knobs,
+    PeerExitCountdown, PeerExitWait, ShutdownDeadline, StopSignals, await_pump_drain,
+    install_panic_hook, join_partial_shard_survivors, resolve_shard_assignments, run_shard_thread,
+    spawn_shutdown_watchdog, validate_sharding_runtime_knobs,
 };
 use crate::boot::topology::{RosterCells, resolve_tcp_topology};
 use crate::dispatch::partition::make_partition_read_handler;
@@ -60,8 +60,8 @@ use crate::dispatch::reads::read_frontier_budget;
 use crate::dispatch::session_ops::warm_dummy_password_hash;
 use crate::dispatch::submit::make_metadata_submit_handler;
 use crate::dispatch::{
-    make_client_request_handler, make_deferred_client_request_handler,
-    make_deferred_replica_message_handler, make_list_clients_handler,
+    make_deferred_client_request_handler, make_deferred_replica_message_handler,
+    make_list_clients_handler,
 };
 use crate::server_error::ServerError;
 use crate::session_manager::SessionManager;
@@ -278,6 +278,15 @@ pub fn bootstrap(
     // thread body or in a task compio's `spawn` would swallow, escapes it.
     let first_panic = install_panic_hook(Arc::clone(&shutdown_flag));
     let config = Arc::new(config);
+    // One post-shutdown budget for the whole process: the main thread's
+    // shard joins and shard 0's peer wait both count down this instant
+    // instead of each arming a full `shutdown_join_timeout`. The drain
+    // budget floors the peer wait so a spent join budget cannot release the
+    // metadata writer while a peer still reads through it.
+    let shutdown_deadline = Arc::new(ShutdownDeadline::new(
+        config.system.sharding.shutdown_join_timeout.get_duration(),
+        config.system.sharding.shutdown_drain_timeout.get_duration(),
+    ));
     // One owner table per server process, Arc-cloned into every shard's bus so
     // any shard's bus reads the same atomic slots that the owning
     // shard's installer / disconnect path writes.
@@ -301,6 +310,11 @@ pub fn bootstrap(
     // drains them all before binding listeners. Bounded to the peer
     // count so a sender never blocks (each peer sends exactly once).
     let (ready_tx, ready_rx) = crossfire::mpmc::bounded_async::<u16>(metadata_peers);
+
+    // Shard 0 blocks on this at exit until every peer thread is done (see
+    // `PeerExitCountdown`). Sized to the real peer count, not the clamped
+    // channel capacity above: a single-shard server must not wait at all.
+    let peer_exit = Arc::new(PeerExitCountdown::new(shards_count.saturating_sub(1)));
 
     let mut shard_threads: Vec<(u16, thread::JoinHandle<Result<(), ServerError>>)> =
         Vec::with_capacity(shards_count);
@@ -354,6 +368,8 @@ pub fn bootstrap(
         let roster_cells_for_shard = roster_cells.clone();
         let applied_frontier_for_shard = Arc::clone(&metadata_applied_frontier);
         let shard_metrics_for_shard = shard_metrics_all.clone();
+        let peer_exit_for_shard = Arc::clone(&peer_exit);
+        let shutdown_deadline_for_shard = Arc::clone(&shutdown_deadline);
         let handle = match thread::Builder::new()
             .name(format!("shard-{shard_id}"))
             .spawn(move || -> Result<(), ServerError> {
@@ -373,6 +389,8 @@ pub fn bootstrap(
                     roster_cells_for_shard,
                     applied_frontier_for_shard,
                     shard_metrics_for_shard,
+                    peer_exit_for_shard,
+                    shutdown_deadline_for_shard,
                 )
             }) {
             Ok(handle) => handle,
@@ -389,10 +407,12 @@ pub fn bootstrap(
                 drop(metadata_bundle_rx);
                 drop(ready_tx);
                 drop(ready_rx);
-                join_partial_shard_survivors(
-                    shard_threads,
-                    config.system.sharding.shutdown_join_timeout.get_duration(),
-                );
+                // Shard 0 spawns first, so the vec holds it plus every peer
+                // that made it; the rest are peers shard 0's exit wait would
+                // otherwise sit out its whole budget for.
+                let spawned_peers = shard_threads.len().saturating_sub(1);
+                peer_exit.peers_never_spawned(shards_count.saturating_sub(1) - spawned_peers);
+                join_partial_shard_survivors(shard_threads, &shutdown_deadline);
                 return Err(ServerError::ShardSpawnFailed { shard_id, source });
             }
         };
@@ -416,7 +436,7 @@ pub fn bootstrap(
     Ok(ShardHandles {
         shutdown_flag,
         shard_threads,
-        join_timeout: config.system.sharding.shutdown_join_timeout.get_duration(),
+        deadline: shutdown_deadline,
         first_panic,
     })
 }
@@ -441,6 +461,8 @@ async fn shard_main(
     roster_cells: RosterCells,
     metadata_applied_frontier: Arc<AppliedFrontier>,
     shard_metrics_all: Vec<ShardMetrics>,
+    peer_exit: Arc<PeerExitCountdown>,
+    shutdown_deadline: Arc<ShutdownDeadline>,
 ) -> Result<(), ServerError> {
     let topology = resolve_tcp_topology(config, replica_id)?;
     let bus = Rc::new(IggyMessageBus::with_config_and_owner_table(
@@ -478,7 +500,12 @@ async fn shard_main(
     // metadata VSR; per-commit `publish()` (in `WriteCell::apply`)
     // bounds reader staleness to one op.
     let data_dir = Path::new(&config.system.path);
-    let (mux_stm, owner_state) = match metadata_handoff {
+    // The bundle broadcast is deliberately NOT inside the owner arm: it is
+    // the first moment a peer can hold a read handle over shard 0's writer,
+    // so the writer must first be parked in a binding that outlives the peer
+    // wait armed below. A `recover()` failure inside the arm is safe for the
+    // same reason -- no peer holds a handle yet.
+    let (mux_stm, pending_bundle_tx, owner_state) = match metadata_handoff {
         MetadataHandoff::Owner { bundle_tx } => {
             // Root is created locally at boot (never journaled), so replay
             // must start from the same baseline or every WAL-created user
@@ -520,17 +547,9 @@ async fn shard_main(
                     }
                 }
             });
-            broadcast_metadata_bundle(
-                shard_id,
-                &bundle_tx,
-                recovered.mux_stm.factory_bundle(),
-                total_shards.saturating_sub(1),
-                &shutdown_flag_for_handoff,
-                poll_interval,
-            )
-            .await?;
             (
-                recovered.mux_stm,
+                Rc::new(recovered.mux_stm),
+                Some(bundle_tx),
                 Some(RecoveredOwnerState {
                     journal: recovered.journal,
                     snapshot: recovered.snapshot,
@@ -551,9 +570,39 @@ async fn shard_main(
                 poll_interval,
             )
             .await?;
-            (ServerMuxStateMachine::from_factory_bundle(bundle), None)
+            (
+                Rc::new(ServerMuxStateMachine::from_factory_bundle(bundle)),
+                None,
+                None,
+            )
         }
     };
+
+    // Shard 0 owns the metadata state machine's only write handle, and the
+    // peers read through it until their runtimes are gone. Declared after
+    // `mux_stm` so it drops first: every exit from here on, clean or `?`,
+    // waits for the peers before the write side goes. The `Rc` above is what
+    // makes that hold -- the handle no longer travels by value into the
+    // fallible shard build, which would drop it inside the callee.
+    let _peer_exit_wait = (shard_id == 0).then(|| {
+        PeerExitWait::new(
+            peer_exit,
+            Arc::clone(&shutdown_flag_for_handoff),
+            shutdown_deadline,
+        )
+    });
+
+    if let Some(bundle_tx) = pending_bundle_tx {
+        broadcast_metadata_bundle(
+            shard_id,
+            &bundle_tx,
+            mux_stm.factory_bundle(),
+            total_shards.saturating_sub(1),
+            &shutdown_flag_for_handoff,
+            poll_interval,
+        )
+        .await?;
+    }
 
     // Metadata consensus + journal + snapshot live only on shard 0.
     // `IggyShard::tick_metadata` short-circuits when `consensus.is_none()`,
@@ -589,7 +638,7 @@ async fn shard_main(
         journal_for_metadata,
         snapshot_for_metadata,
         superblock_for_metadata,
-        mux_stm,
+        Rc::clone(&mux_stm),
         Some(PathBuf::from(&config.system.path)),
     )
     .with_applied_frontier(metadata_applied_frontier);
@@ -625,7 +674,12 @@ async fn shard_main(
     // Heap-pin like `run_shard_thread` pins `shard_main`: the builder future
     // carries the whole shard construction state machine and outgrew clippy's
     // `large_futures` cap; one allocation per shard startup.
-    let (shard, sessions) = Box::pin(build_shard_for_thread(
+    let ShardBuild {
+        shard,
+        sessions,
+        on_client_request,
+        shard_handle,
+    } = Box::pin(build_shard_for_thread(
         shard_id,
         total_shards,
         config,
@@ -892,11 +946,15 @@ async fn shard_main(
             boot_view,
             shard.plane.metadata().client_table.borrow().client_ids(),
         );
-        let on_client_request = make_client_request_handler(
-            &shard,
-            &sessions,
-            Arc::clone(&config.system),
-            config.personal_access_token.max_tokens_per_user,
+        // The request handler strands every frame until the weak
+        // self-reference is backfilled, so the build must have done that
+        // before the first listener binds.
+        debug_assert!(
+            shard_handle
+                .borrow()
+                .as_ref()
+                .is_some_and(|weak| weak.upgrade().is_some()),
+            "shard self-reference must be backfilled before listeners bind"
         );
         let (accepted_replica, dialed_replica) =
             make_replica_delegation_fns(Rc::clone(&coord), &bus);
--- a/core/server/src/boot/threads.rs
+++ b/core/server/src/boot/threads.rs
@@ -381,7 +381,8 @@ impl ShutdownDeadline {
     }
 
     /// How long shard 0 may block on its peers: what is left of the shared
-    /// budget, floored at one drain budget so a join that already spent the
+    /// budget, floored at a peer's whole exit (one poll interval to observe the
+    /// shutdown flag plus one drain budget) so a join that already spent the
     /// budget cannot release the metadata writer while a peer still reads
     /// through it. Only [`PeerExitWait::drop`]'s panic arm skips the fence.
     fn remaining_for_peer_wait(&self) -> Duration {
@@ -994,8 +995,8 @@ mod tests {
     fn peer_exit_wait_gives_up_at_the_deadline() {
         let countdown = Arc::new(PeerExitCountdown::new(1));
         let _wedged_peer = PeerExitGuard::new(Arc::clone(&countdown));
-        // Budget equal to the floor: the `join_equal_to_drain_is_accepted`
-        // config, where the floor cannot extend the wait.
+        // Degenerate budget == floor: the floor cannot extend the wait, so
+        // the abandon is bounded by the budget alone.
         let timeout = Duration::from_millis(50);
         let started = Instant::now();
         drop(PeerExitWait::new(
--- a/core/server/src/boot/threads.rs
+++ b/core/server/src/boot/threads.rs
@@ -25,7 +25,8 @@ use crate::shard_allocator::{ShardAllocator, ShardInfo};
 use compio::runtime::ResumeUnwind;
 use configs::server::ServerConfig;
 use configs::sharding::{
-    INBOX_CAPACITY_MAX, SHUTDOWN_DRAIN_TIMEOUT_MAX, SHUTDOWN_POLL_INTERVAL_MAX,
+    INBOX_CAPACITY_MAX, RECONCILE_PERIODIC_INTERVAL_MAX, SHUTDOWN_DRAIN_TIMEOUT_MAX,
+    SHUTDOWN_JOIN_TIMEOUT_MAX, SHUTDOWN_POLL_INTERVAL_MAX,
 };
 use message_bus::{IggyMessageBus, ReplicaOwnerTable};
 use metadata::AppliedFrontier;
@@ -36,7 +37,7 @@ use shard::{Receiver as ShardReceiver, Sender, ShardFrame, TaggedSender};
 use std::backtrace::Backtrace;
 use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, OnceLock};
+use std::sync::{Arc, Condvar, Mutex, OnceLock, PoisonError};
 use std::time::{Duration, Instant};
 use std::{panic, thread};
 use tracing::{error, info, warn};
@@ -46,12 +47,12 @@ use tracing::{error, info, warn};
 /// Carries the cross-thread shutdown flag, one OS-thread `JoinHandle`
 /// per shard, and the first panic `install_panic_hook` recorded. The
 /// caller flips the flag via [`Self::install_ctrlc_handler`] and then
-/// drains every shard via [`Self::join_all`], bounded by `join_timeout`
-/// (`system.sharding.shutdown_join_timeout`).
+/// drains every shard via [`Self::join_all`], bounded by the shared
+/// `ShutdownDeadline` (`system.sharding.shutdown_join_timeout`).
 pub struct ShardHandles {
     pub(in crate::boot) shutdown_flag: Arc<AtomicBool>,
     pub(in crate::boot) shard_threads: Vec<(u16, thread::JoinHandle<Result<(), ServerError>>)>,
-    pub(in crate::boot) join_timeout: Duration,
+    pub(in crate::boot) deadline: Arc<ShutdownDeadline>,
     pub(in crate::boot) first_panic: Arc<OnceLock<String>>,
 }
 
@@ -94,7 +95,9 @@ impl ShardHandles {
     /// deadline passes is abandoned (its `JoinHandle` dropped, the OS
     /// thread left to die with the process) and reported as
     /// [`ShardJoinFailureKind::Wedged`]: a wedged pump or listener must
-    /// not block process exit forever.
+    /// not block process exit forever. Shard 0 gets the peer-wait floor
+    /// as grace on top, because its `PeerExitWait` is allowed to hold the
+    /// metadata writer that long past a spent budget.
     ///
     /// # Errors
     ///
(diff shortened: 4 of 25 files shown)


core/server/src/boot/mod.rs
```diff
@@ -627,6 +645,18 @@ async fn shard_main(
     ))
     .await?;
 
+    // The shard above owns the metadata state machine's only write handle,
+    // and the peers read through it until their runtimes are gone. Declared
```
numinnex: **warning** — this comment claims more than the code delivers: "every exit from here on, clean or `?`, waits for the peers" is false for the exits that matter.

`build_shard_for_thread` takes `metadata: ServerMetadata` **by value** (`boot/recovery.rs:75`), so the `?` at `recovery.rs:112`, `:201` and `:237` drop shard 0's only metadata `WriteCell` *inside the callee*, and a failed `.build()` at `:265` drops it inside the builder. All four happen before `_peer_exit_wait` exists. Peers are reading concurrently — their own build's first statement is `metadata.mux_stm.streams().read(..)` (`recovery.rs:84`, again `:144`), then the reconciler's first pass (`partition_reconciler.rs:336` -> `:1115`) — and hit `.expect("read handle should be accessible")` at `metadata/src/stm/mod.rs:129`. Nothing serializes the two builds: `BootstrapBarrier` is signalled after both.

Same shape one level up: `broadcast_metadata_bundle(..).await?` (`boot/mod.rs:525-533`) can `Err` after serving *some* peers (`boot/handoff.rs:153`, `:157`), and `recovered.mux_stm` is a local of that match arm, so no outer binding can drop after it.

Drop-order cannot reach a callee-owned drop, so moving this declaration up does not fix it. Fix: arm the wait right after `broadcast_metadata_bundle`, and keep the writer out of a fallible call that drops it (`Rc` the mux state machine, clone held before the guard). Note returning `metadata` in the `Err` variant does **not** work — the receiving binding is declared after the guard and drops first.

The by-value signature predates this PR; the invariant claim is new, which is what makes it worth fixing here. Consequence today is misattribution, not data loss: panics unwind (no profile sets `panic = "abort"`), so `join_all` still reports shard 0's real `Err` first, but `first_panic` (`threads.rs:205-207`) records a peer's read-handle panic instead.

core/server/src/boot/mod.rs
```diff
@@ -627,6 +645,18 @@ async fn shard_main(
...
+    // after the shard so it drops first: every exit from here on, clean or
+    // `?`, waits for the peers before the write side goes with the shard.
+    let _peer_exit_wait = (shard_id == 0).then(|| {
+        PeerExitWait::new(
+            peer_exit,
+            Arc::clone(&shutdown_flag_for_handoff),
```
numinnex: **warning** — the peer wait and the outer thread join share `shutdown_join_timeout` and **nest**, so shard 0 can be reported as the wedged shard while it is legitimately waiting.

`join_all` arms ONE shared `deadline = now + join_timeout` on the first 25ms poll that sees the shutdown flag (`boot/threads.rs:110-113`, `:236-238`). On graceful Ctrl-C the signal handler sets that flag (`threads.rs:71`) at T0, so the outer clock arms at T0+25ms independent of shard 0. Shard 0 cannot *start* this wait until after `await_pump_drain` (<= `shutdown_drain_timeout`) plus the watchdog's own `bus.shutdown(drain_timeout)` (`threads.rs:729`) — a second drain budget. So shard 0 returns at up to `2*drain + join` = 50s at defaults, against a 30s deadline.

Result: `"shard thread still running at the shutdown join deadline; abandoning it"`, a `Wedged` row, `JoinHandle` dropped while the OS thread still holds the metadata writer, and a non-zero exit on a correct shutdown. Because the deadline is shared and already blown, every shard after shard 0 in the loop is abandoned too.

Fix: bound this wait by the *remaining* join budget (thread a deadline through rather than passing the raw knob). Related: `validate_sharding_runtime_knobs` (`boot/threads.rs:496-536`) re-checks `poll <= drain` but not `join >= drain` (or `join <= MAX`), which `ShardingConfig::validate` does enforce (`configs/src/server_config/sharding.rs:251-268`) — that gap reaches the same abandon-mid-wait outcome by a second route, and `PeerExitWait` is a new consumer of `join`.

core/server/src/boot/threads.rs
```diff
@@ -333,6 +333,132 @@ impl Drop for ShutdownOnDrop {
     }
 }
 
+/// Peer shards still running: each [`PeerExitGuard`] counts one out,
+/// [`PeerExitWait`] blocks shard 0 until the count is zero.
+///
+/// Shard 0 owns the metadata state machine's only write handle and every
+/// peer reads through handles that stop working the moment it drops, so
+/// the shard that owns the writer must outlive every reader.
```
numinnex: **nit (doc)** — stated unconditionally, but `PeerExitWait::drop` skips the wait entirely on `thread::panicking()` (`:447`), and `install_panic_hook` only sets the flag without aborting (`:196-210`). So on a shard-0 panic the writer drops with peers still reading and their pumps panic at `metadata/src/stm/mod.rs:129`.

The early return is the **right** behaviour and should stay — blocking an unwinding thread on a `Condvar` for up to 30s inside `runtime.block_on` would stall the io_uring driver during a panic, and if the panic came from shard 0's pump the peers may be waiting on shard 0, so the wait would burn its whole budget or deadlock. Only the comment needs narrowing: say the invariant holds on every exit **except** the panic path, and why.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,652 @@
...
+//! | [`FrameChannel::Eviction`] | session-terminal Eviction frame with a typed reason | the client must register again: `NoSession`, `MalformedLogin`, heartbeat and login evictions |
+//! | [`FrameChannel::ResyncSentinel`] | status-0 poll reply, body carries `RESYNC_REQUIRED_PARTITION_SENTINEL` | a fenced consumer-group poll: the consumer must re-sync its assignment; HTTP mirrors it as `resync_required_polled_messages` in `crate::http::wire` |
+//! | [`FrameChannel::EmptyFrame`] | status-0 fail-fast body, empty or the 16-byte empty poll | the partition cannot answer yet; the SDK fails fast (empty poll) and retries |
+//! | [`FrameChannel::Reply`] | status-0 success frame | host-built success replies: login/register, ping, logout, non-replicated read bodies, committed metadata replies |
+//! | silent drop | no frame | deliberate only where a reply would be wrong: an undecodable header (nothing to echo), a transient consensus submit failure (the SDK read-timeout replays) |
```
numinnex: **warning** — this row is not exhaustive, and its justification is false for one of the classes it covers.

"an undecodable header (nothing to echo)" — `try_into_typed::<RequestHeader>` (`dispatch/mod.rs:415`) runs `verify_frame()` then `validate()` (`server_common/src/consensus_message.rs:310-311`), and `RequestHeader::FRAME_SEALED = false` (`binary_protocol/src/consensus/header.rs:501`) so `verify_frame` is a no-op — `validate` is the whole check. `validate_request_fields` (`header.rs:442-490`) rejects `client == 0`, `Operation::Reserved`, `Register` with `session != 0`/`request != 0`, and non-register non-`NonReplicated` with `session == 0`/`request == 0`. Every one of those **decodes cleanly**, so `build_deny_reply` has all the fields it needs — there is plenty to echo. Those clients get no frame at all and wedge to their read timeout, and the victim of the `session == 0` case is exactly the lost-session client `RequestClass::UnboundReplicated` was built to fail fast. `dispatch/partition.rs:483`'s own `request.max(1)` comment ("older and internal callers may still send zero") names callers this silently drops, and that normalization sits *after* `validate`.

Two more routes answer nothing and are not listed: `dispatch/mod.rs:663` (`build_raw_pat_reply` `Err` returns with no frame *after* the op committed — both `Err` arms look practically dead, but the row is still missing) and `shard/src/lib.rs:3400-3402` (`stage_transient_deny` sheds on a `try_send` failure and counts a frame drop).

Fix: enumerate the real set, or scope the table explicitly to `crate::dispatch`. Certifying an echoable drop as deliberate is the part that should not ship — it is what stops the next reader from checking.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,652 @@
...
+//! Every frame the dispatch host builds, success or rejection, leaves through
+//! [`send_host_frame`], so the send-failure log has one shape. Which channel a
+//! failure rides is a wire contract with the SDK:
+//!
+//! | channel | carrier | when |
+//! |---|---|---|
+//! | [`FrameChannel::TypedDeny`] | Reply, nonzero status + empty body, or a result-framed rejection body | rejections that must unblock the SDK's lockstep request slot: checksum, authz, pre-consensus rewrite, unknown or unsupported non-replicated code, unbound non-PING read, transient replay hints |
+//! | [`FrameChannel::Eviction`] | session-terminal Eviction frame with a typed reason | the client must register again: `NoSession`, `MalformedLogin`, heartbeat and login evictions |
+//! | [`FrameChannel::ResyncSentinel`] | status-0 poll reply, body carries `RESYNC_REQUIRED_PARTITION_SENTINEL` | a fenced consumer-group poll: the consumer must re-sync its assignment; HTTP mirrors it as `resync_required_polled_messages` in `crate::http::wire` |
+//! | [`FrameChannel::EmptyFrame`] | status-0 fail-fast body, empty or the 16-byte empty poll | the partition cannot answer yet; the SDK fails fast (empty poll) and retries |
```
numinnex: **warning** — "the partition cannot answer yet ... and retries" does not describe two of the cases that ride this channel. An **undecodable client body** is a permanent client error, not a transient partition condition, and there is nothing to retry: `dispatch/partition.rs:514` (poll) and `:698` (consumer-offset) both answer a malformed request with a success-shaped status-0 frame.

The taxonomy has no row for "undecodable client body answered success-shaped". Add one, or give those two sites a distinct `context` — today `:698` and the legitimate no-stored-offset success at `:757` send byte-identical `Bytes::new()` on the same channel with the same `"get_consumer_offset"` label, so nothing in the log tells them apart.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,652 @@
...
+//! | [`FrameChannel::Reply`] | status-0 success frame | host-built success replies: login/register, ping, logout, non-replicated read bodies, committed metadata replies |
+//! | silent drop | no frame | deliberate only where a reply would be wrong: an undecodable header (nothing to echo), a transient consensus submit failure (the SDK read-timeout replays) |
+//! | HTTP status | HTTP status code | the HTTP spine maps the same rejections in `crate::http::error`; it never rides these frames |
+//!
+//! The last two send nothing, so [`FrameChannel`] has no variant for them.
+//! This exit covers HOST-built frames only: the partitions engine builds and
```
numinnex: **warning** — the exemption is scoped to "produce/poll replies", but two `TypedDeny`-shaped client frames are built and sent from outside this module, so "the one send exit every host-built frame takes" holds only for frames built in this crate.

`IggyShard::deny_partition_request_transient` (`shard/src/lib.rs:3346-3358`) and `stage_transient_deny` (`:3382-3390`) both build `build_deny_reply_from_request_header(request_header, IggyError::TransientNotAccepted.as_code())` — a nonzero-status client Reply, exactly the `TypedDeny` shape at `:26` — and deliver it via `bus.send_to_client` and `LifecycleFrame::ForwardClientSend` respectively. A transient deny is neither a produce nor a poll reply.

Fix: widen the exemption to name the shard-crate partition denies, or route them through `send_host_frame`.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,652 @@
...
+        transport_client_id,
+        reply.into_generic().into_frozen(),
+        FrameChannel::EmptyFrame,
+        "empty partition reply",
+    )
+    .await;
+}
+
+// Byte snapshots pinning each channel's frame to the pre-refactor inline
+// construction. DELIBERATELY temporary: they freeze the refactor, not the
```
numinnex: **warning** — repo rule 8 says comments must never reference the current or a future task/PR, and "a later PR removes them" is an unenforceable removal trigger: it leaves ~250 lines of snapshot test with nothing a future reader can evaluate. State the invariant these snapshots pin instead, or file the removal and drop the note.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,652 @@
...
+    frame: impl Into<BusMessage>,
+    channel: FrameChannel,
+    context: &'static str,
+) {
+    if let Err(send_error) = bus.send_to_client(transport_client_id, frame).await {
+        warn!(
+            transport_client_id,
+            error = %send_error,
+            channel = ?channel,
+            context,
```
numinnex, line 105: **warning** — this PR deleted a diagnostic it had in hand. The `send_login_eviction` helper this replaced logged `reason = ?reason` on a send failure; `send_host_frame` logs only `transport_client_id`, `error`, `channel` and `context`. `send_eviction` (`:267-302`) receives `reason` and never logs it, and five call sites in `dispatch/session_ops.rs` (`:339`, `:1186`, `:1204`, `:1302`, `:1317`) all pass the same `"login rejection"` label across four distinct `EvictionReason`s — so `MalformedLogin`, `IncompatibleProtocol` and `InvalidCredentials` are indistinguishable when a send fails. Add `reason` to the eviction send-failure line.

core/server/src/dispatch/authz.rs
```diff
@@ -328,46 +261,17 @@ where
...
-        .await
-    {
-        warn!(
-            transport_client_id,
-            status,
-            error = %error,
-            "failed to surface non-replicated authz denial"
-        );
+        // No on-demand flush primitive exists, so the code stays the builder's
+        // `FeatureUnavailable` (its arm still serves the HTTP caller).
```
numinnex: **warning** — "(its arm still serves the HTTP caller)" is false. `read_local` (`http/reads.rs:104`) is the only HTTP entry to `build_non_replicated_response`, and its eleven call sites pass fixed constants (`http/handlers.rs:290`, `314`, `348`, `382`, `419`, `447`, `485`, `526`, `566`, `594`, `1722`) — `DESCRIBE_OPTIONS`, `GET_STREAM(S)`, `GET_TOPIC(S)`, `GET_USER(S)`, `GET_CONSUMER_GROUP(S)`, `GET_STATS`, `GET_PERSONAL_ACCESS_TOKENS`. No `FLUSH_UNSAVED_BUFFER`, and there are zero flush references anywhere under `core/server/src/http/` — which is why `flush_vsr.rs` is declared `test_client_transport = [Tcp]`.

So this gate arm is now the only thing producing `FeatureUnavailable` for flush (keep it), and `responses.rs:590` is dead. Drop the false clause from the comment.

core/server/src/dispatch/authz.rs
```diff
@@ -328,46 +261,17 @@ where
...
+        FLUSH_UNSAVED_BUFFER_CODE => Err(IggyError::FeatureUnavailable),
+        // A replicated code smuggled inside a `NonReplicated` header keeps the
+        // builder's `FeatureUnavailable`; a table-listed code with no arm above
+        // and an unknown code are both refused as `InvalidCommand`, never
+        // deferred to the builder's empty-ok catch-all.
```
numinnex: **nit** — with this tail refusing every armless code first, `build_non_replicated_response`'s `_ => Ok(NonReplicatedResponse::Empty)` (`responses.rs:598`) is unreachable from **both** callers: the gate refuses it on the TCP side, and HTTP passes only its eleven named codes. The empty-ok catch-all this PR set out to kill therefore survives as a fail-open trap for the next caller that lands.

`Err(IggyError::InvalidCommand)` there is provably safe: all eleven HTTP codes have named arms above the `_`, and the not-found -> 404 mapping comes from *inside* `GET_STREAM`/`GET_TOPIC`/`GET_USER`'s own arms via `response.map_or(Empty, ..)`, not from the catch-all. Sequence it with the flush arm above so flush's answer does not silently flip from `FeatureUnavailable` to `InvalidCommand`.

core/server/src/dispatch/partition.rs
```diff
@@ -509,11 +511,13 @@ pub(in crate::dispatch) async fn handle_poll_messages<B, MJ, S, SB>(
 {
     let Ok(wire) = PollMessagesRequest::decode_from(request_body(request)) else {
         // Undecodable poll: keep the fail-fast empty-poll shape.
+        let (body, channel) = empty_poll_fallback(0);
```
numinnex: **warning** — this is the last fail-open of exactly the class the PR set out to close, and it is on the hottest read path. An undecodable `POLL_MESSAGES` body answers status 0 plus a *valid* 16-byte `PolledMessages` body, which every SDK decodes as a successful 0-message poll, so a version-skewed consumer loops forever with no error — strictly worse than a wedge, which at least surfaces a timeout. Every sibling undecodable-body path in this same diff denies typed (`rewrite.rs` `static_bounds` -> `InvalidCommand`; `mod.rs:640`, whose commit is titled "deny malformed consumer-group requests instead of dropping").

No SDK breaks on the typed deny, and it is not a new wire shape: the server already sends nonzero-status `POLL_MESSAGES` replies at `:539` (authz) and `:566-577` (`PartitionNotFound`/`StreamIdNotFound`/`TopicIdNotFound`), pinned end-to-end in `integration/tests/server/poll_semantics_vsr.rs:70-73,113,145,188-192`. All five native SDKs peek `status` before any body decode — Rust `sdk/src/vsr.rs:204,234`, Go `reply.go:81`, Node `reply.ts:79`, Java `VsrResponseHandler.java:204`, .NET `VsrReplyDecoder.cs:60` — and Python/C++/PHP are FFI over Rust.

The rustdoc at `:496-497` ("empty body so the SDK fails fast on decode") argues for the current shape but contradicts its own function: `:578-580` explains the body is a valid 16-byte poll precisely so the decoder does *not* fail. And `failure.rs:562-593` never enters `handle_poll_messages`, so this change leaves that snapshot byte-identical.

Fix: `send_non_replicated_deny(shard, request, transport_client_id, IggyError::InvalidCommand.as_code())`. Scope it to this undecodable-body branch only — leave the `ConsumerGroupPartitionNotOwned` resync sentinel at `:585-593` alone.

core/server/src/dispatch/partition.rs
```diff
@@ -642,6 +695,7 @@ pub(in crate::dispatch) async fn handle_get_consumer_offset<B, MJ, S, SB>(
             request,
             transport_client_id,
             Bytes::new(),
+            FrameChannel::Reply,
```
numinnex: **nit** — same shape as the undecodable poll: a malformed request body gets a success-shaped status-0 empty frame, which the SDK reads as "no stored offset". It is also labelled `FrameChannel::Reply` with the identical `"get_consumer_offset"` context as the legitimate no-offset success at `:757`, so the two are indistinguishable in the send-failure log by every field. Deny typed here too, or at minimum give this branch its own `context`.

core/server/src/dispatch/reads.rs
```diff
@@ -154,17 +158,14 @@ pub(in crate::dispatch) async fn handle_non_replicated_request<B, MJ, S, SB>(
...
-                .send_to_client(transport_client_id, reply.into_generic().into_frozen())
-                .await
-            {
-                warn!(
-                    transport_client_id,
-                    error = %error,
-                    "failed to send non-replicated ping reply"
-                );
-            }
+            send_host_frame(
```
numinnex, line 374: **nit** — `record_heartbeat` was already called unconditionally in the funnel (`dispatch/mod.rs:457`) before `classify` routed here, so the PING arm's own call at `:153` is redundant: two `Instant::now()` and two `get_mut` per ping. The funnel call strictly covers it. Pre-existing, not introduced here.

core/server/src/dispatch/reads.rs
```diff
@@ -324,18 +333,14 @@ async fn handle_default_non_replicated<B, MJ, S, SB>(
...
-                .await
-            {
-                warn!(
-                    transport_client_id,
-                    code,
-                    error = %error,
-                    "failed to send non-replicated VSR reply"
-                );
-            }
+            send_host_frame(
```
numinnex, line 605: **warning** — the gate-deny branch at `:308-311` emits no log at all, while this builder-`Err` branch 25 lines below `warn!`s with `code` and `error`. `send_non_replicated_deny` only logs on *send* failure, so a successful refusal leaves zero server-side breadcrumb — and the refusal this PR introduces (every armless or unknown non-replicated code, via `authorize_default_read`'s new tail) is exactly the one an operator would need to see. Add a `warn!` with `code` + `error` on the gate-deny branch.

core/server/src/dispatch/mod.rs
```diff
@@ -615,45 +457,34 @@ async fn handle_client_request<B, MJ, S, SB>(
...
                 EvictionReason::MalformedLogin,
+                "legacy login rejection",
             )
             .await;
-            return;
         }
-        let allowed_pre_auth = nr_code == PING_CODE;
-        if !allowed_pre_auth && sessions.borrow().get_session(transport_client_id).is_none() {
+        RequestClass::UnauthenticatedRead => {
+            let nr_code = non_replicated_code(&header);
```
numinnex: **warning** — the rationale just below (`:488-490`, "Foreign SDKs still probe `GET_CLUSTER_METADATA` before login until they are fixed") is stale for every SDK in tree, so a genuine unauthenticated read of the private replica topology now hides at `debug` (`:491-495`).

All seven read the roster only *after* login: Node `client.socket.ts:773` (comment says "deliberately absent: the server auth-gates it"), Go `internal/util/leader_aware.go:52-88`, .NET `TcpMessageStream.Vsr.cs:489`, Java `AsyncIggyTcpClient.java:854-856` (`loginAndSettleOnLeader` composes the roster read onto the *completed* login future), and Python/C++/PHP inherit Rust. Java's poll is the most aggressive of the seven — `LEADERLESS_WAIT_BUDGET = 5s` at `LEADERLESS_POLL_INTERVAL = 250ms` (`LeaderAwareness.java:60-62`), up to 20 reads per login — but all 20 are authenticated, and this arm fires only for an unbound transport, so dropping the special case cannot cause a warn flood.

Fix: drop the special case and let it `warn!` like every other pre-auth denial.

core/server/src/dispatch/mod.rs
```diff
@@ -681,353 +512,179 @@ async fn handle_client_request<B, MJ, S, SB>(
...
+                    // and secret), so the committed reply body is empty. Substitute the
+                    // raw-token response here, on the minting client's home shard, using
+                    // the confirmed commit position from the committed reply.
+                    let reply = match build_raw_pat_reply(&request_header, reply, raw_pat_token) {
+                        Ok(reply) => reply,
+                        Err(error) => {
+                            warn!(
+                                transport_client_id,
+                                error = %error,
+                                "failed to build raw PAT reply"
```
numinnex, line 803: **nit** — this `Err` returns with no frame at all, *after* the metadata op already committed. The lockstep connection then wedges to its read timeout on an op that succeeded server-side, and `failure.rs:31` declares exactly two deliberate silences, of which this is not one.

Both `Err` arms look practically unreachable (`responses.rs:1566` already gates `command == Command::Reply` on a server-built frame, and `WireName::new(raw)` caps at 255 bytes against the fixed short token from `pat.rs:131`), so this is cheap either way: it is reachable only for a PAT mint, an admin-frequency path. Send the typed deny and add the row.

core/server/src/dispatch/mod.rs
```diff
@@ -681,353 +512,179 @@ async fn handle_client_request<B, MJ, S, SB>(
...
+                    return;
+                }
+            };
+            // Enrich consumer-group Join/Leave with the client's VSR id (+ topic
+            // partition count for Join) before replication; see `crate::consumer_group`.
+            let request = match maybe_rewrite_consumer_group_request(shard, request).await {
+                Ok(rewritten) => rewritten,
                 Err(error) => {
-                    warn!(
+                    // The rewrite only ever fails on an undecodable body, so a
```
numinnex: **nit** — worth stating what the reader needs here: the reason a replay cannot help is that both of `maybe_rewrite_consumer_group_request`'s own errors are `InvalidCommand` decode failures (`consumer_group.rs:80`, `:94`). Its third error path, `rewrite_request_body`'s `InvalidConfiguration` (`wire.rs:76-77`), needs a body past `u32::MAX` against a 64 MiB `MAX_MESSAGE_SIZE`, so it is unreachable for any client frame — the deny is correct for all three regardless. Fine as-is if you would rather not expand it; flagging only because the sentence is load-bearing for why this is terminal rather than transient.

core/server/src/dispatch/mod.rs
```diff
@@ -615,45 +457,34 @@ async fn handle_client_request<B, MJ, S, SB>(
...
-        // emits these. Reject them uniformly with a typed MalformedLogin (the
-        // SDK maps it to InvalidFormat) before the session gate, so a legacy or
-        // foreign client fails fast instead of getting the generic
-        // Unauthenticated deny the pre-auth guard would send unbound, or the
-        // silent empty-ok Reply the bound non-replicated path would send.
-        if matches!(
-            nr_code,
-            LOGIN_USER_CODE | LOGIN_WITH_PERSONAL_ACCESS_TOKEN_CODE
-        ) {
+    let bound = sessions.borrow().get_session(transport_client_id);
```
numinnex: **nit (perf)** — `get_session` is now unconditional here, where pre-PR it ran only inside the `NonReplicated` non-pre-auth branch and again after the Register/Logout returns. PING, Register and Logout each pay one extra `RefCell` borrow plus a map lookup. This is the only per-frame cost this diff adds, and it is small — PING is per-heartbeat-interval, Register/Logout are once per connection — while the diff removes more than it adds (five `Rc`/`Arc` clones now sit behind the `active.insert` guard instead of ahead of it, one fewer checked header cast, one bus clone moved to handler-build time). Net per-frame cost is negative. Not blocking; noting so it is a deliberate trade.

Two pre-existing items on the same line region, for a follow-up rather than this PR: `let header = *request.header()` at `:459` copies 256 bytes on every frame when only the `ReplicatedMetadata` arm needs the pre-`transmute_header` snapshot — bind the *reference* once here and keep the by-value copy inside that arm (do not replace it with per-arm `request.header()` calls: `RequestBacking::header` runs `bytemuck::checked::try_from_bytes` + `.expect` per call at `server_common/src/consensus_message.rs:159-163`, so N accessor calls can cost more than the one memcpy — same reason `failure.rs:168-170`, `:325-327` and `reads.rs:157-160` should bind once). And five same-key lookups happen per data-plane frame (`bus.client_meta` plus an `Rc` clone at `:451`, `ensure_connection`, `record_heartbeat` at `:457`, this `get_session`, then `get_user_id`/`read_context`) where `ensure_transport_connection` is a no-op after frame 1, so the `client_meta` lookup is dead work — one `connections` walk mirroring the existing `read_context` would collapse them.

core/server/src/dispatch/mod.rs
```diff
@@ -299,8 +230,19 @@ fn enqueue_client_request<B, MJ, S, SB>(
...
+    let sessions = Rc::clone(sessions);
+    let system_config = Arc::clone(system_config);
+    let queues = Rc::clone(queues);
+    let active = Rc::clone(active);
     bus.spawn(async move {
+        // The handle is set once the shard is built. A frame that beats
+        // it stays queued and the client's next frame drains both, so the
+        // active slot must be released here or that next frame never spawns.
+        let Some(shard) = upgrade_shard_handle(&shard_handle) else {
+            active.borrow_mut().remove(&client_id);
```
numinnex: **nit** — releasing the active slot here is right (otherwise the client's next frame finds the slot taken and nothing ever drains), but the queued frame stays in `queues[client_id]` forever if the client sends nothing more: the connection-lost hook clears neither `queues` nor `active`. Unreachable in production, since listeners bind after the backfill (`boot/mod.rs:912-920` asserts it).

Three pre-existing items share this state machine and want one fix in the connection-lost hook: a panic inside `handle_client_request` leaves `client_id` in `active` forever (compio catches the detached task's panic), so every later frame for that client queues and never drains; `pop_next_client_request` removes the map entry whenever a queue drains to empty (`:303`), so a lockstep SDK pays a `HashMap` insert plus a `VecDeque` alloc/free per request — keeping the entry is safe, `pop_next_client_request` already handles empty-but-present correctly; and the per-client `VecDeque` is unbounded with no end-to-end backpressure, because the bounded socket->dispatch channel drains at spin speed into it (`enqueue_client_request` is synchronous). Also consider `AHashMap` for `queues`/`active` and `SessionManager`'s maps — ids are server-minted so there is no HashDoS surface, `ahash` is already a direct dep, and `message_bus/src/lib.rs:702` sets the precedent (check `iter_clients` ordering first).

core/server/src/dispatch/mod.rs
```diff
@@ -366,193 +308,93 @@ fn pop_next_client_request(
...
-    if partitions_count > MAX_PARTITIONS_PER_REQUEST {
-        return Err(IggyError::TooManyPartitions);
-    }
-    Ok(())
+/// Where the funnel routes a client request. Derived by [`classify`], which
+/// IS the routing: [`handle_client_request`] matches on its result. The
+/// variant ORDER mirrors the order of the checks inside [`classify`], and
+/// that order is semantics (documented there).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RequestClass {
```
numinnex: **nit** — `RequestClass` and `classify` (`:362`) are `pub` but referenced only inside this file, its own tests included; sibling dispatch items use `pub(in crate::dispatch)` and the submodules are private `mod`. Tighten both.

core/server/src/dispatch/mod.rs
```diff
@@ -172,6 +114,12 @@ where
     })
 }
 
+/// Build the shard's one client-request handler: per-client FIFO queues
+/// drained one task per client, and the bus connection-lost hook that
+/// logs a dropped connection out. Every transport on the shard must
+/// share the instance (shard 0 hands it to its local QUIC, TCP-TLS and
+/// WSS listeners as well), or a client's ordering guarantee and the
+/// disconnect hook split by transport.
```
numinnex, line 138: **nit** — the connection-lost hook below puts `remove_connection(client_id)` as the *first* operand of the `&&` chain, so it mutates before the weak upgrade is attempted: a failed upgrade strips the `SessionManager` entry without submitting the replicated `Logout`, leaking the `ClientTable` entry and its consumer-group memberships. The `&&` chain is unchanged from pre-PR, but shard 0 only now reaches it — the deleted `make_client_request_handler` held a strong `Rc<shard>` whose upgrade could not fail. Window is pre-build / post-runtime-drop only, so effectively unreachable, and it self-heals via `recover()`'s `remove_consumer_group_member` on next boot. Cheap fix: upgrade first, then `remove_connection`.

core/server/src/dispatch/session_ops.rs
```diff
@@ -552,35 +475,18 @@ async fn evict_stale_client<B, MJ, S, SB>(
...
-        consensus::EvictionContext {
-            cluster: 0,
-            view: 0,
-            replica: 0,
-        },
-        consensus::EvictionContext::from_consensus,
-    );
-    let eviction = consensus::build_eviction_message(
-        ctx,
+    warn!(
```
numinnex, line 481: **nit** — this `warn!` now runs unconditionally *before* `send_eviction`, where pre-PR it sat in the `else` of the send result and fired only on success. The log now asserts a delivery it did not observe, and the contradicting line says only "failed to send host frame" with `context = "stale-client eviction"`.

core/server/src/dispatch/session_ops.rs
```diff
@@ -331,58 +331,28 @@ async fn surface_login_failure<B, MJ, S, SB>(
     SB: SuperblockStore + 'static,
 {
     if error.is_terminal() {
-        send_login_eviction(
+        send_eviction(
             shard,
             transport_client_id,
             request_header.client,
             eviction_reason_for(error),
+            "login rejection",
```
numinnex: **nit** — this label is passed at four more sites (`:1186`, `:1204`, `:1302`, `:1317`) covering four distinct `EvictionReason`s, so `context` cannot distinguish them; see the `send_host_frame` comment about the dropped `reason` field. More broadly, the one `context` field now carries three naming conventions — snake_case opcodes (`reads.rs:124` `"get_me"`, `partition.rs:551` `"poll_messages"`), kebab-case stages (`rewrite.rs:110` `"personal-access-token"`), prose (`failure.rs:117` `"request denial"`, this line) and hybrids (`partition.rs:1088` `"delete_segments reply"`) — which makes filtering by it impossible. Repo rule 7 wants labels matching literal API names; pick one shape.

core/server/src/rewrite.rs
```diff
@@ -0,0 +1,534 @@
...
+use metadata::impls::metadata::StreamsFrontend;
+use metadata::stm::stream::Streams;
+use server_common::Message;
+use std::cell::RefCell;
+use std::rc::Rc;
+use tracing::warn;
+
+/// A staged pre-consensus rejection: `stage` labels the chain step for the
+/// deny log line, `error` is the typed code the deny reply carries.
+pub struct RewriteDeny {
```
numinnex, line 99: **nit** — the set of pre-consensus stages is split between this struct and a call-site literal. Three live in `stage` (`:110` `"personal-access-token"`, `:123` `"user-password"`, `:127` `"static-bounds"`); the fourth is a bare `"consumer-group"` at `dispatch/mod.rs:640`, fed to the same `context` parameter by the same `send_pre_consensus_deny`. So the type does not own the set it appears to define, and a fifth stage has two plausible homes. One enum (or one const set) that `mod.rs:640` must also name would close it. (`error` is already a typed `IggyError`, so this is about cohesion, not stringly-typed errors.)

core/server/src/rewrite.rs
```diff
@@ -0,0 +1,534 @@
...
+        error,
+    })?;
+    Ok((request, raw_pat_token))
+}
+
+/// The HTTP submit's pre-consensus rewrite chain: the PAT rewrite for the
+/// already-authenticated `user_id`, then the password rewrite. Mirrors
+/// [`tcp_chain`] minus the session lookup (the HTTP listener authenticates
+/// against its own session table and resolves the acting user itself) and
+/// minus the static bounds step: HTTP enforces the same bounds in its
```
numinnex, line 165: **nit** — this sentence newly asserts cross-transport bounds parity, and the two caps it rests on are defined twice with no cross-check: `MAX_PARTITIONS_COUNT = 1000` (`common/src/lib.rs:162`, used only by the HTTP DTO validators) and `MAX_PARTITIONS_PER_REQUEST = 1000` (`binary_protocol/src/lib.rs:107`, used only on the binary path at `:182`). Equivalent today, so no live divergence — but the parity is now documented and nothing enforces it. This repo's own idiom for exactly this drift class is a `const _: () = assert!(..)` (see `boot/recovery.rs:301-350`).

core/integration/tests/server/raw_tcp.rs
```diff
@@ -0,0 +1,192 @@
...
+    let command = frame_command(&reply_header);
+    assert_eq!(
+        command,
+        Command::Reply as u8,
+        "expected a Reply frame, got command byte {command} (an Eviction carries reason {})",
+        reply_header[offset_of!(EvictionHeader, reason)]
+    );
+
+    let total_size = read_size_field(&reply_header).expect("reply size field") as usize;
+    let mut reply_body = vec![0u8; total_size - HEADER_SIZE];
```
numinnex, line 135: **warning** — `total_size - HEADER_SIZE` has no floor, so a reply whose size field is under `HEADER_SIZE` fails in a way that hides the regression under test. `read_size_field` (`binary_protocol/src/consensus/header.rs:73-78`) is a bare 4-byte LE read with no `>= HEADER_SIZE` guard — the SDK adds its own at `sdk/src/vsr.rs:181-183`, which is why this helper cannot inherit it. Debug builds panic "attempt to subtract with overflow"; release test runs have no `overflow-checks` (`[profile.release]` sets only `lto` and `codegen-units`), so it wraps to a ~1.8e19 `vec![0u8; n]` and aborts on allocation failure. Assert `total_size >= HEADER_SIZE` with a message naming the frame.

core/integration/tests/server/legacy_login_vsr.rs
```diff
@@ -60,39 +59,20 @@ async fn given_legacy_pat_login_code_when_sent_raw_should_evict_malformed_login(
...
-    timeout(Duration::from_secs(5), stream.read_exact(&mut reply))
-        .await
-        .expect("server must answer a legacy login code within 5s, not stall")
-        .expect("reading the eviction frame must succeed");
+    let reply = read_frame_header(&mut stream).await;
 
-    let command_offset = offset_of!(RequestHeader, command);
     assert_eq!(
-        reply[command_offset],
+        frame_command(&reply),
```
numinnex, line 72: **nit** — this file still reads the reason byte as `reply[HEADER_SIZE - 1]` against a hardcoded `EVICTION_REASON_MALFORMED_LOGIN: u8 = 15`, while the new harness next door reads the same field as `offset_of!(EvictionHeader, reason)` (`raw_tcp.rs:123`). Both are correct today — `header.rs:755` statically asserts `reason` is the last header byte — but renumber `EvictionReason` and the hardcoded `15` silently checks a different reason while still passing. Export an `eviction_reason()` accessor from `raw_tcp` and compare against `EvictionReason::MalformedLogin as u8`.

numinnex (changes requested):

core/server/src/boot/mod.rs
hubcio: Fixed, with one deviation from the suggested order: the wait is armed immediately **before** `broadcast_metadata_bundle`, not after it. The broadcast can fail after serving some peers, and by then those peers already hold read handles, so arming after it leaves exactly the window you describe one level up.

`IggyMetadata::mux_stm` is now `Rc<M>`, with `new` taking `impl Into<Rc<M>>` so the 21 existing call sites are untouched. `shard_main` keeps a clone in a binding declared before the wait, and the broadcast moved out of the `MetadataHandoff::Owner` arm to just after it. A `recover()` failure inside that arm needs no wait at all: no peer holds a handle yet, they are still parked in `await_metadata_bundle`.

Drop order is now shard, then `metadata` and its `Rc` clone, then the wait, then `shard_main`'s own clone. So the writer outlives every reader on each exit from the broadcast onward, the `?` returns inside the callee and a failed `.build()` included.

core/server/src/dispatch/failure.rs
hubcio: The certification is gone. The row now names the transient consensus submit failure as the one deliberate silence, and calls the `RequestHeader::validate` drop what it is: a gap, not a contract. It says why that matters too, that the fields decode so a deny could be echoed under the transport id, and that the client instead waits out its read timeout.

I did not turn it into a deny in this PR. At that point `try_into_typed::<RequestHeader>` has already failed, so only a `GenericHeader` exists: building the reply would mean reading `operation` and `request` out of a header the server just refused to interpret, and `into_routed()` has not run. That wants its own change with its own test rather than a rider here.

A `Scope:` paragraph under the table also states that it covers `crate::dispatch` only, and names the two neighbours that answer on their own paths, including `stage_transient_deny` shedding the frame when its lifecycle queue is full.

core/server/src/dispatch/failure.rs
hubcio: Taken the third way. Instead of adding a row for "undecodable client body answered success-shaped", both sites now deny typed, so nothing of that shape rides `EmptyFrame` any more. The row says so: a permanent client error never rides this channel, because there is nothing to retry.

While there, `handle_delete_segments_request` had the same shape one function down and was not in the list: an undecodable body acked status 0 with an empty body, which reads as a completed trim. It denies typed now as well, and the `Err(_) => None` arm plus the `if let Some(truncate)` wrapper collapsed into one `Err(error)` arm that carries the real code.

core/server/src/dispatch/failure.rs
hubcio: Fixed, by a different route than a `reason` field on the send-failure line: `FrameChannel::Eviction` now carries the `EvictionReason`, and `FrameChannel` grew a `Display` that renders it, so the line reads `channel=eviction(MalformedLogin)`.

That keeps one shape for the log line while making the reason part of the channel's identity, which is what the taxonomy already claims the channel is. It also means `send_host_frame` needs no eviction-specific parameter, and the five `"login_rejection"` sites become distinguishable without touching their context labels.

core/server/src/dispatch/mod.rs
hubcio: Deny sent. No row added, because with the deny in place this is no longer a silence: it rides `TypedDeny`, which the first row already covers.

The comment states the trade instead, that the op committed and only the reply could not be rendered, so the client is told the request failed and can read the minted token back. Silence on a committed op is the worse of the two.

core/server/src/dispatch/mod.rs
hubcio: Both deferred items are in this PR after all, so the trade lands the other way: the funnel prologue now costs fewer lookups than before the refactor, not one more.

`SessionManager::touch_connection` stamps the heartbeat and returns the bound session, the acting user and the peer address in ONE `connections` walk. That replaces `record_heartbeat`, `get_session`, the `Partition` arm's `get_user_id` and `reads.rs`'s `read_context`, and the `bus.client_meta` lookup now runs only on a transport's first frame, which is the only place the address has to come from the bus. `read_context` and the already-dead `connection_address` are gone.

`header` is bound as a reference; only the `ReplicatedMetadata` arm takes the by-value copy, which it needs for the pre-`transmute_header` echo on a deny. `queues`, `active` and `SessionManager`'s two maps are on `ahash` now: ids are server-minted and neither map is order-sensitive.

core/server/src/dispatch/mod.rs
hubcio: The connection-lost hook now frees both `queues` and `active` for the client, which covers the stranded queue entry and the panic-leaves-the-slot-taken case in one place. `pop_next_client_request` keeps the map entry when the queue drains empty, so a lockstep client no longer pays a `HashMap` insert plus a `VecDeque` alloc/free per request, and the hook is what removes it. Both maps moved to `ahash`.

The unbounded per-client `VecDeque` is deliberately NOT capped here. A cap needs an overflow policy, whether to shed the frame, deny typed, or evict the connection, plus an operator-facing knob to size it. Picking either silently inside a review round is worse than leaving the queue as it is. I can open a follow-up issue if you want it tracked.

core/server/src/dispatch/session_ops.rs
hubcio: Reworded rather than moved. `send_eviction` swallows the send result by design, since it is best-effort and logs its own failure, so there is no result left to branch on.

The line now claims only what it observed: the session was dropped and the `Logout` submitted, and the notice is being sent. A delivery failure still comes from `send_host_frame`, and it now carries the reason through the channel label (see the `send_host_frame` thread).

core/server/src/boot/threads.rs
```diff
@@ -333,6 +320,173 @@ impl Drop for ShutdownOnDrop {
...
+impl Drop for PeerExitWait {
+    fn drop(&mut self) {
+        // The panic is the fault to report and `ShutdownOnDrop` still
+        // flips the flag for the peers; blocking an unwinding thread here
+        // would only delay it.
+        if thread::panicking() {
+            return;
+        }
+        self.shutdown_flag.store(true, Ordering::Relaxed);
+        let remaining = self.deadline.remaining();
```
numinnex: **blocker** — this closes the nested-budget problem by making the peer-exit wait a debtor of the join budget, which reopens the writer-drop race the commit exists to close.

`remaining()` is `saturating_duration_since` (`:349-353`), so it floors at `Duration::ZERO`. `join_all` arms the shared instant at flag-flip (`:229`), and shard 0 cannot reach this drop until `bus.shutdown()` and `await_pump_drain` have finished. When the join leg has spent the budget, `countdown.wait(ZERO)` re-checks the predicate and returns `Err(running)` **without blocking** (`:381-395`), shard 0 releases, the last `Rc<ServerMuxStateMachine>` drops, and any peer inside `LeftRight::read` panics on `.expect("read handle should be accessible")` (`core/metadata/src/stm/mod.rs:129`).

Reachability: `join == drain` is legal and explicitly blessed by `core/configs/src/server_config/sharding.rs:378` (`join_equal_to_drain_is_accepted`), and at that setting one full-length drain exhausts the budget on its own. At the shipped default (join 30s / drain 10s, `core/server/config.toml:888`) zero is not reachable, but the margin still silently drops from a guaranteed 30s to a worst-case ~10s.

The distinction that makes this worth blocking on: the join budget bounds a LIVENESS concern (do not hang process exit on a wedged shard), while this wait enforces a SAFETY invariant (the writer must outlive every reader). Funding the second out of the first lets an exit-latency timeout cancel a correctness fence. The invariant doc at `:359-365` names exactly one exception (panic); deadline exhaustion is now a second one and is not named.

Note the original finding was about `join_all`'s **verdict** — shard 0 being reported `Wedged` on a correct shutdown — not about the peer wait's length. Fix: floor it (`self.deadline.remaining().max(drain_timeout)`), or give the wait its own budget and let `join_all` accept shard 0 returning within `budget + peer_budget` without a `Wedged` row.

core/server/src/boot/threads.rs
```diff
@@ -649,6 +827,188 @@ mod tests {
...
+        let countdown = Arc::new(PeerExitCountdown::new(1));
+        let _wedged_peer = PeerExitGuard::new(Arc::clone(&countdown));
+        let started = Instant::now();
+        drop(PeerExitWait::new(
+            countdown,
+            Arc::new(AtomicBool::new(false)),
+            Arc::clone(&deadline),
+        ));
+        assert!(
+            started.elapsed() < Duration::from_millis(100),
```
numinnex: **blocker (same defect, second half)** — this assertion pins the unsafe release as intended behaviour, which is why the item above cannot be resolved by a comment.

The test builds `PeerExitCountdown::new(1)` and holds `_wedged_peer`, so the peer is **never counted out**. It then spends the entire 200ms budget in `join_until_shutdown_deadline` and asserts that dropping `PeerExitWait` returns in under 100ms. That is an assertion that shard 0 drops the metadata writer while a peer is still running — a green test encoding exactly the race `PeerExitWait` was added to prevent.

The stated intent ("inherit what is left of the budget, not arm a second one") is reasonable; it is the live peer guard that turns it into a specification of the starvation. A future reader will treat this as settled and no test will catch a regression.

Fix: keep a case that pins budget inheritance with the peer already counted out, and add one that asserts the wait still HONOURS a live peer (blocks for the floor, or for its own budget).

core/server/src/dispatch/mod.rs
```diff
@@ -351,208 +308,106 @@ fn pop_next_client_request(
...
-    let mut queues = queues.borrow_mut();
-    let Some(queue) = queues.get_mut(&client_id) else {
-        active.borrow_mut().remove(&client_id);
-        return None;
-    };
-    let message = queue.pop_front();
-    if queue.is_empty() {
-        queues.remove(&client_id);
-    }
+    // The entry survives draining to empty: removing it would cost a
```
numinnex: **warning** — moving the free to the connection-lost hook makes the hook the sole owner, but the hook is not the last toucher of this map, so an entry can now leak for process lifetime.

The hook fires from the TRANSPORT task's scopeguard (`core/message_bus/src/installer/tcp.rs:118-130`), while the separate DISPATCH task still holds buffered frames in `in_rx`. `ctx.in_tx` drops when `conn.run(ctx)` returns, and `async_channel` delivers every buffered item before `recv()` reports `Closed` — so `client_dispatch_loop` (`tcp.rs:259-289`) still calls `on_request` after the hook ran, and `enqueue_client_request` re-creates the entry via `.entry(client_id).or_default()` (`:239-243`). Nothing frees it again: the hook is once-only (`message_bus/src/lib.rs:983` guards on `remove(..).is_some()`) and `client_id` is monotonic `(shard << 112) | seq`, never reused. The two guards that could stop the loop — `aborted` (`tcp.rs:275`) and `conn_token` — are both insert-race-only, so this is not a rare interleaving; it happens whenever a frame was buffered at close.

Pre-PR there was no leak, because this branch freed the entry every drain.

Fix: restore the removal here and keep the hook's removal for the mid-flight case. That also resolves the sticky-capacity issue below, so no `shrink_to` is needed.

Separately, on the stated motive: `VecDeque::pop_front` never releases capacity, so as written a client's PEAK depth stays resident (no `shrink_to`/`shrink_to_fit` exists anywhere in `core/server` or `core/shard`). For the in-tree SDKs that is small — they are lockstep, so depth stays at the 4-slot `RawVec` minimum, ~96 B per connection — but it interacts badly with the uncapped-queue item you deferred: an unbounded transient spike becomes a permanent high-water mark. Restoring the removal makes both moot.

core/server/src/dispatch/partition.rs
```diff
@@ -725,13 +781,15 @@ pub(in crate::dispatch) async fn handle_get_consumer_offset<B, MJ, S, SB>(
...
+        // "no offset stored" answer, so a malformed request must not send
+        // one. Byte-identical frames on the same channel with the same
+        // context would also be indistinguishable in the send-failure log.
+        send_non_replicated_deny(
             shard,
             request,
             transport_client_id,
-            Bytes::new(),
-            "get_consumer_offset",
+            IggyError::InvalidCommand.as_code(),
```
numinnex, line 792: **warning** — the undecodable-body fail-open is closed here, but the not-found one nine lines below is still open, and it is the only finding in this round reachable from a **well-formed request through a public SDK method**.

`:827` denies typed only on `IggyError::PartitionNotFound`; `StreamIdNotFound` and `TopicIdNotFound` from `resolve_partition_namespace` (`core/server/src/responses.rs:466-476`) fall through to the generic `Err` arm at `:836` and answer `Bytes::new()` — which every SDK decodes as "no offset stored":

- Rust `get_consumer_offset(..) -> Result<Option<ConsumerOffsetInfo>, IggyError>` returns `Ok(None)`, minted at `core/common/src/traits/binary_impls/consumer_offsets.rs:85`
- Go `GetConsumerOffset(..)` returns `(nil, nil)` (`binary_serialization/binary_response_deserializer.go:41-43`)
- C# `GetOffsetAsync(..)` returns `null` (`Iggy_SDK/IggyClient/Implementations/TcpMessageStream.cs:386-389`)
- Java `Optional.empty()`, Node `null`

So `get_consumer_offset` against a typo'd or deleted stream is indistinguishable from a fresh consumer: it resumes from its configured default and silently reprocesses, with no error anywhere. Two contrasts make it hard to defend: the SAME typo through `poll_messages` returns `Err(StreamIdNotFound)` (`:654-659` denies all three variants), and the SAME read over REST 404s (`core/server/src/http/handlers.rs:1312`). One lookup, three answers.

Fix: extend the `:827` match to `StreamIdNotFound | TopicIdNotFound`, mirroring `:654-659`. Error-path only, and cheaper than today since it drops the `Bytes::new()` + `send_non_replicated_bytes` for a header-only deny.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,677 @@
...
+//!
+//! | channel | carrier | when |
+//! |---|---|---|
+//! | [`FrameChannel::TypedDeny`] | Reply, nonzero status + empty body, or a result-framed rejection body | rejections that must unblock the SDK's lockstep request slot: checksum, authz, pre-consensus rewrite, unknown or unsupported non-replicated code, unbound non-PING read, transient replay hints |
+//! | [`FrameChannel::Eviction`] | session-terminal Eviction frame with a typed reason | the client must register again: `NoSession`, `MalformedLogin`, heartbeat and login evictions. The reason rides the channel label, since one `context` covers four of them |
+//! | [`FrameChannel::ResyncSentinel`] | status-0 poll reply, body carries `RESYNC_REQUIRED_PARTITION_SENTINEL` | a fenced consumer-group poll: the consumer must re-sync its assignment; HTTP mirrors it as `resync_required_polled_messages` in `crate::http::wire` |
+//! | [`FrameChannel::EmptyFrame`] | status-0 fail-fast body, empty or the 16-byte empty poll | the partition cannot answer yet; the SDK fails fast (empty poll) and retries. A permanent client error never rides this channel: an undecodable body denies typed, because there is nothing to retry |
```
numinnex: **warning** — "A permanent client error never rides this channel" is false as written, and it is a rule this commit authored.

A consumer-group poll that omits the partition id decodes cleanly (`core/binary_protocol/src/requests/messages/poll_messages.rs:95` maps `partition_flag != 1` to `None`), then `resolve_poll_request` returns `IggyError::InvalidIdentifier` (`partition.rs:946`). That variant is absent from the deny set at `partition.rs:654-659`, so it falls to `empty_poll_fallback(0)` (`:754`) and rides `EmptyFrame` as a status-0 16-byte empty poll — a permanent client error answered as a successful 0-message read, which is the exact symptom your commit message describes.

Reachability is foreign/hand-rolled clients only (every in-tree SDK resolves the partition client-side, e.g. `core/common/src/traits/binary_impls/messages.rs:317-320`) — but that is the same reachability class as the three fail-opens this commit closed, so it cannot be the reason to leave the sentence absolute.

Fix: add `InvalidIdentifier` to the `:654-659` match (one pattern, error path, strictly cheaper than the current `Bytes` allocation), or drop the absolute claim from this row.

core/server/src/dispatch/partition.rs
```diff
@@ -581,8 +583,10 @@ async fn relay_partition_reply<B, MJ, S, SB>(
 /// the owning shard ([`shard::IggyShard::partition_read`]), and re-encode
 /// the stored batches into the legacy wire `PolledMessages` body.
 ///
-/// Failures reply with an empty body so the SDK fails fast on decode
-/// instead of hanging until its read timeout.
+/// A partition that cannot answer yet replies with the 16-byte empty poll,
+/// which the SDK reads as 0 messages and retries. Permanent client errors
+/// (undecodable body, authz, unresolved target) deny with a nonzero status
```
numinnex: **warning** — this sentence, added by this commit, says permanent client errors including an "unresolved target" deny with a nonzero status. `IggyError::InvalidIdentifier` IS an unresolved target and does not: it is missing from the deny match at `:654-659` and falls to `empty_poll_fallback(0)`. Same defect as the `failure.rs:29` row. Fix the match or narrow the sentence.

core/server/src/dispatch/partition.rs
```diff
@@ -695,18 +677,92 @@ pub(in crate::dispatch) async fn handle_poll_messages<B, MJ, S, SB>(
...
+    let channel = if partition_id == RESYNC_REQUIRED_PARTITION_SENTINEL {
+        FrameChannel::ResyncSentinel
+    } else {
+        FrameChannel::EmptyFrame
+    };
+    (empty_polled_messages_body(partition_id), channel)
+}
+
+/// Serve `get_consumer_offset`. An empty status-0 body decodes as `None` on
+/// the SDK side (no offset stored / partition unknown), so it is reserved for
```
numinnex: **warning** — this docstring, rewritten by this commit, is contradicted by its own function body. It reserves the empty status-0 body for "no offset stored / partition unknown", but `:827` denies `PartitionNotFound` typed, so partition-unknown never takes that path. Meanwhile the case that DOES take it — an unresolved stream or topic — is not mentioned.

Fix the `:827` match (see the comment there) and the sentence becomes true as written. If the match is deferred instead, this needs to say plainly that an unresolved stream or topic reads as "no offset stored", with an issue link — a documented gap is defensible, this sentence is not.

core/server/src/dispatch/mod.rs
```diff
@@ -188,50 +137,46 @@ where
...
-            .borrow_mut()
-            .remove_connection(client_id)
-            && let Some(shard) = upgrade_shard_handle(&shard_handle_for_disconnect)
+        // The socket is gone: nothing will drain what is still queued and no
+        // later frame will release the active slot, so both entries go here.
+        // This is also the only recovery from a panic inside the drain task
+        // (compio catches it), which would otherwise leave the slot taken and
+        // every later frame for the client queued forever.
+        queues_for_disconnect.borrow_mut().remove(&client_id);
+        active_for_disconnect.borrow_mut().remove(&client_id);
```
numinnex: **nit** — clearing `active` from the hook breaks the one-drain-per-client invariant this handler's doc promises at `:118-123`. The hook can fire while drain #1 is suspended at an `.await` inside `handle_client_request`, so the post-hook `enqueue_client_request` (see the `:311` comment for why one can still arrive) wins `active.insert` at `:244` and spawns drain #2 concurrently. Two drains then pop the same `VecDeque`.

Impact is limited and I want to be accurate about it: because this closure is synchronous with no await between here and `remove_connection` at `:161`, the session is already stripped, and the hook fires FROM `remove_client_meta`, so `ensure_transport_connection` finds nothing and `classify(header, false)` routes every replicated op to `UnboundReplicated`. Drain #2 therefore commits nothing — it only emits `Eviction(NoSession)` frames at a dead socket. No op loss, no reordering.

Fix: release the slot from a scopeguard inside the spawned drain task instead. That still covers the panic case this line was written for, without racing the hook.

core/server/src/dispatch/mod.rs
```diff
@@ -681,353 +532,193 @@ async fn handle_client_request<B, MJ, S, SB>(
...
+                            warn!(
+                                transport_client_id,
+                                error = %error,
+                                "failed to build raw PAT reply"
+                            );
+                            // The op COMMITTED; only the reply could not be
+                            // rendered. A typed deny is still the right frame:
+                            // silence wedges the lockstep connection on a
+                            // request that succeeded server-side, and the
+                            // client can read the minted token back.
```
numinnex: **nit** — "the client can read the minted token back" is false, so the comment justifies the right frame with a recovery path that does not exist. `PersonalAccessTokenResponse` carries `name` + `expiry_at` only (`core/binary_protocol/src/responses/personal_access_tokens/get_personal_access_tokens.rs:30-33`, `core/common/src/types/permissions/personal_access_token.rs:35-40`); the raw secret exists in that one reply and nowhere else. A retry hits `ClientTable` dedup and replays the empty committed body, so it is unrecoverable and the caller must delete and re-create.

The deny is still correct — silence on a committed op is worse. Just drop the last clause.

core/server/src/dispatch/partition.rs
```diff
@@ -1176,8 +1186,7 @@ pub(in crate::dispatch) async fn handle_delete_segments_request<B, MJ, S, SB>(
 /// truncate commits under. A resolvable namespace with nothing sealed to delete
 /// still yields a `TruncatePartition(up_to_offset = 0)` so the metadata request
 /// sequence stays contiguous. `Err` on a malformed body or an unresolved
-/// namespace: the TCP caller drops it to a silent replay, the HTTP caller renders
-/// the error.
+/// namespace: the TCP caller denies typed, the HTTP caller renders the error.
```
numinnex: **nit** — two stale rustdoc sentences around the lines this commit edited.

Here: "`Err` on a malformed body or an unresolved namespace" — an unresolved namespace returns `Ok(build_truncate_partition_client_message_with_identifiers(..))` at `:1223-1230`. The only `Err` values are `InvalidCommand` and `TransientNotAccepted`.

And at `:1075-1076`: "Only a malformed / unresolvable request is acked empty without a commit" — now false on both halves, since malformed denies typed at `:1143` and unresolvable returns `Ok`.

core/server/src/dispatch/mod.rs
```diff
@@ -188,50 +137,46 @@ where
...
+        // every later frame for the client queued forever.
+        queues_for_disconnect.borrow_mut().remove(&client_id);
+        active_for_disconnect.borrow_mut().remove(&client_id);
+        // Upgrade FIRST: `remove_connection` strips the `SessionManager`
+        // entry, so running it ahead of a failed upgrade would drop the
+        // binding without ever submitting the replicated `Logout`, leaking
+        // the `ClientTable` entry and its consumer-group memberships. The
+        // window is pre-build / post-runtime-drop only.
+        if let Some(shard) = upgrade_shard_handle(&shard_handle_for_disconnect)
+            && let Some((vsr_client_id, session)) = sessions_for_disconnect
```
numinnex: **nit** — upgrading before `remove_connection` is the right trade (a leaked local `SessionManager` entry beats a leaked replicated `ClientTable` entry), but the failure is now silent: if the upgrade fails, `connections` and `client_to_connection` keep their entries and nothing says so. Worth noting the heartbeat verifier does not always reap it — it is gated on `config.heartbeat.enabled` (`boot/mod.rs:806`) and `collect_stale` only considers `Bound`/`Authenticated` (`session_manager.rs:196-199`), so a `Connected` leak is never reaped. Window is pre-build / post-runtime-drop only, so this is about observability, not incidence. Fix: `error!` on the upgrade-failed branch.

core/server/src/boot/threads.rs
```diff
@@ -406,6 +560,22 @@ pub(in crate::boot) fn validate_sharding_runtime_knobs(
             drain: drain_timeout,
         });
     }
+    let join_timeout = sharding.shutdown_join_timeout.get_duration();
```
numinnex, line 605: **nit** — the restored `join >= drain` and `join <= MAX` checks match `ShardingConfig::validate` exactly (`core/configs/src/server_config/sharding.rs:251-267`), but the doc at `:519-522` claims this fn "Mirrors `ShardingConfig::validate`" and one clause is still missing: `reconcile_periodic_interval`, which `validate` rejects at zero and above `RECONCILE_PERIODIC_INTERVAL_MAX` (`sharding.rs:269-287`).

That one is not cosmetic for a caller that bypasses `validate`: zero feeds `run_reconciler`'s `ctx.shard.bus.sleep(periodic)` inside an unconditional loop (`core/server/src/partition_reconciler.rs:338-355`), so the sleep arm is ready every iteration and `reconcile_once` runs back-to-back, starving the pump on that core. Add the two checks, or narrow the doc to the knobs it actually mirrors.

core/server/src/dispatch/authz.rs
```diff
@@ -328,46 +261,19 @@ where
...
-    {
-        warn!(
-            transport_client_id,
-            status,
-            error = %error,
-            "failed to surface non-replicated authz denial"
-        );
+        // No on-demand flush primitive exists. This arm is the only thing
+        // answering `FeatureUnavailable` for flush now: the HTTP path reaches
+        // the builder through `read_local`, whose call sites pass eleven fixed
```
numinnex: **nit** — the corrected rationale is true today but rests on a cross-module count nothing enforces: "whose call sites pass eleven fixed read codes, none of them flush". I counted eleven `read_local` sites too (`core/server/src/http/handlers.rs:286,310,344,378,415,443,481,522,562,590,1718`), so it is accurate — but a twelfth site silently falsifies it, and the reader has no link to check. Prefer stating the local requirement (flush has no HTTP route, so this arm is the only source of `FeatureUnavailable` for it) without the number.

core/server/src/dispatch/failure.rs
```diff
@@ -0,0 +1,677 @@
...
+    )
+    .await;
+}
+
+/// Ack a consumer-offset op whose body could not be rewritten for the
+/// partition plane with an empty Reply. The SDK connection processes replies
+/// in lockstep, so a silent drop wedges every subsequent request on that
+/// connection.
+#[allow(clippy::future_not_send)]
+pub(in crate::dispatch) async fn send_empty_partition_reply<B, MJ, S, SB>(
```
numinnex: **nit** — `send_empty_partition_reply` appears to have no reachable producer. Its only caller is the `Err` arm of `maybe_rewrite_consumer_offset_request` (`partition.rs:472-483`), but `resolve_partition_request_namespace` already decodes the same bytes with the same `decode_from` earlier at `partition.rs:377` and denies typed at `:396-405` (`core/server/src/responses.rs:361-384`), so an undecodable body never reaches the rewrite. The second decode is on unmutated bytes, and `rewrite_request_body`'s only other error needs a body past `u32::MAX` against a 64 MiB cap.

If that holds, the `EmptyFrame` row's "empty" carrier documents a shape nothing produces. Worth confirming and then either deleting this helper or noting why it is kept.

core/server/src/boot/threads.rs
```diff
@@ -649,6 +827,188 @@ mod tests {
...
+        // budget). With a budget each, a clean shutdown reported shard 0 as
+        // wedged and exited non-zero.
+        let deadline = Arc::new(ShutdownDeadline::new(Duration::from_millis(200)));
+        let shutdown_flag = AtomicBool::new(true);
+        // Never finishes: stands in for the slow drain that arms and then
+        // spends the shared budget. The thread leaks into the test process.
+        let wedged_shard = thread::spawn(|| -> Result<(), ServerError> {
+            loop {
+                thread::sleep(Duration::from_secs(1));
+            }
```
numinnex: **nit** — this test spawns `thread::spawn(|| loop { thread::sleep(1s) })` with no stop path, so the thread leaks for the life of the test binary (the comment acknowledges it). Parked in `sleep`, so it costs stack rather than CPU. Note `:1143`'s pre-existing `abandons_a_shard_still_running_past_the_deadline` does the same thing, so this copied an existing pattern — worth an `AtomicBool` stop flag in both or neither, not just the new one.

core/server/src/session_manager.rs
```diff
@@ -26,13 +26,30 @@
...
 use std::rc::Rc;
 use std::time::{Duration, Instant};
 
+/// What the request funnel resolves from one `connections` lookup per frame.
+///
+/// The bound consensus session, the acting user, and the transport peer
+/// address. `Default` (everything absent, no address) stands for a
+/// connection neither this map nor the bus knows.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct ConnectionContext {
```
numinnex: **nit** — `ConnectionContext` is bare `pub` in `pub mod session_manager` (`core/server/src/lib.rs:49`), in the same commit that tightened `RequestClass`/`classify` to `pub(in crate::dispatch)`. Harmless (`server` is `publish = false`) but inconsistent. `pub(crate)` fits.

(`RewriteStage` looks like the same issue but is not — its `pub` is forced by `pub struct RewriteDeny { pub stage: RewriteStage }` in a `pub(crate)` module, so narrowing it trips `private_interfaces`.)

core/server/src/dispatch/mod.rs
```diff
@@ -299,8 +245,19 @@ fn enqueue_client_request<B, MJ, S, SB>(
         return;
     }
 
-    let bus = shard.bus.clone();
+    let shard_handle = Rc::clone(shard_handle);
+    let sessions = Rc::clone(sessions);
+    let system_config = Arc::clone(system_config);
```
numinnex, line 281: **nit, pre-existing, follow-up material** — recording the measurement next to the queue-cap item you deferred, since they belong in the same change.

A lockstep client pays one `compio::runtime::spawn` task allocation plus 4 `Rc::clone` and 1 `Arc::clone` (atomic RMW) per request, because `drain_client_requests` returns whenever the queue empties (`:290-303`) so `active.insert` wins again on the next frame and respawns. The boxed future holds ~512 B of headers live across the consensus await in the `ReplicatedMetadata` arm, so it is ~1 KB+ — an order of magnitude larger than the `VecDeque` allocation the `:311` change was made to avoid.

Not introduced here and not worth acting on alone. But it is the reason restoring the `:311` removal is the right trade: the allocation it reintroduces is the smaller half of a pair this path already pays. Folding `active` into the queue entry would recover both, and drop 2 of the 4 per-frame hash lookups.

numinnex (changes requested):

spetz (approved):

numinnex (approved):

PR #4046, opened 2026-09-03
fix(metadata): roll deleted partitions and topics out of parent stats

numinnex:
Update the `Stats` from metadata layer when partitions get removed.


hubcio:
please add integration test (either enhance current one if it exists or create new one)

core/server/src/partition_reconciler.rs
```diff
@@ -928,10 +926,21 @@ async fn tear_down_owned_partition(
...
+    // outlive the partition, and a retention tick subtracting from counters the
+    // apply had already zeroed. A frame already past the gate can still land
+    // after this settle, which is why the rollback clamps rather than trusting
+    // the amount it is handed.
+    let live_stats = partitions.with_partition(&ns, |partition| Arc::clone(&partition.stats));
     if !partitions.is_tombstoned(&ns) {
         partitions.tombstone(ns);
     }
     shards_table.remove(&ns);
+    settle_partition_stats(ctx, ns, live_stats.as_ref());
```
hubcio: warning: a pump handler suspended mid-append resumes after this settle, and its increment lands in the parent totals with nothing left to roll it back. settle off the removed partition in the `ConfirmRemove` arm instead.

core/server/src/partition_reconciler.rs
```diff
@@ -1137,18 +1146,68 @@ fn current_revision(ctx: &ReconcilerCtx) -> u64 {
...
+    // Registry cloned out rather than mutated inside the read closure: the
+    // rollback cascades into parent totals, which the STM read has no part in.
+    let registry = ctx
+        .shard
+        .plane
+        .metadata()
+        .mux_stm
+        .streams()
+        .read(|inner| Arc::clone(&inner.stats_registry));
+    registry.remove_partition(ns.stream_id(), ns.topic_id(), ns.partition_id());
```
hubcio: warning: evicting the whole entry drops `purged_generation`, so the rebuild mints 0 and a still-pending purge then wipes everything appended since. seed the gate from the committed `Partition::purge_generation` - keeping the entry would break the recreate test.

core/server/src/boot/mod.rs
```diff
@@ -552,9 +552,7 @@ async fn shard_main(
             // and decrement the parent `StreamStats` by it.
             let () = recovered.mux_stm.streams().read(|inner| {
```
hubcio: warning: this runs after the wal replay, so a replayed `DeleteTopic` over a torn checkpoint clamps against the stream and trips the new underflow counter on every such boot. zero the aggregates before replay.

core/server/src/http/metrics.rs
```diff
@@ -45,6 +45,16 @@ pub(in crate::http) struct HttpMetrics {
     pub(in crate::http) messages: Gauge,
     pub(in crate::http) users: Gauge,
     pub(in crate::http) clients: Gauge,
+    /// Not a legacy-parity metric: the running count of rollup decrements that
```
hubcio, line 48: warning: "alert on any increase" isn't actionable - a retention decrement against a settled partition trips it, and the new unit test drives exactly that case. restate as a bounded rate outside delete, purge, teardown and restore windows.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,86 @@ use std::sync::{
...
+use tracing::warn;
+
+/// Number of rollup decrements that could not be covered by the counter they
+/// were subtracted from.
+///
+/// A decrement bigger than the total it targets means the tree lost
+/// `parent >= sum(children)` somewhere upstream. The counters are unsigned, so
+/// without a clamp that subtraction wraps to ~1.8e19 and every reader
+/// (`get_stream`, `get_topic`, `/stats`, `/metrics`) serves the wrapped value
+/// until the process restarts. Clamping keeps the total merely low, which the
```
hubcio: warning: "the next write corrects" is wrong - increments are `fetch_add`, so they stack on the clamped base and the shortfall sticks until a rebuild or restart. this is what tells an operator whether the alarm needs a response.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,86 @@ use std::sync::{
...
+}
+
+fn report_rollup_underflow(scope: &'static str, counter: &'static str, shortfall: u64) {
+    let total = ROLLUP_UNDERFLOWS.fetch_add(1, Ordering::Relaxed) + 1;
+    // One line for the first, then at powers of two. A skewed tree emits up to
+    // three of these per counter per rollback, and a bulk delete would turn a
+    // diagnostic into a log flood that buries the first one. The counter above
+    // stays exact, and it is what an alert reads.
+    if total.is_power_of_two() {
+        warn!(
```
hubcio: warning: no stream, topic or partition id here, so an operator alerting on the counter can't locate the divergence. return the shortfall and log where the ids are known - retention and purge are the likeliest source.

core/metadata/src/stm/stream.rs
```diff
@@ -384,10 +389,29 @@ impl Stream {
...
+///
+/// * `WriteCell::apply` is `append(cmd).publish()`, so a batch is one op. Batch
+///   two and a delete could be absorbed on the second buffer AFTER a create
+///   that re-minted the same id, and the delete would then evict the new
+///   partition's entry.
+/// * `left_right` drains the previous batch's `absorb_second` before the
+///   current batch's `absorb_first` (0.11.8 `write.rs`), which is an
+///   implementation detail, not a contract.
+///
+/// Independently of both, `fetch_partition_stats` refuses to register a
```
hubcio: warning: this doesn't close the live route - the check reads the left-right read side, but the eviction happens in `absorb_first` before `publish()`. a reader in that window recreates the entry the apply just dropped.

core/metadata/src/stm/stream.rs
```diff
@@ -2360,6 +2553,10 @@ impl StreamsInner {
         // that slot next.
         registry.retain_from_snapshot(&snapshot);
         *self = Self::inner_from_snapshot(snapshot, registry);
+        // Last, over the totals `inner_from_snapshot` just stored: the donor's
+        // aggregates describe the donor's partitions, and this node kept its
+        // own the whole time.
+        self.stats_registry.rebuild_parent_totals(&self.items);
```
hubcio, line 2590: warning: `retain_from_snapshot` keeps entries by slab key with no identity check, so behind a donor that recycled a key the receiver keeps the dead occupant's stats and this rebuild makes them authoritative. compare `created_revision` per key first.

core/server/src/http/metrics.rs
```diff
@@ -45,6 +45,16 @@ pub(in crate::http) struct HttpMetrics {
...
+    /// had to be clamped at zero. Non-zero means a topic or stream total was
+    /// asked to give back more than it held, so the aggregate it now reports is
+    /// low. Alert on any increase.
+    ///
+    /// A `Counter`, not a `Gauge`: it only ever climbs within a process, so
+    /// `rate()` and `increase()` are the queries an operator wants, and both are
+    /// counter-only. The scrape handler samples the source and advances this by
+    /// the difference, which is why the source is monotonic per process.
+    stats_rollup_underflows: Counter,
```
hubcio, line 71: warning: this only advances inside the `/metrics` scrape, which needs both `http.enabled` and `http.metrics.enabled`. a tcp-only or quic-only deployment gets no counter at all, leaving one throttled warn line as the whole signal.

core/server/src/boot/mod.rs
```diff
@@ -1032,6 +1030,27 @@ async fn shard_main(
 /// and dropped (the periodic tick recovers). Installed via
 /// [`metadata::IggyMetadata::set_commit_notifier`] on shard 0 only, the
 /// sole writer of the metadata state machine.
```
hubcio, line 1018: nit: `clear_snapshot_totals` landed between this doc block and the fn it describes, so both blocks now attach to the stats helper and `make_metadata_commit_notifier` ships undocumented. move it above line 1024 or below the notifier.

core/server/src/boot/mod.rs
```diff
@@ -1121,4 +1140,55 @@ mod tests {
...
+            Arc::clone(&stream_stats),
+        );
+        let topic_stats = Arc::clone(&topic.stats);
+        stream.topics.insert(topic);
+
+        // Topics summing above their stream is the torn shape: the stream was
+        // read first, and the topic kept counting before its own read.
+        stream_stats.store_from_snapshot(100, 1, 1);
+        topic_stats.store_from_snapshot(200, 2, 2);
+        let underflows_before = iggy_common::rollup_underflows();
```
hubcio: nit: `rollup_underflows()` is a process-global static, so this exact-equality assert breaks at a distance under `cargo test -p server`, where the whole crate shares one process. assert on the totals instead.

core/server/src/partition_reconciler.rs
```diff
@@ -1137,18 +1146,68 @@ fn current_revision(ctx: &ReconcilerCtx) -> u64 {
...
+/// check: the committed topic has to still list this partition, because a pass
+/// captures its targets once and then awaits disk work per namespace, so
+/// without it a target captured before a `DeletePartitions` re-creates the
+/// entry that delete just evicted -- and the apply on the second left-right
+/// buffer, deferred to a later publish, then zeroes a live partition's counters
+/// and drops its entry again.
+///
+/// `None` if the stream, the topic, or the partition vanished between the
+/// target snapshot and this read.
+fn fetch_partition_build_inputs(
```
hubcio, line 1189: nit: three references still name the old `fetch_partition_stats` - `stream.rs:411`, `responses.rs:1385`, and the dead intra-doc link at line 1107. that comment also still claims stats are fetched lazily, which this merge made false.

core/server/src/partition_reconciler.rs
```diff
@@ -1137,18 +1146,68 @@ fn current_revision(ctx: &ReconcilerCtx) -> u64 {
...
     iggy_common::TopicRuntimeOptions,
+    Partition,
 )> {
     ctx.shard.plane.metadata().mux_stm.streams().read(|inner| {
         let stream = inner.items.get(ns.stream_id())?;
         let topic = stream.topics.get(ns.topic_id())?;
+        let partition = topic
+            .partitions
+            .iter()
+            .find(|partition| partition.id == ns.partition_id())?
```
hubcio, line 1203: nit: this `find` now runs on every build, not just the disk-load branch, so materialising a k-partition topic is quadratic under the read guard. cheap next to the mkdir, but `iggy_partitions.rs:365` is the same shape.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,86 @@ use std::sync::{
...
+/// (`get_stream`, `get_topic`, `/stats`, `/metrics`) serves the wrapped value
+/// until the process restarts. Clamping keeps the total merely low, which the
+/// next write corrects; this counter plus the `warn!` below is what tells an
+/// operator the divergence happened at all.
+static ROLLUP_UNDERFLOWS: AtomicU64 = AtomicU64::new(0);
+
+/// Monotonic count of clamped rollup decrements. Surfaced on `/metrics` so the
+/// clamp is alertable rather than log-grep-able.
+#[must_use]
+pub fn rollup_underflows() -> u64 {
```
hubcio, line 48: nit: this reaches the root of the published `iggy_common` through the glob re-export, so a server-only diagnostic becomes semver-locked api. `#[doc(hidden)]` it.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,86 @@ use std::sync::{
...
+    ROLLUP_UNDERFLOWS.load(Ordering::Relaxed)
+}
+
+fn report_rollup_underflow(scope: &'static str, counter: &'static str, shortfall: u64) {
+    let total = ROLLUP_UNDERFLOWS.fetch_add(1, Ordering::Relaxed) + 1;
+    // One line for the first, then at powers of two. A skewed tree emits up to
+    // three of these per counter per rollback, and a bulk delete would turn a
+    // diagnostic into a log flood that buries the first one. The counter above
+    // stays exact, and it is what an alert reads.
+    if total.is_power_of_two() {
```
hubcio: nit: the throttle keys on one global count across all six scope/counter pairs, so a first-ever divergence in a fresh scope stays silent for up to 1023 events. key it per pair, or on a time window.

core/server/src/boot/mod.rs
```diff
@@ -1121,4 +1140,55 @@ mod tests {
...
         );
     }
+
+    /// A checkpoint loads a stream's total and its topics' as separate reads
+    /// while the partition plane keeps counting, so the two can disagree in
+    /// either direction. Boot has to drop both, or the surviving residue sits
+    /// under a total nothing on disk backs.
+    #[test]
+    fn clear_snapshot_totals_drops_both_levels_of_a_torn_checkpoint() {
+        use iggy_common::{
```
hubcio: nit: imports go at the top of the module - the sibling test above has none.

core/server/src/partition_reconciler.rs
```diff
@@ -928,10 +926,21 @@ async fn tear_down_owned_partition(
...
+    // Take the counters handle BEFORE the tombstone (every partition accessor
+    // is tombstone-gated) and settle it AFTER, once the tombstone has stopped
+    // new frames from resolving the partition. The metadata apply already
+    // rolled it out at commit time, so what is left here is whatever landed in
+    // the window between the two: in-flight appends adding to parents that
+    // outlive the partition, and a retention tick subtracting from counters the
+    // apply had already zeroed. A frame already past the gate can still land
+    // after this settle, which is why the rollback clamps rather than trusting
+    // the amount it is handed.
+    let live_stats = partitions.with_partition(&ns, |partition| Arc::clone(&partition.stats));
```
hubcio: nit: "take the counters handle BEFORE the tombstone" reads like the retry path is covered, but `with_partition` is tombstone-gated, so on retry this is always `None` and the settle is a no-op. say so.

core/server/src/partition_reconciler.rs
```diff
@@ -1137,18 +1146,68 @@ fn current_revision(ctx: &ReconcilerCtx) -> u64 {
...
+/// but a stale incarnation being torn down after a slab-key reuse still has
+/// one, and it must not survive into the rebuild: `partition` is a
+/// get-or-create, so the rebuild would inherit the dead incarnation's counters
+/// and its `purged_generation`. `live` is the mounted partition's own handle,
+/// which is the only place the post-commit residue lives. When both name the
+/// same `Arc` the second zeroing rolls back 0.
+fn settle_partition_stats(
+    ctx: &ReconcilerCtx,
+    ns: IggyNamespace,
+    live: Option<&Arc<iggy_common::PartitionStats>>,
```
hubcio: nit: `&Arc<PartitionStats>` buys nothing over `&PartitionStats` here - the body only calls `zero_out_all()`. the call site becomes `live_stats.as_deref()`.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+//! rollback and eviction, `server::partition_reconciler` for the teardown
+//! settle and the membership gate — because the integration layer cannot hold
+//! the reconciler off by configuration: the tick is capped at 30s, cannot be
+//! set to zero, and the commit wake bypasses it regardless.
+//!
+//! Verifying any of this needs `cargo build --bin iggy-server` first. The
+//! harness spawns the built binary, so a source-only mutation runs against the
+//! previous build and reports a green that means nothing.
+//!
+//! * `given_counted_partition_when_apply_delete_partitions_should_roll_it_out_of_the_parents`
```
hubcio: nit: this hand-kept list already misses three of the nine tests this commit adds (`stream.rs:3542`, `:3656`, `:3759`), and it's split from the paragraph it belongs to. name the module instead.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+//! What this guards, and what it does not. Two mechanisms roll a deleted scope
+//! out of its parents: the metadata STM evicts the registry entries at commit
+//! and zeroes them into their parents, and the reconciler's partition teardown
+//! settles whatever landed after that. The STM half runs inside the apply that
+//! produces the ack, on counters shared across every shard and both left-right
+//! buffers, so it alone satisfies every assertion here. The reconciler runs as
+//! a separate task, but a delete commit wakes it (`signal_reconcile_wake`), so
+//! it too finishes before a client can complete another round trip.
+//!
+//! Measured, not reasoned: with only the STM rollback disabled this scenario
```
hubcio, line 34: nit: the literal "7 unit tests fail" is a number nothing maintains - it goes stale on the next test added to that module. keep the conclusion, drop the count.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+//!
+//! The reads after each delete are deliberately single-shot. Retrying until the
+//! numbers converge, the way the pre-delete assertions do for the async send
+//! folding, would also wait out a broken rollback and pass regardless.
+
+use crate::server::scenarios::PARTITIONS_COUNT;
+use bytes::Bytes;
+use iggy::prelude::*;
+use integration::harness::{TestHarness, assert_clean_system, login_root};
+use std::str::FromStr;
```
hubcio: nit: `Identifier::named` is what the other scenarios use - `from_str` reinterprets a numeric-looking name as an id. no live bug with these names, just the wrong constructor.

also at the other 15 call sites in this file.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+        MSGS_SIZE * retained,
+        "the deleted partitions' bytes must leave the stream total too"
+    );
+    assert_eq!(stream.messages_count, MSGS_COUNT * retained);
+
+    client
+        .delete_stream(&Identifier::from_str(STREAM_NAME).unwrap())
+        .await
+        .unwrap();
+}
```
hubcio, line 146: nit: missing blank line before the next doc comment - every other item boundary in this file has one.

core/server/src/partition_reconciler.rs
```diff
@@ -1817,6 +1862,96 @@ mod tests {
...
+            "and the refusal must not have created its entry on the way"
+        );
+    }
+
+    /// The metadata apply rolls a deleted partition out of its parents at
+    /// commit, but the partition stays mounted until the reconciler tears it
+    /// down, and appends in that window land in parents with no entry left to
+    /// account for them. Teardown is where that residue gets settled.
+    #[compio::test]
+    async fn teardown_settles_the_residue_the_metadata_delete_could_not_reach() {
```
hubcio, line 1897: nit: this still passes with the `registry.remove_partition` at line 1174 deleted, because the metadata apply already evicted the entry in this fixture. the stale-incarnation eviction has no test.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+use iggy::prelude::*;
+use integration::harness::{TestHarness, assert_clean_system, login_root};
+use std::str::FromStr;
+use std::time::{Duration, Instant};
+use tokio::time::sleep;
+
+// Committed partition ops fold into the shared stats on the owning shard, so a
+// read can race the apply window. Retry until the expectation holds, then make
+// the terminal assertion for a real mismatch.
+const STATS_CONVERGENCE_TIMEOUT: Duration = Duration::from_secs(10);
```
hubcio: simplification: the consts, `create_messages` and the three validators are copies of `stream_size_validation_scenario.rs` - `validate_system_stats` byte for byte. hoist them into `scenarios/mod.rs`, which already owns the shared consts and helpers.

core/server/src/partition_reconciler.rs
```diff
@@ -1137,18 +1146,68 @@ fn current_revision(ctx: &ReconcilerCtx) -> u64 {
...
+/// which is the only place the post-commit residue lives. When both name the
+/// same `Arc` the second zeroing rolls back 0.
+fn settle_partition_stats(
+    ctx: &ReconcilerCtx,
+    ns: IggyNamespace,
+    live: Option<&Arc<iggy_common::PartitionStats>>,
+) {
+    // Registry cloned out rather than mutated inside the read closure: the
+    // rollback cascades into parent totals, which the STM read has no part in.
+    let registry = ctx
```
hubcio: simplification: one full `streams().read()` per torn-down namespace just to clone the registry `Arc`. build it once into `ReconcilerCtx` - it's never re-minted after boot, and a topic delete opens one reader epoch per partition.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,86 @@ use std::sync::{
...
+/// written from the metadata shard and from whichever shard owns the partition,
+/// so a separate load leaves a window where the clamp reads one value and
+/// subtracts from another.
+fn clamped_sub_u64(
+    counter: &AtomicU64,
+    amount: u64,
+    scope: &'static str,
+    name: &'static str,
+) -> u64 {
+    // The closure always yields `Some`, so `Err` carries the same previous
```
hubcio: simplification: the `unwrap_or_else` arm can't run - the closure always returns `Some`, so `fetch_update` never gives `Err`. four lines of comment justify a dead branch, twice.

core/metadata/src/stm/stream.rs
```diff
@@ -550,23 +578,109 @@ impl StatsRegistry {
     }
 
     fn remove_topic(&self, stream_id: usize, topic_id: usize) {
-        self.topics
+        // Partitions first: each one's rollback cascades through its parent
+        // topic into the stream, so zeroing the topic ahead of them would
+        // subtract the same bytes from the stream twice.
+        self.remove_partitions_where(|(sid, tid, _)| *sid == stream_id && *tid == topic_id);
```
hubcio: simplification: this still uses the predicate sweep the comment at line 608 argues against, and `remove_stream` does too. the ids are in hand at `:2234`, and every registry insert takes them from the committed topic, so keying is safe.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,86 @@ use std::sync::{
...
+}
+
+/// Subtract `amount`, clamping at zero instead of wrapping. Returns what was
+/// actually taken, which is what the caller passes on to its parent.
+///
+/// `fetch_update` rather than a load followed by a subtract: the counters are
+/// written from the metadata shard and from whichever shard owns the partition,
+/// so a separate load leaves a window where the clamp reads one value and
+/// subtracts from another.
+fn clamped_sub_u64(
```
hubcio: simplification: `clamped_sub_u64` and `clamped_sub_u32` are the same 16 lines twice, differing only in the atomic type. one `macro_rules!` covers both - a sealed trait would be longer than the duplication.

core/metadata/src/stm/stream.rs
```diff
@@ -550,23 +578,109 @@ impl StatsRegistry {
...
+    /// in the topic and stream totals. And zeroing alone leaves an entry that
+    /// outlives its ids: a topic's slab key is recycled by the next
+    /// `create_topic`, and `DeletePartitions` truncates the tail so the next
+    /// `CreatePartitions` mints the freed ids again. The survivor would hand
+    /// its successor both its counters and its `purged_generation`, and that
+    /// stale generation makes the next purge's gate skip the reset.
+    ///
+    /// # Panics
+    /// If the registry mutex is poisoned.
+    fn remove_partitions_where(&self, should_drop: impl Fn(&(usize, usize, usize)) -> bool) {
```
hubcio: simplification: generic over a predicate with exactly one caller. inline it as `remove_topic_partitions(stream_id, topic_id)` and drop the single-instantiation generic.

core/metadata/src/stm/stream.rs
```diff
@@ -550,23 +578,109 @@ impl StatsRegistry {
...
+    /// sibling's data.
+    ///
+    /// The metadata apply has usually evicted the entry already, in which case
+    /// this is a no-op and the caller's own handle carries whatever landed
+    /// after; a stale incarnation being torn down after a slab-key reuse still
+    /// has its entry, and it must not survive into the rebuild.
+    ///
+    /// # Panics
+    /// If the registry mutex is poisoned.
+    pub fn remove_partition(&self, stream_id: usize, topic_id: usize, partition_id: usize) {
```
hubcio: simplification: three-line wrapper over `remove_partitions` with one caller. call it directly with a one-element slice - `remove_partitions` has to go `pub`, so the public item count is unchanged.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+
+// The server accounts the on-disk batch framing: one command header per append
+// pass plus a per-message header. Mirrors `stream_size_validation_scenario`.
+const NG_BATCH_HEADER_SIZE: u64 = 256;
+const NG_MESSAGE_HEADER_SIZE: u64 = 48;
+const MSGS_SIZE: u64 =
+    NG_BATCH_HEADER_SIZE + (NG_MESSAGE_HEADER_SIZE + MESSAGE_PAYLOAD_SIZE_BYTES) * MSGS_COUNT;
+
+pub async fn run(harness: &TestHarness) {
+    let client = harness
```
hubcio: simplification: this re-implements `scenarios::create_client` verbatim, panic string included. import it instead.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,346 @@
...
+    validate_topic(client, STREAM_NAME, KEPT_TOPIC, MSGS_SIZE, MSGS_COUNT).await;
+    validate_system_stats(client, MSGS_SIZE, MSGS_COUNT).await;
+
+    client
+        .delete_stream(&Identifier::from_str(STREAM_NAME).unwrap())
+        .await
+        .unwrap();
+}
+
+async fn create_stream(client: &IggyClient, stream_name: &str) {
```
hubcio: simplification: one-line passthrough with two call sites. inline it.

hubcio (changes requested):
one finding sits outside the diff, so it goes here.

`core/shard/src/lib.rs:8180` - `fence_partition_for_rebuild` tombstones and enqueues `ConfirmRemove` without settling the partition counters. metadata never deleted the partition, so the registry entry survives and the rebuild's get-or-create hands back the fenced incarnation's numbers, still sitting in the parent totals. `build_partition_fresh` only zeroes on its error path, so journal repair then counts the same data twice.

the second exit at `:8160-8178`, where quarantine fails, never enqueues `ConfirmRemove` at all, so a settle in that arm would not reach it. both arms need one.

this path predates the pr, but the invariant it breaks is the one this pr establishes.

core/metadata/src/stm/stream.rs
```diff
@@ -603,7 +719,82 @@ impl StatsRegistry {
         self.partitions
             .lock()
             .expect("stats registry mutex poisoned")
-            .retain(|key, _| live_partitions.contains(key));
+            .retain(|key, entry| {
```
hubcio: critical: this eviction doesn't zero the entry's stats, and the still-mounted stale incarnation holds the same `Arc` - its later `ConfirmRemove` subtracts those bytes again out of live siblings. zero the dropped entries here, like `remove_partitions` does.

core/server/src/partition_reconciler.rs
```diff
@@ -966,6 +977,12 @@ async fn tear_down_owned_partition(
         partitions.tombstone(ns);
     }
     shards_table.remove(&ns);
+    // Registry entry only. The mounted partition's OWN counters are settled by
+    // the `ConfirmRemove` arm, which is the point where the partition value is
+    // dropped: a handler suspended mid-append resumes and increments through
+    // its cached handle, and anything settled before that drop leaves the
+    // increment in the parent totals with nothing left to roll it back.
+    settle_partition_stats(ctx, ns);
```
hubcio: warning: settle runs before the disk delete, so a failed delete returns with no `ConfirmRemove` and the still-mounted partition then reports one empty segment at offset 0 while its files remain. move it down next to the enqueue.

core/server/src/partition_reconciler.rs
```diff
@@ -694,11 +707,12 @@ async fn reconcile_additions(
...
-        // Resolve the shared stats `Arc` only for namespaces actually
-        // built, not once per committed partition every pass. A topic that
-        // vanished between the target snapshot and this read defers to the
-        // next pass.
-        let Some((partition_stats, topic_runtime)) = fetch_partition_stats(ctx, ns) else {
+        // Resolved only for namespaces actually built, not once per committed
+        // partition every pass. A stream, topic or partition that vanished
+        // between the target snapshot and this read defers to the next pass.
+        let Some((partition_stats, topic_runtime, partition_metadata)) =
+            fetch_partition_build_inputs(ctx, ns)
```
hubcio, line 723: warning: `created_revision` is in hand here but the build and row still get the snapshot `epoch`, so a mid-pass recreate lands a dead-epoch row. skip on mismatch only when the dir is absent - a bare `continue` lets the next pass adopt the old segments.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,110 @@ use std::sync::{
...
+static ROLLUP_UNDERFLOWS: AtomicU64 = AtomicU64::new(0);
+
+/// Monotonic count of clamped rollup decrements. Surfaced on `/metrics` so the
+/// clamp is alertable rather than log-grep-able.
+///
+/// Hidden from the docs: `iggy_common` re-exports this module at its root, and
+/// a server-only diagnostic has no business in the published SDK surface.
+#[doc(hidden)]
+#[must_use]
+pub fn rollup_underflows() -> u64 {
```
hubcio: warning: `doc(hidden)` hides this from rustdoc but not from the crate's public API, and the glob in `lib.rs` puts it at the `iggy_common` root with nothing in CI to catch it. rename to `stats_rollup_underflows()`, or reach it through a named `pub mod`.

core/metadata/src/stm/stream.rs
```diff
@@ -603,7 +719,82 @@ impl StatsRegistry {
...
+    /// saturating rollback absorbs it; quiescing the data plane for a metadata
+    /// install would cost far more than it buys.
+    ///
+    /// # Panics
+    /// If the registry mutex is poisoned.
+    // One acquisition for the whole walk: the stores it makes are plain atomic
+    // writes on the topic and stream `Arc`s, with no parent cascade and no way
+    // back into the map, so there is nothing for a tighter guard to protect.
+    #[allow(clippy::significant_drop_tightening)]
+    fn rebuild_parent_totals(&self, streams: &IdSlab<Stream>) {
```
hubcio, line 788: warning: this holds the `partitions` mutex across a walk of every partition on the node, and other threads take that same lock through the get-or-create. collect under the guard, drop it, then store.

core/server/src/partition_reconciler.rs
```diff
@@ -1244,47 +1261,69 @@ fn current_revision(ctx: &ReconcilerCtx) -> u64 {
...
+/// Usually a no-op: the metadata apply evicts the entry on the commit that
+/// acked the delete. What is left for this call is a stale incarnation being
+/// torn down after a slab-key reuse, whose entry the apply never named. It must
+/// not survive into the rebuild -- `StatsRegistry::partition` is a
+/// get-or-create, so the rebuild would inherit the dead incarnation's counters.
+/// (Its `purged_generation` no longer rides along either way: a fresh entry
+/// seeds that gate from the committed partition.)
+///
+/// The mounted partition's own handle is settled later, on `ConfirmRemove`.
+fn settle_partition_stats(ctx: &ReconcilerCtx, ns: IggyNamespace) {
```
hubcio, line 1325: warning: on a successful teardown this eviction leaves the partition reporting the registry-miss shape, and the parents low by its whole size, until the rebuild re-folds the disk delta a pass later. document the window, or carry a materialisation signal.

core/shard/src/lib.rs
```diff
@@ -2537,6 +2537,17 @@ where
...
+                        // parent topic and stream. Here and not in the
+                        // reconciler's teardown: a handler suspended mid-append
+                        // holds its own `Arc` past the tombstone, and an earlier
+                        // settle leaves its increment in the parents with the
+                        // partition already gone. This is the drop point, so
+                        // nothing can add through that handle afterwards. The
+                        // rollback clamps, so the usual case -- the metadata
+                        // apply already zeroed these counters at commit -- takes
+                        // nothing.
+                        partition.stats.zero_out_all();
```
hubcio, line 2550: nit: this drop point rolls its counters back, but the two discard arms above at 2469 and 2493 just `drop(partition)` and leave what the build already folded into the shared stats. both are unreachable today - fold at adoption in `IggyPartitions::insert` instead.

core/partitions/src/iggy_partition.rs
```diff
@@ -7263,6 +7260,39 @@ pub struct SegmentRemoval {
...
+    segment: &Segment,
+    messages_count: u64,
+) {
+    let size_shortfall = stats.decrement_size_bytes(segment.size.as_bytes_u64());
+    let segments_shortfall = stats.decrement_segments_count(1);
+    let messages_shortfall = stats.decrement_messages_count(messages_count);
+    if size_shortfall == 0 && segments_shortfall == 0 && messages_shortfall == 0 {
+        return;
+    }
+    warn!(
```
hubcio: nit: this warn fires per segment with no throttle, and `UnderflowSite::report` already logs the same event power-of-two throttled plus counts it. emit one line per `clean_*` call with the removed offset range.

core/metadata/src/stm/stream.rs
```diff
@@ -384,10 +389,32 @@ impl Stream {
...
+///
+/// Every eviction here therefore runs twice, once per buffer, and the second
+/// run is deferred. Two properties keep that from doing damage, and neither is
+/// guaranteed by the `Absorb` trait:
+///
+/// * `WriteCell::apply` is `append(cmd).publish()`, so a batch is one op. Batch
+///   two and a delete could be absorbed on the second buffer AFTER a create
+///   that re-minted the same id, and the delete would then evict the new
+///   partition's entry.
+/// * `left_right` drains the previous batch's `absorb_second` before the
```
hubcio, line 407: nit: the keyed evictions lean on this drain order, and `left-right = "0.11"` in the manifest floats any 0.11.x. pin `=0.11.8`, or add an `Absorb` impl that records the call sequence.

core/server/src/http/metrics.rs
```diff
@@ -45,6 +45,30 @@ pub(in crate::http) struct HttpMetrics {
     pub(in crate::http) messages: Gauge,
     pub(in crate::http) users: Gauge,
     pub(in crate::http) clients: Gauge,
+    /// Not a legacy-parity metric: the running count of rollup decrements that
+    /// had to be clamped at zero. Non-zero means a topic or stream total was
```
hubcio: nit: says "a topic or stream total", but the three partition sites feed the same counter - including the example the doc cites, where the partition clamps and no parent does. say "partition, topic or stream".

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,110 @@ use std::sync::{
...
+/// (`get_stream`, `get_topic`, `/stats`, `/metrics`) serves the wrapped value
+/// until the process restarts.
+///
+/// Clamping trades that for a total that is merely low, and low is where it
+/// stays: increments are `fetch_add`, so every later write stacks on the
+/// clamped base and carries the shortfall with it. Only a rebuild
+/// (`rebuild_parent_totals`) or a restart puts the total back. This counter
+/// plus the `warn!` below is what tells an operator the divergence happened,
+/// and that the aggregate needs a rebuild rather than time.
+static ROLLUP_UNDERFLOWS: AtomicU64 = AtomicU64::new(0);
```
hubcio, line 40: nit: this static is process-global, which the docs never say here. worth stating, since the throttle right below is deliberately split per pair to stop a noisy counter hiding a quiet one's first divergence.

core/server/src/http/metrics.rs
```diff
@@ -203,6 +245,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn rollup_underflow_total_tracks_the_sampled_count() {
+        let metrics = HttpMetrics::init(&[]);
+        metrics.observe_rollup_underflows(3);
+        metrics.observe_rollup_underflows(5);
+        // A restart resets the source; the counter must not run backwards.
```
hubcio: nit: a restart can't produce this state - the static and `HttpMetrics` are recreated together, so `recorded` never exceeds `total`. say the call guards a non-monotonic source defensively, or drop it.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,110 @@ use std::sync::{
...
+    ///
+    /// The throttle is per pair, not global: a skewed tree emits up to three
+    /// lines per counter per rollback and a bulk delete turns that into a
+    /// flood, so a shared count would hold a quiet pair's first-ever
+    /// divergence back for up to 1023 events behind a noisy one.
+    fn report(&self, shortfall: u64) {
+        ROLLUP_UNDERFLOWS.fetch_add(1, Ordering::Relaxed);
+        let clamped = self.clamped.fetch_add(1, Ordering::Relaxed) + 1;
+        if clamped.is_power_of_two() {
+            warn!(
```
hubcio, line 87: nit: no `target:` here, so this can't be filtered with the other diag streams. a scope-neutral target like `iggy.stats.diag` would do - the site logs stream and topic scopes too, so the partitions target is wrong.

core/server/src/http/metrics.rs
```diff
@@ -45,6 +45,30 @@ pub(in crate::http) struct HttpMetrics {
...
     pub(in crate::http) clients: Gauge,
+    /// Not a legacy-parity metric: the running count of rollup decrements that
+    /// had to be clamped at zero. Non-zero means a topic or stream total was
+    /// asked to give back more than it held, so the aggregate it now reports is
+    /// low, and stays low until a rebuild or a restart.
+    ///
+    /// Not "alert on any increase". A delete, a purge, a partition teardown and
+    /// a snapshot restore each open a window where a retention pass hands back
+    /// bytes the parents have already given up, and the clamp is the intended
+    /// outcome there -- the `given_a_late_decrement_on_a_rolled_back_partition`
```
hubcio: nit: the test is `given_a_late_decrement_on_a_rolled_back_partition_should_leave_siblings_alone`, and `iggy_common::streaming_stats` isn't a reachable path since the module is `pub(crate)`. cite the full name plus the file.

core/common/src/types/streaming_stats.rs
```diff
@@ -357,3 +492,115 @@ impl PartitionStats {
...
+        assert_eq!(topic.size_bytes_inconsistent(), 0);
+        assert_eq!(stream.size_bytes_inconsistent(), 0);
+    }
+
+    /// A deleted partition keeps its handle until the reconciler tears it down,
+    /// so a retention sweep can decrement counters the delete already rolled
+    /// back. Those bytes are not in the parents any more, and taking them again
+    /// would take a live sibling's instead.
+    #[test]
+    fn given_a_late_decrement_on_a_rolled_back_partition_should_leave_siblings_alone() {
```
hubcio: nit: no `when_` in this name, while the other four new tests here have one.

core/integration/tests/data_integrity/verify_after_server_restart.rs
```diff
@@ -520,3 +557,17 @@ async fn should_handle_resource_deletion_and_restart() {
...
 }
+
+/// Payload for `should_handle_resource_deletion_and_restart`: small, fixed and
+/// non-empty, so a scope that gets deleted has bytes worth rolling back.
+fn deletion_test_messages() -> Vec<IggyMessage> {
+    (0..8u128)
+        .map(|offset| {
+            IggyMessage::builder()
+                .id(offset + 1)
+                .payload(bytes::Bytes::from_static(b"deletion-stats-payload"))
```
hubcio: nit: `Bytes` isn't in scope anywhere in this file and the rest of it imports at the top. add `use bytes::Bytes;` - this PR added the same import to `scenarios/mod.rs`.

core/metadata/src/stm/stream.rs
```diff
@@ -439,22 +474,30 @@ impl StatsRegistry {
     /// reads the same `Arc`, so partition-plane counters are visible
     /// cross-shard without a gather.
     ///
+    /// Takes the committed record rather than a bare id, because a fresh entry
```
hubcio, line 478: nit: the doc says committed record, but not that it has to be the record for this `(stream_id, topic_id)` key, or that `parent` must be the committed topic's `Arc`. neither is checked, and a fresh entry inherits both.

core/partitions/src/iggy_partition.rs
```diff
@@ -7263,6 +7260,39 @@ pub struct SegmentRemoval {
...
 
+/// Roll one cleaned-up segment out of the partition counters, naming the
+/// partition when the rollback could not be covered.
+///
+/// Retention is the likeliest source of a clamped rollback: a partition on its
+/// way out keeps serving cleanup passes after the delete already settled its
+/// counters into the parents. The `stats_rollup_underflows` metric says a
+/// divergence happened and carries no ids; this says which partition and which
+/// segment, which is what an operator needs to act on it.
+fn settle_cleaned_segment(
```
hubcio: nit: `messages_count` is derived from the same `&segment` one line above the call. derive it inside and return it, then feed `removal.messages` from the return.

core/server/src/partition_reconciler.rs
```diff
@@ -1926,6 +1965,148 @@ mod tests {
...
+        stats.increment_size_bytes(100);
+        assert_eq!(stream_size(&ctx), 100);
+
+        reconcile_once(&ctx).await;
+        ctx.shard.apply_reconcile_ops();
+
+        assert_eq!(
+            stream_size(&ctx),
+            0,
+            "teardown must settle what landed after the commit that acked the delete"
```
hubcio: nit: the metadata apply already evicted the entry, so this passes on `zero_out_all()` at the drop point alone - delete `settle_partition_stats` and it stays green. the sibling test is the one guarding the settle, so rename this for the drop point.

core/integration/tests/server/scenarios/delete_stats_rollback_scenario.rs
```diff
@@ -0,0 +1,221 @@
...
+//! Verifying any of this needs `cargo build --bin iggy-server` first. The
+//! harness spawns the built binary, so a source-only mutation runs against the
+//! previous build and reports a green that means nothing.
+//!
+//! The reads after each delete are deliberately single-shot. Retrying until the
+//! numbers converge, the way the pre-delete assertions do for the async send
+//! folding, would also wait out a broken rollback and pass regardless.
+
+use crate::server::scenarios::{
+    PARTITIONS_COUNT, batch_size, create_client, create_messages, validate_stream,
```
hubcio, line 56: nit: the scenario never checks server-wide totals go back to zero after the deletes, which is the clearest statement of what this fixes. `assert_clean_system` reads no stats - add `validate_system_stats(client, 0, 0)` before it.

core/common/src/types/streaming_stats.rs
```diff
@@ -357,3 +492,115 @@ impl PartitionStats {
...
+        let live_partition = Arc::new(PartitionStats::new(live_topic.clone()));
+        live_partition.increment_size_bytes(900);
+        // Landed through the cached handle after the delete evicted its entry.
+        doomed_partition.increment_size_bytes(100);
+        assert_eq!(stream.size_bytes_inconsistent(), 1000);
+
+        doomed_topic.zero_out_all();
+        assert_eq!(stream.size_bytes_inconsistent(), 900);
+
+        doomed_partition.zero_out_all();
```
hubcio, line 525: nit: both widths expand from the same `clamped_sub!` body using `saturating_sub`, so nothing modulus-specific runs - say it covers the u32 wiring instead. same wording at `stream.rs:3480`.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,110 @@ use std::sync::{
...
+    /// Count the clamp globally, then log this pair's first and every power of
+    /// two after it.
+    ///
+    /// The throttle is per pair, not global: a skewed tree emits up to three
+    /// lines per counter per rollback and a bulk delete turns that into a
+    /// flood, so a shared count would hold a quiet pair's first-ever
+    /// divergence back for up to 1023 events behind a noisy one.
+    fn report(&self, shortfall: u64) {
+        ROLLUP_UNDERFLOWS.fetch_add(1, Ordering::Relaxed);
+        let clamped = self.clamped.fetch_add(1, Ordering::Relaxed) + 1;
```
hubcio, line 85: nit: `clamped` is the per-pair count but the metric it bumps is process-global and carries no scope or counter label, so the two numbers can't be reconciled. log both, or label the metric.

core/metadata/src/stm/stream.rs
```diff
@@ -2338,7 +2547,22 @@ impl Snapshotable for Streams {
...
+        // topics' as separate loads while the partition plane keeps counting, so
+        // they can disagree in either direction. Left in place, a replayed
+        // `DeleteTopic` over that torn shape rolls the topic back against a
+        // stream that never held it, clamps, and raises the rollup-underflow
+        // alarm on every boot with no real divergence behind it.
+        //
+        // The registry is empty here, so this stores zero at both levels; it is
+        // the same walk state transfer uses, which is why there is no second
+        // spelling of it.
+        inner.stats_registry.rebuild_parent_totals(&inner.items);
```
hubcio: simplification: this rebuild overwrites the two `store_from_snapshot` calls in `inner_from_snapshot` on every non-test path, and over a fresh registry it's storing 0 over 0. drop the stores and this call, about 25 lines - the tests at 4110 and 3573 go vacuous.

core/common/src/types/streaming_stats.rs
```diff
@@ -42,18 +146,21 @@ impl StreamStats {
...
+    /// hold -- so a caller that knows the ids can name where the divergence is.
+    /// Zero whenever the decrement was covered.
+    pub fn decrement_size_bytes(&self, size_bytes: u64) -> u64 {
+        size_bytes - clamped_sub_u64(&self.size_bytes, size_bytes, &STREAM_SIZE_BYTES)
     }
 
-    pub fn decrement_messages_count(&self, messages_count: u64) {
-        self.messages_count
-            .fetch_sub(messages_count, Ordering::AcqRel);
+    pub fn decrement_messages_count(&self, messages_count: u64) -> u64 {
```
hubcio: simplification: nothing reads the stream-level or topic-level shortfall anywhere - each level computes its own from `taken`. return `()` at both levels and keep the three on `PartitionStats`, which also cuts the public signature churn from nine to three.

core/integration/tests/server/general.rs
```diff
@@ -88,6 +88,17 @@ async fn stream_size_validation(harness: &TestHarness) {
     stream_size_validation_scenario::run(harness).await;
 }
 
+#[iggy_harness(
```
hubcio: simplification: the scenario's own doc calls it transport-irrelevant and says it guards neither half, and every assertion goes through `get_topic`, `get_stream` or `get_stats`, all already covered on four transports. `[Tcp]` drops 9 server processes.

core/server/src/http/metrics.rs
```diff
@@ -106,6 +137,17 @@ impl HttpMetrics {
         self.http_requests.clone()
     }
 
+    /// Advance the rollup-underflow counter to `total`, the process-wide count
```
hubcio: simplification: the get, `checked_sub`, `inc_by` dance exists only because `Counter` has no setter, and it guards a state the doc calls impossible. a ~12-line `Collector` with `ConstCounter` emits the same `_total` series - add a test for it, `PARITY_METRIC_NAMES` skips this one.

core/common/src/types/streaming_stats.rs
```diff
@@ -282,21 +399,39 @@ impl PartitionStats {
...
+        let taken = clamped_sub_u32(
+            &self.segments_count,
+            segments_count,
+            &PARTITION_SEGMENTS_COUNT,
+        );
+        self.decrement_parent_segments_count(taken);
+        segments_count - taken
     }
 
     pub fn decrement_parent_size_bytes(&self, size_bytes: u64) {
```
hubcio: simplification: these twelve `increment_parent_*` and `decrement_parent_*` wrappers have no caller outside this file, and they're public on a published crate. inline them - `self.parent.increment_size_bytes(x)` is the same one line.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,110 @@ use std::sync::{
...
+/// `fetch_update` rather than a load followed by a subtract: the counters are
+/// written from the metadata shard and from whichever shard owns the partition,
+/// so a separate load leaves a window where the clamp reads one value and
+/// subtracts from another.
+macro_rules! clamped_sub {
+    ($name:ident, $counter:ty, $amount:ty) => {
+        fn $name(counter: &$counter, amount: $amount, site: &'static UnderflowSite) -> $amount {
+            let previous = counter
+                .fetch_update(Ordering::AcqRel, Ordering::Acquire, |current| {
+                    Some(current.saturating_sub(amount))
```
hubcio: simplification: the closure always returns `Some`, so the `Err` arm can't happen. `(current != 0).then(|| current.saturating_sub(amount))` gives it a meaning and skips the CAS on an already-zero counter.

core/common/src/types/streaming_stats.rs
```diff
@@ -19,6 +19,110 @@ use std::sync::{
...
+                clamped,
+                "rollup decrement exceeded the total it was subtracted from; clamped at zero"
+            );
+        }
+    }
+}
+
+static STREAM_SIZE_BYTES: UnderflowSite = UnderflowSite::new("stream", "size_bytes");
+static STREAM_MESSAGES_COUNT: UnderflowSite = UnderflowSite::new("stream", "messages_count");
+static STREAM_SEGMENTS_COUNT: UnderflowSite = UnderflowSite::new("stream", "segments_count");
```
hubcio, line 102: simplification: nine per-pair statics where three per-scope sites plus a `counter` argument would do, about 9 lines. costs you the property the doc right above defends - a noisy counter would then swallow a quiet one's first divergence.

core/integration/tests/server/scenarios/mod.rs
```diff
@@ -72,6 +74,133 @@ const USERNAME_3: &str = "user3";
 const CONSUMER_KIND: ConsumerKind = ConsumerKind::Consumer;
 const MESSAGES_COUNT: u32 = 1337;
 
+// The partition plane applies committed ops asynchronously on the owning shard
+// (sends fold into the shared stats at commit-apply; purge/delete zero them
+// when the reconciler drives the wipe), so a read racing that window can see a
+// pre-apply value. Retry until the expectation holds, then make the terminal
+// assertion for a real mismatch.
+const STATS_CONVERGENCE_TIMEOUT: Duration = Duration::from_secs(10);
```
hubcio: simplification: same 10s and 100ms as `POLL_CONVERGENCE_TIMEOUT` and `POLL_RETRY_INTERVAL` at 65, and both wait out the same convergence. keep one neutral pair and update the doc links at 114 and 205.

hubcio (changes requested):
`retain_from_snapshot` (stream.rs:722) evicts a partition entry without zeroing its stats, so the still-mounted stale incarnation's later `ConfirmRemove` takes those bytes a second time out of live siblings. No clamp fires, so nothing surfaces it. Reachable on a donor delete-then-recreate installed ahead of the receiver's frontier. Breaks the invariant at stream.rs:738, and `created_revision` has no test.

On unchanged lines, so no inline comment:

- `boot/recovery.rs:184` points at the zeroing loop this change deleted from `boot/mod.rs`. Retarget at `Streams::from_snapshot`.
- `streaming_stats.rs:313-326` and `:469-482`: `zero_out_*` cascades unconditionally, so the new settle runs an amount-0 rollback per deleted partition. `if prev == 0 { return; }` after each swap.
- Nine public `iggy_common` return types change from `()`. A major break with no semver gate - worth a release note.

core/shard/src/lib.rs
numinnex: Leaving this one, with a reason rather than a fix.

The discarded build shares the namespace's `PartitionStats` Arc with the live incarnation, so neither arm can roll back what it folded: zeroing there is exactly the clobber that already cost a live partition its `current_offset` once.

Folding at adoption is the right shape, but it means the loader and the builder stop touching the shared Arc at all and hand `insert` a local delta instead. `load_partition` counts as it walks the chain, and `load_partition_or_fence` zeroes the shared Arc on a refusal, so that is a boot-path change rather than a local one. Both arms are unreachable today and already log plus count as damage control, so the refactor is out of proportion to this PR. Worth a separate issue.

core/common/src/types/streaming_stats.rs
numinnex: Keeping the nine statics.

The trade you name is the deciding one: three sites plus a `counter` argument drops the per-pair throttle, and that property is why the throttle is not global already. A bulk delete floods one pair, and a shared count would hold a quiet pair's first-ever divergence back behind it for up to 1023 events.

Nine flat statics is the cheapest shape that keeps it. An array per scope needs a counter enum plus index plumbing at every call site, which is more code than the nine lines it saves.

hubcio (approved):

mmodzelewski (approved):

PR #4055, opened 2026-09-04
feat(server): generic external auth callout

### LINKED ISSUE #3929: feat(server): Generic external auth callout
### Description

Currently Iggy supports built-in credentials and external JWT verification for HTTP only. A trusted JWT issuer maps all accepted tokens to a configured Iggy user, after which authorization uses that user’s locally stored permissions.

There is no mechanism for an external service to authenticate a client and return its effective permissions.

This limits integration with centralized identity and policy systems, especially for service-to-service and device-oriented protocols.

For example, consider a scenario of fleet of embedded devices which support MQTT. During `CONNECT` iggy may resolve its effective permissions without provisioning a persistent Iggy user for every device.

### Affected area / component

Wire protocol / API

### Proposed solution

During client login or protocol handshake:

1. Iggy sends one request to the configured auth service
2. auth service authenticates the client and returns permissions
3. Iggy creates an authenticated session
4. existing operation-specific permission rules enforce the permissions

  Sample request (schema too be refined):

```json
{
  "transport": "tcp",
  "connection_id": "0195f4d7-57d3-7c21-a711-08a147ad6c90",
  "remote_address": "192.0.2.10:54321",
  "credentials": {
    "type": "username_password",
    "username": "application-service",
    "secret": "<redacted>"
  }
}
```

The credential object can be a tagged enum so other variants can be supported present and future, viz bearer tokens, X509 certs, etc.

Example response using session-scoped permissions:

```json
{
  "authenticated": true,
  "principal": "service:analytics-ingestor",
  "authorization": {
    "type": "permissions",
    "permissions": {
      "global": {
        "manage_servers": false,
        "read_servers": false,
        "manage_users": false,
        "read_users": false,
        "manage_streams": false,
        "read_streams": false,
        "manage_topics": false,
        "read_topics": false,
        "poll_messages": false,
        "send_messages": false
      },
      "streams": {
        "42": {
          "manage_stream": false,
          "read_stream": false,
          "manage_topics": false,
          "read_topics": false,
          "poll_messages": false,
          "send_messages": true,
          "topics": null
        }
      }
    }
  },
  "expires_at": 1787212800
}
```

Alternatively, the provider could select an existing Iggy user:

```json
{
  "authenticated": true,
  "principal": "service:analytics-ingestor",
  "authorization": {
    "type": "iggy_user",
    "user_id": 12
  },
  "expires_at": 1787212800
}
```

A denied response could be:

```json
{
  "authenticated": false,
  "reason": "credential revoked"
}
```
Session-scoped permissions should reuse Iggy’s existing Permissions, GlobalPermissions, StreamPermissions, and TopicPermissions structures. They must not create users or modify persisted permissions.

The provider must fail closed on timeout, invalid response, explicit denial, or unavailability.

Decision left to maintainers on whether to follow fixed schema or define via some form of templating.

### Alternatives considered

  - Provision an Iggy user for every external identity: Works with the current permission model but creates lifecycle and synchronization overhead, particularly for large device fleets.
  - Map each JWT issuer to one fixed Iggy user: Already supported for HTTP trusted issuers, but it cannot assign different permissions to individual external subjects.
  - Call the authorization service for every operation: Supports immediate policy changes but adds latency and availability dependencies to message hot paths.
  - Use client-certificate authentication: Suitable for some service identities, but it still requires identity-to-permission mapping and does not cover all credential types.

### Contribution

### Good first issue

sansmoraxz:

Closes #3929

## Rationale

Iggy supports only built-in credentials and HTTP-only JWT verification. There is no mechanism for an external service to authenticate a client and return its effective permissions, which blocks integration with centralized identity/policy systems (like Azure AD) or automated machine fleets.

## What changed?

Login attempts on all transports (TCP, QUIC, WebSocket, HTTP) can forwarded to a configured HTTP endpoint. The service authenticates the client and returns one of three decisions: map to an existing Iggy user, grant a session-scoped identity with explicit permissions, or deny. Session-scoped identities use synthetic user IDs that are never persisted and are restricted to data-plane operations; (capacity defined as `1_000_000`). The users ids within this pool are rotated (and next avaiable free user ids would be allocated). This can be adjusted if necessary.

### Request/response contract

```json
// Request (POST to configured URL)
{
  "credential_type": "password" | "personal_access_token",
  "credential": "...",
  "username": "alice",
  "transport": "tcp" | "http" | "quic" | "websocket",
  "client_address": "10.0.0.1:5000"
}

// Response
{
  "decision": "iggy_user" | "inline_grant" | "deny",
  "user_id": 42,
  "principal": "device-1234",
  "permissions": { "global": { ... } },
  "expires_at": 1700000000,
  "reason": optional string ("certificate revoked")
}
```

### Configuration

```toml
[external_auth]
enabled = false
url = "https://auth.example.com/validate"
timeout = "5 s"
on_error = "deny"          # "deny" | "fallback"
forward_credentials = true
```

### E2E validation

Built the server and validated the full external auth flow against a mock auth service.

**Startup:** Server loaded the config and logged the expected warning:

**Test 1:  Login `ext-alice` (inline_grant):** Mock received the callout with `credential_type`, `credential` (forwarded), `username`, `transport: "http"`, `client_address`. Responded with `inline_grant` granting `read_streams`, `read_topics`, `poll_messages`, `send_messages`. Server returned HTTP 200 with a JWT for synthetic `user_id: 4294967295` (u32::MAX, first minted ID).

**Test 2:  Login `ext-deny` (deny):** Mock responded `{"decision":"deny","reason":"blocked by policy"}`. Server returned HTTP 401.

**Test 3:  GET /streams with alice's token (allowed):** HTTP 200, body `[]`. The grant's `read_streams: true` allowed the read.

**Test 4:  POST /streams with alice's token (denied):** HTTP 403. The grant's `manage_streams: false` blocked the mutation.

## Local Execution

- Passed
- Pre-commit hooks ran

## AI Usage

1. Which tools? Claude, Github Copilot
2. Scope of usage? Research and implementation (some agents, some auto completes)
3. How did you verify the generated code works correctly? Also unit tests and validation against mock servers.

### DIFF
--- a/core/common/src/types/permissions/permissions_global.rs
+++ b/core/common/src/types/permissions/permissions_global.rs
@@ -26,6 +26,7 @@ use std::fmt::Display;
 /// Global permissions are applied to all streams.
 /// Stream permissions are applied to a specific stream.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+#[serde(default)]
 pub struct Permissions {
     /// Global permissions are applied to all streams.
     pub global: GlobalPermissions,
@@ -36,6 +37,7 @@ pub struct Permissions {
 
 /// `GlobalPermissions` are applied to all streams without a need to specify them one by one in the `streams` field.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+#[serde(default)]
 pub struct GlobalPermissions {
     /// `manage_servers` permission allows to manage the servers and includes all the permissions of `read_servers`.
     pub manage_servers: bool,
@@ -104,6 +106,7 @@ pub struct GlobalPermissions {
 /// `StreamPermissions` are applied to a specific stream and its all topics. If you want to define granular permissions for each topic, use the `topics` field.
 /// These permissions do not override the global permissions, but extend them, and allow more granular control over the streams and the users that can access them.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+#[serde(default)]
 pub struct StreamPermissions {
     /// `manage_stream` permission allows to manage the stream and includes all the permissions of `read_stream`.
     /// Also, it allows to manage all the topics of a stream, thus it has all the permissions of `manage_topics`.
@@ -143,6 +146,7 @@ pub struct StreamPermissions {
 
 /// `TopicPermissions` are applied to a specific topic of a stream. This is the lowest level of permissions.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+#[serde(default)]
 pub struct TopicPermissions {
     /// `manage_topic` permission allows to manage the topic and includes all the permissions of `read_topic`.
     pub manage_topic: bool,
--- a/core/configs/src/server_config/external_auth.rs
+++ b/core/configs/src/server_config/external_auth.rs
@@ -45,6 +45,16 @@ pub struct ExternalAuthConfig {
     /// inline grants share this single ID; the external principal string
     /// distinguishes sessions. Must not be 0 (root). The replicated
     /// state-machine gate recognizes this ID at boot.
+    ///
+    /// **Changing this value between restarts is safe.** External auth
+    /// sessions are connection-scoped and never persisted, so there is no
+    /// stale state referencing the old value. The server validates at boot
+    /// that the new value does not collide with any existing user in the
+    /// slab. Within a running process the value is immutable (set-once via
+    /// `OnceLock`); a hot-reload that attempts a different value panics.
+    /// In a cluster, all nodes should use the same value to avoid
+    /// operational confusion, though cross-node correctness is not
+    /// affected (session state is node-local).
     #[serde(default = "default_external_auth_user_id")]
     #[config_env(leaf)]
     pub user_id: u32,
--- a/core/configs/src/server_config/external_auth.rs
+++ b/core/configs/src/server_config/external_auth.rs
@@ -1,0 +1,59 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use configs::ConfigEnv;
+use iggy_common::IggyDuration;
+use serde::{Deserialize, Serialize};
+use serde_with::DisplayFromStr;
+use serde_with::serde_as;
+
+/// External authentication callout configuration.
+///
+/// When enabled, login attempts that fail built-in credential verification
+/// are forwarded to an external HTTP service. The service returns a grant
+/// (with inline permissions or by mapping to an existing Iggy user) or a
+/// denial. Built-in users never hit the callout.
+#[serde_as]
+#[derive(Debug, Deserialize, Serialize, Clone, ConfigEnv)]
+pub struct ExternalAuthConfig {
+    #[serde(default)]
+    pub enabled: bool,
+    #[serde(default, skip_serializing)]
+    #[config_env(secret)]
+    pub url: String,
+    #[serde_as(as = "DisplayFromStr")]
+    #[serde(default = "default_external_auth_timeout")]
+    #[config_env(leaf)]
+    pub timeout: IggyDuration,
+    #[serde(default)]
+    pub forward_credentials: bool,
+    /// Reserved user ID for inline-grant sessions. All external auth
+    /// inline grants share this single ID; the external principal string
+    /// distinguishes sessions. Must not be 0 (root). The replicated
+    /// state-machine gate recognizes this ID at boot.
+    #[serde(default = "default_external_auth_user_id")]
+    #[config_env(leaf)]
+    pub user_id: u32,
+}
+
+fn default_external_auth_timeout() -> IggyDuration {
+    "5 s".parse().expect("hardcoded timeout")
+}
+
+const fn default_external_auth_user_id() -> u32 {
+    u32::MAX
+}
--- a/core/metadata/src/stm/authz.rs
+++ b/core/metadata/src/stm/authz.rs
@@ -77,7 +77,12 @@ where
     M: StreamsFrontend
         + StateMachine<Input = Message<PrepareHeader>, Output = ApplyReply, Error = IggyError>,
 {
-    if let Some(denied) = authorize(&prepare, mux.users(), mux.streams()) {
+    if let Some(denied) = authorize(
+        &prepare,
+        mux.users(),
+        mux.streams(),
+        mux.external_auth_user_id(),
+    ) {
         return Ok(denied);
     }
     mux.update(prepare)
@@ -129,6 +134,7 @@ pub(crate) fn authorize(
     prepare: &Message<PrepareHeader>,
     users: &Users,
     streams: &Streams,
+    external_auth_user_id: Option<u32>,
 ) -> Option<ApplyReply> {
     let header = prepare.header();
     let user_id = header.user_id;
@@ -136,7 +142,30 @@ pub(crate) fn authorize(
     if user_id == ROOT_USER_ID {
         return None;
     }
+
     let body = prepare.body();
+
+    // The external auth inline-grant user may execute the replicated
+    // data-plane ops listed below. Each is also gated at dispatch time by
+    // session-scoped permissions (authorize_partition_op for SendMessages
+    // and consumer offsets; authorize_consumer_group_op for CG join/leave).
+    // Non-replicated reads (PollMessages, GetConsumerOffset) never reach
+    // the STM and are gated only at dispatch time.
+    // SYNC: if a new REPLICATED data-plane operation is added, add it
+    // here AND add a dispatch-time session-scoped permission check in
+    // dispatch/authz.rs or dispatch/mod.rs.
+    if external_auth_user_id.is_some_and(|id| user_id == id) {
+        return match header.operation {
+            Operation::Register
+            | Operation::Logout
+            | Operation::SendMessages
+            | Operation::StoreConsumerOffset
+            | Operation::DeleteConsumerOffset
+            | Operation::JoinConsumerGroup
+            | Operation::LeaveConsumerGroup => None,
+            _ => Some(ApplyReply::err(IggyError::Unauthorized.as_code())),
+        };
+    }
 
     match header.operation {
         // Streams. `create_stream` is unscoped; the rest resolve the stream id.
@@ -417,11 +446,16 @@ mod tests {
     const ROOT: u32 = 0;
     const ALICE: u32 = 1;
 
+    /// External auth user ID used by tests that exercise the inline-grant gate.
+    const EXT_AUTH_USER_ID: u32 = 99_999;
+
     fn mux() -> MuxStateMachine<variadic!(Users, Streams)> {
         let users = Users::default();
         users.ensure_root_user("iggy", "hash");
         let streams = Streams::default();
-        MuxStateMachine::new(variadic!(users, streams))
+        let m = MuxStateMachine::new(variadic!(users, streams));
+        m.set_external_auth_user_id(EXT_AUTH_USER_ID);
+        m
     }
 
     fn make_prepare(
@@ -685,7 +719,13 @@ mod tests {
         let mux = mux();
         let prepare = make_prepare(1, Operation::CompleteConsumerGroupRevocation, ROOT, &[]);
         assert!(
-            authorize(&prepare, mux.users(), mux.streams()).is_none(),
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
             "server-originated revocation must not be gated"
         );
     }
@@ -703,7 +743,13 @@ mod tests {
 
         let prepare = make_prepare(2, Operation::CreatePersonalAccessToken, ALICE, &[]);
         assert!(
-            authorize(&prepare, mux.users(), mux.streams()).is_none(),
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
             "PAT ops are self-scoped (parity: no legacy RBAC)"
         );
     }
@@ -754,5 +800,153 @@ mod tests {
         assert_eq!(primary_codes[2], UNAUTHORIZED, "alice's create is denied");
         assert_eq!(primary.streams().read(|inner| inner.items.len()), 1);
         assert_eq!(replay.streams().read(|inner| inner.items.len()), 1);
+    }
+
+    #[test]
+    fn given_external_auth_user_when_register_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::Register, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to Register"
+        );
+    }
+
+    #[test]
+    fn given_external_auth_user_when_send_messages_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::SendMessages, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to SendMessages"
+        );
+    }
+
+    #[test]
+    fn given_external_auth_user_when_join_consumer_group_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::JoinConsumerGroup, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to JoinConsumerGroup"
+        );
+    }
+
+    #[test]
+    fn given_external_auth_user_when_create_stream_should_deny() {
+        let mux = mux();
+        let prepare = make_prepare(
+            1,
+            Operation::CreateStream,
+            EXT_AUTH_USER_ID,
+            &create_stream_body("s"),
+        );
+        let reply = authorize(
+            &prepare,
+            mux.users(),
+            mux.streams(),
+            mux.external_auth_user_id(),
+        );
+        assert!(
+            reply.is_some(),
+            "external auth user must not create streams"
+        );
+        assert_eq!(reply.unwrap().code, UNAUTHORIZED);
+    }
+
+    #[test]
+    fn given_external_auth_user_when_create_user_should_deny() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::CreateUser, EXT_AUTH_USER_ID, &[]);
+        let reply = authorize(
+            &prepare,
+            mux.users(),
+            mux.streams(),
+            mux.external_auth_user_id(),
+        );
+        assert!(reply.is_some(), "external auth user must not create users");
+        assert_eq!(reply.unwrap().code, UNAUTHORIZED);
+    }
+
+    #[test]
+    fn given_external_auth_user_when_logout_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::Logout, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to Logout"
+        );
+    }
+
+    #[test]
+    fn given_external_auth_user_when_store_consumer_offset_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::StoreConsumerOffset, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to StoreConsumerOffset"
+        );
+    }
+
+    #[test]
+    fn given_external_auth_user_when_delete_consumer_offset_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::DeleteConsumerOffset, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to DeleteConsumerOffset"
+        );
+    }
+
+    #[test]
+    fn given_external_auth_user_when_leave_consumer_group_should_allow() {
+        let mux = mux();
+        let prepare = make_prepare(1, Operation::LeaveConsumerGroup, EXT_AUTH_USER_ID, &[]);
+        assert!(
+            authorize(
+                &prepare,
+                mux.users(),
+                mux.streams(),
+                mux.external_auth_user_id()
+            )
+            .is_none(),
+            "external auth user must be allowed to LeaveConsumerGroup"
+        );
     }
 }
--- a/core/server/src/boot/mod.rs
+++ b/core/server/src/boot/mod.rs
@@ -107,6 +107,7 @@ pub fn wire_shell_handlers<B, MJ, S, SB>(
     shard_handle: &ShellShardHandle<B, MJ, S, SB>,
     system_config: Arc<ServerSystemConfig>,
     max_tokens_per_user: u32,
+    external_auth: Arc<configs::external_auth::ExternalAuthConfig>,
 ) -> ShellHandlers
 where
     B: ShellBus,
@@ -124,6 +125,7 @@ where
             &sessions,
             system_config,
             max_tokens_per_user,
+            external_auth,
         ),
         on_metadata_submit: make_metadata_submit_handler(shard_handle),
         on_list_clients: make_list_clients_handler(&sessions),
@@ -255,6 +257,22 @@ pub fn bootstrap(
     // run this bootstrap; after this line those are no-ops.
     install_default_crypto_provider();
     validate_root_credentials_env(&config)?;
+    crate::external_auth::validate_config(&config.external_auth)?;
+    if config.external_auth.enabled {
+        for issuer in config.http.jwt.trusted_issuers.iter().flatten() {
+            if issuer.user_id == config.external_auth.user_id {
+                return Err(
+                    crate::server_error::ServerError::InvalidExternalAuthConfig {
+                        reason: format!(
+                            "trusted issuer '{}' user_id {} collides with external_auth.user_id",
+                            issuer.issuer, issuer.user_id
+                        ),
+                    },
+                );
+            }
+        }
+    }
+    crate::external_auth::warn_insecure_url(&config.external_auth);
     warm_dummy_password_hash();
     // The sync GetStats read path has no access to server config, so capture
     // the data directory here for its disk-usage reporting.
@@ -531,6 +549,9 @@ async fn shard_main(
                 config.metadata.clients_table_max,
                 |mux_stm| {
                     ensure_default_root_user(mux_stm);
+                    if config.external_auth.enabled {
+                        mux_stm.set_external_auth_user_id(config.external_auth.user_id);
+                    }
                 },
                 |mux_stm, client, stamp| {
                     mux_stm
@@ -627,6 +648,29 @@ async fn shard_main(
     } else {
         (None, None, None, None, (0, 0), None)
     };
+    // Peer shards (Waiter path) and snapshot-restored mux_stm need the
+    // external auth user_id too. The Owner path sets it in seed_baseline
+    // (before WAL replay); the Waiter path gets it here after the bundle.
+    // set_external_auth_user_id is idempotent for the same value, so the
+    // Owner path's earlier set is a no-op here.
+    if config.external_auth.enabled {
+        mux_stm.set_external_auth_user_id(config.external_auth.user_id);
+        // The reserved user_id must not collide with a real user, or the
+        // authz gate would block that user from all metadata operations.
+        let uid = config.external_auth.user_id;
+        let collides = mux_stm
+            .users()
+            .read(|users| users.items.contains(uid as usize));
+        if collides {
+            return Err(ServerError::InvalidExternalAuthConfig {
+                reason: format!(
+                    "external_auth.user_id {uid} collides with an existing user in the slab; \
+                     pick a user_id that is not assigned to any Iggy user (default: {})",
+                    u32::MAX
+                ),
+            });
+        }
+    }
     let metadata = ServerMetadata::new(
         metadata_consensus,
         journal_for_metadata,
@@ -668,6 +712,7 @@ async fn shard_main(
     // Heap-pin like `run_shard_thread` pins `shard_main`: the builder future
     // carries the whole shard construction state machine and outgrew clippy's
     // `large_futures` cap; one allocation per shard startup.
+    let external_auth_config = Arc::new(config.external_auth.clone());
     let ShardBuild {
         shard,
         sessions,
@@ -685,6 +730,7 @@ async fn shard_main(
         reply_inbox,
         shard_metrics,
         &roster_cells,
+        Arc::clone(&external_auth_config),
     ))
     .await?;
 
@@ -964,6 +1010,7 @@ async fn shard_main(
             dialed_replica,
             accepted_client,
             &shard_metrics_all,
+            external_auth_config,
         )
         .await
         {
--- a/core/server/src/dispatch/authz.rs
+++ b/core/server/src/dispatch/authz.rs
@@ -35,13 +35,14 @@ use iggy_binary_protocol::codes::{
     GET_USER_CODE, GET_USERS_CODE,
 };
 use iggy_binary_protocol::requests::consumer_groups::{
-    GetConsumerGroupRequest, GetConsumerGroupsRequest,
+    GetConsumerGroupRequest, GetConsumerGroupsRequest, JoinConsumerGroupRequest,
+    LeaveConsumerGroupRequest,
 };
 use iggy_binary_protocol::requests::streams::GetStreamRequest;
 use iggy_binary_protocol::requests::topics::{GetTopicRequest, GetTopicsRequest};
 use iggy_binary_protocol::requests::users::GetUserRequest;
 use iggy_binary_protocol::{Operation, PrepareHeader, WireDecode, WireIdentifier, lookup_command};
-use iggy_common::IggyError;
+use iggy_common::{IggyError, Permissions};
 use journal::superblock::SuperblockStore;
 use journal::{Journal, JournalHandle};
 use metadata::impls::metadata::StreamsFrontend;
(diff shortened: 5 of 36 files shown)


hubcio:
hi,
please rebase this PR :)

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+use configs::external_auth::{ExternalAuthConfig, ExternalAuthErrorStrategy};
+use iggy_common::Permissions;
+use serde::{Deserialize, Serialize};
+use tracing::warn;
+
+const MAX_RESPONSE_BODY_BYTES: usize = 1_048_576;
+
+thread_local! {
+    static HTTP_CLIENT: cyper::Client =
+        cyper::Client::new().expect("failed to build cyper HTTP client for external auth");
```
hubcio: critical: `cyper::Client::new()` follows redirects by default and resends the body on 307/308, so the plaintext credential goes to whatever host `Location` names and that host's reply becomes the auth decision. use `.redirect(Policy::none())`, same as `http/forward.rs:159`.

core/server/src/http/state.rs
```diff
@@ -382,6 +426,15 @@ impl HttpInner {
...
     pub(in crate::http) fn forget_session(&self, session: &Rc<HttpSession>) {
         let torn = forget_if_same(&mut self.sessions.borrow_mut(), session);
+        if torn.is_some()
+            && self
+                .synthetic_permissions
+                .borrow_mut()
+                .remove(&session.user_id)
+                .is_some()
+        {
+            self.free_synthetic_ids.borrow_mut().insert(session.user_id);
```
hubcio: critical: the synthetic id goes back on the free list while its jwt is still valid, and there is no revocation list. the next inline grant gets that id, and the old token then authorizes as the new principal.

core/server/src/http/state.rs
```diff
@@ -382,6 +426,15 @@ impl HttpInner {
     /// request re-register cleanly through the barrier.
     pub(in crate::http) fn forget_session(&self, session: &Rc<HttpSession>) {
         let torn = forget_if_same(&mut self.sessions.borrow_mut(), session);
+        if torn.is_some()
```
hubcio: critical: `forget_if_same` returns `None` when `registry_token` is unset even though it already removed the table entry, so this cleanup is skipped. any session that never did an acked produce leaks its permissions entry and burns an id.

core/server/src/dispatch/session_ops.rs
```diff
@@ -1309,6 +1312,192 @@ pub(in crate::dispatch) async fn handle_login_register_request<B, MJ, S, SB>(
...
+                    send_login_eviction(
+                        shard,
+                        transport_client_id,
+                        vsr_client_id,
+                        EvictionReason::InvalidCredentials,
+                    )
+                    .await;
+                    return;
+                }
+                if crate::external_auth::is_synthetic_user_id(user_id) {
```
hubcio: critical: nothing checks that this user exists or is active. the built-in path and the http twin both check, config.toml:1085 promises it, and user ids get recycled - so a stale mapping logs in as whoever took that slot.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+/// Process-wide counter for minting synthetic user IDs. Wraps an
+/// `Arc<AtomicU32>` so every transport (TCP, QUIC, WS, HTTP) draws from
+/// the same sequence and no two transports can mint the same ID.
+#[derive(Clone)]
+pub struct SyntheticUserIdCounter(Arc<AtomicU32>);
+
+impl SyntheticUserIdCounter {
+    #[must_use]
+    pub fn new() -> Self {
+        Self(Arc::new(AtomicU32::new(u32::MAX)))
```
hubcio: critical: this counter is per process but the jwt key is cluster-wide, so two nodes both mint `u32::MAX` first for different principals. a token minted on one node then resolves against the other node's grant.

core/server/src/http/handlers.rs
```diff
@@ -1799,6 +1866,103 @@ pub(in crate::http) async fn delete_pat(
...
+        ExternalAuthDecision::InlineGrant {
+            principal: _,
+            permissions,
+            expires_at,
+        } => {
+            let Some(synthetic_user_id) = state.mint_synthetic_user_id() else {
+                tracing::error!("synthetic user ID space exhausted");
+                return Err(IggyError::Unauthenticated.into());
+            };
+            state
```
hubcio: critical: this entry is only removed when a registered `HttpSession` expires or is forgotten, but read routes use `Identity` and register nothing. a poll-only client never creates a session, so its id leaks for the process lifetime.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+        .and_then(|v| v.to_str().ok())
+        .and_then(|s| s.parse::<usize>().ok())
+        && len > MAX_RESPONSE_BODY_BYTES
+    {
+        return Err(ExternalAuthError::BadResponse(
+            "response body too large".to_owned(),
+        ));
+    }
+
+    let bytes = compio::time::timeout(timeout, response.bytes())
```
hubcio: critical: `bytes()` buffers the whole body before the 1 MiB check below, and the content-length precheck is skipped on a chunked reply. there is no actual cap, though config.toml:1101 says there is.

core/server/src/dispatch/session_ops.rs
```diff
@@ -1309,6 +1312,192 @@ pub(in crate::dispatch) async fn handle_login_register_request<B, MJ, S, SB>(
...
+                        .await;
+                }
+                return;
+            }
+            Ok(Some(crate::external_auth::ExternalAuthDecision::InlineGrant {
+                principal,
+                permissions,
+                expires_at,
+            })) => {
+                let synthetic_id = sessions.borrow_mut().mint_synthetic_user_id();
```
hubcio: critical: a second `Register` on a bound connection is accepted, since `dispatch/mod.rs:702` runs before the bound check. it mints an id nothing frees, overwrites `conn.session_permissions`, and leaves the old bound id resolving to the new grant.

core/metadata/src/stm/authz.rs
```diff
@@ -136,6 +136,11 @@ pub(crate) fn authorize(
     if user_id == ROOT_USER_ID {
         return None;
     }
+
+    if user_id > SYNTHETIC_USER_ID_THRESHOLD {
```
hubcio: critical: this runs before the operation match, so it also denies `JoinConsumerGroup` and `LeaveConsumerGroup`, which config.toml:1089 tells operators inline grants can do. the denial is post-commit, so each retry costs a journal append and a quorum round.

core/server/src/dispatch/session_ops.rs
```diff
@@ -1309,6 +1312,192 @@ pub(in crate::dispatch) async fn handle_login_register_request<B, MJ, S, SB>(
...
+                    &LoginRegisterError::ExternalAuthDenied(reason),
+                )
+                .await;
+                return;
+            }
+            Ok(None) => {
+                // on_error = fallback: fall through to built-in credential check
+            }
+            Err(error) => {
+                warn!(
```
hubcio, line 1505: warning: a timeout or a dead auth service reaches the client as `InvalidCredentials`, same as a wrong password, and `fail_sign_in` then wipes remembered credentials. add a non-terminal variant mapped to a transient code.

core/configs/src/server_config/external_auth.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub enabled: bool,
+    #[serde(default, skip_serializing)]
+    #[config_env(secret)]
+    pub url: String,
+    #[serde_as(as = "DisplayFromStr")]
+    #[serde(default = "default_external_auth_timeout")]
+    #[config_env(leaf)]
+    pub timeout: IggyDuration,
+    #[serde(default)]
+    pub on_error: ExternalAuthErrorStrategy,
```
hubcio: warning: `on_error` has no `#[config_env(leaf)]`, and the enum's unit variants make `env_mappings()` empty, so `IGGY_EXTERNAL_AUTH_ON_ERROR` does not exist. setting it is ignored in release and panics boot in a debug build.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+
+use configs::external_auth::{ExternalAuthConfig, ExternalAuthErrorStrategy};
+use iggy_common::Permissions;
+use serde::{Deserialize, Serialize};
+use tracing::warn;
+
+const MAX_RESPONSE_BODY_BYTES: usize = 1_048_576;
+
+thread_local! {
+    static HTTP_CLIENT: cyper::Client =
```
hubcio, line 48: warning: this `expect` fires when the rustls backend finds no system CA store, and the panic hook sets the shutdown flag - so the first login drops the whole server. build it in `bootstrap()` and return `ServerError`.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+        Ok(client
+            .post(&config.url)
+            .map_err(|e| ExternalAuthError::HttpError(format!("failed to build request: {e}")))?
+            .header("content-type", "application/json")
+            .map_err(|e| ExternalAuthError::HttpError(format!("failed to set header: {e}")))?
+            .body(body))
+    };
+    let request_builder = build()?;
+
+    let response = compio::time::timeout(timeout, request_builder.send())
```
hubcio: warning: the timeout wraps `send()` here and `bytes()` again below, so one login can take twice the configured budget. use a single deadline across both.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+    }
+}
+
+/// Call the external auth service and parse the response.
+///
+/// # Errors
+///
+/// Returns [`ExternalAuthError`] on network/timeout/parse failure. The
+/// caller applies the configured `on_error` strategy.
+pub async fn callout_external_auth(
```
hubcio, line 261: warning: every login attempt fires a POST before any credential check, with no in-flight cap and no single-flight. on the http route that is unauthenticated and unbounded - `jwks.rs:112` already solves this shape.

core/server/src/http/handlers.rs
```diff
@@ -384,6 +426,7 @@ pub(in crate::http) async fn get_topics(
         |permissioner, uid| {
             scope.map_or(Ok(()), |stream_id| permissioner.get_topics(uid, stream_id))
         },
+        |p| scope.is_none_or(|sid| can_read_stream(p, sid)),
```
hubcio: warning: this passes `can_read_stream`, but the binary arm uses `can_list_topics`, which is what `Permissioner::get_topics` actually accepts. a grant holding only `read_topics` lists topics over tcp and gets 403 here.

core/server/src/session_manager.rs
```diff
@@ -351,6 +394,88 @@ impl SessionManager {
...
+                return None;
+            }
+            Some(&sp.permissions)
+        })
+    }
+
+    /// Connection IDs whose session permissions have expired. The heartbeat
+    /// verifier or a periodic sweep evicts these.
+    #[must_use]
+    pub fn collect_expired_sessions(&self, now_secs: u64) -> Vec<u128> {
```
hubcio: warning: `collect_expired_sessions` and `is_session_expired` have no callers, so nothing evicts a binary inline-grant connection at `expires_at`. gated ops fail closed, but the connection keeps its client-table slot and the auth-only codes.

core/server/src/dispatch/authz.rs
```diff
@@ -292,41 +460,54 @@ where
         // transport with an `Unauthenticated` Reply before it reaches the
         // builder, so this arm only ever fires if that gate is bypassed.
         GET_CLUSTER_METADATA_CODE => user_id.map(|_| ()).ok_or(IggyError::Unauthenticated),
```
hubcio, line 445: warning: this gates on authentication alone, so an inline grant with every bool false reads the cluster topology - node names, replica ips, endpoints. config.toml:1087 says these sessions are restricted to data-plane ops.

core/server/src/http/state.rs
```diff
@@ -122,6 +123,17 @@ pub(in crate::http) struct HttpInner {
     /// Legacy-parity metric registry served by the scrape route; the router's
     /// counting layer holds a clone of its request counter.
     pub(in crate::http) metrics: HttpMetrics,
+    /// External authentication callout config. Shared across all handlers.
+    pub(in crate::http) external_auth: Arc<ExternalAuthConfig>,
+    /// Session-scoped permissions for synthetic user IDs (external auth
+    /// inline-grant sessions). Keyed by synthetic `user_id`. Cleaned up on
+    /// session expiry sweep.
+    pub(in crate::http) synthetic_permissions: RefCell<HashMap<u32, Permissions>>,
```
hubcio: warning: this map is per node, so a follower-minted grant relayed to the primary misses and 403s with a misleading unauthorized. worse with the id collision, since the primary may hold a different principal under that id.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+        DecisionTag::InlineGrant => {
+            let principal = resp.principal.ok_or_else(|| {
+                ExternalAuthError::BadResponse("inline_grant decision missing principal".to_owned())
+            })?;
+            let permissions = resp.permissions.ok_or_else(|| {
+                ExternalAuthError::BadResponse(
+                    "inline_grant decision missing permissions".to_owned(),
+                )
+            })?;
+            let expires_at = resp.expires_at.unwrap_or(u64::MAX);
```
hubcio, line 367: warning: a missing `expires_at` becomes `u64::MAX`, so the grant is bounded only by the connection or the default token ttl. fail-open default on a security field - bound it, or require the field.

core/server/src/dispatch/session_ops.rs
```diff
@@ -1309,6 +1312,192 @@ pub(in crate::dispatch) async fn handle_login_register_request<B, MJ, S, SB>(
...
+        {
+            crate::external_auth::ExternalAuthRequest {
+                credential_type: crate::external_auth::CredentialType::Password,
+                credential: if external_auth.forward_credentials {
+                    Some(wire_request.password.expose_secret().to_owned())
+                } else {
+                    None
+                },
+                username: wire_request.username.to_string(),
+                transport: "binary".to_owned(),
```
hubcio: warning: `transport` is hardcoded `"binary"` for tcp, quic and websocket, but the documented contract is `tcp`/`quic`/`websocket`. the real kind sits on the connection - `get_connection()` was added for this and never called.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,495 @@
...
+            "external auth URL uses plain HTTP; credentials will be sent in cleartext"
+        );
+    }
+}
+
+/// Session-scoped permissions from an external auth inline grant.
+/// Carried on the connection/session, never persisted.
+#[derive(Debug, Clone)]
+pub struct SessionPermissions {
+    pub principal: String,
```
hubcio, line 250: warning: `principal` is written and never read anywhere, and the http path drops it entirely. it is the only link from a synthetic id back to a real identity, so log it at grant time or drop the field.

core/server/src/session_manager.rs
```diff
@@ -104,15 +112,31 @@ pub struct SessionManager {
     /// per-shard context already threaded to the non-replicated read path;
     /// installed once at bootstrap, disabled until then.
     cluster_roster: Rc<ClusterRoster>,
+    /// Shared counter for minting synthetic user IDs. All transports draw
+    /// from the same sequence so no two can mint the same ID.
+    synthetic_counter: SyntheticUserIdCounter,
+    /// Freed synthetic user IDs available for reuse.
+    free_synthetic_ids: BTreeSet<u32>,
```
hubcio: warning: each shard and the http listener keep their own free list over one shared counter, so freed ids only return to whoever minted them. once the counter drains, a busy shard mints `None` while ids sit free elsewhere.

core/server/src/dispatch/reads.rs
```diff
@@ -142,6 +142,9 @@ pub(in crate::dispatch) async fn handle_non_replicated_request<B, MJ, S, SB>(
     // connection lookup. `user_id` is `None` only on the pre-auth path
     // (PING), which serves ungated codes; the gated arms fail closed on it.
     let (user_id, client_address) = sessions.borrow().read_context(transport_client_id);
+    let session_perms = user_id
+        .filter(|&uid| crate::external_auth::is_synthetic_user_id(uid))
+        .and_then(|uid| sessions.borrow().session_permissions_for_user(uid).cloned());
```
hubcio: warning: this deep-clones `Permissions` on every request, including `PING` and `GET_ME` which never use it, and `streams` is a nested `BTreeMap`. store an `Rc` and clone that instead.

core/server/src/http/handlers.rs
```diff
@@ -1799,6 +1866,103 @@ pub(in crate::http) async fn delete_pat(
...
+        } => {
+            let Some(synthetic_user_id) = state.mint_synthetic_user_id() else {
+                tracing::error!("synthetic user ID space exhausted");
+                return Err(IggyError::Unauthenticated.into());
+            };
+            state
+                .synthetic_permissions
+                .borrow_mut()
+                .insert(synthetic_user_id, permissions);
+            let result = issue_identity_capped(state, synthetic_user_id, expires_at);
```
hubcio: warning: `expires_at` is never compared to now, so a past value mints an already-dead token, still returns 200, and skips the rollback below. that permanently burns a synthetic id per call.

core/server/src/http/state.rs
```diff
@@ -373,6 +382,41 @@ impl HttpInner {
...
+    }
+
+    /// Return a synthetic user ID to the free list without removing a session.
+    pub(in crate::http) fn reclaim_synthetic_user_id(&self, id: u32) {
+        self.free_synthetic_ids.borrow_mut().insert(id);
+    }
+
+    /// Look up session-scoped permissions for a synthetic user ID. Returns
+    /// `None` for non-synthetic users or when no permissions are stored.
+    pub(in crate::http) fn get_synthetic_permissions(&self, user_id: u32) -> Option<Permissions> {
```
hubcio: warning: this hands back the grant without checking `expires_at`, while the binary twin at `session_manager.rs:449` does check it. on http an expired grant keeps authorizing until the token itself expires.

core/configs/src/server_config/external_auth.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct ExternalAuthConfig {
+    #[serde(default)]
+    pub enabled: bool,
+    #[serde(default, skip_serializing)]
+    #[config_env(secret)]
+    pub url: String,
+    #[serde_as(as = "DisplayFromStr")]
+    #[serde(default = "default_external_auth_timeout")]
+    #[config_env(leaf)]
+    pub timeout: IggyDuration,
```
hubcio, line 41: warning: `timeout = "unlimited"` (also `0`, `none`, `disabled`) parses to `Duration::ZERO`, and the compio timeout then fires on the first poll. with the default `on_error = deny` that is a total login outage, and nothing validates it.

core/server/src/dispatch/authz.rs
```diff
@@ -42,30 +42,155 @@ use iggy_binary_protocol::requests::users::GetUserRequest;
...
+    }
+    match session_perms {
+        Some(perms) if check(perms) => Some(Ok(())),
+        _ => Some(Err(IggyError::Unauthorized)),
+    }
+}
+
+/// Check if inline permissions allow sending messages to (stream, topic),
+/// mirroring the `Permissioner::append_messages` inheritance chain.
+pub fn can_send_messages(perms: &Permissions, stream_id: usize, topic_id: usize) -> bool {
```
hubcio, line 73: simplification: these five predicates re-implement `Permissioner` rules by hand. `init_permissions_for_user` builds a one-user permissioner from the same `Permissions`, dropping ~100 lines plus the closure param threaded through the gates. watch the `_ => false` at line 211 - the permissioner arm allows there.

core/server/src/http/handlers.rs
```diff
@@ -1799,6 +1866,103 @@ pub(in crate::http) async fn delete_pat(
...
+        credential_type,
+        credential,
+        username: username.to_owned(),
+        transport: "http".to_owned(),
+        client_address: client_address.to_owned(),
+    };
+    let decision = match try_external_auth(&state.external_auth, request).await {
+        Ok(Some(decision)) => decision,
+        Ok(None) => return None,
+        Err(_) => {
```
hubcio: simplification: `try_external_auth` already applied `on_error` - it returns `Ok(None)` for fallback and `Err` only for deny, so the fallback arm here is unreachable. collapse to `Err(_) => Some(Err(IggyError::Unauthenticated.into()))`.

hubcio (changes requested):
Needs another pass. The callout plumbing is mostly fine; nearly everything below is in the identity and ownership half.

Blockers are inline: the callout follows redirects and resends the credential body, the response cap runs after the body is buffered, HTTP recycles a synthetic id while its JWT is live, the counter is per process under a cluster-wide JWT key, a second `Register` splits identity from permissions, two paths leak ids until the 1M pool is gone, the binary transport skips the existence and `Active` checks HTTP enforces, and the metadata gate denies the consumer-group join/leave config.toml advertises.

Rest, on lines the diff does not cover or too small to inline:

- `permissions_global.rs:38` - all ten global bools are mandatory and none of the three levels sets `deny_unknown_fields`, so a typo silently weakens a grant and an 11th permission breaks every grant service. Parse into a `#[serde(default)]` DTO.
- `dispatch/reads.rs:172` - `GET_ME` ungated: an all-false grant gets client id, transport, peer address, consumer-group memberships.
- `dispatch/reads.rs:288` - `SYNC_CONSUMER_GROUP` ungated. Self-scoped, so nothing crosses callers, but still served to an all-false grant.
- `http/handlers.rs:665` - metrics route takes `_identity` with no permission check. Predates this PR, but the PR creates a principal class with no permissions.
- `http/jwt.rs:130` - trusted issuers guard `user_id == 0` but not the synthetic range, so an issuer mapped there inherits whatever grant holds that id.
- `dispatch/authz.rs:578` - `gate_user_scoped` returns `Ok(())` on an unresolvable target while HTTP runs the predicate, so `GET /users/me` 403s on HTTP and returns not-found on binary.
- `external_auth.rs:160,171` - `url` is marked secret, then printed in full in the boot error and the plain-HTTP warning. Userinfo in the URL lands in the log.
- Synthetic ids are persisted, against "never persisted" in the description: `Register` prepares carry them into the WAL, checkpoints and the recovered client table. The counter restarts at `u32::MAX`, so after a restart the same id means a different principal.
- `common/types/user/mod.rs:25` - the threshold and predicate are server-internal but sit in the published `iggy_common`. The only outside consumer is `core/metadata`, which already depends on `server_common`.
- `session_ops.rs:1317` - external auth picks PAT vs password by first match, while the built-in path below deliberately falls back, with a comment saying the shapes collide. A colliding body fails here and succeeds there.
- `config.toml:1108` - `forward_credentials = true` by default, so enabling the feature POSTs plaintext credentials unless the operator opts out, and the doc block does not say so. Worth flipping to opt-in.
- `http/handlers.rs:198` - `client_address` is the raw TCP peer, so behind a proxy every login reports the proxy. Honour a trusted-proxy header or document which address it is.
- No test covers a login on any transport, either authorization plane, the id lifecycle, or `on_error`. The JSON tests assert on the DTO and never reach the decision mapping, so dropping the required-field checks leaves them green. Three blockers above only show up when you read two transports side by side.

Not blocking: single-use `build` closure at `external_auth.rs:240`; `thread_local` client duplicates `http/jwks.rs:44`; `check_session_permission` takes an `Option` all callers fill; mint and free-list logic written twice; `get_connection`, `collect_expired_sessions`, `is_session_expired` have no callers; `session_permissions_for_user` walks two maps when callers hold the connection id; `stm/authz.rs:140` open-codes `is_synthetic_user_id`; `handlers.rs:1910` re-imports `MetadataHandle`.

core/server/src/external_auth.rs
sansmoraxz: Ouch! Wasn't even aware of that behaviour. It's there in jwks too. Addressing...

hubcio:
just to add: please wait with rebase until #4092 is merged.

sansmoraxz:
OK rethinking this, regarding the synthetic IDs probably better to rearchitect.

I think we can have a synthetic ID provider (mapped to specific ID) and that basically enforces a custom ID that's provider generated. (Of course the provider here has to enforce the unique mapping).

 This can also remove all the extra plumbing necessary for rotating IDs and such. Especially in cluster mode.

sansmoraxz:
> * `permissions_global.rs:38` - all ten global bools are mandatory and none of the three levels sets `deny_unknown_fields`, so a typo silently weakens a grant and an 11th permission breaks every grant service. Parse into a `#[serde(default)]` DTO.

This would likely turn this PR into a breaking change.

hubcio:
alright, you can ignore this comment for now. i'll check this from my side later this week.

sansmoraxz:
OK I rebased and went with a different approach now. There's one dedicated ID reserved for external users (which should not collide with local users) and is configurable (but not persisted). Default value if external users enables  - `u32::MAX`. Server boot should crash if ID collision detected at startup and not after server is booted. For external users the main identifiers are unique strings.

Also the auth flow is now `local -> external` instead of external first. This prevents the potential lag of normal iggy hitting and getting failures on the auth server and then falling back to normal.

Both  single node and cluster modes are validated to work without issues.

_Side note:_ `segment_recovery` tests are currently broken at master which I rebased this from.

sansmoraxz:
Also some pre-exsting lint issues in other languages (go etc.)

Had to skip verify.

core/server/src/http/state.rs
sansmoraxz: not relevant anymore

core/server/src/http/handlers.rs
sansmoraxz: done

core/server/src/external_auth.rs
sansmoraxz: fixed

core/server/src/http/handlers.rs
sansmoraxz: no longer relevant

core/server/src/http/handlers.rs
sansmoraxz: not relevant anymore

core/metadata/src/stm/authz.rs
sansmoraxz: no longer relevant

core/configs/src/server_config/external_auth.rs
sansmoraxz: no longer relevant.

core/server/src/external_auth.rs
sansmoraxz: Am mirroring from `jwks.rs`.

Probably worth sharing a HTTP client instead of having seperate dedicated HTTP clients.

core/server/src/external_auth.rs
sansmoraxz: Fixed. Using a single round_trip future object.

core/server/src/external_auth.rs
sansmoraxz: Logic is flipped now. Anyway probably worth another pass to lift up some common logic from jwks.

Will do after forementioned PR gets merged.

core/configs/src/server_config/external_auth.rs
sansmoraxz: Zero timeout is now rejected at boot.

core/server/src/http/state.rs
sansmoraxz: There should be proper cleanup now.

core/server/src/dispatch/session_ops.rs
sansmoraxz: External auth is now fallen through if local validation fails.

core/server/src/http/handlers.rs
sansmoraxz: Fixed!

core/server/src/http/state.rs
sansmoraxz: Not relevant anymore

core/server/src/dispatch/session_ops.rs
sansmoraxz: Fixed.

core/server/src/session_manager.rs
sansmoraxz: not relevant anymore

sansmoraxz:
Review comments should be addressed now.

core/server/src/http/state.rs
```diff
@@ -283,16 +296,21 @@ impl HttpInner {
...
                             table.insert(key.clone(), Rc::clone(&fresh));
                             (Some(fresh), torn)
                         }
                     };
+                    // Reclaim orphaned grants: either the expiry has passed
+                    // or the session table no longer holds a matching entry
+                    // (the JWT expired and was swept above).
+                    {
+                        let sessions = self.sessions.borrow();
+                        self.session_grants.borrow_mut().retain(|key, grant| {
```
hubcio, line 310: critical: a login grant has no session entry until its holder's first write, so this retain drops it as soon as any other client registers - the holder then gets 403 on every read. drop the session-table clause and let expiry reclaim.

core/metadata/src/stm/authz.rs
```diff
@@ -129,15 +134,39 @@ pub(crate) fn authorize(
...
+    // The external auth inline-grant user may execute the replicated
+    // data-plane ops listed below. Each is also gated at dispatch time by
+    // session-scoped permissions (authorize_partition_op for SendMessages
+    // and consumer offsets; authorize_consumer_group_op for CG join/leave).
+    // Non-replicated reads (PollMessages, GetConsumerOffset) never reach
+    // the STM and are gated only at dispatch time.
+    // SYNC: if a new REPLICATED data-plane operation is added, add it
+    // here AND add a dispatch-time session-scoped permission check in
+    // dispatch/authz.rs or dispatch/mod.rs.
+    if external_auth_user_id.is_some_and(|id| user_id == id) {
```
hubcio, line 157: critical: this reads node-local config inside a replicated apply - a node with external auth off, or a snapshot restart that replays before boot sets the id, denies the join a peer applied and group state diverges. carry the id in replicated state instead.

core/server/src/dispatch/authz.rs
```diff
@@ -128,12 +264,60 @@ where
...
+        }
+        Operation::LeaveConsumerGroup => {
+            let Ok(req) = LeaveConsumerGroupRequest::decode_from(body) else {
+                return None;
+            };
+            (req.stream_id, req.topic_id)
+        }
+        _ => return None,
+    };
+    let (sid, tid) = resolve_topic_scope(shard, &stream_id, &topic_id)?;
```
hubcio, line 302: critical: the `?` here returns `None`, which means proceed, so a join on a topic this shard's mirror has not seen yet reaches the primary unchecked and the STM allow-list admits it. deny on a miss like the other ext-auth gates do.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+    pub expires_at: u64,
+}
+
+/// Call the external auth service and parse the response.
+///
+/// # Errors
+///
+/// Returns [`ExternalAuthError`] on network/timeout/parse failure.
+/// Fail-closed: every error variant denies the login.
+pub async fn callout_external_auth(
```
hubcio, line 279: warning: nothing bounds concurrent callouts - every failed login on an unauthenticated route opens an outbound socket and parks a handler for up to the timeout. cap in-flight callouts the way the partition-write admission guard does.

core/common/src/types/permissions/permissions_global.rs
```diff
@@ -26,6 +26,7 @@ use std::fmt::Display;
 /// Global permissions are applied to all streams.
 /// Stream permissions are applied to a specific stream.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone, Default)]
+#[serde(default)]
```
hubcio, line 29: warning: `#[serde(default)]` here also relaxes the create-user and update-permissions REST bodies, which parse this same type - a body with `global` missing or misspelled now parses as all-false instead of failing. scope the default to the callout's grant type or document the break.

core/server/src/dispatch/authz.rs
```diff
@@ -128,12 +264,60 @@ where
...
+    if can_poll_messages(perms, sid, tid) {
```
hubcio, line 303: warning: join and leave map to `get_topic` in the permissioner, but this tests `can_poll_messages`, which also admits a bare `poll_messages` grant. use `can_read_topic` for both.

core/server/src/dispatch/reads.rs
```diff
@@ -399,12 +411,27 @@ pub(in crate::dispatch) async fn handle_non_replicated_request<B, MJ, S, SB>(
...
-            // any other metadata read.
+            // any other metadata read. External auth sessions with no stream
+            // grants are denied: an all-false inline grant should not learn
+            // transport details or consumer-group memberships.
             if let Err(error) = authorize_and_hold_read(shard, code, watermark, || Ok(())).await {
                 send_non_replicated_deny(shard, &request, transport_client_id, error.as_code())
                     .await;
                 return;
             }
+            if session_perms
```
hubcio, line 422: warning: this checks the grant's shape, not its bits - a global-only `read_streams` grant can list every stream yet GET_ME denies it, while a missing or expired grant passes. gate on read permission, deny an all-false grant, and deny a missing one.

also at line 636.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+    }
+}
+
+impl std::error::Error for ExternalAuthError {}
+
+fn redact_url_userinfo(url: &str) -> std::borrow::Cow<'_, str> {
+    let Some((scheme, rest)) = url.split_once("://") else {
+        return std::borrow::Cow::Borrowed(url);
+    };
+    match rest.split_once('@') {
```
hubcio, line 168: warning: this splits at the first `@`, so a password with one leaks its tail into the boot error and the warn. use `rsplit_once`.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+///
+/// Returns [`ExternalAuthError`] on network/timeout/parse failure.
+/// Fail-closed: every error variant denies the login.
+pub async fn callout_external_auth(
+    config: &ExternalAuthConfig,
+    request: ExternalAuthRequest,
+) -> Result<ExternalAuthDecision, ExternalAuthError> {
+    use futures::StreamExt;
+
+    if let Some(ref cred) = request.credential
```
hubcio, line 285: warning: only `credential` is capped. on HTTP an over-long `username` fails as invalid credentials, which is exactly the arm that calls out, so it rides the whole 2 MB body to the service - cap it the same way.

core/server/src/http/state.rs
```diff
@@ -466,6 +523,9 @@ impl HttpInner {
     /// request re-register cleanly through the barrier.
     pub(in crate::http) fn forget_session(&self, session: &Rc<HttpSession>) {
         let torn = forget_if_same(&mut self.sessions.borrow_mut(), session);
+        if let Some(sk) = crate::http::extractor::SessionKey::from_table_key(&session.key) {
```
hubcio, line 526: warning: eviction is transient, but this drops the grant for good - the bearer keeps a valid token and gets 403 until it expires. keep the grant here and let expiry reclaim it.

core/server/src/http/state.rs
```diff
@@ -193,6 +198,14 @@ pub(in crate::http) struct HttpInner {
     /// Behind `Rc` because the write path records from a detached task that
     /// outlives its handler by design (see `submit_committed`).
     pub(in crate::http) metadata_watermarks: Rc<MetadataWatermarks>,
+    /// External authentication callout config. Shared across all handlers.
+    pub(in crate::http) external_auth: Arc<ExternalAuthConfig>,
+    /// Pending inline-grant permissions keyed by session key (`jwt:{jti}`).
+    /// Populated at login when an inline grant is issued; consumed by the
+    /// session resolution path to attach the grant to the `HttpSession`.
+    pub(in crate::http) session_grants: RefCell<
```
hubcio, line 206: warning: grants live only on the node that issued the token, and the token verifies on every node, so a request that lands elsewhere gets 403. carry the grant in the token or replicate it.

core/server/src/http/handlers.rs
```diff
@@ -1870,6 +1964,110 @@ pub(in crate::http) async fn delete_pat(
     Ok(StatusCode::NO_CONTENT)
 }
 
+/// Try external auth for an HTTP login. Returns `Some(result)` when the
```
hubcio, line 1967: nit: this doc promises `Some`/`None` but the function returns a `Result` and never falls through. drop the stale paragraph.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+            },
+        );
+    }
+    // The callout runs inline on the single-threaded shard reactor; a very
+    // large timeout blocks the entire shard for its duration. Warn above 30 s.
+    if config.timeout.get_duration().as_secs() > 30 {
+        tracing::warn!(
+            timeout = %config.timeout,
+            "external_auth.timeout is unusually large (> 30 s); \
+             the callout blocks the shard reactor for its full duration"
```
hubcio, line 229: nit: an awaited callout parks its own task, not the reactor, so this warning overstates the cost. say it stalls that login for the timeout instead.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+//!
+//! When enabled, the server POSTs credential and connection metadata to an
+//! external HTTP service during login. The service decides whether to grant
+//! access (with inline permissions or by mapping to an existing Iggy user)
+//! or deny it. This module owns the request/response types, the HTTP
+//! callout, and the session-scoped permission carrier.
+//!
+//! Known limitation: there is no per-client rate limit on callout attempts.
+//! Every failed built-in login triggers a callout, so a brute-force attack
+//! can amplify traffic to the external service. The shard reactor's
```
hubcio, line 28: nit: every binary client drains on its own task, so the reactor does not bound callout concurrency either. drop that sentence.

core/simulator/src/replica.rs
```diff
@@ -372,6 +372,12 @@ pub fn new_shard(
     }
     // Same seed the server bootstrap runs after its own replay.
     metadata.seed_applied_frontier_from_consensus();
+    // NOTE: The simulator uses ExternalAuthConfig::default() (enabled: false),
```
hubcio, line 375: nit: backwards - with the id unset the gate routes the reserved user to the permissioner, which denies its join and leave, so the sim diverges from nodes that set it. and there is no `seed_baseline` closure here, the mux is built inline above.

core/server/src/boot/mod.rs
```diff
@@ -624,6 +645,29 @@ async fn shard_main(
     } else {
         (None, None, None, None, (0, 0), None)
     };
+    // Peer shards (Waiter path) and snapshot-restored mux_stm need the
+    // external auth user_id too. The Owner path sets it in seed_baseline
```
hubcio, line 649: nit: only true without a snapshot - recovery skips the seed closure when it restores one, so the id stays unset for that whole replay. fix the comment.

core/server/src/http/handlers.rs
```diff
@@ -1271,6 +1350,10 @@ pub(in crate::http) async fn poll_messages(
                     permissioner.poll_messages(uid, stream_id, topic_id)
                 })
         },
+        |p| {
+            resolve_gate_topic_ids(&state, &stream_id, &topic_id)
+                .is_none_or(|(sid, tid)| can_poll_messages(p, sid, tid))
```
hubcio, line 1355: nit: a resolution miss passes here but the same miss on the binary path denies, so one unknown topic answers 404 over HTTP and 403 over TCP. pick one.

also at line 1441.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+}
+
+/// Try external auth and map the result to a decision the login flow
+/// can act on. Fail-closed: a callout failure (timeout, network error,
+/// bad response) denies the login.
+///
+/// # Errors
+///
+/// Returns [`ExternalAuthError`] when the callout itself failed.
+pub async fn try_external_auth(
```
hubcio, line 407: simplification: this only logs and re-raises, and the binary caller logs the same message again with more context. delete it and put the warn in the HTTP caller, which drops the error today.

core/server/src/http/reads.rs
```diff
@@ -383,10 +395,36 @@ pub(in crate::http) fn resolve_gate_topic_ids(
...
+        let sk = crate::http::extractor::SessionKey::from_table_key(session_key)
+            .ok_or(IggyError::Unauthorized)?;
+        let perms = state
+            .session_grant_permissions(&sk)
+            .ok_or(IggyError::Unauthorized)?;
+        // Fail-closed: for external auth users the session-scoped check below
+        // is the only topic-level gate (the STM authz gate allows all
+        // data-plane ops by op-code). A resolution miss must deny, not fall
+        // through to the handler's own not-found path.
+        let (Ok(wire_stream), Ok(wire_topic)) =
```
hubcio, line 414: simplification: both arms convert and resolve the same two ids. resolve once above the branch and keep only the miss mapping per arm.

core/server/src/http/extractor.rs
```diff
@@ -42,6 +42,35 @@ const BEARER: &str = "Bearer ";
 const JWT_KEY_PREFIX: &str = "jwt:";
 const PAT_KEY_PREFIX: &str = "pat:";
 
+/// Session table key identifying a credential. Two variants so JWTs and PATs
+/// occupy disjoint key spaces.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
hubcio, line 47: simplification: `SessionKey` is formatted to a string on every write and parsed back at four sites, and every `from_table_key` miss branch is dead. return the prefixed string from `resolve_credential` and pass `&str` around.

core/server/src/http/submit.rs
```diff
@@ -406,6 +406,13 @@ pub(in crate::http) async fn partition_write_replicated(
             );
             PartitionWriteError::Unavailable
         })?;
+    let session_perms = (state.external_auth.enabled
```
hubcio, line 409: simplification: this block is repeated verbatim at line 474. pull it into one `HttpInner` helper.

core/server/src/http/jwt.rs
```diff
@@ -79,6 +79,9 @@ pub struct JwtManager {
     /// `[[http.jwt.trusted_issuers]]` is configured, in which case `decode`
     /// takes the JWKS verification path for tokens from these issuers.
     trusted_issuers: HashMap<String, TrustedIssuerConfig>,
+    /// User IDs that trusted issuers must never map to (defense in depth).
+    /// Contains the external auth reserved `user_id` when that feature is enabled.
+    forbidden_user_ids: Vec<u32>,
```
hubcio, line 84: simplification: this holds at most one id - the only writer collects an `Option<u32>`. store the `Option` and compare it at line 282.

core/server/src/external_auth.rs
```diff
@@ -0,0 +1,772 @@
...
+    }
+}
+
+/// Session-scoped permissions from an external auth inline grant.
+/// Carried on the connection/session, never persisted. The permissions
+/// are `Arc`-wrapped so dispatch-time lookups share rather than clone
+/// the full `BTreeMap` tree on every request.
+#[derive(Debug, Clone)]
+pub struct SessionPermissions {
+    pub principal: String,
```
hubcio, line 268: simplification: `principal` is written by both login paths and read by nothing. drop the field.

core/server/src/dispatch/authz.rs
```diff
@@ -51,17 +52,143 @@ use server_common::Message;
...
+        Some(perms) if check(perms) => Some(Ok(())),
+        Some(_) => Some(Err(IggyError::Unauthorized)),
+        None => None,
+    }
+}
+
+/// Check if inline permissions allow sending messages to (stream, topic),
+/// mirroring the `Permissioner::append_messages` inheritance chain.
+// SYNC: keep in lockstep with Permissioner::append_messages
+pub fn can_send_messages(perms: &Permissions, stream_id: usize, topic_id: usize) -> bool {
```
hubcio, line 73: simplification: these five predicates re-implement the permissioner rules by hand and already drift, see the join/leave gate. pull the rule bodies out as free functions over `Option<&GlobalPermissions>` and `Option<&StreamPermissions>` and call them from both sides.

core/server/src/dispatch/reads.rs
```diff
@@ -565,17 +613,39 @@ pub(in crate::dispatch) async fn handle_non_replicated_request<B, MJ, S, SB>(
...
             // header client id, so it carries no permissioner rule. The
             // assignment itself is metadata-STM state, hence the gate.
+            // External auth sessions with no stream grants are denied (same
+            // rationale as GET_ME above).
             if let Err(error) = authorize_and_hold_read(shard, code, watermark, || Ok(())).await {
                 send_non_replicated_deny(shard, &request, transport_client_id, error.as_code())
                     .await;
                 return;
             }
+            if session_perms
```
hubcio, line 636: simplification: same predicate as line 422. one helper, so the fix lands once.

core/server/src/dispatch/authz.rs
```diff
@@ -128,12 +264,60 @@ where
...
+            };
+            (req.stream_id, req.topic_id)
+        }
+        Operation::LeaveConsumerGroup => {
+            let Ok(req) = LeaveConsumerGroupRequest::decode_from(body) else {
+                return None;
+            };
+            (req.stream_id, req.topic_id)
+        }
+        _ => return None,
```
hubcio, line 300: simplification: unreachable, the caller only routes join and leave here. drop the arm.

hubcio (changes requested):
the description does not match the code: `on_error` is not a config key and the server drops it silently, there is no synthetic id pool or rotation - one fixed `user_id` with default `u32::MAX`, which the description omits - `transport` can also be `"binary"`, and `credential` is only sent when `forward_credentials` is true. please fix it before the squash merge, since it becomes the commit message.

`core/server/src/http.rs:289` - nit: the login handlers now send this address to the callout, which can deny on it, so 'never authorization' no longer holds. say it also reaches the callout.

NOTE: this review was generated using skill `/team-review-slim` from this repository and hundreds of my personal markdowns that i use to work with iggy. i manually fact-checked criticals and warnings.

core/server/src/http/state.rs
sansmoraxz: removed the sessions.contains_key clause that dropped grants before their holder's first write; now only expiry reclaims

core/server/src/dispatch/authz.rs
sansmoraxz: changed to `Unauthorized`

core/server/src/dispatch/authz.rs
sansmoraxz: Unauthorized

core/server/src/dispatch/reads.rs
sansmoraxz: agrreed

core/server/src/http/jwt.rs
sansmoraxz: It's better to be a Vec, root and external users are both denied. Also this maybe expanded in future so rather than a breaking change current behaviour makes more sense.

core/common/src/types/permissions/permissions_global.rs
sansmoraxz: ok reverted. thought some more on this better to have this as a non-breaking change. just that the ext auth implementations has to be exhaustive.

Issue #3848, opened 2026-08-08
fix(connectors): restart loads the newest connector config, not the active one


mlevkov:
## Summary

On the default local configuration provider, a connector runs its **active** configuration after a process start and its **highest-numbered** configuration after `POST /{sinks,sources}/{key}/restart`. Rolling back with `PUT .../configs/active` therefore appears to work until the next restart, which silently reinstates the newest version.

The HTTP provider does not have this split.

## Where the two paths diverge

Startup resolves the active configuration:

```rust
// core/connectors/runtime/src/main.rs:153
let connectors_config = connectors_config_provider.get_active_configs().await?;
```

Restart asks for `None` instead:

```rust
// core/connectors/runtime/src/manager/sink.rs:261   (source.rs:280 is identical)
.get_sink_config(key, None)
```

and the local provider resolves `None` to the highest version number rather than the active one:

```rust
// core/connectors/runtime/src/configs/connectors/local_provider.rs
async fn get_sink_config(&self, key: &str, version: Option<u64>) -> ... {
    if let Some(version) = version {
        ...
    } else {
        Ok(self
            .get_sink_configs(key)
            .await?
            .into_iter()
            .max_by_key(|config| config.version))
    }
}
```

`get_source_config` is the same. Meanwhile the HTTP provider resolves the identical `None` to the active config:

```rust
// core/connectors/runtime/src/configs/connectors/http_provider.rs:291
None => self.url_builder.build(TemplateKeys::GET_ACTIVE_SINK_CONFIG, &vars),
```

So `version: None` means "active" in one provider and "newest" in the other, and the restart path is the caller that notices.

## The local provider already tracks this

It is not that the local provider lacks the concept. `set_active_sink_version` persists it, and `get_active_configs` reads it back and selects on it:

```rust
let active_config = if let Some(&version) = active_versions.sinks.get(key) {
    config_files.iter().find(|c| c.config.version == version)
```

Only the `get_{sink,source}_config(key, None)` accessor skips it, which is exactly the one `restart_connector` uses.

## Why it matters

The intended rollback flow does not survive a restart:

1. `POST /sinks/{key}/configs` publishes v2, which turns out to be bad.
2. `PUT /sinks/{key}/configs/active` sets v1 active. `GET .../configs/active` confirms v1, and a process restart runs v1.
3. `POST /sinks/{key}/restart` runs **v2** again.

An operator rolling back an incident sees the rollback take effect, then get undone by the very call they would reach for to apply it.

[...]

PR #4064, opened 2026-09-05
fix(connectors): close the source instance a failed start leaves behind

### LINKED ISSUE #4062: connectors: source restart leaks the plugin instance when producer setup fails
## What happens

`SourceManager::start_connector`, which `restart_connector` calls after stopping the old instance, calls `source::init_source` and then `source::setup_source_producer`:

```rust
// core/connectors/runtime/src/manager/source.rs
source::init_source(&container, &config.plugin_config.clone().unwrap_or_default(), plugin_id, state)?;   // :232
info!("Source connector with ID: {plugin_id} for plugin: {key} initialized successfully.");

let (producer, encoder, transforms) =
    source::setup_source_producer(key, config, iggy_client).await?;                                       // :241
```

When `setup_source_producer` fails, the `?` returns with `init_source` already succeeded, so the plugin holds an initialized instance under `plugin_id` that nothing will ever close. `details.info.id = plugin_id` is not reached until `:261`, so the stored id still names the previous instance and a later stop closes that one instead. The orphan lives for the rest of the process.

The boot path handles the identical failure correctly, and is the model for the fix:

```rust
// core/connectors/runtime/src/source.rs, the setup_source_producer error arm
Err(error) => {                                                                                          // :245
    let message = format!("Failed to set up source producer: {error}");
    error!("Source: {name} ({key}) - {message}");
    let connector = source_connectors.get_mut(&path).expect("source connector was inserted above");
    let close_result = (connector.container.iggy_source_close)(plugin_id);                                // :251
    ...
}
```

## Why it is worth fixing rather than tolerating

For a plugin whose `init_source` only allocates, an orphan is wasted memory. For one that acquires a process-global resource in `open()`, it is a live fault. The HTTP source connector in #3798 is the worked example: instances share one listener keyed by address, so the orphan stays joined and keeps answering `200` into a bridge that no `poll()` drains, holds `/health` at `503` because a joined instance is not polling, and makes every subsequent restart attempt fail on the duplicate `instance_name` it never released. A transient broker blip at restart time turns into an instance that is unreachable until the whole runtime is bounced.

## Fix

Mirror the boot path: close the plugin instance on the `setup_source_producer` failure branch before returning. Setting `details.info.id` before the fallible call, or carrying the id out of the error arm, is what keeps a later stop from closing the wrong instance.

One thing that does **not** work, in case it looks tempting: evicting on a `poll_is_live`-style signal from the plugin side. That flag is false for the whole producer-setup window, so a plugin cannot distinguish "setup still in progress" from "setup failed and I have been orphaned".

## Provenance

Raised by @hubcio in review on #3798, twice, as out of scope for that PR, with the fix above. Filing it so it is tracked once #3798 merges and its review thread stops being where this lives.

### Contribution

mlevkov:
Closes #4062.

## The leak

`SourceManager::start_connector` takes a fresh `plugin_id`, calls `init_source`, and records that id on `SourceDetails` only after the handler tasks are spawned. In between, the instance exists inside the plugin and nothing outside it knows the id: `stop_connector` closes whatever `details.info.id` holds, which is still the previous instance. `setup_source_producer` returning early through `?` therefore stranded the new one for the life of the process. `source::init` already cleaned up on the identical failure, which is the asymmetry @hubcio pointed at.

For a plugin whose open only allocates, the orphan is wasted memory. For one that takes a process-global resource it is a live fault: a shared listener stays bound and answering into a queue nothing drains, and every retried restart then fails on the identity the orphan never released.

## A guard, not a cleanup branch

This deviates from the fix in the review, which was to mirror `source::init`'s error arm at the call site, so it is worth saying why rather than leaving it to be found.

The window is defined by the two statements that open the instance and record its id, not by which call between them happens to be fallible today. A cleanup branch is correct only for the one `?` that exists now, and silently wrong for the next one somebody adds. `SourceInstanceGuard` is armed at `init_source` and disarmed once the id is recorded, so every path out of that window closes the instance, including a panic.

It also made the behaviour testable. `Container<SourceApi>` only comes from `dlopen`, so `start_connector` cannot be exercised in a unit test at all, while a guard holding the bare `extern "C" fn` can be driven directly.

The close-and-report itself is now one function shared with `source::init`, so the two sites cannot drift. `source::init` keeps its existing control flow; only the duplicated body moved.

No `cleanup_sender` on this path: `spawn_source_handler` is what registers the sender, and it has not run yet.

## Tests

Three, each mutation-checked, each mutant confirmed to compile first:

- an armed guard closes, and closes *its own* id, since closing another would leave this instance open and tear down a live one
- a disarmed guard does not close, or a source that just started successfully would be torn down
- a refused close (`-1`, the code the SDK returns for an unknown id) is reported and not propagated, because unwinding out of `drop` would be worse than the leak it is cleaning up after

Each test owns its stub and statics rather than sharing a pair, which would have made two of them race in the same process.

## What is not covered, and why

**The guard's placement in `start_connector` has no test.** I verified that rather than assuming it: disarming the guard immediately after construction restores the original leak, compiles, and the suite still passes.

Reaching that path needs a real `Container`, so it cannot be a unit test, and the only route into `start_connector` is `POST /sources/{key}/restart`. Making `setup_source_producer` fail there means either a config the local provider will serve on restart but not at boot, which today works only because of the version selection in #3848 and would break when that is fixed, or stopping the broker mid-test. Both couple this regression test to something unrelated to it, so I left it out rather than write a test that fails for the wrong reason later. Happy to add either if you would rather have the coverage than the independence.

`source::init`'s cleanup remains covered only by `error_isolation.rs` asserting the connector reports `Error`, which it did before this change too.

## Verification

`cargo fmt`, `cargo sort --no-format`, clippy at both feature sets, rustdoc under `-D warnings`, 198 unit tests in `iggy-connectors`, and `stdout_sink` + `random_source` still build.


core/connectors/runtime/src/manager/source.rs
```diff
@@ -265,6 +271,8 @@ impl SourceManager {
             details.handler_tasks = handler_tasks;
             metrics.increment_sources_running();
         }
+        // `details.info.id` now names this instance, so a later stop reaches it.
+        instance.disarm();
```
hubcio: warning: cancel at the `details.lock().await` above (client disconnect drops the axum handler future) and the guard closes the instance but leaves the `SOURCE_SENDERS` entry and both spawned tasks behind, so the forwarding loop runs forever. take the lock before `spawn_source_handler` and set `info.id` there.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -265,6 +271,8 @@ impl SourceManager {
             details.handler_tasks = handler_tasks;
             metrics.increment_sources_running();
```
hubcio: warning: the forwarding loop's first `update_status(Running)` already bumped this gauge, and stop only decrements once, so `sources_running` ratchets up per restart. drop the direct status write and this increment, let `update_status` own it.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+    /// close it. Call only after the plugin id is durably recorded.
+    pub(crate) fn disarm(mut self) {
+        self.armed = false;
+    }
+}
+
+impl Drop for SourceInstanceGuard<'_> {
+    fn drop(&mut self) {
+        if self.armed {
+            close_failed_source(self.close, self.plugin_id, self.key);
```
hubcio: warning: this can fire after `spawn_source_handler` ran, so `iggy_source_close` hits `block_on(handle)` and `block_on(source.close())` - unbounded plugin teardown on a tokio worker inside drop glue, where no timeout fits. either document that contract on the type or move cleanup to an explicit `finish()` on the error arms.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -236,6 +236,12 @@ impl SourceManager {
             state,
         )?;
         info!("Source connector with ID: {plugin_id} for plugin: {key} initialized successfully.");
+        // Armed from here until the id is recorded below. Until then nothing
+        // outside the plugin knows this instance exists, so any early return
+        // would strand it: `stop_connector` closes `details.info.id`, which
+        // still names the previous one.
+        let instance =
```
hubcio: nit: `instance` holds a guard, so `instance.disarm()` below reads as disarming the instance. `instance_guard` matches `shutdown_guard` and `tmp_guard` elsewhere in the repo.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+/// it knows the id: `stop_connector` closes whatever `details.info.id` holds,
+/// which is still the previous instance. An early return in that window
+/// stranded the new one for the life of the process.
+///
+/// A guard rather than a cleanup branch at the one call site that can fail
+/// today, because the window is defined by the two statements that open and
+/// record the instance, not by which call between them happens to be fallible.
+/// Adding a `?` inside it stays correct.
+pub(crate) struct SourceInstanceGuard<'a> {
+    close: extern "C" fn(u32) -> i32,
```
hubcio: nit: the fn pointer has no lifetime tie to the `Container` that owns the `.so` - it works only because `container` is declared before the guard and so drops after it. worth stating that in the doc comment.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+/// `SourceDetails`, the instance exists inside the plugin and nothing outside
+/// it knows the id: `stop_connector` closes whatever `details.info.id` holds,
+/// which is still the previous instance. An early return in that window
+/// stranded the new one for the life of the process.
+///
+/// A guard rather than a cleanup branch at the one call site that can fail
+/// today, because the window is defined by the two statements that open and
+/// record the instance, not by which call between them happens to be fallible.
+/// Adding a `?` inside it stays correct.
+pub(crate) struct SourceInstanceGuard<'a> {
```
hubcio: nit: add `#[must_use = "dropping an armed guard closes the source instance"]`. a bare `SourceInstanceGuard::new(..);` statement closes the instance on the spot, and with `warnings = "deny"` the attribute turns that into a build error.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+        Self {
+            close,
+            plugin_id,
+            key,
+            armed: true,
+        }
+    }
+
+    /// Hands ownership of the instance to the caller, once something else can
+    /// close it. Call only after the plugin id is durably recorded.
```
hubcio: nit: "durably recorded" promises a disk guarantee that doesn't exist - `SourceDetails` is memory only and `plugin_id` never reaches the state store. the struct doc above already says "recorded on `SourceDetails`", reuse that.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+        if self.armed {
+            close_failed_source(self.close, self.plugin_id, self.key);
+        }
+    }
+}
+
+/// Closes an instance whose setup did not finish, reporting a refusal rather
+/// than returning it: both callers are already on a failure path and have an
+/// error of their own to surface.
+pub(crate) fn close_failed_source(close: extern "C" fn(u32) -> i32, plugin_id: u32, key: &str) {
```
hubcio: nit: this closes a plugin instance, not a source connector - `close_source_instance` reads right, and would also suit the stop path at `manager/source.rs:167`. it can be a plain `fn`, both callers are in this file.

core/connectors/runtime/src/source.rs
```diff
@@ -961,6 +1013,88 @@ mod tests {
...
+
+    static REFUSED_CALLS: AtomicU32 = AtomicU32::new(0);
+
+    extern "C" fn refusing_close(_id: u32) -> i32 {
+        REFUSED_CALLS.fetch_add(1, Ordering::SeqCst);
+        -1
+    }
+
+    #[test]
+    fn given_an_armed_guard_when_dropped_should_close_the_instance() {
```
hubcio: nit: the articles are new here - every other test in the module reads `given_serialized_batch_...`, `given_missing_sender_...`. `given_armed_guard_when_dropped_should_close_instance` and siblings.

also at lines 1066, 1082.

core/connectors/runtime/src/source.rs
```diff
@@ -961,6 +1013,88 @@ mod tests {
...
+
+        assert_eq!(
+            DISARMED_CALLS.load(Ordering::SeqCst),
+            0,
+            "the instance is the manager's once its id is recorded"
+        );
+    }
+
+    #[test]
+    fn given_a_refused_close_when_guard_drops_should_not_panic() {
```
hubcio: nit: the harness gives no-panic for free. what line 1095 actually proves is that close ran once and the -1 was swallowed - `given_refused_close_when_guard_drops_should_close_once_and_swallow_refusal`.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+/// Closes a source instance that `iggy_source_open` created and nothing else
+/// will ever reach.
+///
+/// Between `init_source` succeeding and the plugin id being recorded on
+/// `SourceDetails`, the instance exists inside the plugin and nothing outside
+/// it knows the id: `stop_connector` closes whatever `details.info.id` holds,
+/// which is still the previous instance. An early return in that window
+/// stranded the new one for the life of the process.
+///
+/// A guard rather than a cleanup branch at the one call site that can fail
```
hubcio: nit: `init` at line 251 is exactly the cleanup branch this paragraph argues against, and it's still there. use the guard there too, or cut the paragraph.

core/connectors/runtime/src/source.rs
```diff
@@ -961,6 +1013,88 @@ mod tests {
...
+        );
+        assert_eq!(
+            ARMED_CLOSED_ID.load(Ordering::SeqCst),
+            plugin_id,
+            "closing any other id would leave this instance open and kill a live one"
+        );
+    }
+
+    #[test]
+    fn given_a_disarmed_guard_when_dropped_should_leave_the_instance_open() {
```
hubcio: nit: this passes even if the guard is never built - `DISARMED_CALLS` starts at 0 and nothing else touches it. close an armed guard through the same stub first, then disarm one and assert the count didn't move.

core/connectors/runtime/src/source.rs
```diff
@@ -961,6 +1013,88 @@ mod tests {
...
+    }
+
+    static REFUSED_CALLS: AtomicU32 = AtomicU32::new(0);
+
+    extern "C" fn refusing_close(_id: u32) -> i32 {
+        REFUSED_CALLS.fetch_add(1, Ordering::SeqCst);
+        -1
+    }
+
+    #[test]
```
hubcio: nit: all three tests drop or disarm inline, which the compiler already guarantees. none covers the shape the guard exists for - a `?` returning early with the guard still armed.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+///
+/// A guard rather than a cleanup branch at the one call site that can fail
+/// today, because the window is defined by the two statements that open and
+/// record the instance, not by which call between them happens to be fallible.
+/// Adding a `?` inside it stays correct.
+pub(crate) struct SourceInstanceGuard<'a> {
+    close: extern "C" fn(u32) -> i32,
+    plugin_id: u32,
+    key: &'a str,
+    armed: bool,
```
hubcio: simplification: `armed` only models "consumed". `disarm(self) { std::mem::forget(self) }` with an unconditional `Drop` removes the field, the init and the branch - every field is `Copy` so nothing leaks. same shape as `AVecRawGuard::defuse`.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
     }
 }
 
+/// Closes a source instance that `iggy_source_open` created and nothing else
+/// will ever reach.
+///
+/// Between `init_source` succeeding and the plugin id being recorded on
```
hubcio: simplification: this paragraph is repeated almost word for word at `manager/source.rs:239-242`. keep it here and cut the call-site copy to the one fact it adds.

core/connectors/runtime/src/source.rs
```diff
@@ -961,6 +1013,88 @@ mod tests {
         }
     }
 
+    /// Records what `SourceInstanceGuard` passed to the FFI. A stub has to be a
```
hubcio: simplification: `next_plugin_id()` already hands each test a unique id, so a shared recorder can't race - the caveat guards against a design nobody's using. one id-keyed map plus `ok_close` and `refusing_close` replaces three stubs and four statics.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+/// than returning it: both callers are already on a failure path and have an
+/// error of their own to surface.
+pub(crate) fn close_failed_source(close: extern "C" fn(u32) -> i32, plugin_id: u32, key: &str) {
+    let close_result = close(plugin_id);
+    if close_result != 0 {
+        warn!(
+            "iggy_source_close returned {close_result} while cleaning up failed source connector with ID: {plugin_id} ({key})"
+        );
+    }
+}
```
hubcio: simplification: `sink.rs:185-190` still has this body inline, one word apart. both close pointers are `extern "C" fn(u32) -> i32`, so one helper taking a "source"/"sink" label covers both.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +300,63 @@ pub(crate) fn init_source(
...
+/// stranded the new one for the life of the process.
+///
+/// A guard rather than a cleanup branch at the one call site that can fail
+/// today, because the window is defined by the two statements that open and
+/// record the instance, not by which call between them happens to be fallible.
+/// Adding a `?` inside it stays correct.
+pub(crate) struct SourceInstanceGuard<'a> {
+    close: extern "C" fn(u32) -> i32,
+    plugin_id: u32,
+    key: &'a str,
```
hubcio: simplification: `key` exists to label one `warn!`, and it drags in the lifetime param and the `impl<'a>`. both call sites already log the key on the same failure, and `plugin_id` identifies the instance.

core/connectors/runtime/src/manager/source.rs
hubcio: warning: cancel at the `details.lock().await` above (client disconnect drops the axum handler future) and the guard closes the instance but leaves the `SOURCE_SENDERS` entry and both spawned tasks behind, so the forwarding loop runs forever. take the lock before `spawn_source_handler` and set `info.id` there.

core/connectors/runtime/src/manager/source.rs
hubcio: warning: the forwarding loop's first `update_status(Running)` already bumped this gauge, and stop only decrements once, so `sources_running` ratchets up per restart. drop the direct status write and this increment, let `update_status` own it.

core/connectors/runtime/src/source.rs
hubcio: warning: this can fire after `spawn_source_handler` ran, so `iggy_source_close` hits `block_on(handle)` and `block_on(source.close())` - unbounded plugin teardown on a tokio worker inside drop glue, where no timeout fits. either document that contract on the type or move cleanup to an explicit `finish()` on the error arms.

core/connectors/runtime/src/manager/source.rs
hubcio: nit: `instance` holds a guard, so `instance.disarm()` below reads as disarming the instance. `instance_guard` matches `shutdown_guard` and `tmp_guard` elsewhere in the repo.

core/connectors/runtime/src/source.rs
hubcio: nit: the fn pointer has no lifetime tie to the `Container` that owns the `.so` - it works only because `container` is declared before the guard and so drops after it. hold an `Arc<Container<SourceApi>>` and read `iggy_source_close` inside `drop`, rather than leaning on declaration order.

core/connectors/runtime/src/source.rs
hubcio: nit: add `#[must_use = "dropping an armed guard closes the source instance"]`. a bare `SourceInstanceGuard::new(..);` statement closes the instance on the spot, and with `warnings = "deny"` the attribute turns that into a build error.

core/connectors/runtime/src/source.rs
hubcio: nit: "durably recorded" promises a disk guarantee that doesn't exist - `SourceDetails` is memory only and `plugin_id` never reaches the state store. the struct doc above already says "recorded on `SourceDetails`", reuse that.

core/connectors/runtime/src/source.rs
hubcio: nit: this closes a plugin instance, not a source connector - `close_source_instance` reads right, and would also suit the stop path at `manager/source.rs:167`. it can be a plain `fn`, both callers are in this file.

core/connectors/runtime/src/source.rs
hubcio: nit: the articles are new here - every other test in the module reads `given_serialized_batch_...`, `given_missing_sender_...`. `given_armed_guard_when_dropped_should_close_instance` and siblings.

also at lines 1066, 1082.

core/connectors/runtime/src/source.rs
hubcio: nit: the harness gives no-panic for free. what line 1095 actually proves is that close ran once and the -1 was swallowed - `given_refused_close_when_guard_drops_should_close_once_and_swallow_refusal`.

core/connectors/runtime/src/source.rs
hubcio: nit: `init` at line 251 is exactly the cleanup branch this paragraph argues against, and it's still there. use the guard there too, or cut the paragraph.

core/connectors/runtime/src/source.rs
hubcio: nit: this passes even if the guard is never built - `DISARMED_CALLS` starts at 0 and nothing else touches it. close an armed guard through the same stub first, then disarm one and assert the count didn't move.

core/connectors/runtime/src/source.rs
hubcio, line 1127: nit: all three tests drop or disarm inline, which the compiler already guarantees. none covers the shape the guard exists for - a `?` returning early with the guard still armed.

core/connectors/runtime/src/source.rs
hubcio, line 322: simplification: this paragraph is repeated almost word for word at `manager/source.rs:239-242`. keep it here and cut the call-site copy to the one fact it adds.

core/connectors/runtime/src/source.rs
hubcio: simplification: `next_plugin_id()` already hands each test a unique id, so a shared recorder can't race - the caveat guards against a design nobody's using. one id-keyed map plus `ok_close` and `refusing_close` replaces three stubs and four statics.

core/connectors/runtime/src/source.rs
hubcio, line 427: simplification: `sink.rs:185-190` still has this body inline, one word apart. both close pointers are `extern "C" fn(u32) -> i32`, so one helper taking a "source"/"sink" label covers both.

core/connectors/runtime/src/source.rs
hubcio: simplification: `key` exists to label one `warn!`, and it drags in the lifetime param and the `impl<'a>`. both call sites already log the key on the same failure, and `plugin_id` identifies the instance.

hubcio (changes requested):
a few things outside the diff, none of them blocking this PR:

- `core/connectors/sdk/src/source.rs:597` - `iggy_source_open` inserts into `INSTANCES` even when `open()` failed, and `SourceContainer::open` stores the source before returning 1. a plugin whose `open()` bound a listener and then errored keeps it for the life of the process. same leak class as this PR, one statement earlier. the fix is to skip the insert and drop the container, not to close from the runtime side - the SDK already stored the source, so that would run `Source::close()` on an instance that never opened.
- `core/connectors/runtime/src/source.rs:71` - `SOURCE_SENDERS` being a process global is why an orphaned forwarding loop can't die. the sink owns its `watch::Sender` in `SinkDetails`, so its identical window self-heals. an RAII registration stored next to `handler_tasks` would fix the source side properly.
- `core/connectors/runtime/src/manager/sink.rs:200-223` - same unrecorded-instance window on the sink path, `details.info.id` only set at 223. narrower than the source case, since the consume tasks exit when the `watch::Sender` drops, so the same guard alone is enough there.
- `core/connectors/runtime/src/manager/source.rs:186-190` - `stop_connector` never clears `details.info.id`, so after a failed start every later stop re-closes a dead id and line 168 logs "Closed" for it. the shutdown sweep hits this too.
- `core/connectors/runtime/src/manager/source.rs:167` - the stop path drops the `iggy_source_close` result and logs "Closed" unconditionally. a -1 there means teardown was skipped while the `INSTANCES` entry is already gone, so nothing can retry.
- `core/connectors/runtime/src/manager/source.rs:311` - a failed restart leaves the connector Stopped with `last_error` cleared, so `GET /sources` shows nothing wrong. one `set_error` on `start_connector(..).await?` covers all five fallible steps.
- `core/connectors/runtime/src/source.rs:384-427` - `setup_source_producer` builds and `init()`s a producer per configured stream but keeps only the last, so two configured streams silently produce to one.

core/connectors/runtime/src/source.rs
hubcio: one more option, if you take the `Arc` suggestion on line 317: move that `Arc<Container<SourceApi>>` into a `spawn_blocking` and read `iggy_source_close` there. `Container` is `Send + Sync`, so the task keeps the library mapped and the close stops parking a worker. tradeoff is the teardown becomes unordered against the `Err` return.

mlevkov:
All 18 inline findings are in, at `898c6a9a1`. Ten commits, one per finding
where that made sense. Gate green: fmt, sort, clippy at both feature sets,
rustdoc `-D warnings`, `--locked`, **212 unit tests** (was 206) over three clean
runs, plus **64 connectors integration tests** across `runtime`, `api`, `random`,
`stdout`, `http_config_provider` and `postgres`.

| batch | findings | commit |
| --- | --- | --- |
| A3 | close pointer tied to its library | `1f2c2a490` |
| A2 and its follow-up | teardown off the worker | `20d03bc7d` |
| A1 | cancelled start leaking both tasks | `a1722c5bf` |
| A4 | the `sources_running` ratchet | `abacd38d2` |
| B1-B4 | the guard tests | `cbc4a31e0` |
| C1, C5, C8 | one close for a plugin instance | `0dcf36957` |
| C2, C3, C4, C6 | naming, `must_use`, docs | `9c7bb0d21` |
| A4 follow-up | a gauge test, and what it showed | `d90b72fa5` |
| A1 follow-up | an await in the start window is now a build error | `0c4b859f8` |
| @spetz | the status transition and the gauge move together | `898c6a9a1` |

You were right that the fix was incomplete both ways. A1 is a bit worse than
"the tasks are left behind": dropping the `JoinHandle`s detaches them, so the
forwarding loop keeps running against a channel nothing feeds. The lock now
comes before the spawn, so nothing can await between registering the tasks and
recording the id that reaches them.

Rebased onto master twice since this was posted, so the shas above are the
current ones. A tenth commit answers @spetz's review; see the reply below.

### Five places I did not do exactly what you asked

**A3, the `Arc`.** Held as `Arc<dyn Fn(u32) -> i32 + Send + Sync>`, built by a
`for_container` constructor that captures the `Arc<Container<SourceApi>>`,
rather than as an `Arc<Container<SourceApi>>` field. A `Container` only exists
via dlopen, so a guard carrying that field cannot be built in a unit test at
all, and B1 and B2 ask for more guard tests, not fewer. The single production
constructor takes the container, so the ownership stays enforced by the API.
Same lifetime tie, still testable.

**A2, both of your options rather than one.** `drop` defers to the blocking
pool, and the error arms await an explicit `close()`. Deferring on its own costs
the ordering you named, and that ordering matters here: a restart retried inside
the teardown window would collide with the instance being torn down, which is
the failure #4062 exists to prevent. `drop` stays the net for a cancellation and
for any `?` added in the window later.

**A4, reported rather than deleted.** Dropping the status write outright leaves a
connector that just started answering `Stopped` to `GET /sources` until the
forwarding loop takes the lock. `start_connector` now reports through
`update_status`, so the gauge still has exactly one owner.

**C7, the `key` field.** The lifetime param and the `impl<'a>` are gone, which
was the objection. `key` stays, as an owned `String`: now that the close can be
deferred, its `warn!` is emitted outside the caller's log context, so the key is
the only thing tying that line to a connector. Happy to drop it for `plugin_id`
alone if you would rather.

**C1, not the stop path.** You noted the helper would also suit
`manager/source.rs:167`. That path drops the `iggy_source_close` result and logs
"Closed" regardless, which is your own out-of-diff finding, so wiring it in here
would fix a separate bug inside this PR.

### What the integration tests do and do not reach

They do reach more than I first credited them with. `source_with_invalid_config`
in `error_isolation` fails inside `setup_source_producer`, which is the arm the
guard now sits on in `init`, and `random_source_produces_messages` is what would
break if the `disarm()` in the other arm were wrong. For the restart path,
`given_restart_when_state_exists_should_resume_from_served_state` and the two
CDC restart tests POST `/sources/{key}/restart` and then wait for the source to
reload state and resume, so A1's restructure is exercised end to end.

Two things are still not pinned by any test:

- **A1's cancellation window.** Nothing can drop the `start_connector` future at
  that one await, so it still has no test. It is no longer held by a comment
  either: `SourceDetails::record_started` takes the spawn as a closure and is
  deliberately not `async`, so an await added between the spawn and the id
  record does not compile. Checked by adding one, `error[E0728]`. Same argument
  you made for `#[must_use]`, applied to the window rather than the guard. It
  also made the step testable, which nothing in that region was.
- **A4's arithmetic.** There is now a test for it, and it is worth telling you
  what it showed. `sources_running_does_not_climb_across_restarts` restarts the
  random source three times and requires the gauge back at 1. It passes, but it
  does not isolate the fix: revert the report through `update_status` and it
  still passes 3 of 3, because taking the lock before the spawn already makes
  `start_connector` win the race to that lock. Revert that too, so the code is
  what you reviewed, and it still passes 6 of 6. `start_connector` takes an
  uncontended mutex immediately, while the spawned loop has to be scheduled and
  walk a DashMap to reach the same lock, so it does not get there first.

  So the double count is real in the code and, as far as I can make it behave,
  latent. The fix stands on not wanting correctness to rest on that lock
  ordering, not on a ratchet I could reproduce. Said plainly rather than let the
  commit imply I had seen it climb.

Everything else was mutation-checked, with each mutant confirmed to compile
first: closing inline instead of deferring, `close()` not awaiting (10 rounds of
10), `disarm()` moved ahead of the fallible step, a guard that never arms, the
transition guard removed from `update_status`, and a bare
`SourceInstanceGuard::new(..);` statement, which `#[must_use]` now rejects at
build time.

### Out of diff

Filed the multi-stream producer bug as #4097. `setup_source_producer` `init()`s a
producer per `[[streams]]` entry and keeps only the last, so two configured
streams silently produce to one. Both auto-create flags default on, so every
configured stream and topic really is created and all but the last stay empty,
which is what makes it hard to spot. The other six are not filed yet.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -236,35 +236,65 @@ impl SourceManager {
...
+        // `details.info.id` now names this instance, so a later stop reaches it.
+        instance_guard.disarm();
+
+        // Through `update_status`, which is the only thing that moves the gauge
+        // and moves it only on a real transition. Writing the status here and
+        // incrementing by hand as well meant two mechanisms counting one start:
+        // the forwarding loop reports `Running` too, for the boot path that
+        // never comes through here, and a stop decrements once. Reported twice
+        // and taken back once, the gauge climbed with every restart.
+        self.update_status(key, ConnectorStatus::Running, Some(metrics))
```
spetz: This `update_status(Running)` runs after releasing the startup lock, so the forwarding loop can report `Running` and then call `set_error` for its first batch before this call executes. The delayed update overwrites `Error`, clears `last_error`, and increments `sources_running` again because `set_error` does not decrement it.

A controlled interleaving reproduced `Running`, no `last_error`, and `sources_running = 2` for one instance. Performing the initial transition under the startup lock preserved `Error`, its details, and a count of `1`.

Please apply the initial status transition under the existing `details` lock, sharing the transition logic with `update_status`, and remove this trailing unconditional update. Add a regression test with an error between the two `Running` reports. The current consecutive-`Running` test cannot catch this ordering

spetz (changes requested):

mlevkov:
Fixed at `898c6a9a1`, rebased onto master first. 215 unit tests and 9 connectors
integration tests green.

You were right, and the second half of your sentence is the part I had missed.
`set_error` moved the status out of `Running` without moving the gauge, so the
increment was never given back. That is why the count reached 2 rather than
merely being set late, and it also means the loop's later `Stopped` could not
correct it: by then the old status was `Error`, and neither branch of the gauge
rule fires.

`SourceDetails::apply_status` now owns the transition and the gauge move
together. `update_status` takes the lock and delegates, `set_error` delegates and
then sets its message, and `start_connector` calls it in the same hold as the id
record. The trailing unconditional report is gone.

The regression test puts the error between the two `Running` reports. Removing
the gauge move from `set_error` leaves the gauge reading 1 where it should read
0, and the test fails on that. You were right that the consecutive-`Running`
test cannot reach this: there the second report finds the status already
`Running`, so nothing crosses and no arithmetic runs.

One thing I did not change and one I could not test.

`SinkManager::set_error` has the same gauge asymmetry. It is outside this PR's
two files, and the sink consume loop never reports status at all, so the shape of
the fix differs. Happy to fold it in if you would rather it went here.

The ordering itself has no test. `start_connector` needs a dlopened container, so
what holds it is that the transition sits in the same lock hold as the id record
with no await between them. The second test pins `set_error`'s observable
contract and its comment says plainly what it does not pin, since the message is
assigned after the transition and a mutant widening the clear survives.

The rebase moved every sha, so the ones in my earlier comment on this PR are
stale; I have refreshed them there.

hubcio (approved):

spetz (dismissed):

core/integration/tests/connectors/random/random_source.rs
```diff
@@ -97,6 +97,79 @@ async fn state_save_failure_preserves_state_and_source_recovers(harness: &TestHa
...
+async fn sources_running_does_not_climb_across_restarts(harness: &TestHarness) {
+    // The gauge counts running instances, and a restart takes one down before
+    // bringing one up, so one configured source stays at one however often it
+    // is restarted. It used to be reported by two mechanisms and taken back by
+    // one.
+    let api_url = harness
+        .connectors_runtime()
+        .expect("connectors runtime")
+        .http_url();
+    let http = Client::new();
```
hubcio: warning: requests outside the polling timeout can hang when the runtime stalls. set `.timeout(WAIT_TIMEOUT)` on the client and report the last completed poll without another request.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -463,6 +532,122 @@ mod tests {
...
+        assert_eq!(
+            details.handler_tasks.len(),
+            1,
+            "a stop drains the tasks recorded here, so they cannot be dropped"
+        );
+    }
+
+    #[tokio::test]
+    async fn should_not_double_count_when_an_error_falls_between_two_running_reports() {
+        // The interleaving spetz measured on #4064: the forwarding loop reports
```
hubcio: nit: this comment records PR history instead of the test invariant. remove the PR number and person's name.

core/connectors/runtime/src/main.rs
```diff
@@ -456,8 +456,28 @@ struct SinkConnectorWrapper {
...
 
+/// Closes a plugin instance whose setup did not finish, reporting a refusal
+/// rather than returning it: every caller is already on a failure path with an
+/// error of its own to surface.
+///
+/// `kind` is "source" or "sink". The two sides had this body inline, one word
+/// apart.
+pub(crate) fn close_plugin_instance(
+    close: &dyn Fn(u32) -> i32,
+    kind: &str,
```
hubcio: nit: `kind` accepts arbitrary strings even though `ConnectorType` already defines the labels. use that enum and make `as_label` visible within the crate.

core/connectors/runtime/src/main.rs
```diff
@@ -456,8 +456,28 @@ struct SinkConnectorWrapper {
     plugins: Vec<SinkConnectorPlugin>,
 }
 
+/// Closes a plugin instance whose setup did not finish, reporting a refusal
+/// rather than returning it: every caller is already on a failure path with an
+/// error of its own to surface.
+///
+/// `kind` is "source" or "sink". The two sides had this body inline, one word
+/// apart.
+pub(crate) fn close_plugin_instance(
```
hubcio, line 467: nit: this helper splits the connector struct declarations. move it above the grouped types.

core/integration/tests/connectors/random/random_source.rs
```diff
@@ -97,6 +97,79 @@ async fn state_save_failure_preserves_state_and_source_recovers(harness: &TestHa
...
+            204,
+            "restart {round} should be accepted"
+        );
+
+        wait_for_sources_running(&http, &api_url, 1).await;
+    }
+
+    // Read it once more after the gauge has settled. A report that lands after
+    // the poll above would otherwise go unseen.
+    sleep(STATE_STABILITY_WINDOW).await;
```
hubcio: nit: this gauge wait reuses a constant named for state storage. use `GAUGE_SETTLE_WINDOW`, or a neutral shared name if both waits should use the same policy.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -236,35 +231,60 @@ impl SourceManager {
             state,
         )?;
```
hubcio, line 232: warning: this pre-existing failed open leaves the fresh instance in the SDK registry before the guard exists. roll back failed opens while preserving duplicate-id rejection.

core/connectors/runtime/src/manager/source.rs
```diff
@@ -328,6 +348,55 @@ pub struct SourceDetails {
...
+
+    /// Records an instance that has just started, spawning its handlers in the
+    /// same breath.
+    ///
+    /// Deliberately not `async`, and that is the point. The id has to be
+    /// recorded under the same lock hold as the spawn: a cancellation between
+    /// the two strands the `SOURCE_SENDERS` entry and both tasks with nothing
+    /// naming them, which no guard can reach. Taking `spawn` as a closure is
+    /// what lets the compiler refuse an await added between them.
+    fn record_started(
```
hubcio: simplification: this callback adds indirection around startup assignments. inline it under the existing lock and remove its test, but keep registration, status and disarming free of `await` because the compiler will stop enforcing that.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +311,121 @@ pub(crate) fn init_source(
     }
 }
 
+/// A plugin's `iggy_source_close` together with whatever keeps the library that
+/// exports it mapped. Held instead of the bare `extern "C" fn` so the call stays
+/// valid once it is deferred off the calling thread.
+pub(crate) type SourceClose = Arc<dyn Fn(u32) -> i32 + Send + Sync>;
+
+/// Closes a source instance that `iggy_source_open` created and nothing else
```
hubcio, line 320: simplification: trim the repeated cleanup prose, but keep both startup and restart handoffs, library lifetime, worker offloading, and the difference between awaited close and deferred `Drop`.

also at `core/connectors/runtime/src/manager/source.rs:92`.

core/connectors/runtime/src/source.rs
```diff
@@ -305,6 +311,121 @@ pub(crate) fn init_source(
...
+/// A guard rather than a cleanup branch on each fallible call, because the
+/// window is defined by the two statements that open and record the instance,
+/// not by which call between them happens to be fallible today. Adding a `?`
+/// inside it stays correct. Both call sites use it.
+#[must_use = "dropping an armed guard closes the source instance"]
+pub(crate) struct SourceInstanceGuard {
+    close: SourceClose,
+    plugin_id: u32,
+    key: String,
+    armed: bool,
```
hubcio: simplification: the flag duplicates cleanup ownership and forces callback clones. use `Option<SourceClose>`, clearing it on disarm and taking it during teardown, while preserving awaited close and deferred `Drop`.

core/connectors/runtime/src/source.rs
```diff
@@ -961,6 +1093,193 @@ mod tests {
...
+
+        assert_eq!(
+            *closed.lock().expect("close recorder"),
+            vec![plugin_id],
+            "a guard still armed owns the instance and must close exactly it"
+        );
+    }
+
+    #[test]
+    fn given_disarmed_guard_when_dropped_should_leave_the_instance_open() {
```
hubcio: simplification: the successful helper test already covers this disarmed guard. remove this duplicate, keeping the helper success/error cases and the separate armed-drop, runtime-drop and awaited-close tests.

core/integration/tests/connectors/random/random_source.rs
```diff
@@ -97,6 +97,79 @@ async fn state_save_failure_preserves_state_and_source_recovers(harness: &TestHa
...
+    assert_eq!(
+        sources_running(&http, &api_url).await,
+        1,
+        "one running source must stay counted once, whatever it took to restart it"
+    );
+}
+
+async fn sources_running(http: &Client, api_url: &str) -> u32 {
```
hubcio: simplification: the stats helpers repeat the same request and decoding. share a helper returning `reqwest::Result<ConnectorRuntimeStats>` so callers that retry errors keep doing so, while immediate readers retain their existing checks.

hubcio (changes requested):
warning: `core/connectors/runtime/src/manager/sink.rs:207` - this pre-existing setup failure strands the opened sink. guard it, await cleanup on failure, and keep task spawning and id registration under the manager lock before disarming.

warning: `core/connectors/runtime/src/manager/source.rs:162` - this pre-existing close call blocks an async worker. use `spawn_blocking` with the library owner, keeping the restart lock and remaining cleanup alive if the caller is cancelled.

warning: `core/connectors/runtime/src/manager/source.rs:183` - this pre-existing race leaves the gauge elevated when an initial `Running` report follows `Stopping` and batch draining is aborted. apply the final `Stopped` through `apply_status`.

warning: `core/connectors/runtime/src/sink.rs:737` - this pre-existing call ignores declared sink failures and skips runtime error accounting. propagate nonzero results before success accounting, and update sink status and gauge together when enabling that error path.

core/connectors/runtime/src/manager/source.rs
mlevkov: Confirmed and fixed in `fix(connectors): do not register a source or sink whose open failed`.

It is a bit worse than the registry entry. `SourceContainer::open` assigns `self.source` before it looks at the result, so the stranded instance is fully constructed, not a half-built one. The macro then inserts it whatever the open returned, and the runtime has its error back before it has recorded the plugin id, so nothing outside the plugin can name it to close it.

The rollback is not inserting: the container drops on the failure path and releases the instance the way any other failed construction does. The duplicate id guard above is untouched, so reopening the same id still refuses rather than replacing a live instance.

**`iggy_sink_open` had the same shape and I fixed it too.** Same unconditional insert after the same kind of open. Say if you would rather that rode separately.

No test, and I would rather say so than imply otherwise: the FFI entry points are `cfg(not(test))` so a unit test cannot call them, and the map lives inside the plugin where the runtime cannot see it. The change is three lines, which is the argument for reading it rather than around it.

core/integration/tests/connectors/random/random_source.rs
mlevkov: Fixed in `test(connectors): stop the runtime stats waits from hanging on a stalled runtime`. Both halves, and the second one is nastier than "can hang".

The unbudgeted request is the one in the assertion message. It is only evaluated once the wait has already timed out, which is exactly the moment the runtime is stalled, and with no timeout on the client it never returns. So the test that should fail with "never reached 1" hangs instead and reports nothing at all. The loop now carries the value it saw and the message prints that.

The client is built with `WAIT_TIMEOUT` on it, so no request in the file can outlive the wait it belongs to.

core/connectors/runtime/src/manager/source.rs
mlevkov: Inlined in `refactor(connectors): give the instance guard one representation of ownership`. Registration, the status transition and the disarm are all still inside the one lock hold, in that order.

One thing worth putting in front of you rather than just doing. The closure was not incidental indirection: it arrived as `0c4b859f8 fix(connectors): make an await in the start window a compile error`, and taking the spawn as a closure into a non-async fn is what made the compiler refuse an await between the spawn and the id record. Your note says to keep that property knowing the compiler will stop enforcing it, so I have written the requirement where the three statements now sit. Flagging it in case the commit changes your mind; happy either way.

Its test went with it.

core/connectors/runtime/src/source.rs
mlevkov: Done in `refactor(connectors): give the instance guard one representation of ownership`.

`Option<SourceClose>` throughout: `disarm` clears it, the awaited close and `Drop` each `take()` it, and neither clones the callback any more. The flag is gone, so there is no second thing that has to agree with the first.

Awaited close still finishes before its caller returns, and `Drop` still hands the work to the blocking pool with the container captured so the library stays mapped.

Checked it rather than assumed: making `disarm` a no-op turns `given_fallible_step_when_it_succeeds_should_leave_the_instance_open` red, so the ownership release is still the thing under test.

core/connectors/runtime/src/source.rs
mlevkov: Consolidated in `refactor(connectors): type the cleanup label and say the argument once`.

The argument now lives on `SourceInstanceGuard` once and keeps everything you listed: the window and why it is a guard rather than a cleanup branch per fallible call, that startup and restart both hand off through it, that the library has to stay mapped for a deferred call, that `Drop` offloads to the blocking pool, and that awaited close and `Drop` are not interchangeable because only one gives the caller an ordering. `SourceClose`, `for_container`, `close`, the `Drop` body and the call sites point at it instead of restating it.

**One question.** I could not place the second location. `manager/source.rs:92` in the current head is inside `set_error`, which is the gauge transition comment rather than cleanup prose, and the branch has taken master three times since you looked, so I suspect the anchor drifted. I trimmed the two call-site comments in that file that do repeat the cleanup argument. If you meant somewhere else, point me at it and I will do that one too.

core/connectors/runtime/src/source.rs
mlevkov: Removed in `refactor(connectors): give the instance guard one representation of ownership`. You are right that it is covered: `given_fallible_step_when_it_succeeds_should_leave_the_instance_open` runs a guard through `disarm` and asserts nothing was closed, and I verified it by making `disarm` a no-op, which turns that test red.

The armed-drop, runtime-drop, awaited-close and both helper cases are all still there.

One thing to hand back, since it is the reason the deleted test was written that way. The surviving success case asserts an **empty** recorder with nothing in it proving the recorder can move, which is the vacuity the removed test guarded against by dropping an armed guard first. It is saved only by its error-case sibling using the same type in a different test. That holds today; it stops holding if that sibling ever goes.

core/integration/tests/connectors/random/random_source.rs
mlevkov: Done in `test(connectors): stop the runtime stats waits from hanging on a stalled runtime`.

`fetch_stats` is the one request and the one decode, and it hands back `reqwest::Result` exactly as you described, so `wait_for_source_error_after` keeps treating a failed read as "not yet" while `sources_running` and `source_errors` keep failing on it.

core/connectors/runtime/src/manager/source.rs
mlevkov: Fixed in `refactor(connectors): type the cleanup label and say the argument once`. It states the interleaving now, with no name and no PR number.

Fair catch, and it was our own rule: CLAUDE.md says comments explain why and never reference the current task or PR. I wrote that one.

core/connectors/runtime/src/main.rs
mlevkov: Done in `refactor(connectors): type the cleanup label and say the argument once`. `close_plugin_instance` takes `ConnectorType` and `as_label` is `pub(crate)`.

The old doc comment stated the constraint the signature did not enforce, which is the tell.

core/connectors/runtime/src/main.rs
mlevkov: Moved in `refactor(connectors): type the cleanup label and say the argument once`. It sits above the grouped connector types now instead of between two of them.

core/integration/tests/connectors/random/random_source.rs
mlevkov: Renamed in `test(connectors): stop the runtime stats waits from hanging on a stalled runtime`.

Took the shared option: both waits are waiting on the same thing, a report that may land just after the poll before it, so it is `SETTLE_WINDOW` and the doc comment says that is why they share it.

mlevkov:
All 11 addressed and each one answered on its own thread. Four commits: the failed-open rollback, the stats-wait fixes, the guard and start window, and the naming and prose pass.

Two things I did not decide on my own, both on their threads and repeated here so they are not missed:

**`manager/source.rs:92`** in the S2 comment does not land on cleanup prose in the current head; it is inside `set_error`, the gauge transition comment. The branch has taken master three times since you looked, so I think the anchor drifted. I consolidated at `source.rs:319` and trimmed the two call-site comments in `manager/source.rs` that do repeat the argument. Point me at the right line if I missed one.

**The `record_started` closure** arrived as `0c4b859f8 fix(connectors): make an await in the start window a compile error` — passing the spawn into a non-async fn is what made the compiler refuse an await between the spawn and the id record. Your note says to keep that property knowing the compiler stops enforcing it, so I inlined it and wrote the requirement where the statements now are. Flagging the commit in case it changes your mind.

One thing beyond what you asked: **`iggy_sink_open` had the same unconditional insert as the source**, so a failed sink open stranded an instance the same way. Fixed alongside. Happy to split it out if you would rather it rode separately.

213 runtime and 165 SDK unit tests pass, and all 26 `connectors::random` and `connectors::runtime` integration tests pass.

hubcio (approved):

spetz (approved):

Issue #3893, opened 2026-08-16
Tracking Issue: Align the Python SDK API with the Rust (native) SDK API


jiengup:
### Description

This issue **works as a tracking of the process of aligning Python SDK API with the Rust core SDK API**.

**Goal:** Bring the Python SDK in line with the Rust native SDK under `core/sdk`. The table below covers both low-level and high-level APIs, with current Python status, the gap, and an adaptation difficulty estimate.

### A. Fully aligned (no work required)

| # | API | Layer |
|---|---|---|
| 1 | `connect` | Lifecycle |
| 2 | `ping` | System |
| 3 | `describe_options` | System |
| 4 | `get_user` | User |
| 5 | `get_users` | User |
| 6 | `create_user` | User |
| 7 | `delete_user` | User |
| 8 | `update_permissions` | User |
| 9 | `change_password` | User |
| 10 | `logout_user` | User |
| 11 | `get_stream` | Stream |
| 12 | `get_topic` | Topic |
| 13 | `get_topics` | Topic |
| 14 | `update_topic` | Topic |
| 15 | `delete_topic` | Topic |
| 16 | `purge_topic` | Topic |
| 17 | `get_consumer_group` | ConsumerGroup |
| 18 | `get_consumer_groups` | ConsumerGroup |
| 19 | `delete_consumer_group` | ConsumerGroup |
| 20 | `join_consumer_group` | ConsumerGroup |
| 21 | `leave_consumer_group` | ConsumerGroup |
| 22 | `send_binary_request` | Raw |
| 23 | `consumer_group()` + `IggyConsumer` | High-level |

### B. Remaining gaps (sorted by Priority, then Layer)

| # | API | Layer | Python status | Gap | Priority | Issue | PR | Difficulty |
|---|---|---|---|---|---|---|---|---|
| 1 | `get_streams` | Stream | ✅ (done) | Needs `Stream` type | High | #3520 | #3701  | Medium |
| 2 | `send_messages` | Message | ✅ (done) | `partition_id(u32)` only; missing `balanced()` / `messages_key()` | High | #3896 | #3927 | Medium |
| 3 | `poll_messages` | Message | ✅ (done) | `partition_id` mandatory `u32`; `Consumer` fixed to default; no named consumer/group | High  | #3876 | #3877  | Medium |
| 4 | `store_consumer_offset` | Offset | ❌ | Needs `Consumer` type exposed | High | #3997 | TBD | Medium |
[...]

PR #4065, opened 2026-09-05
feat(python): expose get_me, get_client and get_clients

### LINKED ISSUE #4015: Python SDK: expose get_me, get_client and get_clients

yummyPancake2607:
The Rust SDK's SystemClient trait can report the currently connected client and enumerate all connected clients, but the Python SDK exposed none of it, leaving operators unable to see connection state that every other SDK already surfaces.

Adds ClientInfo/ClientInfoDetails/ConsumerGroupInfo bindings following the user.rs pattern, wires get_me/get_client/get_clients onto IggyClient, and re-exports ClientInfo/ConsumerGroupInfo from the SDK prelude since only ClientInfoDetails was public before.

Closes #4015

## Rationale

Every other Apache Iggy SDK exposes `get_me`/`get_client`/`get_clients` from `SystemClient`, so a Python operator has no way to see which client they're connected as or what else is connected to the server — parity gap tracked under #3893, this issue's specific ask.

## What changed?

`foreign/python/src/client_info.rs` gained real bindings for `ClientInfo`, `ClientInfoDetails`, and `ConsumerGroupInfo` — `inner`-wrapped structs with `From<Rust*>` conversions and `#[gen_stub_pymethods]` getters, mirroring `user.rs` rather than the bare-struct copy that was there before. `address`/`transport` return `&str` instead of cloning, and `consumer_groups` documents that each read allocates a fresh list.

`IggyClient` in `client.rs` gained `get_me`, `get_client`, `get_clients`, placed right after `ping` (get_me needs authentication only; get_client/get_clients need `read_servers` or `manage_servers` and raise `Unauthorized` without it — both stated explicitly in the docstrings since the wire error is generic).

`core/sdk/src/prelude.rs` now re-exports `ClientInfo` and `ConsumerGroupInfo` alongside the already-exported `ClientInfoDetails` — needed for `get_clients()`'s return type to be nameable from the Python crate without pulling in `iggy_common` directly (which would have re-introduced a version pin `bump-version.sh` doesn't touch).

`tests/test_client_info.py` covers: `get_me` against the fixture's TCP transport and logged-in user id, `get_me` reporting a joined consumer group (with a helper that forces stream/topic/group ids to be pairwise distinct, since a fresh stream would otherwise hand its first topic and group the same id and let a swapped field mapping pass silently — verified by manually swapping `stream_id`/`topic_id` and confirming the test fails), `get_clients` containing the caller's own entry, `get_client` matching `get_me` and returning `None` for an unknown id, and the `read_servers` permission gate (granted and denied cases).

## Local Execution

- Passed
- Pre-commit hooks not ran — HawkEye isn't installed locally, so I ran the equivalent steps directly instead of through `prek`: `cargo fmt --all --check`, `cargo clippy --all-targets -- -D warnings` (Python crate, plus `-p iggy` for the prelude change), `ruff check`/`ruff format --check`/`pyrefly check .`, and `cargo run --bin stub_gen` to confirm `apache_iggy.pyi` has no drift. Full `pytest tests/` (excluding the Docker-only `test_tls.py`) passed against a real `iggy-server` instance: 510 passed.

## AI Usage

1. Claude Code (Claude Opus).
2. Full implementation: the binding module, the three client methods, the prelude re-export, and the test file were AI-generated, following the existing `user.rs`/`get_user`/`get_users` patterns in this crate.
3. Verified by running the code, not just reading it: built and ran a real `iggy-server`, exercised every new method and the permission gate against it, and for the stream/topic/group id-collision risk specifically, mutated the field mapping (swapped `stream_id`/`topic_id`) and confirmed the test suite catches it before reverting.
4. Yes.


yummyPancake2607:
@hubcio i have created a new PR. I have implemented what you have told me in the previous PR. I have make sure everything works and written accordingly.

slbotbm (changes requested):
The following comments from #4020 were not resolved:
- https://github.com/apache/iggy/pull/4020#discussion_r3922437697
- https://github.com/apache/iggy/pull/4020#discussion_r3922437708
- https://github.com/apache/iggy/pull/4020#discussion_r3922437749
- https://github.com/apache/iggy/pull/4020#discussion_r3922437739

foreign/python/tests/test_client_info.py
```diff
@@ -0,0 +1,212 @@
...
+from .utils import login_fresh_client, unique_credentials
+
+
+def unused_client_id(clients: list[ClientInfo]) -> int:
+    """Return a client id that no currently connected client holds."""
+    return max((client.client_id for client in clients), default=0) + 1_000_000
+
+
+async def group_with_distinct_ids(client: IggyClient, unique_name):
+    """Create a stream, topic and consumer group with pairwise distinct ids.
```
slbotbm: Inline these. No need for functions that are only called once.

slbotbm (changes requested):

hubcio:
why did you close previous PR?

yummyPancake2607:
@hubcio I closed the previous PR because I was having some issues committing the necessary fixes to that branch. Since I wasn't able to properly update the code there, I closed it and created a new PR with the corrected changes.

yummyPancake2607:
@slbotbm 
## get_client: OverflowError vs RuntimeError for out-of-range client_id

You're right — `client_id: u32` is converted by pyo3 before the coroutine
even exists, so `get_client(-1)` raises `OverflowError` synchronously, not
`RuntimeError` from within the request.

Fixed in `foreign/python/src/client.rs`:
- The `Args:` section now says the conversion happens before the awaitable
  exists and names `OverflowError` as the result for an out-of-range value.
- Added `OverflowError` as its own line in `Raises:`, ahead of the
  `RuntimeError` line.

Added `tests/test_client_info.py::TestGetClient::test_get_client_out_of_range_id_raises_overflow_error`,
parametrized over `[-1, 2**32]` (`ids=["negative", "above-u32"]`), following
`test_client_config.py:130`'s `out_of_range` pattern rather than
`test_user.py`'s (that one asserts `TypeError` because `get_user` takes the
`str | int` union `PyIdentifier`, not a plain `u32`). Verified it fails
before the fix and passes after, against a live server.

## get_me: unimplemented over HTTP

Confirmed — `core/sdk/src/http/system.rs:46` is
`async fn get_me(&self) -> ... { Err(IggyError::FeatureUnavailable) }`
unconditionally, while `get_client`/`get_clients` do issue real requests
over HTTP (`GET /clients/{id}` and `GET /clients`).

Added a line to `get_me`'s docstring stating it's unimplemented over HTTP
and always raises, unlike `get_client`/`get_clients`, and updated `Raises:`
to name `Feature is unavailable` as the unconditional HTTP outcome.

## get_clients completeness assumption

Confirmed against `core/shard/src/lib.rs:1840` (`list_all_clients`): it's a
scatter-gather across shards bounded by `LIST_CLIENTS_GATHER_TIMEOUT` (3s,
`core/shard/src/lib.rs:312`), and a shard that misses the deadline is
dropped from the result — "callers should treat the result as
best-effort-complete" per that function's own doc comment.

Added a comment explaining why completeness holds anyway at each place the
tests rely on it:
- `test_get_clients_contains_this_client` — the membership check now uses
  `next((c for c in clients if ...), None)` instead of a list-length
  assertion, with the comment explaining the test server is unloaded and
  single-node so every shard replies well within the 3s budget.
- `test_user_with_read_servers_can_list_clients` — same assumption, shorter
  comment pointing back to the first one rather than repeating it.

foreign/python/tests/test_client_info.py
```diff
@@ -0,0 +1,235 @@
...
+        """Test get_client and get_clients are denied without read_servers."""
+        username, password = unique_credentials(unique_name)
+        created = await iggy_client.create_user(username, password)
+
+        client = await login_fresh_client(username, password)
+        # get_me needs authentication only, so it stays available.
+        me = await client.get_me()
+        assert me.user_id == created.id
+
+        with pytest.raises(RuntimeError):
```
hubcio: warning: a bare `pytest.raises(RuntimeError)` passes on any failure, since every error in this binding becomes `RuntimeError`. use `match="Unauthorized"` - the docstring promises that string and it survives the wire round trip.

also at line 232.

foreign/python/tests/test_client_info.py
```diff
@@ -0,0 +1,235 @@
...
+        assert me.consumer_groups_count == 1
+        assert len(me.consumer_groups) == 1
+
+        joined = me.consumer_groups[0]
+        assert joined.stream_id == stream.id
+        assert joined.topic_id == topic.id
+        assert joined.group_id == group.id
+
+        await member.leave_consumer_group(stream_name, topic_name, group_name)
+        await iggy_client.delete_consumer_group(stream_name, topic_name, group_name)
```
hubcio: warning: cleanup runs only on the success path, so a failed assert leaks the consumer group or the user. wrap in `try/finally`.

also at lines 215, 235.

foreign/python/src/client.rs
```diff
@@ -157,6 +158,93 @@ impl IggyClient {
...
+            Ok(client.map(PyClientInfoDetails::from))
+        })
+    }
+
+    /// Get the info about all the currently connected clients.
+    ///
+    /// Requires the `read_servers` or `manage_servers` global permission.
+    ///
+    /// Returns:
+    ///     An awaitable that resolves to `list[ClientInfo]`.
```
hubcio, line 237: nit: the best-effort caveat lives only in a test comment. python callers read this docstring, so say here that a shard missing the gather timeout is dropped from the list instead of failing the call.

foreign/python/tests/test_client_info.py
```diff
@@ -0,0 +1,235 @@
...
+        name = unique_name()
+        await client.create_topic(
+            stream=stream_name,
+            name=name,
+            partitions_count=1,
+        )
+        details = await client.get_topic(stream_name, name)
+        assert details is not None
+        topics.append((name, details))
+    topic_name, topic = next((n, t) for n, t in topics if t.id != stream.id)
```
hubcio: nit: bare `next()` cannot fire today, but if it ever does the `StopIteration` escaping an `async def` becomes `RuntimeError: coroutine raised StopIteration`, the same type the permission tests expect. give it a `None` default and assert, like line 146.

also at line 71.

foreign/python/tests/test_client_info.py
```diff
@@ -0,0 +1,235 @@
...
+    IggyClient,
+    Permissions,
+)
+
+from .utils import login_fresh_client, unique_credentials
+
+
+def unused_client_id(clients: list[ClientInfo]) -> int:
+    """Return a client id that no currently connected client holds."""
+    return max((client.client_id for client in clients), default=0) + 1_000_000
```
hubcio: nit: `1_000_000` is an unnamed literal. name it, same for the `3` in `range(3)` at lines 52 and 65.

foreign/python/src/client_info.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+#[gen_stub_pyclass]
+#[pyclass]
+pub struct ConsumerGroupInfo {
+    pub(crate) inner: RustConsumerGroupInfo,
+}
+
+impl From<&RustConsumerGroupInfo> for ConsumerGroupInfo {
```
hubcio, line 147: simplification: this hand-rolled copy exists only because `ConsumerGroupInfo` in `iggy_common` has no `Clone`. derive it there and this collapses to `inner: group.clone()`.

foreign/python/src/client_info.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn consumer_groups_count(&self) -> u32 {
+        self.inner.consumer_groups_count
+    }
+
+    /// The collection of consumer groups the client is part of.
+    ///
+    /// Each read rebuilds the list and every `ConsumerGroupInfo` in it, so
+    /// bind it once (`groups = details.consumer_groups`) rather than
+    /// subscripting the attribute repeatedly: two reads never share object
+    /// identity, and mutating the returned list does not affect the client.
```
hubcio: nit: the part a caller actually hits is that `ConsumerGroupInfo` has no `__eq__`, so `==` between groups from two reads is `False`. say that instead of the list-mutation sentence.

foreign/python/src/client_info.rs
```diff
@@ -0,0 +1,178 @@
...
+#[pymethods]
+impl ClientInfo {
+    /// The unique identifier of the client.
+    #[getter]
+    pub fn client_id(&self) -> u32 {
+        self.inner.client_id
+    }
+
+    /// The unique identifier of the user, or `None` while the client is
+    /// connected but not yet authenticated.
```
hubcio, line 47: nit: nothing tests the `None` branch, though a client that connects without `login_user` shows up in `get_clients` with `user_id` `None`. deliberate, or worth a test?

foreign/python/tests/test_client_info.py
```diff
@@ -0,0 +1,235 @@
...
+        # get_clients() is a best-effort scatter-gather across shards: one
+        # that misses the server's LIST_CLIENTS_GATHER_TIMEOUT (3s) budget is
+        # dropped from the result rather than failing the whole read. This
+        # test server is unloaded and single-node, so every shard replies
+        # well within budget and the caller's own entry is always present;
+        # that would not hold under load or across a busier cluster.
+        mine = next((c for c in clients if c.client_id == me.client_id), None)
+        assert mine is not None
+        assert mine.address == me.address
+        assert mine.transport == me.transport
```
hubcio: nit: equality with `me.transport` cannot catch a wire-code shift, both sides come from the same `transport_kind_to_wire`. assert `== "TCP"` here too, like line 89.

also at line 165.

foreign/python/tests/test_client_info.py
yummyPancake2607: both were only called from one place each.

- `unused_client_id(clients)` is now inlined directly into
  `test_get_client_unknown_id_returns_none` as a local
  `unused_id = max((c.client_id for c in clients), default=0) + 1_000_000`.
- `group_with_distinct_ids(client, unique_name)` is now inlined into
  `test_get_me_reports_joined_consumer_group`, using `iggy_client` directly
  instead of a generic `client` param; its docstring became a comment above
  the stream/topic/group setup explaining why three candidates are created
  to force distinct ids.

hubcio (changes requested):

hubcio:
i will approve CI once you fix everything on your side.

yummyPancake2607:
@hubcio 
## Bare pytest.raises(RuntimeError)

Fixed — both occurrences in `test_user_without_read_servers_cannot_list_clients`
now use `pytest.raises(RuntimeError, match="Unauthorized")`. Confirmed
`IggyError::Unauthorized`'s `Display` is exactly `"Unauthorized"`
(`core/common/src/error/iggy_error.rs:96`) and the binding forwards
`e.to_string()` verbatim, so it survives the wire round trip unchanged.

## Cleanup only on the success path

Fixed at all three spots:
- `test_get_me_reports_joined_consumer_group`: join, assertions and leave
  now run inside `try`, with `delete_consumer_group` in `finally`.
- `test_user_with_read_servers_can_list_clients` (line 215) and
  `test_user_without_read_servers_cannot_list_clients` (line 235): user
  creation is followed by `try`/`finally: await iggy_client.delete_user(...)`.

## get_clients best-effort caveat missing from the docstring

Added to both `get_client` and `get_clients` in `client.rs` (confirmed both
go through the same `shard.list_all_clients()` scatter-gather in
`core/server/src/dispatch/reads.rs`, bounded by
`LIST_CLIENTS_GATHER_TIMEOUT` — 3s — in `core/shard/src/lib.rs:312`):
`get_clients` states a shard missing the gather timeout is dropped from the
result; `get_client` states a connected client on such a shard can
transiently resolve to `None`. Regenerated the stub, both notes are in
`apache_iggy.pyi` now.

## Bare next() risking StopIteration

Fixed both (the topic pick and the group pick in
`test_get_me_reports_joined_consumer_group`): each now takes `None` as the
`next()` default and asserts not-None before unpacking, same shape as the
`mine = next(..., None); assert mine is not None` pattern already used for
`get_clients()` lookups.

## Unnamed literals (1_000_000, 3)

Named as module-level constants: `UNUSED_CLIENT_ID_MARGIN = 1_000_000` and
`DISTINCT_ID_ATTEMPTS = 3`, replacing all three call sites (`+ 1_000_000`
and both `range(3)` loops).

## Hand-rolled ConsumerGroupInfo copy

Derived `Clone, Copy` on `iggy_common::ConsumerGroupInfo`
(`core/common/src/types/client/client_info.rs`) — it's three plain `u32`
fields, nothing blocked it. `client_info.rs`'s `From<&RustConsumerGroupInfo>`
now collapses to `Self { inner: *group }`.

## consumer_groups docstring naming the wrong surprise

Rewritten to say what a caller actually hits: `ConsumerGroupInfo` has no
`__eq__`, so two instances from separate reads compare unequal even with
identical fields — bind the list once and compare fields, not objects.
Dropped the list-mutation sentence, which wasn't the real gotcha.

## user_id None branch untested

Not deliberate — added
`test_get_clients_reports_none_user_id_before_login`: connects a second
client without `login_user`, confirms it shows up in `get_clients()`
(diffed against the pre-connect snapshot to avoid depending on client-id
ordering) with `user_id is None`.

## transport equality can't catch a wire-code shift

Fixed at both call sites (`test_get_clients_contains_this_client` and
`test_get_client_by_id_matches_get_me`): both now assert `== "TCP"` instead
of comparing against `me.transport`, which comes from the same
`transport_kind_to_wire` mapping and so couldn't have caught a shift.

hubcio:
stop overusing AI.

hubcio:
image

yummyPancake2607:
@hubcio i will use AI only for help now can you open this PR?

hubcio:
no, please find yourself other project to contribute.

Issue #4056, opened 2026-09-04
csharp and Java SDKs: string identifier length uses length in UTF-16 code units


richardcocks:
### Bug description

The C# SDK and Java SDK build the length prefix for a string identifier from the string's `.Length` / `.length()`, which in each case count UTF-16 code units. They then write the value as UTF-8 encoded bytes in C# or platform default encoded in Java 17. For any non-ASCII name these diverge and the identifier is encoded wrong.

There is also a size guard in the C# SDK which checks the code unit count rather than the byte count.

The server expects a 1 to 255 **byte** UTF-8 name (`WireName`, `core/binary_protocol/src/primitives/identifier.rs`), so this is a client-side encoding bug; nothing on the server changes.

## C# `Iggy_SDK/Identifier.cs`

```csharp
Length = value.Length,                  // UTF-16 code units
Value  = Encoding.UTF8.GetBytes(value)  // UTF-8 bytes
```

Callers in `Contracts/Tcp/TcpContracts.cs` size their buffers from `Length`, so a non-ASCII id is either truncated (`GetBytesFromIdentifier`), throws `ArgumentException: Destination is too short` (tight single-field buffers like `GetUser`), or overruns into the next field (multi-field buffers like `UpdateStream`). In the last case the server reads a shorter, different identifier.

Reflecting the internal serializers off the shipping DLL:

```
"café"       len 4, 5 bytes   -> truncated to invalid UTF-8, server rejects
"naïve-café" len 10, 12 bytes -> UpdateStream sends streamId "naïve-caf"
"日本語"      len 3, 9 bytes    -> UpdateStream sends streamId "日"
```

## Java `BytesSerializer.java`

In Java, the buffer grows, so it doesn't throw, instead it writes all the bytes but declares the code-unit count, so every non-ASCII id produces a desynced frame that the server misreads.

```java
ByteBuf buffer = Unpooled.buffer(2 + identifier.getName().length()); // UTF-16 code units
buffer.writeByte(identifier.getName().length());                     // length prefix
buffer.writeBytes(identifier.getName().getBytes());                  // UTF-8 bytes ( or platform-default on Java 17 ) 
```
Produces:
```
café       len 4, 5 bytes  -> 02 04 63 61 66 C3 A9      (declares 4, 5 follow; A9 spills)
naïve-café len 10, 12 bytes -> server reads id "naïve-caf", C3 A9 spill into next field
[...]

PR #4068, opened 2026-09-05
fix(java): size wire strings by UTF-8 byte length, not char count

mmodzelewski:
Names, keys and passwords with non-ASCII characters went out with a
length prefix counted in UTF-16 chars while the bytes came from the
platform default charset, so the prefix and payload disagreed and the
server misread the frame. The 255 cap was also checked in chars, so a
short name could still overflow the u8 prefix, and uncapped strings
wrapped it silently.

Every wire string is now encoded as UTF-8 once and prefixed with its
byte length. The serializer rejects empty or over-long strings and
names the field. Identifier caches the encoded name, and Partitioning
checks the value shape the server expects for each kind. The SDK
version is cut at a code point boundary. Message.of, the example
consumer and the javadoc all use UTF-8 explicitly.

The login payload no longer carries version and context strings; the
VSR codec drops them and sends the SDK version itself. The codec now
validates fields before allocating, closing a pooled-buffer leak on
an empty username. Usernames, passwords and token names are checked
against the server's own bounds before the round trip.

Message keys are hashed as UTF-8 bytes, so on a JVM whose default
charset is not UTF-8 a non-ASCII key may land on a different
partition than before. Needs a 0.9.0 release note. Non-ASCII names
are tested end to end over TCP only; the HTTP client does not
percent-encode paths yet.

Refs #4056


foreign/java/java-sdk/src/main/java/org/apache/iggy/identifier/Identifier.java
```diff
@@ -68,13 +80,21 @@ public String getName() {
         return name;
     }
 
+    /**
+     * The name as UTF-8 wire bytes, encoded once at construction; {@code null} for a numeric
+     * identifier. The array is shared rather than copied, so callers must not modify it.
+     */
+    @Nullable public byte[] getEncodedName() {
```
spetz, line 87: Consider returning a defensive copy or a read-only view. This exposes the cached serialization bytes: `id.getEncodedName()[0] = 'b'` changes `StreamId.of("foo")` into `"boo"` on the wire while `getName()`, logging, and routing-cache keys still identify `"foo"`. The Javadoc documents the restriction, but enforcing it would preserve identifier consistency

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/async/tcp/vsr/VsrLoginCodecTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @ParameterizedTest
+    @NullSource
+    @EmptySource
+    void sdkVersionFieldFallsBackToUnknownWhenVersionIsMissing(String version) {
+        assertThat(VsrLoginCodec.sdkVersionField(version)).isEqualTo("unknown".getBytes(StandardCharsets.UTF_8));
+    }
+
+    @Test
+    void sdkVersionFieldTruncatesOnCodePointBoundaryWithinU8Prefix() {
```
spetz, line 47: Please add surrogate-pair boundary cases. The existing `é` test covers multibyte UTF-8 but not supplementary code points. `"a".repeat(251) + "\uD83D\uDE00"` should encode to 255 bytes. With 252 leading ASCII characters, the result should contain only those 252 bytes. This verifies that truncation preserves complete code points

spetz (changes requested):

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/UsersTcpClient.java
```diff
@@ -155,20 +157,23 @@ public CompletableFuture<IdentityInfo> login(String username, String password) {
         return routingHook.loginOnLeader(() -> loginWithoutRedirect(username, password));
     }
 
+    static ByteBuf loginPayload(String username, String password, String version, String context) {
+        byte[] versionBytes = version.getBytes(StandardCharsets.UTF_8);
+        byte[] contextBytes = context.getBytes(StandardCharsets.UTF_8);
+        var payload = Unpooled.buffer();
+        payload.writeBytes(toBytes(username));
+        payload.writeBytes(toBytes(password));
+        payload.writeIntLE(versionBytes.length);
```
hubcio: warning: these bytes never reach the wire - `VsrLoginCodec.rewriteUserLogin` reads username and password only, then hardcodes `writeIntLE(0)` for the context. drop version and context here, or thread `context` through the codec.

foreign/java/java-sdk/src/main/java/org/apache/iggy/message/Partitioning.java
```diff
@@ -23,8 +23,20 @@
...
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 public record Partitioning(PartitioningKind kind, byte[] value) {
+
+    /** Server-side cap on a messages key, in encoded bytes, matching its u8 length prefix. */
+    public static final int MAX_MESSAGES_KEY_LENGTH = 255;
+
+    public Partitioning {
+        if (value.length > MAX_MESSAGES_KEY_LENGTH) {
```
hubcio: warning: this only checks the ceiling, and it checks it for every kind. an empty `MessagesKey` and a `PartitionId` that isn't 4 bytes both pass here and get rejected by the server - switch on `kind` instead.

foreign/java/java-sdk/src/main/java/org/apache/iggy/message/Partitioning.java
```diff
@@ -38,10 +50,15 @@ public static Partitioning partitionId(Long id) {
     }
 
     public static Partitioning messagesKey(String key) {
-        if (key == null || key.isBlank() || key.length() > 255) {
-            throw new IggyInvalidArgumentException("Key must be non-empty and less than 255 characters long");
+        if (key == null || key.isBlank()) {
+            throw new IggyInvalidArgumentException("Key must be non-empty");
+        }
+        byte[] encoded = key.getBytes(StandardCharsets.UTF_8);
```
hubcio: warning: this changes message-key routing - `MessagesTcpClient` hashes these bytes client-side, so a non-ASCII key lands on a different partition after the upgrade on any JVM whose default charset isn't UTF-8. correct change, but it needs a release note.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
```diff
@@ -63,6 +63,9 @@ public final class BytesSerializer {
      */
     private static final int MAX_HEADER_FIELD_LENGTH = 255;
 
+    /** Bound on a u8-length-prefixed wire string, in encoded bytes. */
+    private static final int MAX_U8_STRING_LENGTH = 255;
```
hubcio, line 65: warning: 255 is the wire cap, but the server bounds username at 3..50, password at 3..100 and PAT name at 3..30. a 200-byte username passes here and then dies at the server - bound them at the call sites.

foreign/java/java-sdk/src/test/java/org/apache/iggy/message/MessageTest.java
```diff
@@ -72,6 +73,14 @@ void ofCreatesMessageWhenGivenMessageHeaderPayloadAndNullUserHeaders() {
         assertThat(message.userHeaders().size()).isEqualTo(0);
     }
 
+    @Test
+    void ofEncodesStringPayloadAsUtf8() {
```
hubcio, line 77: warning: this passes against pre-PR code on any UTF-8-default JVM, and the Test task sets no `defaultCharacterEncoding`, so CI never runs the case it's named for. same at `PartitioningTest.java:81`. fork the test JVM with `file.encoding=ISO-8859-1`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
```diff
@@ -210,9 +214,12 @@ public static ByteBuf toBytes(TopicPermissions permissions) {
     }
 
     public static ByteBuf toBytes(String value) {
-        int bufferLength = 1 + value.length();
-        ByteBuf buffer = Unpooled.buffer(bufferLength);
         byte[] stringBytes = value.getBytes(StandardCharsets.UTF_8);
+        if (stringBytes.length > MAX_U8_STRING_LENGTH) {
```
hubcio: nit: seven fields funnel through here - stream, topic and group names, PAT name, username, password, PAT token - so this message tells the caller nothing about which one broke. `checkFieldLength` below takes a field label for exactly that reason.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
```diff
@@ -210,9 +214,12 @@ public static ByteBuf toBytes(TopicPermissions permissions) {
...
+            throw new IggyInvalidArgumentException("String must be at most " + MAX_U8_STRING_LENGTH
+                    + " bytes when UTF-8 encoded, got " + stringBytes.length);
+        }
+        ByteBuf buffer = Unpooled.buffer(1 + stringBytes.length);
         buffer.writeByte(stringBytes.length);
```
hubcio, line 215: nit: no lower bound - `toBytes("")` still writes `[0x00]`, and the server rejects a zero-length wire name, so `createStream("")` builds a frame it always refuses. pre-existing, and adding a floor means flipping `shouldSerializeEmptyString`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/identifier/Identifier.java
hubcio: nit: new public API on a public abstract class that hands back the live array. a caller that mutates it keeps the length prefix valid and silently sends a different name than `getName()` reports.

foreign/java/java-sdk/src/main/java/org/apache/iggy/message/Partitioning.java
```diff
@@ -23,8 +23,20 @@
...
 
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 public record Partitioning(PartitioningKind kind, byte[] value) {
+
+    /** Server-side cap on a messages key, in encoded bytes, matching its u8 length prefix. */
+    public static final int MAX_MESSAGES_KEY_LENGTH = 255;
+
+    public Partitioning {
```
hubcio, line 35: nit: `new Partitioning(kind, null)` throws a bare NPE at the length check instead of `IggyInvalidArgumentException`. same shape at `partitionId(null)`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrLoginCodec.java
```diff
@@ -94,15 +103,22 @@ static long readSessionEpoch(ByteBuf registerBody) {
...
+     * always fits its u8 prefix and still decodes as UTF-8 on the server.
+     */
+    static byte[] sdkVersionField(String version) {
+        String value = version == null || version.isEmpty() ? UNKNOWN_SDK_VERSION : version;
+        ByteBuffer encoded = ByteBuffer.allocate(MAX_SHORT_FIELD_LENGTH);
+        StandardCharsets.UTF_8
+                .newEncoder()
+                .onMalformedInput(CodingErrorAction.REPLACE)
+                .onUnmappableCharacter(CodingErrorAction.REPLACE)
+                .encode(CharBuffer.wrap(value), encoded, true);
```
hubcio: nit: the encoder contract wants `flush(out)` after `encode(in, out, true)`. it works today only because the UTF-8 encoder's `implFlush` is a no-op - append `.flush(encoded)`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrLoginCodec.java
```diff
@@ -94,15 +103,22 @@ static long readSessionEpoch(ByteBuf registerBody) {
     private static void writeVersionInfo(ByteBuf body) {
         body.writeIntLE(PROTOCOL_VERSION);
         writeShortField(body, SDK_NAME.getBytes(StandardCharsets.UTF_8));
```
hubcio: nit: `SDK_NAME` and the version are both constants, re-encoded on every login and reconnect. hoist them to `private static final byte[]`, keeping `sdkVersionField` package-visible for the test.

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/blocking/tcp/BytesSerializerTest.java
```diff
@@ -239,6 +267,44 @@ void shouldSerializeStringIdentifier() {
...
+            assertThat(result.readableBytes()).isEqualTo(0);
+        }
+
+        @ParameterizedTest
+        @CsvSource({
+            "café,       02 05 63 61 66 C3 A9",
+            "naïve-café, 02 0C 6E 61 C3 AF 76 65 2D 63 61 66 C3 A9",
+            "日本語,     02 09 E6 97 A5 E6 9C AC E8 AA 9E",
+        })
+        void shouldMatchServerWireFormatForNonAsciiNames(String name, String expectedHex) {
```
hubcio: nit: the name says server wire format, but the body compares hand-written hex with nothing tying it to the Rust encoder. rename to `shouldMatchExpectedWireLayoutForNonAsciiNames`, or move the parity claim into `bdd/`.

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/blocking/tcp/BytesSerializerTest.java
```diff
@@ -239,6 +267,44 @@ void shouldSerializeStringIdentifier() {
             result.readBytes(nameBytes);
             assertThat(new String(nameBytes)).isEqualTo("test-stream");
```
hubcio: nit: `new String(nameBytes)` decodes on the platform default, the same conflation this PR is removing. pass `StandardCharsets.UTF_8`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/message/Partitioning.java
```diff
@@ -38,10 +50,15 @@ public static Partitioning partitionId(Long id) {
     }
 
     public static Partitioning messagesKey(String key) {
-        if (key == null || key.isBlank() || key.length() > 255) {
-            throw new IggyInvalidArgumentException("Key must be non-empty and less than 255 characters long");
+        if (key == null || key.isBlank()) {
+            throw new IggyInvalidArgumentException("Key must be non-empty");
+        }
+        byte[] encoded = key.getBytes(StandardCharsets.UTF_8);
+        if (encoded.length > MAX_MESSAGES_KEY_LENGTH) {
```
hubcio: simplification: the compact constructor applies this same bound four lines down. drop the check here and let the constructor throw - `hasMessageContaining("300")` still passes.

foreign/java/java-sdk/src/main/java/org/apache/iggy/identifier/Identifier.java
```diff
@@ -23,10 +23,15 @@
 import org.apache.iggy.exception.IggyInvalidArgumentException;
 
 import javax.annotation.Nullable;
+import java.nio.charset.StandardCharsets;
 
 public abstract class Identifier {
 
+    /** Server-side cap on a wire name, in UTF-8 bytes, matching its u8 length prefix. */
+    public static final int MAX_NAME_LENGTH = 255;
```
hubcio: simplification: nothing outside this file uses `MAX_NAME_LENGTH`, and the same holds for `Partitioning.MAX_MESSAGES_KEY_LENGTH`. both are new, so making them private breaks nothing.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/ConsumerGroupsTcpClient.java
```diff
@@ -106,11 +106,13 @@ public CompletableFuture<ConsumerGroupDetails> createConsumerGroup(
             StreamId streamId, TopicId topicId, String name) {
         var streamIdBytes = BytesSerializer.toBytes(streamId);
         var topicIdBytes = BytesSerializer.toBytes(topicId);
-        var payload = Unpooled.buffer(1 + streamIdBytes.readableBytes() + topicIdBytes.readableBytes() + name.length());
+        var nameBytes = BytesSerializer.toBytes(name);
+        var payload = Unpooled.buffer(
```
hubcio: simplification: `toBytes(streamId)` followed by two `writeBytes` calls does the same thing without the manual sizing, matching `:61` and `:84` in this file. same shape at `StreamsTcpClient.java:104`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/StreamsTcpClient.java
```diff
@@ -102,12 +99,12 @@ public CompletableFuture<List<StreamBase>> getStreams() {
...
-        var payload = Unpooled.buffer(payloadSize + idBytes.capacity());
+        var nameBytes = BytesSerializer.toBytes(name);
+        var payload = Unpooled.buffer(idBytes.readableBytes() + nameBytes.readableBytes());
 
         payload.writeBytes(idBytes);
-        payload.writeBytes(BytesSerializer.toBytes(name));
+        payload.writeBytes(nameBytes);
         // Trailing options block. Streams have no catalog keys yet, so the
         // server rejects every key; the empty block is the extension point.
         payload.writeBytes(BytesSerializer.toBytes(Map.<HeaderKey, HeaderValue>of()));
```
hubcio: simplification: `toBytes` of an empty map returns `EMPTY_BUFFER`, so this writes nothing. the comment above already marks the extension point - same at `UsersTcpClient.java:125`.

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/vsr/VsrLoginCodec.java
```diff
@@ -94,15 +103,22 @@ static long readSessionEpoch(ByteBuf registerBody) {
...
-        String version = IggyVersion.getInstance().getVersion();
-        if (version == null || version.isEmpty()) {
-            return "unknown";
-        }
-        return version.length() > 255 ? version.substring(0, 255) : version;
+    /**
+     * An over-long version is cut on the encoded bytes at a code point boundary, so the field
+     * always fits its u8 prefix and still decodes as UTF-8 on the server.
+     */
+    static byte[] sdkVersionField(String version) {
```
hubcio, line 115: simplification: `getBytes(UTF_8)` already replaces unpaired surrogates, so this reduces to encode once, return if it's 255 bytes or under, else walk back off the continuation bytes. drops the flush issue too.

hubcio (changes requested):
Four things that aren't on a changed line:

- **The login third fixes a path with no bug.** `VsrLoginCodec.rewriteUserLogin` reads username and password only and hardcodes `writeIntLE(0)` for the context at `:76`, so those bytes never left the process pre-PR either. `UsersTcpClientPayloadTest` now pins a layout no server sees.
- **Pooled buffer leak at `VsrLoginCodec.java:71`.** `writeShortField` throws on an empty username at `:138`, and `VsrRequestEncoder.java:58` calls the codec outside its try/finally. Reachable via `login("", pw)` - move the length check above the allocation.
- **Breaks a round trip in the examples.** `Message.of` now encodes UTF-8, but its partner `AsyncConsumer.java:315` still decodes on the platform default. Same line in the `MessagesClient.java:62` javadoc.
- **No non-ASCII test reaches a real server.** `MessagesTcpClientTest.java:34` already covers the `metadataLength` sum end to end, but with ASCII names only.

For the description: `toBytes(String)` now rejects over-long usernames, passwords and PAT fields too, and the message-key routing change needs a line in the 0.9.0 notes.

foreign/java/java-sdk/src/main/java/org/apache/iggy/identifier/Identifier.java
mmodzelewski: raw bytes array is no longer shared

foreign/java/java-sdk/src/test/java/org/apache/iggy/client/async/tcp/vsr/VsrLoginCodecTest.java
mmodzelewski: added new test cases

foreign/java/java-sdk/src/main/java/org/apache/iggy/message/Partitioning.java
mmodzelewski: added info to the PR description

foreign/java/java-sdk/src/test/java/org/apache/iggy/message/MessageTest.java
mmodzelewski: I think it's not worth it spawning a separate test suite just for it. From JDK 18, UTF-8 is the default. Before 18, ISO-8859-1 would only be set on Windows.

hubcio (approved):

spetz (approved):

PR #4072, opened 2026-09-05
fix(csharp): update string length validation to use UTF-8 byte count

lukaszzborek:
Identifier, Partitioning and header types checked string.Length,
so a non-ASCII name under 255 characters could exceed 255 bytes
and be truncated into the one byte length prefix, sending a
frame the server parses as a shorter name followed by garbage.
Most TcpContracts skipped the check entirely. Identifier also
compared and hashed Value by reference.

Every length-prefixed name, credential, token and header field
now goes through one WireName rule mirroring the server's 1 to
255 byte bound. Identifier and Partitioning derive Length from
Value and reject oversize arrays in the initializer. Identifier
equality uses byte contents. User contracts validate credentials
against CredentialBounds before serializing.

Contract tests cover empty, at-limit, over-limit and non-ASCII
inputs for each entry point.

#4056

### DIFF
--- a/foreign/csharp/Iggy_SDK/Identifier.cs
+++ b/foreign/csharp/Iggy_SDK/Identifier.cs
@@ -33,24 +33,34 @@ public readonly struct Identifier : IEquatable<Identifier>
     public required IdKind Kind { get; init; }
 
     /// <summary>
-    ///     Identifier length in bytes.
+    ///     Identifier length in bytes, always derived from <see cref="Value" />.
+    ///     The initializer is kept for compatibility and its value is ignored.
     /// </summary>
-    public int Length => Value.Length;
+    public int Length
+    {
+        get => _value.Length;
+        init { }
+    }
 
     /// <summary>
-    ///     Identifier value as bytes, at most 255 of them.
+    ///     Copy of the identifier value as bytes, at most 255 of them.
     /// </summary>
     /// <exception cref="ArgumentOutOfRangeException">Thrown when the value is longer than 255 bytes.</exception>
     public required byte[] Value
     {
-        get => _value;
+        get => _value.ToArray();
         init
         {
             ArgumentNullException.ThrowIfNull(value);
             ArgumentOutOfRangeException.ThrowIfGreaterThan(value.Length, WireName.MAX_LENGTH, nameof(Value));
-            _value = value;
+            _value = value.ToArray();
         }
     }
+
+    /// <summary>
+    ///     Read-only view of the value bytes for serialization, without the defensive copy of <see cref="Value" />.
+    /// </summary>
+    internal ReadOnlySpan<byte> Bytes => _value;
 
     private readonly byte[] _value;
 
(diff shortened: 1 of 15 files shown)


foreign/csharp/Iggy_SDK/Kinds/Partitioning.cs
```diff
@@ -33,12 +34,24 @@ public readonly struct Partitioning
     /// <summary>
     ///     Length of the partitioning value.
     /// </summary>
-    public required int Length { get; init; }
+    public int Length => Value.Length;
```
spetz: Removing `init` breaks existing object initializers that supplied the previously required `Length`. Recompiling produces `CS0200`, while previously compiled consumers throw `MissingMethodException` for `set_Length`. Both are reproduced. Keep the getter derived from `Value`, but preserve a compatible initializer so this serialization fix does not require downstream source changes. It also relates to `Identifier`.

foreign/csharp/Iggy_SDK/Identifier.cs
```diff
@@ -151,6 +160,9 @@ public override bool Equals(object? obj)
     /// <inheritdoc />
     public override int GetHashCode()
     {
-        return HashCode.Combine((int)Kind, Value);
+        var hash = new HashCode();
+        hash.Add(Kind);
+        hash.AddBytes(Value);
```
spetz: Content-based hashing now depends on a publicly mutable array. After inserting an identifier into a dictionary, changing `id.Value[0]` makes lookup of that same identifier fail. Consider protecting the backing bytes from mutation through both the initializer input and the getter.

foreign/csharp/Iggy_SDK_Tests/UtilityTests/IdentifiersByteSerializationTests.cs
```diff
@@ -30,6 +31,27 @@ public void StringIdentifier_WithInvalidLength_ShouldThrowArgumentException()
         Assert.Throws<ArgumentException>(() => Identifier.String(val));
     }
 
+    [Theory]
+    [InlineData("café", 5)]
+    [InlineData("naïve-café", 12)]
+    [InlineData("日本語", 9)]
+    public void StringIdentifier_WithNonAscii_ShouldUseUtf8ByteLength(string value, int expectedLength)
```
spetz, line 40: These assertions check identifier properties, while the new contract matrix uses numeric identifiers. Add exact-byte assertions for `GetUser(Identifier.String("café"))` and `UpdateStream(Identifier.String("café"), "topic")`. These cover the original tight-buffer failure and corruption of the following field.

spetz (changes requested):

foreign/csharp/Iggy_SDK/Kinds/Partitioning.cs
lukaszzborek: init was reverted with empty body, so value passed here will be ignored in both places

foreign/csharp/Iggy_SDK/Identifier.cs
lukaszzborek: added defensive copy on creating and return data. Introduced internal field to get data without additional allocation

spetz (approved):

hubcio (approved):

PR #4076, opened 2026-09-06
perf(node, csharp, java, go): generate message ids from random bytes instead of via UUIDs 

### LINKED ISSUE #4066: all clients: Performance of messageId generation - Do they need to be crypto-secure?
I've not tagged this a bug, because current behaviour is as-designed, but the current state of play is that client SDKs have a feature to generate messageId values rather than passing 0 and leaving it up to the server.

The message Id is 16 opaque random bytes. It doesn't appear to be used for idempotency, it doesn't get keyed off, and it doesn't appear to need to be secret.

As far as I can tell, it does not therefore need to be crypto-secure, and indeed clients are free to pass sequential IDs if they wish to do so.

The default behaviour however is the current generation schemes:

| client | trigger | generator | RNG | form → 16 bytes | overhead |
|---|---|---|---|---|---|
| **Rust** | `id == 0` | `Uuid::new_v4().to_u128_le()` (`core/common/src/utils/random_id.rs:22`) | `getrandom` (OS) | native `u128` | ~none (no alloc, no string) |
| **C#** | `Header.Id == 0` (`Iggy_SDK/Contracts/Tcp/TcpContracts.cs:350`) | `Guid.NewGuid().ToUInt128()` | OS RNG | `UInt128` reinterpret (struct) | ~none (value type, no alloc) |
| **Go** | `Header.Id == zero` (`foreign/go/internal/command/message.go:98`) | `uuid.NewRandom()` (google/uuid) | `crypto/rand` (OS) | fills a `[16]byte` directly | low (one small value) |
| **Java** | id `signum == 0` (`serde/BytesSerializer.java:342`) | `UUID.randomUUID()` | shared `SecureRandom` (contended) | `UUID` obj → `byte[16]` copy | higher: crypto RNG + object + `byte[]` alloc |
| **Node** | id absent/zero (`wire/message/message.utils.ts:127`) | `uuidv4()` (uuidv7 pkg) | `getRandomValues` (OS), inside | string → `BigInt('0x…')` → `u128ToBuf` | worst: ~2 strings + a BigInt per msg |

Node is by far the worst on my ( admittedly very artificial ) benchmark, my throughput was ~1.8M/sec on go/rust/csharp, ~1M/sec on Java and just 0.183M/sec on Node, with a large part of the different attributable to message Id generation, with a 2.3x throughput improvement by switching to sequential IDs.

If it's confirmed that message Ids do not need to be crypto-secure, then there are much better options for message IDs. Twitter developed Snowflake ID for multi-node collision resistant monotonic ID generation, and I'd recommend switching to that.

Another alternative to consider would be UUIDv7, which generates monotonic but still unique IDs, but snowflake has the advantage of removing the need for random generation at all which puts it on a par with sequential IDs for performance.

Sticking to entirely random IDs, performance improvements could be made across the board by switching from CSPRNGs to regular random sources. In the case of Node, the difference seems especially dramatic, with the default xoroshift128 being close to a sequential counter in performance.

richardcocks:
Closes #4066

## Rationale

Profiling the small-message hot path showed a large amount of time spent in generating message IDs. This was most noticeable in the Node SDK where it was impacting throughput severely.

## What changed?

The SDKs were minting Ids by constructing UUIDv4 values and then transforming them to extract the 16 bytes the message IDs needed.  The message IDs are opaque bytes which do not need to match a UUID format, nor are they cryptographically sensitive.

This updates the SDKs to generate bytes directly from faster random sources where available. 

For Node, to achieve a fast throughput required a new random pool, which was fastest to fill from the CSPRNG, because the non-CSPRNG doesn't have any way to get a byte array from it. Other SDKs have PRNGs which are fast without the complexity of managing a pool.

The Node path now throws on a numeric id ≥ 2¹²⁸ (via a new bound check in serializeMessageId), where the old hex round-trip silently truncated to a garbage value. This is in line with other SDKs. 

### Producer throughput — small (~6 B) messages

`id=0` mint path, measured with my harness (`MODE=producer
MSG_ID=uuid REPEAT=10 BITS=42`, TCP_NODELAY on, server + producer co-located).
Mean messages/sec over 10 runs (stddev %).

| SDK     | master (before)  | branch (after)   |            change |
| :------ | ---------------: | ---------------: | ----------------: |
| Node    |   196,809 (1.6%) |   528,076 (2.4%) | **+168% (2.68×)** |
| Java    | 1,027,080 (6.5%) | 1,375,820 (9.1%) |        **+34.0%** |
| C#\*    | 1,870,353 (5.5%) | 2,201,070 (9.8%) |        **+17.7%** |
| Go      | 1,630,751 (8.2%) | 1,915,466 (9.4%) |        **+17.5%** |
| Rust\*\*| 1,785,892 (6.7%) | 1,824,243 (8.3%) |      +2.1% (within noise)  |

\* C# Compiled with .NET 11 AOT, which emits warnings against the SDK's HTTP transport, which is unused. Both master and branch are compiled in the same way.

\*\* Rust is a control: its mint path is unchanged, so the tie (within noise)
confirms the harness attributes the other deltas to the code, not the machine.

### Producer throughput — ~1 KB messages

Same test at a realistic payload (`pack=167` → 1,002-byte messages). Mean
messages/sec over 10 runs (stddev %); the mint runs once per message, so its
benefit shrinks as payloads grow.

| SDK  | master (before) | branch (after)  |          change |
| :--- | --------------: | --------------: | --------------: |
| Node |   75,565 (3.4%) |  107,641 (2.2%) |    **+42.4%**   |
| Java |  172,647 (5.1%) |  203,600 (4.1%) |    **+17.9%**   |
| Go   |  368,275 (5.1%) |  384,620 (6.5%) |  +4.4% (noise)  |
| C#   |  394,541 (4.5%) |  407,785 (4.0%) |  +3.4% (noise)  |
| Rust\*  |  323,347 (7.4%) |  324,836 (10.1%)|   +0.5% (tie)   |

The C# and Go results collapse to technically within noise, but given the single-message improvements is likely to be a real effect that I can't prove statistically without ~50 repeated runs for each branch.

The Java and Node results are well above the noise floor.

\* Rust is a control, unchanged between master and branch.

Please do not try to compare languages, absolute throughput is not comparable across SDKs, each producer is different and are not comparable, the benchmarks are tuned for comparing differences across builds of the same SDK, not between languages or SDKs.

## AI Usage

Claude Opus 4.8 was used extensively both for writing code, analysis of ideas, in benchmarking and in verification of results.
Every line has been hand-reviewed and I fully understand the changes, the choices made and the rationale for those choices.

Verification was mostly through end-to-end benchmarking tests which also verify correctness. Additional correctness tests and micro-benchmarks were scaffolded during development to check assumptions, some of which were out of scope of this PR and have been excluded.

### DIFF
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
@@ -33,7 +33,6 @@ import org.apache.iggy.message.MessageHeader;
 import org.apache.iggy.message.MessageId;
 import org.apache.iggy.message.Partitioning;
 import org.apache.iggy.message.PollingStrategy;
-import org.apache.iggy.message.UuidMessageId;
 import org.apache.iggy.user.GlobalPermissions;
 import org.apache.iggy.user.Permissions;
 import org.apache.iggy.user.StreamPermissions;
@@ -45,7 +44,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.UUID;
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * Unified serializer for both blocking and async clients.
@@ -340,7 +339,12 @@ public final class BytesSerializer {
      */
     private static byte[] encodedMessageId(MessageId id) {
         if (id.toBigInteger().signum() == 0) {
-            return readAllBytes(new UuidMessageId(UUID.randomUUID()).toBytes());
+            // A zero id is filled with random bytes directly (fast, non-crypto):
+            // the id is opaque, not keyed on, and 128 bits keeps collisions far
+            // below the birthday bound at any realistic message rate.
+            byte[] minted = new byte[16];
+            ThreadLocalRandom.current().nextBytes(minted);
+            return minted;
         }
         return readAllBytes(id.toBytes());
     }
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
@@ -339,9 +339,6 @@ public final class BytesSerializer {
      */
     private static byte[] encodedMessageId(MessageId id) {
         if (id.toBigInteger().signum() == 0) {
-            // A zero id is filled with random bytes directly (fast, non-crypto):
-            // the id is opaque, not keyed on, and 128 bits keeps collisions far
-            // below the birthday bound at any realistic message rate.
             byte[] minted = new byte[16];
             ThreadLocalRandom.current().nextBytes(minted);
             return minted;
--- a/foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
+++ b/foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
@@ -44,7 +44,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * Unified serializer for both blocking and async clients.
@@ -339,9 +338,7 @@ public final class BytesSerializer {
      */
     private static byte[] encodedMessageId(MessageId id) {
         if (id.toBigInteger().signum() == 0) {
-            byte[] minted = new byte[16];
-            ThreadLocalRandom.current().nextBytes(minted);
-            return minted;
+            return MessageIdGenerator.mint();
         }
         return readAllBytes(id.toBytes());
     }
--- a/foreign/csharp/Iggy_SDK/Contracts/Tcp/TcpContracts.cs
+++ b/foreign/csharp/Iggy_SDK/Contracts/Tcp/TcpContracts.cs
@@ -344,12 +344,17 @@ internal static class TcpContracts
         position += 4;
 
         // The producer owns message ids: a zero id is minted before the frame checksum covers it.
+        // A zero id is filled with random bytes directly (fast, non-crypto): the id is opaque,
+        // not keyed on, and 128 bits keeps collisions far below the birthday bound. The scratch
+        // buffer is hoisted out of the loop so the stackalloc runs once (CA2014).
         var originTimestamp = ulong.MaxValue;
+        Span<byte> idBytes = stackalloc byte[16];
         foreach (var message in messages)
         {
             if (message.Header.Id == 0)
             {
-                message.Header = message.Header with { Id = Guid.NewGuid().ToUInt128() };
+                Random.Shared.NextBytes(idBytes);
+                message.Header = message.Header with { Id = BinaryPrimitives.ReadUInt128LittleEndian(idBytes) };
             }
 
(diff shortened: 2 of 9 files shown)


T1B0:
Thanks for this PR @richardcocks ! the nodejs part looks good to me :+1:

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
```diff
@@ -340,7 +339,9 @@ private static long batchChecksum(
      */
     private static byte[] encodedMessageId(MessageId id) {
         if (id.toBigInteger().signum() == 0) {
-            return readAllBytes(new UuidMessageId(UUID.randomUUID()).toBytes());
+            byte[] minted = new byte[16];
+            ThreadLocalRandom.current().nextBytes(minted);
```
spetz: `ThreadLocalRandom` has only 64 bits of random state per thread. Producing 16 bytes does not preserve UUIDv4’s collision resistance across JVMs. Please use a generator with at least 128 bits of independently seeded state, or pooled `SecureRandom` bytes

spetz (changes requested):

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
richardcocks: Thank you for spotting this, I'll go back to a CSPRNG in Java too.

Unlike in node/v8, I seem to be hitting a performance quirk where asking for more bytes out of the Java SecureRandom class seems closer to a linear cost, so pooling isn't helping. I'll investigate this a little more because it doesn't feel right to me.

What I can do instead is use AES-CTR to generate a collision resistant stream, as that will be both fast and sufficiently random. This is the primitive that `SecureRandom` itself uses, but it also does extra re-seeding, which isn't necessary for collision resistance. 

I've checked C# and Go and they both have higher random state in their PRNGs ( 256 bits ), so have collision resistance,

However, with Java now also going back to CSPRNG source, I would understand a desire to keep the SDKs more consistent, so I can source their message Ids from their CSPRNGs too if desired. ( This has the benefit of being better for anyone downstream who might accidentally treat message IDs as being securely generated. )

spetz: @richardcocks sounds good. I think the other SDKs are good to go, so we probably don't need to change these, and regarding the Java one, if we could have it done today, that'd be great, as we'd like to make it a part of the voting release.

hubcio (approved):

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
richardcocks: @spetz  - I have pushed a new MessageIdGenerator class for Java with test coverage.  I've also done end-to-end tests to confirm the performance is good.

I can explain what it is doing and why it works, although some of my Java knowledge is slightly confused by 20 years of C#, so I sometimes make bad assumptions about the details in Java.

I have not had time to de-claude the language in the comments for the class, but I have set "allow maintainer fixes" for this branch.

spetz: Thanks, should be good enough!

spetz (approved):

PR #4077, opened 2026-09-06
ci(connectors): run a plugin's integration suite when it changes

ethanlin01x:

Relates to #3996

## Rationale

Integration tests for a connector plugin do not run when that plugin changes.

Cause: `cargo rail plan` scopes tests via the cargo dependency graph. A plugin is a `cdylib` the runtime loads with `dlopen`, so cargo sees no edge between the plugin and the `integration` crate, and `integration::connectors::<name>` falls outside scope. This affects all 19 plugins. Doris is the only exception, because `core/integration` happens to have a path dependency on it.

## What changed?

The pre-merge action now maps each affected plugin to its integration suite: the nextest filter gains `| (package(integration) & test(/^connectors::<name>::/))`, and `integration` joins the build list. 

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

1. Which tools? Claude Code
2. Scope of usage? Implementation and local verification
3. How did you verify the generated code works correctly? Ran the shell under the same `bash -eo pipefail` flags GitHub Actions uses, and checked the generated expressions with `cargo nextest list -E`


ethanlin01x:
The CI should be fixed after #4074 is merged.

spetz (approved):
LGTM

hubcio (approved):

Issue #4021, opened 2026-09-01
Add an OpenMessaging Benchmark driver for Iggy


hubcio:
## Why

[OpenMessaging Benchmark](https://github.com/openmessaging/benchmark) (OMB) is the suite behind Confluent's 2020 Kafka vs Pulsar vs RabbitMQ comparison and Redpanda's 2022 Redpanda vs Kafka comparison (each ran its own fork, confluentinc/openmessaging-benchmark and redpanda-data/openmessaging-benchmark). It ships 15 drivers, 130 workloads under `workloads/`, and a result JSON that `bin/create_charts.py` turns into the charts people compare. Iggy has no driver, so there's no like-for-like number for it. #217 (2023) asked for a Kafka-style single-node benchmark and linked the OMB docs. It was closed as completed in 2025 with iggy-bench, without an OMB driver.

`iggy-bench` stays the in-house tool (8 benchmark kinds, one microsecond latency sample per batch, `report.json` for the dashboard). OMB is coarser on the consumer side (end-to-end latency is a wall-clock millisecond difference, 0 ms samples are dropped) and reports in 10-second windows, but it's the baseline other systems already publish against.

## When

After the Iggy 0.9.0 release. That release puts Java SDK 0.9.0 on Maven Central, the first release that speaks the VSR wire protocol (#3841). The current 0.8.0 (on Maven Central since 2026-04-22) uses the old TCP framing, which the server removed in #3856, so it can't even log in. OMB's build resolves third-party artifacts from Maven Central only and its CI runs without a custom `settings.xml`, so the driver isn't buildable upstream before then.

## What a driver has to provide

`driver-api` (`io.openmessaging.benchmark:driver-api:0.0.1-SNAPSHOT`, compiled with `--release 17`) is four interfaces plus a `ResourceCreator` helper nobody needs. Compiling against it needs only that jar and its transitive `bookkeeper-stats-api`.

`BenchmarkDriver` has `initialize(File yaml, StatsLogger)`, `getTopicNamePrefix()`, `createTopic(name, partitions)`, `createProducer(topic)`, `createConsumer(topic, subscriptionName, callback)` and `close()`. The worker instantiates it by reflection from `driverClass` in the driver yaml, through a no-arg constructor.

[...]

PR #4096, opened 2026-09-08
feat(java): let TCP clients run one I/O thread or share a group

mmodzelewski:
Every AsyncIggyTcpClient built its own event loop group with
Netty's default of 2 x CPU threads. Its pool holds one channel,
and a channel lives on one loop, so the other threads sat idle.
Nothing let clients share a group. An application that opened one
client per producer and consumer paid a thread count that grew
with the client count.

The group that a client creates for itself now has one thread by
default, sized through ioThreads on the builder. eventLoopGroup
registers the channel on a caller-owned group instead, and the
client never shuts that group down. The builder rejects a group
that cannot drive NIO channels or that already started to shut
down, and ignores ioThreads when a group is supplied. The
blocking builder forwards both options.

The connection takes one loop from the group at construction and
pins both the channel and the heartbeat to it. A next() call per
heartbeat tick rotates the timer across loops the channel never
uses. A second next() call, such as a pool handed the whole
group, takes a slot of the group's round-robin counter and lands
later connections off the sequence.

Shutting the owned group down swept every channel. A shared
group stays up, so close() now closes the tracked channels
itself before it closes the pool. The pool closes only idle
channels, and a login holds its lease until the reply, so the
pool alone left that channel open. If a caller shuts the group
down under a live connection, the heartbeat stops with a
warning and close() completes instead of failing on a
terminated loop.

Refs #4021


foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncTcpConnection.java
```diff
@@ -920,18 +958,26 @@ public CompletableFuture<Void> close() {
...
-                .closeAsync()
-                .addListener(f -> eventLoopGroup.shutdownGracefully().addListener(sf -> {
-                    if (sf.isSuccess()) {
-                        shutdownFuture.complete(null);
-                    } else {
-                        shutdownFuture.completeExceptionally(sf.cause());
-                    }
-                }));
+        channelPool.closeAsync().addListener(poolClosed -> {
+            if (!ownsEventLoopGroup) {
```
spetz: Closing the pool only closes idle channels. A pending login or logout still holds its channel, so this branch completes `close()` while that socket and request remain alive. I reproduced this by withholding a login reply: `close().get()` returned, but the peer socket stayed open and the login future remained pending. Previously, shutting down the owned group closed that channel too. Please explicitly close this connection’s channels, including acquired ones, before completing shutdown

foreign/java/java-sdk/src/main/java/org/apache/iggy/client/async/tcp/AsyncIggyTcpClientBuilder.java
```diff
@@ -308,6 +359,28 @@ private void validateMaxVsrFrameSize() {
...
+        if (ioThreads < 1) {
+            throw new IggyInvalidArgumentException("IoThreads must be at least 1");
+        }
+    }
+
+    private void validateEventLoopGroup() {
+        if (eventLoopGroup == null) {
+            return;
+        }
+        if (!eventLoopGroup.isCompatible(NioIoHandle.class)) {
```
spetz: Netty’s [`isCompatible()`](https://github.com/netty/netty/blob/netty-4.2.17.Final/transport/src/main/java/io/netty/channel/IoEventLoopGroup.java) calls `next()`, so validation advances the shared chooser. Pool construction, acquisition and channel registration also advance it. I reproduced six sequential `build()` / `connect().join()` calls on a two-thread group: all six channels landed on the same loop. Pinning the heartbeat does not prevent this. Please validate without advancing the chooser, then select one loop per connection and use it for both the bootstrap and pool

spetz (changes requested):

spetz (approved):

numinnex (approved):

Issue #4084, opened 2026-09-07
Retry backoff has an off-by-one delay and an unclamped post-jitter overshoot


mattp5657:
### Description

`exponential_backoff`/`jitter` in `core/connectors/sdk/src/retry.rs` have two
bugs, both reproduced by every call site that composes them by hand:

1. **Off-by-one doubles the first retry's delay.** `exponential_backoff(base, attempt, max_delay)` expects a 0-based `attempt` (`0` → `base`). Most call sites increment their counter *before* calling it, so the first retry passes `1` and waits `base * 2` instead of `base`, and every subsequent retry is skewed one step ahead too.
2. **Jitter is applied after the cap, so it can overshoot `max_delay`.** `exponential_backoff` clamps to `max_delay` internally, but `jitter(...)` (±20%) wraps that already-capped value with no re-clamp: a retry can sleep up to `max_delay * 1.2`, silently exceeding the configured ceiling.

### Affected area / component

Connectors, Rust SDK

| Location | Off-by-one | Missing clamp |
|---|---|---|
| SDK `retry.rs`: `HttpRetryMiddleware::handle`, `check_connectivity_with_retry` | Yes | Yes |
| `influxdb_sink` / `influxdb_source` (inherit the SDK bug directly) | Yes | Yes |
| `meilisearch_sink::check_connectivity` (3 call sites) | Yes | Yes |
| `core/connectors/sources/iggy_source` (passes `consecutive_failures` directly) | Yes | Yes |
| `s3_sink` upload retry loop | Fixed (`attempt - 1`) | Still overshoots |
| `surrealdb_sink` write retry loop | Fixed (`saturating_sub(1)`) | Still overshoots |
| `doris_sink` Stream Load retry | Fixed | Fixed |
| `opensearch_sink::sleep_before_retry` | Fixed (`retries - 1`) | Fixed |
| SDK `source.rs::nack_retry_delay` | Fixed (`saturating_sub(1)`) | N/A (never calls `jitter`) |
| `clickhouse_sink`'s own local `jittered_backoff` helper (4 call sites) | Yes (separate reimplementation) | N/A (self-bounding by construction) |

### Proposed solution

Fix the interface, not every call site: change `exponential_backoff` to
accept a 1-based total-attempt count internally (subtract 1 before computing
the exponent), so the call every caller already makes naturally (an
incremented counter) becomes correct.

[...]

Issue #3702, opened 2026-07-19
feat(connectors): shared retry_async helper in iggy_connector_sdk::retry


ryankert01:
### Description

The connectors codebase has three hand-written copies of the same retry skeleton: `HttpRetryMiddleware::handle` and `check_connectivity_with_retry` in `core/connectors/sdk/src/retry.rs`, and `DorisSink::load_batch` in `core/connectors/sinks/doris_sink/src/lib.rs`.

The Doris sink cannot reuse the middleware: Doris Stream Load signals failure as HTTP 200 with `{"Status":"Fail"}` in the body, and the middleware never inspects response bodies. Any connector whose backend reports errors in-band has the same problem, so each one ends up hand-rolling the loop.

The copies have already diverged on backoff convention: the SDK's two loops pass the already-incremented attempt to `exponential_backoff`, so their first retry waits `2 * base`, while the Doris sink passes `attempt - 1` so the first retry waits the configured base delay, matching the documented meaning of `retry_delay`.

Raised as a non-blocking follow-up in the review of #3574: https://github.com/apache/iggy/pull/3574#discussion_r3507968371

### Affected area / component

Connectors

### Proposed solution

Expose a generic `retry_async(operation, policy)` in `iggy_connector_sdk::retry` that owns attempt counting, the transient-error predicate, capped exponential backoff with jitter, and per-retry logging, then collapse the three call sites onto it.

The shared helper should keep the base-delay-first convention (`base * 2^(k-1)` for retry `k`), since that is what the `retry_delay` config field documents.

### Alternatives considered

- Reuse `HttpRetryMiddleware` in the Doris sink: not possible, the middleware classifies on status codes and never inspects response bodies, so an in-band `{"Status":"Fail"}` looks like success to it.
- Keep per-connector loops: the exponent divergence above shows the conventions drift as soon as the skeleton is copied.

### Contribution

- [x] I'm willing to submit a pull request to implement this feature

### Good first issue

- [ ] I think this could be a good first issue for a new contributor

ryankert01:
There will be 3 PRs. 1 main, 2 follow-ups.

PR #4104, opened 2026-09-09
feat(connectors)!: add shared retry_async helper to the connector SDK

ryankert01:

Relates to #3702
Related to #4084

## Rationale

`exponential_backoff(base, attempt, max)` takes a 0-based exponent, but retry loops count from one. Eight of twelve call sites passed the un-decremented counter and waited twice the configured `retry_delay` before the first retry.

## What changed?

Seven connectors hand-rolled the same retry loop, because `HttpRetryMiddleware` classifies on HTTP status alone and a backend reporting failure in-band (Doris: HTTP 200 with `{"Status":"Fail"}`) cannot use it. The copies then drifted on the convention above.

`retry_async` now owns the loop for failures that surface as `Err`, and `retry_backoff` is the one place that knows a 1-based retry means `base × 2^(k−1)`. Doris and the startup probe move onto it; Meilisearch, S3, SurrealDB, RabbitMQ and the middleware keep their loops but take their timing from it.

**Breaking:** `ConnectivityConfig` removed — field-for-field `RetryPolicy` (`max_open_retries`→`max_attempts`, `retry_delay`→`base_delay`, `open_retry_max_delay`→`max_delay`). Both are `(u32, Duration, Duration)` with the delay roles crossed, so a field-by-field rename compiles and swaps base for cap. No FFI or `repr(C)` change, so built plugins are unaffected.

**Operator-visible:** first retry now waits `retry_delay`, not twice it, for Meilisearch (startup 17s→12.5s, per-op 7s→3.5s), the InfluxDB probe, and the middleware. Attempt counts unchanged.

Follow-ups, kept out to stay reviewable: 
1. jitter is zeroed below 5ms and clamping after jitter puts 260 of 512 saturated draws on the cap exactly
2. `Retry-After` is unbounded and read only on 429. The `Retry-After` ceiling is an operational policy call worth agreeing first.

## Local Execution

- Passed — `fmt`, `clippy -D warnings`, `test` across 8 crates, `cargo doc` under `RUSTDOCFLAGS=-D warnings`.
- `prek` ran and passes, except `license-headers`, `skills` and `taplo`, which need bash ≥ 4.2; ran those directly instead.

## AI Usage

- Claude Code.
- Implementation, tests and docs.
- Local chain above; tests confirmed failing against the old behaviour first; several adversarial review rounds.
- Yes.

### DIFF
--- a/core/connectors/sdk/src/retry.rs
+++ b/core/connectors/sdk/src/retry.rs
@@ -25,10 +25,10 @@
 //! - [`check_connectivity`] — single health-check probe (GET /health)
 //! - [`check_connectivity_with_retry`] — startup probe with exponential backoff
 //! - [`retry_async`] — generic retry loop for fallible async operations
+//! - [`RetryFailure`] — terminal outcome of [`retry_async`], with attempt count
 //! - [`RetryPolicy`] — attempt budget and backoff bounds for [`retry_async`]
 //! - [`is_transient_status`] — transient HTTP status predicate
 //! - [`parse_duration`] — humantime duration parsing with fallback
-//! - [`jitter`] — ±20 % random jitter for retry delays
 //! - [`exponential_backoff`] — capped exponential backoff
 //! - [`retry_backoff`] — jittered backoff for a 1-based retry number
 //! - [`parse_retry_after`] — HTTP `Retry-After` header parsing
@@ -154,7 +154,7 @@ pub fn parse_duration(value: Option<&str>, default_value: &str) -> Duration {
 }
 
 /// Apply ±20 % random jitter to `base` to spread retry storms.
-pub fn jitter(base: Duration) -> Duration {
+pub(crate) fn jitter(base: Duration) -> Duration {
     let millis = base.as_millis() as u64;
     let jitter_range = millis / 5; // 20% of base
     if jitter_range == 0 {
@@ -219,7 +219,7 @@ impl RetryPolicy {
 /// retry waits `base_delay`, the second `2 × base_delay`, and so on, which is
 /// the convention the `retry_delay` config fields document.
 ///
-/// The cap is re-applied after [`jitter`], because ±20 % jitter on an
+/// The cap is re-applied after jittering, because ±20 % jitter on an
 /// already-capped delay can otherwise land above `max_delay`, which the config
 /// fields document as a strict upper bound.
 ///
@@ -233,7 +233,44 @@ pub fn retry_backoff(base_delay: Duration, retry: u32, max_delay: Duration) -> D
     ))
     .min(max_delay)
 }
-/// Run `operation`, retrying while it fails with an error `is_transient`
+
+/// Why [`retry_async`] stopped.
+///
+/// Carries the attempt count so a caller can write its own terminal log: the
+/// helper owns the per-retry line, giving up is the caller's to report.
+#[derive(Debug)]
+pub struct RetryFailure<E> {
+    pub error: E,
+    /// Attempts actually made, including the first.
+    pub attempts: u32,
+    /// `true` when the attempt budget ran out, `false` when `should_retry`
+    /// rejected the error and no further attempt was made.
+    pub exhausted: bool,
+}
+
+impl<E> RetryFailure<E> {
+    /// Discard the attempt bookkeeping and keep the underlying error.
+    pub fn into_error(self) -> E {
+        self.error
+    }
+}
+
+impl<E: fmt::Display> fmt::Display for RetryFailure<E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let reason = if self.exhausted {
+            "ran out of attempts"
+        } else {
+            "hit a non-retryable error"
+        };
+        write!(
+            f,
+            "{reason} after {} attempts: {}",
+            self.attempts, self.error
+        )
+    }
+}
+
+/// Run `operation`, retrying while it fails with an error `should_retry`
 /// accepts and the attempt budget in `policy` is not exhausted.
 ///
 /// This is the retry skeleton for connectors whose failures surface as `Err`,
@@ -245,15 +282,16 @@ pub fn retry_backoff(base_delay: Duration, retry: u32, max_delay: Duration) -> D
 /// emits, e.g. `"Doris sink ID 3 Stream Load (label=abc)"`. Callers build it
 /// before the first attempt, so keep it off per-message paths.
 ///
-/// Retries, recovery and giving up are all logged here, with the attempt
-/// counts callers would otherwise each have to track. The last error is
-/// returned unchanged so callers keep their own error classification.
+/// The per-retry line is logged here. Giving up is not: the returned
+/// [`RetryFailure`] carries the attempt count and which condition ended the
+/// loop, so a caller logs the terminal failure at the level and wording that
+/// suit it. The error itself is returned unchanged.
 pub async fn retry_async<T, E, Op, Fut>(
     policy: RetryPolicy,
     context: &str,
-    is_transient: impl Fn(&E) -> bool,
+    should_retry: impl Fn(&E) -> bool,
     mut operation: Op,
-) -> Result<T, E>
+) -> Result<T, RetryFailure<E>>
 where
     Op: FnMut() -> Fut,
     Fut: Future<Output = Result<T, E>>,
@@ -274,19 +312,14 @@ where
         };
 
         attempt += 1;
-        let retryable = is_transient(&error);
-        let budget_exhausted = attempt >= max_attempts;
-        if budget_exhausted || !retryable {
-            let reason = if retryable {
-                "ran out of attempts"
-            } else {
-                "hit a non-retryable error"
-            };
-            // Both arms are terminal for the operation, so both log at error!.
-            // A caller that aborts on the `Err` (an `open()` probe) drops it at
-            // the FFI boundary, leaving this as the only record of the cause.
-            error!("{context} {reason} after {attempt} attempts: {error}");
-            return Err(error);
+        let retryable = should_retry(&error);
+        let exhausted = attempt >= max_attempts;
+        if exhausted || !retryable {
+            return Err(RetryFailure {
+                error,
+                attempts: attempt,
+                exhausted,
+            });
         }
 
         let delay = policy.backoff(attempt);
@@ -519,6 +552,12 @@ pub async fn check_connectivity_with_retry(
         || check_connectivity(client, url.clone(), connector_label),
     )
     .await
+    .map_err(|failure| {
+        // Sole record of the cause: `open()`'s Err is dropped at the FFI
+        // boundary and the runtime logs only "Plugin initialization failed".
+        error!("{context} {failure}");
+        failure.into_error()
+    })
 }
 
 #[cfg(test)]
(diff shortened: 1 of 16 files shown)


core/connectors/sdk/src/retry.rs
```diff
@@ -256,34 +376,28 @@ impl Middleware for HttpRetryMiddleware {
...
                         // Consume the error body for logging, then retry.
                         let body_text = response.text().await.unwrap_or_default();
-                        let delay = retry_after.unwrap_or_else(|| {
-                            jitter(exponential_backoff(
-                                self.retry_delay,
-                                attempts,
-                                self.max_delay,
-                            ))
-                        });
+                        let delay = retry_after.unwrap_or_else(|| self.policy.backoff(attempts));
```
jiengup, line 439: This logic seems to lack test cases. Let's add `Retry-After` takes precedence over local backoff tests.

core/connectors/sdk/src/retry.rs
```diff
@@ -181,6 +190,114 @@ pub fn parse_retry_after(value: &str) -> Option<Duration> {
...
+/// which classifies on the status code alone.
+///
+/// `context` identifies the connector and operation in every log line this
+/// emits, e.g. `"Doris sink ID 3 Stream Load (label=abc)"`. Callers build it
+/// before the first attempt, so keep it off per-message paths.
+///
+/// Retries, recovery and giving up are all logged here, with the attempt
+/// counts callers would otherwise each have to track. The last error is
+/// returned unchanged so callers keep their own error classification.
+pub async fn retry_async<T, E, Op, Fut>(
```
jiengup: As a common util, I think this interface should be designed to be simpler and easier to use. Consider renaming the `is_transient` parameter and removing the `context` parameter, just letting users to print logs in `Op` themselves.

jiengup (commented):
An almost complete commit, leaving some minor comments. 
In addition, this PR is also related to #4084.

core/connectors/sdk/src/retry.rs
```diff
@@ -181,6 +190,114 @@ pub fn parse_retry_after(value: &str) -> Option<Duration> {
...
+/// Prefer [`retry_async`], which calls this for you. Reach for it directly
+/// only in a loop that cannot be expressed as a retried `Result`.
+pub fn retry_backoff(base_delay: Duration, retry: u32, max_delay: Duration) -> Duration {
+    jitter(exponential_backoff(
+        base_delay,
+        retry.saturating_sub(1),
+        max_delay,
+    ))
+    .min(max_delay)
+}
```
mattp5657, line 233: Nit: Once backoff saturates (base_delay * 2^k >= max_delay), exponential_backoff returns exactly max_delay every time, and this trailing .min(max_delay) clamps away the upper half of jitter's +/-20% draw.

Maybe we should document it, but not a huge issue.

core/connectors/sdk/src/retry.rs
```diff
@@ -181,6 +190,114 @@ pub fn parse_retry_after(value: &str) -> Option<Duration> {
...
+/// the convention the `retry_delay` config fields document.
+///
+/// The cap is re-applied after [`jitter`], because ±20 % jitter on an
+/// already-capped delay can otherwise land above `max_delay`, which the config
+/// fields document as a strict upper bound.
+///
+/// Prefer [`retry_async`], which calls this for you. Reach for it directly
+/// only in a loop that cannot be expressed as a retried `Result`.
+pub fn retry_backoff(base_delay: Duration, retry: u32, max_delay: Duration) -> Duration {
+    jitter(exponential_backoff(
```
mattp5657, line 227: Maybe we should make `jitter()` private, it has no callers anywhere except retry_backoff . Unlike exponential_backoff, kept public deliberately for out-of-tree plugins, nothing relies on jitter being external. pub(crate) closes off the misuse class structurally instead of by convention, at zero cost.

mattp5657 (commented):
Overall LGTM, minor issues.

core/connectors/sdk/src/retry.rs
ryankert01: it will be in one of my followups

core/connectors/sdk/src/retry.rs
ryankert01: Good idea on the rename — `should_retry` is clearer, will do.

I'd keep `context` though. Plugin logs cross FFI via `CallbackLayer`, which drops structured fields (checked: `warn!(attempt = 3, "retrying")` arrives as just `"retrying"`) and sets target to the module path — so retries land as `iggy_connector_sdk::retry` with no instance identity. Doris puts `label=` there; two InfluxDB sinks in one runtime would otherwise be indistinguishable.

Agree the helper shouldn't own the terminal log though — I'll return `RetryFailure { error, attempts, exhausted }` so callers can write their own.

core/connectors/sdk/src/retry.rs
ryankert01: Done,

core/connectors/sdk/src/retry.rs
ryankert01: done.

core/connectors/sdk/src/retry.rs
```diff
@@ -161,16 +165,21 @@ pub fn jitter(base: Duration) -> Duration {
...
     let millis = base
         .as_millis()
         .saturating_mul(factor as u128)
         .min(max_delay.as_millis());
-    let millis_u64 = u64::try_from(millis).unwrap_or(u64::MAX);
-    Duration::from_millis(millis_u64)
+    Duration::from_millis(u64::try_from(millis).unwrap_or(u64::MAX))
 }
 
+/// Ceiling for a server-supplied `Retry-After`.
```
spetz: doc line "Ceiling for a server-supplied `Retry-After`." refers to a constant that does not exist in this PR

core/connectors/sdk/src/retry.rs
```diff
@@ -398,49 +529,296 @@ pub async fn check_connectivity(
...
+        // base + 2 × base, each independently jittered. Passing the retry
+        // number straight to `exponential_backoff` would give 2 + 4 × base.
+        let nominal = BASE * 3;
+        assert!(
+            elapsed >= nominal.mul_f64(JITTER_LOW) && elapsed <= nominal.mul_f64(JITTER_HIGH),
+            "two retries waited {elapsed:?}, expected roughly {nominal:?}"
+        );
+    }
+
+    /// The regression test for the convention this helper exists to unify: the
```
spetz: Comment block on `retry_client` describes tests not present in this module (Retry-After bound, statuses the header is read on).

core/connectors/sdk/src/retry.rs
```diff
@@ -181,6 +190,147 @@ pub fn parse_retry_after(value: &str) -> Option<Duration> {
...
+}
+
+impl<E> RetryFailure<E> {
+    /// Discard the attempt bookkeeping and keep the underlying error.
+    pub fn into_error(self) -> E {
+        self.error
+    }
+}
+
+impl<E: fmt::Display> fmt::Display for RetryFailure<E> {
```
spetz, line 260: `RetryFailure<E>` implements `Display` but not `std::error::Error`, so it cannot be propagated with `?` into `anyhow::Error` or `Box<dyn Error>`.

core/connectors/sdk/src/retry.rs
```diff
@@ -181,6 +190,147 @@ pub fn parse_retry_after(value: &str) -> Option<Duration> {
...
+                    info!("{context} succeeded after {attempt} retries.");
+                }
+                return Ok(value);
+            }
+            Err(error) => error,
+        };
+
+        attempt += 1;
+        let retryable = should_retry(&error);
+        let exhausted = attempt >= max_attempts;
```
spetz: `exhausted` is computed without `retryable`, so a permanent error on the last attempt reports `exhausted: true` and `Display` says "ran out of attempts".

core/connectors/sdk/src/retry.rs
```diff
@@ -181,6 +190,147 @@ pub fn parse_retry_after(value: &str) -> Option<Duration> {
...
+/// before the first attempt, so keep it off per-message paths.
+///
+/// The per-retry line is logged here. Giving up is not: the returned
+/// [`RetryFailure`] carries the attempt count and which condition ended the
+/// loop, so a caller logs the terminal failure at the level and wording that
+/// suit it. The error itself is returned unchanged.
+pub async fn retry_async<T, E, Op, Fut>(
+    policy: RetryPolicy,
+    context: &str,
+    should_retry: impl Fn(&E) -> bool,
```
spetz: `impl Fn(&E) -> bool` in argument position mixed with explicit generics `<T, E, Op, Fut>` makes `retry_async` impossible to turbofish (E0632)

core/connectors/sdk/src/retry.rs
```diff
@@ -398,49 +529,296 @@ pub async fn check_connectivity(
...
+        assert_eq!(response.status(), 200);
+        assert!(
+            elapsed >= base.mul_f64(JITTER_LOW),
+            "first retry waited {elapsed:?}, expected roughly {base:?}"
+        );
+        // Guards the middleware's own `policy.backoff(attempts)` wiring, which
+        // the pure `retry_backoff` tests do not reach: feeding a 1-based
+        // counter to the 0-based `exponential_backoff` doubles this delay.
+        assert!(
+            elapsed < base.mul_f64(1.4),
```
spetz: Upper bound `1.4 × base` leaves 200ms over max jitter (1.2s) for two wiremock round trips; flaky on loaded CI. Bug band starts at 1.6s.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -385,7 +385,9 @@ impl DorisSink {
                     "Doris sink ID {} stream load returned HTTP {status}: {response_for_log}",
                     self.id
                 );
-                error!("{msg}");
+                // Per-attempt detail only: `retry_async` logs each attempt and
```
spetz: Comment says `retry_async` logs the terminal outcome. It does not (`retry.rs:283-286`), and line 438 in this file says `consume` does.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -417,36 +419,25 @@ impl DorisSink {
...
-                "Doris sink ID {} transient Stream Load failure on attempt {attempt}/{} (label={label}): {error}; retrying in {delay:?}",
-                self.id, connected.max_retries
-            );
-            tokio::time::sleep(delay).await;
-        }
+        })
+        .await
+        // `consume` already logs the terminal error for the batch, so the
+        // attempt bookkeeping is dropped here rather than logged twice.
+        .map_err(RetryFailure::into_error)
```
spetz: `into_error` drops `attempts` and `exhausted`, terminal log at line 1178 no longer says whether retries ran out or the error was permanent.

spetz (changes requested):
Also:

`.claude/skills/connector-sink/SKILL.md:86` - "Do not call `exponential_backoff` directly" contradicts the SDK's own `nack_retry_delay` at `core/connectors/sdk/src/source.rs:546`, which calls it.

`.claude/skills/connector-sdk/SKILL.md:153` - same contradiction as `connector-sink/SKILL.md:86`.

numinnex (approved):

spetz (approved):

PR #4109, opened 2026-09-09
fix(consensus): count an unservable header as a nack

krishvishal:
A view change can reach a state where an op is neither recoverable from the `DoViewChange` messages in hand nor provably uncommitted. `merge_dvc_quorum` answers `AwaitingRepair` and waits for every replica to report, so a permanently crashed replica leaves two healthy survivors unable to elect a primary. The view number keeps climbing while `log_view` stays put, and client requests retry forever.

The op was decidable from the quorum already present. A sender that holds the header but cannot serve the body counted as neither a copy nor a nack. A prepare is journaled before it is acked and nothing compacts an op above the commit point, so a missing body proves that sender never acked it, which is what a nack asserts.

`tally_op` now derives a nack from that sender. The quorum is unchanged at `replica_count - quorum_replication + 1`, so replicas provably outside the ack set still leave fewer than a replication quorum inside it, and a sender that can serve the body is untouched. The derived nack stops at the sender's own commit point, where a missing body means compaction rather than absence.

Two unit tests now assert truncation where they asserted a stall, one pins the boundary at a single servable copy, and the undecidable case moves to abstention. A simulator test replays the seed that found this: without the change the cluster reaches view 103 against `log_view` 3 with both survivors caught up, and one request retries 266 times unanswered.


core/consensus/src/dvc_merge.rs
```diff
@@ -221,21 +222,31 @@ fn tally_op<'a>(
...
-            let may_nack_implicitly = dvc.log_view < canonical_log_view;
-            match canonical {
-                // Implicit: the sender holds a DIFFERENT prepare, so not this one.
-                Some(canonical) if may_nack_implicitly && held.checksum != canonical.checksum => {
-                    nacks += 1;
+            if !offers_body && op > dvc.commit {
+                // A prepare is journaled before it is acked, so a header this
+                // sender cannot serve, above its own commit, proves it is outside
+                // the ack set. Below that commit a missing body is compaction.
+                nacks += 1;
```
numinnex, line 229: The derived nack reads the cached `local_dvc_suffix` snapshot, not the journal. Its tag is `(op, commit)` only (`impls.rs:2213`), refreshes are gated on that tag (`shard/lib.rs:10249`, `:10299`), and all five `invalidate_local_dvc_suffix` sites are removal paths, so nothing invalidates when a body lands. A backup that adopts a StartView header-only, caches that shape on a straggler SVC, then appends and acks the retransmitted op still reports it header-only here, and nacks it.

Two such votes are a nack quorum at n=3, and the `op <= commit_max` refusal cannot catch it: `head.commit` sits inside `identity_checksum`, pinning `commit_max <= N-1`. `truncate_from(N)` in `reconcile_metadata_view_divergence` (`shard/lib.rs:6041-6063`) then erases the acked op from the WAL, logged as "uncommitted". Pre-PR the same state returned `AwaitingRepair` and kept the entry.

Suggested fix: fold a journal-mutation generation into the snapshot tag, bumped on append, truncate and drain, on both planes. A directed 3-replica probe hits this in 0.09s; a 1400-seed fuzz sweep never finds the ordering.

numinnex (approved):

spetz (approved):

PR #4092, opened 2026-09-08
feat(cluster)!: let topics require durable acks and flatten config

spetz:
Producer acknowledgments did not wait for recoverable storage. The
`enforce_fsync` option synchronized segment flushes, but replicas sent
`PrepareOk` before prepares reached stable storage. Consumer-offset
synchronization was one server-wide key.

This PR adds two create-only topic options, `durability` and
`consumer_offset_durability`. Each is `replicated` (the default) or
`persisted`, and neither inherits the other. Both policies write to
disk and complete after VSR quorum commit. Persisted also requires
recoverable stable-storage copies on the quorum, or a local sync in a
single-replica group.

In replicated groups, a persisted policy adds a bounded per-partition
prepare WAL, sized by `partition.wal_bytes_max`. WAL records reference
message bodies in segment files, and hard links keep them until
reclamation. Replicas forward each prepare before their own WAL write
completes. Reclamation waits for durable materialized state, and
missing history or storage errors fail closed.

AWS i4i benchmarks found two costs outside the WAL, which this PR
removes. Produce admission zero-filled each request buffer before the
copy, and repair ran for operations that were already resident.

The configuration drops `[system]`. Its `path` key and remaining
tables move to the root, and `IGGY_SYSTEM_*` variables lose `SYSTEM_`.
This PR removes the stream, topic and partition path keys,
`archive_expired`, `recreate_missing_state` and
`consumer_offset_enforce_fsync`. The server refuses to boot with a
relocated key or a stored topic with `enforce_fsync=true`.

The Rust, Java, C#, Go, Node, Python, PHP and C++ SDKs, the CLI and
the benchmark expose both options. The HTTP `Iggy-Durability` header
reports `replicated` or `persisted` instead of `replicated-memory`.
Poll auto-commit and `ack=none` produce stay asynchronous.

A deterministic simulator storage model injects crashes, power loss
and torn writes into WAL tests. Cluster tests cover crashes and
corruption, and a compatibility test boots a baseline data directory.
The in-memory partition simulator does not run persisted topics.

PR #4119, opened 2026-09-10
fix(partitions): complete polls on the owning shard

### LINKED ISSUE #4117: A delayed disk poll can update consumer progress after a topic purge
## Summary

Disk polls run asynchronously so that file I/O does not block the shard that owns a partition. The poll captures the resources it needs, performs the read in a detached task, and completes later.

The detached poll currently also has access to shared consumer progress state. When the read finishes, it can apply an automatic commit or update a consumer group's `last_polled` value before the partition owner establishes that the result still belongs to the current message history.

This becomes incorrect when a topic is purged while the disk read is pending. Purge removes the old messages, clears consumer progress, and starts a new offset space at zero. The pending poll is not revoked. It can therefore finish with information from the history that existed before the purge and apply that information to the history that exists after the purge.

## Example

Consider an individual consumer that polls with `Next` and automatic commit enabled:

| Time | Partition state | Delayed poll |
| --- | --- | --- |
| 1 | Old messages occupy offsets 0 through 2 | A disk poll starts for offsets 0 through 2 |
| 2 | The topic is purged; messages and consumer progress are cleared | The disk read is still pending |
| 3 | Five new messages are appended at offsets 0 through 4 | The pending poll still belongs to the old history |
| 4 | The new history is ready for consumers | The old poll completes and records offset 2 |
| 5 | The consumer issues its next `Next` poll | Polling resumes after offset 2 |

The consumer can now skip the new messages at offsets 0 through 2. Checking only whether offset 2 is within the current range cannot detect the problem because offset values are reused after purge. The same number identifies different messages before and after the purge.

If fewer new messages exist, the stale progress can instead move the consumer beyond the current end of the topic and cause empty polls until the new history catches up.

## Why the current completion checks are insufficient

`PollPlan` is created while the partition is borrowed by its owning shard. To let the read continue after that borrow ends, the plan captures owned file resources and shared handles for consumer offsets, automatic commit capacity, durable offset state, and group `last_polled` state.

For a disk poll, `PollPlan::execute` runs outside the partition owner. After the file read completes, it can update the captured offset map and produce an `AutoCommitApplied` value. The server later borrows the partition again to decide whether to submit the corresponding `StoreConsumerOffset` operation. If that decision rejects the operation, the earlier local update is rolled back.

This protects some changes in partition ownership and consensus state, but it does not prove that the message history is unchanged. A purge clears the existing shared maps rather than replacing every handle captured by older polls. A plan created before the purge can therefore still point at the same allocation after the purge. The allocation is current, but the read result is not.

There is a similar path for consumer groups. A poll without automatic commit can update `last_polled` directly from the detached completion, without returning to the partition owner for history validation.

This means the current flow mixes two responsibilities:

1. Reading message data, which may wait for file I/O and may safely happen outside the owner.
2. Deciding whether the result is still valid and applying consumer progress, which depends on the partition's current history and must be serialized with changes to that history.

The second responsibility cannot be made safe by validating offset numbers alone.

## Observable impact

Depending on the poll type and timing, a completion from before the purge can:

- advance an individual consumer's offset in the new history;
- advance a consumer group's `last_polled` value in the new history;
- enqueue or replicate an automatic commit derived from the old history;
- cause a later `Next` poll to skip messages that were appended after the purge;
- authorize a reply whose message data and current offset were captured from a history that no longer exists.

The race requires a disk read to remain pending across purge, so it may be uncommon in normal tests. Slow storage, a busy runtime, or a deliberately suspended read makes the ordering reproducible.

## Expected behavior

A poll result must only affect the partition history from which it was created. If that history has been replaced before completion:

- the result must be rejected with a retryable error;
- no messages from that result should be returned to the client;
- consumer offsets and group `last_polled` values must remain unchanged;
- no automatic commit or capacity reservation from that result may survive.

This rule must still hold when the stale offset is numerically valid in the new history.

## Suggested test

A deterministic regression test can expose the bug without relying on timing:

1. Persist messages at offsets 0 through 2 so polling uses the disk path.
2. Start a poll for those messages and suspend it at the file read.
3. Purge the topic while the poll is suspended.
4. Append new messages at offsets 0 through 4 and persist them.
5. Resume the old poll and run its completion path.
6. Verify that the completion is rejected and that consumer progress remains empty.
7. Poll with `Next` and verify that all five new messages are returned, starting at offset 0.

The regression should cover an individual consumer with automatic commit and a consumer group whose `last_polled` value would otherwise change. A poll without automatic commit should also be checked to ensure that an obsolete result cannot authorize a reply.

## Possible implementation direction

Keep the asynchronous worker limited to reading messages. It should return the read fragments, relevant immutable request context, and an identity for the message history from which the read was planned. It should not hold handles that allow it to change consumer progress.

The partition owner can then receive the result, compare its history identity with the current one, perform admission and capacity checks, apply progress, and authorize the reply as one synchronous completion step. Any queued automatic commit should retain the same history identity so it can be discarded if the history changes before operation assignment.

The exact representation can be decided in the fix. The required invariant is that work created for an old partition history cannot change progress or authorize a successful poll in a newer history.

diegomrsantos:
Fixes #4117.

A disk poll can remain pending while purge replaces a partition's message history and resets its offsets to zero. Previously, the detached read worker retained shared handles that could advance consumer offsets and group progress after purge. If fresh messages reused those numeric offsets, the stale completion could make a later `Next` poll skip messages from the new history.

Poll workers now return owned read results to the partition owner. The owner validates the captured history before accepting progress or authorizing a successful reply.

## Implementation

- `PollHistoryId` is a process local `u64` identity from a checked counter. History replacement invalidates earlier reads, including when the namespace and numeric offsets are reused.
- Disk polls reserve capacity before I/O in a dedicated completion lane. Running reads and queued results retain that capacity until dequeue or discard. `sharding.poll_completion_capacity` is independent of the ordinary inbox, and exhaustion rejects before reading.
- Resident reads finish inline under the same history and admission rules. The owner services the completion lane alongside ordinary work.
- Automatic commit admission, capacity reservation and progress updates occur synchronously on the owner. A successful poll reply can precede durable offset replication. Once a prepare has been assigned, its continuation must begin immediately on the same pump. A missing partition trips an invariant instead of silently discarding it.
- A reply receiver already disconnected when completion arrives prevents admission. Disconnection can still race with that check, so the API documents recovery from explicit offsets after a lost reply.
- The owner loop reuses its shutdown receive future, avoiding repeated receive setup on each pump turn. Completion metrics are shared during shard setup, and individual polls copy their history identity without accessing the global counter.

## Verification

At `c7e6fb165`, 1,348 affected tests pass on macOS and 1,371 on ARM64 Linux, with eight existing ignored tests on each platform. Strict Clippy with all features and all targets passes for the six affected crates on both platforms. Formatting, dependency sorting with `--no-format`, TOML formatting and the source diff checks pass. GitHub workflows for the measured commit require maintainer approval. The PR subsequently advanced to `8f1c0e260` through a master merge containing #4152; its changes are confined to connectors and their tests. The server code and build inputs used by this comparison are unchanged. The test and benchmark results below remain attributed to `c7e6fb165`.

The new continuation regression proves an automatic commit prepare has been assigned but is not yet in the journal before tombstoning its namespace. It fails with the previous silent drop behavior and passes with the invariant restored. Existing suites cover delayed polls after purge, stale history with valid reused numeric offsets, consumer group progress, completion capacity, caller cancellation and reply release before replication.

The Linux boot smoke test uses `IGGY_SHARDING_CPU_ALLOCATION=all` because the container denies NUMA memory binding. Partition and shard documentation, including private items, builds successfully. The broader server documentation build encounters existing broken links outside these review edits.

## Performance

The measured candidate `c7e6fb165` was compared with upstream master `585ab6922` in the ARM64 Docker Desktop Linux VM. After high Docker memory use was reported, the existing benchmark was rerun with only its runtime container active and no concurrent builds. Earlier observations are retained separately and excluded from the rerun estimates.

The rerun completed 120 measured runs and 2,304,000 successful nonempty polls, plus 20 excluded pilot runs. All workload and existing resource checks passed. There are 12 balanced randomized pairs per case. Positive values below mean slower polls.

| Case | Mean time change, 95% interval | p99 change, 95% interval |
| --- | --- | --- |
| Explicit offset | +0.10%, [-1.35%, +1.69%] | -0.02%, [-2.79%, +2.88%] |
| Explicit offset with writes | +2.32%, [+0.41%, +4.36%] | +3.40%, [-0.62%, +7.63%] |
| Next | -0.48%, [-1.41%, +0.44%] | -0.84%, [-2.37%, +0.69%] |
| Consumer group | +0.45%, [-0.59%, +1.45%] | +0.86%, [-0.89%, +2.65%] |
| Resident | -3.77%, [-5.27%, -2.38%] | -2.33%, [-6.29%, +1.23%] |

Mean intervals use 20,000 bootstrap resamples of whole pairs. They are pointwise 95% intervals without adjustment across cases. p99 is the geometric mean of paired ratios of each run's p99, not a pooled percentile.

The concurrent write case has higher mean time, while resident reads have lower mean time. The other mean intervals span zero. The earlier Next p99 increase was not reproduced. This supports neither a blanket claim that all earlier reports still hold nor a claim that Docker memory caused their differences. Disk reads may use the guest OS cache, and these results describe this laptop VM. Native macOS performance and standalone disk batch and fsync results were not refreshed.

[Full report, raw results, frozen methods and validation logs](https://github.com/diegomrsantos/iggy/releases/tag/pr-4119-final-review-evidence-2026-09-13).

The benchmark tooling is reviewed separately in #4122. The earlier measurements and discussion remain historical evidence for their recorded revisions.

## Metric retirement

`frame_drops_total{variant="partition_auto_commit"}` is retired because the inbox operation it measured no longer exists. `frame_drops_total{variant="partition_poll_completion"}` reports disk completion capacity and return failures, while `partition_consumer_offsets_denied_total` counts consumer offset capacity denials. These measure different events and are not a direct replacement for the retired label. Unexpected poll replies now log the actual reply instead of a missing reply field that was always false.


hubcio:
#4092 will be merged first :)

hubcio:
@diegomrsantos have you joined our discord? if yes, what's your username?

diegomrsantos:
> #4092 will be merged first :)

Thanks for letting me know. I'll rebase this one onto #4092

hubcio:
@diegomrsantos which one should be reviewed/merged first, this one or #4122?

diegomrsantos:
I’d prioritize this one, since it fixes the correctness issue. #4122 contains the benchmark tooling and works against master without this fix. Neither PR depends on the other, so they can be reviewed independently, but I’d merge this one first.

diegomrsantos:
Added a short overview of poll completion and concise docs for the new types, fields, and functions. These explain history validation, reservation lifetimes, admission checks, and why poll replies can precede durable offset commits. This update changes documentation only.

diegomrsantos:
Added concise docs explaining the reservation ownership boundary. Previously, `AutoCommitSubmit` carried an existing guard through the shard inbox, whose `Send` bound applied even to sends to the same shard. Now the inbox carries read results, and the owner creates and retains guards in local request or replication state. This permits `Rc` and `Cell` while preserving synchronous counter updates. The docs also explain why capacity checking and acquisition must stay in one owner operation: individual atomics would not protect admission split across owner turns. This update changes documentation only.

core/server/src/dispatch/partition.rs
```diff
@@ -830,6 +489,19 @@ where
             ReadPolledMessagesError::Fallback(empty_poll_fallback(partition_id))
         }),
         Some(PartitionReadReply::Rejected(error)) => Err(ReadPolledMessagesError::Rejected(error)),
+        None => {
```
spetz, line 492: Maps every `None` from `partition_read` to `ShardCommunicationError` (11001). Three of those cases prove the read never started: `core/shard/src/lib.rs:1998` unroutable namespace, `core/shard/src/lib.rs:2012` no sender, `core/shard/src/lib.rs:2013` full inbox. Base answered an empty poll here.

core/partitions/src/iggy_partition.rs
```diff
@@ -15148,6 +15804,252 @@ mod retention_tests {
...
+        }
+    }
+
+    async fn assert_delayed_poll_preserves_fresh_progress(
+        auto_commit: bool,
+        fresh_message_count: u32,
+    ) {
+        let config = repair_config();
+        let (_directory, mut partition) = Box::pin(disk_poll_partition(&config)).await;
+        let consumer = PollingConsumer::Consumer(7, 0);
```
spetz, line 15851: No-auto-commit purge test uses `PollingConsumer::Consumer(7, 0)`. On base an individual consumer with `auto_commit=false` updates neither `consumer_offsets` nor `last_polled_offsets`, so every assertion passes without the fix.

core/shard/src/poll.rs
```diff
@@ -0,0 +1,425 @@
...
+            };
+            let frame =
+                ShardFrame::lifecycle(LifecycleFrame::PollCompleted(Box::new(PollCompleted {
+                    namespace: self.namespace,
+                    result,
+                    reply: self.reply,
+                    #[cfg(feature = "poll-diagnostics")]
+                    queued_at: Some(std::time::Instant::now()),
+                })));
+            if let Err(error) = inbox.try_send(frame) {
```
spetz, line 295: `PollCompletionSender::complete` calls `try_send` on the main lane. A full inbox discards a finished read and answers `TransientNotAccepted`.

core/shard/src/poll.rs
```diff
@@ -0,0 +1,425 @@
...
+                            tier = "disk",
+                            "partition poll dispatch"
+                        );
+                        let completion = completion::PollCompletionSender::new(
+                            self.senders.get(usize::from(self.id)).cloned(),
+                            namespace,
+                            reply,
+                            self.metrics.clone(),
+                        );
+                        self.bus.spawn(read_poll(namespace, plan, completion));
```
spetz, line 106: Adds one main-lane frame, one pump turn and one heap allocation per disk poll. The completion queues behind arbitrary consensus and write frames. This is the measured `disk-offset` regression.

core/shard/src/poll.rs
```diff
@@ -0,0 +1,425 @@
...
+                namespace_raw = namespace.inner(),
+                phase = "completion",
+                tier = "disk",
+                queue_wait_us = u64::try_from(queued_at.elapsed().as_micros()).unwrap_or(u64::MAX),
+                "partition poll completion"
+            );
+        }
+        let partitions = self.plane.partitions();
+        let consumer_kind = result.consumer_kind();
+        match partitions.complete_poll(&namespace, result) {
```
spetz, line 186: Commits progress at `core/partitions/src/iggy_partition.rs:2901`, then tries the reply at `core/shard/src/poll.rs:195` and discards the result. Budget expiry at `core/shard/src/lib.rs:2021` drops the receiver while the client still waits. Needs `auto_commit=true` with `Next`.

core/server_common/src/poll.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Identity of one serviceable message history. It is never serialized.
+///
+/// The allocation identifies the history even if a rebuilt partition reuses
+/// its namespace and offsets. Pending reads keep it alive, preventing address
+/// reuse while they can still complete. `Arc` also keeps completion frames
+/// `Send`, as required by the shard inbox.
+/// `Default` creates a fresh identity. Only clones compare equal.
+#[derive(Clone, Debug, Default)]
+pub struct PollHistoryId(Arc<()>);
```
spetz, line 32: `PollHistoryId` is an `Arc<()>` compared by address where a `u64` counter serves. Zero lines saved, one allocation removed per poll.

core/server/src/dispatch/partition.rs
```diff
@@ -830,6 +489,19 @@ where
...
+            // A timeout drops the receiver without canceling the owner. Its
+            // eventual completion may still accept progress, so neither an
+            // empty success nor TransientNotAccepted describes this outcome.
+            warn!(
+                transport_client_id,
+                namespace = namespace.inner(),
+                "partition read reply unavailable; acceptance unknown"
+            );
+            Err(ReadPolledMessagesError::Rejected(
+                IggyError::ShardCommunicationError,
```
spetz, line 502: Binary transports answer 11001 while HTTP answers 504 at `core/server/src/http/handlers.rs:1322`. The parity rule is stated at `core/server/src/http/error.rs:474`.

core/simulator/src/lib.rs
```diff
@@ -1701,6 +1706,87 @@ mod tests {
...
+        let owner = Rc::clone(sim.replicas[0].partition_shard(namespace));
+        let consumer = PollingConsumer::Consumer(7, 0);
+        let args = PollingArgs::new(iggy_common::PollingStrategy::next(), 1, true);
+        let plan = owner
+            .plane
+            .partitions()
+            .build_poll_snapshot(&namespace, consumer, &args)
+            .expect("poll snapshot");
+        assert!(
+            !plan.needs_off_pump_io(),
```
spetz, line 1752: The one new simulator test asserts `!plan.needs_off_pump_io()`, and no disk tier exists. The new inbox hop is never modeled.

core/simulator/src/lib.rs
```diff
@@ -1407,12 +1407,16 @@ impl Simulator {
...
-        let (fragments, _commit_offset, auto_commit) = futures::executor::block_on(plan.execute())?;
-        if let Some(applied) = auto_commit {
-            applied.admit(|_| Ok::<(), IggyError>(()))?;
+        let partitions = shard.plane.partitions();
+        let completion =
+            partitions.complete_poll(&namespace, futures::executor::block_on(plan.execute()))?;
+        if let Some(replication) = completion.replication {
+            // SimOutbox sends enqueue immediately, so replication cannot suspend.
+            futures::executor::block_on(
+                partitions.replicate_poll_completion(&namespace, replication),
```
spetz, line 1416: Comment states replica sends never suspend, but `core/simulator/src/bus.rs:234` awaits under `cfg(test)`, which deadlocks the `block_on` at `core/simulator/src/lib.rs:1415`.

spetz (changes requested):

core/shard/src/poll.rs
diegomrsantos: Agreed. With `auto_commit=true`, the internal timeout can drop the receiver while the owner still accepts the result and advances the cursor. A subsequent `Next` can then skip the batch the client never received.

The previous implementation also advanced progress before attempting the reply. On timeout, however, it returned a successful empty poll, even though the detached read could still finish and advance progress. This PR returns `ShardCommunicationError` instead, so the caller is told that the read failed to produce a reply rather than being given an empty success. That improves the reported outcome, but it does not cancel the read or prevent later progress updates. The new completion queue also adds another delay window.

I’ll add a regression that delays a nonempty completion past the timeout, then checks the cursor and a subsequent `Next`. What contract should we guarantee here? Should timeout cancel any poll that has not yet been admitted, ensuring it cannot advance progress afterward? Or is eventual admission acceptable if the caller receives an error indicating an unknown outcome? Once admission has happened, a lost reply can still leave the outcome unknown.

core/server/src/dispatch/partition.rs
diegomrsantos: Good point. `partition_read` already combined submission failures and unavailable replies into `None` before this PR. The binary poll handler previously converted all those cases into a successful empty poll. I changed that to `ShardCommunicationError` because a timeout can leave the owner running and advancing progress, making empty success misleading.

However, that mapping is too broad for the cases you identified. If routing fails, the sender is missing, or the initial inbox rejects the request, we know this poll never reached the owner.

I propose returning `TransientNotAccepted` for those known submission failures, allowing the existing SDK retry logic to handle them. Timeouts and lost replies after submission would retain `ShardCommunicationError`, since acceptance remains uncertain.

Does that distinction make sense to you, or would you prefer a different error mapping or approach?

core/shard/src/poll.rs
diegomrsantos: Agreed. Once the disk read finishes, the `complete` method on `PollCompletionSender` tries to enqueue the result in the owner’s main inbox. If that inbox is full, it drops the completed result and attempts a `TransientNotAccepted` reply. This is deliberate behavior, and the existing test explicitly verifies it.

The rejection preserves progress safety. The owner has not accepted this completion, so this poll has neither advanced the consumer cursor nor admitted an automatic commit. `TransientNotAccepted` is therefore appropriate even though the disk read already happened. The concern is wasted work. Under sustained pressure, retries can repeatedly read and discard the same messages, adding load.

A related failure path existed before this PR. Polls that needed to submit an automatic commit could already reject after reading their messages if the owner’s inbox refused that submission. However, polls without automatic commits, or whose commits needed no submission, could reply directly. This PR makes every disk poll depend on completion inbox capacity. Resident polls still complete inline.

We need to retain owner validation before allowing progress or a successful reply, since purge or partition replacement can happen while disk I/O is pending. I see two possible improvements.

- **Wait asynchronously for inbox capacity from the detached worker.** This retains the completed batch instead of immediately discarding it. The current sender’s ordinary `send` blocks, so using it could deadlock the shard thread that must drain the inbox. We would need an asynchronous sender, with waiting outside the pump. This also requires a bound on outstanding reads or retained bytes. A bounded inbox alone does not bound waiting workers. Completions would still compete with other main inbox traffic.
- **Reserve dedicated completion capacity before starting disk I/O.** Each admitted read would then have space to return its result. If capacity is exhausted, we could defer or reject before doing the read. This addresses wasted I/O and competition with other traffic more directly, but introduces additional admission and scheduling machinery, including fair service for completions.

Either approach needs explicit timeout and shutdown handling, because the requester’s timeout currently does not cancel detached work. History validation must still happen when the owner accepts the result.

The current policy favors immediate rejection and releasing memory. Would you prefer retaining completed reads with bounded asynchronous waiting, reserving completion capacity before reading, or another approach?

core/shard/src/poll.rs
diegomrsantos: Thanks, I agree this extra step deserves attention. In `disk-offset`, the client tracks its next offset itself and automatic commits are disabled. The old worker could therefore reply directly after reading. With this PR, it allocates a completion and waits for the owner to process it before replying, so there is additional overhead in this case.

The measured slowdown is consistent with that extra work. Since the results describe an earlier revision, I’d like to rerun the comparison on the current code and check how much time goes into allocation and waiting for the owner. This case has no concurrent producer, so testing with concurrent writes would also help us understand the queue contention you mentioned.

The completion queue options discussed in the other thread seem worth exploring. We can preserve the owner’s history validation while looking for a cheaper way to return completed reads, then use the measurements to guide the choice.

core/partitions/src/iggy_partition.rs
diegomrsantos: Thanks, that makes sense. With an individual consumer and `auto_commit=false`, the old implementation did not update consumer progress, so the unchanged-offset and `Next` assertions do not demonstrate that the fix prevented a stale update.

The test still checks that the old result is rejected with `TransientNotAccepted`. There is also a separate group test using a constructed result, but we are missing a group test that exercises a real disk read across purge.

I’ll add that case with automatic commits disabled and ensure the delayed read produces a nonempty result. After purge and rejection of the old completion, it should verify that both the group’s last polled and committed offsets remain unset, before any fresh group read changes them. This would catch the old behavior where a delayed read restores a stale last polled mark after purge, potentially delaying partition handoff.

core/simulator/src/lib.rs
diegomrsantos: Thanks, you’re right. That test deliberately uses a resident read to verify that the client receives its reply while replication is stalled. It does not exercise the extra inbox step used by disk completions.

The existing disk purge test calls the completion handler directly, while the channel tests check enqueueing and rejection without processing the completion through the owner. We are missing coverage of those pieces working together.

I’ll add a test that holds back a nonempty result, purges or replaces the partition, and then delivers the old result through the actual completion sender and inbox. The owner should reject it without advancing progress. A fresh result should also succeed through the same route.

We can control when the result is released without building a complete disk model into the simulator, while keeping the existing disk test for actual I/O coverage.

core/simulator/src/lib.rs
diegomrsantos: Thanks for catching this. The comment is too strong. Replica sends normally finish immediately, but the new test mechanism can pause them, so the helper cannot safely assume that replication will never suspend.

Before this PR, the helper used `block_on` for the resident read but did not replicate automatic commits. This PR adds the replication wait alongside the ability to pause a replica send. If a test calls the helper while that pause is active, it can block before reaching the instruction that releases the send.

The existing test avoids this by using the asynchronous path, but the helper should handle that situation safely too. I’ll adjust it to use the simulator’s normal asynchronous execution path so the calling test can continue and release paused work, and add regression coverage for this combination.

core/server_common/src/poll.rs
diegomrsantos: Thanks, I think a numeric ID is worth trying here. The current allocation is created when a history is created or invalidated, and polls clone the existing `Arc` without allocating a new history object. Switching to a number would still remove the reference count updates from each poll.

We can keep the implementation small by using an `AtomicU64` shared by the process to assign fresh history IDs, with a checked update that prevents wrapping and reuse. The counter would run only when the history changes. Each poll would carry a plain `u64`, so it would not access the atomic counter. This also preserves uniqueness when a partition is recreated with the same namespace.

I’ll retain the tests for partition replacement and stale results, and include the change in the planned performance comparison. It removes some polling overhead, although we should measure its effect separately from the completion allocation and queue processing.

core/server/src/dispatch/partition.rs
diegomrsantos: Thanks for pointing this out. HTTP already returned `504` when the partition owner did not reply before this PR, while binary transports returned a successful empty poll. This PR changes the binary response to `ShardCommunicationError` with code `11001`, so both now report a failure, but their error identities remain inconsistent.

The numbers represent different things. `11001` is an Iggy error code, whereas `504` is an HTTP status. However, the HTTP response also uses `504` as its JSON error ID, with `partition_read_timeout` as its code. We could preserve an appropriate HTTP status while exposing the same Iggy error identity in the body.

I suggest aligning this with the distinction discussed in the other thread. When we know the request was never submitted, both transports could return `TransientNotAccepted`, which already maps to HTTP `503`. When submission succeeded but no reply arrived, we should preserve the unknown outcome. For an actual timeout, HTTP could retain `504` while including error ID `11001`, without implying that the original poll had no effect.

This would need an explicit HTTP mapping. Passing `ShardCommunicationError` through the current generic handler would otherwise produce `400 Bad Request`, which would incorrectly attribute the failure to the client.

Does that approach match the contract you have in mind, or would you prefer a different mapping?

core/partitions/src/iggy_partition.rs
diegomrsantos: Done in [70268ab7b](https://github.com/apache/iggy/commit/70268ab7b98c3573b1b7b02ee277cc00fc3c1556). The new test covers a real group disk read across purge with automatic commits disabled. It checks that the delayed result is nonempty and that rejecting it leaves both `last_polled` and the committed offset unset before any fresh read.

core/shard/src/poll.rs
diegomrsantos: Added the regression in [ad9fc761b](https://github.com/apache/iggy/commit/ad9fc761b03ba1cc64ed25897817a1c6cfabbb92). It holds a nonempty result past the requester timeout with `auto_commit=true`, verifies that the reply channel has closed, then delivers the result through the completion sender and owner inbox. The test records the current behavior: local progress advances through offsets 0–2, and a subsequent `Next` returns only offset 3, skipping the messages the caller never received. This documents the behavior while leaving the timeout and cancellation contract open.

core/simulator/src/lib.rs
diegomrsantos: Fixed in [13afb497a](https://github.com/apache/iggy/commit/13afb497ab56f3f49e7810d893cad610f54e6d1e). `Simulator::poll_messages` now returns an owned future and routes the poll through the partition owner, removing both `block_on` calls. The regression uses the helper with a replica send paused, checks that the nonempty reply is available before release, and then verifies that the automatic commit reaches every replica after replication resumes. The simulator demo also uses the asynchronous helper. All 143 simulator tests passed on both macOS and Linux, with five existing ignored tests on each platform.

core/simulator/src/lib.rs
diegomrsantos: Added in [a699ff91c](https://github.com/apache/iggy/commit/a699ff91c6b9126502ea112ac676d96d8d107975). The new regression holds a group read result, replaces the partition with different messages at the same offsets, and delivers the old result through the actual completion sender, inbox, router, and owner pump. It checks `TransientNotAccepted` and both group offsets remaining unset before a fresh result succeeds through the same route. In an isolated copy, disabling the history check made the test fail because the stale result restored `last_polled` to `Some(2)`, confirming the delayed result is nonempty and the regression catches stale progress. All 99 shard tests and strict Clippy passed on both macOS and Linux.

core/shard/src/poll/timeout_tests.rs
```diff
@@ -0,0 +1,277 @@
...
+        panic!("expected the subsequent poll request");
+    };
+    owner.on_partition_read(namespace, read, reply).await;
+    let Some(PartitionReadReply::Poll { fragments, .. }) = next_poll.await else {
+        panic!("subsequent Next must return messages successfully");
+    };
+    assert_eq!(
+        message_offsets(&fragments),
+        vec![3],
+        "Next skips offsets 0 through 2, which the caller never received"
```
spetz: Contract: a poll that cannot deliver its reply must not advance progress. Duplicate beats skip, and that is the invariant this PR restores for purge.

 `crossfire::Sender::is_disconnected()` is on the reply sender. Probe it in `complete_poll` before `admit_poll_auto_commit`. A timed-out caller is already gone when the read lands, so the completion rejects and the next poll re-reads. The race window drops from the whole disk read to a few instructions, and losing it costs a duplicate.

Then flip this assertion to re-delivery. Pinned as-is, the next reader takes the skip for the intended contract.

core/server/src/dispatch/partition.rs
spetz: Yes. Keep 504 as the status and carry 11001 as the body error id. Add the
explicit mapping, the generic handler would give 400 and blame the client.

core/shard/src/poll.rs
```diff
@@ -0,0 +1,461 @@
...
+                    Some(plan) if plan.needs_off_pump_io() => {
+                        #[cfg(feature = "poll-diagnostics")]
+                        tracing::debug!(
+                            target: "iggy.shard.poll_diagnostics",
+                            namespace_raw = namespace.inner(),
+                            phase = "dispatch",
+                            tier = "disk",
+                            "partition poll dispatch"
+                        );
+                        let completion = completion::PollCompletionSender::new(
```
spetz: Needs the re-measurement before merge, with a concurrent producer. This hop is
on every disk poll and the posted numbers describe an earlier revision.

Report writes per poll next to latency. The completion adds a frame and an
allocation, no I/O, so a surviving regression is queueing behind consensus and
write frames, which the reserved lane also fixes.

core/server/src/dispatch/partition.rs
spetz: Yes, split it that way. TransientNotAccepted for the three that prove the read
never started, ShardCommunicationError only once submission succeeded.

One addition. Do not just report the uncertainty, shrink it. Probe
crossfire::Sender::is_disconnected() on the reply sender in complete_poll
before admit_poll_auto_commit. A timed-out caller is already gone when the read
lands, so the completion rejects and nothing advances. ShardCommunicationError
then means a rare lost reply instead of the common case.

core/shard/src/poll.rs
spetz: Cancel it. A poll that cannot deliver its reply must not advance progress.
Duplicate delivery is fine, a silent skip is not, and that is the same
invariant this PR restores for purge.

Eventual admission with an error is not acceptable. The client cannot act on
"unknown outcome" and the skip is silent from its side.

crossfire::Sender::is_disconnected() is on the reply sender. Probe it in
complete_poll before admit_poll_auto_commit and reject when the caller is
gone. That covers the real case, since a timed-out caller is already gone when
the read lands. The remaining race is a few instructions wide and costs a
duplicate, not a skip.

Then flip the new test to assert re-delivery. Pinned as a skip it reads as the
intended contract.

core/shard/src/poll.rs
spetz: Re-measure before merge, with a concurrent producer. Give writes per poll next to latency. The completion adds a frame and an allocation, no I/O, so if the regression survives it is queueing behind consensus and write frames, which the reserved lane in the other thread also fixes.

core/partitions/src/iggy_partition.rs
```diff
@@ -2840,19 +2866,282 @@ where
...
+    /// Accept a read only while its message history still belongs to this owner.
+    /// Validation, admission, and progress updates must stay in one synchronous
+    /// owner turn so purge or recovery cannot interleave between them.
+    /// Nonempty group reads update `last_polled` even without automatic commits.
+    /// Rejection leaves this read's progress unapplied.
+    pub(crate) fn complete_poll(
+        &mut self,
+        result: PollReadResult,
+    ) -> Result<PollCompletion, IggyError> {
+        self.resynchronize_consumer_offset_reservations();
```
spetz: Move this below the history guard. Every rejected stale completion pays the
reconciliation before being thrown away.

core/shard/src/poll.rs
spetz: Do not wait on the inbox. A worker waiting for capacity keeps the batch and the read handle for as long as the pressure lasts, so memory and descriptors grow without bound.

Size the completion lane to the maximum in-flight disk polls instead, which the read budget already caps, or reserve the slot when the poll is dispatched. Then try_send cannot fail for capacity. Leave the unroutable and disconnected arms alone.

core/server_common/src/poll.rs
spetz: Do it here, not in a follow-up. The point is the invariant, not the allocation. Arc::ptr_eq only works because pending reads keep the old allocation alive so the address cannot be reused. A u64 counter cannot alias at all, and it is Send, so the "keeps frames Send" note does not argue for Arc either.

spetz (changes requested):

core/shard/src/poll.rs
diegomrsantos: Agreed. A completion arriving after the requester has timed out should leave progress unchanged. I’ll add the disconnect check before admission and change the regression to verify that the next poll returns the unseen messages.

There is one remaining ordering detail I’d like to account for. The task awaiting the reply and the partition owner can run on different shards. In [`partition_read`](https://github.com/apache/iggy/blob/a699ff91c6b9126502ea112ac676d96d8d107975/core/shard/src/lib.rs#L1992), the requesting shard sends the read to the shard selected by the namespace, then waits on its own reply receiver with a timeout. One thread per shard therefore still allows the requester’s timeout to run concurrently with the owner’s completion handler.

With the proposed check, this sequence remains possible:

1. The owner checks `is_disconnected()` and sees that the receiver is present.
2. The requesting shard times out and drops the receiver.
3. The owner admits the automatic commit and advances progress through offset 2.
4. The owner attempts to return messages 0–2, but the reply cannot be delivered.
5. The next `Next` poll starts at offset 3.

The check prevents the case where the requester is already gone when completion is processed. In the sequence above, however, the remaining race still skips unseen messages because progress advances before the reply attempt. The absence of an `.await` protects the owner’s state from interleaving on its own thread, but it does not prevent another shard from closing the receiver.

To make cancellation and admission mutually exclusive across shards, they need a shared decision: if cancellation wins, the owner must leave progress untouched; if acceptance wins, the requester must preserve the reply path instead of abandoning it when its timer fires. Checking whether the receiver is connected only observes its state at that moment.

I suggest keeping the disconnect check and the redelivery regression, and adding a deterministic test that places timeout between the check and admission. That would let us verify the additional coordination needed for the stated invariant. This concern is specifically about the internal timeout between the requesting shard and the owner.

diegomrsantos: I looked into existing concurrency and messaging contracts to make the shared decision mentioned above more concrete.

If we want **confirmed cancellation to guarantee that the poll cannot advance progress**, the requester and owner need to compete for one explicit state transition:

```text
Pending → Cancelled
Pending → Completing → Completed(result)
```

An atomic compare-and-exchange could decide which side wins:

- **Cancellation wins:** the owner discards the result without updating consumer progress, group `last_polled`, or admitting an automatic commit.
- **Completion wins:** cancellation can no longer succeed. The owner finishes validation and admission, applies the permitted effects, and publishes the outcome. That outcome can still be a rejection if validation or admission fails.

The partition state would remain exclusively owned by its shard. Only the request’s cancellation decision needs to be shared. This follows an established executor pattern: Python’s [`Future.set_running_or_notify_cancel()`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.set_running_or_notify_cancel) arbitrates between cancellation and execution, with cancellation no longer succeeding once execution has won.

The requester’s behavior is essential to that contract. If its timer fires after the owner has claimed completion, it must preserve the receiver and wait for the outcome to maintain the internal handoff guarantee. That may extend the wait beyond the original deadline. Returning an error with an unknown outcome instead would describe the uncertainty, but would not recover the unseen messages or satisfy the no-skip requirement.

I also found that [Crossfire explicitly documents](https://docs.rs/crossfire/latest/crossfire/struct.MTx.html#method.try_send) that a buffered send can succeed while the receiver concurrently disappears, leaving nobody to receive the message. Therefore, moving `try_send()` before the effects would not establish delivery either. It could additionally expose success before admission fails. Rolling back after a failed send would require undoing all associated progress and consensus admission, rather than simply restoring one offset.

This distinguishes two atomicity questions. We can make cancellation and permission to apply effects mutually exclusive within the process. Actual receipt by the external client requires a delivery and recovery protocol: a connection can fail after the owner completes but before the client observes the reply. [gRPC documents that same ambiguity](https://grpc.io/docs/what-is-grpc/core-concepts/#rpc-termination). Our poll reply also currently acknowledges admission rather than a durable offset commit, so admission, durability, and receipt are separate boundaries.

For reliable consumption across those failures, established messaging systems retain eligibility for redelivery until the consumer acknowledges the required work. When processing must be protected, acknowledgement follows processing; failure before acknowledgement can then produce duplicates. [RabbitMQ documents this tradeoff](https://www.rabbitmq.com/docs/confirms#consumer-acknowledgement-modes-and-data-safety-considerations), including the weaker safety of automatic acknowledgement.

Kafka offers a useful comparison, but its automatic commits operate through the consumer client. Its current position and committed recovery position are separate. Obtaining at-least-once behavior with automatic commits requires processing all records returned by `poll()` before subsequent polls or closing. That differs from the server admitting the current batch’s automatic commit before replying. [Kafka consumer documentation](https://kafka.apache.org/41/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html)

Another possible recovery contract is replayable polling: retrying the same request ID recovers the same batch without advancing again. Here, the effects and the recorded request outcome need to be committed together, following the atomicity requirement described in [AWS’s idempotency guidance](https://aws.amazon.com/builders-library/making-retries-safe-with-idempotent-APIs/). For Iggy, this would also require defining batch retention, request expiry, and recovery across owner failure.

Any broader delivery contract must include consumer-group `last_polled`, which advances even without automatic commits. Disabling `auto_commit` alone therefore would not cover every path. The shared cancellation decision addresses the internal timeout; acknowledgements or replayable requests provide recovery when the external reply is lost.

diegomrsantos: Added the disconnect guard in [7cdf64cfa](https://github.com/apache/iggy/commit/7cdf64cfa6096d8a1cf1fa807cd22c41469015bb). `on_poll_completed` now discards the result when the reply receiver is already disconnected, before `complete_poll` can admit an automatic commit or change progress. The same commit updates the timeout regression to require redelivery. The race between the connection check and admission remains as discussed above.

core/shard/src/poll/timeout_tests.rs
diegomrsantos: Updated in [7cdf64cfa](https://github.com/apache/iggy/commit/7cdf64cfa6096d8a1cf1fa807cd22c41469015bb). The regression now requires an unchanged cursor, no automatic commit queued or assigned, and offsets `[0, 1, 2, 3]` on the next poll after the original requester times out. It fails without the disconnect guard and passes with it. This covers timeout before owner completion; the concurrent cancellation race is discussed in the [completion thread](https://github.com/apache/iggy/pull/4119#discussion_r3993707768).

core/shard/src/poll.rs
diegomrsantos: I completed the remeasurement and a separate experiment with reserved completion
capacity. The reserved lane passed its scoped correctness checks on a699ff91c, but it
did not show a performance benefit, so it remains an experimental patch.

The first comparison used a699ff91c6b9126502ea112ac676d96d8d107975 against
dc2b382097ed7fb6cfeb8a31fed0fbc74af4433e, the master revision already merged
into the PR. Throughput was 4.01% lower without concurrent writes, with a 95%
interval of 1.94% to 6.26% lower, and 5.41% lower with writes, with an interval
of 2.59% to 8.41% lower.

Both campaigns predate 7cdf64cfa6096d8a1cf1fa807cd22c41469015bb, which adds the
disconnected receiver check. These results do not establish the latest PR
revision's performance.

The second comparison used a699ff91c as the baseline and added a reserved
completion lane. It keeps the existing allocation and owner history validation.
Throughput changed by +0.64% without writes, interval −0.64% to +2.40%, and
−3.42% with writes, interval −7.67% to approximately zero. With writes, p99
latency changed by +13.86%, interval +0.31% to +29.94%.

For the concurrent producer cases, medians across runs were:

| Experiment and version | Producer batches per poll | Written messages per poll | Median poll latency | p99 poll latency |
| --- | ---: | ---: | ---: | ---: |
| First comparison, master | 0.586 | 58.60 | 268.5 µs | 575.5 µs |
| First comparison, PR | 0.613 | 61.28 | 278.5 µs | 603.5 µs |
| Lane experiment, PR | 0.564 | 56.37 | 271.0 µs | 507.5 µs |
| Lane experiment, reserved lane | 0.574 | 57.40 | 274.0 µs | 574.0 µs |

The producer target is 50 MB/s in batches of 100 messages. These are logical
producer operations, not disk syscalls. The two experiments ran at different
times, so their absolute values should not be compared across experiments.

Each case has ten pairs of 33,000 explicit offset polls, with five pairs in each
order, one shard and one sequential TCP consumer. Each poll returns one 256 byte
message; automatic commits are disabled. Percentage changes are geometric means
of paired ratios, and the intervals resample whole pairs.
Each experiment also has four baseline repeatability pairs per case. All
1,848,000 clean polls in each experiment passed the accounting audit.

This is a Linux VM on an M1 Pro with disk data that may be cached. One identical
binary pair in the lane experiment differed by 8.54% in throughput. The mixed
throughput interval ends very close to zero, so I would not claim a universal
slowdown from it. It does not support calling this implementation a performance
fix. Separate diagnostics also did not show a handoff improvement.

[Reports and evidence bundles](https://github.com/diegomrsantos/iggy/releases/tag/pr-4119-benchmark-evidence-2026-09-12) contain the source patch, revisions,
build and run commands, configuration, raw observations, diagnostic results,
validation logs, analysis scripts and SHA256 hashes. The evidence is hosted
on my fork; the lane change has not been added to this PR.

core/shard/src/poll.rs
diegomrsantos: I tested the option that reserves capacity when the disk poll is dispatched.
Each pending read owns one slot until its result is dequeued. Completed reads
use a dedicated bounded lane, and the owner processes one completion after
ordinary work so both lanes keep making progress. A read is rejected before
disk I/O if no slot is available. Finished reads do not wait for inbox space.
The missing and disconnected route checks remain, as does owner history
validation.

The new regression fills the ordinary inbox, then returns two nonempty reads.
It verifies delivery and observes an ordinary request between their acceptances.
It fails if completion service is removed or if the entire completion queue
is drained at once. Additional tests cover reservation release and shutdown, plus the behavior
after caller cancellation at a699ff91c. In total, 103 shard tests and 143
simulator tests passed on macOS and Linux.

This experiment used a699ff91c, before 7cdf64cfa added the disconnected receiver
check. It preserves the earlier timeout characterization. Adopting the lane
would require rebasing it and updating that regression for the new policy.

The performance comparison did not show an improvement, and the point estimates
were worse with a concurrent producer. The implementation remains an experimental patch. The [performance reply](https://github.com/apache/iggy/pull/4119#discussion_r3994067578) and the [evidence bundle](https://github.com/diegomrsantos/iggy/releases/tag/pr-4119-benchmark-evidence-2026-09-12) contain the measurements and
[exact patch](https://github.com/diegomrsantos/iggy/releases/download/pr-4119-benchmark-evidence-2026-09-12/reserved-lane-candidate.patch).

core/shard/src/poll.rs
diegomrsantos: Following up on the extra pump turn: profiles pointed to CPU work in the shutdown receive. `run_message_pump` recreates `stop.recv().fuse()` on every turn, repeating Crossfire’s initial backoff and receive registration while the channel is empty. I tested a local patch that pins this future once outside the loop and reuses it. Completions still return to the owner for validation.

Against `7cdf64cf`, I ran four pairs per load on the same ARM64 Linux VM: 16 runs of 33,000 explicit offset polls, one sequential TCP consumer, 256 bytes per message and automatic commits disabled.

| Load | Elapsed µs per poll, before → after | Throughput change | Written messages per poll, before → after |
| --- | ---: | ---: | ---: |
| No producer | 304.27 → 273.41 | +11.27% | 0 → 0 |
| Producer targeting 50 MB/s | 369.44 → 340.29 | +8.14% | 71.51 → 66.30 |

Elapsed time and write counts are means across runs; throughput changes are geometric means of paired ratios. The producer sustained 48.5–50.4 MB/s in batches of 100. All four pairs improved without writes; three improved with writes, where results were noisier. All polls passed the audit. The 244 shard and simulator tests passed, including new shutdown coverage.

The patch is still local and these results are preliminary. Since master also recreates this receive, the next comparison will test preserved master `dc2b38209` and PR `a699ff91c`, each with and without reuse, to measure the remaining gap under both loads.

diegomrsantos: Following up with longer measurements of the shutdown receive reuse patch. The patch is now included in [3d9116daa](https://github.com/apache/iggy/commit/3d9116daab434f79fbe65b2757e1df6e4a349d1a). It creates and pins `stop.recv().fuse()` once before the owner loop. Shutdown stays first in the biased select, and disk completions still return to the owner for validation.

This addresses an existing cost that the PR exercises more often. In the earlier profiles of master `dc2b38209` and PR `a699ff91c`, sampled CPU time in the shutdown receive function rose from about 4.1 to 7.8 µs per poll without concurrent writes, nearly doubling. With writes it rose from 6.83 to 10.35 µs per poll. The extra owner turn can recreate the receive and repeat Crossfire's backoff for an empty channel. These are sampled CPU estimates, not an additive explanation of the entire elapsed time regression.

A fresh comparison of `7cdf64cfa` against that revision plus reuse used 12 pairs per workload. These longer results supersede the preliminary gains above:

| Load | Elapsed µs/poll, before → after | Written messages/poll, before → after | Time reduction, 95% interval |
| --- | ---: | ---: | ---: |
| No producer | 270.009 → 262.870 | 0 → 0 | 2.64% [1.55%, 3.57%] |
| Producer targeting 50 MB/s | 310.369 → 298.296 | 60.97 → 58.73 | 3.89% [2.59%, 5.33%] |

Fresh resident and consumer group checks, with 12 pairs each, also showed smaller mean elapsed times. The patch passed 244 shard and simulator tests, including shutdown wakeup, draining queued completions and sender disconnection. Five existing simulator tests remain ignored.

I then ran a fresh comparison of `7cdf64cfa` plus reuse against preserved master `dc2b38209`, with 64 pairs per workload:

| Load | Master → PR + reuse, µs/poll | Written messages/poll, master → PR + reuse | Slowdown, 95% interval |
| --- | ---: | ---: | ---: |
| No producer | 234.201 → 234.771 | 0 → 0 | +0.24% [−0.08%, +0.61%] |
| Producer targeting 50 MB/s | 258.465 → 260.416 | 50.90 → 51.34 | +0.75% [+0.34%, +1.20%] |

All 256 runs completed: 8,448,000 measured polls, with 33,000 polls after 100 warmup polls per run. Each poll returns one 256 byte message, using explicit offsets with automatic commits disabled. Version order was randomized and balanced within each workload. All runs passed the raw workload audit, and an independent verifier reproduced the primary estimates and intervals. The concurrent producer sustained 50.09–50.60 MB/s in batches of 100 messages.

Elapsed times and written messages per poll are means across runs. The mean time intervals use 20,000 resamples of whole pairs. The two comparisons ran at different times, so absolute values should be compared within each table. Writes are logical producer operations, not disk syscalls.

The same fresh comparison measured p99 poll latency:

| Load | p99 change | Adjusted interval |
| --- | ---: | ---: |
| No producer | +1.57% | −0.79% to +5.20% |
| Producer targeting 50 MB/s | −1.33% | −4.13% to +0.66% |

These are geometric means of paired ratios of p99 values from individual runs. Positive values mean higher latency. The intervals use 100,000 resamples of whole pairs and retain the preregistered 98.75% bounds per workload, accounting for both workloads and two planned checkpoints. With concurrent writes, the upper bound is below the 3% slowdown threshold. Without writes, that threshold remains unresolved. All pairs remain included.

In these ARM64 Linux VM workloads, the estimated remaining mean overhead is below 1%, and both 95% intervals end below 1.3%. The comparison with reuse on both master and PR mentioned above remains outstanding, so the benefit of applying reuse to plain master and the exact fraction of the regression it explains are still unmeasured.

core/server/src/dispatch/partition.rs
diegomrsantos: Fixed in [736c4a025](https://github.com/apache/iggy/commit/736c4a02561f8876be4fdc48dba6f4bef06f9bcb). Missing partition read replies now keep HTTP 504 and carry `ShardCommunicationError` (ID `11001`) in the JSON body. Regression tests cover this mapping and HTTP 503 with `TransientNotAccepted`. All 451 server tests and strict Clippy passed on macOS.

core/shard/src/poll.rs
hubcio: nonblocking docs follow-up: could we spell this out in the public poll api docs? a lost reply can come after the auto-commit cursor has advanced, so blindly retrying `Next` can skip messages the caller never saw. worth documenting recovery from client-known explicit offsets for each partition.

core/partitions/src/iggy_partition.rs
hubcio, line 1: nonblocking: with `AutoCommitCtx` gone, `durable_consumer_offsets` and both capacity trackers no longer need the outer `Rc` wrappers. could we own those fields directly and keep sharing inside the reservation tokens/counters? that would save three allocations and indirections per partition.

core/partitions/src/iggy_partition.rs
diegomrsantos: Fixed in [cbdd9c43c](https://github.com/apache/iggy/commit/cbdd9c43c315523120d75d9f8f0dead297cb72f9). `complete_poll` now checks history and partition availability before reconciling reservations. The regression fails with the previous ordering and verifies that a fresh completion still reconciles before automatic commit admission. All 985 tests across the affected crates passed on macOS, with eight existing ignored tests, and strict Clippy passed.

core/shard/src/poll.rs
```diff
@@ -0,0 +1,466 @@
...
+                            self.senders.get(usize::from(self.id)).cloned(),
+                            namespace,
+                            reply,
+                            self.metrics.clone(),
```
hubcio: optional: each disk poll clones and drops 30 `Arc`-backed metric handles, even though completion metrics are only used when enqueueing fails. could we share the needed metrics once during shard setup? the latency impact is unmeasured, so this isn't a blocker. we'd need to preserve the `const` metrics accessor and `Send` bootstrap inputs.

core/shard/src/poll.rs
diegomrsantos: Agreed that reserving completion capacity before disk I/O would bound outstanding work without making workers wait for inbox space.

Could you point me to the existing budget that caps outstanding disk polls? In the current path, `on_partition_read` builds the snapshot and spawns `read_poll` without acquiring a permit. `PARTITION_READ_TIMEOUT` bounds the requester’s wait but leaves the disk read running. The request inbox also frees its slot when the owner dequeues the request.

For a completion lane of capacity `N`, we need:

```text
running disk polls + completions awaiting the owner <= N
```

Releasing a read permit when I/O finishes would let new reads start while earlier results still occupy the completion lane. Capacity therefore needs to remain accounted for until the completion is dequeued or discarded, even after requester timeout. That can be a permit retained by the operation or a reservation transferred into queue occupancy. Unrelated frames must not consume that capacity.

If an existing budget covers this lifetime, we can reuse it. Otherwise, reserving a completion slot before dispatch establishes the guarantee directly: reject before disk I/O when no slot is available, while preserving the existing handling for missing and disconnected routes.

diegomrsantos: Implemented in [49e2e15d4](https://github.com/apache/iggy/commit/49e2e15d4efce576f0ff0bd867dc546cf75e7e64). Disk polls reserve dedicated completion capacity before I/O and retain it until dequeue or discard, even after requester timeout. Exhaustion rejects before reading; completed reads never wait for inbox space.

`sharding.poll_completion_capacity` is independent of `inbox_capacity`. The pump services completions alongside ordinary work, and missing or disconnected routes still reject.

core/shard/src/poll.rs
diegomrsantos: @hubcio Added in [3b9619ea2](https://github.com/apache/iggy/commit/3b9619ea2346b1909c7d24a55ad345b99f429a64). The public poll API now explains lost replies and recovery from checkpoints for each partition, with links from `Next`, HTTP polling, and the consumer docs. It covers staying with explicit offsets until caught up, retention limits, and why a successful reply does not acknowledge a durable offset commit.

core/partitions/src/iggy_partition.rs
diegomrsantos: Done in [3b9619ea2](https://github.com/apache/iggy/commit/3b9619ea2346b1909c7d24a55ad345b99f429a64). The partition now owns the durable offset table and both capacity trackers directly, removing three allocations per partition. Reservation tokens and counters retain their existing sharing.

core/shard/src/poll.rs
diegomrsantos: Done in [3b9619ea2](https://github.com/apache/iggy/commit/3b9619ea2346b1909c7d24a55ad345b99f429a64). The completion lane shares the frame drop family and cache through its existing state, removing the full `ShardMetrics` clone from each disk read dispatch. The `const` accessor and `Send` bootstrap inputs are preserved. The regression verifies lazy metric creation and that a late completion updates the registered counter after lane teardown. Latency remains unmeasured.

core/server_common/src/poll.rs
diegomrsantos: Implemented in [98a6e0c00](https://github.com/apache/iggy/commit/98a6e0c00accac2e185394f33044ba93e065416f). `PollHistoryId` now holds a copied `u64`, assigned by a checked `AtomicU64` counter shared by the process. Exhaustion cannot wrap or reuse an ID, including across partition recreation. Individual polls only copy the ID; they do not access the counter.

core/server/src/dispatch/partition.rs
spetz: This arm catches every `None`, so `reply_was_none = other.is_none()` in the fallback arm below it can never be true. The fallback now only fires on an unexpected `Some(_)` variant, realistically `PartitionReadReply::NotFound`. An operator reading `reply_was_none=false` on a "partition read failed; replying empt

core/partitions/src/iggy_partitions.rs
```diff
@@ -495,6 +488,46 @@ where
...
+    /// Drive an accepted poll's replication on the owning pump after its reply.
+    /// Use the continuation returned by [`Self::complete_poll`] for this namespace.
+    /// This call may suspend with a partition borrow, so it must remain on the
+    /// owning pump. A missing or tombstoned namespace drops the continuation.
+    pub async fn replicate_poll_completion(
+        &self,
+        namespace: &IggyNamespace,
+        replication: PollReplication,
+    ) {
+        if let Some(partition) = self.get_mut_by_ns(namespace) {
```
spetz: A namespace-lookup miss silently drops the continuation, and the doc above sanctions it. By this point `admit_poll_auto_commit` has already called `request.project(consensus)`, consuming an op number, and pushed the prepare into `prepare_queue`. Only `on_replicate` journals and broadcasts it, so a dropped continuation leaves an unjournaled prepare holding that op and stalls the partition's commit chain there. The sole caller in `core/shard/src/poll.rs` has no suspension point between `complete_poll` and this call, since `reply.try_send` is synchronous, so nothing can tombstone the namespace in that window and it is unreachable today. The concern is that the split between op assignment and journaling is new in this diff and the miss is documented as acceptable rather than asserted against.

core/shard/src/metrics.rs
```diff
@@ -175,6 +177,7 @@ const VARIANTS: [&str; VARIANT_COUNT] = [
     frame_drop_variant::REPLICA_HANDSHAKE_ACK,
     frame_drop_variant::PARTITION_AUTO_COMMIT,
```
spetz: This entry has no producer left. Deleting `submit_auto_commit_offset` and the `LifecycleFrame::AutoCommitSubmit` arm removed the only `record_frame_drop` call for it, which was at `core/shard/src/lib.rs:2129` on the base. A repo-wide grep on this head finds only the constant at line 115 and this array slot. `frame_drops_total{variant="partition_auto_commit"}` goes permanently silent after deploy while the code still reads as a live series, so any dashboard or alert keyed on that label breaks quietly instead of being explicitly retired.

spetz (changes requested):
Some final comments before it's ready to merge, could you also please confirm that the initial reports still hold after all the improvements?

core/server/src/dispatch/partition.rs
diegomrsantos: Changed the fallback to `Some(other)` and replaced `reply_was_none` with `reply = ?other`. The warning now identifies the unexpected reply variant.

Fixed in [c7e6fb165](https://github.com/apache/iggy/commit/c7e6fb1655bd48c18c2fb567917f88ec1bdf7fcc).

core/partitions/src/iggy_partitions.rs
diegomrsantos: Replaced the optional lookup with an invariant assertion and documented that the continuation must run on the same owner pump immediately after the synchronous reply attempt. The new regression checks a tombstoned namespace after acceptance has queued an unjournaled prepare, and was verified to fail with the old silent drop and pass with the assertion.

Fixed in [c7e6fb165](https://github.com/apache/iggy/commit/c7e6fb1655bd48c18c2fb567917f88ec1bdf7fcc).

core/shard/src/metrics.rs
diegomrsantos: Removed `PARTITION_AUTO_COMMIT` and its table entry, retiring `frame_drops_total{variant="partition_auto_commit"}` with the deleted submit path. `frame_drops_total{variant="partition_poll_completion"}` measures failures to reserve or deliver poll completions, while `partition_consumer_offsets_denied_total` measures consumer offset capacity denials. Those are different events, so the retired series has no direct replacement.

Fixed in [c7e6fb165](https://github.com/apache/iggy/commit/c7e6fb1655bd48c18c2fb567917f88ec1bdf7fcc).

spetz (approved):
LGTM, thank you for all the work!

hubcio (approved):

PR #4123, opened 2026-09-10
feat(connectors): add OpenDAL sink

### LINKED ISSUE #4071: Implement OpenDAL connector
### Description

This is a spinoff from the discussion here: https://github.com/apache/iggy/issues/2956

I'd like to investigate this one to come up with a sensible configuration. It would be a nice follow up to: https://github.com/apache/iggy/issues/4070

### Affected area / component

_No response_

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Contribution

### Good first issue

George-Miao:

Closes #4071

## Rationale

OpenDAL provides an uniformed interface for writing to various services, which is useful for iggy to support multiple new sink types with a very small amount of change & maintainance effort.

## What changed?

Added opendal-sink crate that implements a sink adapter for OpenDAL and integration test. I intentionally didn't touch the other part, e.g., CI. Those can be done in another follow-up PR.

## Local Execution

- Passed 
- Pre-commit hooks ran 

## AI Usage

Uses codex & omp for writing tests, reviewing the changes & write some part of the doc.


core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+                "OpenDAL service cannot be empty".to_string(),
+            ));
+        }
+        if self.config.path_template.is_empty() {
+            return Err(Error::InvalidConfigValue(
+                "OpenDAL path_template cannot be empty".to_string(),
+            ));
+        }
+
+        opendal::init_default_registry();
```
ryankert01, line 178: `install_default` calles `init_default_registry()` then install transport, so s3 would not fail. Maybe consider to add a test against minio container.
```suggestion
        opendal::install_default();
```

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+        info!(
+            "Opened {CONNECTOR_NAME} connector ID: {}, service: {service}, root: {}",
+            self.id,
+            operator.info().root()
+        );
+        self.operator = Some(operator);
+        Ok(())
+    }
+
+    async fn consume(
```
mattp5657, line 207: Should these be batched instead of one `operator.write()` per message?
`s3_sink` batches and flushes a combined object per batch, this sink
issues a separate write per message, so at the default `batch_length =
100` that's 100 S3 PUTs instead of 1. Worth aligning with `s3_sink`'s
approach for cost/throughput at scale?

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+
+#[derive(Clone, Deserialize)]
+pub struct OpenDalSinkConfig {
+    pub service: String,
+    #[serde(default)]
+    pub path_prefix: Option<String>,
+    #[serde(default = "default_path_template")]
+    pub path_template: String,
+    #[serde(default)]
+    pub options: BTreeMap<String, String>,
```
mattp5657, line 55: Credentials (S3 access keys, etc.) flow
through `options: BTreeMap<String, String>` instead of `SecretString`,
diverging from this codebase's own stated convention for
credential-bearing config. A hand-rolled `Debug` impl redacts values in
one specific logging path but gives no type-level guarantee anywhere
else.

core/connectors/sinks/opendal_sink/src/path.rs
```diff
@@ -0,0 +1,123 @@
...
+            }
+        })
+        .collect()
+}
+
+fn timestamp_to_datetime(micros: u64) -> Result<DateTime<Utc>, Error> {
+    let seconds = (micros / 1_000_000) as i64;
+    let nanoseconds = ((micros % 1_000_000) * 1_000) as u32;
+    DateTime::<Utc>::from_timestamp(seconds, nanoseconds).ok_or_else(|| {
+        Error::CannotStoreData(format!(
```
mattp5657, line 93: Should an out-of-range timestamp return `Error::InvalidRecordValue`
instead of `Error::CannotStoreData`?

mattp5657 (commented):
2 things:

1. This connector is missing from the release/version-bump tooling. Absent from scripts/bump-version.sh, the connector_plugins default in _build_rust_artifacts.yml, and the connector list in edge-release.yml. It'll pass CI but won't ship in edge or tagged release artifacts, and a future version bump will silently skip it. Not unique to this PR (a handful of existing sinks are missing from these same lists too), but worth fixing here or in an immediate follow-up rather than adding a fourth thing to that pile.

2. I think we could use more unit and integration test coverage, happy and unhappy paths for every function. Untested today: path.rs's render_template and sanitize_path_segment beyond one edge case, lib.rs's consume()/write_message() unhappy path, the retry policy past one attempt, and S3/GCS/Azure end-to-end (fs is the only backend any test exercises)

core/connectors/sinks/opendal_sink/src/lib.rs
George-Miao: Done

core/connectors/sinks/opendal_sink/src/path.rs
```diff
@@ -0,0 +1,150 @@
...
+    Ok(template
+        .replace("{stream}", &sanitize_path_segment(context.stream))
+        .replace("{topic}", &sanitize_path_segment(context.topic))
+        .replace("{partition}", &context.partition_id.to_string())
+        .replace("{date}", &date)
+        .replace("{hour}", &hour)
+        .replace("{timestamp}", &timestamp_millis))
+}
+
+fn sanitize_path_segment(segment: &str) -> String {
```
hubcio, line 73: warning: `a b` and `a_b` both sanitize to `a_b`, so the later write hides the earlier topic's object, and `..` passes too and always fails. percent-encode the segment or hash the raw name.

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+            .collect::<Result<Vec<_>, _>>()?;
+        let data = formatter::finalize_buffer(entries.iter().map(Vec::as_slice), output_format);
+        let buffer = Buffer::from(data);
+        let retry_context = format!("{CONNECTOR_NAME} connector ID {} batch write", self.id);
+
+        retry_async(
+            self.retry_policy,
+            &retry_context,
+            opendal::Error::is_temporary,
+            || operator.write(&path, buffer.clone()),
```
hubcio, line 184: warning: with the shipped `fs` config opendal truncates the final key and writes in place, so a reader can see a partial object and a crash leaves one. set `atomic_write_dir`, and document that fs writes are not atomic.

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+        };
+        let path = object_path(
+            &self.path_prefix,
+            &self.config.path_template,
+            &context,
+            first_message.offset,
+            last_offset,
+            output_format,
+        )?;
+        let entries = messages
```
hubcio, line 163: warning: collects a `Vec<u8>` per message, then `finalize_buffer` copies all of them into a second buffer. format straight into one buffer, and move the `JsonArray` bracket logic there first.

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+            },
+            verbose,
+            output_format: None,
+            operator: None,
+            messages_processed: AtomicU64::new(0),
+            write_errors: AtomicU64::new(0),
+        }
+    }
+
+    async fn write_batch(
```
hubcio, line 135: warning: one object per poll batch means nothing bounds object size in bytes. `s3_sink` rotates at 8 MiB, but the connector guidance says not to buffer across `consume` - which rule wins here?

core/connectors/sinks/opendal_sink/src/formatter.rs
```diff
@@ -0,0 +1,215 @@
...
+            Some(messages_metadata.partition_id)
+        } else {
+            None
+        },
+        headers: if include_headers {
+            message.headers.as_ref().map(serialize_headers)
+        } else {
+            None
+        },
+        payload: payload_to_json_value(&message.payload),
```
hubcio, line 113: warning: rebuilds the payload tree before serializing, which allocates a node per element and a `String` per key. serialize in place with `serde_json::to_writer` and a wrapper that maps the six `Payload` variants.

core/integration/tests/connectors/fixtures/opendal.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub async fn wait_for_object(
+        &self,
+        offset_start: u64,
+        offset_end: u64,
+    ) -> Result<Vec<u8>, TestBinaryError> {
+        let path = self.object_path(offset_start, offset_end);
+
+        for _ in 0..POLL_ATTEMPTS {
+            match tokio::fs::read(&path).await {
+                Ok(payload) => return Ok(payload),
```
hubcio, line 51: warning: returns on the first successful read, but the sink writes in place, so a poll that lands mid-write reads an empty object and the assertion fires instead of retrying. poll on the payload comparison.

core/connectors/sinks/opendal_sink/README.md
```diff
@@ -0,0 +1,161 @@
...
+```bash
+cargo build --release -p iggy_connector_opendal_sink \
+    --no-default-features \
+    --features opendal/services-compfs
+```
+
+When using the sink as a dependency, add `opendal` as a direct dependency and enable additional features:
+
+```toml
+iggy_connector_opendal_sink = { version = "0.5.0-edge.4", default-features = false }
```
hubcio, line 82: warning: this dependency cannot resolve - the crate sets `publish = false`, so nothing is on a registry. replace the snippet with a `cargo build` command.

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+            ))
+        })?;
+
+        if !operator.info().capability().write {
+            return Err(Error::InvalidConfigValue(format!(
+                "OpenDAL service '{service}' does not support writes"
+            )));
+        }
+
+        operator.check().await.map_err(|error| {
```
hubcio, line 233: warning: `operator.check()` sends a list request, so a write-only credential fails here and the connector never starts. is listing a documented requirement, or does this need a write probe?

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+                    .fetch_add(message_count, Ordering::Relaxed);
+                debug!(
+                    "{CONNECTOR_NAME} connector ID: {} uploaded one object with {} messages",
+                    self.id, message_count
+                );
+                Ok(())
+            }
+            Err(write_error) => {
+                self.write_errors
+                    .fetch_add(message_count, Ordering::Relaxed);
```
hubcio, line 304: nit: `write_errors` counts messages, but `close()` logs it under an errors label. rename the label - counting batches instead breaks the test at line 609.

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+    #[serde(default)]
+    pub options: BTreeMap<String, SecretString>,
+    #[serde(default = "default_output_format")]
+    pub output_format: String,
+    #[serde(default = "default_true")]
+    pub include_metadata: bool,
+    #[serde(default)]
+    pub include_headers: bool,
+    #[serde(default)]
+    pub max_attempts: Option<u32>,
```
hubcio, line 66: nit: `s3_sink` accepts both `max_attempts` and `max_retries`, and five other sinks use `max_retries`. add `#[serde(alias = "max_retries")]` so a copied config is not silently ignored.

core/connectors/sinks/opendal_sink/src/lib.rs
```diff
@@ -0,0 +1,759 @@
...
+    JsonArray,
+    Raw,
+}
+
+impl TryFrom<&str> for OutputFormat {
+    type Error = Error;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        match value.to_lowercase().as_str() {
+            "json_lines" | "jsonl" | "jsonlines" => Ok(Self::JsonLines),
```
hubcio, line 351: nit: `jsonl` and `jsonlines` are accepted here but the README lists only three names. add them to the table, since `s3_sink` documents the same pair.

core/connectors/sinks/opendal_sink/README.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+### Plugin options
+
+| Option                  | Type     | Default                          | Description                                   |
+| ----------------------- | -------- | -------------------------------- | --------------------------------------------- |
+| `service`               | String   | **required**                     | OpenDAL service name                          |
+| `path_prefix`           | String   | empty                            | Prefix before the rendered object path        |
+| `path_template`         | String   | `{stream}/{topic}/{date}/{hour}` | Object directory template                     |
+| `options`               | Map      | empty                            | Secret OpenDAL service options                |
```
hubcio, line 48: nit: values must be quoted strings in TOML, since `SecretString` deserializes through `String`, so a bare number or bool fails at load. say so in the row.

core/connectors/sinks/opendal_sink/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[package]
+name = "iggy_connector_opendal_sink"
+version = "0.5.0-edge.4"
```
hubcio, line 20: nit: `0.5.0-edge.4` is the only prerelease among the connector plugins, and `bump-version.sh` does not list this crate, so a release bump skips it. use `0.5.0` and add the crate to `CONNECTOR_SINK_COMPONENTS`.

hubcio (changes requested):
the plugin does not reach release artifacts yet: `_build_rust_artifacts.yml:49` and `edge-release.yml:108` do not list it, `scripts/bump-version.sh:90` omits the crate, and the sinks table in `core/connectors/sinks/README.md` has no row. you noted that the CI part is a follow-up PR - can you confirm it covers all four?

Issue #4128, opened 2026-09-11
Purge can record completion after consumer offset directory sync fails


diegomrsantos:
### Bug description

Purge can durably record its applied generation even when synchronizing a consumer offset directory fails. If a subsequent power loss loses the offset file deletions but preserves the purge generation, restart can restore the old consumer offsets without retrying the purge. Consumers can then skip messages written after the purge.

This affects both individual consumers and consumer groups. The failure path remains after #4092, including when consumer offset durability is `persisted`.

This report is based on code inspection. I have not injected the synchronization failure or reproduced the power loss sequence.

### Affected component and versions

Iggy server. Inspected `master` at `e0027506f2a0379ee3bb5508427ff570e3a9a435` and #4119 at `d89b81f0297cbcb6577a5a69cd44857c94b0cdf0`. The links below refer to the inspected `master` commit.

### Evidence

1. After deleting consumer offset files, `purge()` synchronizes both offset directories. A failure [only logs a warning](https://github.com/apache/iggy/blob/e0027506f2a0379ee3bb5508427ff570e3a9a435/core/partitions/src/iggy_partition.rs#L7548-L7574). It does not return an error or preserve a pending cleanup obligation.
2. Execution can then [persist `purge.gen` and advance `applied_purge_generation`](https://github.com/apache/iggy/blob/e0027506f2a0379ee3bb5508427ff570e3a9a435/core/partitions/src/iggy_partition.rs#L7627-L7654). Recording that generation synchronizes its own parent directory, which does not establish durability for deletions inside the separate consumer offset directories.
3. On restart, recovery [loads consumer and group offset files and seeds their values into partition state](https://github.com/apache/iggy/blob/e0027506f2a0379ee3bb5508427ff570e3a9a435/core/server/src/partition_helpers.rs#L181-L257). It can clamp a value to the recovered offset range, but does not discard it based on the purge generation. A clamped offset is still a consumer entry; "consumed through 0" is different from having no committed offset.
4. The reconciler [only requests another purge when the committed generation exceeds the applied generation](https://github.com/apache/iggy/blob/e0027506f2a0379ee3bb5508427ff570e3a9a435/core/server/src/partition_reconciler.rs#L1430-L1453). A surviving `purge.gen` therefore prevents the cleanup from being retried for that generation.

[...]

Issue #3793, opened 2026-08-01
server: partition journal repair resurrects purged data after restart


krishvishal:
## Summary

  `PurgeTopic` mutates partition state out of band (metadata op → reconciler → pump) and never
  touches the partition journal. Superseded ops stay repairable. After a node restart, partition
  journal repair replays them and undoes the purge.

  The partition plane has no checkpoint, so there is no repair floor a purge can advance.

  ## Repro

  ```
  cargo nextest run --features vsr --retries 0 \
    -E 'test(/should_purge_topic_and_clear_consumer_offsets.*restart_on/)'
  ```

  ~30% fail. Control: `restart_off` 24/24 pass. Three sites fire depending on interleaving:

  | site | assertion | observed |
  |---|---|---|
  | `purge_delete_scenario.rs:992` | consumer offset cleared | offset returns |
  | `purge_delete_scenario.rs:1006` | group offset cleared | offset returns |
  | `purge_delete_scenario.rs:1331` | single segment at offset 0 | `[0, 5, 10]`, `[0, 5, 10, 15]` |

  ## Evidence

  Restarted node:

  ```
  .412599  purge-partition reset partition to empty  shard=1 generation=1
  .429027  repair prepare received  shard=1 op=1
     ...   ops 2..26
  .519273  partition journal repair complete  through_op=26  (commit_min_before=0)
  ```

  Peer: `.428320  served partition repair range  from_op=1 to_op=26 served_through=26`

jiengup:
Any instructions for fixing this issue?
Assign it to me if you're not going to work on it in the short term.

jphjsoares:
Hi @krishvishal interested on looking into this, could you assign it to me?

jphjsoares:
Hey @krishvishal, I was able to reproduce it, so I don't think it's fixed yet (even though I saw some promising PRs getting merged the last month).

Repro: `cargo nextest run -p integration --retries 0 -E "test(/should_purge_topic_and_clear_consumer_offsets.*restart_on/)"` failed 1 in 79 runs for me on `d693eb93d`. Two setup notes: `--features vsr` no longer exists (I ran the bare command), and the test is currently pinned to `cluster_nodes = 1`, which can't hit the repair path — I ran it with a local override back to 3 nodes.

The failure looks like what's reported here: the surviving peer served the full pre-purge window (`from_op=1`, served through 27, no eviction notice), the restarted node completed the repair, and the partition came back with all segments plus both consumer offsets at their old values.

[...]

PR #4130, opened 2026-09-11
test(simulator): reproduce purge offset sync failure after power loss

diegomrsantos:
Purge can return success and record its applied generation after an offset directory sync fails. Simulated power loss then restores bookmark `2` while generation `5` survives, so `Next` returns only messages `3` and `4` from fresh history `0–4`.

Refs #4128. This draft uses the shared production purge completion and recovery paths introduced by #4240, on `master` at `7bd69670f9ad9946f6490359443ec25bf7e2a089`. It contains no production fix.

Eight independent regression tests cover consumer and group directory failures under both `Replicated` and `Persisted` offset policies. Four verify that a reached sync failure follows successful unlinking, returns an error, and leaves generation `4` in memory and storage. Four persist fresh history, simulate power loss, reload the marker and bookmarks, and complete real `Next` polls for both consumers before asserting. Recovery must either retain the earlier generation after a failed purge or return all fresh messages without old bookmarks after a successful purge. The existing recovery and successful purge controls remain in place.

Run the focused reproduction with:

```sh
cargo test -p simulator --lib storage::purge -- --nocapture
```

Local validation on macOS:

- The focused run produced two passing controls and eight expected failures.
- The full simulator library suite produced 150 passes, the same eight expected failures, and four ignored tests.
- A temporary experiment that propagated the directory sync error made all ten focused tests pass. That change was removed.
- Workspace formatting, dependency ordering, Clippy with all features and targets, and TOML formatting passed.

The eight regressions intentionally remain red until the bug is fixed. The harness starts after message reset and persists fresh history through a fixture. It does not exercise the full purge retry path or prove preservation of acknowledged fresh writes during retries.

### DIFF
--- a/core/partitions/src/iggy_partition.rs
+++ b/core/partitions/src/iggy_partition.rs
@@ -235,6 +235,8 @@ where
     #[cfg(test)]
     consumer_offset_dir_sync_fault: Cell<Option<usize>>,
     #[cfg(test)]
+    consumer_offset_dir_sync_fault_count: Cell<usize>,
+    #[cfg(test)]
     offset_dir_sync_count: Cell<usize>,
     /// Highest `PurgeTopic` generation this replica has locally applied (reset
     /// the partition to empty). The reconciler compares the committed metadata
@@ -603,6 +605,8 @@ where
             consumer_offset_dirs_touched: [Cell::new(None), Cell::new(None)],
             #[cfg(test)]
             consumer_offset_dir_sync_fault: Cell::new(None),
+            #[cfg(test)]
+            consumer_offset_dir_sync_fault_count: Cell::new(0),
             #[cfg(test)]
             offset_dir_sync_count: Cell::new(0),
             applied_purge_generation: 0,
@@ -6095,13 +6099,8 @@ where
             if !kinds[index] || !self.consumer_offset_dirs_dirty[index].get() {
                 continue;
             }
-            #[cfg(test)]
-            if self.consumer_offset_dir_sync_fault.get() == Some(index) {
-                failed[index] = true;
-                continue;
-            }
             if let Some(dir) = dir {
-                match crate::state_transfer::fsync_dir(dir).await {
+                match self.sync_consumer_offset_directory(index, dir).await {
                     Ok(()) => {}
                     Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
                         // The directory disappeared after the final unlink.
@@ -6141,6 +6140,24 @@ where
             self.consumer_offset_dirs_dirty[index].set(false);
         }
         failed
+    }
+
+    async fn sync_consumer_offset_directory(
+        &self,
+        kind_index: usize,
+        directory: &str,
+    ) -> std::io::Result<()> {
+        #[cfg(not(test))]
+        let _ = kind_index;
+        #[cfg(test)]
+        if self.consumer_offset_dir_sync_fault.get() == Some(kind_index) {
+            self.consumer_offset_dir_sync_fault_count
+                .set(self.consumer_offset_dir_sync_fault_count.get() + 1);
+            return Err(std::io::Error::other(
+                "injected consumer offset directory sync failure",
+            ));
+        }
+        crate::state_transfer::fsync_dir(directory).await
     }
 
     fn mark_consumer_offset_dir_dirty(&self, kind: ConsumerKind) {
@@ -7553,13 +7570,17 @@ where
         // failure, sharper than the partition-dir fsync above: the generation
         // write below still runs, so a crash would resurrect these files with
         // no boot re-purge left to clear them.
-        for dir in self
-            .consumer_offsets_path
-            .clone()
-            .into_iter()
-            .chain(self.consumer_group_offsets_path.clone())
+        for (index, directory) in [
+            self.consumer_offsets_path.as_deref(),
+            self.consumer_group_offsets_path.as_deref(),
+        ]
+        .into_iter()
+        .enumerate()
         {
-            if let Err(error) = crate::state_transfer::fsync_dir(&dir).await {
+            let Some(dir) = directory else {
+                continue;
+            };
+            if let Err(error) = self.sync_consumer_offset_directory(index, dir).await {
                 warn!(
                     target: "iggy.partitions.diag",
                     plane = "partitions",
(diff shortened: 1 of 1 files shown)


diegomrsantos:
@hubcio @spetz, working on [#4130](https://github.com/apache/iggy/pull/4130) raised a broader question about the guarantees we want from purge.

Today, purge clears messages, restarts their offsets at zero, and separately removes consumer bookmarks. If a bookmark deletion is not durable, an old bookmark at `2` can return after power loss and silently skip new messages `0–2`. It still looks valid because those offsets now exist in a different history. Retrying the entire purge can also delete fresh messages accepted after the original reset. That is why this PR separates message reset from bookmark cleanup and records their progress independently.

**The fundamental requirement is that messages, bookmarks, and recovery metadata agree about which history they describe.** Physical file deletion can take several steps, provided recovery restores a coherent logical state. A persisted cutoff is only meaningful if the sequence it describes is also recovered consistently.

There is a related defect in the current PR. A local partition regression with three replicas and both durability policies set to `Replicated` restores the old operation cutoff, but reopens no durable operation journal. When operations restart at `1`, fresh committed messages at or below the old cutoff disappear from polls. These are internal operation numbers, separate from public message offsets. Existing tests miss this combination because they either restore the previous operation counter or exercise restart with a singleton.

I looked at how other systems handle this.

- **Kafka** advances the earliest available offset while preserving message numbering. After deleting messages through `100`, new messages continue from `101`. A bookmark at `80` is recognizably outside the available range, and the consumer can resume at the beginning, jump to the end, or report an error according to its configuration. Consumer offsets are managed separately, so message deletion does not require atomically rewriting every bookmark. [Deletion implementation](https://github.com/apache/kafka/blob/4.3.1/core/src/main/scala/kafka/cluster/Partition.scala#L1514-L1539), [consumer behavior](https://kafka.apache.org/43/configuration/consumer-configs/#auto.offset.reset)
- **NATS JetStream** preserves numbering even when a stream becomes empty and records the exact boundary of a clustered purge. Recovery replays “purge through 100,” so a later message `101` survives. It adjusts consumer state separately. File cleanup uses prepared replacement storage and recovery logic that can finish an interrupted switch, with old files removed afterward. [Replay protection](https://github.com/nats-io/nats-server/blob/v2.15.0/server/jetstream_cluster.go#L5397-L5443), [storage recovery](https://github.com/nats-io/nats-server/blob/v2.15.0/server/filestore.go#L10619-L10804)
- **Redpanda** replicates a command advancing the partition's starting offset, requests a disk flush, and waits for the command to be applied. The committed boundary determines which messages remain available, while background cleanup reclaims obsolete storage. This gives recovery an authoritative deletion decision without requiring all files to disappear together. [Implementation](https://github.com/redpanda-data/redpanda/blob/v26.2.2/src/v/cluster/log_eviction_stm.cc#L240-L304)
- **Pulsar** identifies positions by ledger and entry, like a volume number and page number. Entry zero in a new ledger cannot be mistaken for entry zero in the old one. Truncation advances subscription bookmarks before trimming unused ledgers. This is not one transaction across all bookmarks, but it preserves the identity of their positions. [Position model](https://bookkeeper.apache.org/docs/getting-started/concepts/), [truncate implementation](https://github.com/apache/pulsar/blob/v4.0.13/managed-ledger/src/main/java/org/apache/bookkeeper/mledger/impl/ManagedLedgerImpl.java#L4772)
- **RabbitMQ quorum queues** order purge with message operations in their replicated history. Replaying “receive A, purge, receive B” leaves B intact because purge retains its place in that history. Physical compaction happens separately. Its semantics differ from Iggy's, since consumer state and messages already delivered but awaiting acknowledgment survive purge. [Purge submission](https://github.com/rabbitmq/rabbitmq-server/blob/v4.3.6/deps/rabbit/src/rabbit_fifo_client.erl#L527), [queue state transition](https://github.com/rabbitmq/rabbitmq-server/blob/v4.3.6/deps/rabbit/src/rabbit_fifo.erl#L501)

These systems have different atomicity and durability guarantees. The common ideas worth considering are stable identities, precise replay boundaries, and separating the committed logical change from storage reclamation.

I'd appreciate your input on three points.

1. Is restarting public message offsets at zero after purge a required compatibility guarantee, or could we preserve numbering and advance the available boundary?
2. If we keep the reset semantics, would a committed generation shared by message and bookmark validity give us a clearer recovery contract?
3. What should a successful purge guarantee after restart or power loss under each durability policy, and is that guarantee per partition?

My preference is to fix the concrete recovery defect in this PR and add the missing restart regression, then evaluate the broader design separately. The tests should also cover failed directory synchronization, retries after fresh writes, and another crash during recovery.

hubcio:
thanks, i actually had reply yesterday, i just forgot to post :(

I agree with your plan: fix that defect here and add the restart regression, plus failed dir sync, a retry after fresh writes, and a crash during recovery. three design questions fit #4145 better, i'll answer them there now.

diegomrsantos:
Thanks @hubcio. Agreed on keeping the fix and regression tests here and discussing the broader design in #4145.

I also created a small [Quint](https://quint.sh/docs/quint) model of purge and recovery to investigate the edge cases. Quint is a language for writing executable descriptions of a system’s behavior. We describe what survives a crash, which actions can happen, and which rules must always hold. Here, those rules include keeping purged messages and bookmarks from returning, while ensuring an old purge cutoff cannot hide fresh messages.

Using TLC, a model checker supported by Quint, we can explore every reachable state of a small finite configuration. When a rule fails, it produces the sequence of events that caused the failure. That helps us examine combinations of crashes, retries, and replica recovery that are easy to overlook when writing individual tests.

The model reproduces the cutoff mismatch we discussed and demonstrates problems with some apparent fixes. For example, clearing the cutoff on every restart can let a recovering replica replay purged messages and bookmarks from another replica that still retains the old journal.

I think this could be a useful addition alongside the regression tests. It makes the recovery assumptions explicit, helps compare proposed fixes, and gives us concrete failure sequences to reproduce against the Rust implementation. Its conclusions depend on the model’s scope and assumptions, including how consensus selects a history, so it does not establish that the implementation is correct.

Would you be open to keeping a small model alongside the tests as an executable explanation of the recovery rules?

There is also a possible next step with [Quint Connect](https://github.com/quint-co/quint-connect). It can replay scenarios generated from a Quint model against real Rust code and compare the relevant state after each step. We would need a test driver that maps model actions to Iggy operations, controls the modeled failures, and translates Iggy’s state into the form the model expects. Those tests could run through `cargo test` in CI, helping catch cases where the implementation diverges from the recovery rules. That integration would still need to be built.

hubcio:
i like quint approach. AFAIR @krishvishal mentioned it to me couple of weeks ago, didnt have time to dig into it.

we can keep it, but in its own PR so the fix here doesn't wait on tooling, and with `check.py --suite=quick` in CI, otherwise it quietly drifts from the code.

the model also caught something not in your comment: `journal.clear_all()` never resets `poll_floor` (only `clear_poll_index` sets it), so after a state transfer install the old floor stays and can hide fresh ops. it's on master too. fix here or separate issue?

quint connect: interesting. do you have previous experience with it?

diegomrsantos:
@hubcio, agreed. I’ll move the model and supporting files into a separate draft PR.

One clarification about CI. `check.py --suite=quick` checks only the Quint models, so it can catch model regressions but cannot detect Rust drifting from them. A Rust change could leave those checks passing. The current model compares failure cases and candidate recovery rules, rather than defining an agreed implementation contract. CI would still be useful, with review and Rust regressions needed to maintain that connection.

Good catch on `poll_floor`. I’d first reproduce the state transfer case in a Rust regression before deciding where the fix belongs.

Yes, I tried Quint Connect a few months ago in a personal project and it seemed promising. I checked it again, and public activity looks fairly quiet, with the [latest commit in May](https://github.com/quint-co/quint-connect/commit/4f018f54fc7dd4cef341d10111427bab59d3b307). I’d treat an integration as an experiment for now.

diegomrsantos:
@hubcio, I’ve pushed fixes for the replicated restart defect and the `poll_floor` issue, with regression tests. The local suites pass, but I kept finding additional edge cases while working through the changes, and I’m concerned about how much recovery machinery this fix now requires.

The PR adds almost 3,000 lines, roughly 80% of which are tests and fixtures. We can simplify some repetition, but the larger concern is the production design. `purge.reset` remembers both unfinished cleanup and an operation cutoff that remains after cleanup completes. Keeping that cutoff consistent when operation history is lost or replaced requires additional fencing, history selection, rebasing and rollback handling.

This overlaps with the concern @krishvishal raised in #3793 about persisting locally chosen cutoffs, and his recommendation to order purge through the partition’s own consensus history. Your preference in #4145 for preserving message numbering also seems promising. Old bookmarks would remain distinguishable from positions in new messages, removing much of the cleanup problem. Those approaches address complementary concerns, while #4266 covers the broader conditions for safely resuming service.

The repeated discoveries are why I’ve been raising verification tools. The regressions give evidence for the cases they exercise, but I’m still uncomfortable treating them as sufficient confidence in the overall recovery behavior. Adding another testing framework would not resolve the underlying contract questions.

I know we agreed to finish the concrete fix here and discuss the broader design separately. Given how much the implementation has grown, I’d like to revisit that separation before introducing more durable state that a subsequent design may replace.

My preference now is to keep this PR as a draft, stop expanding it, and coordinate the design decision with the existing proposal in #3793. If we agree to pursue that direction, I’d be comfortable closing this PR as superseded and carrying its useful regressions into the replacement. The independent journal fixes could also be extracted.

Would you be comfortable taking that approach, or is there an urgent need for a fix that preserves the current behavior before the design work can land?

PR #4132, opened 2026-09-11
test(simulator): cover durability failures no existing test can reach

numinnex:
Adds extra tests scenarios after #4092 PR.

### DIFF
--- a/core/integration/tests/cluster/crash_durability.rs
+++ b/core/integration/tests/cluster/crash_durability.rs
@@ -859,3 +859,40 @@ async fn verify_transferred_quorum(
         sleep(POLL_INTERVAL).await;
     }
 }
+
+/// Every persisted case in this file kills with SIGKILL, and the module doc
+/// above states why that cannot reach fsync ordering. This pins the premise:
+/// a completed `write()` lives in the page cache, which the kernel owns, so
+/// process death cannot lose it. Consequently none of the barriers this PR
+/// adds -- `file.sync()`, `sync_data()`, `fsync_dir()` -- changes the outcome
+/// of a single test here, and `persisted` differs from `replicated` only in
+/// surviving writes that were never synced.
+///
+/// Covering the real contract needs a fault that discards unsynced pages:
+/// either the deterministic simulator (drop the persisted-topic assert at
+/// `core/shard/src/lib.rs:4193-4198` and route partition storage through
+/// `DurableStorage`) or `dm-log-writes` / `dm-flakey --drop_writes` under the
+/// data directory.
+#[test]
+#[ignore = "PR #4092 review: SIGKILL cannot lose a completed write, so no test in this file distinguishes a correct barrier order from a deleted one"]
+fn given_a_completed_write_when_the_process_is_sigkilled_then_the_bytes_should_not_survive() {
+    let directory = tempfile::tempdir().unwrap();
+    let path = directory.path().join("unsynced");
+    let status = std::process::Command::new("sh")
+        .arg("-c")
+        .arg(r#"printf 'acknowledged' > "$0"; kill -9 $$"#)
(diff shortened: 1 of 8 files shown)


core/simulator/src/storage.rs
```diff
@@ -205,6 +265,13 @@ impl SimStorage {
 impl DurableStorage for SimStorage {
     type File = SimFile;
 
+    fn writer_identity(&self, path: &Path) -> io::Result<Option<std::path::PathBuf>> {
```
spetz, line 268: Include the simulated process epoch in writer ownership. Cancelling an in-flight writer leaves an interrupted entry in `WRITERS`, and `Crash::Process` preserves the same identity. Reopening therefore still fails with “requires process restart” after the simulated restart. Reproduced locally. Namespace leases by the process epoch or reset their process-local state on crash.

core/simulator/src/storage/tests.rs
```diff
@@ -2517,3 +2518,251 @@ fn prepare_with_payload(op: u64, parent: u128, payload: &[u8]) -> Message<Prepar
...
+        let first = prepare(1, 0);
+        journal.append(first.clone().into_frozen()).await.unwrap();
+        let second = prepare(2, first.header().checksum);
+        journal
+            .append_buffered(second.clone().into_frozen())
+            .await
+            .unwrap();
+        drop(journal);
+
+        // Background writeback preserved the later pages of the buffered record
```
spetz, line 2559: Create the page-level hole this test claims to cover. The first prepare occupies four already-synced blocks, so writing back from block 1 persists the entire second prepare. I verified that every WAL byte survives unchanged. Select a page inside the unsynced record and verify its missing prefix and surviving tail before testing recovery.

core/simulator/src/storage/tests.rs
```diff
@@ -2517,3 +2518,251 @@ fn prepare_with_payload(op: u64, parent: u128, payload: &[u8]) -> Message<Prepar
...
+        writer.write(0, b"committed-index".to_vec()).await.unwrap();
+
+        // The device drops the dirty pages. The writer that issued them is the
+        // only handle that can observe it.
+        storage.fail_writeback(path).unwrap();
+
+        // `checkpoint_files` opens its own descriptor and syncs through that,
+        // so it samples errseq after the failure and sees success over bytes
+        // that are already gone.
+        let checkpoint_handle = storage.open(path, OpenMode::Read).await.unwrap();
```
spetz: Exercise the actual checkpoint path. This test calls only `SimStorage`, whose fresh handle deliberately samples the current error sequence and succeeds. Fixing `checkpoint_files` cannot make this assertion pass. Inject writeback failure through checkpointing and assert that it reports failure and preserves the required WAL history.

core/integration/tests/cluster/crash_durability.rs
```diff
@@ -859,3 +859,40 @@ async fn verify_transferred_quorum(
...
+        .arg(r#"printf 'acknowledged' > "$0"; kill -9 $$"#)
+        .arg(&path)
+        .status()
+        .expect("spawn a writer that dies before any barrier");
+    assert!(
+        !status.success(),
+        "the writer exited normally, so it is not modelling a crash"
+    );
+
+    let survived = std::fs::read(&path).unwrap_or_default();
```
spetz, line 891: Do not require SIGKILL to discard completed writes. This test runs no Apache Iggy code and expects process death to clear the kernel’s cached bytes. Locally, the writer dies while `acknowledged` remains readable. Assert survival to document that limitation, or inject a fault that actually discards unsynced data.

spetz (changes requested):

spetz (approved):

hubcio (approved):

PR #4146, opened 2026-09-11
perf(partitions): cut persisted-mode disk traffic per acknowledgment

spetz:
A persisted topic drove far more disk traffic than the bytes it
ingested, and the excess barely moved when the ingest rate rose.
The cost was per acknowledgment, not per byte.

Publishing the WAL frontier created a temporary inode, wrote one
block, fdatasynced it, renamed it over the frontier and fsynced
the directory, once per batch. Three barriers on three inodes at
three instants, so none could share a filesystem journal commit.
It was also the fourth serialized barrier in a path that needs two.

The frontier is now two slots in one file, published by
overwriting the older slot in place: one aligned block into a
file that never resizes, so no create, no rename, no directory
barrier. The body and WAL barriers overlap and land in one
journal window.

Dropping the publication entirely was tried first and rejected.
Without a durable length, a zeroed acknowledged record and one
never written are indistinguishable, and recovery silently
truncates. Two slots keep that length at no metadata cost: a
torn slot fails its checksum and its partner holds the previous
publication, as the rename did. An unreadable newest slot is the
one case that loses, so recovery then walks past the frontier it
could read and adopts the records already on disk, refusing
rather than truncating when any fails to verify.

The per-flush index fdatasync moves to the seal and checkpoint
barriers WAL reclamation already requires, through the original
writer handle so a writeback error still reaches it, and a failed
sync fences before any history is reclaimed. The group-commit
budget stops counting body bytes the WAL only references.

A cheaper barrier groups fewer prepares, since grouping was only
ever whatever queued during the previous barrier. Latency falls.
Write count does not, until grouping is made explicit.


core/journal/src/partition_journal.rs
```diff
@@ -1018,21 +1265,41 @@ impl<S: DurableStorage> PartitionPrepareJournal<S> {
...
-        file.write(0, state.encode()).await?;
-        file.sync().await?;
-        self.storage
-            .rename(&temporary, &self.directory.join("frontier"))
+        let sequence = self
+            .frontier_sequence
+            .checked_add(1)
+            .ok_or_else(|| invalid("WAL frontier sequence exhausted"))?;
+        self.frontier
+            .write_aligned(frontier_offset(sequence), state.encode(sequence))
```
hubcio, line 1314: critical: after an acknowledged append lands in slot 1, a crash and binary rollback can make the old reader accept stale slot 0 and truncate acknowledged data. version the two-slot format so old readers refuse it, retaining one-slot migration.

hubcio (changes requested):

core/journal/src/partition_journal.rs
```diff
@@ -234,27 +269,121 @@ impl<S: DurableStorage> PartitionPrepareJournal<S> {
...
-            journal.publish(state).await?;
-        }
-        journal.discover_obsolete().await?;
-        loop {
-            let remaining = journal.obsolete.len();
-            journal.cleanup_obsolete().await;
-            if journal.obsolete.is_empty() || journal.obsolete.len() == remaining {
-                break;
-            }
+        if recovered != state {
```
hubcio: warning: when `recovered == state`, a damaged frontier slot survives reopen and can make a later partial, unacknowledged tail prevent recovery. republish whenever `verified < slots`, even when the recovered state is unchanged.

core/journal/src/partition_journal.rs
```diff
@@ -790,9 +933,110 @@ impl<S: DurableStorage> PartitionPrepareJournal<S> {
...
+    /// has to fence and rebuild from its peers rather than open a truncated log.
+    /// When both slots verify the frontier is exact and the tail is discarded as
+    /// it always was.
+    async fn recover_unpublished_tail(
+        &mut self,
+        mut previous: u64,
+        mut checksum: u128,
+    ) -> io::Result<()> {
+        let limit = self.file.length().await?;
+        if limit > self.capacity {
```
hubcio: warning: lowering WAL capacity can make valid history fail to reopen when a frontier slot is damaged. remove both admission-capacity checks from recovery while keeping the file and protocol bounds.

hubcio (changes requested):
warning: in `core/journal/src/partition_journal.rs:1467`, a torn publication during `rewrite()` can make recovery reject an unsynced old-generation tail. sync `self.file` through its original writing handle before publishing the replacement.

hubcio (approved):

mmodzelewski (approved):

PR #4150, opened 2026-09-12
fix(server): preserve write responses and validate CLI input

hubcio:
Forwarding can lose durability headers, accepted CLI input can panic
or parse incorrectly, and benchmarks can expose secrets or join the
wrong consumer group.

Preserve response headers, validate CLI input, redact benchmark
metadata and select the matching group. Avoid the telemetry reload
deadlock and align protocol, configuration and command guidance with
the implemented behavior.


spetz (dismissed):

README.md
```diff
@@ -89,10 +89,10 @@ The name is an abbreviation for the Italian Greyhound - small yet extremely fast
   are reserved for future disk/network compression support; use message headers
   for manual compression today (see `examples/rust/src/message-headers/message-compression`).
 - Optional **data backups and archiving** to disk or **S3** compatible cloud storage (e.g. AWS S3)
-- Support for **OpenTelemetry** logs & traces + Prometheus metrics
+- Prometheus metrics and **OpenTelemetry** logs & traces in the connectors runtime. Server OTLP export is unavailable pending runtime integration.
```
spetz: Claims Prometheus metrics and OpenTelemetry exist only in the connectors runtime and that server OTLP export is unavailable. The server calls `logging.late_init(..., &TelemetrySettings::from(&config.telemetry))` at `core/server/src/boot/mod.rs:176` with working OTLP gRPC and HTTP exporters, and `core/server/src/http/metrics.rs` serves the Prometheus registry. The PR's own new test at `core/integration/tests/server/telemetry.rs:24` boots the server with `IGGY_TELEMETRY_LOGS_TRANSPORT=http` against a mock collector, so the PR contradicts itself.

core/cli/src/args/topic.rs
```diff
@@ -166,18 +166,18 @@ pub(crate) struct TopicUpdateArgs {
...
     pub(crate) name: String,
-    /// Compression algorithm for the topic, set to "none" for no compression
+    /// Compression metadata (none or gzip). Payload compression is not implemented
     #[arg(value_parser = clap::value_parser!(CompressionAlgorithm), verbatim_doc_comment)]
     pub(crate) compression_algorithm: CompressionAlgorithm,
     /// New max topic size in human-readable format like "unlimited" or "15GB"
     ///
-    /// "server_default" or skipping parameter makes CLI to use server default (from current server config) max topic size
-    /// Can't be lower than segment size in the config.
+    /// Skipping this parameter or using "server_default" sends the legacy zero sentinel. Pass an explicit value to avoid inconsistent option reporting.
```
spetz: CLI help now tells the operator to work around a bug instead of fixing it: "Skipping this parameter or using server_default sends the legacy zero sentinel. Pass an explicit value to avoid inconsistent option reporting." Same text at line 180 for `message_expiry`. The divergence is real. `TopicCreateOptions::parse_committed` maps the sentinel to `None` at `core/common/src/types/options/mod.rs:825`, while `resource_options_from_wire` keeps the raw value and `topic.options.extend(updated_options)` stores it. After `iggy topic update` without `-m`, the typed `max_topic_size` keeps its old value while `GetTopic`'s option map reports `server_defaul`

core/metadata/src/stm/stream.rs
```diff
@@ -2244,10 +2244,8 @@ impl StateHandler for UpdateTopicRequest {
 
         stream.topic_index.remove(&topic.name);
         topic.name = new_name_arc.clone();
-        // Settings arrive only through the options block now, so the typed
-        // fields are a projection of it and cannot drift. Absent means absent:
-        // a client that sends just a rename leaves every setting alone, and one
-        // built before a key existed cannot erase it.
+        // Parsing treats default sentinels as absent for these typed fields,
```
spetz: Same defect, other end. The replaced comment documents the divergence ("Parsing treats default sentinels as absent for these typed fields, while the stored option map below retains their raw values") rather than reconciling the two writes three lines apart.

core/bench/src/utils/mod.rs
```diff
@@ -187,7 +190,11 @@ fn add_environment_variables(parts: &mut Vec<String>, server_address: &str) {
 
     if is_localhost {
         let iggy_vars: Vec<_> = std::env::vars()
-            .filter(|(k, _)| k.starts_with("IGGY_"))
+            .filter(|(name, _)| {
+                RUNTIME_ENV_VARS.contains(&name.as_str())
```
spetz: The new allowlist keeps `ServerConfig` env mappings plus two hand-copied shard variables, so `IGGY_CONFIG_PATH` and `IGGY_ENV_PATH` no longer reach the report's reproduction command. Both change what server the run measured. `SERVER_PROCESS_ENV_VARS` at `core/configs/src/server_config/server.rs:52` already enumerates the non-`ServerConfig` set including those two.

core/integration/src/bench_utils.rs
```diff
@@ -180,3 +180,92 @@ pub fn get_random_path() -> String {
     .display()
     .to_string()
 }
+
+#[cfg(test)]
+mod tests {
```
spetz: Two new tests that spawn a real server and run a full benchmark sit in the library's `#[cfg(test)]` module, so they land in the `--lib` binary rather than the `tests/mod.rs` binary. They get none of the `#[ctor]` harness from `core/integration/tests/mod.rs:169`: no tracing subscriber, no per-test file under `test_logs/`, no failure log dump on panic. A CI failure in either produces no diagnostic output.

core/server/src/http/forward.rs
```diff
@@ -135,8 +136,13 @@ const RESPONSE_CAPACITY_HINT: usize = 64 * 1024;
 /// applied op, not this follower's (the response layer only fills either when
 /// absent); the applied op is also what this node records as the caller's
 /// read-your-writes floor, so dropping it here would reopen the stale read.
-const RELAYED_RESPONSE_HEADERS: [HeaderName; 4] =
-    [CONTENT_TYPE, RETRY_AFTER, VIEW_HEADER, APPLIED_OP_HEADER];
+const RELAYED_RESPONSE_HEADERS: [HeaderName; 5] = [
```
spetz, line 141: `RELAYED_RESPONSE_HEADERS` grows to five entries. The doc block above it explains why each of `content-type`, `retry-after`, `iggy-view` and `iggy-applied-op` is relayed, and was not extended for `iggy-durability`.

spetz (changes requested):

spetz (approved):

mmodzelewski (approved):

PR #4152, opened 2026-09-12
fix(connectors): report failures and correct format conversion

hubcio:
The runtime can count failed sink deliveries as processed messages,
misclassify transformation errors and leave failed instances in
running metrics. Protobuf and Avro conversion can corrupt encoded data
or apply mappings incorrectly.

Inspect FFI outcomes, correct error and lifecycle accounting, redact
state URL secrets and run HTTP telemetry on the existing runtime.
Repair codec and transform behavior with regression coverage and
matching configuration guidance.

Failed sink batches are logged and counted as errors. Runtime batch replay is not introduced by these fixes.

Validation: formatting, manifest sorting, Clippy, build, Markdown and TOML checks passed. Selected crate suites passed 470 tests, and runtime/HTTP integration passed 32 tests.

CI owns the remaining validation.


core/connectors/sdk/src/decoders/proto.rs
```diff
@@ -384,18 +407,32 @@ impl ProtoStreamDecoder {
...
                 Ok((json_value, new_cursor))
             }
+            1 | 5 => {
+                let (value, new_cursor) = Self::parse_fixed_integer(data, cursor, wire_type)?;
+                let json_value = match (wire_type, field_desc.r#type()) {
+                    (1, Type::Fixed64) => simd_json::OwnedValue::from(value),
+                    (1, Type::Sfixed64) => simd_json::OwnedValue::from(value as i64),
+                    (5, Type::Fixed32) => simd_json::OwnedValue::from(value as u32),
+                    (5, Type::Sfixed32) => simd_json::OwnedValue::from(value as u32 as i32),
+                    _ => simd_json::OwnedValue::String("unsupported_wire_type".into()),
```
spetz, line 425: `Type::Float` (wire type 5) and `Type::Double` (wire type 1) are absent from the `(wire_type, field_desc.r#type())` match, so both fall to this arm and decode to the literal string `"unsupported_wire_type"`. The same PR teaches `ProtoStreamEncoder` and `ProtoConvert` to write those two types as little-endian fixed32 and fixed64, so a JSON to protobuf to JSON round trip now loses exactly the two types the encoder just learned to emit.

core/connectors/runtime/src/state/http.rs
```diff
@@ -298,7 +298,7 @@ impl HttpStateProvider {
                     return Ok(LoadResponse::NotFound);
                 }
                 Ok(response) => return Ok(LoadResponse::Failure(response)),
-                Err(send_error) => TransientFailure::Send(send_error),
+                Err(send_error) => TransientFailure::Send(send_error.without_url()),
```
spetz, line 303: The `without_url()` redaction is applied to both `TransientFailure::Send` arms but not to the `TransientFailure::Read` arm at `core/connectors/runtime/src/state/http.rs:294`, which stores `read_error.to_string()` and feeds it verbatim into `describe` at `core/connectors/runtime/src/state/http.rs:619`. In reqwest 0.13.4, the pinned workspace version, `Response::bytes()` delegates to `do_bytes`, whose error path is `crate::error::decode(err).with_url(*self.url)`, and `impl Display for Error` ends with `if let Some(url) = &self.inner.url { write!(f, " for url ({url})")?; }`. A body-read failure on the load path logs the full state URL with its query string, the exact leak this change closes for send failures. `resource_label` is already redacted through `url_label`, so this is the only remaining vector, and the new `given_unavailable_state_store_when_booting_should_fail_startup` assertion exercises only the `Send` path.

core/connectors/sdk/src/decoders/proto.rs
```diff
@@ -83,26 +83,32 @@ impl ProtoStreamDecoder {
...
+        Ok(())
     }
 
     pub fn load_schema(&mut self) -> Result<(), Error> {
         let schema_path = self.config.schema_path.clone();
         let descriptor_set = self.config.descriptor_set.clone();
 
-        if let Some(path) = schema_path {
-            self.compile_schema_internal(&path)?;
+        let old_message_descriptor = self.message_descriptor.take();
```
spetz, line 98: `load_schema` takes the descriptors up front and restores them only on `Err`, but `compile_schema_internal` returns `Ok(())` after logging "Falling back to Any wrapper mode" for both an unreadable file and a protox compile failure. A broken schema keeps the last good descriptor while a missing one discards it and downgrades output to the Any envelope, both reporting success. `given_loaded_schemas_when_file_changes_should_preserve_errors_and_clear_fallback` asserts exactly that split, so it is deliberate. Same shape at `core/connectors/sdk/src/encoders/proto.rs:120` and `core/connectors/sdk/src/transforms/proto_convert.rs:114`.

spetz (changes requested):

spetz (approved):

mmodzelewski (approved):

PR #4153, opened 2026-09-12
fix(connectors): defer source checkpoints and surface sink failures

hubcio:
Elasticsearch/InfluxDB sources could advance checkpoints before delivery acknowledgement, while sinks could report partially rejected batches as successful.

- Defer Elasticsearch/InfluxDB checkpoints until acknowledgement and fix source query/payload handling, including PostgreSQL tracking keys.
- Surface PostgreSQL/Elasticsearch sink failures, retry only rejected Elasticsearch items, and cap PostgreSQL inserts at the bind limit.
- Fix S3 write probes, share Quickwit readiness retries, enforce ClickHouse numeric bounds, and clean up buffer, authorization and routing configuration.

Preserves PostgreSQL acknowledgement/CDC behavior from merged #3957. Terminal sink failures still have no runtime replay.

Local validation: 360 source + 398 sink unit tests and 23 PostgreSQL source + 50 sink integration tests passed, along with build, fmt, sort, Clippy, TOML and Markdown checks.


core/connectors/sinks/s3_sink/src/sink.rs
```diff
@@ -45,7 +46,27 @@ impl Sink for S3Sink {
...
         let bucket = crate::client::create_bucket(&self.config).await?;
 
-        crate::client::verify_bucket(&bucket).await?;
+        let prefix = self
+            .config
+            .prefix
+            .as_deref()
+            .unwrap_or_default()
+            .trim_matches('/');
+        let probe_key = if prefix.is_empty() {
```
spetz: The startup probe moved from the bucket root into the configured prefix so it exercises the same write scope as data uploads, but cleanup is still best-effort `let _ = bucket.delete_object(&probe_key)`. The probe exists for write-only IAM policies, and such a policy normally grants `s3:PutObject` without `s3:DeleteObject`, so on exactly the deployments this targets a zero-byte `.iggy-sink-probe` now lands inside the data prefix, survives, and is rewritten on every open. A consumer listing the prefix sees an object the sink did not write as data. At the bucket root it was outside the reader's path.

core/connectors/sinks/elasticsearch_sink/src/lib.rs
```diff
@@ -38,6 +37,27 @@ use tracing::{error, info, warn};
 sink_connector!(ElasticsearchSink);
 
 const DEFAULT_TIMEOUT_SECONDS: u64 = 30;
+const BULK_RETRY_POLICY: RetryPolicy = RetryPolicy {
```
spetz: `BULK_RETRY_POLICY` is a hardcoded const of three attempts, one second base, five second cap, while every other HTTP-shaped sink takes `max_retries` and `retry_delay` from config. With the default 30 second client timeout an operator facing a slow cluster has no knob on a consume call that can now run for roughly three timeouts plus backoff.

spetz (changes requested):

spetz (approved):

mmodzelewski (approved):

numinnex (approved):

PR #4169, opened 2026-09-13
fix(cluster): preserve group membership and reduce produce/poll overhead

spetz:
Metadata and partition primaries can differ. Auto-commit polls and Rust
manual/interval offset writes could move the coordinator connection and
lose consumer-group membership. Backups previously stored automatic
offsets locally without replication.

- Route auto-commit polls to partition primaries while retaining the
  coordinator in Rust (TCP/QUIC/WebSocket), Java, C#, Go and Node.js.
- Route Rust offset stores/deletes through the same primary connection pool.
- Add session attachment and routing commands (14/103/104/123), with
  session, ownership and metadata fences; return ambiguous data outcomes
  without replay.
- Harden SDK connection recovery, credentials, topology discovery and
  deadlines. Forward HTTP polls with bounded streaming and refusal retries.
- Reduce request copies/lookups, adapt disk reads to demand, and improve
  Java batch encoding and flush handling.
- Defer obsolete-file cleanup and protect reused WAL and segment paths.
- Raise defaults: inbox 65,536; peer queue 4,096; repair ring 65,536 entries
  and 64 MiB per partition. Add poll and WAL diagnostics.

Validation: rebuilt server binaries; Rust build/fmt/sort/Clippy; protocol,
storage and SDK suites; split-primary TCP/QUIC/WebSocket/HTTP regressions;
Java/C#/Go/Node tests, including protocol parity and deadline coverage.

Upgrade all servers before binary SDKs, then restart group consumers.
Existing store/delete layouts remain unchanged; routing needs the new commands.


core/partitions/src/iggy_partition.rs
```diff
@@ -2987,12 +2993,18 @@ where
...
-                .durable_consumer_offsets
-                .covers(kind, consumer_id, offset)
+        // A replica that cannot originate the prepare cannot record this
+        // progress anywhere a peer will ever see. `Ok(None)` would leave
+        // `complete_poll` applying the offset to local state alone, so the
+        // poll would report progress the group never agreed, and a later
+        // read on the primary would hand the same messages out again.
+        // Refusing keeps the outcome retriable on a replica that can commit.
+        if !consensus.is_primary() || !consensus.is_normal() || consensus.is_transferring() {
+            return Err(IggyError::TransientNotAccepted);
```
numinnex, line 3018: This refusal reaches HTTP as a 503 (`http/error.rs:83`), but `GET .../messages` is in neither `partition_write_routes` nor `forwardable_routes`, and `forward.rs:265` short-circuits every GET, so no HTTP client can reach a node that would serve it.

TCP/QUIC recover via the roster walk. All four HTTP surfaces (Rust, Java, .NET, Python) treat code 58 as fatal, and the Rust/Python high-level consumers send `auto_commit=true` at stock settings (`consumer_builder.rs:69` -> `consumer.rs:745` -> `http/messages.rs:40`).

On any cluster with >=2 replicas, replica 0 is primary for every partition until a view change, so 2 of 3 HTTP endpoints now fail every auto-commit poll that previously returned messages.

core/partitions/src/poll_plan.rs
```diff
@@ -383,16 +387,82 @@ pub enum DiskReadOutcome {
...
+    /// starting offset can sit inside a batch the index resolved before it,
+    /// and an underestimate only costs another read of the next chunk.
+    fn chunk_len(&self, remaining: u32) -> u64 {
+        let Some(bytes_per_message) = self.bytes_per_message else {
+            return DISK_POLL_CHUNK_MAX;
+        };
+        u64::from(bytes_per_message)
+            .saturating_mul(u64::from(remaining))
+            .saturating_add(COMMAND_HEADER_SIZE as u64)
+            .clamp(DISK_POLL_CHUNK_MIN, DISK_POLL_CHUNK_MAX)
```
numinnex, line 489: `chunk_len(c) = mean*c + COMMAND_HEADER_SIZE` equals the producer's batch size exactly at `c == N` (the +256 cancels the floor loss). One message below that, the chunk lands under a whole batch, `consumed == 0` discards the entire read, and the `*4` ladder at :636 starts.

Against the old flat 1 MiB (one read for any `c`): 4.99x the bytes at `c = N-1`, 2.50x at `c = 500`, 2.10x at `c = 100`, with 2-3x the syscalls.

The SDK rustdoc (`consumer.rs:314`) and `examples/rust/src/stream-builder/stream-consumer-config` both use `.batch_length(100)`, while `bench/src/actors/consumer/client/high_level.rs:138` sets the consumer's batch length from the producer's, so iggy-bench only ever generates the single neutral point `c == N`.

numinnex (dismissed):

core/partitions/src/poll_plan.rs
```diff
@@ -532,6 +574,74 @@ impl DiskReadPlan {
...
+    #[allow(clippy::cast_possible_truncation)]
+    async fn walk_segment(
+        &self,
+        file: &compio::fs::File,
+        query: MessageLookup,
+        count: u32,
+        persisted: u64,
+        walk: &mut DiskWalk,
+    ) -> SegmentWalk {
+        let mut chunk_len = self.chunk_len(count - walk.matched);
```
hubcio: warning: the first read ignores the gap from the index entry to the target offset, so a poll deep in a flush group walks it in 64 KiB reads. add that gap to the first read size.

core/partitions/src/poll_plan.rs
```diff
@@ -532,6 +574,74 @@ impl DiskReadPlan {
         }
     }
 
+    /// Read one segment from `walk.position` until the count is filled, the
+    /// segment is exhausted, or the walk must fail closed.
+    ///
+    /// The chunk length is already bounded by `DISK_POLL_CHUNK_MAX` and by the
```
hubcio: nit: this says `DISK_POLL_CHUNK_MAX` bounds the chunk, but line 636 grows `chunk_len` past it. keep only the `persisted` bound here.

core/journal/src/partition_journal.rs
```diff
@@ -1075,6 +1074,24 @@ impl<S: DurableStorage> PartitionPrepareJournal<S> {
...
 
+    /// Whether [`Self::reclaim_obsolete`] has anything left to remove.
+    #[must_use]
+    pub fn has_obsolete(&self) -> bool {
+        !self.obsolete.is_empty() || self.cleanup_directory_dirty
+    }
+
+    /// Remove a bounded batch of the files no published generation retains.
+    ///
+    /// Separate from the append path so an acknowledgement never waits on
```
hubcio: nit: an ack can wait on the unlinks, because the writer loop reclaims before it pops the next append. and `rewrite` queues the old generation for `truncate_from`, `reset` and `reset_with_prepare` too, not only the checkpoint and boot scan.

core/journal/src/partition_journal.rs
```diff
@@ -1075,6 +1074,24 @@ impl<S: DurableStorage> PartitionPrepareJournal<S> {
...
+    /// unlinks and a directory barrier for history it does not depend on. Only
+    /// a checkpoint and the boot scan ever queue work here, and both do so
+    /// after the publication that excludes those files, so the queue holds
+    /// nothing a reader or a recovery could still need. The writer owns the
+    /// journal, so this runs between mutations and never beside one.
+    pub async fn reclaim_obsolete(&mut self) {
```
hubcio: simplification: `reclaim_obsolete` only forwards to `cleanup_obsolete`, and `has_obsolete` guards a call that is already a no-op on an empty queue. make `cleanup_obsolete` public under this name and drop both.

core/partitions/src/persistence.rs
```diff
@@ -1140,49 +1148,66 @@ impl<S: DurableStorage> PartitionPersistence<S> {
...
-                }
+                self.publish_mutation(journal, rebuild_references);
+            }
+            // After the notification, never before the barrier it would delay.
+            // A checkpoint queues the generation it replaced, so reclaiming it
+            // here keeps the unlinks and the directory barrier out of the
+            // acknowledgement the next append is waiting on, while still
+            // running once per mutation so a busy partition reclaims as
+            // promptly as an idle one.
+            if journal.has_obsolete() {
```
hubcio: nit: this runs before the next append is popped, so that append still waits on the unlinks and the directory fsync, contrary to the comment above. reclaim only on an empty queue, with a mutation-count bound for busy partitions.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
```diff
@@ -268,69 +283,112 @@ public static ByteBuf toMessagesBatch(List<Message> messages) {
...
-            batch.writeIntLE(message.userHeaders().length);
-            batch.writeIntLE(message.payload().length);
-            batch.writeLongLE(0); // reserved
-            batch.writeBytes(message.payload());
-            batch.writeBytes(message.userHeaders());
-            batch.setLongLE(
-                    frameStart, xxHash3(batch, frameStart + Long.BYTES, batch.writerIndex() - frameStart - Long.BYTES));
         }
+        if (latestTimestamp.subtract(originTimestamp).compareTo(MAX_TIMESTAMP_DELTA_MICROS) > 0) {
+            throw new IggyInvalidArgumentException("Message origin timestamp delta exceeds unsigned 32-bit range");
```
hubcio: nit: the old message and the server's `InvalidMessageTimestampDelta` error both name the offending delta, this one drops it. append the delta and the message index.

foreign/java/java-sdk/src/main/java/org/apache/iggy/serde/BytesSerializer.java
```diff
@@ -354,9 +412,13 @@ private static long xxHash3(ByteBuf buffer, int index, int length) {
...
     private static byte[] readAllBytes(ByteBuf buffer) {
-        var bytes = new byte[buffer.readableBytes()];
-        buffer.readBytes(bytes);
-        return bytes;
+        try {
+            var bytes = new byte[buffer.readableBytes()];
+            buffer.readBytes(bytes);
+            return bytes;
+        } finally {
+            buffer.release();
```
hubcio, line 450: nit: `readAllBytes` now releases the shared `Unpooled.EMPTY_BUFFER` that `toBytes(Map)` returns for empty headers, which only works because that release is a no-op. return an empty `byte[]` for empty headers before any buffer exists.

core/partitions/src/poll_plan.rs
hubcio: warning: the decoder already returns the batch length in `UnexpectedEof { need, .. }`, and line 898 throws it away before the `*4` ladder. return `need` from `walk_disk_chunk` and grow to `max(chunk_len * 4, need)`, so the second read is exactly the batch.

core/partitions/src/iggy_partition.rs
hubcio: nit: the role check runs before `covers`, so a caught-up backup refuses a poll whose offset the durable table already covers. test `covers` first and keep the refusal for uncovered offsets only.

hubcio (changes requested):
warning: `core/journal/src/partition_journal.rs:879` - after one failed unlink the path stays in `self.obsolete`, so the next append reuses the link and the reclaim then removes it. drop the retained path from `self.obsolete` before reusing it (only `append_batch_referenced_buffered` reaches this).

core/sdk/src/poll_routing.rs
```diff
@@ -0,0 +1,1284 @@
...
+                }
+                self.routes
+                    .lock()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner)
+                    .remove(&key);
+                if Instant::now() + retry_interval >= deadline {
+                    return Err(IggyError::TransientNotAccepted);
+                }
+                sleep(retry_interval).await;
+                retry_interval = (retry_interval * 2).min(ROUTING_RETRY_MAX_INTERVAL);
```
spetz, line 299: The retry interval now doubles from 50 ms to a 1 second ceiling, which cuts a refused poll from roughly 600 coordinator round trips over its 30 second budget to roughly 30. The route is still removed on every `TransientNotAccepted` retry at `core/sdk/src/poll_routing.rs:292`, so a refusal unrelated to leadership still re-resolves a route that did not change.

core/consensus/src/client_table.rs
```diff
@@ -1653,6 +1674,27 @@ impl ClientTable {
         self.slots[slot_idx].as_ref().map(|entry| entry.epoch)
     }
 
+    /// Attach only after the caller has authenticated `user_id` and waited
+    /// for the local metadata frontier to cover the requested session.
+    /// The registered user owns the session, matching authenticated login
```
spetz, line 1679: A doc comment now states that the registered user owns the session and that client ids and epochs are identifiers rather than authentication secrets. The check is unchanged, so a second connection authenticated as the same user can still adopt another group member's identity and take its partitions. The sweep validator called this resolution out in advance: documenting it records the behavior as intended rather than closing it. The request doc at `core/binary_protocol/src/requests/system/attach_consumer_session.rs:25` is untouched and still reads "The server verifies ownership and the exact parent epoch", so the two documents now disagree about the same guarantee

core/partitions/src/persistence.rs
```diff
@@ -1140,49 +1155,69 @@ impl<S: DurableStorage> PartitionPersistence<S> {
...
+            }
+            // Reclaim after publishing the mutation when the queue is empty.
+            // Appends arriving during reclaim still wait for its unlinks and
+            // directory barrier.
+            // A partition that never drains would then never reclaim, so force
+            // a pass every `RECLAIM_MUTATIONS_MAX` mutations and accept that
+            // one group's latency.
+            mutations_since_reclaim += 1;
+            let idle = self.queue.borrow().is_empty();
+            if idle || mutations_since_reclaim >= RECLAIM_MUTATIONS_MAX {
```
spetz, line 1168: At the legal 1 MiB minimum segment size a checkpoint queues roughly 64 links, so 64 mutations enqueue about 520 obsolete paths against the 16 a pass removes.

core/server/config.toml
```diff
@@ -993,12 +997,16 @@ offset_reservation_lease = 65536
...
 
 # Byte ceiling for the evicted ring per partition; whichever ring cap (this or
 # evicted_ring_capacity) trips first evicts. Bounds the ring memory a burst of
 # large batches can pin. Must be > 0 and <= "256 MiB".
-evicted_ring_bytes_max = "16 MiB"
+# PER PARTITION and per replica, so the node-wide ceiling is this times the partitions it
+# hosts, pinned as anonymous memory beside the same bytes in page cache. Raising it on a
+# node with many partitions is a large memory decision; partition_repair_ring_bytes on the
+# metrics endpoint reports what the rings on each shard actually hold.
+evicted_ring_bytes_max = "64 MiB"
```
spetz, line 1002: `evicted_ring_bytes_max` is still 64 MiB per partition per replica with no node-wide budget. The `BTreeMap` change removed the CPU cost of the larger ring but not its memory ceiling.

core/server/config.toml
```diff
@@ -726,7 +726,8 @@ pin_cores = true
 # ~ the prepare queue depth of the planes the shard hosts ([metadata] on
 # shard 0, [partition] elsewhere) times replica_count times directions.
 # Both depths are tunable, so raising either raises the capacity needed here.
-inbox_capacity = 1024
+# Slots are allocated at startup per shard, even before traffic arrives.
+inbox_capacity = 65536
```
spetz, line 725: `inbox_capacity` is still 65536 preallocated slots per shard, against a comment whose own sizing rule does not produce that number.

core/sdk/src/poll_routing.rs
```diff
@@ -0,0 +1,1284 @@
...
+                if connections.len() >= MAX_DATA_CONNECTIONS {
+                    error!(endpoint = route.endpoint, protocol = %T::PROTOCOL, limit = MAX_DATA_CONNECTIONS, "primary poll connection pool is full");
+                    return Err(IggyError::InvalidConfiguration);
+                }
+                let slot = Arc::default();
+                connections.insert(route.endpoint.clone(), Arc::clone(&slot));
+                slot
+            }
+        };
+        let mut connection = slot.lock().await;
```
spetz, line 341: One connection slot per endpoint, still locked across the attach and the poll, so a client gets at most one concurrent auto-commit poll per node.

core/configs/src/server_config/partition.rs
```diff
@@ -115,7 +115,7 @@ fn default_offset_reservation_lease() -> NonZeroU32 {
 }
 
 /// Mirrors `partitions::EVICTED_RING_CAPACITY`.
-pub const DEFAULT_EVICTED_RING_CAPACITY: usize = 4096;
+pub const DEFAULT_EVICTED_RING_CAPACITY: usize = 65536;
```
spetz, line 118: `DEFAULT_EVICTED_RING_CAPACITY` still equals `MAX_EVICTED_RING_CAPACITY`.

core/sdk/src/poll_routing.rs
```diff
@@ -0,0 +1,1284 @@
...
+            routes: Mutex::default(),
+            connections: Mutex::default(),
+            credentials: Mutex::default(),
+            next_heartbeat: Mutex::default(),
+        }
+    }
+}
+
+impl<T> PollRouter<T> {
+    pub(crate) fn clear_session(&self) {
```
spetz, line 130: `clear_session` still leaves `roster_size` set.

core/sdk/src/poll_routing.rs
```diff
@@ -0,0 +1,1284 @@
...
+        }
+        let payload = request.to_bytes();
+        let parameters_size = request.strategy.encoded_size() + size_of::<u32>() + size_of::<u8>();
+        let key = (
+            GET_POLL_ROUTING_CODE,
+            payload.slice(..payload.len() - parameters_size),
+        );
+        self.send_routed(coordinator, POLL_MESSAGES_ON_PRIMARY_CODE, key, payload)
+            .await
+    }
```
spetz, line 220: The shared wire prefix is still unasserted.

core/sdk/src/poll_routing.rs
hubcio: Keeping route invalidation here. Code 58 guarantees non-admission but does not
identify the cause: leadership, membership/metadata fencing and capacity can
produce the same response. Reusing the route on an assumed capacity refusal
could keep hitting an obsolete primary or attachment for the entire retry
budget. Backoff bounds discovery traffic while preserving recovery. Avoiding
discovery safely needs a distinct refusal reason or a trustworthy route hint;
that protocol extension is outside this change.

core/consensus/src/client_table.rs
hubcio: Agreed that the request documentation was ambiguous. It now states the actual
contract: the authenticated user must match the parent session's registered
user, the epoch must match, and attaching does not extend the parent's lifetime.
Connections using the same account can share its group membership. This is
consistent with login/register, which already allows that user to rebind its
client id (`core/metadata/src/impls/metadata.rs`, `submit_register_in_process`).
This documents the existing user-level boundary; it does not introduce
isolation between processes sharing credentials. Providing that isolation
needs a session capability and consistent changes to both attachment and
login/resume, so I am keeping that redesign separate.

core/partitions/src/persistence.rs
hubcio: The 16-path limit was removed in a64a966c5. `cleanup_obsolete` snapshots
`self.obsolete.len()` and attempts every path from that pass. The 64-mutation
constant only schedules cleanup; it does not limit how many paths are removed.
Failed unlinks are queued for the next pass. The checkpoint-generation
regression verifies that one cleanup pass drains the backlog and reopening
preserves the published history.

core/server/config.toml
hubcio: Correct: the default remains 64 MiB per partition per replica, and there is no
node-wide budget. It is also a payload-byte budget, not an allocation or RSS
ceiling: spare buffer capacity, alignment and map overhead are additional,
and the newest entry is retained even if it exceeds the byte budget alone.
I corrected the config comments, occupancy documentation and metric help to
make those limits explicit. The configured defaults are intentional and remain
unchanged. Allocation-aware accounting and a node-wide eviction policy need
separate work; the BTreeMap change addresses lookup cost, not those concerns.

core/server/config.toml
hubcio: The 65,536-slot default is intentional. I removed the misleading per-plane
formula: sizing must account for the combined traffic of all consensus groups
hosted by the shard, replica fan-out and burst headroom. The comment still
states that slots are preallocated per shard. A single group's prepare depth
does not derive this default.

core/sdk/src/poll_routing.rs
hubcio: Correct: one router has one serialized data connection per endpoint. The lock
protects attachment plus the following request, and cancellation marks an
unfinished exchange unusable before another caller can reuse it. Simply
releasing the lock would break that ownership guarantee; the underlying
transports also serialize exchanges. A bounded pool of independent sessions
could increase concurrency, but needs measured workload requirements and its
own attachment/cancellation lifecycle. Keeping that pool design separate from
this membership fix.

core/configs/src/server_config/partition.rs
hubcio: Keeping both at 65,536. The default is an intentional tuning choice; the
validation maximum remains a guard against still larger allocations. Equality
does not disable either the entry cap or byte cap. Increasing the maximum just
to leave headroom would allow a larger per-partition footprint without a
requirement for it; operators can still configure a lower capacity.

core/sdk/src/poll_routing.rs
hubcio: Keeping the known roster across a session reset because it describes cluster
topology, not group membership. `clear_session` advances the generation and
clears routes and data connections; in-flight replies and attachments are
fenced against that generation. Topology discovery updates the cached roster
when it obtains a valid response. Clearing it on every session invalidation
adds discovery I/O without strengthening the membership fence. The reconnect
regressions cover late route replies and pending data-connection users.

core/sdk/src/poll_routing.rs
hubcio: This is asserted in a64a966c5 by
`offset_writes_share_the_routing_request_prefix` in
`core/binary_protocol/src/requests/consumer_offsets/mod.rs`. It independently
encodes store and delete requests, decodes their prefix as the routing request,
and checks both bytes and consumed length. The combinations cover both
consumer kinds, numeric/named identifiers, absent/zero/maximum partition ids,
and both acknowledgement levels. No additional wire change is needed.

mmodzelewski (approved):

hubcio (approved):

hubcio (approved):

PR #4173, opened 2026-09-14
feat(python): expose update_user options

### LINKED ISSUE #4164: Python SDK: expose update_user options
### Description

The Rust `update_user()` API accepts `UserUpdateOptions`, but the Python wrapper always passes `UserUpdateOptions::default()`:

```rust
async fn update_user(
    &self,
    user_id: &Identifier,
    username: Option<&str>,
    status: Option<UserStatus>,
    options: &UserUpdateOptions,
) -> Result<(), IggyError>;
```

The current server has no updatable user option keys, so every non-empty map is presently rejected. The raw options field nevertheless exists to allow future server-side keys without another wire or SDK signature change.

The Python stream and topic update APIs already expose equivalent raw option maps.

This covers B32 of the Python SDK parity work tracked in #3893.

### Affected area / component

Python SDK

### Proposed solution

- Add an optional keyword argument `options: dict[str, str] | None = None` to `IggyClient.update_user()`.
- Build `UserUpdateOptions { raw: options.unwrap_or_default() }`, following the existing `update_stream()` pattern.
- Preserve the current behavior when `options` is omitted or `None`.
- Regenerate `foreign/python/apache_iggy.pyi` and update the method documentation.
- Extend `foreign/python/tests/test_user.py` to verify:
  - Existing username and status updates still work when `options` is omitted.
  - An empty options map succeeds.
  - An unknown non-empty option is forwarded and rejected by the current server.

Introducing typed user options is outside this issue.

### Alternatives considered

Keep the options hardcoded until the server defines its first user option. That would require another Python API change when such an option is introduced.

### Contribution

### Good first issue

Jorge-Polanco-Roque:
Closes #4164.

The Python SDK's `update_user()` always passed `UserUpdateOptions::default()`, so there was no way to actually set the options the Rust API accepts. This exposes an optional `options` kwarg and forwards it into `UserUpdateOptions`, following the exact pattern already used by `update_stream()`. Omitting it keeps the previous behavior.

**Tests:** added `test_update_user_with_empty_options_succeeds` and `test_update_user_forwards_options` (mirroring `test_stream.py`); the full `test_user.py` suite (53) passes against a local server, and `cargo clippy` + `ruff` are clean. The `.pyi` stub was regenerated with `stub_gen` + `ruff format`.

---
_Disclosure: per CONTRIBUTING's AI-assistance note — I use AI to help write and review code, and I've reviewed and verified every line myself._


foreign/python/src/client.rs
```diff
@@ -354,35 +354,37 @@ impl IggyClient {
     ///     user_id: User identifier as `str | int`.
     ///     username: New username as `str | None`; unchanged when `None`.
     ///     status: New status as `UserStatus | None`; unchanged when `None`.
+    ///     options: Additional option keys as `dict[str, str] | None`, forwarded
```
jiengup: Let's explicitly state that this parameter is reserved for future use.

jiengup (commented):
LGTM, a minor suggestion.

foreign/python/tests/test_user.py
```diff
@@ -361,6 +361,39 @@ async def test_update_username_and_status_together(
...
+
+    @pytest.mark.asyncio
+    async def test_update_user_forwards_options(
+        self, iggy_client: IggyClient, unique_name
+    ):
+        """Test update_user forwards option keys to the server."""
+        username, password = unique_credentials(unique_name)
+        created = await iggy_client.create_user(username, password)
+
+        with pytest.raises(RuntimeError):
```
spetz: Bare `pytest.raises(RuntimeError)` passes on any RuntimeError, including an unrelated failure. Mirrors `test_update_stream_forwards_options` so it's consistent with the existing suite.

foreign/python/src/client.rs
```diff
@@ -354,35 +354,37 @@ impl IggyClient {
...
     ///
     /// Returns:
     ///     An awaitable that resolves to `None` when the user is updated.
     ///
     /// Raises:
     ///     ValueError: If a string identifier is invalid.
     ///     RuntimeError: If the request fails.
-    #[pyo3(signature = (user_id, username=None, status=None))]
+    #[pyo3(signature = (user_id, username=None, status=None, options=None))]
     #[gen_stub(override_return_type(type_repr="collections.abc.Awaitable[None]", imports=("collections.abc")))]
```
spetz, line 368: Signature spacing is `status=None, options=None` here but `options = None` in `update_stream`. Cosmetic, rustfmt does not touch attribute contents.

spetz (dismissed):
LGTM, just minor improvements but not blocking.

spetz (approved):
Fixed minor issues.

hubcio (approved):

PR #4183, opened 2026-09-14
feat(python): expose client disconnect and shutdown

### LINKED ISSUE #4163: Python SDK: expose client disconnect and shutdown lifecycle methods
### Description

The Rust `Client` trait exposes both `disconnect()` and `shutdown()`, while the Python `IggyClient` currently exposes only `connect()`.

`disconnect()` closes the current connection while allowing the client to connect again. `shutdown()` releases the underlying client resources and is terminal for binary transports.

This covers B7 and B8 of the Python SDK parity work tracked in #3893.

```rust
async fn disconnect(&self) -> Result<(), IggyError>;
async fn shutdown(&self) -> Result<(), IggyError>;
```

### Affected area / component

Python SDK

### Proposed solution

- Add asynchronous `disconnect()` and `shutdown()` methods to the `#[pymethods]` block in `foreign/python/src/client.rs`.
- Delegate directly to the corresponding Rust `IggyClient` methods and map failures through the existing `to_runtime_error` helper.
- Regenerate `foreign/python/apache_iggy.pyi`, with both methods returning `Awaitable[None]`.
- Extend `foreign/python/tests/test_connectivity.py` to verify:
  - `disconnect()` is idempotent.
  - Requests fail while disconnected.
  - A disconnected client can connect and authenticate again.
  - `shutdown()` is idempotent.
  - Requests after shutdown fail with the underlying client-shutdown error.
- Document transport-specific behavior, including the no-op lifecycle used by HTTP.

No new wrapper types are needed.

### Alternatives considered

Expose only `disconnect()`. This would leave Python without the terminal cleanup operation provided by the Rust client.

### Contribution

### Good first issue

dedsec-terminal:

Closes #4163

## Rationale

The Rust Client trait exposes disconnect() and shutdown(), but the Python IggyClient only exposed connect(), so there was no way to close the connection for reconnect or release the client without dropping it.

## What changed?

Added async disconnect() and shutdown() to the Python binding, following the exact pattern already used by connect(). disconnect() closes the current connection while allowing a later connect(), though the previous login is cleared so login_user() has to run again. shutdown() releases the client and rejects further calls with Client shutdown. Both are idempotent, and both are no-ops over HTTP where there is no connection to manage. Stub entries and connectivity tests are included.

## Local Execution

- Passed: ruff check and ruff format --check on the touched test file
- Not run: cargo fmt / clippy / stub_gen and the live-server pytest suite (no Rust toolchain or running server on this machine); the .pyi entries were written by hand to match the new methods


slbotbm:
Closing due to author not following CONTRIBUTING.md guidelines

justinmclean:
This PR says it wasn't run locally. CONTRIBUTING.md lists "Code not ran and tested locally" as a close reason. The guide also says that using an AI assistant is fine, but you're responsible for the code you submit. You need to explain every part of the change and answer review questions yourself.

PR #4184, opened 2026-09-14
feat(python): expose consumer shutdown and offset drain timeout

### LINKED ISSUE #4165: Python SDK: expose consumer shutdown and offset drain timeout
### Description

The Rust high-level consumer exposes an asynchronous `shutdown()` method that drains pending offset work, performs final commits when enabled, leaves the consumer group, and stops its connection watcher.

The Python `IggyConsumer` currently has no corresponding method. Its `consumer_group()` builder also does not expose the Rust builder's `offset_drain_timeout`, which controls how long shutdown waits for background offset tasks.

This covers B22 and the currently applicable part of B23 in the Python SDK parity work tracked in #3893.

The `commit_failed_messages` setting listed in B23 is not present in the current Rust SDK and should not be added to the Python API.

### Affected area / component

Python SDK

### Proposed solution

- Add `IggyConsumer.shutdown()` as an asynchronous method that locks the wrapped Rust consumer, calls its `shutdown()`, and maps any error to `RuntimeError`.
- Preserve the Rust method's idempotent behavior.
- Add an optional `offset_drain_timeout: datetime.timedelta | None` argument to `IggyClient.consumer_group()`.
- When provided, convert it to `IggyDuration` and pass it to `IggyConsumerBuilder::offset_drain_timeout()`.
- When omitted, preserve the Rust builder default of five seconds.
- Regenerate `foreign/python/apache_iggy.pyi` and document the supported shutdown call order.
- Extend `foreign/python/tests/test_consumer_group.py` to verify:
  - A consumer can be shut down after its consumption loop exits.
  - Repeated shutdown calls succeed.
  - A custom offset-drain timeout is accepted and propagated.
  - Shutdown leaves a consumer group promptly so another member can receive its partitions.

Adding a replacement for the removed `commit_failed_messages` option is outside this issue.

### Alternatives considered

Continue relying only on the `asyncio.Event` accepted by `consume_messages()`. That event stops one consumption loop but does not expose the Rust consumer's resource and offset cleanup operation.

### Contribution

### Good first issue

dedsec-terminal:

Closes #4165

## Rationale

The Rust high-level consumer exposes an async shutdown() that drains pending offsets, does the final commit, leaves the group and stops the watcher. The Python IggyConsumer had no equivalent, and consumer_group() did not expose offset_drain_timeout, so there was no way to trigger that cleanup from Python.

## What changed?

Added async IggyConsumer.shutdown(), following the same lock-and-map-to-RuntimeError pattern as the other consumer methods. It is idempotent through the Rust implementation. Added an optional offset_drain_timeout to IggyClient.consumer_group(), converted to IggyDuration and passed to IggyConsumerBuilder::offset_drain_timeout(). Omitted keeps the Rust default of five seconds. Documented the supported order in consume_messages(), iter_messages(), consumer_group() and shutdown(): signal the event, let consume_messages() return or finish iterating, then call shutdown(). Updated the stub, the high-level consumer example, and added shutdown and timeout tests to test_consumer_group.py.

## Local Execution

- Passed: ruff check and ruff format --check on the touched Python files, py_compile on the tests and example, ast parse on the .pyi, git diff --check
- Not run: cargo fmt / clippy / stub_gen and the live-server pytest suite (no Rust toolchain or running server on this machine); the .pyi entries were written by hand to match the new methods


slbotbm:
Closing due to author not following CONTRIBUTING.md guidelines

justinmclean:
This PR says it wasn't run locally. CONTRIBUTING.md lists "Code not ran and tested locally" as a close reason. The guide also says that using an AI assistant is fine, but you're responsible for the code you submit. You need to explain every part of the change and answer review questions yourself.

PR #4193, opened 2026-09-15
fix(server): distribute TCP-TLS and WSS connections across shards

hubcio:
TCP-TLS and WSS pinned handshakes and encrypted I/O to shard 0.

Reuse TCP/WS fd delegation to transfer raw sockets and listener
configuration before either handshake. Preserve placement policy,
client ID allocation and the existing inbox frame size.
Clear metadata on shutdown and reject late connection installs.
Verify shard ownership, encrypted round trips and failure cleanup.


spetz (dismissed):
LGTM, good catch!

core/integration/tests/server/specific.rs
```diff
@@ -20,33 +20,56 @@ use crate::server::scenarios::{reconnect_after_restart_scenario, restart_offset_
...
+
+const TLS_TEST_SHARDS: usize = 4;
+const TLS_TEST_CLIENTS: usize = TLS_TEST_SHARDS * 2;
+const TLS_CLEANUP_TIMEOUT: Duration = Duration::from_secs(10);
+const TLS_POLL_INTERVAL: Duration = Duration::from_millis(20);
 
 #[iggy_harness(
     test_client_transport = TcpTlsGenerated,
-    server(tls = generated)
+    server(tls = generated, sharding.cpu_allocation = "0..4", logging.level = "info")
```
numinnex: `sharding.cpu_allocation = "0..4"` lands in `extra_envs`, and `harness/handle/server.rs:374-376` applies it with `insert` — hard-overriding the adaptive `"all"` fallback at `:297-313`. Below 4 usable CPUs `validators.rs:218-223` rejects `Range(0,4)` and the server refuses to boot, reporting a cpu_allocation error that names nothing about TLS. At exactly 4 CPUs it passes with zero margin while three tests each demand four shards under concurrent nextest. Same string at `:51` and `:64`.

core/integration/tests/server/specific.rs
```diff
@@ -141,3 +164,99 @@ async fn restart_offset_skip(harness: &mut TestHarness) {
...
+                );
+                assert!(
+                    fields.contains(&format!("shard-{owner}").as_str()),
+                    "{line}"
+                );
+                assert!(installed.insert(wire_id), "client installed twice: {line}");
+                counts[owner] += 1;
+            }
+            if installed == client_ids {
+                assert_eq!(
```
numinnex: Exact equality pins the round-robin *phase*, not just the distribution. `ship_client_fd` advances the shared `client_rr` before `peer_addr()` and `dup_fd()` can fail (`coordinator.rs:397-401`), so a single transient accept failure on any of the four client transports shifts the phase and this assertion fails. `TLS_TEST_SHARDS` is also hard-coded to 4, independently of what the server actually bound.

core/integration/tests/server/specific.rs
```diff
@@ -141,3 +164,99 @@ async fn restart_offset_skip(harness: &mut TestHarness) {
...
+        client.disconnect().await.unwrap();
+    }
+    // Polling opens a separate SDK connection. A fresh observer has none,
+    // so an exact count also detects leaked partition-client sessions.
+    let observer = harness.root_client().await.unwrap();
+    let observer_id = observer.get_me().await.unwrap().client_id;
+    tokio::time::timeout(TLS_CLEANUP_TIMEOUT, async {
+        loop {
+            let connected = observer.get_clients().await.unwrap();
+            if connected.len() == 1 {
```
numinnex: This can pass vacuously. `IggyShard::list_all_clients` (`core/shard/src/lib.rs:1979-1987`, `:2014-2024`) under-reports warn-only when a shard's `ListClients` try_send is rejected or the 3s gather times out, so `len() == 1` is satisfiable while another shard still holds a leaked session — and the loop breaks on the first such iteration. Pre-PR every TLS session lived on shard 0, so no cross-shard gather was involved.

core/shard/src/coordinator.rs
```diff
@@ -358,31 +344,78 @@ impl ShardZeroCoordinator {
...
+            LifecycleFrame::ClientWssConnectionSetup { fd, meta, config }
+        })
+    }
+
+    #[must_use]
+    pub const fn total_shards(&self) -> u16 {
+        self.total_shards
+    }
+
+    fn ship_client_fd(
```
numinnex, line 391: No in-flight cap on un-handshaken client delegations, where the replica plane gates every accept on `try_acquire_replica_handshake_slot` (`message_bus/src/lib.rs:893`, cap 256). Each delegated connection commits an inbox slot, ~18 KiB of ring buffer, two compio tasks, a `clients()` entry and a `client_meta` entry before one TLS byte and before LOGIN, held up to `handshake_grace`. A TLS-only node previously had no unauthenticated path into a worker shard at all.

core/shard/src/coordinator.rs
```diff
@@ -358,31 +344,78 @@ impl ShardZeroCoordinator {
...
+        &self,
+        stream: TcpStream,
+        transport: ClientTransportKind,
+        build_frame: impl FnOnce(fd_transfer::DupedFd, ClientConnMeta) -> LifecycleFrame,
+    ) -> Result<u128, SendError> {
         let target = self.next_client_target();
```
numinnex, line 397: One `client_rr` cursor serves all four client transports. With k transport classes interleaved on one cursor and N shards, each class reaches only N/gcd(k,N) distinct shards — so an alternating accept pattern on 4 shards pins TLS connections to a subset, for the life of each connection. The matrix test at `:951` cannot observe this: it rotates all four transports and asserts the aggregate sequence, not per-transport placement.

core/shard/src/coordinator.rs
```diff
@@ -358,31 +344,78 @@ impl ShardZeroCoordinator {
...
         let client_id = self.mint_client_id(target);
         let peer_addr = stream.peer_addr().map_err(SendError::DupFailed)?;
```
numinnex, line 399: `peer_addr()` and `dup_fd()` failures share one `SendError::DupFailed` and record no metric, now on all four client planes. `fcntl(F_DUPFD_CLOEXEC)` failing is EMFILE/ENFILE — the fd-exhaustion signal — and it is indistinguishable here from getpeername on a reset peer. Only the `try_send` branch below reaches `record_frame_drop`; successful delegations are counted nowhere, which is why the new integration test has to scrape stdout to observe placement.

core/shard/src/coordinator.rs
```diff
@@ -358,31 +344,78 @@ impl ShardZeroCoordinator {
...
 
         let fd = fd_transfer::dup_fd(&stream).map_err(SendError::DupFailed)?;
-        let meta = ClientConnMeta::new(client_id, peer_addr, ClientTransportKind::Ws);
-        let setup = LifecycleFrame::ClientWsConnectionSetup { fd, meta };
+        let meta = ClientConnMeta::new(client_id, peer_addr, transport);
+        let setup = build_frame(fd, meta);
         if let Err(e) = self.senders[target as usize].try_send(ShardFrame::lifecycle(setup)) {
```
numinnex, line 404: Setup frames ride the main lane (`self.senders[target]` derefs to `inner`, `core/shard/src/lib.rs:580-586`) — the same bounded queue as `ShardFrame::Consensus`, reply lane unused, no priority. The queue does not need to fill to hurt: every setup frame the destination pump dequeues costs that shard tens of microseconds during which no consensus frame on it moves. Pre-PR that cost landed on shard 0 only; it now lands on the shards carrying consensus.

core/shard/src/lib.rs
```diff
@@ -734,6 +734,21 @@ pub enum LifecycleFrame {
         fd: DupedFd,
         meta: ClientConnMeta,
     },
+    /// Delegate TCP-TLS before reading TLS bytes. The destination wraps the
```
numinnex, line 737: Both new variants omit the routing invariant their neighbours state — that the owning shard is encoded in the top 16 bits of `meta.client_id` — which the reply path depends on and which this PR's own test asserts by hand-rolling `>> 112` (`specific.rs:199`). Neither names the receiving installer the way `ClientWsConnectionSetup` names `install_client_ws_fd`, nor records that dropping the frame releases both the fd and a config refcount.

core/message_bus/src/lib.rs
```diff
@@ -403,42 +403,26 @@ pub type AcceptedQuicClientFn = std::rc::Rc<dyn Fn(AcceptedQuicConn)>;
 /// requires a dupable plaintext fd) stays well-defined.
 pub type AcceptedWsClientFn = std::rc::Rc<dyn Fn(compio::net::TcpStream)>;
 
+/// Listener TLS configuration shared with the shard owning each connection.
+/// Per-connection TLS state is created only on the destination runtime.
+pub type SharedTlsServerConfig = std::sync::Arc<rustls::ServerConfig>;
```
numinnex, line 408: Applied at 5 sites while ~20 on the same delegation path still spell `Arc<rustls::ServerConfig>` (`installer/tcp_tls.rs:54`, `installer/wss.rs:53`, `client_listener/{tcp_tls,wss}.rs:67,73,93,99`, `transports/tcp_tls.rs:143`, `transports/wss.rs:101`). The alias also hides the `Arc` at the one boundary where the shared-nothing invariant makes it load-bearing: a reader of `config: SharedTlsServerConfig` (`shard/src/lib.rs:742`) cannot see that it is atomically refcounted, crosses threads, and carries shared interior state.

core/message_bus/src/client_listener/tcp_tls.rs
```diff
@@ -86,11 +82,7 @@ pub fn bind(
     // cannot enable it accidentally.
     cfg.max_early_data_size = 0;
```
numinnex, line 83: Only `max_early_data_size` is overridden; the remaining rustls defaults now cross shards with this config. `session_storage` is `ServerSessionMemoryCache::new(256)` behind a `std::sync::Mutex`, and with `NeverProducesTickets` + `send_tls13_tickets = 2` the stateful branch runs two blocking `put()`s per successful TLS 1.3 handshake. That lock was single-threaded while TLS terminated on shard 0; it is now contended across N shard threads on a runtime with no blocking pool, where a futex sleep parks the whole shard.

core/message_bus/src/client_listener/tcp_tls.rs
```diff
@@ -86,11 +82,7 @@ pub fn bind(
...
 
-    let listener = bind_reusable_tcp_listener(addr)
-        .map_err(|_| IggyError::CannotBindToSocket(addr.to_string()))?;
-    let actual = listener
-        .local_addr()
-        .map_err(|e| IggyError::IoError(e.to_string()))?;
+    let (listener, actual) = bind_nodelay_listener(addr)?;
```
numinnex, line 85: This moves TCP-TLS onto `bind_nodelay_listener`, but `wss::bind` (`client_listener/wss.rs:75`) still uses `bind_reusable_tcp_listener`. The two TLS listeners were symmetric before this PR, and WSS is now the only client plane with no listener-level Nagle setting — its sole mechanism is the warn-and-continue per-connection call at `installer/wss.rs:58-63`. The updated helper doc at `mod.rs:101-102` and the new test at `mod.rs:151-164` both exclude WSS.

core/message_bus/src/transports/tcp_tls.rs
```diff
@@ -218,7 +218,11 @@ impl TransportConn for TcpTlsTransportConn {
         let mut tls = match self.state {
             ConnState::Established(tls) => *tls,
             ConnState::Pending { stream, role } => {
-                match compio::time::timeout(handshake_grace, handshake(role, stream)).await {
+                let outcome = futures::select_biased! {
+                    () = ctx.shutdown.wait().fuse() => return,
```
numinnex, line 222: This arm returns with no log and no comment, where every sibling exit in this match warns with `%label, %peer`. The router has already emitted an INFO `installing delegated TCP-TLS client fd` for the connection, which then disappears with no record of why.

core/message_bus/src/transports/wss.rs
```diff
@@ -174,12 +174,14 @@ impl TransportConn for WssTransportConn {
...
         // state machines) and would otherwise push `run`'s state past
         // the `clippy::large_futures` threshold.
-        let tls_stream = match compio::time::timeout(
-            handshake_grace,
-            Box::pin(tls_handshake(role, self.stream)),
-        )
-        .await
-        {
+        let tls_outcome = futures::select_biased! {
+            () = ctx.shutdown.wait().fuse() => return,
```
numinnex, line 178: Same shape as the TCP-TLS arm: a silent `return` with no log, in a function where every other exit warns with `%label, %peer`. The second `select_biased!` at `:218` repeats it, and that one drops an already-established `tls_stream`.

core/message_bus/src/installer/replica.rs
```diff
@@ -369,8 +369,6 @@ pub fn install_replica_conn<C: TransportConn>(
     // `notify_connection_lost` stands down whenever a live registry entry
     // exists - but the loser also drives `replica_dispatch_loop`, which
     // must never hand the winner's replica id to `on_message`.
-    // `compio::runtime::JoinHandle::drop` does not cancel the spawned
-    // task, so we have to tell the tasks to stand down in-band.
     let install_aborted = Rc::new(Cell::new(false));
```
numinnex, line 372: The rationale for `install_aborted` was deleted rather than corrected. The flag is still required: `drain_rejected_registration` (`installer/common.rs:38-51`) awaits both handles under a timeout rather than dropping them, so the loser's tasks run cooperatively to completion and must stand down in-band or they hand the winner's replica id to `on_message`. The three sibling `JoinHandle` comments in this PR were corrected in place; this one now carries no stated reason at all.

core/integration/tests/server/specific.rs
hubcio: the local update removes the fixed CPU range and reads the server's shard count from startup. the tests use the harness fallback and still check every shard.

core/integration/tests/server/specific.rs
hubcio: a starting-phase shift keeps eight consecutive allocations evenly split across four shards. the local update reads the actual shard count and retains the exact distribution check.

core/integration/tests/server/specific.rs
hubcio: the local update keeps one observer per shard and requires the exact set of observer IDs. a missing shard reply or leaked session can no longer satisfy the cleanup check.

core/shard/src/coordinator.rs
hubcio: the client path has no global handshake cap, and moving it to workers widens the impact. this needs a separate client-admission change with explicit limits and cleanup.

core/shard/src/coordinator.rs
hubcio: alternating TCP and TLS accepts can concentrate TLS on a subset of shards while total connection counts stay balanced. per-transport balancing needs a separate placement-policy change.

core/shard/src/coordinator.rs
hubcio: `DupFailed` preserves `io::Error`, and the accept callback logs it, so `EMFILE` remains visible. separate failure counters would help monitoring, but the error details are not lost.

core/shard/src/coordinator.rs
hubcio: setup shares the existing TCP/WS queue, while TLS handshakes run in spawned tasks. queue contention under connection churn needs measurement before changing scheduling.

core/shard/src/lib.rs
hubcio: the owner-bit rule is shared with the adjacent setup variants. a common comment would clarify it, but documenting fd and `Arc` drops repeats standard ownership behavior.

core/message_bus/src/lib.rs
hubcio: `SharedTlsServerConfig` lets `shard` carry the configuration without adding a direct rustls dependency. replacing the local `Arc<rustls::ServerConfig>` signatures would add no type-safety benefit.

core/message_bus/src/client_listener/tcp_tls.rs
hubcio: the shared session cache uses a blocking mutex during handshakes and resumption, outside steady-state message I/O. contention needs benchmarking before choosing a change that could reduce resumption performance.

core/message_bus/src/client_listener/tcp_tls.rs
hubcio: accepted WSS sockets already enable `TCP_NODELAY` before TLS. setting it on the listener would align startup error handling with TCP/TCP-TLS; normal WSS connections already disable nagle.

core/message_bus/src/transports/tcp_tls.rs
hubcio: this branch handles requested shutdown, so a warning would report normal cleanup as a failure. the installer still removes the connection state; a debug message is optional.

core/message_bus/src/transports/wss.rs
hubcio: both branches handle shutdown during connection setup, including after TLS completes. dropping the stream and cleaning the registry is intentional; a debug message is optional.

core/message_bus/src/installer/replica.rs
hubcio: the preceding comment still explains why losing tasks must not dispatch using the winner's replica ID. only the incorrect claim that dropping a `JoinHandle` does not cancel its task was removed.

spetz (approved):

numinnex (approved):

PR #4194, opened 2026-09-15
fix(cluster): prevent spurious heartbeat elections

hubcio:
A former primary can retain a higher heartbeat counter than its
successor and start another election despite valid heartbeats.
Scope freshness to the view so a new primary's heartbeats refresh
liveness while replayed heartbeats still cannot prevent failover.

The five-second mesh reconnect sweep also matches the heartbeat
timeout. Retry every second by default so a dial that races peer
startup gets another attempt before that timeout, and bound each
dial by the same period: the sweep joins every dial, so a peer
that drops SYNs would otherwise hold it for the kernel connect
timeout and starve the retries. Read the period from the bus
configuration in tests and benchmarks so the default has a single
source in TOML.


core/consensus/src/impls.rs
```diff
@@ -1198,9 +1198,9 @@ where
 
     timeouts: RefCell<TimeoutManager>,
 
-    /// Monotonic timestamp from the most recent accepted commit heartbeat.
-    /// Old/replayed commit messages with a lower timestamp are ignored.
-    heartbeat_timestamp: Cell<u64>,
+    /// View and counter of the last emitted or accepted heartbeat. A new
+    /// primary's counter can be lower, so freshness is scoped to its view.
```
numinnex, line 1202: Maybe if we received a lower timestamp than the previous one, we can instead take max(current, new). In the VSR when message with a strictly non monotonic StartViewChange/DoViewChange is received, the default mode is to assume that it's the correct view.

hubcio: max() on a lower counter either keeps the timer stale (bug stays) or refreshes it on any heartbeat (replay guard gone)

we compare (view, counter), so a new view's first heartbeat always wins and a lower counter inside one view is still a replay. Same as TigerBeetle, which zeroes heartbeat_timestamp on every view transition and keeps the strict compare in on_commit.

i'd leave it as is

spetz (approved):

numinnex (approved):

PR #4204, opened 2026-09-16
fix(connectors): tag sink batches with the payload's schema

### LINKED ISSUE #4053: connectors: batches are tagged with the decoder's schema, not the payload's, so avro/proto/flatbuffer reach sinks as the wrong Payload variant
### Bug description

When a sink's `[[streams]]` entry sets `schema = "avro"`, `"proto"` or `"flatbuffer"`, the sink receives its messages as the wrong `Payload` variant.

The runtime tags each batch with the decoder's schema rather than the type the decoder returned. Those three decoders all return `Payload::Json` under the config the runtime gives them, so the tag and the bytes disagree. The SDK rebuilds the payload from the tag, so a sink is handed a `Payload::Avro` holding JSON.

What happens next depends on the sink. Sinks that only handle `Payload::Json` — ClickHouse, Delta, Elasticsearch, Meilisearch, Quickwit, Iceberg — log the message as an unsupported payload type and drop it. Sinks that accept the raw variants — S3, HTTP, SurrealDB — base64-encode the JSON text and store it as though it were Avro bytes.

Expected: the sink receives `Payload::Json`, because that is what the decoder produced.

**Root cause**

```rust
// core/connectors/runtime/src/sink.rs:359-363
let messages_metadata = MessagesMetadata {
    partition_id,
    current_offset,
    schema: decoder.schema(),   // Schema::Avro
};
```

The bytes sent alongside come from `Payload::try_into_vec()`, which for the `Payload::Json` the Avro decoder returned gives JSON (`core/connectors/sdk/src/lib.rs:163-165`). The same tag is put on `RawMessages` at `:731`. The SDK rebuilds from it (`sdk/src/sink.rs:199`, then `lib.rs:263`: `Schema::Avro => Payload::Avro(value)`).

No connector config avoids this. The runtime builds the Avro decoder with `..AvroConfig::default()` (`sink.rs:532-536`), so `extract_as_json` cannot be turned off from a connector file, and Proto and FlatBuffer get `Schema::decoder()` defaults with no knob at all.

Sources are unaffected: a source plugin tags its own `ProducedMessages.schema` and the encoder takes the `Payload` directly (`source.rs:430-435`), with no rebuild step.

**Which schemas are affected**

| `schema` | decoder returns | batch tagged | sink receives |
|---|---|---|---|
| `json`, `text`, `raw` | matching variant | matching | works |
| `avro` | `Payload::Json` (`decoders/avro.rs:169`) | `Avro` | `Payload::Avro` |
| `flatbuffer` | `Payload::Json` (`decoders/flatbuffer.rs:109`) | `FlatBuffer` | `Payload::FlatBuffer` |
| `proto` | `Payload::Json`, if the message is an `Any` (`decoders/proto.rs:529`) | `Proto` | `Payload::Raw` |

The three that work are the ones whose decoder output matches its own `schema()`. Avro and FlatBuffer default to `extract_as_json: true`; Proto with no schema configured falls through to `decode_as_any` (`proto.rs:264-265`), which also returns JSON. Proto ends up as `Payload::Raw` rather than `Payload::Proto` because `try_into_payload` re-decodes the bytes as a `prost_types::Any` and falls back to `Raw` when that fails (`lib.rs:260`).

**Impact**

For the Json-only sinks the batch is lost without an error being raised. The ClickHouse sink ends up with an empty body and returns `Ok(())` (`sinks/clickhouse_sink/src/sink.rs:146-152`), and offsets are committed with the poll request (`AutoCommit::When(AutoCommitWhen::PollingMessages)`, `runtime/src/sink.rs:522`) — the SDK documents that mode as sending "the commit with the poll request itself, before your code sees the batch" (`core/sdk/src/clients/consumer.rs:501`). So the offsets advance before the sink runs, and restarting the connector does not redeliver.

The batch is still counted as consumed. A run of 100 messages logs `Consumed 100 messages`, and the connector reports `Processed 0 messages.` at close.

Related to #3950 and PR #3954, which bind the sink's return status and let a sink defer the offset commit. Neither covers this: the ClickHouse sink returns `Ok(())` on an empty body, so even with `offset_commit = after_consuming` the batch would be committed as successfully written.

**Suggested fix**

Tag the batch with the payload's type rather than the decoder's. Changing each decoder's `schema()` would not be enough on its own: transforms run after decoding and can change the type again, and the Proto decoder returns either `Json` or `Raw` depending on the path it takes.

Adding a `Payload::schema()` and using it where `MessagesMetadata` and `RawMessages` are built (`runtime/src/sink.rs:362` and `:731`) covers all three formats in one place. I can send a PR either way; say which you prefer.

**Compatibility**

The fix changes what three sinks write. S3, HTTP and SurrealDB match on `Payload::Avro` and base64-encode it, so an avro stream reaches them today as base64 of JSON text, stored as though it were Avro bytes. Nothing reports this. After the fix they receive `Payload::Json` and write a JSON document, so anyone consuming their current output would see it change shape. That output is corrupt.

The Json-only sinks go from dropping the batch to writing it correctly. Streams configured `json`, `text` or `raw` are unaffected, because the decoder's schema and the payload's type already agree and the tag does not move.

Two related gaps:

- `StreamConsumerConfig` only exposes `avro_schema_json` and `avro_schema_path`, so flatbuffer and proto cannot be given a schema through the runtime at all.
- Nothing exercises a non-JSON schema through the runtime's sink path, and all ten configs under `runtime/example_config/connectors/` use `schema = "json"`. A test across the decoders would catch this class of problem.

### Affected area / component

Connectors

### Deployment

Compiled from source

### Versions

master @ `412014a` — `iggy-connectors` 0.5.0-edge.6, `iggy_connector_sdk` 0.4.0-edge.3, `iggy_connector_clickhouse_sink` 0.2.0-edge.4

### Hardware / environment

macOS arm64, local build; not hardware-dependent.

### Sample code

```toml
type = "sink"
key = "clickhouse"
enabled = true
version = 0
name = "ClickHouse sink"
path = "target/release/libiggy_connector_clickhouse_sink"
verbose = true

[[streams]]
stream = "repro"
topics = ["avro"]
schema = "avro"
avro_schema_json = '{"type":"record","name":"Event","fields":[{"name":"id","type":"long"},{"name":"name","type":"string"}]}'
batch_length = 100
poll_interval = "5ms"
consumer_group = "repro-sink"

[plugin_config]
url = "http://localhost:8123"
database = "default"
table = "repro"
insert_format = "row_binary"
```

Producer — an ordinary `IggyProducer`, with each payload an Avro datum for that schema:

```rust
let schema = apache_avro::Schema::parse_str(SCHEMA)?;
let record = apache_avro::types::Value::Record(vec![
    ("id".to_owned(), apache_avro::types::Value::Long(id)),
    ("name".to_owned(), apache_avro::types::Value::String(format!("row-{id}"))),
]);
let datum = apache_avro::writer::datum::GenericDatumWriter::builder(&schema)
    .build()?
    .write_value_to_vec(record)?;
let message = IggyMessage::builder().payload(datum.into()).build()?;
```

### Logs

100 Avro messages into a single-partition topic, ClickHouse sink in `row_binary` mode:

```
INFO  iggy_connectors::sink: Processing 100 messages for sink connector with ID: 1
ERROR connector: connector_target="iggy_connector_clickhouse_sink::body"
      RowBinary mode: skipping unsupported payload type at offset 0
ERROR ... at offset 1, 2, ... 99
ERROR connector: connector_target="iggy_connector_clickhouse_sink::sink"
      ClickHouse sink ID: 1 — no serialisable messages in batch of 100
INFO  iggy_connectors::sink: Consumed 100 messages in 675µs for sink connector with ID: 1
INFO  connector: ClickHouse sink ID: 1 closed. Processed 0 messages.
```

`SELECT count() FROM default.repro` returns 0.

Restarting the connector against the same consumer group redelivers nothing: no batch, no errors, still no rows. The offsets were committed at poll time, so the 100 messages cannot be recovered without resetting them by hand.

Tagging the batch from the payload's type rather than the decoder's makes the same configuration land all 100 rows with the expected values, which confirms the cause.

### Iggy server config

Default server config; not involved.

### Reproduction

1. Run an Iggy server and a ClickHouse instance with
`CREATE TABLE repro (id Int64, name String) ENGINE = MergeTree ORDER BY id`.
2. Produce Avro-encoded records matching the schema above to `repro/avro`.
3. Run `iggy-connectors` with the sink config above.
4. The log fills with `skipping unsupported payload type` and the table stays
empty.
5. Restart the connector: nothing is redelivered, and the table is still empty.

### Contribution

### Good first issue

MarcusKainth:

Closes #4053

## Rationale

Sinks configured `avro`, `proto` or `flatbuffer` receive the wrong `Payload` variant. JSON-only sinks drop the batch after the offset has already been committed, so those messages are lost with no redelivery, and sinks that accept the raw variants store base64 of JSON text as though it were Avro bytes.

## What changed?

The runtime tagged each batch with the decoder's schema, which names the wire format a decoder reads rather than the variant it returns. All three decoders extract to `Payload::Json` under the configuration the runtime gives them, and the SDK rebuilds the payload from that tag alone, so a sink was handed a `Payload::Avro` holding JSON.

The tag now comes from the payload itself through `Payload::schema()`, read after transforms run. One `Schema` covers a whole FFI call, so messages are grouped into contiguous runs of the same variant, and a uniform batch stays one run and one `consume()` call. The new integration test covers `avro` only, because `StreamConsumerConfig` has no schema configuration for flatbuffer or proto.

## Compatibility

Three of the six stream `schema` values change tag. `avro`, `flat_buffer` and `proto` all extract to JSON under the runtime's default decoder settings (`avro.rs` `L40`, `flatbuffer.rs` `L42`, `proto.rs` `L48`), so a batch from any of them now arrives tagged `json`. S3, HTTP and SurrealDB will write a JSON document for those streams instead of base64 of JSON text labelled as the wire format. That output was wrong, so this is a correction, but anyone reading it will see the shape change. A `proto` stream into Doris improves outright: it failed every poll before and writes rows now. `json`, `text` and `raw` streams are unaffected, because the decoder's schema and the payload's variant already agree.

`Payload::Proto` now reaches a sink, which it never did before. A `proto_convert` transform hands it over whenever it has no descriptor, and also whenever it has one and the message is not a top-level JSON object, because `encode_json_with_schema` returns `Err` for any array or scalar (`proto_convert.rs` `L285`). On a `json` stream the same messages previously arrived as `Payload::Json`. Every sink that writes documents now reads a `Payload::Proto` that holds JSON as that document: ClickHouse (JSONEachRow and RowBinary), Doris, Delta, Iceberg, Elasticsearch, Meilisearch, Quickwit, S3, SurrealDB and HTTP. Proto text that is not JSON is handled the way each sink handles `Text`. RowBinary keeps failing a batch on a non-object document, as it did on `master`. Two smaller shifts remain: a `text` or `raw` stream with `proto_convert` now hands the sink `Payload::Proto` where it handed `Text` or `Raw`, which changes nothing now that both are treated alike; and a descriptor-backed `proto_convert` on a `json` stream, which the SDK dropped on `master` because the `json` tag sent protobuf bytes through the JSON parser, now arrives as `Raw`.

`iggy_connector_sdk` moves from 0.4.0 to 0.5.0. No FFI signature or `Schema` variant changed, but the `proto` tag changed meaning on the wire. A plugin built against 0.4.0 is better off under this runtime on `avro`, `flat_buffer` and `proto` streams, because it now receives a `json` tag it can read, and worse off on one pipeline: a `json` stream with `proto_convert`, where it rebuilds the `proto`-tagged run as `Payload::Raw` where `master` handed it `Payload::Json`. Plugins built against 0.4.0 or earlier must be rebuilt.

`avro`, `flat_buffer` and `proto` streams now pay a per-message JSON parse inside the plugin. They avoided it before by handing the plugin a mislabelled payload it could not use.

One poll can reach a plugin as several `consume()` calls, one per contiguous payload variant, all repeating the same `current_offset`. `iggy_connector_sink_runs_total` counts those calls; the `ffi` stage histogram stays one sample per batch, summed over its runs. SurrealDB's `iggy_schema` column records the variant received, so rows written before this change on `avro`, `flat_buffer` and `proto` streams disagree with new ones.

## Local Execution

- Passed. `cargo fmt`, `cargo sort --no-format`, Clippy with all features and all targets on every touched crate and the integration crate, `taplo`, `markdownlint`, license headers, trailing whitespace and newline, `typos` and `cargo machete` all pass. Unit tests pass for the runtime (236), the SDK (185) and the ten sinks touched, including the new tests for each. The `schema_tagging`, Elasticsearch, ClickHouse and Doris `proto_text` integration tests pass against a real server, runtime and containers.
- The same test files were run against a `master` worktree, `d8a8ece` and this branch. The three pipeline tests pass on `master`, fail on `d8a8ece` and pass here; the schema-tagging and sink container tests fail on `master` and pass on both later refs; every `json`, `text`, `raw` and source-path control passes on all three. The full table is in the review summary comment.
- Pre-commit hooks ran on every commit.

## AI Usage

1. Claude Opus 5.
2. Review and rebase, not implementation. I wrote the code. Claude reviewed it afterwards and I applied its suggestions with my own reasoning. Claude also rebased the branch onto current master and resolved the conflicts against #4152.
3. The unit tests pin the schema tag per variant, the one-FFI-call-per-uniform-batch invariant and the `Proto` round-trip. The integration test sends real Avro datums through a running server and connectors runtime and asserts the sink is handed `Payload::Json`. The local checks listed above were run.
4. Yes.


core/connectors/sdk/src/lib.rs
```diff
@@ -157,6 +157,26 @@ pub enum Payload {
...
+    /// is set, and the Proto decoder returns `Payload::Json` or `Payload::Raw`
+    /// depending on the path it takes. A transform may change the variant again
+    /// after that. Anything tagging a payload for transport has to read the tag
+    /// off the payload it actually holds.
+    pub const fn schema(&self) -> Schema {
+        match self {
+            Payload::Json(_) => Schema::Json,
+            Payload::Raw(_) => Schema::Raw,
+            Payload::Text(_) => Schema::Text,
+            Payload::Proto(_) => Schema::Proto,
```
rohankumardubey, line 174: Payload::Proto contains text, but tagging it as Schema::Proto does not round-trip through the sink SDK: Schema::Proto::try_into_payload() reconstructs it as Raw (or Json for a protobuf Any), never Proto. The existing JSON→Proto transform can produce this text, so this change can cause a sink to receive Raw where it previously received Json and silently skip the message. Could we preserve the intended payload type across the FFI boundary and add a regression test for that transform path?

core/connectors/runtime/src/sink.rs
```diff
@@ -599,8 +596,14 @@ async fn process_messages(
     }
     let decode_elapsed = decode_start.elapsed();
 
-    let mut messages = Vec::with_capacity(decoded.len());
+    // One `Schema` tag covers a whole FFI call, so messages are grouped into
+    // contiguous runs of the same payload variant and each run is sent on its
+    // own. Every decoder and transform in tree is deterministic per instance, so
+    // a run holds the entire batch in practice and there is one call.
```
rohankumardubey: The claim that an in-tree transform always yields one run is not quite true. `ProtoConvert::json_to_protobuf()` can return `Raw` for a successful schema conversion and `Proto` for a fallback on another message in the same batch. Could we remove the “one call” assertion and test a mixed-output batch?

rohankumardubey (approved):

core/connectors/sdk/src/lib.rs
jiengup, line 174: The PR promises that the tag describes the actual payload after transforms. That does not hold for `Payload::Proto`:

1. proto_convert.rs#L627-L657 can return `Payload::Proto(json_string)` when no descriptor is configured or schema encoding fails.
2. The new runtime logic tags its UTF-8 bytes as `Schema::Proto`.
3. The receiver’s Schema::try_into_payload interprets those bytes as binary protobuf `Any`, falling back to `Payload::Raw`, never reconstructing `Payload::Proto`.

rohankumardubey (dismissed):
@MarcusKainth left couple of comments to address. Thanks for contributing.

core/connectors/sdk/src/lib.rs
MarcusKainth: Correct, and it is a regression against master, not just a round-trip gap: on a `schema = "json"` stream with a `proto_convert` transform the sink got `Payload::Json` on master and `Payload::Raw` on `31c0773`. ClickHouse skips those rows; Delta, Doris and Iceberg fail the batch.

`try_into_payload` had two callers meaning different things by `Schema::Proto` — the sink SDK meant the payload's variant, `runtime/src/source.rs:609` means the wire format a plugin sent. Split in d8a8ece: the sink rebuilds through `Payload::try_from_schema`, so all six variants keep theirs. Elasticsearch, Meilisearch and the ClickHouse string passthrough now take proto text as text; the JSON-only sinks still reject it, noted in Compatibility.

One correction: `json_to_protobuf` only falls back to `Payload::Proto` with no descriptor (`proto_convert.rs:656`) or non-object JSON (`:285`); with both it returns `Payload::Raw(binary)`, which round-trips.

Regression tests in d8a8ece: the round-trip test no longer exempts Proto, and a new Elasticsearch integration test drives a `proto_convert` pipeline through a real runtime and container.

core/connectors/runtime/src/sink.rs
MarcusKainth: Agreed, the comment overreached. Dropped in d8a8ece; it now names `ProtoConvert` and claims only that a uniform batch stays one run.

`json_to_protobuf` returns `Payload::Raw` when `encode_json_with_schema` succeeds (`proto_convert.rs:646`) and `Payload::Proto` when it does not (`:652`), discriminated per message by whether the JSON is a top-level object (`:285`), so one instance with a descriptor alternates. It also contradicted the branch's second commit, which caps the per-run reservation for exactly that case.

The existing split test drove the mix with a stub transform. d8a8ece adds two using the real one: a `proto_convert` unit test pinning both variants from one instance, and a runtime test running `ProtoConvert` through `process_messages` asserting three runs tagged Raw, Proto, Raw with nothing lost.

core/connectors/sdk/src/lib.rs
MarcusKainth: Confirmed, all three steps hold. `Schema::try_into_payload` served two callers that meant different things by `Schema::Proto`: the sink SDK meant the payload's variant, `runtime/src/source.rs:609` means the wire format a plugin sent. Split in d8a8ece, so the sink rebuilds through `Payload::try_from_schema` and the source arm is untouched. Detail in @rohankumardubey's thread.

One addition to step 1: with a descriptor loaded, `encode_json_with_schema` also falls back on non-object JSON (`proto_convert.rs:285`), so one instance can emit both variants.

mlevkov (commented):
Adversarial multi-agent review of `d8a8ece9`: four independent reviewers, then clean-room validators who saw only the raw claims and the diff, then tiebreaks on the three contested items. 10 of 22 initial claims were corrected and 1 was removed before anything reached this comment.

First, the parts that hold. The round trip is correct for all six `Payload` variants, the split of `Schema::try_into_payload` into two inverses is the right shape, and every line number in your reply to @rohankumardubey and @jiengup checks out. The earlier regression is genuinely fixed.

**Three findings look like merge blockers**, all the same shape: this PR establishes a remedy and applies it to three sinks out of six.

You added `| Payload::Proto(text)` to the `Payload::Text` arm in Elasticsearch, Meilisearch and the ClickHouse passthrough. `build_json_body`, `build_row_binary_body`, Doris, Delta and Iceberg did not get it, and all of them lose committed messages as a result. The offset commits at poll time (`runtime/src/sink.rs:515`), there is no replay and no dead-letter queue, so a failed or skipped batch is gone. `iggy_connector_messages_processed_total` counts the dropped rows and `iggy_connector_errors_total` stays flat, so the dashboard reads the same before and after the upgrade. Details inline on `body.rs` and `doris_sink/src/lib.rs`.

The Compatibility section is right that this surfaces a misconfiguration. The part that does not follow is the surfacing: on master these pipelines delivered correct rows, and after the upgrade they deliver none, with metrics reporting success.

Two corrections to the section itself: `flat_buffer` and `proto` streams also change tag and appear in neither the affected nor the unaffected list, and the descriptor-less case is not the only trigger. `encode_json_with_schema` returns `Err` for any non-object top-level JSON (`proto_convert.rs:285`), so a fully configured pipeline takes the fallback on every array, string, number, boolean and null message.

Also worth a look, not blocking:

- Four test gaps. The round-trip test compares only the variant, the per-run subtraction has no failing-run test, the new Elasticsearch test asserts the flattened shape, and the `Any` test does not check that `type_url` and `value` survived. Inline where they anchor.
- `surrealdb_sink/src/lib.rs:838` persists `messages_metadata.schema` into `iggy_schema`, so new rows on an avro stream now say `json`. Operator-visible, worth a README line.
- `http_sink/src/lib.rs:1197` says a nonzero return records no processed messages for the batch. The runtime now subtracts only the failed run.
- `.claude/skills/connector-sdk/SKILL.md:82` points item 12 at a schema list in `core/connectors/README.md`. That file has none; the list is in `sdk/README.md`.
- Pre-existing, exposed rather than caused: `avro_convert.rs:125` builds an encoder per message, `proto_convert.rs:630` serializes a document it then drops, `proto_convert.rs:296` can produce an empty `Payload::Raw` that now gets written, and `elasticsearch_sink/src/lib.rs:373` clones and parses per Raw message.

Simplifications, all optional: `runtime/src/sink.rs:606` can use one flat `Vec` with run offsets instead of a Vec per run (postcard writes a borrowed slice and an owned `Vec` identically, so the FFI bytes are unchanged); `sdk/src/lib.rs:294` can delegate five of six arms to `Payload::try_from_schema`, though it needs the variants named rather than a `_` arm or it removes the exhaustiveness check, and `mut value` then trips `-D warnings`; `runtime/src/sink.rs:751` can accumulate instead of subtracting. At `flatbuffer_convert.rs:146` I would keep the `Err` return rather than `unreachable!()`, since the error drops one message and a panic kills the consumer task.

core/connectors/sinks/clickhouse_sink/src/body.rs
```diff
@@ -98,7 +98,10 @@ pub(crate) fn build_string_body(
     let mut buf = Vec::with_capacity(messages.len() * 64);
     for msg in messages {
         match &msg.payload {
-            Payload::Text(s) => {
+            // `Payload::Proto` holds proto text, so a passthrough body takes
+            // it the same way it takes `Payload::Text`. The JSON and RowBinary
+            // builders above cannot: they need a parsed document.
```
mlevkov, line 103: This comment is the stated reason the JSON and RowBinary builders still skip `Payload::Proto`, and it is wrong for the JSON one. `build_json_body` writes a serialized document and a newline (`body.rs:46`), and on this path the proto text is already a serialized JSON document (`proto_convert.rs:630`). The RowBinary half is correct, because `binary.rs:52` needs a parsed object.

The result is silent loss. `JsonEachRow` is `#[default]` (`clickhouse_sink/src/lib.rs:60`), the skip still returns `Ok` (`sink.rs:146`), and the runtime adds the whole run to `iggy_connector_messages_processed_total` while `iggy_connector_errors_total` stays flat. `AutoCommitWhen::PollingMessages` (`runtime/src/sink.rs:515`) commits the offset before the plugin sees the batch, so nothing replays. On master these rows loaded, because the `Schema::Json` tag reparsed the text. RowBinary also loses a signal it had: that path failed the batch and counted an error before, and now drops the row quietly.

This does not need a missing descriptor. `encode_json_with_schema` returns `Err` for any non-object top-level JSON (`proto_convert.rs:285`), so a fully configured pipeline takes the same fallback on every array, string, number, boolean and null message.

Fix: parse `Payload::Proto(text)` as JSON in both builders, then take the existing paths. Parse and re-serialize rather than pass the text through, because `pretty_json` (`proto_convert.rs:48`) puts newlines in it and JSONEachRow needs one object per line.

core/connectors/sinks/doris_sink/src/lib.rs
```diff
@@ -1098,8 +1098,9 @@ impl Sink for DorisSink {
         //
         // The lone hard-abort is a non-JSON payload (via `?`): a stream-wide
         // schema-contract violation, not a transient chunk failure. Under the
-        // documented `schema = "json"` config the SDK drops non-JSON before
-        // consume() is called, so this stands as a defensive guard.
+        // documented `schema = "json"` config the runtime's JSON decoder drops
+        // non-JSON bytes before consume() is called, so this stands as a
+        // defensive guard unless a format-converting transform is configured.
```
mlevkov, line 1103: This edit names the hole and leaves it open. `proto_convert` is a format-converting transform, and with no descriptor it returns `Payload::Proto` holding a JSON document (`proto_convert.rs:656`).

Doris then takes the `_` arm at `:1109`, and the `?` at `:1117` ends the whole `consume()` call, so no chunk is written. The offset already committed (`runtime/src/sink.rs:515`), and the runtime has no replay and no dead-letter queue, so the batch is gone and the loss repeats every poll. On master the same configuration loaded rows, because the `Schema::Json` tag reparsed the text. The config is not hypothetical: this PR adds one at `proto_text_config/elasticsearch_sink.toml:41`.

Delta (`delta_sink/src/sink.rs:110`) and Iceberg (`iceberg_sink/src/router/mod.rs:152`) have the same shape. Elasticsearch, Meilisearch and the ClickHouse passthrough already take this payload, so three sinks accept it and three fail.

Fix: parse `Payload::Proto` as JSON in Doris, Delta and Iceberg, and keep the abort when the parse fails. Do not treat all proto text as JSON: `proto_convert.rs:695` and `:714` produce genuine proto text that is not JSON.

core/connectors/runtime/src/sink.rs
```diff
@@ -674,6 +680,10 @@ async fn process_messages(
             continue;
         };
 
+        // Read the tag off the payload before `try_into_vec` consumes it. The
+        // decoder's own schema names the format it reads, not the variant it
+        // returned, and a transform may have changed the variant since.
+        let schema = message.payload.schema();
```
mlevkov, line 686: Three of the six stream `schema` values change the tag here, and the Compatibility section names one. Under their defaults, `avro` (`avro.rs:40`), `flat_buffer` (`flatbuffer.rs:42`) and `proto` (`proto.rs:48`) all decode to `Payload::Json`, so all three change. `json`, `text` and `raw` do not, which the section states correctly.

For `flat_buffer` this is the same output change in the same three sinks the section already describes for avro: S3, HTTP and SurrealDB stop base64-encoding and write the document. For `proto` the effect differs, and one case improves, because a `proto` stream failed every poll into Doris before and now writes rows.

Fix: add `flat_buffer` and `proto` to the Compatibility section.

core/connectors/sinks/README.md
```diff
@@ -249,6 +249,8 @@ for message in messages {
 
 While the schema of messages (that will be consumed from the Iggy stream), cannot be controlled by the sink connector itself, the built-in configuration allows to decide what's the expected format of the messages (the particular `StreamDecoder` will be used).
 
+`messages_metadata.schema` names the variant each `Payload` in the batch actually holds, which is not always the stream's `schema` setting. A decoder may return a different form than the wire format it reads: with `schema = "avro"` the Avro decoder extracts to JSON by default, so the batch arrives as `Payload::Json`. A configured transform can change the variant again. Match on the `Payload` itself and treat `messages_metadata.schema` as a description of what arrived, not of how the stream was configured.
```
mlevkov, line 252: One poll can now reach a plugin as several `consume()` calls, and this paragraph does not say so. `runtime/src/sink.rs:761` loops over the runs and `:764` gives every run the same `current_offset`, so a sink that commits per call applies part of a poll. Delta commits a table version per call, and Iceberg commits per call.

Nothing in the repository breaks today, because no sink keys its output on `current_offset`. It is still a new contract for plugin authors.

One exception is worth a sentence: a fully dropped batch is tagged with the decoder wire format (`runtime/src/sink.rs:748`), so an avro stream's tag flips between `avro` and `json`.

Fix: state that a mixed batch arrives as several `consume()` calls that repeat one offset, and note the empty-batch case.

core/connectors/sdk/src/sink.rs
```diff
@@ -196,7 +198,12 @@ impl<T: Sink + std::fmt::Debug> SinkContainer<T> {
                     }
                 };
 
-                let payload = match messages_metadata.schema.try_into_payload(message.payload) {
+                // The runtime tags each run from `Payload::schema`, so the
+                // tag names a variant here rather than a wire format.
+                let payload = match Payload::try_from_schema(
```
mlevkov, line 203: Two things to record against this line.

The meaning of `Schema::Proto` in `MessagesMetadata` changed and `iggy_connector_sdk` stays at 0.4.0, so a plugin built before this change still calls `Schema::try_into_payload` and rebuilds a Proto run as `Payload::Raw`. A version bump alone does not repair that, because a rebuilt Doris, Delta or ClickHouse sink still rejects the `Payload::Proto` it now receives. A runtime check is not available either: `iggy_sink_version` (`sdk/src/sink.rs:328`) expands `CARGO_PKG_VERSION` inside the plugin crate, so it reports the plugin version, and the runtime only stores the string.

This line also adds a per-message JSON parse on avro and flatbuffer streams. The old `Schema::Avro` arm moved the bytes without reading them, and the new `Schema::Json` arm runs `simd_json::to_owned_value` per message. That is the price of the fix and not a defect, but operators sizing sink CPU need to know.

Fix: bump to 0.5.0, and document both the rebuild requirement and the new parse.

core/connectors/sdk/README.md
```diff
@@ -175,7 +177,7 @@ These are SDK configuration fields, not Random or Stdout `plugin_config` keys. T
 #### Transform Options
 
 - **`proto_convert`**: Transform for converting between protobuf and other formats
-- **`source_format`** / **`target_format`**: Formats to convert between - any schema value (`json`, `raw`, `text`, `proto`, `flat_buffer`, `avro`)
+- **`source_format`** / **`target_format`**: Formats to convert between - any schema value (`json`, `raw`, `text`, `proto`, `flat_buffer`, `avro`). `source_format` must match the variant the decoder or a preceding transform actually produced, or the message is rejected
```
mlevkov, line 180: Two problems in this sentence.

`proto_convert` does not reject a mismatched `source_format`. `ProtoConvert::transform` (`proto_convert.rs:737`) dispatches on the format pair with no up-front guard, and the example at `README.md:128` depends on that: `schema = "raw"` with `source_format = "proto"` reaches `protobuf_to_json`, whose `Payload::Raw` arm always returns `Ok`. `flatbuffer_convert.rs:114` is the only transform that has the guard, and `avro_convert` has none either.

"A preceding transform" also implies an order the runtime does not guarantee. `runtime/src/transform.rs:26` builds the chain from a `HashMap`, so the order varies per process.

Fix: limit the claim to `flatbuffer_convert`, and either drop the ordering wording or state that the chain order is not defined.

core/connectors/sinks/elasticsearch_sink/src/lib.rs
```diff
@@ -381,7 +381,9 @@ impl Sink for ElasticsearchSink {
                         }
                     }
                 }
-                Payload::Text(text) => simd_json::json!({
+                // `Payload::Proto` holds proto text, so it indexes the
+                // same way `Payload::Text` does.
+                Payload::Text(text) | Payload::Proto(text) => simd_json::json!({
```
mlevkov, line 386: This arm indexes proto text as one opaque `text` field. On a `schema = "json"` stream with a descriptor-less `proto_convert`, the old tag reparsed the same bytes and Elasticsearch indexed a structured document, so for that pipeline this is a searchability regression rather than a new capability.

Fix: try a JSON parse on `Payload::Proto` first, and fall back to this arm when it fails.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -360,7 +360,9 @@ impl MeilisearchSink {
                     ]),
                 }
             }
-            Payload::Text(text) => Map::from_iter([
+            // `Payload::Proto` holds proto text, so it indexes the same
+            // way `Payload::Text` does.
+            Payload::Text(text) | Payload::Proto(text) => Map::from_iter([
```
mlevkov, line 365: Same shape as the Elasticsearch arm: proto text is stored under a single `text` key, so the document loses every original field. The same descriptor-less `proto_convert` pipeline stored a structured document before.

Fix: try a JSON parse on `Payload::Proto` first, and fall back to this arm when it fails.

core/integration/tests/connectors/elasticsearch/proto_text.rs
```diff
@@ -0,0 +1,110 @@
...
+        .expect("Failed to search documents");
+
+    assert_eq!(
+        search_result.hits.total.value, TEST_MESSAGE_COUNT,
+        "Expected {TEST_MESSAGE_COUNT} documents in Elasticsearch"
+    );
+
+    for hit in &search_result.hits.hits {
+        assert_eq!(
+            hit.source.get("data_type").and_then(|v| v.as_str()),
```
mlevkov, line 92: This assertion locks in the flattened result as the expected one. It asserts `data_type == "text"`, which is the shape the document takes only after the structured fields are gone, so the test passes precisely when the regression happens.

Fix: assert that the indexed document keeps the original message fields.

core/connectors/runtime/src/sink.rs
```diff
@@ -721,43 +757,50 @@ async fn process_messages(
...
+            messages_meta.as_ptr(),
+            messages_meta.len(),
+            messages.as_ptr(),
+            messages.len(),
         );
-        metrics.inc_errors_with_labels(&labels.counter);
-        0
-    };
+        ffi_elapsed += ffi_start.elapsed();
+        if result != 0 {
```
mlevkov, line 795: The new per-run subtraction has no coverage. No test in this module fails a run, so `processed_count -= run_len` never executes, and a wrong sign or a wrong operand would not fail the suite.

Fix: add a stub sink that fails one run of a split batch, and assert the resulting `processed_count`.

core/connectors/sdk/src/lib.rs
```diff
@@ -471,3 +522,103 @@ pub enum Error {
...
+    #[test]
+    fn given_a_payload_when_round_tripped_through_its_own_schema_should_keep_the_variant() {
+        for (payload, schema) in all_payloads() {
+            let bytes = payload
+                .try_into_vec()
+                .unwrap_or_else(|error| panic!("failed to serialize {schema} payload: {error}"));
+            let rebuilt = Payload::try_from_schema(schema, bytes)
+                .unwrap_or_else(|error| panic!("failed to rebuild {schema} payload: {error}"));
+
+            assert_eq!(
```
mlevkov, line 569: This round-trip test compares only the rebuilt variant, so a `try_from_schema` that drops or corrupts the value still passes. The test proves the tag survives, not the payload.

Fix: compare the rebuilt payload bytes against the original bytes.

core/connectors/runtime/src/sink.rs
```diff
@@ -721,43 +757,50 @@ async fn process_messages(
...
     })?;
 
-    let messages_meta = postcard::to_allocvec(&messages_metadata).map_err(|error| {
-        error!(
-            "Failed to serialize messages metadata for sink connector with ID: {plugin_id}. {error}"
-        );
-        RuntimeError::FailedToSerializeMessagesMetadata
-    })?;
+    let mut ffi_elapsed = Duration::ZERO;
+    for (schema, run) in runs {
```
mlevkov, line 761: Nothing records that a batch split. `stage_ffi` (`:392`) records total duration, and the per-run log at `:797` fires only on failure, so an operator cannot see one poll become N calls. A batch that alternates variants makes one FFI call and one sink round trip per message.

Fix: record the run count per batch as a metric. A log line on every split gets noisy once a descriptor-backed `proto_convert` splits most batches, so the metric is the useful half.

core/connectors/sinks/clickhouse_sink/src/body.rs
MarcusKainth: I agree on the JSON builder, of all three this was the loudest. Both builders now parse `Payload::Proto` and take the existing path on success. Text that is not JSON is skipped, with an accompanying log line, so the existing CSV and TSV passthrough cases from `raw_to_protobuf` are unaffected. One nuance: `pretty_json` defaults to false in `proto_convert.rs` `L75`, so the default fallback text is already one line. I have still gone with parse and re-serialise as you suggested, since it also validates the document before ClickHouse sees it.

One correction on RowBinary. It did not lose a failure signal. `build_row_binary_body` on `master` is byte-identical to this branch, and it never failed on a variant mismatch. On `master` the reparsed JSON reached `serialize_row` and was inserted, or failed the batch only on a table mismatch. The regression was data loss rather than a lost error, and the fix is the same either way.

One item I want to flag which matches the behaviour on `master`. The string passthrough writes text verbatim, so a pipeline configured with `pretty_json = true` and `string_format = "json_each_row"` produces a multi-line JSON document which ClickHouse will reject. A `text` stream carrying pretty JSON has the same exposure on `master`; the `Payload::Proto` arm added in this PR makes it reachable from a `proto_convert` pipeline. I have added documentation in the `README.md` and left both arms as is, unless you'd rather I re-serialise `Payload::Text`? I decided against this because it rewrites operator bytes on master's path. Let me know if you want this resolved within this PR, and if so, how.

Fixed in `30216eb`. Unit tests cover both builders and the string arm, and a new `clickhouse/proto_text` integration test on the default JSONEachRow format loads zero rows and counts zero errors on `d8a8ece`. The `master` vs `HEAD` matrix is in the summary comment.

core/connectors/sinks/doris_sink/src/lib.rs
MarcusKainth: I agree, the comment edit named the hole and then left it open. Doris now parses proto text per chunk and keeps the abort for text that is not JSON, so the schema-contract guard still fires for the `text_to_protobuf` and `raw_to_protobuf` outputs in `proto_convert.rs` `L695` and `L714`. Delta takes the same parse and keeps its batch error.

One correction on Iceberg. It is not the same shape as Doris and Delta. `router/mod.rs` `L149-169` is a `filter_map` with a `warn!`, so it silently dropped the non-JSON rows and only failed when every row was non-JSON, which is worse, and there is a second site in `router/dynamic_router.rs` `L85-96` where the route field returns `None`. Both now parse proto text. The silent-drop shape itself predates this PR and I have left it as is.

Fixed in `30216eb`, with unit tests per sink and a `doris/proto_text` integration test that loads zero rows on `d8a8ece`. Two things worth having on record. No decoder produces `Payload::Proto` (`decoders/proto.rs` only emits `Json` and `Raw`), so every one of these rows is transform-only exposure. And `master` had its own silent loss on the same configuration once a descriptor is set: `json_to_protobuf` returns binary `Payload::Raw` at `proto_convert.rs` `L646`, the `json` tag sent it through `simd_json::to_owned_value`, and the SDK dropped it before `consume()` ran (`sdk/src/sink.rs` `L203-214`). This branch tags it `Raw` and delivers it. That row is in the matrix too.

core/connectors/runtime/src/sink.rs
MarcusKainth: Correct on both. `flat_buffer` (`flatbuffer.rs` `L42`) and `proto` (`proto.rs` `L48`) extract to JSON under the runtime's defaults the same way `avro` does, so all three change tag, and the section now says so, including that a `proto` stream into Doris improves. I have added the non-object trigger as well: `encode_json_with_schema` returns `Err` for any top-level array or scalar at `proto_convert.rs` `L285`, so a fully configured pipeline takes the fallback for those messages too. The section is rewritten in the PR body.

core/connectors/sinks/README.md
MarcusKainth: Added. The paragraph now states three things. A poll can arrive as several `consume()` calls, one per contiguous variant run, and every call repeats the same `current_offset`. That offset is the partition's high-water offset from the poll (`sink.rs` `L344`) and was batch-scoped on `master` too, so repeating it per run is the correct contract rather than a new one, though a sink keying output on it has to tolerate the repeat. And an empty batch keeps the stream's configured schema. `e58b1cf`.

core/connectors/sdk/src/sink.rs
MarcusKainth: I agree, and this is the finding I most wanted a second pair of eyes on. No FFI signature changed and no variant was added, but the `Proto` discriminant changed meaning. The skew is narrower than it first looks and worth stating precisely: a plugin built against 0.4.0 under this runtime is better off on `avro`, `flat_buffer` and `proto` streams, because it now receives a `json` tag it can read, and worse off on exactly one pipeline, a `json` stream with `proto_convert`, where it rebuilds the `proto`-tagged run as `Payload::Raw` where `master` handed it `Payload::Json`. That one pipeline is enough to want the boundary marked. Bumped to 0.5.0 in `1742d52`. The rebuild requirement is stated in the PR body's Compatibility section, and the connector-sdk skill now lists a `Schema` variant that changes meaning alongside a new variant as a breaking change.

On the parse, it is real and now documented, but it comes from the runtime tag rather than the SDK swap. The `Json` arm of `Payload::try_from_schema` is character-identical to `Schema::try_into_payload`'s (`lib.rs` `L189-191` and `L296-298`); the arm that runs changed because the tag did. On `master` the runtime paid `simd_json::to_vec` and then handed the plugin a `Payload::Avro` holding JSON bytes it could not use, so the cost was only avoided by being broken.

core/connectors/sdk/README.md
MarcusKainth: Both right. Only `flatbuffer_convert` has the guard (`flatbuffer_convert.rs` `L114`), and the README's own example at `L128` relies on the absence of one in `proto_convert`. I have limited the sentence to `flatbuffer_convert` and added a line that chain order is not defined, because `TransformsConfig` is a `HashMap` (`configs/connectors.rs` `L250-254`). Changing that to an ordered map is a separate behaviour change and I think it deserves its own issue rather than folding it in here. Happy to raise it if you agree. `e58b1cf`.

core/connectors/sinks/elasticsearch_sink/src/lib.rs
MarcusKainth: I agree, and the arm was also inconsistent with its `Raw` sibling three lines up, which already parses first (`elasticsearch_sink/src/lib.rs` `L372-383`). Both sinks now try the parse and only fall back to the text arm for proto text that is not JSON. Quickwit had the same flattening before this PR (`quickwit_sink/src/lib.rs` `L346-349`), so I have fixed it in its own commit, `c858418`, for consistency. Easy to drop if you would rather keep this PR narrower. `30216eb` for the two you raised.

core/connectors/sinks/meilisearch_sink/src/lib.rs
MarcusKainth: Same as the Elasticsearch arm, fixed in the same commit (`30216eb`): the parse runs first and the text arm is only the fallback for proto text that is not JSON.

core/integration/tests/connectors/elasticsearch/proto_text.rs
MarcusKainth: Yes, it asserted the regression. I have rewritten it to assert the original fields are indexed and no `data_type` wrapper exists. The useful property: it passes on `master`, fails on `d8a8ece`, and passes again after `30216eb`.

core/connectors/runtime/src/sink.rs
MarcusKainth: Added two tests through a stub that fails whichever run's schema the test registers. One failing run in a split batch debits only that run and the later runs still execute, and every run failing reports zero processed with one error per run. The subtraction is gone as well; `processed_count` now accumulates in the success branch, so the underflow class cannot come back. `2dbebc4`.

core/connectors/sdk/src/lib.rs
MarcusKainth: Correct, it proved the tag and not the payload. It now compares `try_to_bytes()` of the rebuilt payload against the original for all fourteen rows (`30216eb`). I have kept the two inverses separate rather than delegating five arms, because the docs frame them as inverses of two different tags and the byte-level test is now the drift guard. The `Any` test also asserts `type_url` and the base64 `value` rather than key presence. `d94bedd`.

core/connectors/runtime/src/sink.rs
MarcusKainth: Added `iggy_connector_sink_runs`, a counter incremented by the run count of each batch. `iggy_connector_stage_duration_seconds_count{stage="total"}` already counts batches, so runs per batch is the ratio of the two, and a histogram would have cost eleven series per sink key for the same answer. The `ffi` stage sample stays one per batch with the summed duration, which is what you described and is identical to `master` for a uniform batch; the README now says so next to the counter. The benchmark event carries a `runs` field too. `2dbebc4`.

MarcusKainth:
Thanks for the depth here. All twelve threads are addressed and the per-thread replies cite the commits. Three things I want on record at the top level.

First, the evidence. Rather than argue the before and after, I ran the same test files against a `master` worktree, `d8a8ece`, and this branch.

| test | `master` | `d8a8ece` | this branch |
|---|---|---|---|
| **Fix proof: fails on `master`, passes here** | | | |
| `given_a_proto_tagged_run_when_the_container_consumes_should_hand_the_sink_a_proto_payload` | FAIL | PASS | PASS |
| `given_an_avro_stream_when_the_sink_consumes_should_receive_json_payloads` | FAIL | PASS | PASS |
| `proto_payloads_that_are_not_json_are_stored_as_text` | FAIL | PASS | PASS |
| `string_body_proto_payload_is_written_as_text` | FAIL | PASS | PASS |
| **Blockers demonstrated: passed on `master`, broke on `d8a8ece`, pass here** | | | |
| `given_a_proto_convert_transform_when_the_sink_consumes_should_index_the_document` | PASS | FAIL | PASS |
| `given_a_proto_convert_transform_when_the_sink_consumes_should_store_the_rows` | PASS | FAIL | PASS |
| `given_proto_text_messages_should_store` | PASS | FAIL | PASS |
| **Round-two gaps closed: failed on both earlier refs, pass here** | | | |
| `given_auto_payload_proto_holding_json_should_store_queryable_json` | FAIL | FAIL | PASS |
| `given_proto_payload_holding_json_should_send_the_document` | FAIL | FAIL | PASS |
| `given_proto_text_holding_json_when_extracted_should_preserve_the_document` | FAIL | FAIL | PASS |
| `given_proto_text_payloads_holding_json_should_write_them_as_rows` | FAIL | FAIL | PASS |
| `given_proto_text_that_is_not_json_should_send_it_as_a_string` | FAIL | FAIL | PASS |
| `json_body_pretty_proto_json_is_written_as_one_line` | FAIL | FAIL | PASS |
| `json_body_proto_payload_holding_json_is_written` | FAIL | FAIL | PASS |
| `proto_payload_holding_json_is_written_as_a_document` | FAIL | FAIL | PASS |
| `proto_payloads_holding_json_are_indexed_as_documents` | FAIL | FAIL | PASS |
| `row_binary_body_proto_payload_holding_json_writes_bytes` | FAIL | FAIL | PASS |
| **Unchanged behaviour: pass everywhere** | | | |
| `given_a_json_tagged_run_when_the_container_consumes_should_hand_the_sink_a_json_payload` | PASS | PASS | PASS |
| `given_a_raw_tagged_run_when_the_container_consumes_should_hand_the_sink_a_raw_payload` | PASS | PASS | PASS |
| `given_a_text_tagged_run_when_the_container_consumes_should_hand_the_sink_a_text_payload` | PASS | PASS | PASS |
| `given_auto_payload_proto_text_should_store_text` | PASS | PASS | PASS |
| `given_bytes_that_are_not_an_any_when_rebuilt_from_a_wire_tag_should_fall_back_to_raw` | PASS | PASS | PASS |
| `given_only_proto_text_that_is_not_json_should_fail_with_invalid_payload_type` | PASS | PASS | PASS |
| `given_protobuf_wire_bytes_when_rebuilt_from_a_wire_tag_should_stay_an_any_document` | PASS | PASS | PASS |
| `json_body_non_json_proto_payload_is_skipped` | PASS | PASS | PASS |
| `proto_text_that_is_not_json_is_written_as_a_string` | PASS | PASS | PASS |
| `row_binary_body_non_json_proto_payload_is_skipped` | PASS | PASS | PASS |

The rows that fail on `master` and pass here are the fix. Two of them fail on `master` only because it had no `Payload::Proto` arm at all, in Meilisearch and the ClickHouse string passthrough, so they are there for completeness rather than as proof. The middle group is the three blockers demonstrated end to end: ClickHouse, Doris and Elasticsearch each pass on `master`, fail on `d8a8ece`, and pass again now. The rows that pass on all three refs are the streams and paths this PR claims not to touch.

Building that matrix surfaced two additions to the Compatibility section. A `text` or `raw` stream with `proto_convert` now hands the sink `Payload::Proto` where `master` handed it `Text` or `Raw`; every sink now treats proto text that is not JSON the way it treats `Text`, including HTTP, which used to base64-encode it. And a descriptor-backed `proto_convert` on a `json` stream was dropped by the SDK on `master`, because the `json` tag sent protobuf bytes through the JSON parser; it now arrives as `Raw`.

A second pass on my side after the fixes turned up the same defect in three more sinks and one metric detail, so for completeness:

- S3, SurrealDB's auto format and HTTP wrote proto text holding JSON as a string or base64 where `master` wrote the document. All three now read it as the document (`c1bbb4c`); HTTP sends proto text that is not JSON as a plain string, the way it sends `Text`.
- `iggy_connector_sink_runs` increments after the calls are made rather than before, so it counts calls that happened (`2dbebc4`).
- RowBinary is unchanged on purpose: non-object proto text fails the batch, which is what the same array did on `master` when it arrived as JSON, and whole-batch failure on a bad row is the documented contract.
- A run that holds no document now fails in Iceberg where a mixed batch used to skip those messages beside their siblings. They were dropped either way; now they are counted as an error.
- No warn log on a split batch. A threshold log would fire on every batch of a descriptor-backed `proto_convert`, and the counter carries the signal.

Second, four attributions I would correct. RowBinary never had a failure signal to lose; the builder on `master` is byte-identical and the regression was loss rather than silence. Iceberg is not the same shape as Doris and Delta: it silently drops via `filter_map` and only errors when every row is non-JSON, which is worse, and it has a second site in the dynamic router. Both are fixed. The `unreachable!()` in `flatbuffer_convert.rs` is on `master` at `L172` verbatim; this PR deleted a helper and left the panic where it was. I agree it should be an `Err`, so I have made it one. The SKILL.md item 12 pointer was already stale on `master` as item 10, and I have fixed it as a drive-by.

Third, the SDK bump. Agreed and done, 0.4.0 to 0.5.0, with the rebuild requirement in the PR body, and the skill now treats a `Schema` variant changing meaning as breaking. SurrealDB's `iggy_schema` column is covered in its README: it now records the variant received, so avro streams write `json` next to a real document where they previously wrote `avro` next to base64 of JSON text. The http sink comment is updated to per-run wording. The pre-existing items you listed (the encoder per message in `avro_convert`, the dropped serialisation in `proto_convert`, the empty `Payload::Raw`, and the per-Raw clone in Elasticsearch) are all real. The Elasticsearch clone is needed because simd_json mutates its input even on failure; the rest I have left for a follow-up so this PR stays a fix.

core/connectors/sdk/src/lib.rs
```diff
@@ -157,6 +158,72 @@ pub enum Payload {
 }
 
 impl Payload {
+    /// The `Schema` describing this payload's variant.
```
hubcio, line 167: nit: the run contract - several `consume()` calls per poll, each repeating the same `current_offset` - lives only in the sinks README. put it on `Sink::consume` (line 138) and `MessagesMetadata` (line 376), where plugin authors read.

core/connectors/sdk/README.md
```diff
@@ -156,12 +156,16 @@ Create the stream and topic, then start the runtime from the repository root:
 IGGY_CONNECTORS_CONFIG_PATH=connectors.toml ./target/release/iggy-connectors
 ```
 
-The source sends 100 records, then continues polling without new messages. Stdout logs message offsets and the serialized JSON envelope bytes, containing `type_url` and base64 `value`. The sink's `raw` schema also determines how the plugin receives those transformed bytes.
+The source sends 100 records, then continues polling without new messages. Stdout logs message offsets and the serialized JSON envelope bytes, containing `type_url` and base64 `value`. The batch handed to the plugin is tagged with the payload's own schema rather than the stream's configured `raw`: `proto_convert` leaves a `Payload::Json`, so the plugin receives a `json` batch.
```
hubcio, line 161: nit: this tag change breaks plugins built against 0.4.0, but the breaking-change list at line 57 still says two items. add it there with the rebuild rule.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -360,7 +367,7 @@ impl MeilisearchSink {
                     ]),
                 }
             }
-            Payload::Text(text) => Map::from_iter([
+            Payload::Text(text) | Payload::Proto(text) => Map::from_iter([
```
hubcio, line 364: nit: the README behavior paragraph (line 59) still says proto is skipped as unsupported. say that proto text holding JSON indexes as a document and other proto text lands in `text`.

core/connectors/sinks/meilisearch_sink/src/lib.rs
```diff
@@ -339,6 +339,13 @@ impl MeilisearchSink {
             payload,
         } = message;
 
+        // The descriptor-less `proto_convert` fallback puts a JSON document in
+        // `Payload::Proto`. It is indexed as that document, and the text arm
+        // below is kept for proto text that is not JSON.
+        let payload = match payload.json_document() {
```
hubcio: simplification: this four-line normalizer is copied into five sinks. add one consuming helper on `Payload` next to `json_document` and call it at each site.

also at elasticsearch_sink/src/lib.rs:322, http_sink/src/lib.rs:386, quickwit_sink/src/lib.rs:320, surrealdb_sink/src/lib.rs:901.

core/connectors/sinks/iceberg_sink/src/router/dynamic_router.rs
```diff
@@ -83,19 +83,25 @@ impl DynamicRouter {
...
+        route_value(&message.payload, &self.route_field)
     }
 }
 
+/// The routing value a payload carries under `route_field`, read from the JSON
+/// document it holds. Proto text holding JSON is the descriptor-less
+/// `proto_convert` fallback and routes like the document it holds; a payload
+/// with no document does not route.
+fn route_value(payload: &Payload, route_field: &str) -> Option<String> {
+    let Some(document) = payload.json_document() else {
```
hubcio, line 95: nit: `route_value` parses the proto text and drops the document, then `write_data_files` parses the same text again at router/mod.rs:156. normalize the payload to `Payload::Json` once at the top of the `route_data` loop.

core/connectors/runtime/src/sink.rs
```diff
@@ -721,53 +761,556 @@ async fn process_messages(
...
+
+        assert_eq!(batches.len(), 1);
+        assert_eq!(batches[0].metadata_schema, Schema::Avro);
+        assert_eq!(batches[0].messages_schema, Schema::Avro);
+    }
+
+    #[tokio::test]
+    async fn given_mixed_payload_variants_when_batch_is_processed_should_split_into_runs() {
+        let plugin_id = next_plugin_id();
+        let transform = Arc::new(RetaggingTransform {
```
hubcio: simplification: this is the same four-payload transform `split_batch()` builds at line 1068. call the helper like the other two tests do.

hubcio (changes requested):
for history: #3669 attempted the same fix in july and closed stale without a verdict, worth a link in the description.

overall i like what you did here. just nits below, once you fix i think we can merge

core/connectors/sdk/src/lib.rs
MarcusKainth: `Sink::consume` `L138` now states the run split, the repeated `current_offset`, and the empty batch that still arrives as one call carrying no messages. `MessagesMetadata` `L392` documents all three fields, with `current_offset` as the partition's high-water offset at the time of the poll and `schema` as the variant the payloads hold rather than the stream's configured format. `sinks/README.md` `L252-258` keeps the long form. A new runtime test pins the empty-batch call, since the rustdoc now promises it. `093f5eb`.

core/connectors/sdk/README.md
MarcusKainth: The lead-in at `sdk/README.md` `L57` now reads as two removals and one field changing meaning, and a paragraph at `L66` records that `MessagesMetadata.schema` and `RawMessages.schema` name the variant the `Payload` holds rather than the wire format the stream's decoder reads. `Schema::Proto` means a `Payload::Proto` string on the sink path where it meant protobuf wire bytes before, so a plugin built against 0.4.0 rebuilds the run as the wrong variant with no error anywhere. The rebuild rule stays where it was, in the lead-in. `093f5eb`.

core/connectors/sinks/meilisearch_sink/src/lib.rs
MarcusKainth: The `## Behavior` paragraph now says proto text holding a JSON document is indexed as that document, and proto text that is not JSON is indexed in the `text` field alongside text payloads. That leaves Avro and FlatBuffer as the schemas skipped with a warning, which matches `lib.rs` `L368-373`. `093f5eb`.

core/connectors/sinks/meilisearch_sink/src/lib.rs
MarcusKainth: `Payload::into_json_document` sits next to `json_document` in `sdk/src/lib.rs` `L241`, and the five sites call it: Meilisearch `L342`, Elasticsearch `L317`, HTTP `L381`, Quickwit `L316`, SurrealDB `L896`. Behaviour is unchanged at every site and the existing per-sink tests cover it. Four SDK unit tests pin proto text holding JSON, proto text that is not JSON with the text left byte-identical, a JSON payload returned as it was, and every other variant returned with both its schema and its bytes intact. `535214c`.

core/connectors/sinks/iceberg_sink/src/router/dynamic_router.rs
MarcusKainth: `route_data` normalises each message at the top of the loop, so `route_value` and `write_data_files` both read the document off a `Payload::Json`, and proto text holding JSON is parsed once on the dynamic path rather than twice. `route_value`'s doc at `L90-93` is updated to match. The `json_document` call at `router/mod.rs` `L156` serves `StaticRouter::route_data`, which reaches `write_data_files` with no routing parse and so parses once either way. The three `route_value` tests at `L183-202`, `router/mod.rs` `L448`, and the `connectors::iceberg` suite pass unchanged. `535214c`.

core/connectors/runtime/src/sink.rs
MarcusKainth: The test calls `split_batch()`. `e587e4d`.

MarcusKainth:
The description now links #3669.

All six threads are addressed in `535214c`, `e587e4d` and `093f5eb`. `origin/master` is merged in at `233211c` and the verification chain is green on the merged tree: fmt, sort, workspace clippy with all features and all targets, taplo, markdownlint, license headers, whitespace and newline, typos, unit tests for the SDK, the runtime and every touched sink, and the `schema_tagging`, Meilisearch, Elasticsearch, Quickwit, HTTP, SurrealDB and Iceberg integration suites against a real server, runtime and containers.

hubcio (approved):

mmodzelewski (approved):

Issue #3545, opened 2026-06-23
Kafka gateway: Kafka transactional APIs — UNSUPPORTED_VERSION stub and idempotence docs


ryerraguntla:
Part of #3560.

## Summary

Phase 3: handle transactional producer APIs with explicit unsupported responses until Iggy supports transactional writes.

## Motivation

Iggy does not support transactional writes today. Workaround: return `UNSUPPORTED_VERSION` (35) for InitProducerId (22), AddPartitionsToTxn (24), AddOffsetsToTxn (25), EndTxn (26), TxnOffsetCommit (28).

Document that Kafka 3.x idempotent producers fall back to at-least-once.

## Depends on

- #3539 — Phase 1 (minimum)

## Scope

- [ ] Handlers return UNSUPPORTED_VERSION or documented stub behavior
- [ ] README section: transactions not supported; idempotent producer behavior
- [ ] Tests assert error codes, no panic

## References

- Rollout plan: https://github.com/apache/iggy/discussions/3253#discussioncomment-17181030

numinnex:
I am picking this one up

Issue #3540, opened 2026-06-23
Kafka gateway: Design Consumer offset persistence strategy for Kafka gateway Phase 2


ryerraguntla:
Part of #3560.

## Summary

Decide and document how the Kafka gateway stores consumer group offsets for Phase 2.

## Motivation

Open question from https://github.com/apache/iggy/discussions/3253#discussioncomment-17181030:

| Option | Description |
|--------|-------------|
| **A (recommended)** | Iggy-backed `__consumer_offsets` internal topic |
| **B** | SQLite file on bridge host |
| **C** | In-memory only (resets on restart) |

Phase 2 is blocked until this decision is recorded.

## Depends on

- #3539 — Phase 1 complete (or near-complete)

## Deliverables

- [ ] ADR or design comment on #3253 with chosen option
- [ ] Data model: key = `(group, topic, partition)` → committed offset
- [ ] Durability, backup, multi-bridge-instance behavior documented
- [ ] Partition assignment strategy decision: range / round-robin / sticky vs Iggy balanced

## Acceptance criteria

- [ ] Maintainers sign off on Option A/B/C
- [ ] Phase 2 implementation issues unblocked

numinnex:
I am picking this one up

Issue #3536, opened 2026-06-23
Kafka gateway: Kafka Fetch API (key 1) — bridge to Iggy offset-based polling


ryerraguntla:
Part of #3560.

## Summary

Wire Kafka **Fetch (API key 1)** to `IggyBridge::fetch()` — return messages from Iggy by partition offset without consumer groups (Phase 1 assign/manual offset).

## Motivation

Enables simple consumers and `kafka-console-consumer.sh` without `--group` for Phase 1 MVP.

## Depends on

- #3421
- #3533 — IggyBridge core
- #3535 — Produce path (for E2E data)

## Scope

- [ ] Decode Fetch v4–v12 (scoped range)
- [ ] Map fetch offset + max bytes → Iggy poll/fetch by offset
- [ ] Encode RecordBatch response (opaque bytes from Iggy payload)
- [ ] High-watermark / log end offset in response
- [ ] Integration test: produce then fetch same partition/offset

## Acceptance criteria

- [ ] `kafka-console-consumer.sh --from-beginning --partition 0` works **without** consumer group (Phase 1)
- [ ] Offset progression matches Iggy partition offsets

## References

- Rollout plan: https://github.com/apache/iggy/discussions/3253#discussioncomment-17181030

krishvishal:
I am picking this up.

Issue #3535, opened 2026-06-23
Kafka gateway: Kafka Produce API (key 0) — bridge to Iggy send_messages


ryerraguntla:
Part of #3560.

## Summary

Wire Kafka **Produce (API key 0)** to `IggyBridge::produce()` — decode RecordBatch payloads and send to Iggy with balanced partitioning.

## Motivation

Core write path for Kafka migration: existing producers must work without code changes.

## Depends on

- #3421
- #3533 — IggyBridge core

## Scope

- [ ] Decode Produce v3–v9 (scoped range); preserve RecordBatch as opaque bytes where possible (hot path)
- [ ] Map Kafka topic + partition → Iggy stream/topic/partition
- [ ] Partitioning: balanced / keyed per Kafka record (document semantics)
- [ ] Produce response with correct base offset / error per partition
- [ ] KAFKA-18659: ApiVersions advertises Produce min=0 for librdkafka compatibility (encode layer)
- [ ] Integration test: produce via gateway → verify message in Iggy (SDK or CLI)

## Acceptance criteria

- [ ] `kafka-console-producer.sh` or `kcat -P` can send to a topic through the gateway
- [ ] Messages visible in Iggy stream/topic
- [ ] At-least-once documented (Phase 1)

## References

- Rollout plan: https://github.com/apache/iggy/discussions/3253#discussioncomment-17181030

krishvishal:
I'm picking this up.

PR #4205, opened 2026-09-16
docs(gateways): settle the Kafka record mapping, producer ids, offsets

krishvishal:
`BRIDGE_MAPPING.md` is the last unchecked item in #3533 and is needed to design #3535 and #3536. The maintainer thread agreed on 16 September that each Kafka record should become one message in Iggy's native format. Four questions remain, with proposed defaults:

1. Use an envelope for records Iggy cannot represent, or reject them with `MESSAGE_TOO_LARGE`? Default: envelope.
2. Keep the `kafka.` prefix? `kafka.h.` uses 8 bytes of the 255-byte header-name limit. Default: keep it.
3. Use a placeholder byte for tombstones, even though native consumers would see a byte the producer never sent? Default: keep it.
4. Recompress on Fetch or return uncompressed records? Default: uncompressed.

`IDEMPOTENCE.md` (#3545) allocates producer IDs but ignores sequence fields because a stock Java producer treats unsupported `InitProducerId` as fatal instead of falling back. `OFFSET_STORAGE.md` (#3540) uses Iggy's native consumer offsets, one per partition—an option missing from the A/B/C list. The diff also corrects two inaccurate claims in `SCOPE.md`.


gateways/kafka/docs/BRIDGE_MAPPING.md
```diff
@@ -0,0 +1,204 @@
...
+## Field mapping
+
+Produce, per record:
+
+| Kafka | Iggy |
+| ------- | ------ |
+| record value | `payload` |
+| record key | `kafka.key` user header, `Raw` |
+| record header `name` | `kafka.h.<name>` user header, `Raw` |
+| record timestamp (CreateTime) | `origin_timestamp` |
```
ryerraguntla: While converting, make sure of the resolution, milli seconds in Kafka to microSeconds in Iggy

gateways/kafka/docs/BRIDGE_MAPPING.md
```diff
@@ -0,0 +1,204 @@
...
+| record offset | partition offset, assigned by Iggy |
+| partition index | partition index, both 0-based |
+| topic | stream and topic per `TopicMapping` |
+
+Fetch reverses it. A message with no `kafka.*` headers is a message an Iggy client wrote, and
+encodes as a Kafka record with a null key, its Iggy user headers as Kafka headers, and
+`origin_timestamp` as the record timestamp (falling back to the server-assigned `timestamp`
```
ryerraguntla, line 48: Refer earlier comment while decoding

gateways/kafka/docs/BRIDGE_MAPPING.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Everything else: the envelope fallback
+
+A record takes the fallback when any of these hold:
+
+- the key is longer than 255 bytes
+- a header name, prefixed with `kafka.h.`, is longer than 255 bytes
+- a header value is null, empty, or longer than 255 bytes
+- two headers share a name
+- the headers together would exceed the 100 KB user-header budget
```
ryerraguntla, line 99: Is 100Kb an optimal number or decided with any data point?

gateways/kafka/docs/BRIDGE_MAPPING.md
```diff
@@ -0,0 +1,204 @@
...
+
+Such a record is stored with a `kafka.envelope` header carrying the format version, and the
```
ryerraguntla: Could the format/schema for such messages be decided upfront with bytes count.

gateways/kafka/docs/BRIDGE_MAPPING.md
```diff
@@ -0,0 +1,204 @@
...
+  [#3545](https://github.com/apache/iggy/issues/3545) ever grows past a stub, producer id,
+  epoch and sequence need their own tracking.
+- The bytes a consumer receives are not the bytes the producer sent, so anything comparing
+  batches byte for byte across the gateway will differ.
+
+Whether producer id `-1` is what Fetch actually sends depends on the InitProducerId decision in
+[`IDEMPOTENCE.md`](IDEMPOTENCE.md). Allocating producer ids does not change what is stored, only
+what Produce accepts, so this section holds under either answer.
+
+Produce decompresses gzip, snappy, lz4 and zstd batches, which means turning those features back
```
ryerraguntla, line 151: How to handle decompressed message exceeds IGGY_KAFKA_MAX_FRAME_SIZE ?

gateways/kafka/docs/IDEMPOTENCE.md
```diff
@@ -0,0 +1,92 @@
...
+## Behavior
+
+Add key 22 to `SUPPORTED_RANGES` in `src/protocol/api.rs` and advertise it through ApiVersions.
+Without both, the producer never sends the request. `kafka-protocol` 0.18 carries the schemas,
+request v0 to v5 and response v0 to v6, flexible from v2.
+
+InitProducerId with no `transactional_id`:
+
+- allocate the next producer id, return it with epoch 0 and error code 0
+- draw ids from a counter seeded per gateway instance, so two instances never hand out the same
```
ryerraguntla: Good idea. Should it be a just a counter value or should it be a instance id (something similar) + counter?

gateways/kafka/docs/IDEMPOTENCE.md
```diff
@@ -0,0 +1,92 @@
...
+  id. Nothing reads the id today. Seeding it now is what stops a later deduplication layer from
+  being born broken
+
+InitProducerId with a `transactional_id`:
+
+- answer `UNSUPPORTED_VERSION` (35), unchanged. Transactions stay out of scope, and so do
```
ryerraguntla, line 133: Should it be 35 or Use a feature error (INVALID_REQUEST / transactions-not-supported)?

gateways/kafka/docs/OFFSET_STORAGE.md
```diff
@@ -0,0 +1,132 @@
...
+
+## What is stored
+
+The Kafka committed offset, verbatim, with no conversion.
+
+The two systems mean different things by the number. Kafka commits the next offset to read.
+Iggy stores the last offset processed, and `PollingKind::Next` resumes at the stored value plus
+one (`core/partitions/src/iggy_partition.rs:3835`). A Kafka offset stored in an Iggy key is
+therefore one greater than Iggy's own convention for that key.
+
```
ryerraguntla, line 61: How is Kafka s OffsetCommit offset -1 (forget commit) mapped in Iggy?

gateways/kafka/docs/OFFSET_STORAGE.md
```diff
@@ -0,0 +1,132 @@
...
+## What is dropped
+
+Kafka lets a client attach a metadata string to a commit. Iggy stores a number and nothing else.
+The string is dropped on commit, and OffsetFetch returns an empty string.
+
+`committed_leader_epoch` is dropped the same way. The gateway reports `-1`.
+
+## Limits
+
+A partition admits a bounded number of distinct offset keys per consumer kind. The default is
```
ryerraguntla, line 123: 4096 offset keys/partition/kind → Kafka -1 UNKNOWN_SERVER_ERROR. Multi-tenant group count will hit this with no useful client retry. Call the operator knob (partition.consumer_offsets_max) in README

ryerraguntla (commented):
Good documentation of mapping ,  assumption and gaps.

gateways/kafka/docs/BRIDGE_MAPPING.md
krishvishal: Kafka is milliseconds, `origin_timestamp` is microseconds (`iggy_message.rs:191`), so produce multiplies by 1000 and fetch divides. Now in the field mapping, along with the -1 case.

gateways/kafka/docs/BRIDGE_MAPPING.md
krishvishal: Same factor both ways. The only asymmetry is an iggy-written message losing sub-millisecond precision, since kafka has nowhere to put it.

gateways/kafka/docs/BRIDGE_MAPPING.md
krishvishal: Not our number, it is `MAX_USER_HEADERS_SIZE` (`iggy_message.rs:58`), fixed with no knob and no rationale recorded next to it. Reworded as inherited, and it works out to ~350 headers at the value cap, so the key-length rule is what actually trips.

gateways/kafka/docs/BRIDGE_MAPPING.md
krishvishal: `kafka_protocol` decodes to a `Record` and keeps no raw slice, so there was never a verbatim body to store. Layout is pinned in the doc now, 13 bytes fixed plus 9 per header.

gateways/kafka/docs/IDEMPOTENCE.md
krishvishal: 35 because INVALID_REQUEST is not in `maybeTransitionToErrorState`'s fatal set, so a transactional producer goes abortable and is told to retry something that can never work. COORDINATOR_NOT_AVAILABLE is worse, it is retriable.

gateways/kafka/docs/OFFSET_STORAGE.md
krishvishal: Kafka does not validate the sign, it stores -1 and hands it back. We cannot (u64), so -1 maps to `delete_consumer_offset` for the same observable result, with 3021 counted as success.

gateways/kafka/docs/OFFSET_STORAGE.md
krishvishal: Done, `partition.consumer_offsets_max` is in the README with the 4096 default and the 262144 ceiling. There is no kafka code for the condition, so it stays UNKNOWN_SERVER_ERROR with the real iggy error logged.

gkoszyk:
Storage direction is right, and the doc is unusually well grounded. I checked the Iggy-side claims against the tree and they hold, including the load-bearing one: a named consumer hashes to a `u32` and can collide, a named consumer group resolves through metadata to a monotonic id.

Two substantive things.

## Idempotence: point it at our dedup instead of allocating ids

We don't do Kafka-style idempotence and won't. What we do have, landed and on by default, is partition-plane dedup keyed on `(client_id, user_id, request)`, where `client_id` is a `u128` the client generates. The watermark is per consensus group, so per partition, with a 128-deep committed window below it. Sized by `[partition] dedup_clients_max`, 4096 per group, ceiling 65536. It covers retried produces explicitly.

That is close to Kafka's key shape. Producer id plus a per-partition sequence maps onto session id plus request id, per partition on both sides, and our window is wider than the 5 batches Kafka's producer keeps.

The catch is which hop it protects. Our dedup covers gateway to Iggy. Kafka idempotence covers producer to gateway. A producer retry arrives as a fresh Produce request, the gateway makes a fresh Iggy send with a fresh request id, and nothing absorbs it. To make this count, the gateway has to derive both halves of the key from the Kafka request: session id from the producer id, request id from the base sequence.

Two SDK gaps sit in the way. `ConsensusSession::with_client_id` exists and is `pub`, but `TcpClient` hardcodes `ConsensusSession::new()` (`tcp_client.rs:378`, `:583`), so nothing plumbs a chosen id through the builder. And `next_request_id` is an internal monotonic counter with no caller-supplied form. Both live in the same area as the open session-resume work.

What the mapping does and does not buy, so the doc can be honest about it:

- Duplicate suppression on retry: yes, and that is what `enable.idempotence=true` buys most users.
- The original offset back on an absorbed duplicate: no. A dedup hit answers the operation's empty success with no base offset, so the retry surfaces an unknown offset.
- Gap detection: no. The watermark accepts anything above it without noticing a hole, so never sending 45 or 46 stays correct.
- Fencing: no. Partition slices carry no epoch, so there is no producer-fenced equivalent.
- It degrades quietly. At `dedup_clients_max` the oldest client entry is evicted and its next replay re-executes, so the guarantee weakens as producer count per partition grows.

Your producer id layout gets more useful under this, not less. The producer id is what a session id would be derived from, and cross-instance uniqueness is what stops two gateways sharing a dedup key. Please also say whether that counter survives a gateway restart, since a counter that resets reissues ids and breaks the same layer the layout exists to protect.

Transactions stay out, permanently. The unsupported-version answer for a transactional id and the reasoning about the Java client's fatal set are both right. Keep them.

## Four things in the mapping doc

1. An empty record key has no path. Kafka separates a null key from a zero-length one, the key is stored as a header value, and an empty header value is rejected at `user_headers.rs:631`, the same line cited for the 255-byte cap. The fallback list covers a key that is too long but not an empty one, and the null-and-empty section covers values only. Such a record either fails to store or comes back with a null key.

2. The decompression bound is per batch where it has to be per request. One frame can carry many batches, bounded only by the 4096-element budget in `bounds_guard.rs`, so the real ceiling is that times `max_frame_size` rather than `max_frame_size`. This is the same product-across-nested-arrays shape `MAX_REQUEST_ELEMENTS` was added for, and that module's doc has the history.

3. Header order is not preserved and the doc does not say so. User headers live in a `BTreeMap` ordered by kind then bytes, so Fetch emits them sorted by name rather than in producer order. Duplicates are handled by the envelope, order is not mentioned anywhere.

4. OffsetFetch with a null topic list cannot reverse the topic mapping. `TopicMapping::resolve` is one way and an override can point at a different stream, so enumerating one stream both misses topics and hands back Iggy-side names where Kafka names are required.

## Process

#3533 is closed, so the mapping doc has no open issue to land under, and #3540 asks for sign-off on options A, B or C while this proposes a fourth. Worth reopening something or filing a docs issue so the epic checklist tracks these three files. I would drop the 22 September default-if-no-answer clause, since #3540's own acceptance criteria call for maintainer sign-off.

numinnex:
A quick primer on the status quo is needed here, because it changes what this document should aim at.

We never implemented Kafka's producer id and sequence machinery and do not intend to. What we do have, landed and on by default, is deduplication on the partition plane keyed on `(client_id, user_id, request)`, where the client id is a `u128` the client itself generates. Each partition group holds a per-client watermark with a 128-deep committed window below it, sized by `[partition] dedup_clients_max` at 4096 per group. Retried produces are what it exists to absorb, and the shape is nearer Kafka's than it looks: a producer id with a per-partition sequence maps onto a session id with a request id, both per partition, and our window is deeper than the five batches a Kafka producer keeps in flight.

What breaks the mapping is which hop each side protects. Ours covers gateway to Iggy, Kafka's covers producer to gateway. A retrying producer sends a fresh Produce request, the gateway turns it into a fresh Iggy send carrying a fresh request id, and nothing recognises the replay. Closing that means deriving both halves of our key from the Kafka request instead of letting the SDK mint them, session id from the producer id and request id from the base sequence. `ConsensusSession::with_client_id` is public, but `TcpClient` builds its session at `tcp_client.rs:378` and `:583` with no way to pass one in, and `next_request_id` has no caller-supplied form, so the SDK needs work before the gateway can lean on any of this.

Worth stating plainly what it buys, because it is not the whole contract. Retries stop duplicating, but an absorbed duplicate answers empty success with no base offset, so the retry reports an unknown one, and the watermark accepts anything above it without noticing a gap, which is why never sending 45 or 46 stays right.

That makes your producer id layout more load-bearing rather than less, since the id is what a session key would derive from. Does the counter survive a restart? Transactions stay out permanently, and the unsupported-version answer is the one to keep.

Four smaller things in the mapping doc: an empty record key has no path, since a zero-length key is not a null one and an empty header value is rejected at the same line cited for the 255-byte cap. The decompression bound is per batch where it needs to be per request, as one frame carries up to the 4096-element budget's worth of batches. Header order is lost to the `BTreeMap` and the doc never says so. OffsetFetch with a null topic list cannot reverse `TopicMapping::resolve`.

numinnex (approved):

hubcio (approved):

PR #4216, opened 2026-09-17
fix(ci): retry flaky downloads and stop dry runs promising new tags

hubcio:
A single transient curl error failed a whole CI job. One-shot
downloads now use --retry 5 --retry-all-errors, with -f so an
HTTP 5xx retries instead of being saved as the script or binary.
Downloads piped into tar or gunzip stage to a temp file first,
because curl cannot rewind a pipe on retry. Polling loops keep
their own backoff and are unchanged.

Dry runs and skip_tag_creation runs listed tags as "Will create"
although neither pushes a tag. They now say "Would create".

The source release tarball now carries rust-toolchain.toml, so a
build from it uses the pinned toolchain.

Also quote the SC2086 sites that do not rely on word splitting.


hubcio:
don't merge this until 0.9.0 is released

krishvishal (approved):

spetz (approved):

PR #4217, opened 2026-09-17
fix(server): note root password reset in generated password log

### LINKED ISSUE #4211: Say how to reset the root password in the generated password log message
### Description

When no root credentials are set, the server generates a root password on first start and logs it once:

```
Generated root user password: <password>
```

If the operator misses that line, there's no way to read the password again. The only recovery is to start over with an empty data directory, and only the README says so. The log message should say it.

Split out of #4209.

### Affected area / component

Iggy server

### Proposed solution

In `create_root_credentials()` in `core/server/src/boot/credentials.rs`, extend the `warn!` that logs the generated password. Say that it is shown only once, and how to reset it on a single-node server, for example:

```
Generated root user password: <password>. It is shown only once. To reset it, stop the server and delete the data directory, or start the server with --fresh. Both delete all stored data.
```

Done looks like:

- The message says the password is shown once and how to reset it.
- `cargo test -p server` passes. Check the tests at the end of `credentials.rs` still pass, and update any that match the old message.

### Alternatives considered

Log the reset hint as a separate line. One message is easier to spot and keeps the hint next to the password.

### Contribution

### Good first issue

de-authority:

Closes #4211

## What changed?
The generated root password is only logged once, and the message did not say so or explain how to recover. Extend the warning to state that the password is shown only once and that resetting it requires stopping the server and deleting the data directory, or starting with --fresh.

## Local Execution

- Passed 
- Pre-commit hooks not ran

## AI Usage

1. Which tools? (e.g., GitHub Copilot, Claude, ChatGPT)
Deepseek/gemini
To verify the implementation.
By running test locally, 
Yes


justinmclean:
Thanks @de-authority, and thanks for asking on #4211 before starting. The new message matches what the issue asked for. The CI hasn't run yet because workflows from first-time contributors need a maintainer to approve them. @hubcio, could you approve the run?

hubcio (approved):

mmodzelewski (approved):

